PR #1, opened 2022-03-04
initial Client stub and unit tests

chrismaree:


src/SpokePoolEventClient.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (searchConfig[0] > searchConfig[1]) return;
+
+    const [depositEvents, speedUpEvents, fillEvents] = await Promise.all([
+      await this.spokePool.queryFilter(this.spokePool.filters.FundsDeposited(), ...searchConfig),
+      await this.spokePool.queryFilter(this.spokePool.filters.RequestedSpeedUpDeposit(), ...searchConfig),
+      await this.spokePool.queryFilter(this.spokePool.filters.FilledRelay(), ...searchConfig),
+    ]);
+
+    for (const event of depositEvents) {
+      const augmentedEvent: Deposit = { ...spreadEvent(event), originChainId: this.chainId };
```
chrismaree, line 53: notice how clean this is...the `spreadEvent` is doing all the heavy lifting that in All our previous clients we would have needed to do something like this:
```ts
const relayData: Relay = {
          relayId: Number(depositRelayedEvent.returnValues.relay.relayId),
          chainId: Number(depositRelayedEvent.returnValues.depositData.chainId),
          depositId: Number(depositRelayedEvent.returnValues.depositData.depositId),
          l2Sender: depositRelayedEvent.returnValues.depositData.l2Sender,
          slowRelayer: depositRelayedEvent.returnValues.relay.slowRelayer,
          l1Recipient: depositRelayedEvent.returnValues.depositData.l1Recipient,
          l1Token: l1Token,
          amount: depositRelayedEvent.returnValues.depositData.amount,
          slowRelayFeePct: depositRelayedEvent.returnValues.depositData.slowRelayFeePct,
          instantRelayFeePct: depositRelayedEvent.returnValues.depositData.instantRelayFeePct,
          quoteTimestamp: Number(depositRelayedEvent.returnValues.depositData.quoteTimestamp),
          realizedLpFeePct: depositRelayedEvent.returnValues.relay.realizedLpFeePct,
          priceRequestTime: Number(depositRelayedEvent.returnValues.relay.priceRequestTime),
          depositHash: depositRelayedEvent.returnValues.depositHash,
          relayState: ClientRelayState.Pending, // Should be equal to depositRelayedEvent.returnValues.relay.relayState
          relayAncillaryDataHash: depositRelayedEvent.returnValues.relayAncillaryDataHash,
          proposerBond: depositRelayedEvent.returnValues.relay.proposerBond,
          finalFee: depositRelayedEvent.returnValues.relay.finalFee,
          settleable: SettleableRelay.CannotSettle,
          blockNumber: depositRelayedEvent.blockNumber,
        };
```

src/SpokePoolEventClient.ts
```diff
@@ -0,0 +1,69 @@
...
+      assign(this.deposits, [augmentedEvent.destinationChainId], [augmentedEvent]);
```
chrismaree, line 54: assign is a helper that I spent some time building that enables you to deep insert objects into the nested structure without needing to see if they are initialized. it is generalized to an n depth object. it's not super useful here (you could achieve the same thing in 1 line) but in the future, it'll be great when populating fresh, deep structures.

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,17 @@
+import { Deposit } from "../interfaces/SpokePool";
+
+import { Event } from "ethers";
+
+export function spreadEvent(event: Event): Deposit {
```
chrismaree, line 5: this is the function that extracts types from the events. basically, remove all non-string keys( remove things that are `isNan(+key)` true.

src/utils/ProviderUtils.ts
```diff
@@ -0,0 +1,7 @@
+import { ethers } from "ethers";
+
+export function getProvider(networkId: number) {
```
chrismaree, line 3: for now, here is the minimal provider and signer utils. this will be extended to include re-try logic once we've built out that part of the bot.

src/utils/index.ts
```diff
@@ -0,0 +1,6 @@
+export { BigNumber, Signer, Contract, ContractFactory } from "ethers";
```
chrismaree, line 1: a lot of this PR, as you can see, is setting up this kind of plumbing. we now have a clean structure for using these kinds of utils through the relayer and soon to be data worker.

test/SpokePoolEventClient.deposits.ts
```diff
@@ -0,0 +1,76 @@
+import { expect, ethers, Contract, SignerWithAddress } from "across-contracts";
+import { spokePoolFixture, deposit, enableRoutes, destinationChainId } from "across-contracts";
```
chrismaree, line 2: this is NB NB! note that I am importing logic from the `across-contracts` package! this means we can re-use the fixtures and helper functions! tests like this before would have been hundreds of lines and now are short, clean and simple.

package.json
```diff
@@ -8,7 +8,10 @@
   "license": "AGPL-3.0",
   "private": true,
   "dependencies": {
+    "@defi-wonderland/smock": "^2.0.7",
+    "@uma/contracts-node": "^0.3.1",
     "@uma/financial-templates-lib": "2.25.0",
+    "across-contracts": "npm:@across-protocol/contracts-v2@^0.0.17",
```
chrismaree, line 14: notice that this has been added AND alisted. this pulls in all the contract artifacts and helpers used throughout the tests.

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  let returnedObject: any = {};
+  keys.forEach((key: string) => (returnedObject[key] = event.args[key]));
+
+  // ChainID information, if included in an event, should be cast to a number rather than a BigNumber.
+  if (returnedObject.destinationChainId) returnedObject.destinationChainId = Number(returnedObject.destinationChainId);
+  if (returnedObject.originChainId) returnedObject.originChainId = Number(returnedObject.originChainId);
+  if (returnedObject.repaymentChainId) returnedObject.repaymentChainId = Number(returnedObject.repaymentChainId);
+
+  return returnedObject as Deposit;
```
mrice32, line 16: This only works for a deposit? Should we make this function generic?
```
spreadEvent<EventType>(....)
```

src/interfaces/SpokePool.ts
```diff
@@ -0,0 +1,51 @@
+import { BigNumber } from "ethers";
+
+export interface Deposit {
```
mrice32, line 3: Can we pull these definitions from typechain?

src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export async function run(logger: winston.Logger): Promise<void> {
+  const config = new RelayerConfig(process.env);
+  logger.debug({ at: "Relayer#index", message: "Relayer starting🚀", config });
+
+  const provider = getProvider(1);
+  console.log("provider", provider);
+
+  const signer = getSigner();
+  console.log("signer", signer);
```
mrice32, line 17: Why are we printing these objects? Should we be printing the url and address instead of the entire objects? Especially to avoid leaking private keys accidentally.

mrice32 (approved):
LGTM! A few minor comments, but overall this is great!

src/index.ts
chrismaree: this was just to show that they are initialised correctly. they dont do anything (yet).

src/interfaces/SpokePool.ts
chrismaree: likely, yes! will look into do this.

src/utils/EventUtils.ts
chrismaree: o ye, good catch. should not be only `Deposit` type!

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,17 @@
+import { Deposit } from "../interfaces/SpokePool";
+
+import { Event } from "ethers";
+
+export function spreadEvent(event: Event): Deposit {
+  const keys = Object.keys(event.args).filter((key: string) => isNaN(+key)); // Extract non-numeric keys.
```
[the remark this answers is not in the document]
chrismaree: ye, I think this is ok. no situation where these keys would end up in this function.

src/SpokePoolEventClient.ts
```diff
@@ -0,0 +1,69 @@
...
+      const augmentedEvent: Deposit = { ...spreadEvent(event), originChainId: this.chainId };
+      assign(this.speedUps, [augmentedEvent.destinationChainId], [augmentedEvent]);
+    }
+
+    for (const event of fillEvents) {
+      const augmentedEvent: Deposit = { ...spreadEvent(event), destinationChainId: this.chainId };
+      assign(this.fills, [augmentedEvent.destinationChainId], [augmentedEvent]);
+    }
+
+    console.log("this.f1ills", this.fills);
```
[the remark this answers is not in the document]
chrismaree: removed. was just there to test some of the `assign` logic.

src/utils/ObjectUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+  const lastKeyIndex = keyPath.length - 1;
+  for (var i = 0; i < lastKeyIndex; ++i) {
+    const key = keyPath[i];
+    if (!(key in obj)) obj[key] = {};
+    obj = obj[key];
+  }
+  // If the object at the deep path does not exist then set to the value.
+  if (!obj[keyPath[lastKeyIndex]]) obj[keyPath[lastKeyIndex]] = value;
+  // If the object at the deep path is an array then append array wise.
+  else if (Array.isArray(value)) obj[keyPath[lastKeyIndex]] = [...obj[keyPath[lastKeyIndex]], ...value];
```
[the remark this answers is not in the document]
chrismaree: ye... this is a great point. I tested it and it behaves correctly when working with bignumbers but I'm not convinced that in all cases this will work as expected. we should keep an eye on this.

note that the implementation here is somewhat underutilized vs what it can do. it's really just used to assign single level keys but is really designed to go much deeper than that.

PR #3, opened 2022-03-11
feat: Add intial relayer unfilled deposits filtering

chrismaree:
Adds initial relayer fill implementation.


package.json
```diff
@@ -11,7 +11,7 @@
     "@defi-wonderland/smock": "^2.0.7",
     "@uma/contracts-node": "^0.3.1",
     "@uma/financial-templates-lib": "2.25.0",
-    "across-contracts": "npm:@across-protocol/contracts-v2@^0.0.17",
+    "@across-protocol/contracts-v2":"^0.0.25",
```
chrismaree, line 14: I removed the alias. it was a bit confusing.

src/MulticallBundler.ts
```diff
@@ -0,0 +1,5 @@
+import winston from "winston";
+
+export class MulticallBundler {
```
chrismaree, line 3: stub

test/Relayer.UnfilledDeposits.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Relayer } from "../src/Relayer";
+
+let spokePool_1: Contract, erc20_1: Contract, spokePool_2: Contract, erc20_2: Contract;
+let depositor_1: SignerWithAddress, depositor_2: SignerWithAddress, relayer_1: SignerWithAddress;
+
+let spy: sinon.SinonSpy, spyLogger: winston.Logger;
+let spokePoolClient_1: SpokePoolEventClient, spokePoolClient_2: SpokePoolEventClient;
+let relayer: Relayer;
+
+describe("Relayer Unfilled Deposits", async function () {
```
chrismaree, line 14: see these test cases on how this method works and how it can be used to find the unfilled deposits.

src/Relayer.ts
```diff
@@ -0,0 +1,36 @@
...
+      for (const destinationChain of chainIds) {
+        if (originChain === destinationChain) continue;
+        const destinationClient = this.spokePoolEventClients[destinationChain];
+        const depositsForDestinationChain = originClient.getDepositsForDestinationChain(destinationChain);
+        unfilledDeposits.push(
+          ...depositsForDestinationChain // append all deposits for the current destination chain that are unfilled.
+            .map((deposit) => {
+              return { unfilledAmount: destinationClient.getUnfilledAmountForDeposit(deposit), deposit };
+            }) // remove any deposits that have no unfilled amount (i.e that are 100% filled).
+            .filter((deposit) => deposit.unfilledAmount.gt(0))
```
nicholaspai, line 30: Should we sort these deposits by size or just keep them ordered by default, which is I assume by timestamp?

chrismaree: it is right now ordered by timestamp. we can change this to size but perhaps leave this for now and get the simplest posible relayer up and running first.

src/SpokePoolEventClient.ts
```diff
@@ -39,9 +43,18 @@ export class SpokePoolEventClient {
       .filter((fill: Fill) => fill.relayer === relayer);
   }
 
+  getUnfilledAmountForDeposit(deposit: Deposit) {
+    const fills = this.getFillsForOriginChain(deposit.originChainId).filter(
```
nicholaspai, line 47: Where do you check whether the `fills` are valid? You could imagine counting some fills where the `originChainId` matches but the fill is invalid for some other reason

nicholaspai (approved):
Great start lgtm

src/SpokePoolEventClient.ts
chrismaree: great call out! will work on this next.

PR #5, opened 2022-03-22
Add new interface support and realized LPFeePct computation

chrismaree:
This PR updates the relayer to work with the latest contracts, adds realizedLpFeePct computation to relay logic (hence adds the rateModelStore) and adds additional tests & helpers.


src/HubPoolEventClient.ts
```diff
@@ -1,31 +1,76 @@
...
+    return this.getDestinationTokenForL1TokenAndDestinationChainId(l1Token, deposit.destinationChainId);
+  }
+
+  getL1TokensToDestinationTokens() {
+    return this.l1TokensToDestinationTokens;
   }
 
-  getWhitelistedRoutes() {
-    return this.whitelistedRoutes;
+  getL1TokenForDeposit(deposit: Deposit) {
```
chrismaree, line 63: this bit of logic is needed given the latest changes to the contract's whitelist structure.

test/HubPoolEventClient.DepositToDestinationToken.ts
```diff
@@ -0,0 +1,61 @@
+import { getContractFactory, expect, ethers, Contract, SignerWithAddress, originChainId } from "./utils";
```
chrismaree, line 1: replaces the previous set of whitelisting tests. works the same way with the new structure.

test/Relayer.ItterativeFill.ts
```diff
@@ -0,0 +1,88 @@
+import { hubPoolFixture, deployIterativeSpokePoolsAndToken, createSpyLogger, lastSpyLogIncludes } from "./utils";
```
chrismaree, line 1: iterative test that mocks out 5 spoke pools (mimicking 5 chains) with one token on each. I will add more iterative tests as we go along.

test/utils.ts
```diff
@@ -31,13 +35,154 @@ export function createSpyLogger() {
...
+
+  if (enableRoute)
+    await utils.enableRoutes(spokePool, [
+      { originToken: erc20.address, destinationChainId: toChainId == 0 ? utils.destinationChainId : toChainId },
+      { originToken: weth.address, destinationChainId: toChainId == 0 ? utils.destinationChainId : toChainId },
+    ]);
   return { timer, weth, erc20, spokePool, unwhitelistedErc20, destErc20 };
 }
+
+export async function deployRateModelStore(signer: utils.SignerWithAddress, tokensToAdd: utils.Contract[]) {
```
chrismaree, line 55: more helpers! rate model & iterative test helpers.

src/HubPoolEventClient.ts
```diff
@@ -1,31 +1,76 @@
...
+    let l1Token = null;
+    Object.keys(this.l1TokensToDestinationTokens).forEach((key) => {
```
nicholaspai, line 65: nit: rename `key` to `l1Token` to make it more clear what we're looking up

nicholaspai (approved):
LGTM logic looks good, just left some nits

src/HubPoolEventClient.ts
```diff
@@ -36,20 +81,26 @@ export class HubPoolEventClient {
   async update() {
     const searchConfig = [this.firstBlockToSearch, this.endingBlock || (await this.getBlockNumber())];
     if (searchConfig[0] > searchConfig[1]) return; // If the starting block is greater than the ending block return.
-    const [whitelistRouteEvents] = await Promise.all([
-      await this.hubPool.queryFilter(this.hubPool.filters.WhitelistRoute()),
+    const [PoolRebalanceRouteEvents, rateModelStoreEvents] = await Promise.all([
```
mrice32, line 84:
```suggestion
    const [poolRebalanceRouteEvents, rateModelStoreEvents] = await Promise.all([
```

src/utils/ObjectUtils.ts
```diff
@@ -9,7 +9,7 @@ export function assign(obj: any, keyPath: any[], value: any) {
     obj = obj[key];
   }
   // If the object at the deep path does not exist then set to the value.
-  if (!obj[keyPath[lastKeyIndex]]) obj[keyPath[lastKeyIndex]] = value;
+  if (!obj[keyPath[lastKeyIndex]] || typeof obj[keyPath[lastKeyIndex]] == "string") obj[keyPath[lastKeyIndex]] = value;
```
mrice32, line 12: Can you explain this change?

mrice32 (approved):
LGTM!

src/utils/ObjectUtils.ts
chrismaree: if the object at the end of the assignment path is a string then the previous implementation would spread the string into an array of char for each element. for example if you had `{key1:{key2:0x1234}}` after trying to `assign` the address `0x3456` to `key1.key2` this would return `{key1:{key2:[0,x,1,2,3,4,5,6]}}` whereas you want to replace the element at `key1.key2`

PR #6, opened 2022-03-22
feat: refactor clients and spokePool Validate fill

nicholaspai:
This PR does a number of things:
- Introduces a `RateModelClient`
- re-names all clients to remove the word `Event` from the file and instance.
- Adds the ability for the `SpokePoolClient` to asynchronously fetch realizedLpFeePct along with each deposit.
	- This is the key change of the PR as it drastically decreases upstream logic.
	- It does have some implications, though. Both the `RateModelClient` and `HubPoolClient` need to be updated first before the first run of the `SpokePoolClient`. This is ok though as in production the intention is to run this bot in a "long running" format wherein we store state for an extended period of time.


src/clients/RateModelClient.ts
```diff
@@ -0,0 +1,70 @@
+import { spreadEvent, assign, Contract, toBNWei, Block, BigNumber, toBN, utils } from "../utils";
+import { Deposit, Fill } from "../interfaces/SpokePool";
+import { lpFeeCalculator } from "@across-protocol/sdk-v2";
+import { BlockFinder, across } from "@uma/sdk";
+import { HubPoolClient } from "./HubPoolClient";
+
+export class RateModelClient {
```
chrismaree, line 7: this is new! it now stores the rate model information and events.

src/clients/SpokePoolClient.ts
```diff
@@ -77,25 +77,50 @@ export class SpokePoolEventClient {
...
     ]);
 
-    for (const event of depositEvents) {
-      const augmentedEvent: Deposit = { ...spreadEvent(event), originChainId: this.chainId };
-      assign(this.deposits, [augmentedEvent.destinationChainId], [augmentedEvent]);
+    // For each depositEvent, compute the realizedLpFeePct. Note this means that we are only finding this value on the
+    // new deposits that were found in the searchConfig (new from the previous run). This is important as this operation
+    // is heavy as there is a fair bit of block number lookups that need to happen. Note this call REQUIRES that the
+    // hubPoolClient is updated on the first before this call as this needed the the L1 token mapping to each L2 token.
+    const realizedLpFeePcts = await Promise.all(depositEvents.map((event) => this.computeRealizedLpFeePct(event)));
```
chrismaree, line 84: there is a really important design decision that I made here: when calling `update` on the `SpokePoolClient` the client will pull in the `realizedLpFeePct` for _each_ deposit, from the last time this method was called. This is vital as it means that `RateModelClient` can actually verify _all_ props of the deposit, without needing to do any extra async calls!

src/Relayer.ts
```diff
@@ -23,50 +24,41 @@ export class Relayer {
       this.logger.debug({ at: "Relayer", message: "Filling deposits", number: unfilledDeposits.length });
     else this.logger.debug({ at: "Relayer", message: "No unfilled deposits" });
 
-    // Fetch the realizedLpFeePct for each unfilled deposit. Execute this in parallel as each requires an async call.
-    const realizedLpFeePcts = await Promise.all(
```
chrismaree, line 27: this is the key part of the change from the PR: the `Relayer` no longer needs to do any extra logic for the `realizedLpFeePct`. it is all from the client and can be synchronously fetched due to it being fetched when `SpokePoolClient` is `update`ed.

chrismaree (approved):

PR #7, opened 2022-03-23
feat: Add Dataworker skeleton

nicholaspai:
This PR creates the initial skeleton for the data worker without function implementations or tests. I'd like to reach consensus on this structure before proceeding to implementations.


src/Dataworker.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  async buildRelayerRefundRoot(blockRange: BlockRange) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Group remaining fills by destination chain ID
+    // Construct leaf for destination chain ID
+    // Construct root
+  }
```
chrismaree, line 38: again comment on refund. order by relayer refund size.

src/Dataworker.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  async buildRelayerRefundRoot(blockRange: BlockRange) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Group remaining fills by destination chain ID
```
chrismaree, line 35: there is an additional grouping you need to do: you need to also group by Relayer. there is one Relayer refund leaf per relayer per l2 token per chain. see required data structure:
```solidity
struct RelayerRefundLeaf {
        // This is the amount to return to the HubPool. This occurs when there is a PoolRebalanceLeaf netSendAmount that
        uint256 amountToReturn;
        uint256 chainId;
        uint256[] refundAmounts;
        uint32 leafId;
        address l2TokenAddress;
        address[] refundAddresses; // you need to group by the relayer address to populate this field.
    }
```

src/Dataworker.ts
```diff
@@ -0,0 +1,53 @@
+import { BigNumber, winston, buildFillRelayProps } from "./utils";
+import { SpokePoolEventClient } from "./SpokePoolEventClient";
+import { HubPoolEventClient } from "./HubPoolEventClient";
+import { MulticallBundler } from "./MulticallBundler";
+import { Deposit } from "./interfaces/SpokePool";
+
+export interface BlockRange {
```
chrismaree, line 7: nice. put this in the `interfaces` file. we should use it in other places as well.

src/Dataworker.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Group remaining fills by destination chain ID
+    // For each destination chain ID:
+    // Group fills by L1 token
+    // For each L1 token:
+    // Sum total relay refund amount ==> netSendAmount
+    // Sum total realized LP fee % ==> bundleLpFee
+    // Construct leaf for destination chain ID with fills grouped by L1 token
```
chrismaree, line 50: you also need to add in the `groupIndex` this will be 0 for now but will expand later when we add a module for this. we should think about how this additional filtering will be applied. it could be within this module or perhaps an external one. this will be linked to the settings defined in the umip.

chrismaree: for now just add some indication of `groupIndex`

chrismaree (approved):
This is a nice way of showing the intention of future code to write and will help parallelise the process. 

each of these methods should be implemented by a separate module that connects to the DataWorker.ts file. I also have some ideas of re-structuring the repo but this will be easier to show once this PR is in.

src/Dataworker.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  async buildSlowRelayRoot(blockRange: BlockRange) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Filter out fills that fully fill a deposit
+    // Group remaining fills by destination chain ID
+    // Construct leaf for destination chain ID
+    // Construct root
+  }
```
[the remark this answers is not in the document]
nicholaspai: Should this be ordered by amount of the fill remaining or just deposit size?

src/Dataworker.ts
nicholaspai: Added to `interfaces/HubPool.ts`. Also changed this to be 

`type BundleEvaluationBlockNumbers = number[]` since we might not want all block #'s within a range

src/Dataworker.ts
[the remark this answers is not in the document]
chrismaree: good point. remaining fill makes the most sence.

src/Dataworker.ts
chrismaree: perfect. we should look at pulling in these from typechain

src/Dataworker.ts
```diff
@@ -0,0 +1,56 @@
...
+    readonly multicallBundler: MulticallBundler | any
+  ) {}
+
+  async buildSlowRelayRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Filter out fills that fully fill a deposit
+    // Group remaining fills by destination chain ID
+    // Order fills by remaining fill s ize
```
mrice32, line 24:
```suggestion
    // Order fills by remaining fill size
```

src/Dataworker.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Construct leaf for destination chain ID
+    // Construct root
+  }
+
+  async buildRelayerRefundRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
```
mrice32, line 32: Looks like we're validating fills in two places. Could make sense to either a) find all valid fills within the block range _outside_ of these two methods or b) use a common method/code to do it and possibly cache the result so it doesn't require repeated queries.

src/Dataworker.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  async buildPoolRebalanceRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
+    // Grab all deposits from SpokePoolEventClient
+    // Grab all fills from SpokePoolEventClient
+    // Filter out invalid fills for deposits
+    // Group remaining fills by destination chain ID
+    // For each destination chain ID:
+    // Group fills by L1 token
+    // For each L1 token:
+    // Sum total relay refund amount ==> netSendAmount
```
mrice32, line 48: Do we also need to add the slow relays?

mrice32 (approved):
Looks great! Really like the idea of adding comments to show the flow that's expected.

src/Dataworker.ts
nicholaspai: Great point, adding to an update() skeleton

src/Dataworker.ts
nicholaspai: oh true...

nicholaspai: right good point

PR #11, opened 2022-03-25
feat: Add RateModelClient tests

nicholaspai:
Client missing unit tests


test/RateModelClient.ts
```diff
@@ -1 +1,136 @@
...
+                relayerFeePct: toBN(0),
+                quoteTimestamp: initialRateModelUpdateTime,
+                // Quote time needs to be >= first rate model event time
+              };
+            await rateModelClient.update();
+
+            // TODO: Validate that this rate is expected
+            expect(
+              await rateModelClient.computeRealizedLpFeePct(depositData, l1Token.address)
+            ).to.equal(toBNWei("0.000835322155994299"));
```
nicholaspai, line 103: Is this an expected rate for a deposit that is 10% of the hub pool liquidity? Would this be different from sending away 10% of the liquidity via a pool rebalance? I notice also that this is a bit diff from v1's expected fee at 10% liquidity, but I understanad the contract calculation is a bit different: https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js#L1037

src/clients/RateModelClient.ts
```diff
@@ -49,6 +49,8 @@ export class RateModelClient {
   }
 
   async update() {
+    if (!this.hubPoolClient.isUpdated()) throw new Error("hubpool not updated");
```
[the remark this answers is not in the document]
chrismaree: I think it's good. there is a core assumption in both this client and in the spoke client that the dependent upstream clients have been updated.

test/RateModelClient.ts
```diff
@@ -1 +1,136 @@
...
+        await rateModelClient.update();
+        expect(rateModelClient.cumulativeRateModelEvents.length).to.equal(0);
+
+        // Add new RateModelEvents and check that updating again pulls in new events.
+        await rateModelStore.updateRateModel(l1Token.address, JSON.stringify(sampleRateModel));
+        await rateModelClient.update();
+        expect(rateModelClient.cumulativeRateModelEvents.length).to.equal(1);
+    })
+
+    describe("RateModelStore contract has 1 initial update", function() {
```
chrismaree, line 46: I've avoided having nested `describes` in all the other tests to try keep them as flat as posible. can you remove this to be consistant?

test/RateModelClient.ts
```diff
@@ -1 +1,182 @@
...
+      rateModelClient.getRateModelForBlockNumber(l1Token.address, initialRateModelUpdate.blockNumber)
+    ).to.deep.equal(sampleRateModel);
+
+    // Block number when there is no rate model
+    try {
+      rateModelClient.getRateModelForBlockNumber(l1Token.address, initialRateModelUpdate.blockNumber - 1);
+      assert(false);
+    } catch (err) {
+      assert.isTrue(err.message.includes("before first UpdatedRateModel event"));
+    }
```
chrismaree, line 88: can we not use the chai should throw: https://www.chaijs.com/api/bdd/#method_throw

test/Relayer.BasicFill.ts
```diff
@@ -101,10 +101,8 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
 });
 
 async function updateAllClients() {
-  await Promise.all([
```
chrismaree, line 104: you did this because the hub must be updated before the ratemode/spoke, right?

chrismaree (approved):

test/RateModelClient.ts
nicholaspai: Good point, done

test/Relayer.BasicFill.ts
nicholaspai: exactly

PR #12, opened 2022-03-28
feat: Factor in speedup in relays, add more general entry point and add docker file

chrismaree:
This PR updates the following:

- `SpokePoolClient` to consider speedups when dealing with relays.
- adds a more general entry point to actually run the relayer as well as separating out the entry points between the relayer and data worker.
- adds the implementation of the relayer `index.ts` file
- adds a `ClientHelper` to build all clients needed for all bot types.
- adds longer to all clients.
- adds better error handling
- adds a Dockerfile to enable us to start setting up the production environment for the bot.


src/clients/HubPoolClient.ts
```diff
@@ -8,14 +8,18 @@ export class HubPoolClient {
...
     readonly hubPool: Contract,
     readonly startingBlock: number = 0,
     readonly endingBlock: number | null = null
   ) {}
 
   getDestinationTokenForDeposit(deposit: Deposit) {
     const l1Token = this.getL1TokenForDeposit(deposit);
-    return this.getDestinationTokenForL1TokenAndDestinationChainId(l1Token, deposit.destinationChainId);
+    const destinationToken = this.getDestinationTokenForL1TokenDestinationChainId(l1Token, deposit.destinationChainId);
+    if (!destinationToken)
```
chrismaree, line 20: when testing against Kovan this was not defined. Added better error logging to indicate this situation.

src/clients/RateModelClient.ts
```diff
@@ -12,29 +12,38 @@ export class RateModelClient {
...
+    readonly hubPoolClient: HubPoolClient
+  ) {
+    this.blockFinder = new BlockFinder(this.rateModelStore.provider.getBlock.bind(this.rateModelStore.provider));
     this.rateModelDictionary = new across.rateModel.RateModelDictionary();
   }
 
   async computeRealizedLpFeePct(deposit: Deposit, l1Token: string) {
-    const quoteBlockNumber = (await this.blockFinder.getBlockForTimestamp(deposit.quoteTimestamp)).number;
+    this.logger.debug({ at: "RateModelClient", message: "Computing realizedLPFeePct", deposit, l1Token });
+    // const quoteBlockNumber = (await this.blockFinder.getBlockForTimestamp(deposit.quoteTimestamp)).number;
```
chrismaree, line 26: temp work around util we re-deploy these contracts. method does work correctly if the deployment order is correct.

src/clients/SpokePoolClient.ts
```diff
@@ -81,22 +96,32 @@ export class SpokePoolClient {
...
     }
 
     for (const event of speedUpEvents) {
       const speedUp: SpeedUp = { ...spreadEvent(event), originChainId: this.chainId };
-      // assign(this.speedUps, [speedUp.destinationChainId], [speedUp]); todo: add associated reverse lookup for this.
+      assign(this.speedUps, [speedUp.depositor, speedUp.depositId], [speedUp]);
     }
 
+    // Traverse all deposit events and update them with associated speedups, If they exist.
+    for (const destinationChainId of Object.keys(this.deposits))
```
chrismaree, line 114: speedup logic.

src/clients/SpokePoolClient.ts
```diff
@@ -45,6 +46,16 @@ export class SpokePoolClient {
       .filter((fill: Fill) => fill.relayer === relayer);
   }
 
+  appendMaxSpeedUpSignatureToDeposit(deposit: Deposit) {
```
chrismaree, line 49: selects highest speedup setting and adds it.

src/relayer/index.ts
```diff
@@ -0,0 +1,29 @@
+import { config } from "dotenv";
```
chrismaree, line 1: look how short and clean this is. this is the entire relayer entry point. still need to add more error handling and potentially re-try logic but going to ignore this for now.

index.ts
```diff
@@ -0,0 +1,20 @@
+import { Logger } from "@uma/financial-templates-lib";
+
+import { runRelayer } from "./src/relayer";
+
+export async function run(): Promise<void> {
+  if (process.argv.includes("--relayer")) await runRelayer(Logger);
+  else if (process.argv.includes("--dataworker")) console.log("NOT YET IMPLEMENTED");
+  else console.log("Select either relayer OR dataworker");
```
nicholaspai, line 8: You can't run both?

src/MulticallBundler.ts
```diff
@@ -1,11 +1,14 @@
...
+export class MultiCallBundler {
   private transactions: Transaction[] = [];
   constructor(readonly logger: winston.Logger, readonly gasEstimator: any) {}
 
+  // Adds defined transaction to the transaction queue.
   addTransaction(transaction: Transaction) {
-    this.transactions.push(transaction);
+    console.log("ADDING TX", transaction);
+    if (transaction) this.transactions.push(transaction);
+    console.log("this.transactions", this.transactions);
```
nicholaspai, line 11: Some leftover logs

src/clients/HubPoolClient.ts
nicholaspai: Should we instead make `getDestinationTokenForL1TokenDestinationChainId` throw an error if destination token can't be found?

index.ts
chrismaree: in my current design I think they should be run separately.

PR #15, opened 2022-03-30
feat(dataworker): Add function to load and format data used in all functions

nicholaspai:
All data worker functions will need some combination of:
- unfilled deposits grouped by destination chain
    - Will be used to create `SlowRelayRoots` and `PoolRebalanceRoots`
- validated fill data grouped by repayment chain
    - Will be used to create `RelayerRefundRoots` and `PoolRebalanceRoots`


src/clients/SpokePoolClient.ts
```diff
@@ -30,7 +31,11 @@ export class SpokePoolClient {
       .filter((deposit: Deposit) => deposit.depositor === depositor); // Select only deposits where the depositor is the same.
   }
 
-  getFillsForDestinationChain(destinationChainId: number) {
+  getFills(): Fill[] {
+    return Object.values(this.fills).flat();
+  }
+
+  getFillsForDestinationChain(destinationChainId: number): Fill[] {
```
nicholaspai, line 38: I added `getFills` because I think I just need ALL the fill data for a given destination. Why would filtering fills on destination chain (i.e. calling `getFillsForDestinationChain`) ever be useful? When would fills for a given spoke pool ever have a different destination chain than the spoke pool's chain?

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,113 @@ export class Dataworker {
...
+        }
+
+        // Grab all valid fills submitted to the destination spoke pool.
+        const fillsOnDestinationChain = destinationClient.getFills()
+        const validFillsOnDestinationChain = fillsOnDestinationChain.filter((fill) => {
+            // For each fill, see if we can find a deposit sent from the origin client that matches it.
+            for (const deposit of depositsForDestinationChain) {
+              // @dev: It doesn't matter which client we call validateFillForDeposit() on as the logic is 
+              // chain agnostic.
+              if (destinationClient.validateFillForDeposit(fill, deposit)) return true;
```
nicholaspai, line 75: IIUC, `validateFillForDeposit` is chain agnostic. Should we move this to a more common client, perhaps to a `ClientTools` class?

src/clients/RateModelClient.ts
```diff
@@ -54,11 +54,6 @@ export class RateModelClient {
     return this.rateModelDictionary.getRateModelForBlockNumber(l1Token, blockNumber);
   }
 
-  async validateRealizedLpFeePctForFill(fill: Fill, deposit: Deposit) {
```
chrismaree, line 57: 👍

src/clients/SpokePoolClient.ts
chrismaree: `getFillsForDestinationChain` is currently not used but you do raise a good point...the data structure:
```ts
private fills: { [DestinationChainId: number]: Fill[] } = {};
```
does not actually make sence as all fills seen on this client will, be definition, be at the destination chain of _this_ client.

`getDepositsForDestinationChain` is obviously diffrent as it's saying: from the perspective of _this_ client's spokePool the target is this other chain.

src/clients/SpokePoolClient.ts
```diff
@@ -56,7 +61,7 @@ export class SpokePoolClient {
     return { ...deposit, speedUpSignature: maxSpeedUp.depositorSignature, relayerFeePct: maxSpeedUp.newRelayerFeePct };
   }
 
-  getValidUnfilledAmountForDeposit(deposit: Deposit) {
+  getValidUnfilledAmountForDeposit(deposit: Deposit): BigNumber {
```
chrismaree, line 64: thanks for adding the return type. useful. we will have a separate PR later to go in and make this compliant all over the place.

The PR I'm working on also adds a bunch of additional logging spots.

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+    //             Attempt to map fill to an UnfilledDeposit sent from the origin chain
+    //             If a match is found, save fill and key by repaymentChainId => refundAddress
+
+    const unfilledDeposits: UnfilledDeposits = {};
+    const fillsToRefund: FillsToRefund = {};
+
+    for (const originChainId of Object.keys(this.spokePoolClients)) {
+      const originClient = this.spokePoolClients[originChainId];
+      if (!originClient.isUpdated())
+        throw new Error(`origin spokepoolclient with chain ID ${originChainId} not updated`);
```
chrismaree, line 40: silly nit but let's try keep the total line length on these kinds of errors down. ideally it's just a one liner.

```suggestion
      if (!originClient.isUpdated()) throw new Error(`origin spokepoolclient on chain ${originChainId} not updated`);
```

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+
+      // Loop over all other SpokePoolClient's to find deposits whose destination chain is the selected origin chain.
+      this.logger.debug({ at: "Dataworker", message: `Looking up data for origin spoke pool`, originChainId });
+      for (const destinationChainId of Object.keys(this.spokePoolClients)) {
+        if (originChainId === destinationChainId) continue;
+
+        const destinationClient = this.spokePoolClients[destinationChainId];
+        if (!destinationClient.isUpdated())
```
chrismaree, line 48: same nit about this log. one line pllleeease

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+        this.logger.debug({
+          at: "Dataworker",
+          message: `Found ${depositsForDestinationChain.length} deposits for destination chain ${destinationChainId}`,
+          originChainId,
+          destinationChainId,
+        });
+
+        const unfilledDepositsForDestinationChain = depositsForDestinationChain
+          .map((deposit) => {
+            return { data: deposit, unfilledAmount: destinationClient.getValidUnfilledAmountForDeposit(deposit) };
```
chrismaree, line 65: why key this by `data`? would it not be better to key it by just `deposit` then it will become:
```ts
return { deposit, unfilledAmount: destinationClient.getValidUnfilledAmountForDeposit(deposit) };
```

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+        for (const fill of validFillsOnDestinationChain) {
+          assign(fillsToRefund, [fill.repaymentChainId, fill.relayer], [fill]);
+        }
+      }
+    }
+
+    return {
+      fillsToRefund,
+      unfilledDeposits,
+    };
```
chrismaree, line 114:
```suggestion
    return { fillsToRefund, unfilledDeposits };
```

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+        this.logger.debug({
+          at: "Dataworker",
+          message: `Found ${validFillsOnDestinationChain.length} fills on destination ${destinationChainId} matching origin ${originChainId}`,
+          originChainId,
+          destinationChainId,
+        });
+
+        for (const fill of validFillsOnDestinationChain) {
+          assign(fillsToRefund, [fill.repaymentChainId, fill.relayer], [fill]);
+        }
```
chrismaree, line 107: might need to be listed. not sure if this is too long.
```suggestion
        for (const fill of validFillsOnDestinationChain)assign(fillsToRefund, [fill.repaymentChainId, fill.relayer], [fill]);
```

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+        // Grab all valid fills submitted to the destination spoke pool.
+        const fillsOnDestinationChain = destinationClient.getFills();
+        const validFillsOnDestinationChain = fillsOnDestinationChain.filter((fill) => {
+          // For each fill, see if we can find a deposit sent from the origin client that matches it.
+          for (const deposit of depositsForDestinationChain) {
+            // @dev: It doesn't matter which client we call validateFillForDeposit() on as the logic is
+            // chain agnostic.
+            // @dev: All of the deposits returned by `getDepositsForDestinationChain` will include the expected realized
+            // lp fee % for the deposit quote time. If this fill does not have the same realized lp fee %, then it will
+            // be ignored.
```
chrismaree, line 89: this commenting is non-standard with the `@dev` flag. we should not mix and match commenting styles methinks.
```suggestion
            // Note: it doesn't matter which client we call validateFillForDeposit() on as the logic is chain agnostic.
            // All of the deposits returned by getDepositsForDestinationChain will include the expected realized
            // lp fee % for the deposit quote time. If this fill does not have the same realized lp fee %, then it will be ignored.
```

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
+import { deploySpokePoolWithToken, enableRoutesOnHubPool, destinationChainId, originChainId, sinon } from "./utils";
```
chrismaree, line 1: some of these should be moved to the `const` import such as `destinationChainId, originChainId` then you'll be able to pull the imports from the following line into this line and decrease the line count in this bulky import.

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    // Set the spokePool's time to the provider time. This is done to enable the block utility time finder identify a
+    // "reasonable" block number based off the block time when looking at quote timestamps.
+    await spokePool_1.setCurrentTime(await getLastBlockTime(spokePool_1.provider));
+    await spokePool_2.setCurrentTime(await getLastBlockTime(spokePool_2.provider));
+  });
+
+  it("Default conditions", async function () {
+    // Throws error if spoke pool client not updated.
+    expect(() => dataworkerInstance._loadData()).to.throw(/not updated/);
```
chrismaree, line 137: ooo this is great!

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+    await updateAllClients();
+
+    const deposit1 = await buildDeposit(
+      spokePool_1,
+      erc20_1,
+      depositor,
+      depositor,
+      destinationChainId,
+      amountToDeposit
+    );
```
chrismaree, line 156: I know you are going to hate me for this but do you think we can decrease the line length on this so it is not 8 lines repeated over and over again? the simplest modification I can see is to change `buildDeposit` to make the `depositor` and `recipient` the same address. this is basically what I did in `fillWithRealizedLpFeePct` here: https://github.com/across-protocol/relayer-v2/blob/b39b050a7d6b93fae3417aebf9237c0a4962b3a2/test/Relayer.UnfilledDeposits.ts#L143

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+    await buildFill(
+      spokePool_1,
+      erc20_1,
+      depositor,
+      depositor,
+      relayer,
+      { ...deposit2, realizedLpFeePct: deposit2.realizedLpFeePct.div(toBN(2)) },
+      0.25
+    );
+    // @dev: This fill has identical deposit data to fill2 except for the destination token being different
```
chrismaree, line 288:
```suggestion
    // This fill has identical deposit data to fill2 except for the destination token being different
```

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+    const fill2 = await buildFill(spokePool_1, erc20_1, depositor, depositor, relayer, deposit2, 0.25);
+    const fill3 = await buildFill(spokePool_2, erc20_2, depositor, depositor, depositor, deposit1, 1);
+    await updateAllClients();
+    const data2 = dataworkerInstance._loadData();
+    expect(data2.fillsToRefund).to.deep.equal({
+      [repaymentChainId]: { [relayer.address]: [fill1, fill2], [depositor.address]: [fill3] },
+    });
+
+    // Submit fills without matching deposits. These should be ignored by the client.
+    // @dev: Switch the deposit data to make fills invalid.
```
chrismaree, line 268: same nit as before around using the `@dev` syntax.
```suggestion
    // Switch the deposit data to make fills invalid.
```

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+    await buildFill(spokePool_2, erc20_2, depositor, depositor, relayer, deposit2, 0.5);
+    await buildFill(spokePool_2, erc20_2, depositor, depositor, depositor, deposit2, 1);
+    await updateAllClients();
+    const data3 = dataworkerInstance._loadData();
+    expect(data3.fillsToRefund).to.deep.equal(data2.fillsToRefund);
+
+    // Submit fills that match deposit in all properties except for realized lp fee % or l1 token. These should be
+    // ignored because the rate model client deems them invalid. These are the two properties added to the deposit
+    // object by the spoke pool client.
+    // @dev: This fill has identical deposit data to fill2 except for the realized lp fee %.
```
chrismaree, line 278:
```suggestion
    // This fill has identical deposit data to fill2 except for the RealizedLpFeePct.
```

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+): Promise<Deposit> {
+  const _deposit = await deposit(
+    spokePool,
+    tokenToDeposit,
+    recipient,
+    depositor,
+    _destinationChainId,
+    _amountToDeposit
+  );
+  return {
```
chrismaree, line 46: is the whole point of this method just to append extra data onto the deposit object? if so, this would be useful in other places I do something similar here: https://github.com/across-protocol/relayer-v2/blob/b39b050a7d6b93fae3417aebf9237c0a4962b3a2/test/Relayer.UnfilledDeposits.ts#L70 where I append these other props directly onto the object in the exact same format you are doing it here. if you refactored this into utils we could simplify those other tests as well.

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+        if (unfilledDepositsForDestinationChain.length > 0) {
+          assign(unfilledDeposits, [destinationChainId], unfilledDepositsForDestinationChain);
+        } else {
+          this.logger.debug({
+            at: "Dataworker",
+            message: `All deposits are filled`,
+            originChainId,
+            destinationChainId,
+          });
+        }
```
chrismaree, line 78:
```suggestion
        if (unfilledDepositsForDestinationChain.length > 0) assign(unfilledDeposits, [destinationChainId], unfilledDepositsForDestinationChain);
        else
          this.logger.debug({
            at: "Dataworker",
            message: `All deposits are filled`,
            originChainId,
            destinationChainId,
          });
```

chrismaree (commented):
this is really awesome work @nicholaspai ! I'm sorry I left so many nits...I just really want to keep this repo as consistant and clean as posible as we add more and more complex code. Please try some of the refactors I proposed and I'll re review. I want to go through the logic on the filtering a bit more closely once the other changes are in.

src/clients/SpokePoolClient.ts
nicholaspai: Yeah so should we remove the `[DestinationChainId]` key?

nicholaspai: We could then flatten the fills to be an array of `Fill[]`

src/clients/SpokePoolClient.ts
nicholaspai: Sounds good I just added where I used it

src/dataworker/Dataworker.ts
```diff
@@ -1,9 +1,17 @@
...
-import { Deposit } from "../interfaces/SpokePool";
+import { BigNumber, winston, assign } from "../utils";
+import { SpokePoolClient, HubPoolClient, MultiCallBundler } from "../clients";
+import { Deposit, Fill } from "../interfaces/SpokePool";
 import { BundleEvaluationBlockNumbers } from "../interfaces/HubPool";
 
-// @notice Constructs roots to submit to HubPool on L1
+// @notice Constructs roots to submit to HubPool on L1. Fetches all data synchronously from SpokePool/HubPool clients
+// so this class assumes that those upstream clients are already updated and have fetched on-chain data from RPC's.
+type UnfilledDeposit = {
```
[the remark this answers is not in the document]
nicholaspai: Ah good catch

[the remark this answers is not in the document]
nicholaspai: I don't really understand the diff between types and interfaces but [this post](https://stackoverflow.com/questions/37233735/interfaces-vs-types-in-typescript) seems to suggest that types are much better for what we're doing since we're not trying to extend these interfaces in our class implementations.

src/dataworker/Dataworker.ts
```diff
@@ -13,40 +21,134 @@ export class Dataworker {
...
+    //         Store deposits that are sent from origin chain to destination chain as an UnfilledDeposit
+    //             Associate this UnfilledDeposit with the destination chain and set fillAmountRemaining
+    //         Grab all fills on destination client
+    //             Attempt to map fill to an UnfilledDeposit sent from the origin chain
+    //             If a match is found, save fill and key by repaymentChainId => refundAddress
+
+    const unfilledDeposits: UnfilledDeposits = {};
+    const fillsToRefund: FillsToRefund = {};
+
+    for (const originChainId of Object.keys(this.spokePoolClients)) {
```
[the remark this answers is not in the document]
nicholaspai: Good call. Done

src/dataworker/Dataworker.ts
nicholaspai: sure

src/dataworker/Dataworker.ts
nicholaspai: would a `forEach` make you happy?

test/Dataworker.loadData.ts
nicholaspai: Fixed

test/Dataworker.loadData.ts
```diff
@@ -0,0 +1,301 @@
...
+    depositor,
+    relayer,
+    deposit.depositId,
+    deposit.originChainId,
+    deposit.amount,
+    deposit.amount.mul(toBNWei(pctOfDepositToFill)).div(toBNWei(1)),
+    deposit.realizedLpFeePct,
+    deposit.relayerFeePct
+  );
+}
```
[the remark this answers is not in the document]
nicholaspai: Hmmm probably yeah

test/Dataworker.loadData.ts
nicholaspai: cool done where possible

test/Dataworker.loadData.ts
nicholaspai: Fixed

test/Dataworker.loadData.ts
nicholaspai: Fixed

test/Dataworker.loadData.ts
nicholaspai: Fixed

test/Dataworker.loadData.ts
nicholaspai: Also applied this util function in all other tests

nicholaspai:
> this is really awesome work @nicholaspai ! I'm sorry I left so many nits...I just really want to keep this repo as consistant and clean as posible as we add more and more complex code. Please try some of the refactors I proposed and I'll re review. I want to go through the logic on the filtering a bit more closely once the other changes are in.

Added all syntactical nits and added the helper functions to `test/utils`! Hopefully now you can review the logic but please do point out other stylistical inconsistencies

chrismaree (approved):
This is sick! thanks for going through all my feedback.

src/dataworker/Dataworker.ts
chrismaree: what you have is fine. just you dont need the extra pair of `{` & `}` :)

src/clients/SpokePoolClient.ts
nicholaspai: yeah so far I haven't had to use `getFillsForRepaymentChain` so I won't add that in this PR. But I will flatten the `fills` structure a bit.

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: according to that style these should be interfaces then so I'll change them and move to `interfaces/SpokePool.ts` for now

src/dataworker/Dataworker.ts
nicholaspai: gotcha i ended up using `forEach`

PR #16, opened 2022-03-31
feat: add multicall transaction processing logic

chrismaree:


src/clients/HubPoolClient.ts
```diff
@@ -63,14 +62,8 @@ export class HubPoolClient {
...
     }
 
-    this._isUpdated = true;
-
     this.firstBlockToSearch = searchConfig[1] + 1; // Next iteration should start off from where this one ended.
 
     this.logger.debug({ at: "HubPoolClient", message: "Client updated!" });
   }
-
-  isUpdated(): Boolean {
```
nicholaspai, line 73: why remove? Is there another way to detect if this has updated?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +24,94 @@ export class MultiCallBundler {
...
+      this.logger.debug({ at: "MultiCallBundler", message: "Executing tx bundle", number: this.transactions.length });
+
+      // Simulate the transaction execution for the whole queue.
+      const transactionsWillSucceed = await Promise.all(
+        this.transactions.map(async (tx: preSignedTransaction) => this.willSucceed(tx.contract, tx.method, tx.args))
+      );
+      console.log("transactionsWillSucceed", transactionsWillSucceed);
+
+      // If any transactions will revert then log the reason and remove them from the transaction queue.
+      if (transactionsWillSucceed.every((willSucceed) => !willSucceed))
```
nicholaspai, line 42: Shouldn't this be `every((willSucceed) => willSucceed)`?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +24,94 @@ export class MultiCallBundler {
...
+    const feeData = await provider.getFeeData();
+    if (feeData.maxFeePerGas && feeData.maxPriorityFeePerGas)
+      return {
+        maxFeePerGas: feeData.maxFeePerGas.mul(priorityScaler),
+        maxPriorityFeePerGas: feeData.maxPriorityFeePerGas.mul(priorityScaler),
+      };
+    else return { gasPrice: feeData.gasPrice.mul(priorityScaler) };
+  }
+
+  async willSucceed(contract: Contract, method: string, args: any) {
```
nicholaspai, line 109: What if the `method` is a read only method? If read only methods don't work with this then maybe rename this function to `willWriteMethodSucceed` to be more clear

src/clients/RateModelClient.ts
```diff
@@ -60,8 +60,6 @@ export class RateModelClient {
   }
 
   async update() {
-    if (!this.hubPoolClient.isUpdated()) throw new Error("hubpool not updated");
```
nicholaspai, line 63: why remove this?

src/clients/RateModelClient.ts
```diff
@@ -25,9 +25,9 @@ export class RateModelClient {
 
   async computeRealizedLpFeePct(deposit: Deposit, l1Token: string) {
     this.logger.debug({ at: "RateModelClient", message: "Computing realizedLPFeePct", deposit, l1Token });
-    const quoteBlockNumber = (await this.blockFinder.getBlockForTimestamp(deposit.quoteTimestamp)).number;
+    let quoteBlockNumber = (await this.blockFinder.getBlockForTimestamp(deposit.quoteTimestamp)).number;
     // Set to this temporarily until we re-deploy. The RateModelStore was deployed after the spokePool's deposits.
-    // const quoteBlockNumber = 30626071;
+    if (quoteBlockNumber < 30626070) quoteBlockNumber = 30626070;
```
nicholaspai, line 30: Why this magic number?

src/relayer/Relayer.ts
```diff
@@ -44,6 +51,18 @@ export class Relayer {
     return this.spokePoolClients[deposit.destinationChainId].spokePool;
   }
 
+  constructRelayFilledMrkdwn(deposit: Deposit, repaymentChainId: number, amountToFill: BigNumber) {
```
chrismaree, line 54: I also built out the full log for the deposit. will add multi decimal support in a following PR when I make the balance monitor.

src/utils/ContractInstance.ts
```diff
@@ -1,8 +1,24 @@
-import { ethers } from "ethers";
-import * as typechain from "@across-protocol/contracts-v2"; // This imports all typechain artifacts from the contracts-v2 package.
+import { getNetworkName, Contract, Wallet } from "./";
 
-export function contractAt(name: string, address: string, signer: ethers.Wallet) {
-  const typeChainArtifact = typechain[`${name}__factory`];
-  if (!typeChainArtifact) throw new Error(`No contract factory for ${name}`);
-  return typeChainArtifact.connect(address, signer);
+import { getContractArtifact } from "@across-protocol/contracts-v2";
```
chrismaree, line 3: this is the new export method from my across-v2 PR: see here: https://github.com/across-protocol/contracts-v2/pull/129

src/utils/TransactionUtils.ts
```diff
@@ -0,0 +1,39 @@
+import { winston, Contract, toBN } from "../utils";
+
+// Note that this function will throw if the call to the contract on method for given args reverts. Implementers
+// of this method should be considerate of this and catch the response to deal with the error accordingly.
+export async function runTransaction(logger: winston.Logger, contract: Contract, method: string, args: any) {
```
chrismaree, line 5: put there here to be used to `runTransaction` for all transactions in all bots.

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
     this.transactions = [];
   }
 
+  getTarget(index: number) {
+    return { target: this.transactions[index].contract.address };
+  }
+
   async executeTransactionQueue() {
```
chrismaree, line 32: main logic in this PR.

src/utils/ContractInstance.ts
```diff
@@ -1,8 +1,24 @@
...
+  if (contractName === "SpokePool") contractName = castSpokePoolName(networkId);
+  console.log("contractName", contractName);
+  const artifact = getContractArtifact(contractName, networkId);
+  if (!artifact) throw new Error(`Could not find artifact for contract ${contractName} on ${networkId}`);
+  return new Contract(artifact.address, artifact.abi, signer);
+}
+
+// If the name of the contract is SpokePool then we need to apply a transformation on the name to get the correct
+// contract name. For example, if the network is "mainnet" then the contract is called Ethereum_SpokePool.
+export function castSpokePoolName(networkId: number): string {
```
chrismaree, line 17: spoke pools need a bit of logic to work out the contract name based on the network name. i.e if the network is `mainnet` then we want `Ethereum_spokePool`

src/clients/ClientHelper.ts
```diff
@@ -35,6 +35,16 @@ export function constructClients(logger: winston.Logger, config: RelayerConfig)
   // const gasEstimator = new GasEstimator() // todo when this is implemented in the SDK.
   const multiCallBundler = new MultiCallBundler(logger, null);
 
+  logger.debug({
+    at: "constructClients",
+    message: "Clients constructed",
+    hubPoolAddress: hubPool.address,
```
nicholaspai, line 41: nit: maybe just rename from `hubPoolAddress` to `hubPool` since its assumed this will just log an address

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
-      message: "All transactions executed",
-      hashes: transactionReceipts.map((receipt) => receipt.transactionHash),
-    });
-    return transactionReceipts;
+    try {
+      this.logger.debug({ at: "MultiCallBundler", message: "Executing tx bundle", number: this.transactions.length });
+
+      // Simulate the transaction execution for the whole queue.
+      const transactionsSucceed = await Promise.all(
+        this.transactions.map(async (tx: preSignedTransaction) => willSucceed(tx.contract, tx.method, tx.args))
```
nicholaspai, line 38: This should be `await willSucceed` right?
```suggestion
        this.transactions.map(async (tx: preSignedTransaction) => await willSucceed(tx.contract, tx.method, tx.args))
```

test/Relayer.IterativeFill.ts
```diff
@@ -18,7 +18,7 @@ describe("Relayer: Iterative fill", async function () {
     [relayer_signer] = await ethers.getSigners(); // note we use relayer_signer as the owner as well to simplify the test.
     ({ hubPool, mockAdapter } = await hubPoolFixture());
   });
-  it("One token on multiple chains", async function () {
+  it.only("One token on multiple chains", async function () {
```
nicholaspai, line 21: leftover .only
```suggestion
  it("One token on multiple chains", async function () {
```

test/Relayer.UnfilledDeposits.ts
```diff
@@ -141,7 +139,7 @@ async function updateAllClients() {
 }
 
 async function fillWithRealizedLpFeePct(spokePool, relayer, depositor, deposit, relayAmount = amountToRelay) {
-  const realizedLpFeePctForDeposit = await rateModelClient.computeRealizedLpFeePct(deposit, l1Token.address);
+  const realizedLpFeePctForDeposit = await rateModelClient.computeRealizedLpFeePct(deposit, l1Token.address)
```
nicholaspai, line 142: did you forget to lint this?

test/utils/utils.ts
```diff
@@ -9,6 +9,7 @@ import { RateModelClient } from "../../src/clients/RateModelClient";
 import winston from "winston";
 import sinon from "sinon";
 import chai from "chai";
+import { AnyNsRecord } from "dns";
```
nicholaspai, line 12: what is this?

src/clients/MultiCallBundler.ts
chrismaree: aha no. the map should not be async seeing this is within a `Promise.all` it must return a promise and not await it.

test/Relayer.UnfilledDeposits.ts
chrismaree: fixed.

test/utils/utils.ts
chrismaree: my IDE added it...have removed.

src/clients/MultiCallBundler.ts
```diff
@@ -1,12 +1,20 @@
...
+  contract: Contract;
+  method: string;
+  args: any;
+  message: string;
+  mrkdwn: string;
+}
 
 export class MultiCallBundler {
-  private transactions: Transaction[] = [];
+  private transactions: preSignedTransaction[] = [];
```
mrice32, line 11: nit: why are these called "presigned"? They don't seem to be signed at all, no?

src/clients/MultiCallBundler.ts
```diff
@@ -1,12 +1,20 @@
-import { winston, Transaction } from "../utils";
+import { winston, getNetworkName, assign, Contract, runTransaction, willSucceed, etherscanLink } from "../utils";
+interface preSignedTransaction {
```
mrice32, line 2: nit: types should be capitalized.

src/clients/MultiCallBundler.ts
```diff
@@ -1,12 +1,20 @@
...
-  private transactions: Transaction[] = [];
+  private transactions: preSignedTransaction[] = [];
   constructor(readonly logger: winston.Logger, readonly gasEstimator: any) {}
 
-  // Adds defined transaction to the transaction queue.
-  addTransaction(transaction: Transaction) {
-    if (transaction) this.transactions.push(transaction);
+  // Adds all information associated with a transaction to the transaction queue. This is the intention of the
+  // caller to send a transaction. The transaction might not be executable, which should be filtered later.
+  enqueueTransaction(contract: Contract, method: string, args: any, message: string, mrkdwn: string) {
```
mrice32, line 16: Any reason you decided to spread these args rather than just taking a transaction object?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
     this.transactions = [];
   }
 
+  getTarget(index: number) {
```
mrice32, line 28: nit: private if you only need to use this in this class.

src/clients/MultiCallBundler.ts
```diff
@@ -1,12 +1,20 @@
...
+  private transactions: preSignedTransaction[] = [];
   constructor(readonly logger: winston.Logger, readonly gasEstimator: any) {}
 
-  // Adds defined transaction to the transaction queue.
-  addTransaction(transaction: Transaction) {
-    if (transaction) this.transactions.push(transaction);
+  // Adds all information associated with a transaction to the transaction queue. This is the intention of the
+  // caller to send a transaction. The transaction might not be executable, which should be filtered later.
+  enqueueTransaction(contract: Contract, method: string, args: any, message: string, mrkdwn: string) {
+    if (contract) this.transactions.push({ contract, method, args, message, mrkdwn });
```
mrice32, line 17: Why do you check `if(contract)`? Doesn't the type say that contract cannot be undefined?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+    try {
+      this.logger.debug({ at: "MultiCallBundler", message: "Executing tx bundle", number: this.transactions.length });
+
+      // Simulate the transaction execution for the whole queue.
+      const transactionsSucceed = await Promise.all(
+        this.transactions.map((tx: preSignedTransaction) => willSucceed(tx.contract, tx.method, tx.args))
+      );
+
+      // If any transactions will revert then log the reason and remove them from the transaction queue.
+      if (transactionsSucceed.some((willSucceed) => !willSucceed.succeed))
```
mrice32, line 42:
```suggestion
      if (transactionsSucceed.some(({ succeed }) => !succeed))
```

Just because I think it's a little confusing shadowing the function name with the variable name.

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+        this.logger.error({
+          at: "MultiCallBundler",
+          message: "Some transaction in the queue are reverting!",
+          revertingTransactions: transactionsSucceed.flatMap((willSucceed, index) =>
+            willSucceed.succeed
+              ? [] // return blank array if the transaction is succeeding. Else, return information on why reverted.
+              : { target: this.getTarget(index), reason: willSucceed.reason, message: this.transactions[index].message }
```
mrice32, line 49: Instead of having a separate getTarget method, could we just add the original txn object to each transactionsSucceed element above?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
-      at: "MultiCallBundler",
-      message: "All transactions executed",
-      hashes: transactionReceipts.map((receipt) => receipt.transactionHash),
-    });
-    return transactionReceipts;
+    try {
+      this.logger.debug({ at: "MultiCallBundler", message: "Executing tx bundle", number: this.transactions.length });
+
+      // Simulate the transaction execution for the whole queue.
+      const transactionsSucceed = await Promise.all(
```
mrice32, line 37: As mentioned elsewhere, I think augmenting the transaction object here would make this a bit simpler:
```ts
      const transactionsSucceed = await Promise.all(
        this.transactions.map((tx: preSignedTransaction) => ({ ...willSucceed(tx.contract, tx.method, tx.args)), ...tx })
      );

File: src/clients/MultiCallBundler.ts [hunk 24]
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+        this.logger.error({
+          at: "MultiCallBundler",
+          message: "Some transaction in the queue are reverting!",
+          revertingTransactions: transactionsSucceed.flatMap((willSucceed, index) =>
+            willSucceed.succeed
+              ? [] // return blank array if the transaction is succeeding. Else, return information on why reverted.
+              : { target: this.getTarget(index), reason: willSucceed.reason, message: this.transactions[index].message }
+          ),
+        });
+      const validTransactions: preSignedTransaction[] = transactionsSucceed.flatMap((willSucceed, i) =>
```
mrice32, line 52: If you augment the transaction above, this could be a filter operation:
```ts
const validTransactions = transactionsSucceed.filter(tx => tx.succeed);
```

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+        this.logger.debug({ at: "MultiCallBundler", message: "No valid transactions in the queue" });
+        return;
+      }
+
+      // Group by target chain. Note that there is NO grouping by target contract. The relayer will only ever use this
+      // multiCallBundler to send multiple transactions to one target contract on a given target chain and so we dont
+      // need to group by target contract. This can be further refactored with another group by if this is needed.
+      let groupedTransactions = {};
+      for (const transaction of validTransactions)
+        assign(groupedTransactions, [(transaction.contract.provider as any).network.chainId], [transaction]);
```
mrice32, line 66: nit: from the docs, it doesn't appear as if `.network` is a standard field on ethers providers: https://docs.ethers.io/v5/api/providers/provider/#Provider-getNetwork.

Should we call `(await provider.getNetwork()).chainId` instead?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+
+      if (validTransactions.length == 0) {
+        this.logger.debug({ at: "MultiCallBundler", message: "No valid transactions in the queue" });
+        return;
+      }
+
+      // Group by target chain. Note that there is NO grouping by target contract. The relayer will only ever use this
+      // multiCallBundler to send multiple transactions to one target contract on a given target chain and so we dont
+      // need to group by target contract. This can be further refactored with another group by if this is needed.
+      let groupedTransactions = {};
```
mrice32, line 64: nit const?

And should we add a type to this?

src/clients/MultiCallBundler.ts
```diff
@@ -17,18 +25,97 @@ export class MultiCallBundler {
...
+        txs: Object.keys(groupedTransactions).map((chainId) => ({ chainId, num: groupedTransactions[chainId].length })),
+      });
+
+      // Construct multiCall transaction for each target chain.
+      const multiCallTransactionsResult = await Promise.allSettled(
+        Object.keys(groupedTransactions).map((chainId) => this.buildMultiCallBundle(groupedTransactions[chainId]))
+      );
+
+      const transactionReceipts = await Promise.allSettled(
+        multiCallTransactionsResult.map((transaction) => (transaction as any).value.wait())
```
mrice32, line 80: Would providing an explicit return value for buildMultiCallBundle help typescript deduce the type of transaction? What type is it inferring now?

mrice32 (commented):

src/clients/MultiCallBundler.ts
chrismaree: fair. have renamed to `TransactionObject`.

src/clients/MultiCallBundler.ts
chrismaree: I wanted to take in the full `contract.method` but this is limiting due to how ethers works. we then, for example, can't do things like `contract.callStatic.method` or `contract.interface.encodeFunctionData`. this way we get best of both worlds, and we can log the args.

src/clients/MultiCallBundler.ts
chrismaree: indeed it is not. it is the only synchronous version where you can get this prop. this is why I needed to cast. I dont want to do the async call or we need to call `getNetwork` on _each_ transaction in the bundle. this I'd like to avoid, no?

src/clients/MultiCallBundler.ts
chrismaree: o, ye. I forgot you can append to a const object/array.

I have added the typing.

src/clients/MultiCallBundler.ts
chrismaree: right now its not sure of the return type. 

I have added typing on the return.

src/clients/MultiCallBundler.ts
chrismaree: I'm not exactly sure how this would work. can you provide some more context? sorry if I'm being thick.

src/clients/MultiCallBundler.ts
chrismaree: this was lingering from the previous implementation. have removed it.

src/clients/MultiCallBundler.ts
mrice32: Will this work? My understanding is that this _cannot_ work in all cases because of the following situation:

```ts
const { chainId } = new ethers.providers.JsonRpcProvider(url);
```

There would be no way that the provider could get the chainId from the URL before it has to return it, right?

I'm also just concerned that this chainId prop will not exist on every provider since it doesn't seem to be documented anywhere.

Would it help if we put this inside a Promise.all so it was parallelized?

mrice32 (approved):
Looks great! Just added to a few ongoing threads.

src/clients/MultiCallBundler.ts
chrismaree: ye so there are some issues with this technique, like you laid out. I tried creating an `AugmentedProvider` that has the `chainId` prop set when you create the provider. this works fine but issue comes in with some testing related things when we work with a mock spokePool. 

my solution now is to simply pass this prop into the `enqueueTransaction` and add it as part of the `AugmentedTransaction` type.

src/utils/ContractInstance.ts
chrismaree: ye, this would be good. let's add this later.

src/clients/MultiCallBundler.ts
chrismaree: o yes! you right haha. have changed it.

src/clients/MultiCallBundler.ts
chrismaree: fixed.

src/clients/MultiCallBundler.ts
chrismaree: ye, this is a great point. only issue with your proposal is we'll loose some of the other props we need. Simple solution is to modify `willSucceed` to take in  an `AugmentedTransaction` and return `AugmentedTransaction + `succeed` + `reason`.

src/clients/MultiCallBundler.ts
chrismaree: Further renamed to `AugmentedTransaction`

src/clients/MultiCallBundler.ts
chrismaree: changing this actually had some very positive ramifications throughout the code. we now pas around an `AugmentedTransaction` which made some of the code in the `Relayer.ts` easier to reason about. good recommendation.

src/clients/MultiCallBundler.ts
chrismaree: ok I did this. it worked out pretty well. we now pass `AugmentedTransaction` into the `willSucceed` method and return `transaction: AugmentedTransaction; succeed: boolean; reason: string` so I no longer need to do any indexing stuff later down the line.

src/clients/MultiCallBundler.ts
chrismaree: 100%. worked great.

src/clients/MultiCallBundler.ts
chrismaree: ye, in the event the `this.buildMultiCallBundle` returns null (due to mismatching targets) this will error, you are right. I've refactored it to check it is not null and to execute a no-op in this case. we can make this more robust later by improving some of the type declarations used here. for now I'm not going to do this as this will require a meaningful refactor of some other logic.

PR #18, opened 2022-04-01
refactor: Flatten fills property in spokepool-client

nicholaspai:
From the user of the SpokePoolClient's perspective, this PR replaces the redundant function `getFillsForDestinationChain` with `getFillsForRepaymentChainId`. Note that `getFillsForDestinationChain` is redundant because all of the fills for the current `SpokePoolClient` are by definition using the same `destinationChainId`--that of the `SpokePool`.

Moreover, keying `fills` by `destinationChainId` is also redundant since they all share the same `destinationChainId`.

I can see an arugment potentially for keying by `repaymentChainId` since a dataworker might only want to send refunds for a specific chain, but the dataworker would still probably want to loop over all `spokePoolClients` and concatenate the results of `getFillsForRepaymentChainId` for each of these clients. This is why I've included a `getFillsForRepaymentChainId` function.


src/clients/SpokePoolClient.ts
```diff
@@ -4,7 +4,7 @@ import { Deposit, Fill, SpeedUp } from "../interfaces/SpokePool";
 
 export class SpokePoolClient {
   private deposits: { [DestinationChainId: number]: Deposit[] } = {};
-  private fills: { [DestinationChainId: number]: Fill[] } = {};
+  private fills: Fill[] = [];
```
chrismaree, line 7: good.

chrismaree (approved):
lgtm. thanks for simplifying this.

PR #19, opened 2022-04-04
feat: Implement slow relay root builder

nicholaspai:
- Ignores `bundleEvaluationBlockNumbers` for now
- Refactors common logic shared with all `Dataworker` tests into a `fixtures/Dataworker` file


src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
...
+    // Key leaves temporarily by destination chain ID which we'll re-order later.
+    const leavesByChainId: { [chainId: number]: RelayData[] } = {};
+
+    for (const destinationChainId of Object.keys(unfilledDeposits)) {
+      // We need to re-order leaves deterministically such that the same data returned by _loadData() always produces
+      // the same merkle root. So we'll order the deposits for each chain ID by unfilled amount since this will order
+      // leaves by largest token outflows from the SpokePool to fulfill slow relays. (Recall that slow relay leaf
+      // executions send to the recipient the unfilled amount of their deposit, not just the deposit amount).
+      unfilledDeposits[destinationChainId].sort(function (a: UnfilledDeposit, b: UnfilledDeposit) {
+        return a.unfilledAmount.gt(b.unfilledAmount) ? -1 : 1;
```
nicholaspai, line 126: I was kind of hesitant here so I'd like some specific feedback. Its a bit strange to sort `SlowRelay` leaves by `unfilledAmont` which is a value that doesn't actually exist on the leaf. instead, this property is defined by the `SlowRelay.amount` and all the `FillRelay` events associated with it. Is this potentially going to cause pain for the dataworker later?

I ultimately decided to keep it because no matter what the data worker is going to have to be aware of the `unfilledAmounts` when constructing the `poolRebalanceRoot` and also in deciding which slow relay leaves to add to this root.

test/Dataworker.loadData.ts
```diff
@@ -1,79 +1,37 @@
-import { deploySpokePoolWithToken, enableRoutesOnHubPool, expect, ethers, Contract } from "./utils";
```
nicholaspai, line 1: All changes in this file are just copy pasta or removal of no-longer needed imports due to the `Dataworker.Fixture` refactor!

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
...
+      });
+
+      // Associate relay data implied by unfilled deposit with destination chain ID.
+      leavesByChainId[destinationChainId] = [];
```
chrismaree, line 130: the assumption here is that this runs from blank each time, right?

src/dataworker/Dataworker.ts
```diff
@@ -1,7 +1,9 @@
 import { winston, assign } from "../utils";
+import { buildSlowRelayTree, RelayData } from "@across-protocol/contracts-v2/dist/test-utils";
```
chrismaree, line 2: this should not be done. we must not import anything from the `test-utils`. we can _only_ import directly from `@across-protocol/contracts-v2` base package and none of the nested exports. for example see here: https://github.com/across-protocol/relayer-v2/blob/6c80b5dfd123424e4b8a1b7d0c7ed5bb6fbbf48c/src/utils/ContractInstance.ts#L3

what you've done here will drastically bloat the package size and cause the overall bot to be much slower when starting up. I've made a clear separation in the other parts of the bot to segment between the imports of `@across-protocol/contracts-v2` for the bot and ``@across-protocol/contracts-v2/dist/test-utils` ONLY for the test of the bot.

The correct solution is to add the `MerkleTree.ts` to the `index.ts` and export it from the `@across-protocol/contracts-v2` package, in the same way that we export the type chain exports and how I export `getContractArtifact` in this PR: https://github.com/across-protocol/contracts-v2/pull/129

As a second point: I really dont think we want to use `RelayData`. this is very loosely typed vs what this bot has. for example, it stores all props as a _string_ which it is fine for the tests but really should not be the case in the bots. Please see the strict type declaration for the same information here https://github.com/across-protocol/relayer-v2/blob/6c80b5dfd123424e4b8a1b7d0c7ed5bb6fbbf48c/src/interfaces/SpokePool.ts#L17 that you should be using.

src/dataworker/Dataworker.ts
```diff
@@ -1,7 +1,9 @@
...
 import { SpokePoolClient, HubPoolClient, MultiCallBundler } from "../clients";
-import { UnfilledDeposits, FillsToRefund } from "../interfaces/SpokePool";
+import { UnfilledDeposits, FillsToRefund, UnfilledDeposit } from "../interfaces/SpokePool";
 import { BundleEvaluationBlockNumbers } from "../interfaces/HubPool";
+import { MerkleTree } from "@across-protocol/contracts-v2/dist/utils/MerkleTree";
```
chrismaree, line 6: linked to above comment. we should not be importing like this. we need to modify the export in the `@across-protocol/contracts-v2` package to correctly export this class. I can help you with this.

chrismaree (commented):
initial code looks great! I'm a bit worried about the typing & how some things are being imported. I'm going to add the `MerkleLib` to the export on the other package and we can re-review once that's done.

src/dataworker/Dataworker.ts
mrice32: Why do we sort by the unfilled amount? Should we sort by something simpler (in the UMIP, I sort by origin chain id and deposit id since that tuple is guaranteed to be unique) since the ordering doesn't really matter here?

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
...
+      // leaves by largest token outflows from the SpokePool to fulfill slow relays. (Recall that slow relay leaf
+      // executions send to the recipient the unfilled amount of their deposit, not just the deposit amount).
+      unfilledDeposits[destinationChainId].sort(function (a: UnfilledDeposit, b: UnfilledDeposit) {
+        return a.unfilledAmount.gt(b.unfilledAmount) ? -1 : 1;
+      });
+
+      // Associate relay data implied by unfilled deposit with destination chain ID.
+      leavesByChainId[destinationChainId] = [];
+      unfilledDeposits[destinationChainId].forEach((deposit) => {
+        leavesByChainId[destinationChainId].push({
```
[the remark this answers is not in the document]
mrice32: Sorry, maybe I'm underthinking this, but this seems like it should just be a normal map, with no assigns or pushing?

```ts
leavesByChainId[destinationChainId] = unfilledDeposits[destinationChainId].map(deposit => ({
  depositor: deposit.deposit.depositor,
  ...
  })
);
```

I only say this because it seems like a pretty standard array transformation, which is what .map is designed for.

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
...
+      .sort(function (a, b) {
+        return Number(b) - Number(a);
+      })
+      .reduce(function (prev: RelayData[], current: string) {
+        return prev.concat(leavesByChainId[Number(current)]);
+      }, []);
+
+    return leaves.length > 0 ? await buildSlowRelayTree(leaves) : null;
+
+    // TODO: Figure out how to store merkle trees. IPFS?
```
mrice32, line 158: Hmmm, do we need to store? Or should we regenerate them whenever we need to use them for execution?

Note: I think we'll have to have a way to regenerate them to build the disputer anyway.

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
...
+      });
+    }
 
-    // For each destination chain ID key in unfilledDeposits
-    //     Order by fillAmountRemaining
-    //     Make Leaf for fill
-    // Construct root
+    // Flatten leaves deterministically so that the same root is always produced from the same _loadData return value.
+    // To do that we'll sort keys by chain ID and then concat sorted leaves.
+    const leaves: RelayData[] = Object.keys(leavesByChainId)
```
mrice32, line 148: nit: rather than flattening here, any reason we shouldn't just flatten all the unfilled deposits into a single array, process them one-by-one, and then call sort once at the end with a total sorting function?

Only reason I bring this up is that it's a little confusing to try to reason about how the sort operation above and this one will interact with one another (I do think it's correct based on the sorting criteria you've outlined, just a little hard to parse). I also think it might reduce the number of nested loops, which could make this function easier to reason about.

src/dataworker/Dataworker.ts
nicholaspai: Yeah this function is stateless in the sense that it always should return the same MerkleRoot for the same data loaded by `_loadData`

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +108,54 @@ export class Dataworker {
     };
   }
 
-  async buildSlowRelayRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
-    this._loadData();
+  async buildSlowRelayRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers): Promise<MerkleTree<RelayData>> | null {
+    const { unfilledDeposits } = this._loadData();
+    // TODO: Use `bundleBlockNumbers` to decide how to filter which blocks to keep in `unfilledDeposits`.
+
+    if (Object.keys(unfilledDeposits).length === 0) return null;
```
mrice32, line 115: Do unfilled deposits include deposits that haven't been relayed at all? I think we should only consider _partial relays_. The reason is a little nuanced, but basically, the frame of reference for the bundleEvaluationBlockNumbers is the _destination_ chain, not the _origin_ chain. For instance, if a deposit is outside of the bundle evaluation block numbers but the relay is included, we include it in the relayer refunds. If we use the deposits as the source of truth here, it's inconsistent with how we use these block numbers when relaying.

I'm not 100% sure if this causes any problems, but I can see weird scenarios where a deposit was completely filled in one shot, but the fill wasn't in the bundle evaluation block number for the destination chain, but the deposit was included in the bundle evaluation block number for the origin chain, so it's treated as a slow relay.

Thoughts?

mrice32 (commented):
Looks awesome! Very nice work on this. A few comments

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: But the slow relay root needs to accept an array of `RelayData` not fill, so we need to define a new `RelayData` type

src/dataworker/Dataworker.ts
nicholaspai: Good point will do

src/dataworker/Dataworker.ts
nicholaspai: Ah yes great point I'm going to add to this [PR](https://github.com/across-protocol/relayer-v2/pull/20) which makes it easier to figure out whether a deposit is completely unfilled. It adds a function `getValidFillsForDeposit` which returns fills associated with a deposit.

src/dataworker/Dataworker.ts
chrismaree: ok I've added some details in this PR that should be useful: https://github.com/across-protocol/relayer-v2/pull/21/files

this includes: `RelayData` type and utils version of `buildSlowRelayTree` and `MerkleTree`

src/dataworker/Dataworker.ts
nicholaspai: I ended up reordering them according to matt's suggestion https://github.com/across-protocol/relayer-v2/pull/19#discussion_r843099315

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +110,42 @@ export class Dataworker {
...
+            recipient: deposit.deposit.recipient,
+            destinationToken: deposit.deposit.depositor,
+            amount: deposit.deposit.amount.toString(),
+            originChainId: deposit.deposit.originChainId.toString(),
+            destinationChainId: deposit.deposit.destinationChainId.toString(),
+            realizedLpFeePct: deposit.deposit.realizedLpFeePct.toString(),
+            relayerFeePct: deposit.deposit.relayerFeePct.toString(),
+            depositId: deposit.deposit.depositId.toString(),
+          };
+        })
```
chrismaree, line 134: for this you can simply cast the deposit to a `RelayData` like this:
```suggestion
        ...unfilledDeposits[destinationChainId].map((deposit) => deposit as DepositData)
```

nicholaspai: As discussed offline, typescript casting is really a type assertion and doesn't change the object at runtime https://stackoverflow.com/questions/52296152/typescript-cast-using-as

src/dataworker/Dataworker.ts
```diff
@@ -106,13 +110,42 @@ export class Dataworker {
...
+            amount: deposit.deposit.amount,
+            originChainId: deposit.deposit.originChainId,
+            destinationChainId: deposit.deposit.destinationChainId,
+            realizedLpFeePct: deposit.deposit.realizedLpFeePct,
+            relayerFeePct: deposit.deposit.relayerFeePct,
+            depositId: deposit.deposit.depositId,
+          };
+        })
+      );
+    });
```
mrice32, line 136: You can take or leave this suggestion. I just thought the nested map + .flat was a little simpler, but I could see the other way too.
```suggestion
    const leaves: RelayData[] = Object.values(unfilledDeposits).map((deposits) => deposits.map((deposit: UnfilledDeposit): RelayData => ({
            depositor: deposit.deposit.depositor,
            recipient: deposit.deposit.recipient,
            destinationToken: deposit.deposit.depositor,
            amount: deposit.deposit.amount,
            originChainId: deposit.deposit.originChainId,
            destinationChainId: deposit.deposit.destinationChainId,
            realizedLpFeePct: deposit.deposit.realizedLpFeePct,
            relayerFeePct: deposit.deposit.relayerFeePct,
            depositId: deposit.deposit.depositId,
        })
      ).flat();
```

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
chrismaree: I really dont like this syntax. @nicholaspai and I spoke about doing some casting between `Deposit` and `RelayData` but there are ts issues with this.

 Ideally, we want the ability to do the following:
1) extract all the keys within `RelayData`
2) for each key assign a value in the `relayData` from the given key from `Deposit`. This basically makes this spread, assign and cast operation programatic, rather than manual the way you have it now. Only issue is I have not found a predictable way to extract keys from ts interfaces. I have tried using this: https://github.com/kimamula/ts-transformer-keys but it was not consistent. 

Another option would be to have a const in the interfaces file that defines all the keys you are expecting within the `RelayData`. this can then be traversed and spread from the deposit object.

src/dataworker/Dataworker.ts
chrismaree: we could choose to store this. an idea @nicholaspai and I have spoken about a bit is to add a prop to `proposeRootBundle` that is something like `ipfsHash` that lets you add a structure to IPFS which contains the proofs and tree structures of all three roots within some json structure.  this prop would be simply emitted.

The point of including this would be to primarily create a dashboard that tracks the data worker can now show the contents of the bundle as it can fetch both the leafs (manual verification) and proofs (manual submission). obviously, though, the bots should always be re-generating the full tree when proposing bundles/executing leafs or disputing other proposals.

src/dataworker/Dataworker.ts
nicholaspai: @chrismaree what is your suggestion for replacing this syntax? I think this is type safe as this will fail if the manual creation of the RelayData forgets a key

src/dataworker/Dataworker.ts
nicholaspai: I switched these imports to use the new ones you created!

src/dataworker/Dataworker.ts
nicholaspai: Alternatively, we could run an instance of the dataworker bot that doesn't submit transactions on-chain but only publishes the roots to some persistent storage layer like `IPFS` and can be consumed by dashboards. What i'm suggesting is to make this "publish" step optional and configurable by the user. WDYT? I don't want to implement this now but leave a TODO so let's reach consensus on the best comment to leave

nicholaspai: For now, I've added a `publishRoots` public function that could be called to build all roots and then publish them. This could be used a no-on-chain-write version of the dataworker that just publishes the roots to some data layer

PR #20, opened 2022-04-05
refactor: Remove redundant `validateFill` loops and change logic for fetching unfilled deposit amount to match UMIP

nicholaspai:
There are three features in this PR:

1. Remove an unnecessary loop of `SpokePoolClient.validateFillForDeposit()` calls
2. Change how `getValidUnfilledAmountForDeposit` is implemented
3. Filter out deposits with NO matching fills

Details:
1. `Remove an unnecessary loop of `SpokePoolClient.validateFillForDeposit()` calls`
    - Currently the `dataworker._loadData()` method loops through the `depositsForDestinationClient` array twice, once to store the `unfilledAmount` per deposit and once to validate fills. This PR changes the `_loadData()` function to only loop through `depositsForDestinationClient` once and simultaneously stores the `unfilledAmount` and validates fills.
    - This is possible by adding a new method to the `SpokePoolClient` that returns both the `unfilledAmount` for a deposit on the destination client, and returns all `fills` that match the deposit`.

2. `Change how `getValidUnfilledAmountForDeposit` is implemented`
    - Finally, the SpokePoolClient's `getValidUnfilledAmountForDeposit` method no longer backs into the unfilled amount by subtracting all fill amounts from the deposit amount. Instead, it orders all fills associated with a deposit by `totalFilledAmount` from largest to smallest, and uses the largest one to simply return `amount - totalFilledAmount = unfilledAmount`

3. `Filter out deposits with NO matching fills`
    - Dataworker should skip deposits with no matching fills when constructing an `unfilledDeposit` array. This avoids the case where a fill for a deposit was included previously, and a slow relay for the deposit was already included.


src/dataworker/Dataworker.ts
```diff
@@ -46,19 +43,41 @@ export class Dataworker {
...
+        // latest valid fill for the deposit did not completely fill it, then store the deposit as an "unfilled" deposit
+        // that we'll want to include in a list of slow relays to execute.
+        const validFillsOnDestinationChain: Fill[] = [];
+        const unfilledDepositsForDestinationChain: UnfilledDeposit[] = [];
+        depositsForDestinationChain.forEach((deposit) => {
+          const { fills: validFillsForDeposit, unfilledAmount } = destinationClient.getValidFillsForDeposits(deposit);
+
+          // If there are no fills associated with this deposit, then we will ignore it and not treat it as an unfilled
+          // deposit.
+          if (validFillsForDeposit.length > 0) {
```
nicholaspai, line 65: All of this function is simply refactored except for this line where we ignore any `deposits` with no matching `validFills`.

src/dataworker/Dataworker.ts
```diff
@@ -46,19 +43,42 @@ export class Dataworker {
...
           originChainId,
           destinationChainId,
         });
 
-        const unfilledDepositsForDestinationChain = depositsForDestinationChain
-          .map((deposit) => {
-            return { deposit, unfilledAmount: destinationClient.getValidUnfilledAmountForDeposit(deposit) };
-          })
-          .filter((deposit) => deposit.unfilledAmount.gt(0));
+        // For each deposit on the destination chain, fetch all valid fills associated with it and store it to a
```
mrice32, line 54: General comment: the way I envisioned this working was that we would _start_ with the fills in the block range and then look up the associated deposit for each to determine validity.

I tried to think of edge cases with the current design, though, and as long as we start with _all_ deposits and only search for matching fills _within the block range_, it should ultimately result in the same output.

I do see a potential efficiency issue with this approach: you will _always_ have to search _all the way back_ on every origin chain for deposits. If you start with the fills within the range and look for associated deposits, you can have a fairly short lookback for deposits to start and only go further back on a specific chain if you find a fill where you can't find an associated deposit (which should be rare).

Thoughts?

mrice32 (approved):
LGTM! A few high level comments

src/dataworker/Dataworker.ts
```diff
@@ -46,19 +43,42 @@ export class Dataworker {
         // amount remaining equal to the full deposit amount minus any valid fill amounts.
         // Remove any deposits that have no unfilled amount (i.e that have an unfilled amount of 0) and append the
         // remaining deposits to the unfilledDeposits array.
-        const depositsForDestinationChain = originClient.getDepositsForDestinationChain(destinationChainId);
+        const depositsForDestinationChain: Deposit[] = originClient.getDepositsForDestinationChain(destinationChainId);
```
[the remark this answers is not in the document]
nicholaspai: There will be a lookback but i haven't incorporated `bundleBlockNumbers` into this function yet. I was going to add all that timestamping stuff in a separate PR see the comment at top of function

chrismaree (approved):

src/dataworker/Dataworker.ts
```diff
@@ -42,96 +37,71 @@ export class Dataworker {
...
-            // fee % will be reset to 0 by the SpokePool contract, however we still need to explicitly filter slow
-            // relays out of the relayer refund array because its possible that a deposit is submitted with a relayer
-            // fee % set to 0.
-            if (fill.isSlowRelay) continue;
-            else if (destinationClient.validateFillForDeposit(fill, deposit)) return true;
-            else continue;
+        fillsForOriginChain.forEach((fill) => {
+          const matchedDeposit: Deposit = originClient.getDepositForFill(fill);
+          if (matchedDeposit) {
+            // FillRelay events emitted by slow relay executions will usually not match with any deposits because the
```
mrice32, line 52: I'm a little confused as to why this is true?

src/dataworker/Dataworker.ts
```diff
@@ -42,96 +37,71 @@ export class Dataworker {
...
+            // relayer fee % will be reset to 0 by the SpokePool contract, however we still need to explicitly filter slow
+            // relays out because its possible that a deposit is submitted with a relayer fee % set to 0.
+            if (!fill.isSlowRelay) assign(fillsToRefund, [fill.repaymentChainId, fill.relayer], [fill]);
+            // Save this deposit as an unfilled deposit and update its unfilled amount if this fill occurred later
```
mrice32, line 56: Not sure if we should handle this here, but I think we need to deal with the following case:
1. Epoch 1: deposit and partial fill
2. Epoch 2: liveness for epoch 1's bundle + another partial fill

The bundle for epoch 2 will not realize that a slow relay will be coming after the epoch 1's liveness ends. This might cause us to send duplicate slow relays for the same deposit.

I think the way to prevent this (and I need to double check that this is in the UMIP) is _only_ slow relaying for a particular bundle if:
1. One or more fills occur in the epoch for that deposit.
2. None of those relays have `totalFilledAmount` == `amount` (aka full fill)
3. None of those relays have `fillAmount` == `totalFilledAmount` && `fillAmount > 0`.

Thoughts?

Not sure if this filtering logic is meant to be in this method or should be done by the caller, but just wanted to call it out.

mrice32 (approved):
LGTM! Only minor questions/comments

src/dataworker/Dataworker.ts
```diff
@@ -42,96 +37,71 @@ export class Dataworker {
...
-            // fee % set to 0.
-            if (fill.isSlowRelay) continue;
-            else if (destinationClient.validateFillForDeposit(fill, deposit)) return true;
-            else continue;
+        fillsForOriginChain.forEach((fill) => {
+          const matchedDeposit: Deposit = originClient.getDepositForFill(fill);
+          if (matchedDeposit) {
+            // FillRelay events emitted by slow relay executions will usually not match with any deposits because the
+            // relayer fee % will be reset to 0 by the SpokePool contract, however we still need to explicitly filter slow
+            // relays out because its possible that a deposit is submitted with a relayer fee % set to 0.
```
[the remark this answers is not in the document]
nicholaspai: No there isn't, why would we filter that out?

src/dataworker/Dataworker.ts
nicholaspai: Its not really an issue and the comment is probably misleading. Basically we should filter out any relays from the `fillsToRefund` array that are slow relays (i.e. `isSlowRelay == true`) right? So, usually any slow relays won't match with a deposit because the slow relay event is emitted with  an overridden `relayerFeePct == 0` which will mismatch with any deposit where `relayerFeePct` was set to something different to start. However, if a deposit has set `relayerFeePct = 0`, then it will match with a slow fill (i.e. `isSlowRelay == true`) because all the deposit props will match with the fill props. Does that make sense?

src/dataworker/Dataworker.ts
nicholaspai: > I think the way to prevent this (and I need to double check that this is in the UMIP) is only slow relaying for a particular bundle if:

Cases (1) and (2) are already implemented, right? As for case (3), I don't understand the point of this fully. 

Also, I don't think this is the worst case. Duplicate slow relays are harmless, as long as slow relay merkle root can be constructed deterministically for a bundle block range:

> This might cause us to send duplicate slow relays for the same deposit.

mrice32: They aren't completely harmless IMO, as they send more funds than necessary. Also, if there were duplicate slow relays, we would need to account for that in future bundles to ensure that money gets returned to mainnet.

mrice32: 3 basically ensures that this is the _first_ partial fill. By doing this, we make sure that we only slow relay once. If it was slow relayed in a previous bundle, then all the fills in the current bundle would not be taking the fill from 0 to nonzero.

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/20#discussion_r846389134

src/dataworker/Dataworker.ts
```diff
@@ -42,68 +37,72 @@ export class Dataworker {
...
-          assign(fillsToRefund, [fill.repaymentChainId, fill.relayer], [fill])
-        );
       }
     }
 
+    // For each deposit with a matched fill, figure out the unfilled amount that we need to slow relay. We will filter
+    // out any deposits that are fully filled, or any deposits that were already slow relayed in a previous epoch.
+    const unfilledDeposits = Object.keys(unfilledDepositsForOriginChain)
+      .map((_originChainIdPlusDepositId: string): UnfilledDeposit => {
+        const _unfilledDeposits = unfilledDepositsForOriginChain[_originChainIdPlusDepositId];
```
mrice32, line 86:
```suggestion
    const unfilledDeposits = Object.values(unfilledDepositsForOriginChain)
      .map((_unfilledDeposits): UnfilledDeposit => {
```

src/dataworker/Dataworker.ts
```diff
@@ -42,68 +37,72 @@ export class Dataworker {
...
+        // Remove deposits where there isn't a fill with fillAmount == totalFilledAmount && fillAmount > 0. This ensures
+        // that we'll only be slow relaying deposits where the first fill
+        // (i.e. the one with fillAmount == totalFilledAmount) is in this epoch.
+        if (
+          !_unfilledDeposits.some((_unfilledDeposit: UnfilledDeposit) => _unfilledDeposit.hasFirstPartialFill === true)
+        )
+          return { unfilledAmount: toBN(0), deposit: undefined };
+        // Take the smallest unfilled amount since each fill can only decrease the unfilled amount.
+        _unfilledDeposits.sort((unfilledDepositA, unfilledDepositB) =>
+          unfilledDepositA.unfilledAmount.gt(unfilledDepositB.unfilledAmount) ? 1 : -1
```
mrice32, line 99: nit: in general, prefer sorting functions that don't depend on the ordering of the inputs. `cmp` returns 0 if they are equal: https://www.npmjs.com/package/bn.js. This can be possible I think if there are 0-sized fills.

```suggestion
          unfilledDepositA.unfilledAmount.cmp(unfilledDepositB.unfilledAmount)
```

src/clients/SpokePoolClient.ts
```diff
@@ -70,21 +70,33 @@ export class SpokePoolClient {
...
       .filter((fill) => fill.depositId === deposit.depositId) // Only select the associated fill for the deposit.
       .filter((fill) => this.validateFillForDeposit(fill, deposit)); // Validate that the fill was valid for the deposit.
 
-    if (!fills.length) return deposit.amount; // If no fills then the full amount is remaining.
-    return deposit.amount.sub(fills.reduce((total: BigNumber, fill: Fill) => total.add(fill.fillAmount), toBN(0)));
+    if (fills.length === 0) return deposit.amount; // If no fills then the full amount is remaining.
+
+    // Order fills by totalFilledAmount and then return the first fill's full deposit amount minus total filled amount.
+    const fillsOrderedByTotalFilledAmount = fills.sort((fillA, fillB) =>
+      fillB.totalFilledAmount.gt(fillA.totalFilledAmount) ? 1 : -1
```
mrice32, line 88:
```suggestion
      fillB.totalFilledAmount.cmp(fillA.totalFilledAmount)
```

mrice32 (approved):
LGTM! A few minor comments, but should be basically ready to go!

src/dataworker/Dataworker.ts
nicholaspai: Implemented separately

src/dataworker/Dataworker.ts
nicholaspai: So it looks like ethers' `BigNumber` library doesn't expose `cmp`, would it be equivalent to just add the additional return value of 0 if they are equal?

src/clients/SpokePoolClient.ts
nicholaspai: Because ethers' BigNumber doesn't expose `cmp` from the underlying `BN.js` lib I think this sorting function is equivalent:

```js
const fillsOrderedByTotalFilledAmount = fills.sort((fillA, fillB) =>
      fillB.totalFilledAmount.gt(fillA.totalFilledAmount)
        ? 1
        : fillB.totalFilledAmount.lt(fillA.totalFilledAmount)
        ? -1
        : 0
    );
```

src/dataworker/Dataworker.ts
mrice32: Yep!

mrice32:
Agree, you can just return 0 if equal.

mrice32 (approved):
https://github.com/across-protocol/relayer-v2/pull/20#issuecomment-1094595814

src/clients/SpokePoolClient.ts
```diff
@@ -70,21 +70,37 @@ export class SpokePoolClient {
...
+    // Order fills by totalFilledAmount and then return the first fill's full deposit amount minus total filled amount.
+    const fillsOrderedByTotalFilledAmount = fills.sort((fillA, fillB) =>
+      fillB.totalFilledAmount.gt(fillA.totalFilledAmount)
+        ? 1
+        : fillB.totalFilledAmount.lt(fillA.totalFilledAmount)
+        ? -1
+        : 0
+    );
+    const lastFill = fillsOrderedByTotalFilledAmount[0];
+    return lastFill.amount.sub(lastFill.totalFilledAmount);
```
chrismaree, line 95: ^

PR #23, opened 2022-04-08
feat: add tokenClient and additional refactor

chrismaree:
This PR does the following:
1. Bump the contract importing logic to use the latest implementation by pulling from typechain
2. refactor how pool utilization is calculated to be more consistant between the clients
3. adds "L1 token info" logic to the hub client to add details to logs.
4. refacts how "is updated" is defined in the clients.
5. moved the relayer client model to using a ` RelayerClients` type that includes all clients (relevant as we now have 5 clients kinds) 
6. exported deposit routes from spoke clients
7. create a token client that stores the `balance` and `allowance` of each token exported from 6.
8. using 7, monitors the relayer decides if it has enough balance when doing a relay. 
9. if 8 is insufficient, the token client tracks this and logs it. 
10. added approval logic for each origin token on each chain including pretty logs.
11. added lots of tests for the above.


test/RateModelClient.RealizedLpFeePct.ts
```diff
@@ -1,21 +1,6 @@
...
-import {
-  expect,
-  ethers,
-  Contract,
-  SignerWithAddress,
-  setupTokensForWallet,
-  assertPromiseError,
-  toBNWei,
-  toWei,
-} from "./utils";
```
chrismaree, line 18: This was triggering me so I changed it.

src/relayer/Relayer.ts
```diff
@@ -20,70 +19,101 @@ export class Relayer {
       this.logger.debug({ at: "Relayer", message: "Filling deposits", number: unfilledDeposits.length });
     else this.logger.debug({ at: "Relayer", message: "No unfilled deposits" });
 
-    // Iterate over all unfilled deposits. For each unfilled deposit add a fillRelay tx to the multiCallBundler.
-    for (const unfilledDeposit of unfilledDeposits) {
-      // TODO: right now this method will fill the whole amount of the relay. Next iteration should consider the wallet balance.
-      this.multiCallBundler.enqueueTransaction(this.fillRelay(unfilledDeposit));
+    // Iterate over all unfilled deposits. For each unfilled deposit: a) check that the token balance client has enough
```
chrismaree, line 22: note this logic change.

src/relayer/Relayer.ts
```diff
@@ -20,70 +19,101 @@ export class Relayer {
...
-        // remaining deposits to the unfilledDeposits array.
+        // Remove any deposits that have no unfilled amount and append the remaining deposits to unfilledDeposits array.
         unfilledDeposits.push(...unfilledDepositsForDestinationChain.filter((deposit) => deposit.unfilledAmount.gt(0)));
       }
     }
 
     return unfilledDeposits;
   }
+
+  private handleTokenShortfall() {
```
chrismaree, line 83: note this log production and how we handle token shortfall.

src/clients/TokenClient.ts
```diff
@@ -0,0 +1,147 @@
+import { BigNumber, winston, assign, ERC20, Contract, toBN, MAX_SAFE_ALLOWANCE } from "../utils";
```
chrismaree, line 1: new client.

src/clients/HubPoolClient.ts
```diff
@@ -67,6 +65,11 @@ export class HubPoolClient {
     return this.getTokenInfoForL1Token(this.getL1TokenForDeposit(deposit));
   }
 
+  getTokenInfo(chainId: number | string, tokenAddress: string): L1Token {
+    const deposit = { originChainId: parseInt(chainId.toString()), originToken: tokenAddress } as Deposit;
```
nicholaspai, line 69: What is this syntax? How can you only define `orighinChainId` and `originToken` to create a deposit? Also its probably cleaner to just expect `chainId` to be a number or string and the `parseInt(x.toString())` in the calling code

chrismaree: I'm creating an instance of the `Deposit` type and only providing `originChainId` and `originToken`. When calling `getTokenInfoForDeposit` these are the only two props used. this lets me simply pass in the things I _have_ (the `TokenClient` does not store anything other than these two props so this lets me use the same method via a cast).

src/utils/ContractUtils.ts
```diff
@@ -1,15 +1,20 @@
...
+    if (contractName === "SpokePool") contractName = castSpokePoolName(networkId);
 
-  const artifact = getContractArtifact(contractName, networkId);
-  if (!artifact) throw new Error(`Could not find artifact for contract ${contractName} on ${networkId}`);
-  return new Contract(artifact.address, artifact.abi, signer);
+    const address = getDeployedAddress(contractName, networkId);
+    const artifact = typechain[`${[contractName.replace("_", "")]}__factory`];
+    return new Contract(address, artifact.abi, signer);
+  } catch (error) {
+    console.log("e", error);
```
nicholaspai, line 15: leftover log

nicholaspai (approved):
LGTM

PR #24, opened 2022-04-11
Add profitability module

chrismaree:
Adds profitability module to relayer to consider relay profit before executing.


src/clients/ProfitClient.ts
```diff
@@ -0,0 +1,122 @@
+import { BigNumber, winston, toBNWei, toBN, assign } from "../utils";
+import { HubPoolClient } from ".";
+import { Deposit, L1Token } from "../interfaces";
+import { Coingecko } from "@uma/sdk";
+
+// Define the minimum revenue, in USD, that a relay must yield in order to be considered "profitable". This is a short
+// term solution to enable us to avoid DOS relays that yield negative profits. In the future this should be updated
+// to actual factor in the cost of sending transactions on the associated target chains.
+const chainIdToMinRevenue = {
```
chrismaree, line 9: this is the temp workaround we've spoken about before. at minimum each of these chains must produce the amount of USD revenue defined in this const object.

src/relayer/Relayer.ts
```diff
@@ -99,8 +108,35 @@ export class Relayer {
...
+          `${getNetworkName(deposit.originChainId)} to ${getNetworkName(deposit.destinationChainId)}` +
+          ` and an unfilled amount of  ${formatFunction(fillAmount)} ${symbol} is unprofitable!\n`;
+      });
+    });
+
+    this.logger.warn({ at: "Relayer", message: "Not relaying unprofitable deposits 🙅‍♂️!", mrkdwn });
+
+    this.clients.profitClient.clearUnprofitableFills();
 
+    // Note that this implementation right now will result in each run of the bot logging for insufficient token balance.
```
chrismaree, line 138: this note is important.

src/relayer/Relayer.ts
```diff
@@ -81,12 +87,15 @@ export class Relayer {
     return unfilledDeposits;
   }
 
+  // TODO: that the implementations below for both methods will produce logs on each iteration of the bot. This should
```
chrismaree, line 90: this note is important. we can fix both of these in a subsequent PR.

src/clients/ProfitClient.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  getAllPrices() {
+    return this.tokenPrices;
+  }
+
+  getPriceOfToken(token: string) {
+    if (!this.tokenPrices[token]) {
+      this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
+      return toBN(0);
```
mrice32, line 45: Just double checking my understanding, a price of 0 means we will _not_ relay this token, right?

src/clients/ProfitClient.ts
```diff
@@ -0,0 +1,122 @@
...
+  isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
+    if (toBN(this.relayerDiscount).eq(toBNWei(1))) {
+      this.logger.debug({ at: "ProfitClient", message: "Relayer discount set to 100%. Accepting relay" });
+      return true;
+    }
+
+    if (toBN(deposit.relayerFeePct).eq(toBN(0))) {
+      this.logger.debug({ at: "ProfitClient", message: "Deposit set 0 relayerFeePct. Rejecting relay" });
+      return false;
+    }
```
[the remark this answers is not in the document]
chrismaree: yup! that's why I added it :)

src/clients/ProfitClient.ts
```diff
@@ -0,0 +1,122 @@
...
+    const { decimals, address: l1Token } = this.hubPoolClient.getTokenInfoForDeposit(deposit);
+    const tokenPriceInUsd = this.getPriceOfToken(l1Token);
+    const fillRevenueInRelayedToken = toBN(deposit.relayerFeePct).mul(fillAmount).div(toBN(10).pow(decimals));
+    const fillRevenueInUsd = fillRevenueInRelayedToken.mul(tokenPriceInUsd).div(toBNWei(1));
+    const discount = toBNWei(1).sub(this.relayerDiscount);
```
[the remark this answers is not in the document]
chrismaree: not quite. consider it in the following line:
```
const discount = toBNWei(1).sub(this.relayerDiscount);
    const minimumAcceptableRevenue = chainIdToMinRevenue[deposit.destinationChainId].mul(discount).div(toBNWei(1));
```
what I've called "discount" is actually how much the `chainIdToMinRevenue` is scaled _down_ by, indicating the minimum acceptable revenue. I think the variable name `discount` is what is making this confusing.

I've renamed this to `revenueScalar` and added a comment to be more clear.

src/clients/ProfitClient.ts
chrismaree: we can always increase it. my thinking with 10 USD is it is above what you'd do to spam the bot but is ~ what a deposit is likely to cost. (an erc20 transfer at 30 gwei is like 7 USD)

PR #25, opened 2022-04-12
feat: Implement relayer refund root construction

nicholaspai:
- Change key of `fillsToRefund` from `relayerAddress` to `destinationToken` since we need to group refunds for a given `repaymentChainId` by `destinationToken`

Follow up PR will implement:
- Implementing `amountToReturn` computation and logic

### DIFF
--- a/src/clients/RelayerClientHelper.ts
+++ b/src/clients/RelayerClientHelper.ts
@@ -2,13 +2,10 @@ import winston from "winston";
 import { getProvider, getSigner, getDeployedContract, Contract } from "../utils";
 import { SpokePoolClient, HubPoolClient, RateModelClient, TokenClient, MultiCallerClient } from ".";
 import { RelayerConfig } from "../relayer/RelayerConfig";
+import { Clients } from "./ClientHelper";
 
-export interface RelayerClients {
-  spokePoolClients: { [chainId: number]: SpokePoolClient };
-  hubPoolClient: HubPoolClient;
-  rateModelClient: RateModelClient;
+export interface RelayerClients extends Clients {
   tokenClient: TokenClient;
-  multiCallerClient: MultiCallerClient;
 }
 
 export function constructRelayerClients(logger: winston.Logger, config: RelayerConfig): RelayerClients {
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -213,7 +213,7 @@ export class Dataworker {
     return indexedLeaves.length > 0 ? buildRelayerRefundTree(indexedLeaves) : null;
   }
 
-  async buildPoolRebalanceRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
+  buildPoolRebalanceRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
     const { fillsToRefund, deposits, unfilledDeposits } = this._loadData();
 
     // For each destination chain ID key in unfilledDeposits
@@ -263,6 +263,4 @@ export class Dataworker {
   async executeRelayerRefundLeaves() {
     // TODO:
   }
-
-  // Private helper functions
 }
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -9,7 +9,6 @@ import {
   buildRelayerRefundTree,
   toBNWei,
 } from "../utils";
-import { SpokePoolClient, HubPoolClient } from "../clients";
 import { SpokePoolClient, HubPoolClient, MultiCallerClient } from "../clients";
 import { FillsToRefund, RelayData, UnfilledDeposit, Deposit, Fill } from "../interfaces/SpokePool";
 import { BundleEvaluationBlockNumbers } from "../interfaces/HubPool";
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -170,7 +170,7 @@ export class Dataworker {
         const sortedRefundAddresses = Object.keys(refunds).sort((addressA, addressB) => {
           if (refunds[addressA].gt(refunds[addressB])) return -1;
           if (refunds[addressA].lt(refunds[addressB])) return 1;
-          const sortOutput = this.sortAddresses(addressA, addressB);
+          const sortOutput = this.compareAddresses(addressA, addressB);
           if (sortOutput !== 0) return sortOutput;
           else throw new Error("Unexpected matching address");
         });
@@ -199,9 +199,10 @@ export class Dataworker {
       .sort((leafA, leafB) => {
         if (leafA.chainId !== leafB.chainId) {
           return leafA.chainId - leafB.chainId;
-        } else if (this.sortAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress) !== 0) {
-          return this.sortAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress);
-        } else return leafA.groupIndex - leafB.groupIndex;
+        } else if (this.compareAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress) !== 0) {
+          return this.compareAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress);
+        } else if (leafA.groupIndex !== leafB.groupIndex) return leafA.groupIndex - leafB.groupIndex;
+        else throw new Error("Unexpected leaf group indices match");
       })
       .map((leaf: RelayerRefundLeafWithGroup, i: number): RelayerRefundLeaf => {
         delete leaf.groupIndex; // Delete group index now that we've used it to sort leaves for the same
@@ -264,7 +265,7 @@ export class Dataworker {
   }
 
   // Private helper functions
-  sortAddresses(addressA: string, addressB: string) {
+  compareAddresses(addressA: string, addressB: string) {
     // Convert address strings to BigNumbers and then sort numerical value of the BigNumber.
     const bnAddressA = BigNumber.from(addressA);
     const bnAddressB = BigNumber.from(addressB);
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -5,6 +5,7 @@ import {
   MerkleTree,
   toBN,
   RelayerRefundLeaf,
+  RelayerRefundLeafWithGroup,
   BigNumber,
   buildRelayerRefundTree,
   toBNWei,
@@ -21,18 +22,21 @@ export class Dataworker {
     readonly logger: winston.Logger,
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly hubPoolClient: HubPoolClient,
-    readonly multiCallerClient: MultiCallerClient | any
+    readonly multiCallerClient: MultiCallerClient | any,
+    readonly maxRefundsPerLeaf: number = 25
   ) {}
 
   // Common data re-formatting logic shared across all data worker public functions.
   _loadData(/* bundleBlockNumbers: BundleEvaluationBlockNumbers */): {
     unfilledDeposits: UnfilledDeposit[];
     fillsToRefund: FillsToRefund;
+    deposits: Deposit[];
   } {
     if (!this.hubPoolClient.isUpdated) throw new Error(`HubPoolClient not updated`);
 
     const unfilledDepositsForOriginChain: { [originChainIdPlusDepositId: string]: UnfilledDeposit[] } = {};
     const fillsToRefund: FillsToRefund = {};
+    const deposits: Deposit[] = [];
 
     const allChainIds = Object.keys(this.spokePoolClients);
     this.logger.debug({ at: "Dataworker", message: `Loading deposit and fill data`, chainIds: allChainIds });
@@ -48,6 +52,9 @@ export class Dataworker {
         const destinationClient = this.spokePoolClients[destinationChainId];
         if (!destinationClient.isUpdated)
           throw new Error(`destination SpokePoolClient with chain ID ${destinationChainId} not updated`);
+
+        // Store all deposits, for use in constructing a pool rebalance root.
+        deposits.push(...originClient.getDepositsForDestinationChain(destinationChainId));
 
         // For each fill within the block range, look up associated deposit.
         const fillsForOriginChain: Fill[] = destinationClient.getFillsForOriginChain(Number(originChainId));
@@ -119,6 +126,7 @@ export class Dataworker {
     // Remove deposits that have been fully filled from unfilled deposit array
     return {
       fillsToRefund,
+      deposits,
       unfilledDeposits,
     };
   }
@@ -166,7 +174,7 @@ export class Dataworker {
     const { fillsToRefund } = this._loadData();
     if (Object.keys(fillsToRefund).length === 0) return null;
 
-    const leaves: RelayerRefundLeaf[] = [];
+    const leaves: RelayerRefundLeafWithGroup[] = [];
 
     // We'll construct a new leaf for each { repaymentChainId, L2TokenAddress } unique combination.
     Object.keys(fillsToRefund).forEach((repaymentChainId: string) => {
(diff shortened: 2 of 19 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -148,14 +160,62 @@ export class Dataworker {
...
 
-    // For each repayment chain ID key in fillsToRefund
-    //     Group by refundAddress, and for each refund address
-    //         Order fills by fillAmount
-    //     Make Leaf for repayment chain ID
-    // Construct root
+    const leaves: RelayerRefundLeaf[] = [];
+
+    // We'll construct a new leaf for each { repaymentChainId, L2TokenAddress } unique combination.
+    Object.keys(fillsToRefund).forEach((repaymentChainId: string) => {
```
nicholaspai, line 172: I tried for a long time to think of a more concise way to express this loop, any ideas?

src/dataworker/Dataworker.ts
```diff
@@ -1,4 +1,15 @@
-import { winston, assign, buildSlowRelayTree, MerkleTree, toBN } from "../utils";
+import {
```
chrismaree, line 1: nit: please do two imports to make this 2 lines rather than 12. something like this: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/test/Relayer.IterativeFill.ts#L1

src/dataworker/Dataworker.ts
```diff
@@ -1,4 +1,15 @@
...
+  toBN,
+  RelayerRefundLeaf,
+  RelayerRefundLeafWithGroup,
+  BigNumber,
+  buildRelayerRefundTree,
+  toBNWei,
+} from "../utils";
 import { SpokePoolClient, HubPoolClient, MultiCallerClient } from "../clients";
 import { FillsToRefund, RelayData, UnfilledDeposit, Deposit, Fill } from "../interfaces/SpokePool";
 import { BundleEvaluationBlockNumbers } from "../interfaces/HubPool";
```
chrismaree, line 15: nit: both of these lines can be condensed to just importing from `interfaces` as the interface directory has an export `index.ts` file here: https://github.com/across-protocol/relayer-v2/blob/master/src/interfaces/index.ts

src/dataworker/Dataworker.ts
```diff
@@ -11,16 +22,21 @@ export class Dataworker {
     readonly logger: winston.Logger,
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly hubPoolClient: HubPoolClient,
-    readonly multiCallerClient: MultiCallerClient | any
+    readonly multiCallerClient: MultiCallerClient | any,
```
chrismaree, line 25: nit (I know this was done before this PR) but we should revert to a pattern of passing around an object of clients. it's much cleaner. see here: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/Relayer.ts#L10

ideally, you define a `DataWorkerClientHelper`, just like the `RelayerClientHelper` defined here: https://github.com/across-protocol/relayer-v2/blob/master/src/clients/RelayerClientHelper.ts

that will define the interface of the clients the data worker uses as well as (when you get there) defining a method for ` constructRelayerClients` and `updateRelayerClients` which the `index.ts` file of the `DataWorker` will use.

src/dataworker/Dataworker.ts
```diff
@@ -11,16 +22,21 @@ export class Dataworker {
...
+    readonly maxRefundsPerLeaf: number = 25
```
chrismaree, line 26:
```suggestion
    readonly maxRefundsPerLeaf: number = 25 // TODO: this will be removed in the future when the `ConfigStoreClient` has been implemented.
```

this is the perfect example of where we will use the `ConfigStore` contract: it will contain this exact kind of setting, as well as other settings for defining how bundles are built.

what you have for now is fine but we should just indicate that this is going to change in the future.

src/dataworker/Dataworker.ts
```diff
@@ -107,6 +126,7 @@ export class Dataworker {
     // Remove deposits that have been fully filled from unfilled deposit array
     return {
       fillsToRefund,
+      deposits,
       unfilledDeposits,
     };
```
chrismaree, line 131:
```suggestion
    return { fillsToRefund, deposits, unfilledDeposits };
```

src/utils/MerkleTreeUtils.ts
```diff
@@ -37,4 +37,4 @@ export async function buildRelayerRefundTree(relayerRefundLeaves: RelayerRefundL
   return new MerkleTree<RelayerRefundLeaf>(relayerRefundLeaves, hashFn);
 }
 
-export { MerkleTree };
```
chrismaree, line 40: unrelated to the changes here but it looks like the method `buildRelayerRefundTree` does not need to be `async`. `await` is called on `getParamType` but this method is not async. If you remove this from being `await` then `buildRelayerRefundTree` will no longer be async. This has a pretty big impact in making `buildRelayerRefundRoot` in the `Dataworker.ts` no longer async :)

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+        } else return leafA.groupIndex - leafB.groupIndex;
+      })
+      .map((leaf: RelayerRefundLeafWithGroup, i: number): RelayerRefundLeaf => {
+        const newLeaf = leaf;
+        delete newLeaf.groupIndex; // Delete group index now that we've used it to sort leaves for the same
+        // { repaymentChain, l2TokenAddress } since it doesn't exist in RelayerRefundLeaf
+        return { ...newLeaf, leafId: toBN(i) };
+      });
+
+    return indexedLeaves.length > 0 ? await buildRelayerRefundTree(indexedLeaves) : null;
```
chrismaree, line 232: I'm pretty sure this does not need to be async. this will make this whole method synchronous, which would be ideal, as this is really just processing data from the clients. see related comment: https://github.com/across-protocol/relayer-v2/pull/25/files#r850208970

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+  ): Promise<MerkleTree<RelayerRefundLeaf>> | null {
+    const { fillsToRefund } = this._loadData();
+    if (Object.keys(fillsToRefund).length === 0) return null;
 
-    // For each repayment chain ID key in fillsToRefund
-    //     Group by refundAddress, and for each refund address
-    //         Order fills by fillAmount
-    //     Make Leaf for repayment chain ID
-    // Construct root
+    const leaves: RelayerRefundLeafWithGroup[] = [];
```
chrismaree, line 177: it might help to name this variable `relayerRefundLeafs`

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
 
-  async buildRelayerRefundRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
-    this._loadData();
+  async buildRelayerRefundRoot(
+    bundleBlockNumbers: BundleEvaluationBlockNumbers
+  ): Promise<MerkleTree<RelayerRefundLeaf>> | null {
+    const { fillsToRefund } = this._loadData();
+    if (Object.keys(fillsToRefund).length === 0) return null;
 
-    // For each repayment chain ID key in fillsToRefund
```
chrismaree, line 154: we likely want to keep these comments in some form, no? perhaps above the builder method? wdyt?

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
     // of roots.
   }
 
-  async buildRelayerRefundRoot(bundleBlockNumbers: BundleEvaluationBlockNumbers) {
-    this._loadData();
+  async buildRelayerRefundRoot(
```
chrismaree, line 171: do you think it makes sence to try and pull the data worker related data processing logic into their own logical modules? right now you have `buildSlowRelayRoot` and `buildRelayerRefundRoot`. soon you'll have `buildPoolRebalanceRoot` and later we'll have a whole bunch of logic for actually executing the roots. Perhaps we should create a `SlowRelayRootBuilder` RelayerRefundBuilder` and `PoolRebalanceBuilder` (or could replace the word "Builder" with "handler" or "processor" and then pull these out into logical blocks?

my main argument here is 1) testing will be separated more cleanly and 2) this file won't explode massively. 

you dont _need_ to do this now. I'm more curious what you think about trying to split this up into more smaller logical blocks.

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+            // L2 token address
+            leafId: toBN(0), // Will be updated before inserting into tree when we sort all leaves.
+            chainId: BigNumber.from(repaymentChainId),
+            amountToReturn: toBN(0), // TODO: Derive amountToReturn
+            l2TokenAddress,
+            refundAddresses: sortedRefundAddresses.slice(i, i + this.maxRefundsPerLeaf),
+            refundAmounts: sortedRefundAddresses
+              .slice(i, i + this.maxRefundsPerLeaf)
+              .map((address) => refunds[address]),
+          });
```
chrismaree, line 211: would it make sence to create a private method `buildRelayerRefundLeaf` that would take in all the props to build the leaf? might make this a bit easer to read.

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+          return addressA.localeCompare(addressB);
+        });
+
+        // Create leaf for { repaymentChainId, L2TokenAddress }, split leaves into sub-leaves if there are too many
+        // refunds.
+        for (let i = 0; i < sortedRefundAddresses.length; i += this.maxRefundsPerLeaf)
+          leaves.push({
+            groupIndex: i, // Will delete this group index after using it to sort leaves for the same chain ID and
+            // L2 token address
+            leafId: toBN(0), // Will be updated before inserting into tree when we sort all leaves.
```
chrismaree, line 203: should `leafId` be a big number? all other contexts where we refer to "ID" are not BigNumber.

test/Dataworker.buildRoots.ts
```diff
@@ -123,6 +137,184 @@ describe("Dataworker: Build merkle roots", async function () {
...
+    const deposit6 = await buildDeposit(
+      rateModelClient,
+      hubPoolClient,
+      spokePool_1,
+      erc20_1,
+      l1Token_1,
+      depositor,
+      destinationChainId,
+      amountToDeposit
+    );
```
chrismaree, line 203: I know this is going to irritate you...but can we get this down onto one line per function call? this looks messy to me. same with all the other calls in `buildDeposit` it just does not read well and makes this file waaaay bigger than it should be.

one option could be that you define a `_buildDeposit` method in this test file that always puts the `rateModelClient, hubPoolClient` parameters into the call. this should help in decreasing the line length a bunch.

test/Dataworker.buildRoots.ts
```diff
@@ -1,14 +1,27 @@
-import { expect, ethers, Contract, buildSlowRelayTree, RelayData } from "./utils";
+import {
```
chrismaree, line 1: please do multiple imports.

test/Relayer.BasicFill.ts
```diff
@@ -21,7 +21,7 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
     [owner, depositor, relayer] = await ethers.getSigners();
     ({ spokePool: spokePool_1, erc20: erc20_1 } = await deploySpokePoolWithToken(originChainId, destinationChainId));
     ({ spokePool: spokePool_2, erc20: erc20_2 } = await deploySpokePoolWithToken(destinationChainId, originChainId));
-    ({ hubPool, l1Token } = await deployAndConfigureHubPool(owner, [
```
chrismaree, line 24: why did you need to make this change?

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+          const existingFillAmountForRelayer = refunds[fill.relayer] ? refunds[fill.relayer] : toBN(0);
+          const currentRefundAmount = fill.fillAmount.mul(toBNWei(1).sub(fill.realizedLpFeePct)).div(toBNWei(1));
+          refunds[fill.relayer] = existingFillAmountForRelayer.add(currentRefundAmount);
+        });
+        // We need to sort leaves deterministically so that the same root is always produced from the same _loadData
+        // return value, so sort refund addresses by refund amount (descending) and then address (ascending).
+        const sortedRefundAddresses = Object.keys(refunds).sort((addressA, addressB) => {
+          if (refunds[addressA].gt(refunds[addressB])) return -1;
+          if (refunds[addressA].lt(refunds[addressB])) return 1;
+          return addressA.localeCompare(addressB);
```
chrismaree, line 194: we should create a `sortAddresses` method that does this, rather than doing it inline like this.

chrismaree (approved):
Overall this is awesome! I've left a bunch of nits on structure and re-factors that I'd like to see, if posible. The logic in the core method makes sence. I tried to think of simplifications but...as you pointed out...the data structure is complex and requires multiple traverses. I'll think more on how to simplify it but for now I think this is great!

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+        });
+
+        // Create leaf for { repaymentChainId, L2TokenAddress }, split leaves into sub-leaves if there are too many
+        // refunds.
+        for (let i = 0; i < sortedRefundAddresses.length; i += this.maxRefundsPerLeaf)
+          leaves.push({
+            groupIndex: i, // Will delete this group index after using it to sort leaves for the same chain ID and
```
mrice32, line 201: Won't this end up being 25 instead of 1 on the second iteration? Should this be `Math.floor(i / this.maxRefundsPerLeaf)`?

Alternatively, you could have i increment by 1 and do a multiply to get the 0, 25, 50, etc sequence.

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+    const indexedLeaves: RelayerRefundLeaf[] = [...leaves]
+      .sort((leafA, leafB) => {
+        if (!leafA.chainId.eq(leafB.chainId)) {
+          return leafA.chainId.sub(leafB.chainId).toNumber();
+        } else if (leafA.l2TokenAddress.localeCompare(leafB.l2TokenAddress) !== 0) {
+          return leafA.l2TokenAddress.localeCompare(leafB.l2TokenAddress);
+        } else return leafA.groupIndex - leafB.groupIndex;
+      })
+      .map((leaf: RelayerRefundLeafWithGroup, i: number): RelayerRefundLeaf => {
+        const newLeaf = leaf;
```
mrice32, line 226: nit: this line isn't doing anything. Since this is just a reference to the original leaf object, you're still deleting the groupIndex from the original object. Put differently, this isn't doing a copy like `{ ...leaf }` would.

Something like this might achieve what you want (avoiding modifying the original object), although it's a little complicated:
```ts
const { groupIndex, ...output } = { ...leaf, leafId: toBN(i) };
return output;
```

src/dataworker/Dataworker.ts
nicholaspai: Couldn't you also just take the output of `buildRelayerRefundRoot` and use it as input into `executeRelayerRefundLeaves` and `proposeRoot`?

src/dataworker/Dataworker.ts
nicholaspai: 25 versus 1 doesn'tmatter though right? We're going to delete the `groupIndex` property and we just need to make sure that the `groupIndex` is unique and increases for each sub leaf

src/dataworker/Dataworker.ts
nicholaspai: I'm going to create a stand-in `ConfigStoreClient` actually with a hardcoded `maxRefundsPerLeaf` class var that we can replace later

src/utils/MerkleTreeUtils.ts
nicholaspai: Can these merkle tree methods also be fixed in the `contracts-v2` repo..?

src/dataworker/Dataworker.ts
nicholaspai: But in Contracts-v2 these are big numbers?

test/Dataworker.buildRoots.ts
nicholaspai: This still doesn't condense them into one line per call, because the only two params you're removing are the two clients

test/Relayer.BasicFill.ts
nicholaspai: I originally wanted to try constructing roots with different L1 tokens, but ended up removing that logic, but I still think useful to have the helper `deploy` function return and setup multiple L1 tokens

src/dataworker/Dataworker.ts
nicholaspai: Do you think its worth staying consistent with how the tests in Contracts-V2 define these structures or no?

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +168,72 @@ export class Dataworker {
...
+    });
+
+    // Sort leaves by chain ID and then L2 token address in ascending order. Assign leaves unique, ascending ID's
+    // beginning from 0.
+    const indexedLeaves: RelayerRefundLeaf[] = [...leaves]
+      .sort((leafA, leafB) => {
+        if (!leafA.chainId.eq(leafB.chainId)) {
+          return leafA.chainId.sub(leafB.chainId).toNumber();
+        } else if (leafA.l2TokenAddress.localeCompare(leafB.l2TokenAddress) !== 0) {
+          return leafA.l2TokenAddress.localeCompare(leafB.l2TokenAddress);
```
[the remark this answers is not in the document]
nicholaspai: Fixed with same private helper functionas above

src/dataworker/Dataworker.ts
nicholaspai: ah see I am happy to delete a prop from the original object, I think i can just remove the `newLeaf = leaf` line and just `delete leaf.groupIndex`

mrice32: Oh great! Yeah, much simpler. +1

src/dataworker/Dataworker.ts
mrice32: You're right. When I originally read this I thought this was going into the leaf. No need to change.

src/dataworker/Dataworker.ts
```diff
@@ -208,4 +262,14 @@ export class Dataworker {
   async executeRelayerRefundLeaves() {
     // TODO:
   }
+
+  // Private helper functions
+  sortAddresses(addressA: string, addressB: string) {
```
mrice32, line 267: nit: I would call this `compareAddresses` or `cmpAddresses` since it's not sorting, just returning a comparison value.

src/dataworker/Dataworker.ts
```diff
@@ -148,18 +149,71 @@ export class Dataworker {
...
+
+    // Sort leaves by chain ID and then L2 token address in ascending order. Assign leaves unique, ascending ID's
+    // beginning from 0.
+    const indexedLeaves: RelayerRefundLeaf[] = [...relayerRefundLeaves]
+      .sort((leafA, leafB) => {
+        if (leafA.chainId !== leafB.chainId) {
+          return leafA.chainId - leafB.chainId;
+        } else if (this.sortAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress) !== 0) {
+          return this.sortAddresses(leafA.l2TokenAddress, leafB.l2TokenAddress);
+        } else return leafA.groupIndex - leafB.groupIndex;
```
mrice32, line 204: Should we throw an error if these group indices match?

mrice32 (approved):
Just two minor comments. LGTM!

src/dataworker/Dataworker.ts
mrice32: Yep, I can make it clearer that addresses should be sorted _by their hex value_.

src/dataworker/Dataworker.ts
nicholaspai: yea good point. Do you think we should refactor this whole sorting logic into a utility function and unit test it?

test/Dataworker.buildRoots.ts
chrismaree: Seeing I'm the one who gets triggered by this I'll try fix it once your PR is in :)

src/clients/RelayerClientHelper.ts
```diff
@@ -2,13 +2,10 @@ import winston from "winston";
...
 import { SpokePoolClient, HubPoolClient, RateModelClient, TokenClient, MultiCallerClient } from ".";
 import { RelayerConfig } from "../relayer/RelayerConfig";
+import { Clients } from "./ClientHelper";
 
-export interface RelayerClients {
-  spokePoolClients: { [chainId: number]: SpokePoolClient };
-  hubPoolClient: HubPoolClient;
-  rateModelClient: RateModelClient;
+export interface RelayerClients extends Clients {
   tokenClient: TokenClient;
```
chrismaree, line 8: this is lit

src/dataworker/Dataworker.ts
```diff
@@ -208,4 +263,14 @@ export class Dataworker {
   async executeRelayerRefundLeaves() {
     // TODO:
   }
+
+  // Private helper functions
+  compareAddresses(addressA: string, addressB: string) {
```
chrismaree, line 268: one further nit: this could be pulled into the `utils` directory. you could make a file called `AddressUtils.ts` and put it in there then export it from `index.ts ` in the `utils` dir.

src/utils/MerkleTreeUtils.ts
chrismaree: ye, for me the goal in the bots is that the ONLY async logic that needs to happen is in the clients. ALL other code should not be async and ideally is only processing/adding/subtracting/combinding event data from the clients.

I'll take a look at the other functions you are dependent on oand see if I can help out there as well.

src/dataworker/Dataworker.ts
chrismaree: for sure. this was more about the _seperation_ of logic though. Do you think this would be possible?

PR #26, opened 2022-04-13
feat: add monitor bot

amateima:
This PR adds the implementation of Across V2 monitor bot. It monitors the l1 tokens utilisation percentages and looks for unknown relayers/dataworkers. It is based on the V1 implementation


src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+    await this.computeHubPoolBlocks();
+    await this.computeSpokePoolsBlocks();
+  }
+
+  async checkUtilization(): Promise<void> {
+    this.logger.debug({ at: "AcrossMonitor#Utilization", message: "Checking for pool utilization ratio" });
+    const l1Tokens = Object.keys(this.hubPoolClient.getL1Tokens());
+    const utilizations = await Promise.all(
+      l1Tokens.map(async (l1Token) => {
+        const utilization = await this.hubPoolClient.getCurrentPoolUtilization(l1Token);
```
chrismaree, line 43: nit: this should probs be done in a promise.all to do it in a batch, otherwise this will run slowly. you dont have to do it right now but please add a comment outlining that this is how it _should_ be done with a `TODO`.

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+        return {
+          address: l1Token,
```
chrismaree, line 45: it might be better to keep this variable as `l1Token`, not `address` to be more explicit as to what you are referring to.

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+          poolCollateralSymbol: this.hubPoolClient.getTokenInfoForL1Token(l1Token).symbol,
+          utilization: utilization,
+        };
+      })
+    );
+
+    // Send notification if pool utilization is above configured threshold.
+    for (const utilization of utilizations) {
+      if (
+        toBN(utilization.utilization).gt(
```
chrismaree, line 56: I think you should change the `getCurrentPoolUtilization` method in the `HubPoolClient` to cast the return value to a BN so you dont need to cast it here.

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+  public async update() {
+    await this.hubPoolClient.update();
+    await this.computeHubPoolBlocks();
+    await this.computeSpokePoolsBlocks();
+  }
+
+  async checkUtilization(): Promise<void> {
+    this.logger.debug({ at: "AcrossMonitor#Utilization", message: "Checking for pool utilization ratio" });
+    const l1Tokens = Object.keys(this.hubPoolClient.getL1Tokens());
+    const utilizations = await Promise.all(
```
chrismaree, line 41: perhaps let's call this `l1TokenUtilizations`

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+    // Send notification if pool utilization is above configured threshold.
+    for (const utilization of utilizations) {
+      if (
+        toBN(utilization.utilization).gt(
+          toBN(this.monitorConfig.utilizationThreshold)
+            .mul(toBN(toWei("1")))
+            .div(toBN(100))
+        )
+      ) {
+        const utilizationString = toBN(utilization.utilization as string)
```
chrismaree, line 62: this is great!

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+          .mul(toBN(100))
+          .toString();
+        this.logger.warn({
+          at: "UtilizationMonitor",
+          message: "High pool utilization warning 🏊",
+          mrkdwn:
```
chrismaree, line 68: nit: it might be cleaner for you to define a `mrkdwn` variable above this warning and populate it with all the content on as few lines as posible, like done here: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/Relayer.ts#L102

you should ideally try get the definition of this variable to as few lines as posible, taking advantage of string contamination using the `${}` syntax 

(I know this is a small nitpick but I just want it to be as consistant as posible.

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+    );
+    for (const event of rootBundleEvents) {
+      if (this.monitorConfig.whitelistedAddresses.includes(event.caller)) {
+        continue;
+      }
+
+      this.logger.warn({
+        at: "UnknownRootBundleCaller",
+        message: "Across Hub Pool unknown root bundle caller warning 🥷",
+        mrkdwn:
```
chrismaree, line 98: same comment on trying to get this to be on fewer lines, if posible.

src/monitor/RootBundleProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export interface EventInfo {
+  logIndex: number;
+  blockNumber: number;
+  transactionHash: string;
+  caller: string;
+  action: string;
+}
+
+export class RootBundleProcessor {
```
chrismaree, line 19: this is interesting. a few points:

1) I think that this will overlap with some work that @nicholaspai is doing. we should make sure we can re-use this work as much as posible. @nicholaspai from this set of information you'll require access to `ProposedRootBund` so we should make sure this is as accessible as posible for this.
2) I dont think that this should be a separate module like this. ideally ALL logic that fetches events (and really does any aync calls for events) sits in the `clients` directory.
3) I dont think this should be a separate client as _all_ these events are being requested against the `HubPool`. 

Due to 2 & 3 my recommendation is to pull this logic out and place it within the existing `HubPoolClient`. Ultimately, you are re-using a fair bit of logic that is already defined in the `HubPool`. wdyt?

A counter argument to this is that this information is _only_ needed in the Monitor and so it can stay here but I think to keep things as a) clean and b) consistant as posible this should reside within the `HubPoolClient`.

src/monitor/index.ts
```diff
@@ -0,0 +1,85 @@
...
+import { HubPoolClient } from "../clients";
+import { getDeployedContract } from "../utils/ContractUtils";
+import { SpokePool } from "@across-protocol/contracts-v2";
+
+config();
+
+export async function runMonitor(logger: winston.Logger) {
+  const config = new AcrossMonitorConfig(process.env);
+  // If pollingDelay === 0 then the bot is running in serverless mode and should send a `debug` level log.
+  // Else, if running in loop mode (pollingDelay != 0), then it should send a `info` level log.
```
chrismaree, line 15: For now we won't be running things like that so are going to ignore this structure. we'll potentially update all logger instantiation later for this purpose.
```suggestion
```

src/monitor/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  logger[config.pollingDelay === 0 ? "debug" : "info"]({
```
chrismaree, line 16: I think you can get this onto 1 line as a result.
```suggestion
  logger.info({
```

src/monitor/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  const hubPoolClient = new HubPoolClient(logger, hubPool);
+  const spokeSigners = config.spokePoolChainIds
+    .map((networkId) => getProvider(networkId))
+    .map((provider) => baseSigner.connect(provider));
+  const spokePools = config.spokePoolChainIds.reduce((acc, chainId, idx) => {
+    return {
+      ...acc,
+      [chainId]: getDeployedContract("SpokePool", chainId, spokeSigners[idx]) as SpokePool,
+    };
+  }, {} as Record<number, SpokePool>);
```
chrismaree, line 35: can you please re-factor this into a helper module like done in the relayer? see [here](https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/index.ts#L18) for how the relayer `index.ts` file is constructed and [here](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/RelayerClientHelper.ts) for the `RelayerClientHelper`. Ideally, you create a `MonitorClientHelper` that returns all clients needed for the monitor to run.

As a second point, I think it makes sence to define an interface similar to the `RelayerClients` defined [here](https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/clients/RelayerClientHelper.ts#L6) for the monitor clients. Then, you simply pass around this object to the respective object instantiation.

src/monitor/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const acrossMonitor = new AcrossMonitor(logger, config, hubPoolClient, spokePools);
+
+  for (;;) {
+    await retry(
```
chrismaree, line 40: I'm sorry to make you do this but can you please remove the use of the `retry()` logic (and all `onRetry` related stuff, plus the serverless `pollingDelay === 0` logic and try to mimic what's done here: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/index.ts#L24

For now, we've not exactly figured out how we want to run these in servelress so I want to keep the interfaces, ideally, exactly the same on how the entry point on these bots work.

Equally, you can remove some of the other logs you've added.

src/monitor/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      await delay(5); // Set a delay to let the transports flush fully.
+      break;
+    }
+    logger.debug({
+      at: "AcrossMonitor#index",
+      message: "End of execution loop - waiting polling delay",
+      pollingDelay: `${config.pollingDelay} (s)`,
+    });
+    await delay(Number(config.pollingDelay));
+  }
```
chrismaree, line 84: This block of comments (and the delay logic) should be replaced with this: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/index.ts#L31

src/monitor/.env.sample
```diff
@@ -0,0 +1,13 @@
+POLLING_DELAY=
```
chrismaree, line 1: thanks for adding this!

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  async checkUnknownRootBundleCallers(): Promise<void> {
+    this.logger.debug({ at: "AcrossMonitor#RootBundleCallers", message: "Checking for unknown root bundle callers" });
+
+    const rootBundleEvents: EventInfo[] = await this.rootBundleProcessor.getRootBundleEventsInfo(
+      this.hubPoolStartingBlock,
+      this.hubPoolEndingBlock
+    );
+    for (const event of rootBundleEvents) {
+      if (this.monitorConfig.whitelistedAddresses.includes(event.caller)) {
```
chrismaree, line 91: as mentioned in the config file, I think this should be re-named to `whitelistedDataworkers`

src/monitor/AcrossMonitorConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+      ERROR_RETRIES_TIMEOUT,
+    } = env;
+
+    this.hubPoolChainId = HUB_CHAIN_ID ? Number(HUB_CHAIN_ID) : 1;
+    this.botModes = {
+      utilizationEnabled: UTILIZATION_ENABLED === "true",
+      unknownRootBundleCallersEnabled: UNKNOWN_ROOT_BUNDLE_CALLERS_ENABLED === "true",
+    };
+
+    this.whitelistedAddresses = WHITELISTED_ADDRESSES ? JSON.parse(WHITELISTED_ADDRESSES) : [];
```
chrismaree, line 50: the variable `whitelistedAddresses` is used to whitelist approved Dataworkers, right? i.e people used when producing the log `Across Hub Pool unknown root bundle caller warning`. I think, as a result, we should re-name this variable to `whitelistedDataworkers` and change in the associated places here: https://github.com/across-protocol/relayer-v2/pull/26/files#r850186640

package.json
```diff
@@ -10,8 +10,10 @@
     "@across-protocol/contracts-v2": "^0.0.43",
     "@across-protocol/sdk-v2": "^0.0.1",
     "@defi-wonderland/smock": "^2.0.7",
+    "@uma/common": "^2.19.0",
     "@uma/contracts-node": "^0.3.1",
     "@uma/financial-templates-lib": "2.25.0",
+    "async-retry": "^1.3.3",
```
chrismaree, line 16: Due to some of my comments on changing the structure here I think we dont need this _right now_. however, feel free to leave it in as it'll likely be added once we've converged on the exact execution framework for _all_ bots.

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
...
+        const utilizationString = toBN(utilization.utilization as string)
+          .mul(toBN(100))
+          .toString();
+        this.logger.warn({
+          at: "UtilizationMonitor",
+          message: "High pool utilization warning 🏊",
+          mrkdwn:
+            utilization.poolCollateralSymbol +
+            " bridge pool at " +
+            createEtherscanLinkMarkdown(utilization.address, utilization.chainId) +
```
chrismaree, line 71: please use `etherscanLink` exported from `utils` as exported [here](https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/utils/FormattingUtils.ts#L19) and done here https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/clients/MultiCallerClient.ts#L97

src/monitor/AcrossMonitor.ts
```diff
@@ -0,0 +1,193 @@
+import { SpokePool } from "@across-protocol/contracts-v2";
+import { createEtherscanLinkMarkdown, createFormatFunction, PublicNetworks } from "@uma/common";
+import { providers } from "ethers";
+import { HubPoolClient } from "../clients";
+import { toBN, toWei, winston } from "../utils";
+import { AcrossMonitorConfig } from "./AcrossMonitorConfig";
+import { RelayerProcessor } from "./RelayerProcessor";
+import { EventInfo, RootBundleProcessor } from "./RootBundleProcessor";
```
chrismaree, line 8: a fair number of these can (and should) be refactored:
1) use `etherscanLink` exported from `utils` over `createEtherscanLinkMarkdown`
2) `createFormatFunction` can be imported directly from `utils`
3) `dont import `PublicNetworks` and import `getNetworkName` and use it to get the name of a network from a given network ID

(1,2,3 should remove the imports from @uma/common completely from this file)

4) ideally, add the `provider` export to the `ethers` export from [index.ts](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/index.ts#L5) file in utils then you can import this from `utils` as well.

src/monitor/AcrossMonitorConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+  readonly utilizationThreshold: number;
+  readonly hubPoolStartingBlock: number | undefined;
+  readonly hubPoolEndingBlock: number | undefined;
+  readonly whitelistedAddresses: string[];
+  readonly whitelistedRelayers: string[];
+  readonly botModes: BotModes;
+
+  public constructor(env: ProcessEnv) {
+    const {
+      POLLING_DELAY,
```
chrismaree, line 29: once the changes are done from the `async` logic in the `index.ts` file you should clean this up to remove:
`POLLING_DELAY`, `ERROR_RETRIES`, & `ERROR_RETRIES_TIMEOUT`,

src/monitor/RelayerProcessor.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface EventInfo {
+  logIndex: number;
+  blockNumber: number;
+  transactionHash: string;
+  caller: string;
+  action: string;
+  rootBundleId: number;
+}
+
+export class RelayerProcessor {
```
chrismaree, line 14: here, I have the same comments I made in the `RootBundleProcessor`. I think that this should rather be part of the `SpokePoolClient` and follow the patterns defined there.

src/monitor/AcrossMonitorConfig.ts
chrismaree: Slight correction here: you'll need to leave `POLLING_DELAY` like done here: https://github.com/across-protocol/relayer-v2/blob/05197b9e96c0965b27b8f22cc2f609b58456bbd6/src/relayer/RelayerConfig.ts#L20

src/monitor/RootBundleProcessor.ts
chrismaree: I've thought about this a bit more and I think that I'm leaning more towards the counter-argument wherein we leave this logic within the `RelayBundleProcessor.ts, in the structure you have now. My main point is I don't think we want to complicate the other two bots and leave this logic out of them will overall help a bit with simplfication of state and logic.

src/monitor/RelayerProcessor.ts
chrismaree: see the comment I left here: https://github.com/across-protocol/relayer-v2/pull/26#discussion_r851123040

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/index.ts
amateima: done

src/monitor/index.ts
amateima: done

src/monitor/index.ts
amateima: done

src/monitor/index.ts
amateima: done

src/monitor/index.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/.env.sample
amateima: done

src/monitor/AcrossMonitorConfig.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

package.json
amateima: removed

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitor.ts
amateima: done

src/monitor/AcrossMonitorConfig.ts
amateima: done

src/monitor/RelayerProcessor.ts
amateima: done

package.json
```diff
@@ -6,10 +6,14 @@
   "author": "UMA Team",
   "license": "AGPL-3.0",
   "private": true,
+  "engines": {
```
chrismaree, line 9: why?

chrismaree, line 9: why?

chrismaree (approved):
This is excellent work! thanks for getting this in and addressing all my nits.

src/monitor/Monitor.ts
```diff
@@ -0,0 +1,176 @@
...
+      this.spokePoolsBlocks[chainId].startingBlock = startingBlock;
+      this.spokePoolsBlocks[chainId].endingBlock = endingBlock;
+    }
+  }
+
+  // Compute the starting and ending block for each chain giving the provider and the config values
+  private async computeStartingAndEndingBlock(
+    provider: providers.Provider,
+    startingBlock: number | undefined,
+    endingBlock: number | undefined
```
mrice32, line 149: nit: on the surface this seems a little confusing.

Might be clearer to name these values `configuredStartingBlock` and `configuredEndingBlock` or similar so it's clear that you're taking the configured values and computing the values that you should actually use.

Otherwise, it reads like this function is taking in the same thing that it's computing.

src/monitor/RootBundleProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+        this.hubPoolClient.hubPool.queryFilter(
+          this.hubPoolClient.hubPool.filters.ProposeRootBundleEvent(),
+          startingBlock,
+          endingBlock
+        ),
+        this.hubPoolClient.hubPool.queryFilter(
+          this.hubPoolClient.hubPool.filters.ProposeRootBundleEvent(),
+          startingBlock,
+          endingBlock
+        ),
```
mrice32, line 66: Copy paste error? All these seem like they're querying for `ProposeRootBundleEvent`?

src/monitor/RootBundleProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+        blockNumber: event.blockNumber,
+        logIndex: event.logIndex,
+        transactionHash: event.transactionHash,
+        caller,
+        action: this.eventActions[event.event],
+      };
+      eventsInfo.push(eventInfo);
+    }
+
+    // Primary sort on block number. Secondary sort on logIndex.
```
mrice32, line 105: nit: if you want a total sorting of events, you need to do a secondary sort on transactionIndex, then use logIndex as a tertiary sort. I think logIndex is the index of the log _within a particular transaction_.

mrice32 (approved):
Looks great! A few minor comments and questions.

src/monitor/index.ts
```diff
@@ -0,0 +1,36 @@
+import { config } from "dotenv";
+import { winston, processEndPollingLoop } from "../utils";
+import { Monitor } from "./Monitor";
+import { MonitorConfig } from "./MonitorConfig";
+import { constructMonitorClients } from "../clients/MonitorClientHelper";
+
+config();
+
+export async function runMonitor(logger: winston.Logger) {
```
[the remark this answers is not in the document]
amateima: good point. @chrismaree do you hve anything against making the signer argument optional [here](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/ContractUtils.ts#L7-L13) ? This way I won't need to instantiate a signer for the monitor bot clients

src/clients/MonitorClientHelper.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getDeployedContract, getProvider, getSigner, winston } from "../utils";
+import { HubPoolClient } from "./HubPoolClient";
+
+export interface MonitorClients {
+  hubPoolClient: HubPoolClient;
+  spokePools: { [chainId: number]: SpokePool };
+}
+
+export function constructMonitorClients(config: MonitorConfig, logger: winston.Logger): MonitorClients {
+  const baseSigner = getSigner();
```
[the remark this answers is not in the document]
amateima: good point. @chrismaree do you have anything against making the signer argument optional [here](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/ContractUtils.ts#L7-L13) ? This way I won't need to instantiate a signer for the monitor bot clients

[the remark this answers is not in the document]
amateima: done

src/monitor/RootBundleProcessor.ts
amateima: good catch

src/monitor/Monitor.ts
amateima: 👍

PR #27, opened 2022-04-18
Implement initial runningBalances computation in buildPoolRebalanceRoot

nicholaspai:
This PR does NOT fully implement `buildPoolRebalanceRoot`. It implements steps 1-3 in the list "To determine the amount to modify the running balances:" as described in the [ACROSS-V2 UMIP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md#constructing-the-poolrebalanceroot). 

This PR is already large so I'm going to stop here, and I've included a simple test to demonstrate how it works.

Checklist

- [X] Initiate `runningBalances` to total refund amount for all recipient addresses for `{ repaymentChainId, l2TokenAddress }`
- [X] Set `realizedLpFees`
- [X] Account for `FundsDeposited` events
- [X] Account for slow relays

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,5 +13,7 @@ module.exports = {
   },
   rules: {
     "node/no-unsupported-features/es-syntax": ["error", { ignores: ["modules"] }],
+    camelcase: "off",
+    "@typescript-eslint/camelcase": "off",
   },
 };
--- a/src/clients/DataworkerClientHelper.ts
+++ b/src/clients/DataworkerClientHelper.ts
@@ -4,3 +4,7 @@ import { Clients } from "./ClientHelper";
 export interface DataworkerClients extends Clients {
   configStoreClient: ConfigStoreClient;
 }
+
+// Used for determining which block range corresponsd to which network. In order, the block ranges passed
+// in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
+export const CHAIN_ID_LIST_INDICES = ["1", "10", "137", "288", "42161"];
--- a/src/clients/DataworkerClientHelper.ts
+++ b/src/clients/DataworkerClientHelper.ts
@@ -7,4 +7,4 @@ export interface DataworkerClients extends Clients {
 
 // Used for determining which block range corresponsd to which network. In order, the block ranges passed
 // in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
-export const CHAIN_ID_LIST_INDICES = ["1", "10", "137", "288", "42161"];
+export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -40,10 +40,10 @@ export class HubPoolClient {
   getL1TokenCounterpart(repaymentChainId: string, l2Token: string) {
     let l1Token = null;
     Object.keys(this.l1TokensToDestinationTokens).forEach((_l1Token) => {
-      if (this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token)
-        l1Token = _l1Token;
+      if (this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token) l1Token = _l1Token;
     });
-    if (l1Token === null) throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
+    if (l1Token === null)
+      throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
     return l1Token;
   }
 
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -7,6 +7,7 @@ export class HubPoolClient {
   private l1TokensToDestinationTokens: { [l1Token: string]: { [destinationChainId: number]: string } } = {};
   private l1Tokens: L1Token[] = []; // L1Tokens and their associated info.
   private proposedRootBundles: ProposedRootBundle[] = [];
+  private l1TokensToDestinationTokensWithBlock: { [l1Token: string]: { [destinationChainId: number]: [{ l2Token: string, block: number }] } } = {};
 
   public isUpdated: boolean = false;
   public firstBlockToSearch: number;
@@ -39,12 +40,18 @@ export class HubPoolClient {
     return l1Token;
   }
 
-  getL1TokenCounterpart(repaymentChainId: string, l2Token: string) {
-    const l1Token = Object.keys(this.l1TokensToDestinationTokens).find(
-      (_l1Token) => this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token
+  getL1TokenCounterpartAtBlock(l2ChainId: string, l2Token: string, block: number) {
+    const l1Token = Object.keys(this.l1TokensToDestinationTokensWithBlock).find(
+      (_l1Token) => {
+          // We assume that l2-l1 token mapping events are sorted in descending order, so find the last mapping published
+          // before the target block.
+        return this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId].find(
+          (mapping: { l2Token: string; block: number }) => mapping.l2Token === l2Token && mapping.block <= block
+        )
+      }
     );
-    if (l1Token === null)
-      throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
+    if (!l1Token)
+      throw new Error(`Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${block}!`);
     return l1Token;
   }
 
@@ -84,17 +91,21 @@ export class HubPoolClient {
 
   // This should find the ProposeRootBundle event whose bundle block number for `chain` is closest to the `block`
   // without being smaller. It returns the bundle block number for the chain.
-  getRootBundleEvalBlockNumberContainingBlock(block: number, chain: number, chainIdList: number[]): number {
+  getRootBundleEvalBlockNumberContainingBlock(block: number, chain: number, chainIdList: number[]): number | undefined {
     let endingBlockNumber: number;
-    sortEventsAscending(this.proposedRootBundles).forEach((rootBundle: ProposedRootBundle) => {
-      const bundleEvaluationBlockNumbers: BigNumber[] = rootBundle.bundleEvaluationBlockNumbers;
+    for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+      const bundleEvaluationBlockNumbers: BigNumber[] = (rootBundle as ProposedRootBundle).bundleEvaluationBlockNumbers;
       if (bundleEvaluationBlockNumbers.length !== chainIdList.length)
         throw new Error("Chain ID list and bundle block eval range list length do not match");
       const chainIdIndex = chainIdList.indexOf(chain);
       if (chainIdIndex === -1) throw new Error("Can't find fill.destinationChainId in CHAIN_ID_LIST");
-      if (bundleEvaluationBlockNumbers[chainIdIndex].gte(toBN(block)))
+      if (bundleEvaluationBlockNumbers[chainIdIndex].gte(toBN(block))) {
         endingBlockNumber = bundleEvaluationBlockNumbers[chainIdIndex].toNumber();
-    });
+        // Since events are sorted from oldest to newest, and bundle block ranges should only increase, exit as soon
+        // as we find the first block range that contains the target block.
+        break;
+      }
+    };
     return endingBlockNumber;
   }
 
@@ -112,7 +123,20 @@ export class HubPoolClient {
     for (const event of poolRebalanceRouteEvents) {
       const args = spreadEvent(event);
       assign(this.l1TokensToDestinationTokens, [args.l1Token, args.destinationChainId], args.destinationToken);
+      assign(
+        this.l1TokensToDestinationTokensWithBlock,
+        [args.l1Token, args.destinationChainId],
+        [{ l2Token: args.destinationToken, block: event.blockNumber }]
+      );
+      // Sort l2 token to l1 token mapping events in descending order so we can easily find the first mapping update 
+      // equal to or earlier than a target block. This allows a caller to look up the l1 token counterpart for an l2
+      // token at a specific block height.
+      this.l1TokensToDestinationTokensWithBlock[args.l1Token][args.destinationChainId]
+        .sort((mappingA: { block: number }, mappingB: { block: number }) => {
+            return mappingB.block - mappingA.block;
+        })
     }
+
 
     // For each enabled Lp token fetch the token symbol and decimals from the token contract. Note this logic will
     // only run iff a new token has been enabled. Will only append iff the info is not there already.
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -1,5 +1,6 @@
-import { spreadEvent, assign, Contract, winston, BigNumber, ERC20, Event } from "../utils";
+import { spreadEvent, assign, Contract, winston, BigNumber, ERC20, Event, sortEventsAscending } from "../utils";
 import { Deposit, Fill, L1Token } from "../interfaces";
+import { SpokePoolClient, CHAIN_ID_LIST_INDICES } from ".";
 
 export class HubPoolClient {
   // L1Token -> destinationChainId -> destinationToken
@@ -81,6 +82,31 @@ export class HubPoolClient {
   getTokenInfo(chainId: number | string, tokenAddress: string): L1Token {
     const deposit = { originChainId: parseInt(chainId.toString()), originToken: tokenAddress } as Deposit;
     return this.getTokenInfoForDeposit(deposit);
+  }
+
+  // TODO:
+  // Query CrossChainContractsSet events to find spoke pools at specified `block`.
+  // getCrossChainContractsForBlock(block:number): { [chainId: number]: SpokePoolClient } {
+  // }
+
+  // // Returns an ordered array of block ranges for each spoke pool. For an ordered list of `toBlocks`, this function
+  // // will find the corresponding `fromBlocks` for each chain. The order of chains can be found in CHAIN_ID_LIST.
+  // getRootBundleEvaluationBlockRange(toBlocks: number[]): number[][] {
+  //   // Find latest RootBundleExecuted event with matching chain ID while still being earlier than the toBlock.
+  //   const blockRanges: number[][] = toBlocks.map((block, i) => {
+  //     const chainId = CHAIN_ID_LIST_INDICES[i]
+  //     // Sort event block heights from latest to earliest, so we find the events right before our target block.
+  //     const precedingExecuteRootBundleEvent: Event = this.executeRootBundleEvents.sort((ex, ey) => ey.blockNumber - ex.blockNumber).find(
+  //       (e: Event) => e.args.chainId === chainId && e.blockNumber < block
+  //     );
+  //     const precedingProposeRootBundleEvent: Event = this.proposeRootBundleEvents.sort((ex, ey) => ey.blockNumber - ex.blockNumber).find(
+  //       (e: Event) => e.blockNumber < precedingExecuteRootBundleEvent.blockNumber
+  //     );
+  //   })
+  // }
+
+  getEarliestProposeRootEventAfterBlock(block: number): Event {
+    return sortEventsAscending(this.proposeRootBundleEvents).find((e: Event) => e.blockNumber > block) as Event;
   }
 
   async update() {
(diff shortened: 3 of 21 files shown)


nicholaspai:
@chrismaree @mrice32 requesting a first pass on the pool rebalance root builder before I implement the third part of `runningBalances` which accounts for `SlowRelays`. The last part seems to me the most complex bit and corresponds to step 3 in [these UMIP instructions](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md#constructing-the-poolrebalanceroot)

nicholaspai:
Also, I will refactor this a lot as I know there's a lot of re-used code, especially in the tests, but I wanted to open this up for some reviews before proceeding

src/dataworker/Dataworker.ts
```diff
@@ -42,21 +66,56 @@ export class Dataworker {
...
-            if (!fill.isSlowRelay) assign(fillsToRefund, [fill.repaymentChainId, fill.destinationToken], [fill]);
+            // Fill was validated. Save it under all blocks with the block number so we can sort it by time.
+            allFills.push(fillWithBlock);
+
+            // Handle slow relay where repaymentChainId = 0. Slow relays always pay recipient on destination chain.
+            // So, save the slow fill under the destination chain, and save the fast fill under its repayment chain.
+            const chainToSendRefundTo = fill.isSlowRelay ? fill.destinationChainId : fill.repaymentChainId;
+
+            // Save fill data and associate with repayment chain and token.
+            assign(fillsToRefund, [chainToSendRefundTo, fill.destinationToken, "fills"], [fill]);
```
nicholaspai, line 93: I refactored this method a bit because I realized that I wasn't previously counting slow relays in for the `realizedlpFee` total. I decided to add slow fills to `fillsToRefund` and then just changed `buildRelayerRefundRoot` to filter out slow fills

src/dataworker/Dataworker.ts
```diff
@@ -121,7 +181,7 @@ export class Dataworker {
       (deposit: UnfilledDeposit): RelayData => ({
         depositor: deposit.deposit.depositor,
         recipient: deposit.deposit.recipient,
-        destinationToken: deposit.deposit.depositor,
+        destinationToken: deposit.deposit.destinationToken,
```
nicholaspai, line 184: stupid bug I found, also repeated this bug in the test which made it go unnoticed

.eslintrc.js
```diff
@@ -13,5 +13,7 @@ module.exports = {
   },
   rules: {
     "node/no-unsupported-features/es-syntax": ["error", { ignores: ["modules"] }],
+    camelcase: "off",
```
chrismaree, line 16: ?

src/clients/DataworkerClientHelper.ts
```diff
@@ -4,3 +4,7 @@ import { Clients } from "./ClientHelper";
 export interface DataworkerClients extends Clients {
   configStoreClient: ConfigStoreClient;
 }
+
+// Used for determining which block range corresponsd to which network. In order, the block ranges passed
+// in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
+export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
```
chrismaree, line 10: cool. in next itteration this will be in the config store. fine to be here for now.

src/clients/HubPoolClient.ts
```diff
@@ -93,6 +126,9 @@ export class HubPoolClient {
     );
     for (const info of tokenInfo) if (!this.l1Tokens.includes(info)) this.l1Tokens.push(info);
 
+    this.proposeRootBundleEvents.push(...proposeRootBundleEvents);
```
chrismaree, line 129: nit: does it make sence to store the raw event? would it not be better to extract the relevant data, like done in other places of this contract, by using the `spreadEvent` function which extracts the _props_ of the event and casts the types, where needed?

src/clients/HubPoolClient.ts
```diff
@@ -37,6 +40,16 @@ export class HubPoolClient {
     return l1Token;
   }
 
+  getL1TokenCounterpart(repaymentChainId: string, l2Token: string) {
```
chrismaree, line 43: nit: could the be done with a `.filter`?

src/clients/HubPoolClient.ts
nicholaspai: Good point, i was going to do this eventually once I figured out how exactly I was going to use the ProposeRootBundle events. Thanks for the reminder!

.eslintrc.js
nicholaspai: I get these annoying warnings in the test all the time that `erc20_1` or `Arbitrum_SpokePool` are not camel case. Do you not?

src/clients/DataworkerClientHelper.ts
nicholaspai: Yeah its just a sample. I don't even use in this PR yet.

src/clients/HubPoolClient.ts
```diff
@@ -71,14 +82,33 @@ export class HubPoolClient {
...
+  getRootBundleEvalBlockNumberContainingBlock(block: number, chain: number, chainIdList: number[]): number {
+    let endingBlockNumber: number;
+    sortEventsAscending(this.proposedRootBundles).forEach((rootBundle: ProposedRootBundle) => {
+      const bundleEvaluationBlockNumbers: BigNumber[] = rootBundle.bundleEvaluationBlockNumbers;
+      if (bundleEvaluationBlockNumbers.length !== chainIdList.length)
+        throw new Error("Chain ID list and bundle block eval range list length do not match");
+      const chainIdIndex = chainIdList.indexOf(chain);
+      if (chainIdIndex === -1) throw new Error("Can't find fill.destinationChainId in CHAIN_ID_LIST");
+      if (bundleEvaluationBlockNumbers[chainIdIndex].gt(toBN(block)))
+        // TODO: Can this be a `gte`?
```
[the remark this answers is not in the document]
nicholaspai: yeah I think that's right, but the UMIP sort of suggests its exclusive. Look at the last line referencing "before". Should we change that ?

"Search through past ProposeRootBundle events to find the proposal for this destinationChainId whose ending block number is the closest to this block number without being smaller. Search for all FilledRelay events that have matching amount, originChainId, destinationChainId, relayerFeePct, depositId, recipient, and depositor values before that ending block number. "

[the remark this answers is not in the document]
nicholaspai: I have a TODO referencing this on line 406 in Dataworker.ts

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +250,120 @@ export class Dataworker {
...
+          );
+          assign(
+            realizedLpFees,
+            [repaymentChainId, l1TokenCounterpart],
+            fillsToRefund[repaymentChainId][l2TokenAddress].realizedLpFees
+          );
+
+          // 3a. For any slow fills, we need to adjust the running balance in case a previous root bundle sent too many
+          // tokens to the spoke pool to pay for the slow fill, but a fill was sent before the slow relay could be
+          // executed, resulting in an excess of funds on the spoke pool. For this step, filter out repeat slow fills.
```
nicholaspai, line 282: Question for @chrismaree @mrice32 . I filter out repeat slow fills because this is the language that the UMIP says: 

> perform a historical event lookup to determine if this slow relay has been executed before by searching for past FilledRelay events with isSlowRelay equal to true, matching originChainId and depositId. If any are found, ignore this slow relay

However, I don't think its possible to execute slow fills twice, otherwise the "relay filled" error will throw [here](https://github.com/across-protocol/contracts-v2/blob/master/contracts/SpokePool.sol#L740)

Do you agree? If so should we change the UMIP?

src/dataworker/Dataworker.ts
```diff
@@ -42,21 +49,44 @@ export class Dataworker {
...
-            // FillRelay events emitted by slow relay executions will usually not match with any deposits because the
-            // relayer fee % will be reset to 0 by the SpokePool contract, however we still need to explicitly filter slow
-            // relays out because its possible that a deposit is submitted with a relayer fee % set to 0.
-            if (!fill.isSlowRelay) assign(fillsToRefund, [fill.repaymentChainId, fill.destinationToken], [fill]);
+            // Fill was validated. Save it under all blocks with the block number so we can sort it by time.
+            allFills.push(fillWithBlock);
+
+            // Handle slow relay where repaymentChainId = 0. Slow relays always pay recipient on destination chain.
+            // So, save the slow fill under the destination chain, and save the fast fill under its repayment chain.
+            const chainToSendRefundTo = fill.isSlowRelay ? fill.destinationChainId : fill.repaymentChainId;
```
mrice32, line 73: +1, great catch

src/dataworker/Dataworker.ts
```diff
@@ -42,21 +49,44 @@ export class Dataworker {
...
+
+            // Update realized LP fee and total refund amount accumulators.
+            const refundObj = fillsToRefund[chainToSendRefundTo][fill.destinationToken];
+            const refund = getRefundForFills([fill]);
+            refundObj.totalRefundAmount = refundObj.totalRefundAmount
+              ? refundObj.totalRefundAmount.add(refund)
+              : refund;
+            refundObj.realizedLpFees = refundObj.realizedLpFees
+              ? refundObj.realizedLpFees.add(getRealizedLpFeeForFills([fill]))
+              : getRealizedLpFeeForFills([fill]);
```
mrice32, line 86: I don't see any filtering by slow relay fills (isSlowRelay == true)? Does this mean if we have a slow relay fill, we add a refund amount for that fill? Shouldn't we _not_ send a refund for that fill?

src/dataworker/Dataworker.ts
mrice32: It's a great point. Does this mean we need to modify our strategy entirely then?

We need to be able to detect when a slow relay was sent but unused.

Let's say there's a $100 deposit.

It's partially filled with $50.

We build a bundle to slow relay $50.

Before our bundle passes liveness, the $50 is relayed by someone else.

Bundle is executed and the slow relay $50 is sent to the chain.

We need to be able to detect that that slow relay can never be executed and that the $50 needs to be sent back. How do we detect that? Should we look for all full fills that occur after a partial fill and see if they match a past slow relay?

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +252,120 @@ export class Dataworker {
...
+            (fill: Fill) => fill.isSlowRelay
+          );
+          const firstTimeSlowFills = slowFills.filter(
+            (slowFill: Fill) =>
+              slowFill.fillAmount.gt(toBN(0)) ||
+              !slowFills.some(
+                (otherSlowFill: Fill) =>
+                  otherSlowFill.originChainId === slowFill.originChainId &&
+                  otherSlowFill.depositId === slowFill.depositId
+              )
```
mrice32, line 295: We should probably re-evaluate this in light of the point you raised above.

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +252,120 @@ export class Dataworker {
...
+          );
+          // 3b. for all non-repeat slow fills, find the FilledRelay event that originally triggered this slow relay.
+          // Recall that slow fills for a deposit are only included in the root bundle if a non-zero amount fill was
+          // submitted for that deposit.
+          firstTimeSlowFills.forEach((slowFill: Fill) => {
+            // Find the earliest fill that should have triggered this slow fill to have been published in a root bundle.
+            const fillThatTriggeredSlowFill = sortEventsAscending(allFills).find(
+              (fill: FillWithBlock) =>
+                fill.originChainId === slowFill.originChainId && fill.depositId === slowFill.depositId
```
mrice32, line 304: Are the fills that are passed in to this function "validated"? The UMIP could be wrong here, but we probably need to make sure these fills are _valid_ fills _or_ that they match the slow fill across all relay parameters (to make sure that invalid fills with the same origin chain and deposit id pollute this logic).

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +252,120 @@ export class Dataworker {
...
+              .sub(slowFill.fillAmount);
+            if (extraFundsSent.eq(toBN(0))) return; // Exit early if slow fill left no excess funds.
+            const l1TokenCounterpart = this.clients.hubPoolClient.getL1TokenCounterpart(
+              slowFill.destinationChainId.toString(),
+              slowFill.destinationToken
+            );
+            if (!runningBalances[slowFill.destinationChainId]) runningBalances[slowFill.destinationChainId] = {};
+            const runningBalance = runningBalances[slowFill.destinationChainId][l1TokenCounterpart];
+            if (runningBalance)
+              runningBalances[slowFill.destinationChainId][l1TokenCounterpart] = runningBalance.sub(extraFundsSent);
```
mrice32, line 337: All this looks great!

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +252,120 @@ export class Dataworker {
...
+      if (!runningBalances[deposit.originChainId.toString()]) runningBalances[deposit.originChainId.toString()] = {};
+      const runningBalanceForDeposit = runningBalances[deposit.originChainId.toString()][l1TokenCounterpart];
+      if (runningBalanceForDeposit)
+        runningBalances[deposit.originChainId.toString()][l1TokenCounterpart] = runningBalanceForDeposit.sub(
+          deposit.amount
+        );
+      else runningBalances[deposit.originChainId.toString()][l1TokenCounterpart] = deposit.amount.mul(toBN(-1));
+    });
+
+    // 6. Factor in latest RootBundleExecuted.runningBalance before this one.
```
mrice32, line 359: +1 was gonna ask about this.

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +252,120 @@ export class Dataworker {
...
+              runningBalances[slowFill.destinationChainId][l1TokenCounterpart] = runningBalance.sub(extraFundsSent);
+            else runningBalances[slowFill.destinationChainId][l1TokenCounterpart] = extraFundsSent.mul(toBN(-1));
+          });
+        });
+      });
+    }
+
+    // 4. Map each deposit event to its L1 token and origin chain ID and subtract deposited amounts from running
+    // balances.
+    deposits.forEach((deposit: Deposit) => {
```
mrice32, line 346: Are these _only_ the deposits that have associated relays or are they all deposits?

mrice32 (approved):
This looks awesome! Some comments and questions, but this is amazing tbh.

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
mrice32: Yeah, I think you're right that the language isn't consistent in the UMIP. Will update that.

src/dataworker/Dataworker.ts
nicholaspai: Yeah this is an edge case we don't capture. So for any Fill where `totalFilledAmount == amount && fillAmount < amount`, i.e. any full fills that occurred after at least one partial fill, we need to check whether the last partial fill for this deposit is also included in this epoch. If not, then it was included in a previous epoch, so we need to grab the `unfilledAmount` remaining after that last partial fill and decrease that from running balance since it is excess leftover from an unexecuted slow fill.

Does that algo work? Do you see a way we can re-use some of the existing code to implement that?

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
nicholaspai: So do you agree that we also should make the search for the `FilledRelay` before the `ProposeRootBundle`'s ending block inclusive?

src/dataworker/Dataworker.ts
nicholaspai: Yeah these `totalRefundAmount` and `realizedLpFees` aggregation variables are only used in the pool rebalance root construction, so I should rename them to something else. The `refundObj.refunds` dictionary contains refunds that we'll include in the `RelayerRefundRoot`. Does that make sense?

src/dataworker/Dataworker.ts
nicholaspai: Yeah I think this block can be removed entirely since its only finding non repeat slow fills, which will be all slow fills

src/dataworker/Dataworker.ts
nicholaspai: Yeah so they are added in `_loadData` after the fill is matched with a deposit:

```js
if (matchedDeposit) {
    // Fill was validated. Save it under all blocks with the block number so we can sort it by time.
    allFills.push(fillWithBlock);
}
```

src/dataworker/Dataworker.ts
nicholaspai: All deposits, since all deposits should decrease running balances right?

src/dataworker/Dataworker.ts
mrice32: Gotcha. Perfect.

src/dataworker/Dataworker.ts
mrice32: Yeah, I think that makes sense!

mrice32: These are only deposits in the block range, right?

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
mrice32: Yes. I think all searches for a particular bundle should be exclusive of the previous end block and inclusive of this bundle's end block. Basically like this: `(previousEndBlock, thisEndBlock]`.

Does that match your intuition?

src/dataworker/Dataworker.ts
mrice32: As I understand it, what you're suggesting is:

1. Slow relays that complete a fill. In those cases, we follow the already-described algorithm of going back in time to find the bundle where the slow relay was initiated, finding the last partial fill that was considered to figure out the total amount that was sent and subtract the amount used in the slow relay to figure out the unused portion. Subtract that unused portion from the running balance.
2. For any fills (non slow relay) in the current epoch that “complete” a partial fill, we check if there was a partial fill in a previous epoch (if so that would mean a slow relay). We then determine the amount that was sent for the slow relay and deduct that entire amount from the running balance.

src/clients/HubPoolClient.ts
nicholaspai: Done

src/dataworker/Dataworker.ts
nicholaspai: Exactly, I've implemented those two cases now

src/dataworker/Dataworker.ts
nicholaspai: I use the `totalRefundAmount` to seed the `runninBalance`, so it makes sense I think to add the slow fill amount to the `totalRefundAmount` so that the correct amount of tokens from the LP pool needed to pay the user are sent to the spoke pool:

> If a slow relay occured, that means the pool sent the capital for that fill

Alternatively, I could have added each slow fill amount to the `runningBalance` but for efficiency I added it in the `_loadData` function above

src/dataworker/Dataworker.ts
nicholaspai: yeah, which I'll filter one I implement block range constraints

src/dataworker/Dataworker.ts
mrice32: Hmmm, what I'm confused about is that when slow fills are _executed_, that means the money was sent in the previous bundle, right? So we don't need to send any additional money, no?

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
```diff
@@ -213,23 +261,147 @@ export class Dataworker {
...
+      if (fill.isSlowRelay) {
+        // Since every slow fill should have a partial fill that came before it, we should always be able to find the
+        // last partial fill for the root bundle including the slow fill refund.
+        if (!lastFillBeforeSlowFillIncludedInRoot)
+          throw new Error("Can't find last fill submitted before slow fill was included in root bundle proposal");
+
+        // Recompute how much the matched root bundle sent for this slow fill. Subtract the amount that was
+        // actually executed on the L2 from the amount that was sent. This should give us the excess that was sent.
+        // Subtract that amount from the running balance so we ultimately send it back to L1.
+        const amountToSendBackToHub = fill.fillAmount
```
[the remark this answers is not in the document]
nicholaspai: I'll change the math around to be clearer, I changed it to remove a .mul(toBN(-1)) but I realize that its not clear.

src/dataworker/Dataworker.ts
nicholaspai: This is a good point, I'll remove slow fill from `totalRefundAmount`

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: This should be clearer:

```js
        const amountSentForSlowFill = lastFillBeforeSlowFillIncludedInRoot.amount.sub(
          lastFillBeforeSlowFillIncludedInRoot.totalFilledAmount
        );
        const excess = amountSentForSlowFill.sub(fill.fillAmount)
        if (excess.eq(toBN(0))) return; // Exit early if slow fill left no excess funds.
        this._updateRunningBalanceForFill(runningBalances, fill, excess.mul(toBN(-1)));
```

mrice32 (approved):
LGTM!!

src/clients/RateModelClient.ts
```diff
@@ -55,14 +58,16 @@ export class RateModelClient {
...
-    return realizedLpFeePct;
+    return { realizedLpFeePct, quoteBlock };
   }
 
   getRateModelForBlockNumber(l1Token: string, blockNumber: number | undefined = undefined): across.constants.RateModel {
     return this.rateModelDictionary.getRateModelForBlockNumber(l1Token, blockNumber);
   }
 
   async update() {
+    if (this.hubPoolClient !== null && !this.hubPoolClient.isUpdated) throw new Error("HubPool not updated");
```
chrismaree, line 69: ^

PR #28, opened 2022-04-19
Retry, quorum and event paginated provider

chrismaree:
- bots now only search from deployed block number (exported in contracts-v2 repo)
- bots use a quorum config between multiple providers to ensure that multiple providers must agree before returning any data
- bots have pagination on event lookups.


src/clients/HubPoolClient.ts
```diff
@@ -12,9 +13,10 @@ export class HubPoolClient {
   constructor(
     readonly logger: winston.Logger,
     readonly hubPool: Contract,
-    readonly startingBlock: number = 0,
-    readonly endingBlock: number | null = null
-  ) {}
+    readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 }
```
chrismaree, line 16: this is a bit cleaner: we always refer to a block search range as `EventSearchConfig` which has a `fromBlock`, `toBlock` and `maxBlockLookBack` in it.

src/clients/HubPoolClient.ts
```diff
@@ -72,13 +74,17 @@ export class HubPoolClient {
...
+      maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
+    };
     this.logger.debug({ at: "HubPoolClient", message: "Updating client", searchConfig });
-    if (searchConfig[0] > searchConfig[1]) return; // If the starting block is greater than the ending block return.
+    if (searchConfig.fromBlock > searchConfig.toBlock) return; // If the starting block is greater than the ending block return.
 
     const [poolRebalanceRouteEvents, l1TokensLPEvents] = await Promise.all([
-      this.hubPool.queryFilter(this.hubPool.filters.SetPoolRebalanceRoute(), ...searchConfig),
-      this.hubPool.queryFilter(this.hubPool.filters.L1TokenEnabledForLiquidityProvision(), ...searchConfig),
+      paginatedEventQuery(this.hubPool, this.hubPool.filters.SetPoolRebalanceRoute(), searchConfig),
```
chrismaree, line 86: use of the new utility.

src/clients/RelayerClientHelper.ts
```diff
@@ -14,7 +14,7 @@ export function constructRelayerClients(logger: winston.Logger, config: RelayerC
 
   const hubSigner = baseSigner.connect(getProvider(config.hubPoolChainId));
   const spokeSigners = config.spokePoolChains
-    .map((networkId) => getProvider(networkId))
+    .map((networkId) => getProvider(networkId, config.nodeQuorumThreshold))
```
chrismaree, line 17: inclusion of the node quorum threshold which sets how many nodes must agree before returning results.

src/utils/ContractUtils.ts
```diff
@@ -32,3 +32,12 @@ export function getParamType(contractName: string, functionName: string, paramNa
   const fragment = artifact.abi.find((fragment) => fragment.name === functionName);
   return fragment!.inputs.find((input) => input.name === paramName) || "";
 }
+
+export function getDeploymentBlockNumber(contractName: string, networkId: number) {
```
chrismaree, line 36: wrapper for the contracts method. difference is we need to do the `castSpokePoolName` logic, like done when fetching the contract deployment addresses in this util

src/utils/EventUtils.ts
```diff
@@ -13,3 +13,27 @@ export function spreadEvent(event: Event) {
 
   return returnedObject;
 }
+
+export async function paginatedEventQuery(contract: Contract, filter: EventFilter, searchConfig: EventSearchConfig) {
```
chrismaree, line 17: pagination logic done simply :)

src/utils/ProviderUtils.ts
```diff
@@ -1,7 +1,22 @@
...
-export function getProvider(networkId: number) {
+export function getProvider(networkId: number, nodeQuorumThreshold: number = 1) {
+  if (process.env[`RETRY_CONFIG_${networkId}`]) return getFallbackProvider(networkId, nodeQuorumThreshold);
   const nodeUrl = process.env[`NODE_URL_${networkId}`];
-  if (!nodeUrl) throw new Error(`No node url for network ${networkId}`);
+  if (!nodeUrl) throw new Error(`No NODE_URL_ for network ${networkId}`);
   return new ethers.providers.JsonRpcProvider(nodeUrl);
 }
+
+export function getFallbackProvider(networkId: number, nodeQuorumThreshold: number) {
```
chrismaree, line 10: fallback provider! this acts to enforce a `nodeQuorumThreshold` against all providers defined. see: https://docs.ethers.io/v5/single-page/#/v5/api/providers/other/-%23-FallbackProvider

package.json
```diff
@@ -7,7 +7,7 @@
   "license": "AGPL-3.0",
   "private": true,
   "dependencies": {
-    "@across-protocol/contracts-v2": "^0.0.43",
+    "@across-protocol/contracts-v2": "^0.0.46",
```
chrismaree, line 10: see associated changes here: https://github.com/across-protocol/contracts-v2/pull/143/files

src/clients/HubPoolClient.ts
```diff
@@ -72,13 +74,17 @@ export class HubPoolClient {
...
   async update() {
-    const searchConfig = [this.firstBlockToSearch, this.endingBlock || (await this.hubPool.provider.getBlockNumber())];
+    const searchConfig = {
+      fromBlock: this.firstBlockToSearch,
+      toBlock: this.eventSearchConfig.toBlock || (await this.hubPool.provider.getBlockNumber()),
+      maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
+    };
     this.logger.debug({ at: "HubPoolClient", message: "Updating client", searchConfig });
-    if (searchConfig[0] > searchConfig[1]) return; // If the starting block is greater than the ending block return.
+    if (searchConfig.fromBlock > searchConfig.toBlock) return; // If the starting block is greater than the ending block return.
```
nicholaspai, line 83: what if theyre >=?

src/clients/RelayerClientHelper.ts
nicholaspai: good name

src/utils/EventUtils.ts
```diff
@@ -13,3 +13,27 @@ export function spreadEvent(event: Event) {
...
+  const numberOfQueries = Math.ceil((searchConfig.toBlock - searchConfig.fromBlock) / searchConfig.maxBlockLookBack);
+
+  let promises = [];
+  for (let i = 0; i < numberOfQueries; i++) {
+    const fromBlock = searchConfig.fromBlock + i * searchConfig.maxBlockLookBack;
+    const toBlock = Math.min(searchConfig.fromBlock + (i + 1) * searchConfig.maxBlockLookBack, searchConfig.toBlock);
+    promises.push(contract.queryFilter(filter, fromBlock, toBlock));
+  }
+
+  return (await Promise.all(promises)).flat();
```
nicholaspai, line 32: do we need to sort these events at all? even to be safe?

src/utils/ProviderUtils.ts
```diff
@@ -1,7 +1,26 @@
...
+
+// Create a fallback provider. This provider type enforces that nodeQuorumThreshold of the total retry configs defined
+// all agree in the data they return before returning the data to the provider user. This applies in the context of both
+// events and state reads to ensure consistency and redundancy within the infrastructure. Set the priority to 1.
+// This makes all providers equally weighted and should all be considered. Stall timeout is how long to wait for a node
+// to reply before trying another node in the set provided. This would be the case when the number of nodes is more than
+// the nodeQuorumThreshold (i.e using a 2 of 3 setup).
+export function getFallbackProvider(networkId: number, nodeQuorumThreshold: number) {
+  const nodeUrls = JSON.parse(process.env[`RETRY_CONFIG_${networkId}`]) || [];
+  if (nodeUrls.length == 0) throw new Error(`No RETRY_CONFIG_ for network ${networkId}`);
```
nicholaspai, line 18:
```suggestion
  if (nodeUrls.length === 0) throw new Error(`No RETRY_CONFIG_ for network ${networkId}`);
```

src/utils/ProviderUtils.ts
```diff
@@ -1,7 +1,26 @@
...
+
+  return new providers.FallbackProvider(
+    nodeUrls.map((url) => {
+      return { provider: new ethers.providers.JsonRpcProvider(url), priority: 1, stallTimeout: 2000, weight: 1 };
```
nicholaspai, line 22: Should they all be priority 1, weight 1? I'm not sure what those do by the way, comments could be useful

nicholaspai (approved):
This is a super useful utility, lgtm

src/utils/EventUtils.ts
chrismaree: two questions: 1) how would they be out of order? 2) would them being out of order matter?

src/utils/ProviderUtils.ts
chrismaree: sir please see the comment on line 10 in the file. let me know if it's not sufficient.

PR #29, opened 2022-04-19
fix: add ability to run relayer for testnet

chrismaree:
This PR adds a small temp workaround to enable bots to run without finding block numbers. See comment in code as to why this was done.


nicholaspai (approved):
lgtm

PR #31, opened 2022-04-27
feat: Start running balance at latest running balance value and split PoolRebalanceLeaf if too many L1 tokens for an L2 chain

nicholaspai:
Two key features introduced in this PR:
- `runningBalance` for a specific { chainID, L1Token } is initialized to same running balance in most recent `RootBundleExecuted` event.
- `PoolRebalanceLeaf` is split into multiple sub-leaves with unique `groupIndex` properties if there are too many `l1Tokens` for a `chainId`
- `runningBalance` gets reset to 0, and copied to `netSendAmounts` if over a threshold.
- Builds root from leaves

Remaining features to implement `buildPoolRebalanceRoot`
- Implement `ConfigStoreClient` logic to read token threshold from some config store

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -40,10 +40,10 @@ export class HubPoolClient {
   getL1TokenCounterpart(repaymentChainId: string, l2Token: string) {
     let l1Token = null;
     Object.keys(this.l1TokensToDestinationTokens).forEach((_l1Token) => {
-      if (this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token)
-        l1Token = _l1Token;
+      if (this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token) l1Token = _l1Token;
     });
-    if (l1Token === null) throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
+    if (l1Token === null)
+      throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
     return l1Token;
   }
 
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -7,6 +7,7 @@ export class HubPoolClient {
   private l1TokensToDestinationTokens: { [l1Token: string]: { [destinationChainId: number]: string } } = {};
   private l1Tokens: L1Token[] = []; // L1Tokens and their associated info.
   private proposedRootBundles: ProposedRootBundle[] = [];
+  private l1TokensToDestinationTokensWithBlock: { [l1Token: string]: { [destinationChainId: number]: [{ l2Token: string, block: number }] } } = {};
 
   public isUpdated: boolean = false;
   public firstBlockToSearch: number;
@@ -39,12 +40,18 @@ export class HubPoolClient {
     return l1Token;
   }
 
-  getL1TokenCounterpart(repaymentChainId: string, l2Token: string) {
-    const l1Token = Object.keys(this.l1TokensToDestinationTokens).find(
-      (_l1Token) => this.l1TokensToDestinationTokens[_l1Token][repaymentChainId] === l2Token
+  getL1TokenCounterpartAtBlock(l2ChainId: string, l2Token: string, block: number) {
+    const l1Token = Object.keys(this.l1TokensToDestinationTokensWithBlock).find(
+      (_l1Token) => {
+          // We assume that l2-l1 token mapping events are sorted in descending order, so find the last mapping published
+          // before the target block.
+        return this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId].find(
+          (mapping: { l2Token: string; block: number }) => mapping.l2Token === l2Token && mapping.block <= block
+        )
+      }
     );
-    if (l1Token === null)
-      throw new Error(`Could not find L1 Token for repayment chain ${repaymentChainId} and L2 token ${l2Token}!`);
+    if (!l1Token)
+      throw new Error(`Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${block}!`);
     return l1Token;
   }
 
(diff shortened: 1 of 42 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -297,12 +297,20 @@ export class Dataworker {
             fillsToRefund[repaymentChainId][l2TokenAddress].realizedLpFees
           );
 
+          // Start with latest RootBundleExecuted.runningBalance for {chainId, l1Token} combination if found.
+          const startingRunningBalance =
+            this.clients.hubPoolClient.getRunningBalanceBeforeBlockForChain(
+              1_000_000,
```
nicholaspai, line 303: This block is a placeholder for some very large block so we can grab latest block, which we might replace with some value from the bundle block eval range once implemented

test/Dataworker.buildRoots.ts
```diff
@@ -549,8 +544,29 @@ describe("Dataworker: Build merkle roots", async function () {
...
-      fill7.destinationChainId,
-      l1Token_1.address,
+      [fill7.destinationChainId],
+      [l1Token_1.address],
+      dataworkerInstance.buildPoolRebalanceRoot([])
+    );
+
+    // Execute root bundle so we can propose another bundle. The latest running balance emitted in this event
+    // should be added to running balances.
+    await timer.setCurrentTime(Number(await timer.getCurrentTime()) + refundProposalLiveness + 1);
```
nicholaspai, line 554: This is the main test change: test that running balances is updated after an ExecutedRootBundle event is emitted

src/clients/HubPoolClient.ts
```diff
@@ -111,16 +112,44 @@ export class HubPoolClient {
     return endingBlockNumber;
   }
 
+  getRunningBalanceBeforeBlockForChain(block: number, chain: number, l1Token: string): BigNumber | undefined {
```
mrice32, line 115: Does undefined signal that there was no previous running balance? Any reason not to return 0 instead?

src/dataworker/Dataworker.ts
mrice32: Sorry, I'm confused. If we want latest block, why not just grab the latest block from the node? Or am I misunderstanding the point of this line?

src/dataworker/Dataworker.ts
```diff
@@ -399,15 +406,83 @@ export class Dataworker {
...
+          );
+
+          // If the running balance is greater than the token transfer threshold, then set the net send amount
+          // equal to the running balance and reset the running balance to 0. Otherwise, the net send amount should be
+          // 0, indicating that we do not want the data worker to trigger a token transfer between hub pool and spoke
+          // pool when executing this leaf.
+          const getNetSendAmountOrRunningBalanceForL1Token = (
+            runningBalance: BigNumber,
+            l1Token: string,
+            getRunningBalance: boolean
```
mrice32, line 442: Sorry, I'm a little confused as to what `getRunningBalance` is intended to do?

src/dataworker/Dataworker.ts
```diff
@@ -399,15 +406,83 @@ export class Dataworker {
...
+                  getNetSendAmountOrRunningBalanceForL1Token(runningBalances[chainId][l1Token], l1Token, false)
+                )
+              : Array(l1TokensToIncludeInThisLeaf.length).fill(toBN(0)),
+            l1Tokens: l1TokensToIncludeInThisLeaf,
+          });
+        }
+      });
 
-    // TODO: Add helpful logs everywhere.
+    const tree = leaves.length > 0 ? buildPoolRebalanceLeafTree(leaves) : null;
```
mrice32, line 479: What does a null tree end up doing? Does this produce a valid merkle root?

mrice32 (approved):
Looks great! A few minor questions and comments, but this is really awesome work.

nicholaspai:
> Sorry, I'm confused. If we want latest block, why not just grab the latest block from the node? Or am I misunderstanding the point of this line?

Well, in this line we actually want to find the latest `RootBundleExecuted` event before the block range to be evaluated. This is important in the case where we're verifying a pool rebalance root was constructed successfully, from the disputer's point of view, so we can't just always grab current block. Do you agree?

@mrice32

src/dataworker/Dataworker.ts
nicholaspai: It errors out if leaves are empty, which might be ok. Would you want `buildRoot` methods to error out or throw null when there are no leaves with which to build a tree?

src/dataworker/Dataworker.ts
nicholaspai: Well, in this line we actually want to find the latest `RootBundleExecuted` event before the block range to be evaluated. This is important in the case where we're verifying a pool rebalance root was constructed successfully, from the disputer's point of view, so we can't just always grab current block. Do you agree?

@mrice32

src/dataworker/Dataworker.ts
mrice32: For which merkle tree. For instance, I think it's totally reasonable (and likely) for there to be no slow relays. Less likely but still possible for there to be no relayer repayments. If both are empty, would there be any point in publishing a bundle? Maybe not, but I think just one of them being empty is at least conceivable. Thoughts?

src/dataworker/Dataworker.ts
mrice32: I do, but then shouldn't we be getting that `1_000_000` block number from somewhere? I guess I just don't understand what work the `1_000_000` is doing?

src/dataworker/Dataworker.ts
```diff
@@ -399,15 +405,61 @@ export class Dataworker {
...
+              ? l1TokensToIncludeInThisLeaf.map((l1Token) => realizedLpFees[chainId][l1Token])
+              : Array(l1TokensToIncludeInThisLeaf.length).fill(toBN(0)),
+            runningBalances: runningBalances[chainId]
+              ? l1TokensToIncludeInThisLeaf.map((l1Token) =>
+                  this._getRunningBalanceForL1Token(runningBalances[chainId][l1Token], l1Token)
+                )
+              : Array(l1TokensToIncludeInThisLeaf.length).fill(toBN(0)),
+            netSendAmounts: runningBalances[chainId]
+              ? l1TokensToIncludeInThisLeaf.map((l1Token) =>
+                  this._getNetSendAmounForL1Token(runningBalances[chainId][l1Token], l1Token)
```
mrice32, line 448:
```suggestion
                  this._getNetSendAmountForL1Token(runningBalances[chainId][l1Token], l1Token)
```

src/dataworker/Dataworker.ts
```diff
@@ -443,6 +495,21 @@ export class Dataworker {
     // TODO:
   }
 
+  // If the running balance is greater than the token transfer threshold, then set the net send amount
+  // equal to the running balance and reset the running balance to 0. Otherwise, the net send amount should be
+  // 0, indicating that we do not want the data worker to trigger a token transfer between hub pool and spoke
+  // pool when executing this leaf.
+  _getNetSendAmounForL1Token(runningBalance: BigNumber, l1Token: string): BigNumber {
```
mrice32, line 502:
```suggestion
  _getNetSendAmountForL1Token(runningBalance: BigNumber, l1Token: string): BigNumber {
```

src/dataworker/Dataworker.ts
```diff
@@ -399,15 +405,61 @@ export class Dataworker {
...
+          leaves.push({
+            groupIndex: groupIndexForChainId++,
+            leafId: leaves.length,
+            chainId: Number(chainId),
+            bundleLpFees: realizedLpFees[chainId]
+              ? l1TokensToIncludeInThisLeaf.map((l1Token) => realizedLpFees[chainId][l1Token])
+              : Array(l1TokensToIncludeInThisLeaf.length).fill(toBN(0)),
+            runningBalances: runningBalances[chainId]
+              ? l1TokensToIncludeInThisLeaf.map((l1Token) =>
+                  this._getRunningBalanceForL1Token(runningBalances[chainId][l1Token], l1Token)
```
mrice32, line 443: This makes a lot of sense. +1 Sorry if I was being dense in the prev implementation.

mrice32 (commented):
Responded to a few threads, but looks awesome!

src/dataworker/Dataworker.ts
nicholaspai: no I think i was forcing something before

src/dataworker/Dataworker.ts
mrice32: Sorry, this was my mistake. I didn't realize this was just the pool rebalance leaf tree. I don't think this can ever be empty for a valid proposal. So it seems like we should error if this is what we come up with. Thoughts?

src/dataworker/Dataworker.ts
nicholaspai: yeah from the bundle block range I imagine, the 1_000_000 is a placeholder and I think i have comment about that

PR #33, opened 2022-05-02
feat: Implement bundle block range filtering

nicholaspai:
- Implements `getNextBundleStartBlockNumber` in HubPoolClient which can be used by data worker to construct bundle block range for new root bundle proposal
- Adds unit tests for above function
- Implements first part of `Dataworker.proposeRootBundle` which constructs block range for next root using `getNextBundleStartBlockNumber` and then inputs that block range into root builder functions to construct root for range

Still todo:
- add tests for `Dataworker` methods that check that block range filtering is used correctly


nicholaspai:
Will add some tests after first pass on reviews

src/clients/HubPoolClient.ts
```diff
@@ -113,6 +114,33 @@ export class HubPoolClient {
     return endingBlockNumber;
   }
 
+  getNextBundleStartBlockNumber(chainIdList: number[], latestBlock: number, chainId: number): number {
```
chrismaree, line 117: so your PR works by using the `spreadEventWithBlockNumber` which appends the block number to the event then uses `getNextBundleStartBlockNumber` to do the searching over the provided range. this makes sence. we can think about refining this later to contain ranges and searching within the underlying class structure. what you have for now does make sence.

src/clients/HubPoolClient.ts
```diff
@@ -113,6 +114,33 @@ export class HubPoolClient {
...
+      (event: ExecutedRootBundle) => event.blockNumber <= latestBlock && event.chainId === chainId
+    );
+
+    // TODO: If no event for chain ID, then we can return a conservative default starting block like 0,
+    // or we could throw an Error.
+    if (!latestRootBundleExecutedEvent) return 0;
+
+    // Once that event is found, search for the ProposeRootBundle event that is as late as possible, but earlier than
+    // the RootBundleExecuted event we just identified.
+    const mostRecentProposeRootBundleEventWithChain = sortEventsDescending(this.proposedRootBundles).find(
```
chrismaree, line 130: in what situation would the events not be sorted?

chrismaree (approved):
this LGTM. I think we still need to think about the overall data structures we are using in the HubPoolClient but for now this is ok as a path of least resistance.

PR #34, opened 2022-05-03
feat: HubPoolClient should be able to return cross chain contracts for a specific block height

nicholaspai:
This PR adds a function to the HubPoolClient: `getSpokePoolForBlockNumber`. This can be used to make sure that dataworker is fetching data from correct spoke pools at propose and validation time


src/dataworker/Dataworker.ts
```diff
@@ -527,9 +534,36 @@ export class Dataworker {
       this.clients.spokePoolClients[chainId].latestBlockNumber,
     ]);
 
+    // 2. Construct spoke pool clients using spoke pools deployed at end of block range.
+    // We do make an assumption that the spoke pool contract was not changed during the block range. By using the
+    // spoke pool at this block instead of assuming its the currently deployed one, we can pay refunds for deposits
+    // on deprecated spoke pools.
+    const endBlockForMainnet = this._getBlockRangeForChain(blockRangesForProposal, 1)[1];
```
nicholaspai, line 541: For all dataworker functions we have a choice of which mainnet block (start or end of bundle range) to use when looking up events:
- SpokePool address: `CrossChainContractsSet`
- L1 tokencounterparts set for an L2 chain + L2 token:`PoolRebalanceRoute`
- Config store settings like `UpdatedGlobalConfig` and `UpdatedTokenConfig`

Do you think we should use the end or start block? Or am i thinking about this wrong?

mrice32: I think we should do the end block, generally, since that would be the most up to date. I can't think of a reason we would want to intentionally use an out-of-date one.

The other option is to use the _current_ value when making a proposal, and then evaluate it based on the block number _of the proposal_. I think your way of doing it, however, is safer for the proposer, despite the fact that it might allow a proposer to manipulate which value is "correct" for their proposal.

src/dataworker/Dataworker.ts
```diff
@@ -527,9 +534,36 @@ export class Dataworker {
...
+      );
+      result[chainId] = new SpokePoolClient(
+        this.logger,
+        spokePoolContract,
+        this.clients.configStoreClient,
+        Number(chainId),
+        this.clients.spokePoolClients[chainId].eventSearchConfig
+      );
+      return result;
+    }, {});
```
mrice32, line 561:
```suggestion
    const spokePoolClients = Object.fromEntries(this.chainIdListForBundleEvaluationBlockNumbers.map((chainId) => {
      const spokePoolContract = new Contract(
        this.clients.hubPoolClient.getSpokePoolForBlock(endBlockForMainnet, Number(chainId)),
        SpokePool.abi,
        this.clients.spokePoolClients[chainId].spokePool.signer
      );
      const client = new SpokePoolClient(
        this.logger,
        spokePoolContract,
        this.clients.configStoreClient,
        Number(chainId),
        this.clients.spokePoolClients[chainId].eventSearchConfig
      );
      return [chainId, client];
    }));
```

mrice32 (approved):
LGTM! Only minor comments

src/dataworker/Dataworker.ts
nicholaspai: > despite the fact that it might allow a proposer to manipulate which value is "correct" for their proposal.

This is a great point. We can mitigate this by validating the block range itself, right? i.e. the block range needs to follow from the previous and be current, etc.?

src/dataworker/Dataworker.ts
nicholaspai: woah this is a cool function, never heard of Object.fromEntries

src/common/ClientHelper.ts
```diff
@@ -48,57 +47,12 @@ export async function constructClients(logger: winston.Logger, config: CommonCon
     rateModelClientSearchSettings
   );
 
-  const spokePoolClients = {};
-  spokePools.forEach((obj: { networkId: number; contract: Contract }) => {
```
nicholaspai, line 52: So I realized that the Dataworker doesn't need to construct spoke pool clients in this step, since it will construct the spoke pools using the addresses set for the block range, which in some cases might be a downgraded spoke pool. So removing this setup logic saves a few web3 event requests in the upgrade edge case

mrice32: Interesting. Nice catch

src/dataworker/Dataworker.ts
mrice32: Yeah, exactly. There are limits to what blocks they can pick from, so effectively, the upgrade must have happened between the last proposal and now for them to "manipulate".

PR #35, opened 2022-05-04
feat: Change RateModelClient => ConfigStoreClient to handle new AcrossConfigStore

nicholaspai:
RateModelClient and ConfigStoreClient are folded into the same `AcrossConfigStoreClient` now. Previously hardcoded config vars like `MAX_RELAYER_REPAYMENT_LEAF_SIZE` and `MAX_POOL_REBALANCE_LEAF_SIZE` are now queried from on-chain `AcrossConfigStore` contract.


src/clients/RateModelClient.ts
```diff
@@ -65,33 +78,100 @@ export class RateModelClient {
...
+        }
+
+        // Store RateModel:
+        // TODO: Temporarily reformat the shape of the event that we pass into the sdk.rateModel class to make it fit
+        // the expected shape. This is a fix for now that we should eventually replace when we change the sdk.rateModel
+        // class itself to work with the generalized ConfigStore.
+        const l1Token = args.key;
+        delete args.value;
+        delete args.key;
+        this.cumulativeRateModelUpdates.push({ ...args, rateModel: rateModelForToken, l1Token });
```
chrismaree, line 145: this is a good temp solution to get us up and running without waisting too much time. good work!

src/interfaces/ConfigStore.ts
```diff
@@ -0,0 +1,15 @@
+import { BigNumber } from "../utils";
+import { SortableEvent } from "./Common";
+export interface L1TokenTransferThreshold extends SortableEvent {
```
chrismaree, line 3: excellent.

src/interfaces/ConfigStore.ts
```diff
@@ -0,0 +1,15 @@
+import { BigNumber } from "../utils";
+import { SortableEvent } from "./Common";
```
chrismaree, line 2: it might make sence to join `Common` with `index.ts` for these interfaces but we can change this later.

chrismaree (approved):
this is great! your implementation means that we dont need to make any SDK or protocol changes (right now) and can refactor. this once everything else is done. good work!

src/clients/RateModelClient.ts
nicholaspai: yeah eventually will change the SDK

src/clients/RelayerClientHelper.ts
```diff
@@ -1,6 +1,13 @@
...
 import { getProvider, getSigner, getDeployedContract, getDeploymentBlockNumber, Contract } from "../utils";
-import { SpokePoolClient, HubPoolClient, RateModelClient, TokenClient, ProfitClient, MultiCallerClient } from ".";
+import {
+  SpokePoolClient,
+  HubPoolClient,
+  AcrossConfigStoreClient,
+  TokenClient,
+  ProfitClient,
+  MultiCallerClient,
+} from ".";
```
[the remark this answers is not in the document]
nicholaspai: will fix!

src/interfaces/ConfigStore.ts
nicholaspai: How does that work? You mean just add them to `index.ts`?

PR #36, opened 2022-05-06
feat: mainnet refinement 1

chrismaree:
refinements after running the relayer bot in production over the last few days.


src/clients/MultiCallerClient.ts
```diff
@@ -80,29 +80,41 @@ export class MultiCallerClient {
       const multiCallTransactionsResult = await Promise.allSettled(
         Object.keys(groupedTransactions).map((chainId) => this.buildMultiCallBundle(groupedTransactions[chainId]))
       );
-
+      this.logger.debug({ at: "MultiCaller", message: "Waiting for bundle transaction inclusion" });
       const transactionReceipts = await Promise.allSettled(
-        multiCallTransactionsResult.map((transaction) => (transaction ? (transaction as any).value.wait() : null))
+        multiCallTransactionsResult.map((transaction) =>
```
chrismaree, line 85: this logic came out of testing the bot during a polygon outage. in this case, the transactions will be submitted but never included. The logic below updates the bot to gracefully handles this. now, the bot will produce a log that looks as follows:
image


If there are other networks not timing out the bot will continue to run on them and can still send transactions on it, like this:

src/clients/MultiCallerClient.ts
```diff
@@ -119,8 +131,8 @@ export class MultiCallerClient {
       });
       return null; // If there is a problem in the targets in the bundle return null. This will be a noop.
     }
-    const multiCallData = transactions.map((tx) => tx.contract.interface.encodeFunctionData(tx.method, tx.args));
-    this.logger.debug({ at: "MultiCallerClient", message: "Produced bundle", target: target.address, multiCallData });
-    return runTransaction(this.logger, target, "multicall", [multiCallData]);
+    const callData = transactions.map((tx) => tx.contract.interface.encodeFunctionData(tx.method, tx.args));
+    this.logger.debug({ at: "MultiCallerClient", message: "Made bundle", target: getTarget(target.address), callData });
```
chrismaree, line 135: added the `getTarget` method that returns information about a target for a known address, such as the chain of the target and the contract name. used here and within the `TransactionUtils.ts` file.

src/clients/RateModelClient.ts
```diff
@@ -58,19 +58,8 @@ export class AcrossConfigStoreClient {
     }
 
     const { current, post } = await this.hubPoolClient.getPostRelayPoolUtilization(l1Token, quoteBlock, deposit.amount);
-
     const realizedLpFeePct = lpFeeCalculator.calculateRealizedLpFeePct(rateModel, current, post);
 
-    this.logger.debug({
```
chrismaree, line 64: this log was very noisy. it was produced once per deposit which on startup made a lot of noise

src/clients/TokenClient.ts
```diff
@@ -106,12 +106,12 @@ export class TokenClient {
     for (const { token, chainId } of tokensToApprove) {
       const targetSpokePool = this.spokePoolClients[chainId].spokePool;
       const contract = new Contract(token, ERC20.abi, targetSpokePool.signer);
-      const tx = await runTransaction(this.logger, contract, "approve", [targetSpokePool.address, MAX_SAFE_ALLOWANCE]);
```
chrismaree, line 109: this here was a bug that resulted in the bot re-approving after using any of its allowance.

src/relayer/Relayer.ts
```diff
@@ -147,10 +147,10 @@ export class Relayer {
       `to ${getNetworkName(deposit.destinationChainId)} of ` +
       `${createFormatFunction(2, 4, false, decimals)(deposit.amount.toString())} ${symbol}. ` +
       `with depositor ${etherscanLink(deposit.depositor, deposit.originChainId)}. ` +
-      `Fill amount of ${createFormatFunction(2, 4, false, 18)(fillAmount.toString())} ${symbol} with ` +
```
chrismaree, line 150: logging refinements.

src/common/Config.ts
```diff
@@ -11,17 +11,26 @@ export class CommonConfig {
...
+      MAX_BLOCK_LOOK_BACK,
+      NODE_QUORUM_THRESHOLD,
+      MAX_TX_WAIT_DURATION,
+    } = env;
     this.hubPoolChainId = HUB_CHAIN_ID ? Number(HUB_CHAIN_ID) : 1;
     this.spokePoolChains = CONFIGURED_NETWORKS ? JSON.parse(CONFIGURED_NETWORKS) : Constants.CHAIN_ID_LIST_INDICES;
     this.pollingDelay = POLLING_DELAY ? Number(POLLING_DELAY) : 60;
     this.maxBlockLookBack = MAX_BLOCK_LOOK_BACK ? JSON.parse(MAX_BLOCK_LOOK_BACK) : {};
     if (Object.keys(this.maxBlockLookBack).length > 0)
-      for (const key of Object.keys(this.maxBlockLookBack))
```
chrismaree, line 22: this change fixes a small bug that prevented the bot from having more keys in `MAX_BLOCK_LOOK_BACK` than `CONFIGURED_NETWORKS`, which does not make sence.

src/relayer/RelayerClientHelper.ts
```diff
@@ -17,8 +17,9 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
 
   // Create clients for each contract for each chain.
   const spokePoolSigners = getSpokePoolSigners(baseSigner, config);
-  const spokePools = config.spokePoolChains.map((networkId, index) => {
```
chrismaree, line 20: this contained a bug. the index from `config.spokePoolChains` is incorrect, when keyed in `spokePoolSigners. `spokePoolSigners` is `chainId` to wallet, not index to wallet. this bug was introduced in https://github.com/across-protocol/relayer-v2/pull/34 and fixed here.

evaldofelipe (approved):

PR #37, opened 2022-05-06
feat: Implement amountToReturn in RelayerRefundRoot and initial creation of proposeRootBundle txn

nicholaspai:
Does not actually submit `proposeRootBundle` txn but simulates it if successfully formed. A lot of work is done to make sure that the root bundle is detailed clearly when logged via SlackTransport.

Also implements the last remaining part of the roots, which is the `amountToReturn` property in the `RelayerRefundRoot`. This is dependent on the `netSendAmount` in the `PoolRebalanceRoot`. Implementation details described [here](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md#constructing-relayerrefundroot)


src/clients/HubPoolClient.ts
```diff
@@ -178,6 +180,23 @@ export class HubPoolClient {
     } else return toBN(0);
   }
 
+  async setBondTokenAllowance() {
```
chrismaree, line 183: I dont think this should be here. this should be done via the [TokenClient](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/TokenClient.ts) which has a similar method for adding approvals for the relay whitelisted tokens, see [here](https://github.com/across-protocol/relayer-v2/blob/b3e38babaeed3972d57a3930e99120b91d5c6546/src/clients/TokenClient.ts#L92).

I think you can simply move the method you have here to that file.

additionally, we can add logic there around monitoring the data workers token balance in the `bondToken` to warn us if we run out of tokens to submit new bundles.

src/clients/MultiCallerClient.ts
```diff
@@ -27,10 +27,15 @@ export class MultiCallerClient {
     this.transactions = [];
   }
 
-  async executeTransactionQueue() {
+  async executeTransactionQueue(simulate = false) {
```
chrismaree, line 30:
```suggestion
  async executeTransactionQueue(simulationModeOn = false) {
```

src/clients/MultiCallerClient.ts
```diff
@@ -27,10 +27,15 @@ export class MultiCallerClient {
...
     try {
       if (this.transactions.length === 0) return;
-      this.logger.debug({ at: "MultiCallerClient", message: "Executing tx bundle", number: this.transactions.length });
+      this.logger.debug({
+        at: "MultiCallerClient",
+        message: "Executing tx bundle",
+        number: this.transactions.length,
+        simulationModeOn: simulate,
```
chrismaree, line 37:
```suggestion
        simulationModeOn,
```

src/clients/MultiCallerClient.ts
```diff
@@ -70,6 +75,25 @@ export class MultiCallerClient {
         assign(groupedTransactions, [transaction.chainId], [transaction]);
       }
 
+      if (simulate) {
```
chrismaree, line 78:
```suggestion
      if (simulationModeOn) {
```

src/dataworker/Dataworker.ts
```diff
@@ -580,40 +622,91 @@ export class Dataworker {
...
         // Check if leaf value is list of BN's. For this leaf, there are no BN's not in lists.
         if (BigNumber.isBigNumber(leaf[key][0])) result[key] = leaf[key].map((val) => val.toString());
         else result[key] = leaf[key];
         return result;
       }, {});
-      console.log(prettyLeaf);
+      this.logger.debug({
+        at: "Dataworker",
+        message: `Pool rebalance leaf #${index}`,
+        leaf: prettyLeaf,
```
chrismaree, line 637: this might let the log be done on one line. best to try do this when the file is already over 600 lines long...
```suggestion
        prettyLeaf,
```

src/dataworker/Dataworker.ts
```diff
@@ -632,6 +725,156 @@ export class Dataworker {
...
+      return l1Tokens.map((l1Token) => {
+        return convertTokenAddressToSymbol(hubPoolChainId, l1Token);
+      });
+    };
+    const shortenAddresses = (addresses: string[]) => {
+      return addresses.map((address) => shortenBytes(address));
+    };
+    const shortenBytes = (root: string) => {
+      return root.substring(0, 10);
+    };
```
chrismaree, line 801: none of these should be here and I think a few of them are repeated from the utils directory. please see [FormattingUtils.ts](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/FormattingUtils.ts) and the other exports in [index.ts](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/index.ts). All of these kinds of things should be put in common utils.

chrismaree, line 801: none of these should be here and I think a few of them are repeated from the utils directory. please see [FormattingUtils.ts](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/FormattingUtils.ts) and the other exports in [index.ts](https://github.com/across-protocol/relayer-v2/blob/master/src/utils/index.ts). All of these kinds of things should be put in common utils.

src/dataworker/Dataworker.ts
```diff
@@ -580,40 +622,91 @@ export class Dataworker {
...
-    // 3. Store root + auxillary information somewhere useful for executing leaves
     // 4. Propose roots to HubPool contract.
+    const hubPoolChainId = (await this.clients.hubPoolClient.hubPool.provider.getNetwork()).chainId;
+    this._proposeRootBundle(
+      hubPoolChainId,
+      blockRangesForProposal,
+      poolRebalanceRoot.leaves,
+      poolRebalanceRoot.tree.getHexRoot(), // Tree must not be undefined if leaf count is > 0
+      relayerRefundRoot.leaves,
+      relayerRefundRoot.tree ? relayerRefundRoot.tree.getHexRoot() : EMPTY_MERKLE_ROOT,
```
chrismaree, line 700: would it make sence to refine the `getHexRoot()` method to return `EMPTY_MERKLE_ROOT` in the event it's empty so you dont need to do this switching logic here and in other places?

src/dataworker/Dataworker.ts
```diff
@@ -632,6 +725,156 @@ export class Dataworker {
...
+    );
+    const runningBalanceForLeaf = poolRebalanceRunningBalances[leafChainId][l1TokenCounterpart];
+    const netSendAmountForLeaf = this._getNetSendAmountForL1Token(
+      runningBalanceForLeaf,
+      l1TokenCounterpart,
+      endBlockForMainnet
+    );
+    return netSendAmountForLeaf.mul(toBN(-1)).gt(toBN(0)) ? netSendAmountForLeaf.mul(toBN(-1)) : toBN(0);
+  }
+
```
chrismaree, line 877: I look forward to the point where we can try and split this file into a few other modules. this file is approaching 950 lines with many many helper functions. it would be great if we could split things up into logical units. this is not something we need to do right now but it would really help with the code readability. Once all the logic we are currently working on is in we can work on this.

also, more generally trying to refactor methods to be more concise and shorter.

src/dataworker/Dataworker.ts
nicholaspai: Will do

src/dataworker/Dataworker.ts
nicholaspai: yeah there's a higher level issue here: it's very annoying to have to change code in `contracts-v2`, push a new NPM version, and then update the import here. I'll change that for now but we should discuss strategy going forward it seems unscalable

src/dataworker/Dataworker.ts
nicholaspai: Moved the more general ones that didn't involve calling the `l1Client` to FormattingUtils

src/dataworker/Dataworker.ts
mrice32: > I'll change that for now but we should discuss strategy going forward it seems unscalable

This may be overly-simplistic, but I think this is the basic tradeoff of using shared code without a monorepo. I could see putting the merkle tree lib somewhere else, but I would still imagine it being imported via npm here.

As for the 0x0 issue, I agree @nicholaspai that we'll do switching logic regardless as long as we need to error sometimes on a 0x0 root. However, I think it would be a little simpler for the library to export the `EMPTY_MERKLE_ROOT` variable and maybe have a method called `isEmpty()` to make it easy to check if empty.

Thoughts?

src/dataworker/Dataworker.ts
```diff
@@ -632,6 +729,157 @@ export class Dataworker {
...
+          [...slowRelayLeaves],
+          slowRelayRoot
+        ),
+      });
+    } catch (error) {
+      this.logger.error({ at: "Dataworker", message: "Error creating proposeRootBundleTx", error });
+    }
+  }
+
+  _generateMarkdownForRootBundle(
```
mrice32, line 766: This is a lot of code for slack formatting? Why doesn't the normal json formatter that we used previously work? Do we need to manually go through each field?

If the concern is about length, maybe there's a general JSON splitting algo that we could utilize?

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -28,9 +33,21 @@ export async function constructDataworkerClients(
     })
   );
 
-  return { ...commonClients, spokePoolSigners, spokePoolClientSearchSettings };
+  return { ...commonClients, tokenClient, spokePoolSigners, spokePoolClientSearchSettings };
 }
 
-export async function updateDataworkerClients(clients: DataworkerClients) {
-  await updateClients(clients);
+// If this is the first run then the hubPoolClient will have no whitelisted routes. If this is the case then first
```
mrice32, line 39: Two questions:
1. Aren't we generally going to run this in serverless mode, so isn't the common case that this will be the first run? (Some people may run it in looping mode, but we don't necessarily need to speed-optimize that case, right?)
2. In the multiple run case, if something is whitelisted while running, will this cause a race condition where it's unclear if the new whitelisting is seen before the clients are updated? It seems a bit safer to always execute these in series where dependencies run first and dependents run afterwards (like a DAG).

Thoughts?

mrice32 (commented):
Looks great! A few minor questions

src/clients/HubPoolClient.ts
chrismaree: I dont think so. why would we need to separate them?

src/dataworker/Dataworker.ts
chrismaree: excellent, thanks!

src/dataworker/Dataworker.ts
chrismaree: regarding updating the npm code in contracts-v2 if you have 2factor turned off it does not take more than a minute to do the release (at least for me) but I agree it is irritating.

regarding the export, you're right I think that the error state makes more sense.

src/clients/HubPoolClient.ts
nicholaspai: I wanted to make sure that Dataworker wasn't needlessly updating L2 data but its actually easy to just set `spokePoolClients = {}` for the dataworker and ignore L2 tokens

src/dataworker/Dataworker.ts
nicholaspai: Good point, ptal! https://github.com/across-protocol/contracts-v2/pull/156

src/clients/MultiCallerClient.ts
```diff
@@ -27,10 +27,15 @@ export class MultiCallerClient {
     this.transactions = [];
   }
 
-  async executeTransactionQueue() {
+  async executeTransactionQueue(simulationModeOn = false) {
```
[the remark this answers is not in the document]
nicholaspai: I'm using it locally so I don't have to submit an on-chain transaction but I can still simulate that the transaction goes through. Do you think there is a better way to test it? If so, I plan to add it to the common config vars that can be toggled with `.env` because I often get into the situation where I don't want to submit txns but want to test slack logs + simulate the txns

src/dataworker/Dataworker.ts
```diff
@@ -580,40 +626,91 @@ export class Dataworker {
...
+    if (poolRebalanceRoot.leaves.length === 0) {
+      this.logger.debug({
+        at: "Dataworker",
+        message: "No pool rebalance leaves, cannot propose",
+      });
+      return;
+    }
+
+    // 3. Check if a bundle is pending.
+    // TODO: Move the `rootBundleProposal` call to the HubPool client?
```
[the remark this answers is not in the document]
nicholaspai: Done

src/dataworker/Dataworker.ts
nicholaspai: I do several things differently to improve readability
- I convert wei values to more readable numbers with fewer decimal parts
- I convert token addresses to symbols to make it more readable

src/dataworker/DataworkerClientHelper.ts
nicholaspai: 1. @chrismaree should comment more on that, i was following his pattern here.
2. This is a good point, we should consider just always running in series

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
mrice32: Yeah, not sure about the right pattern for ths. Just didn't know how you were using it. SGTM

src/clients/TokenClient.ts
```diff
@@ -116,12 +120,35 @@ export class TokenClient {
     this.logger.info({ at: "tokenClient", message: `Approved whitelisted tokens! 💰`, mrkdwn });
   }
 
+  async setBondTokenAllowance() {
+    const ownerAddress = await this.hubPoolClient.hubPool.signer.getAddress();
+    const currentCollateralAllowance: BigNumber = await this.bondToken.allowance(
+      ownerAddress,
+      this.hubPoolClient.hubPool.address
+    );
+    if (currentCollateralAllowance.lt(toBN(MAX_SAFE_ALLOWANCE))) {
```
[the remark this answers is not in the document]
mrice32: Awesome, yeah, they should be different numbers.

nicholaspai:
Just added unit tests for `proposeRootBundle`

src/dataworker/Dataworker.ts
mrice32: Seems reasonable. I wish there were a way to do this without completely deconstructing the json, but since you've already written it, no harm done.

mrice32 (approved):
Only remaining comment: https://github.com/across-protocol/relayer-v2/pull/37#discussion_r869585638

Otherwise, LGTM!

src/dataworker/DataworkerClientHelper.ts
nicholaspai: I think you're right that we should always update first

nicholaspai: In the Dataworker case, however, the dataworker doesn't create and update the spoke pool clients until `proposeRootBundle` is called since it doesn't know which blocks to pull the spoke pool contract for (e.g. consider the case where the spoke pol contract was upgraded) so its not really a concern for the dataworker. But its probably a concern for the relayer

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
nicholaspai: I'll potentially remove this in a follow up PR but don't want to send the new bundle yet on mainnet

src/clients/TokenClient.ts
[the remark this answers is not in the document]
nicholaspai: Fixed

src/clients/MultiCallerClient.ts
```diff
@@ -1,4 +1,4 @@
-import { winston, getNetworkName, assign, Contract, runTransaction, getTarget, rejectAfterDelay } from "../utils";
```
chrismaree, line 1: ^^^

src/utils/TransactionUtils.ts
```diff
@@ -1,12 +1,12 @@
 import { AugmentedTransaction } from "../clients";
```
chrismaree, line 1: ^^^

src/utils/index.ts
```diff
@@ -35,4 +35,4 @@ export {
   AcrossConfigStore__factory as AcrossConfigStore,
 } from "@across-protocol/contracts-v2";
 
-export { getDeployedAddress, getDeployedBlockNumber, getContractInfoFromAddress } from "@across-protocol/contracts-v2";
```
chrismaree, line 38: ^^^

src/relayer/RelayerClientHelper.ts
```diff
@@ -37,31 +37,19 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
...
-      clients.tokenClient.update(),
-      clients.profitClient.update(),
-    ]);
-  }
+export async function updateRelayerClients(clients: RelayerClients) {
+  await updateClients(clients);
+  // Profit and SpokePoolClient client requires up to date HubPoolClient and rateModelClient.
+  // Token client requires up to date spokePool clients to fetch token routes.
+  await Promise.all([clients.profitClient.update(), updateSpokePoolClients(clients.spokePoolClients)]);
+  await clients.tokenClient.update();
```
chrismaree, line 52: ^

PR #40, opened 2022-05-11
feat: Make relayer, data worker, monitor run in serverless mode

chrismaree:
This PR updates the implementation to make the bots more compatible with serverless mode. The relayer, in particular, has an optimization added to help with performance when running in serverless mode.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,12 +1,17 @@
-import { delay, Logger } from "@uma/financial-templates-lib";
+import { delay, Logger, winston } from "./src/utils";
 
 import { runRelayer } from "./src/relayer";
 import { runDataworker } from "./src/dataworker";
+import { runMonitor } from "./src/monitor";
+
+let logger: winston.Logger;
 
 export async function run(): Promise<void> {
+  logger = Logger;
   if (process.argv.includes("--relayer")) await runRelayer(Logger);
   else if (process.argv.includes("--dataworker")) await runDataworker(Logger);
-  else console.log("Select either relayer OR dataworker");
+  else if (process.argv.includes("--monitor")) await runMonitor(Logger);
+  else console.log("Select either relayer, dataworker or monitor to run!");
 }
 
 if (require.main === module) {
@@ -15,8 +20,8 @@ if (require.main === module) {
       process.exit(0);
     })
     .catch(async (error) => {
-      console.error("Process exited with", error);
+      logger.error({ at: "InfrastructureEntryPoint", message: "There was an error in the main entry point!", error });
       await delay(5);
-      process.exit(1);
+      await run();
     });
 }
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,7 +1,7 @@
 import { spreadEvent, assign, Contract, BigNumber, EventSearchConfig } from "../utils";
 import { toBN, Event, ZERO_ADDRESS, winston, paginatedEventQuery, spreadEventWithBlockNumber } from "../utils";
 
-import { AcrossConfigStoreClient } from "./RateModelClient";
+import { AcrossConfigStoreClient } from "./ConfigStoreClient";
 import { Deposit, DepositWithBlock, Fill, SpeedUp, FillWithBlock } from "../interfaces/SpokePool";
 
 export class SpokePoolClient {
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -113,7 +113,7 @@ export class SpokePoolClient {
     let isValid = true;
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) {
-        this.log("debug", "Prop mismatch!", { depositVal: deposit[key].toString(), fillValue: fill[key].toString() });
+        this.log("debug", "Prop mismatch!", { key, depositVal: deposit[key], fillValue: fill[key] });
         isValid = false;
       }
     });
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -130,8 +130,11 @@ export class SpokePoolClient {
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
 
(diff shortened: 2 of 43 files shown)


index.ts
```diff
@@ -1,12 +1,17 @@
...
+import { runMonitor } from "./src/monitor";
+
+let logger: winston.Logger;
 
 export async function run(): Promise<void> {
+  logger = Logger;
   if (process.argv.includes("--relayer")) await runRelayer(Logger);
   else if (process.argv.includes("--dataworker")) await runDataworker(Logger);
-  else console.log("Select either relayer OR dataworker");
+  else if (process.argv.includes("--monitor")) await runMonitor(Logger);
```
chrismaree, line 13: added entry point to run the monitor.

src/clients/SpokePoolClient.ts
```diff
@@ -21,7 +21,8 @@ export class SpokePoolClient {
     readonly spokePool: Contract,
     readonly configStoreClient: AcrossConfigStoreClient | null, // Can be excluded. This disables some deposit validation.
     readonly chainId: number,
-    readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 }
+    readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 },
```
chrismaree, line 24: when running in serverless mode the config can define an offset to push the `fromBlock` forward to ignore old events to speed up the execution. however, getting the `EnabledDepositRoute` events should _always_ look from the deployment block as to not exclude deposit routes

src/dataworker/index.ts
```diff
@@ -1,20 +1,16 @@
...
-
 config();
+let logger: winston.Logger;
 
 export async function runDataworker(_logger: winston.Logger): Promise<void> {
   logger = _logger;
   try {
     const config = new DataworkerConfig(process.env);
-    logger.info({ at: "Dataworker#index", message: "Dataworker starting🏃‍♂️", config });
+    logger[bootLogLevel(config)]({ at: "Dataworker#index", message: "Dataworker starting👩‍🔬", config });
```
chrismaree: now log at `debug` if in serverless and `info` if not.

src/monitor/MonitorConfig.ts
```diff
@@ -1,43 +1,35 @@
...
 
 // Set modes to true that you want to enable in the AcrossMonitor bot.
 export interface BotModes {
   utilizationEnabled: boolean; // Monitors pool utilization ratio
   unknownRootBundleCallersEnabled: boolean; // Monitors relay related events triggered by non-whitelisted addresses
 }
 
-export class MonitorConfig {
-  readonly spokePoolChainIds: number[];
+export class MonitorConfig extends CommonConfig {
```
chrismaree, line 10: updated to be more like the other configs where we pull in the common config and only extend, rather than repeating everything here.

src/relayer/RelayerClientHelper.ts
```diff
@@ -15,16 +15,28 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
...
-  // Create clients for each contract for each chain.
   const spokePoolSigners = getSpokePoolSigners(baseSigner, config);
 
   const spokePools = config.spokePoolChains.map((networkId) => {
     return { networkId, contract: getDeployedContract("SpokePool", networkId, spokePoolSigners[networkId]) };
   });
   const spokePoolClients = {};
+
+  // If maxRelayerLookBack is set then offset the fromBlock to the latest block - the maxRelayerLookBack. Used in serverless mode.
+  let fromBlocks = {};
```
chrismaree, line 26: new bit of logic: if the config has `maxRelayerLookBack` then offset the `fromBlock` to the `provider.getBlockNumber() - maxRelayerLookBack` on each spokePool client. This is to avoid looking at deposit events that are too old, thereby avoiding having to query realised LP fee pct and generally decreasing the overall load on the infra when in serverless mode. 

I think a reasonable starting value for this is to try target looking back ~4 hours. Each chain propagates blocks at a diffrent rate so we need to configure the offset used on each chain differently. By looking at the evolution of the last few blocks and examining on etherscan for each chain I came up with the inital starting numbers:
From my testing a `maxRelayerLookBack` of 10000 would result in data of:
- 4 hours on mainnet takes ~ 1100 blocks.
- 4 hours on optimism takes ~ 50000 blocks.
- 4 hours on polygon takes ~ 6000 blocks.
- 4 hours on boba takes ~1100 blocks.
- 4 hours o arbitrum takes ~ 15000 blocks.

Therefore, to run this bot in serverless mode, I propose we start with the following config:
```
MAX_RELAYER_DEPOSIT_LOOK_BACK='{"1":1100,"10":50000,"137":6000,"288":1000,"42161":15000}'
```

We can always increase/decrease this time horizon. 4 hours seems like a "reasonable" window that will mean we dont miss deposits if we have down time, while still optimising the bot hugely.

Note that as this configuration is set here if you exclude a chainId from the `MAX_RELAYER_DEPOSIT_LOOK_BACK` the bot will default the the previous behavour where it will take all deposits from the deposit block.

src/relayer/RelayerClientHelper.ts
```diff
@@ -15,16 +15,29 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
...
   });
   const spokePoolClients = {};
+
+  // If maxRelayerLookBack is set then offset the fromBlock to the latest - maxRelayerLookBack. Used in serverless mode.
+  let fromBlocks = {};
+  if (config.maxRelayerLookBack != {}) {
+    const l2BlockNumbers = await Promise.all(
+      spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
+    );
+    spokePools.forEach((obj: { networkId: number; contract: Contract }, index) => {
```
nicholaspai, line 31: You can probably replace this forEach with a map where each map creates a `let fromBlock` thats initially `undefined`, tries to reset it if there's a `maxRelayerLookBack`, and then uses `fromBlock` to create an `eventSearchSettings`.

Or you could just create `fromBlocks` from a `.map`

nicholaspai (approved):
This change makes sense. Left one nit on using `.map` instead of `.forEach` and some other nits.

PR #42, opened 2022-05-11
fix: Various bugs in Dataworker

nicholaspai:
- Fixes one minor bug where we're reading the wrong `max` config variable in the Dataworker

- removes a very noisy log 
- changes filter in `_loadData` for `deposits` to compare against the ORIGIN chain block range instead of the DESTINATION chain block range

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -112,7 +112,6 @@ export class SpokePoolClient {
     let isValid = true;
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) {
-        this.log("debug", "Prop mismatch!", { depositVal: deposit[key].toString(), fillValue: fill[key].toString() });
         isValid = false;
       }
     });
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -154,7 +154,7 @@ export class SpokePoolClient {
       assign(
         this.depositsWithBlockNumbers,
         [deposit.destinationChainId],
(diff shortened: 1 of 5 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -112,7 +112,6 @@ export class SpokePoolClient {
     let isValid = true;
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) {
-        this.log("debug", "Prop mismatch!", { depositVal: deposit[key].toString(), fillValue: fill[key].toString() });
```
nicholaspai, line 115: This log is too noisy since `validateFIllForDeposit` is called in `getDepositForFill` on every single deposit to find the matching fill. So, even if a fill matches 1 out of the 10 deposits, then 9 of these logs will be emitted. We should leave the client to handle `undefined` return values from this func

src/dataworker/Dataworker.ts
```diff
@@ -75,19 +80,18 @@ export class Dataworker {
         if (!destinationClient.isUpdated)
           throw new Error(`destination SpokePoolClient with chain ID ${destinationChainId} not updated`);
 
-        const blockRangeForChain = this._getBlockRangeForChain(blockRangesForChains, Number(destinationChainId));
-
         // Store all deposits in range, for use in constructing a pool rebalance root. Save deposits with
         // their quote time block numbers so we can pull the L1 token counterparts for the quote timestamp.
         // We can safely filter `deposits` by the bundle block range because its only used to decrement running
         // balances in the pool rebalance root. This array is NOT used when matching fills with deposits. For that,
```
mrice32, line 86: Very helpful comment btw

mrice32 (approved):
LGTM!

PR #43, opened 2022-05-11
feat: validate pending root bundle and refactor

nicholaspai:
Validating pending root bundle has a fair amount of edge cases besides just recomputing roots for a block range. Most of the edge cases have to do with validating the bundle block range itself, and then there are a few other edge cases:
- PoolRebalanceRoot should never be 0x
- PoolRebalance leaf count must match expected


src/clients/HubPoolClient.ts
```diff
@@ -118,20 +120,26 @@ export class HubPoolClient {
...
-        chainIdList
-      );
-      if (bundleEvalBlockNumber >= block) {
-        endingBlockNumber = bundleEvalBlockNumber;
-        // Since events are sorted from oldest to newest, and bundle block ranges should only increase, exit as soon
-        // as we find the first block range that contains the target block.
-        break;
+      try {
+        // `getBundleEndBlockForChain` will error or return the bundle end block for the chain.
+        const bundleEvalBlockNumber = this.getBundleEndBlockForChain(
```
nicholaspai, line 129: Needed to try-catch this because `getBundleEndBlockForChain` throws if a ProposeRootBundle event isn't the right length. This means that this whole function would ALWAYS throw if the latest proposal was invalid, and then either cancelled or disputed. Instead, we should SKIP these invalid proposals and go to the next most recent one and see if it has a bundle end block that meets the function's search criteria.

This is OK to do because the `dataworker` will check if this function returns undefined.

src/clients/HubPoolClient.ts
```diff
@@ -273,6 +275,31 @@ export class HubPoolClient {
       ...executedRootBundleEvents.map((event) => spreadEventWithBlockNumber(event) as ExecutedRootBundle)
     );
 
+    this.pendingRootBundle = {
```
nicholaspai, line 278: We need all this pending root bundle data to evaluate it

src/common/Constants.ts
```diff
@@ -5,7 +5,7 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 export const CHAIN_MAX_BLOCK_LOOKBACK = {
   1: 0, // Note: 0 gets defaulted to infinity lookback
   10: 0,
-  137: 99990,
-  288: 5000,
+  137: 3490,
```
nicholaspai, line 8: Polygon, in times of stress, imposes a 3500 block lookback. So I think we should just set the default to this tighter constraint

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +703,226 @@ export class Dataworker {
     );
   }
 
-  async validateRootBundle(poolRebalanceRoot: string, relayerRefundRoot: string, slowRelayRoot: string) {
-    // Look at latest propose root bundle event earlier than a block number.
-    // Determine start block numbers for end block numbers emitted in ProposedRootBundle event using
-    // HubPoolClient.getNextBundleStartBlockNumber and passing in the mainnet block of the propose root bundle txn
-    // Construct roots locally using class functions and compare with the event we found earlier.
-    // If any roots mismatch, pinpoint the errors to give details to the caller.
+  async validateRootBundle() {
```
nicholaspai, line 706: Core new function in this PR

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +705,247 @@ export class Dataworker {
...
+          expectedSlowRelayRoot.tree.getHexRoot()
+        )
+    );
+  }
+
+  // This returns a possible next block range that could be submitted as a new root bundle, or used as a reference
+  // when evaluating  pending root bundle. The block end numbers must be less than the latest blocks for each chain ID
+  // (because we can't evaluate events in the future), and greater than the the expected start blocks, which are the
+  // greater of 0 and the latest bundle end block for an executed root bundle proposal + 1.
+  async _getWidestPossibleExpectedBlockRange(): Promise<number[][]> {
```
nicholaspai, line 935: Follow up PR will move a bunch a of these internal methods + refactor some of the longer methods into external files. I'm thinking we have a whole utils file called `DataWorkerUtils` or `DataworkerHelpers` in the same directory and just export them all here? WDYT @chrismaree ?

src/common/Constants.ts
```diff
@@ -5,7 +5,14 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
...
+  288: 4990,
   42161: 99990,
 };
+export const BUNDLE_END_BLOCK_BUFFERS = {
```
nicholaspai, line 12: These constants are all based on 5 mins of blocks for each chain.

src/dataworker/Dataworker.ts
```diff
@@ -177,13 +181,7 @@ export class Dataworker {
...
               else refundObj.refunds[fill.relayer] = refund;
             }
-          } else {
-            this.logger.debug({
-              at: "Dataworker",
-              message: `Could not find deposit for fill on origin client`,
-              fill,
-            });
-          }
+          } else allInvalidFills.push(fillWithBlock);
```
nicholaspai, line 184: So I replaced this Log because it was so noisy when there were multiple invalid fills. Instead, I wanted to summarize in the logs all invalid fills. My question for you is: am I double counting fills here?

src/dataworker/Dataworker.ts
```diff
@@ -100,25 +106,24 @@ export class Dataworker {
           );
         deposits.push(...newDeposits);
 
-        // For each fill within the block range, look up associated deposit.
-        const blockRangeForChain = this._getBlockRangeForChain(blockRangesForChains, Number(destinationChainId));
-        const fillsForOriginChain: FillWithBlock[] = destinationClient
-          .getFillsWithBlockForOriginChain(Number(originChainId))
-          .filter(
```
nicholaspai, line 107: This `filter` was actually a bug because its done too early. We shouldnt fill out fills until we store them in `allValidFills` since the pool rebalance root needs to look up ALL fills to figure out if a slow relay was sent for a deposit on a previous root bundle

src/dataworker/Dataworker.ts
```diff
@@ -400,185 +417,53 @@ export class Dataworker {
 
     const { fillsToRefund, deposits, allValidFills } = this._loadData(blockRangesForChains, spokePoolClients);
 
-    // Running balances are the amount of tokens that we need to send to each SpokePool to pay for all instant and
```
nicholaspai, line 403: I began to refactor the code in this PR and wanted to get thoughts on general pattern. I didn't change code but I a lot of the PoolRebalance logic into `PoolRebalanceUtils`, which itself uses functions exported by `FillUtils`

src/dataworker/FillUtils.ts
```diff
@@ -0,0 +1,69 @@
+import { HubPoolClient } from "../clients";
+import { Fill, FillWithBlock } from "../interfaces";
```
nicholaspai, line 2: All copied from `DataWorker.ts` internal methods

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -0,0 +1,243 @@
+import { AcrossConfigStoreClient, HubPoolClient } from "../clients";
```
nicholaspai, line 1: All copied from `DataWorker.ts` internal methods

src/clients/HubPoolClient.ts
```diff
@@ -118,50 +120,70 @@ export class HubPoolClient {
...
+  }
 
-    // Once that event is found, search for the ProposeRootBundle event that is as late as possible, but earlier than
-    // the RootBundleExecuted event we just identified.
-    const mostRecentProposeRootBundleEventWithChain = sortEventsDescending(this.proposedRootBundles).find(
-      (rootBundle: ProposedRootBundle) => rootBundle.blockNumber <= latestRootBundleExecutedEvent.blockNumber
+  getNextBundleStartBlockNumber(chainIdList: number[], latestMainnetBlock: number, chainId: number): number {
+    // Search for latest ProposeRootBundleExecuted event followed by all of its RootBundleExecuted event suggesting
+    // that all pool rebalance leaves were executed. This ignores any proposed bundles that were partially executed.
+    const latestFullyExecutedPoolRebalanceRoot = sortEventsDescending(this.proposedRootBundles).find(
```
nicholaspai, line 157: This change fixes the issue where a chain has fills as their destination chain, but no refunds and no deposits. It finds the latest (i.e. before `latestMainnetBlock` ProposedRootBundle event with ALL leaves executed, and then uses the bundle end blocke emitted by that event.

src/clients/HubPoolClient.ts
```diff
@@ -118,50 +120,70 @@ export class HubPoolClient {
...
-    // Search for the latest RootBundleExecuted event with a matching chainId while still being earlier than the
-    // latest block.
-    const latestRootBundleExecutedEvent = sortEventsDescending(this.executedRootBundles).find(
-      (event: ExecutedRootBundle) => event.blockNumber <= latestMainnetBlock && event.chainId === chainId
-    );
-
-    // TODO: If no event for chain ID, then we can return a conservative default starting block like 0,
-    // or we could throw an Error.
-    if (!latestRootBundleExecutedEvent) return 0;
+  getMostRecentProposedRootBundle(latestBlockToSearch: number) {
```
nicholaspai, line 148: This function used by `validateRootBundle` script only

src/common/ClientHelper.ts
```diff
@@ -17,6 +26,34 @@ export function getSpokePoolSigners(baseSigner: Wallet, config: CommonConfig): {
   );
 }
 
+export async function constructSpokePoolClientsForBlockAndUpdate(
```
nicholaspai, line 29: Copied from Dataworker.ts

src/dataworker/Dataworker.ts
nicholaspai: Update: I refactored a TON more code since this comment, there is just a lot of shifted around code in this PR hence the large diff

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +487,255 @@ export class Dataworker {
     );
   }
 
-  async validateRootBundle(poolRebalanceRoot: string, relayerRefundRoot: string, slowRelayRoot: string) {
-    // Look at latest propose root bundle event earlier than a block number.
-    // Determine start block numbers for end block numbers emitted in ProposedRootBundle event using
-    // HubPoolClient.getNextBundleStartBlockNumber and passing in the mainnet block of the propose root bundle txn
-    // Construct roots locally using class functions and compare with the event we found earlier.
-    // If any roots mismatch, pinpoint the errors to give details to the caller.
+  async validatePendingRootBundle(submitDispute = true) {
```
nicholaspai, line 490: This function and the next are all the new logic that was the original point of this PR: validate the pending root bundle and submit a dispute if invalid

src/dataworker/DataworkerUtils.ts
```diff
@@ -0,0 +1,14 @@
+export function getBlockRangeForChain(
```
nicholaspai, line 1: Copied from Dataworker.ts

mrice32 (approved):
Looks great! Only minor comments and questions.

This is awesome @nicholaspai -- very very cool stuff in this PR.

src/clients/HubPoolClient.ts
```diff
@@ -118,50 +120,70 @@ export class HubPoolClient {
...
+        ) as ProposedRootBundle;
+        const followingProposedRootBundleBlock = followingProposedRootBundle
+          ? followingProposedRootBundle.blockNumber
+          : latestMainnetBlock;
+        // Figure out how many pool rebalance leaves were executed for this root bundle, but only count executed
+        // leaves before the `latestMainnetBlock` since that's the block height snapshot we're querying.
+        const executedPoolRebalanceLeaves = sortEventsAscending(this.executedRootBundles).filter(
+          (executedLeaf: ExecutedRootBundle) =>
+            executedLeaf.blockNumber <= latestMainnetBlock &&
+            executedLeaf.blockNumber >= proposedRootBundle.blockNumber &&
```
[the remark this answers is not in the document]
nicholaspai: Agreed on both these points, i'm glad you looked at these lines as I was going back and forth on the exclusive/inclusive signs

src/clients/HubPoolClient.ts
```diff
@@ -273,6 +289,31 @@ export class HubPoolClient {
...
+      proposalBlockNumber: undefined,
+    };
+    // If pending proposal is active, even if it passed the liveness period, then we can load the following root
+    // bundle properties: `bundleEvaluationBlockNumbers` and `proposalBlockNumber`, otherwise they are not relevant.
+    // We can assume that the pool rebalance root is never empty if there is a pending root on-chain, regardless
+    // if the root is disputable.
+    if (this.pendingRootBundle.poolRebalanceRoot !== EMPTY_MERKLE_ROOT) {
+      // We make a potentially dangerous assumption here that the most recent ProposeRootBundle event found in the
+      // the above event search corresponds to the pending root bundle. This could NOT be the case if
+      // `this.eventSearchConfig.toBlock !== undefined`, but in most cases this is true because we set the `toBlock`
```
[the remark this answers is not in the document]
nicholaspai: I use `pendingRootBundle` only in `DataWorker.validatePendingRootBundle` and `DataWprler.proposeRootBundle` to determine if there is an active root bundle and whether it can propose/dispute. In the latter case, it needs to know which RootBundle props to validate.

So I actually just want to query `HubPool.rootBundleProposal` which I do here. This comment and `if` block here is to load additional information that can't just be queried on chain: for example, the `bundleEvaluationBlockNumbers` I want to store as `numbers` so I perform a `.map`, while the `proposalBlockNumber` isn't stored on-chain so I need to query the events to find the one corresponding to the active root bundle. Since we're calling `HubPool.rootBundleProposal` I think we NEED to always get the most recent one (even if it was changed after the to block in the block search config), because otherwise we don't properly know if we can propose/dispute.

I'm not really sure I understand your question, but this is my intended use case. This is why I comment that "we make a dangerous assumption that the latest ProposeRootBundle event is the most active root bundle, becuase if the `toBlock` is not set to the latest, then this assumption isn't true, and we'll be loading incorrect data into the client.

src/dataworker/Dataworker.ts
```diff
@@ -85,167 +97,100 @@ export class Dataworker {
...
+              blockRangesForChains,
+              this.chainIdListForBundleEvaluationBlockNumbers
             );
+            updateFillsToRefundWithValidFill(fillsToRefund, fill, chainToSendRefundTo, repaymentToken);
+
+            // Save deposit as one that is eligible for a slow fill, since there is a fill
+            // for the deposit in this epoch. We save whether this fill is the first fill for the deposit, because
+            // if a deposit has its first fill in this block range, then we can send a slow fill payment to complete
+            // the deposit.
+            updateUnfilledDepositsWithMatchedDeposit(fill, matchedDeposit, unfilledDepositsForOriginChain);
```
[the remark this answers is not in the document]
nicholaspai: yeah exactly, if the fill is the FIRST for the deposit, then its potentially a slow fill. However, if other fills complete the deposit, then we can remove it later. Another way to have done this is to associate an `unfilledAmount` with the deposit, and keep updating that if a fill's`totalFilledAmount` would decrease the `unfilledAmount`

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +487,255 @@ export class Dataworker {
...
-    // Look at latest propose root bundle event earlier than a block number.
-    // Determine start block numbers for end block numbers emitted in ProposedRootBundle event using
-    // HubPoolClient.getNextBundleStartBlockNumber and passing in the mainnet block of the propose root bundle txn
-    // Construct roots locally using class functions and compare with the event we found earlier.
-    // If any roots mismatch, pinpoint the errors to give details to the caller.
+  async validatePendingRootBundle(submitDispute = true) {
+    if (!this.clients.hubPoolClient.isUpdated) throw new Error(`HubPoolClient not updated`);
+    const hubPoolChainId = (await this.clients.hubPoolClient.hubPool.provider.getNetwork()).chainId;
+
+    // Exit early if a bundle is pending.
```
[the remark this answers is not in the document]
nicholaspai: yep, nice catch thanks

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +487,255 @@ export class Dataworker {
...
+    });
+
+    // Exit early if challenge period timestamp has passed:
+    if (this.clients.hubPoolClient.currentTime > pendingRootBundle.challengePeriodEndTimestamp) {
+      this.logger.debug({
+        at: "Dataworke#validater",
+        message: "Challenge period passed, cannot dispute",
+      });
+      return;
+    }
```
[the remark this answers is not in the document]
nicholaspai: yeah, good things to war games test @chrismaree , lets add to list of lines we want to explicitly stress test

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: when running the `src/scripts/validateRootBundle.ts` scripts :D

This way I can re-use most of this validation logic while allowing someone on DevX to run this script with less set up

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +487,255 @@ export class Dataworker {
...
+
+    const widestPossibleExpectedBlockRange = await PoolRebalanceUtils.getWidestPossibleExpectedBlockRange(
+      this.chainIdListForBundleEvaluationBlockNumbers,
+      this.clients,
+      this.clients.hubPoolClient.latestBlockNumber
+    );
+    await this.validateRootBundle(hubPoolChainId, widestPossibleExpectedBlockRange, pendingRootBundle, submitDispute);
```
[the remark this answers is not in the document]
nicholaspai: Yeah great point, I actually tried to do this originally but got too cute and wanted to also have the `reason` feed into the `logger` calls but that got too complicated since log messages and mrkdwn messages differed and contained diff props, but this suggestion is very clean!

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +487,255 @@ export class Dataworker {
...
+      } else {
+        this.logger.debug({
+          at: "Dataworker#validate",
+          message: "A bundle end block is > latest block but within buffer, skipping",
+          expectedEndBlocks: widestPossibleExpectedBlockRange.map((range) => range[1]),
+          pendingEndBlocks: rootBundle.bundleEvaluationBlockNumbers,
+          endBlockBuffers,
+        });
+      }
+      return false;
```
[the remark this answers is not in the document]
nicholaspai: yeah I'm going to change this to return true

src/clients/HubPoolClient.ts
nicholaspai: Hm so which errors in `getBundleEndBlockForChain` do you think we should actually throw on? I think they're both ok to skip in this case.

nicholaspai: The complexity here with my architecture is that `getBundleEndBlockForChain` is also used in `getNextBundleStartBlockNumber` in which case we actually do want it to error. Do you think it'd be better to check some of the assertions from `getBundleEndBlockForChain` before entering the function?

Also I just realized that `getBundleEndBlockForChain` will error if the inputted chainId list length is diff from the inputted rootbundle's block eval list, which means it will error whenever we add a chain to the list.

nicholaspai: because of that last case, I think its best that `getBundleEndBlockForChain` DOESN'T perform list length checks, only performs the `indexOf` test, and returns 0 if `indexOf == -1`

nicholaspai: Ok here's a commit that I think simplifies this logic a bit by removing errors thrown https://github.com/across-protocol/relayer-v2/commit/75c1839061fca685e4edac47ed65d1dbb31c4e3a

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
mrice32: Ah, that makes sense! +1

src/scripts/validateRootBundle.ts
```diff
@@ -0,0 +1,142 @@
...
+    config.tokenTransferThresholdOverride,
+    config.blockRangeEndBlockBuffer
+  );
+
+  // Construct blockfinder to figure out which block corresponds with the disputed price request time.
+  const blockFinder = new BlockFinder(
+    clients.configStoreClient.configStore.provider.getBlock.bind(clients.configStoreClient.configStore.provider)
+  );
+  const priceRequestBlock = (await blockFinder.getBlockForTimestamp(priceRequestTime)).number;
+  await updateDataworkerClients(clients);
```
chrismaree, line 64: can these be grouped in to a promise.all?

src/scripts/validateRootBundle.ts
```diff
@@ -0,0 +1,142 @@
+// How to run:
+// 1. You might need to set the following environment variables:
+//    HUB_CHAIN_ID=1
+//    MAX_BLOCK_LOOK_BACK='{"1":0,"10":0,"137":3499,"288":4999,"42161":0}'
+//    NODE_URL_1=https://mainnet.infura.io/v3/KEY
+//    NODE_URL_10=https://optimism-mainnet.infura.io/v3/KEY
+//    NODE_URL_137=https://polygon-mainnet.infura.io/v3/KEY
+//    NODE_URL_288=https://mainnet.boba.network/
+//    NODE_URL_42161=https://arb-mainnet.g.alchemy.com/v2/KEY
+// 2. REQUEST_TIME=1652726987 ts-node ./src/scripts/validateRootBundle.ts --wallet mnemonic
```
chrismaree, line 10: wdyt about changing this to a hardhat task like this one: https://github.com/across-protocol/contracts-v2/blob/3790f7fb6559990fc6830051dc8f15b24c965fcb/tasks/enableL1TokenAcrossEcosystem.ts#L7

src/scripts/validateRootBundle.ts
```diff
@@ -0,0 +1,142 @@
+// How to run:
```
chrismaree, line 1: what's awesome about this script is we could literally add this to a slack command that people can run in the slack channel to validate a given bundle.

src/clients/HubPoolClient.ts
```diff
@@ -118,15 +120,19 @@ export class HubPoolClient {
...
     let endingBlockNumber: number;
     for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+      // `getBundleEndBlockForChain` will error or return the bundle end block for the chain.
       const bundleEvalBlockNumber = this.getBundleEndBlockForChain(
         rootBundle as ProposedRootBundle,
         chain,
         chainIdList
       );
+
+      // If chain list doesn't contain chain, then bundleEvalBlockNumber returns 0 and the following check
```
chrismaree, line 134: this is intended so the bot stops running and blows up, right?

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
mrice32: Makes sense! Do you think we should double check we got the right one by checking that all the roots match?

src/clients/HubPoolClient.ts
mrice32: What I'm really trying to get at is that there could be errors that are thrown that are _not directly thrown by your code_. For instance, there could be a bug where you try to call `someMap["abc"].someField` where `someMap["abc"]` is undefined in some unexpected case, causing the whole thing to error.

It seems like you're intending to catch errors that happen in _specific cases_. So for each of those cases, you can throw a specific type of error that you can check for so you don't swallow errors that are a result of bugs and not the intended cases you're trying to capture.

In general, I think we should avoid using try-catch to branch, but sometimes that's unavoidable, so when we do use it, I think it's best to single out the errors you're trying to capture rather than catching everything.

Does that make sense?

src/clients/HubPoolClient.ts
```diff
@@ -124,22 +124,20 @@ export class HubPoolClient {
...
-        if (bundleEvalBlockNumber >= block) {
-          endingBlockNumber = bundleEvalBlockNumber;
-          // Since events are sorted from oldest to newest, and bundle block ranges should only increase, exit as soon
-          // as we find the first block range that contains the target block.
-          break;
-        }
-      } catch (err) {
-        // Found ProposeRootBundleEvent with invalid bundle block list, skip it.
-        continue;
+      // `getBundleEndBlockForChain` will error or return the bundle end block for the chain.
```
mrice32, line 127: Is this still true?

src/clients/HubPoolClient.ts
```diff
@@ -338,10 +337,11 @@ export class HubPoolClient {
     chainIdList: number[]
   ): number {
     const bundleEvaluationBlockNumbers: BigNumber[] = proposeRootBundleEvent.bundleEvaluationBlockNumbers;
-    if (bundleEvaluationBlockNumbers.length !== chainIdList.length)
-      throw new Error("Chain ID list and bundle block eval range list length do not match");
     const chainIdIndex = chainIdList.indexOf(chainId);
-    if (chainIdIndex === -1) throw new Error(`Can't find chainId ${chainId} in chainIdList ${chainIdList}`);
+    if (chainIdIndex === -1) return 0;
```
mrice32, line 348: +1 -- I think 0 makes sense here as there is no end block, so it effectively ends at 0.

nit: I would add a comment on the function that explains the cases where it returns 0.

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
mrice32: I think the way you have it makes sense. Just wanted to double check.

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +496,260 @@ export class Dataworker {
...
+    const widestPossibleExpectedBlockRange = await PoolRebalanceUtils.getWidestPossibleExpectedBlockRange(
+      this.chainIdListForBundleEvaluationBlockNumbers,
+      this.clients,
+      this.clients.hubPoolClient.latestBlockNumber
+    );
+    const { valid, reason } = await this.validateRootBundle(
+      hubPoolChainId,
+      widestPossibleExpectedBlockRange,
+      pendingRootBundle,
+      submitDispute
```
mrice32, line 537: Do we still need to pass in submitDispute here?

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +496,260 @@ export class Dataworker {
...
+    );
+    if (!valid) this._submitDisputeWithMrkdwn(hubPoolChainId, reason);
```
mrice32, line 539: Shouldn't we only do this if submitDispute === true?

src/clients/HubPoolClient.ts
```diff
@@ -137,31 +143,59 @@ export class HubPoolClient {
...
+    const latestFullyExecutedPoolRebalanceRoot = sortEventsDescending(this.proposedRootBundles).find(
+      (proposedRootBundle: ProposedRootBundle) => {
+        if (proposedRootBundle.blockNumber > latestMainnetBlock) return false;
+        const followingProposedRootBundle = sortEventsAscending(this.proposedRootBundles).find(
+          (_followingProposedRootBundle: ProposedRootBundle) =>
+            _followingProposedRootBundle.blockNumber > proposedRootBundle.blockNumber
+        ) as ProposedRootBundle;
+        const followingProposedRootBundleBlock = followingProposedRootBundle
+          ? followingProposedRootBundle.blockNumber
+          : latestMainnetBlock;
```
chrismaree, line 164: I feel that one thing that would help with code readability is if some of these lines were shorter. if each of these were on only one line (due to shorter/more concise variable names) this would be 2 lines instead of 6. I know this is a nit and stylistic I just find it easer when there is massive amount of code. I can spend some time once all the current tasks are done trying to simplify some of this stuff.

chrismaree: the logic, though totally makes sence. thank you for adding such detailed comments throughout this implementation.

src/clients/HubPoolClient.ts
```diff
@@ -137,31 +143,59 @@ export class HubPoolClient {
...
+            executedLeaf.blockNumber <= latestMainnetBlock &&
+            // Note: We can use > instead of >= here because a leaf can never be executed in same block as its root
+            // proposal due to bundle liveness enforced by HubPool. This importantly avoids the edge case
+            // where the execution all leaves occurs in the same block as the next proposal, leading us to think
+            // that the next proposal is fully executed when its not.
+            executedLeaf.blockNumber > proposedRootBundle.blockNumber &&
+            // This <= makes sense because you could have an execution of a leaf in the same block as the proposal of
+            // the next.
+            executedLeaf.blockNumber <= followingProposedRootBundleBlock
+        ) as ExecutedRootBundle[];
```
chrismaree, line 178: would it not make sence to refactor these into filter functions within this class or prehaps in a util?

src/clients/HubPoolClient.ts
nicholaspai: nope!

src/clients/HubPoolClient.ts
chrismaree: can you not use `spreadEvent` and then append `bundleEvaluationBlockNumbers: undefined` and `proposalBlockNumber: undefined` onto it? something like this:
```ts
this.pendingRootBundle = { ... spreadEvent(pendingRootBundleProposal), bundleEvaluationBlockNumbers: undefined, proposalBlockNumber: undefined}
```

src/dataworker/Dataworker.ts
nicholaspai: nope

src/dataworker/Dataworker.ts
nicholaspai: Removed `submitDispute` so can disregard

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
chrismaree: what you describe nick makes sence: you want the latest information when choosing to execute the dispute.

src/dataworker/Dataworker.ts
```diff
@@ -85,167 +101,101 @@ export class Dataworker {
...
-        const existingCount = result[fill.destinationChainId];
-        result[fill.destinationChainId] = existingCount === undefined ? 1 : existingCount + 1;
-        return result;
-      }, {}),
+      depositsInRangeByOriginChain: getDepositCountGroupedByToken(deposits),
+      allValidFillsInRangeByDestinationChain: getFillCountGroupedByToken(allValidFillsInRange),
+      fillsToRefundInRangeByRepaymentChain: getFillsToRefundCountGroupedByRepaymentChain(fillsToRefund),
+      unfilledDepositsByDestinationChain: getUnfilledDepositCountGroupedByProp(unfilledDeposits, "destinationChainId"),
+      allValidFillsByDestinationChain: getFillCountGroupedByProp(allValidFills, "destinationChainId"),
+      allInvalidFillsInRangeByDestinationChain: getFillCountGroupedByToken(allInvalidFillsInRange),
```
chrismaree, line 185: this is excellent.

src/dataworker/Dataworker.ts
chrismaree: thank you for doing this. the new split makes a lot of sence.

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +496,260 @@ export class Dataworker {
...
+    });
+
+    // Exit early if challenge period timestamp has passed:
+    if (this.clients.hubPoolClient.currentTime > pendingRootBundle.challengePeriodEndTimestamp) {
+      this.logger.debug({
+        at: "Dataworke#validater",
+        message: "Challenge period passed, cannot dispute",
+      });
+      return;
+    }
```
chrismaree, line 526: These are all already collapsable lines.
```suggestion
   // Exit early if a bundle is not pending.
    if (!this.clients.hubPoolClient.hasPendingProposal()) {
      this.logger.debug({ at: "Dataworker#validate", message: "No pending proposal, nothing to validate" });
      return;
    }

    const pendingRootBundle = this.clients.hubPoolClient.getPendingRootBundleProposal();
    this.logger.debug({ at: "Dataworker#validate", message: "Found pending proposal", pendingRootBundle });

    // Exit early if challenge period timestamp has passed:
    if (this.clients.hubPoolClient.currentTime > pendingRootBundle.challengePeriodEndTimestamp) {
      this.logger.debug({ at: "Dataworke#validater", message: "Challenge period passed, cannot dispute" });
      return;
    }
```

src/dataworker/Dataworker.ts
```diff
@@ -731,12 +496,260 @@ export class Dataworker {
...
+    if (rootBundle.poolRebalanceRoot === EMPTY_MERKLE_ROOT) {
+      this.logger.debug({
+        at: "Dataworker#validate",
+        message: "Empty pool rebalance root, submitting dispute",
+        rootBundle,
+      });
+      return {
+        valid: false,
+        reason: `Disputed pending root bundle with empty pool rebalance root`,
+      };
```
chrismaree, line 558:
```suggestion
      return { valid: false,  reason: `Disputed pending root bundle with empty pool rebalance root`};
```

chrismaree (approved):
there is an unbelievable amount of excellent work in here. good job @nicholaspai !

I'm still working on reading through a lot of this code in more detail but the critical sections (dispute, validation and interactions with other parts of the system) seem to be good to me!

src/scripts/validateRootBundle.ts
nicholaspai: I kind of like having this look as much as possible as the bot so the setup can be more easily copy and pasted. I might just refactor some of that setup and reuse it here

src/clients/HubPoolClient.ts
nicholaspai: Hm this function is used to find the root bundle containing a fill when evaluating slow fills, so its OK if it hits this edge case. In this case, it means usually that the current fill isn't yet contained by a root bundle, whcih is pretty common when proposing new bundles. So I fear this log would get noisy

src/clients/HubPoolClient.ts
nicholaspai: Great point, done and should be more readable!

src/clients/HubPoolClient.ts
nicholaspai: Will do!

src/common/Constants.ts
```diff
@@ -5,7 +5,14 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
...
+  1: 20, // At 15s/block, 20 blocks = 5 mins
+  10: 20, // At 15s/block, 20 blocks = 5 mins
```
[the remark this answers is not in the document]
nicholaspai: i don't think so, these seem fine. I couldn't easily figure out the L2 block times so I pegged a lot of them at 15s/block

PR #53, opened 2022-05-18
feat: Add demo for finalizer script

nicholaspai:
This PR is created mostly to start a discussion around how we want to organize the L2 finalizer code, but also to play with the `@arbitrum/sdk` module which we'll need to use and replace the soon to be deprecated `arb-ts` module.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -123,8 +123,7 @@ export class HubPoolClient {
     const nextRootBundle = this.getFollowingRootBundle(rootBundle);
     const executedLeafCount = this.getExecutedLeavesForRootBundle(
       rootBundle,
-      latestMainnetBlock,
-      nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock
+      nextRootBundle ? Math.min(nextRootBundle.blockNumber, latestMainnetBlock) : latestMainnetBlock
     );
     return executedLeafCount.length === rootBundle.poolRebalanceLeafCount;
   }
@@ -137,8 +136,6 @@ export class HubPoolClient {
     chain: number,
     chainIdList: number[]
   ): number | undefined {
-    // TODO: Unit test this very important function!
-
     let endingBlockNumber: number;
     for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
       if (!this.isRootBundleValid(rootBundle, latestMainnetBlock)) continue;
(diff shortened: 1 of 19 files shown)


src/finalizer/index.ts
```diff
@@ -0,0 +1,213 @@
...
+import { L2TransactionReceipt, getL2Network, L2ToL1MessageStatus, L2ToL1MessageWriter } from "@arbitrum/sdk";
+import { RelayerConfig } from "../relayer/RelayerConfig";
+import { constructSpokePoolClientsWithLookback } from "../relayer/RelayerClientHelper";
+import { TokensBridged } from "../interfaces";
+
+// How to run:
+// - Set same config you'd need to run dataworker or relayer in terms of L2 node urls
+// - ts-node ./src/finalizer/index.ts --wallet mnemonic
+
+export async function run(logger: winston.Logger, config: RelayerConfig): Promise<void> {
```
nicholaspai, line 17: This is the main script that I added, all the other changes are just refactoring around the dataworker

src/clients/HubPoolClient.ts
```diff
@@ -119,11 +119,30 @@ export class HubPoolClient {
...
+    latestMainnetBlock: number,
+    block: number,
+    chain: number,
+    chainIdList: number[]
+  ): number | undefined {
+    // TODO: Unit test this very important function!
+
     let endingBlockNumber: number;
     for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+      if (!this.isRootBundleValid(rootBundle, latestMainnetBlock)) continue;
```
nicholaspai, line 144: We weren't ignoring invalid bundles and therefore unexpectedly reading false blocks such as a really high arbitrum block that i set while testing out invalid bundles. Valid bundle == one where all leaves are executed

src/dataworker/DataworkerUtils.ts
```diff
@@ -12,3 +27,244 @@ export function getBlockRangeForChain(
...
+    (deposit: UnfilledDeposit): RelayData => ({
+      depositor: deposit.deposit.depositor,
+      recipient: deposit.deposit.recipient,
+      destinationToken: deposit.deposit.destinationToken,
+      amount: deposit.deposit.amount,
+      originChainId: deposit.deposit.originChainId,
+      destinationChainId: deposit.deposit.destinationChainId,
+      realizedLpFeePct: deposit.deposit.realizedLpFeePct,
+      relayerFeePct: deposit.deposit.relayerFeePct,
+      depositId: deposit.deposit.depositId,
```
chrismaree, line 88: can this not be cleaned up with `spreadEvent`

src/finalizer/index.ts
chrismaree: this script is awesome. is the goal to have this running in the same way we run the other bots wherein this is a totally separate bot instance? we'll have a `relayer`, `dataworker`, `monitor` and `finalizer` bot?

if so, there is a lot of refactoring that we'd need to do here. for example all the client related things should be pulled into common utils and client helpers to clean this up. there should be no signers initialized from within this class, either. basically it should follow the same pattern as the other bots. this can be done at a later point as well.

chrismaree (approved):
this is an excellent starting point. I think this'll need to be pretty heavily refactored before it'll be usable within the bots (i.e this only works with arbitrum & the arbitrum/sdk right now) but will be useful to build on!

src/finalizer/index.ts
```diff
@@ -0,0 +1,213 @@
...
```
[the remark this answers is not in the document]
nicholaspai: Totally. Just wanted a starting point that we can iterate from 1->2 on instead of 0->1

src/finalizer/index.ts
nicholaspai: Yeah I imagine we'd have a separate finalizer bot

This script will have a ton of refactoring to do I agree, i'm not really trying to opine on the structure of the `finalizer` bot in this PR (check the PR description where I emphasize this).

I just wanted to get some code started. The script even has a comment at the top that explains that its not supposed to be used in an automated environment in production.

src/dataworker/DataworkerUtils.ts
nicholaspai: Thoughts on changing the input type of `spreadEvent` from `event: Event` to `event: { args: any }`? We only care that the object has `.args` and we extract everything from there.

nicholaspai: The downside to this however is that in order to use `spreadEvent` with `deposit.deposit` you have to do some nasty casting of `deposit.deposit` ==> `deposit.args`

PR #55, opened 2022-05-20
feat: add pool rebalance and relayer refund leaf execution

mrice32:
This PR includes execution logic and simple lifecycle tests for pool rebalance leaves, relayer refund leaves, and slow relay leaves.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -37,6 +37,10 @@ export class HubPoolClient {
     return this.pendingRootBundle !== undefined && this.pendingRootBundle.unclaimedPoolRebalanceLeafCount > 0;
   }
 
+  getProposedRootBundles() {
+    return this.proposedRootBundles;
+  }
+
   getSpokePoolForBlock(block: number, chain: number): string {
     if (!this.crossChainContracts[chain]) throw new Error(`No cross chain contracts set for ${chain}`);
     const mostRecentSpokePoolUpdatebeforeBlock = (
@@ -173,24 +177,26 @@ export class HubPoolClient {
     ) as ExecutedRootBundle[];
   }
 
-  getNextBundleStartBlockNumber(chainIdList: number[], latestMainnetBlock: number, chainId: number): number {
+  getLatestFullyExecutedRootBundle(latestMainnetBlock: number): ProposedRootBundle | undefined {
     // Search for latest ProposeRootBundleExecuted event followed by all of its RootBundleExecuted event suggesting
     // that all pool rebalance leaves were executed. This ignores any proposed bundles that were partially executed.
-    const latestFullyExecutedPoolRebalanceRoot = sortEventsDescending(this.proposedRootBundles).find(
-      (rootBundle: ProposedRootBundle) => {
-        if (rootBundle.blockNumber > latestMainnetBlock) return false;
-        const nextRootBundle = this.getFollowingRootBundle(rootBundle);
-        const followingProposedRootBundleBlock = nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock;
-        // Figure out how many pool rebalance leaves were executed for this root bundle, but only count executed
-        // leaves before the `latestMainnetBlock` since that's the block height snapshot we're querying.
-        const executedPoolRebalanceLeaves = this.getExecutedLeavesForRootBundle(
-          rootBundle,
-          latestMainnetBlock,
-          followingProposedRootBundleBlock
-        );
-        return executedPoolRebalanceLeaves.length === rootBundle.poolRebalanceLeafCount;
-      }
-    ) as ProposedRootBundle;
+    return sortEventsDescending(this.proposedRootBundles).find((rootBundle: ProposedRootBundle) => {
+      if (rootBundle.blockNumber > latestMainnetBlock) return false;
+      const nextRootBundle = this.getFollowingRootBundle(rootBundle);
+      const followingProposedRootBundleBlock = nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock;
+      // Figure out how many pool rebalance leaves were executed for this root bundle, but only count executed
+      // leaves before the `latestMainnetBlock` since that's the block height snapshot we're querying.
+      const executedPoolRebalanceLeaves = this.getExecutedLeavesForRootBundle(
+        rootBundle,
+        latestMainnetBlock,
+        followingProposedRootBundleBlock
+      );
+      return executedPoolRebalanceLeaves.length === rootBundle.poolRebalanceLeafCount;
+    });
+  }
+
+  getNextBundleStartBlockNumber(chainIdList: number[], latestMainnetBlock: number, chainId: number): number {
+    const latestFullyExecutedPoolRebalanceRoot = this.getLatestFullyExecutedRootBundle(latestMainnetBlock);
 
     // If no event, then we can return a conservative default starting block like 0,
     // or we could throw an Error.
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -17,6 +17,7 @@ export class MultiCallerClient {
   // caller to send a transaction. The transaction might not be executable, which should be filtered later.
   enqueueTransaction(transaction: AugmentedTransaction) {
     this.transactions.push(transaction);
+    // console.log("method", transaction.method, "args", transaction.args.map((elt) => elt.toString()).join(","));
   }
 
   transactionCount() {
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -17,7 +17,6 @@ export class MultiCallerClient {
   // caller to send a transaction. The transaction might not be executable, which should be filtered later.
   enqueueTransaction(transaction: AugmentedTransaction) {
     this.transactions.push(transaction);
-    // console.log("method", transaction.method, "args", transaction.args.map((elt) => elt.toString()).join(","));
   }
 
   transactionCount() {
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -145,7 +145,7 @@ export class MultiCallerClient {
   buildMultiCallBundle(transactions: AugmentedTransaction[]) {
     // Validate all transactions in the batch have the same target contract.
     const target = transactions[0].contract;
-    if (transactions.every((tx) => tx.contract.address != target.address)) {
+    if (transactions.every((tx) => tx.contract.address !== target.address)) {
       this.logger.error({
         at: "MultiCallerClient",
         message: "some transactions in the bundle contain different targets",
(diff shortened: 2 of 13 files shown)


test/utils/utils.ts
```diff
@@ -129,7 +129,7 @@ export async function deployAndConfigureHubPool(
     await utils.getContractFactory("HubPool", signer)
   ).deploy(lpTokenFactory.address, finderAddress, zeroAddress, timerAddress);
 
-  const mockAdapter = await (await utils.getContractFactory("Mock_Adapter", signer)).deploy();
+  const mockAdapter = await (await utils.getContractFactory("Ethereum_Adapter", signer)).deploy();
```
mrice32, line 132: This changes very little -- it just means the messages are passed along via calls.

src/clients/HubPoolClient.ts
```diff
@@ -173,24 +177,26 @@ export class HubPoolClient {
...
-        // leaves before the `latestMainnetBlock` since that's the block height snapshot we're querying.
-        const executedPoolRebalanceLeaves = this.getExecutedLeavesForRootBundle(
-          rootBundle,
-          latestMainnetBlock,
-          followingProposedRootBundleBlock
-        );
-        return executedPoolRebalanceLeaves.length === rootBundle.poolRebalanceLeafCount;
-      }
-    ) as ProposedRootBundle;
+    return sortEventsDescending(this.proposedRootBundles).find((rootBundle: ProposedRootBundle) => {
```
mrice32, line 183: I just separated out this functionality for use elsewhere.

src/clients/MultiCallerClient.ts
```diff
@@ -17,6 +17,7 @@ export class MultiCallerClient {
   // caller to send a transaction. The transaction might not be executable, which should be filtered later.
   enqueueTransaction(transaction: AugmentedTransaction) {
     this.transactions.push(transaction);
+    // console.log("method", transaction.method, "args", transaction.args.map((elt) => elt.toString()).join(","));
```
nicholaspai, line 20: Leftover log

src/clients/SpokePoolClient.ts
```diff
@@ -2,13 +2,23 @@ import { spreadEvent, assign, Contract, BigNumber, EventSearchConfig } from "../
...
-import { Deposit, DepositWithBlock, Fill, SpeedUp, FillWithBlock } from "../interfaces/SpokePool";
+import {
+  Deposit,
+  DepositWithBlock,
+  Fill,
+  SpeedUp,
+  FillWithBlock,
+  RootBundleRelayWithBlock,
+  RelayerRefundExecutionWithBlock,
+} from "../interfaces/SpokePool";
```
nicholaspai, line 13: Nit try to flatten these into one line to keep style consistent

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
           [...expectedSlowRelayRoot.leaves],
           expectedSlowRelayRoot.tree.getHexRoot()
         ),
+      expectedTrees,
     };
   }
 
+  // TODO: this method and executeRelayerRefundLeaves have a lot of similarities, but they have some key differences
+  // in both the events they search for and the comparisons they make. We should try to generalize this in the future,
+  // but keeping them separate is probably the simplest for the initial implementation.
```
nicholaspai, line 813: Agreed can refactor later

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
-    // with matching roots.
+    const spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
+      this.chainIdListForBundleEvaluationBlockNumbers,
+      this.clients,
+      this.logger,
+      this.clients.hubPoolClient.latestBlockNumber
+    );
+
+    Object.entries(spokePoolClients).forEach(([chainId, client]) => {
+      let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays());
```
nicholaspai, line 823: Should these getters sort stuff or do you like having the sorting in the caller client?

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
+
+      // Only grab the most recent n roots that have been sent if configured to do so.
+      if (this.spokeRootsLookbackCount !== 0)
```
nicholaspai, line 826: Would it be doable to be able to specify either all the root bundle IDs you want to execute (or ignore and default to execute everything)?

That might be more useful to us but i think the roots look back is better than nothing

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
+            slowRelayRoot: rootBundleRelay.slowRelayRoot,
+          });
+          continue;
+        }
+
+        const prevRootBundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(
+          matchingRootBundle.blockNumber
+        );
+
+        const blockNumberRanges = matchingRootBundle.bundleEvaluationBlockNumbers.map((endBlock, i) => {
```
nicholaspai, line 862: Did you look at any of the hubpool client helpers related to searching for blocks in the bundle eval numbers? I thought I wrote a function that could help here

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
+            this.clients.hubPoolClient.latestBlockNumber;
+
+          const leaves = this.clients.hubPoolClient.getExecutedLeavesForRootBundle(
+            bundle,
+            this.clients.hubPoolClient.latestBlockNumber,
+            followingBlockNumber
+          );
+
+          // Only use this bundle if it had valid leaves returned (meaning it was at least partially executed).
+          return leaves.length > 0;
```
nicholaspai, line 845: Can you not use hubPoolClient.isRootBundleValid ? It may not be on master but it's in my outstanding PR and super useful!

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
+    if (!valid) {
+      this.logger.error({
+        at: "Dataworke#executePoolRebalanceLeaves",
+        message: "Found invalid proposal after challenge period!",
+        reason,
+      });
+      return;
+    }
+
+    if (valid && !expectedTrees) {
```
nicholaspai, line 964: How would you get to this edge case..?

src/dataworker/Dataworker.ts
```diff
@@ -763,21 +804,295 @@ export class Dataworker {
...
+        ],
+        message: "Executed PoolRebalanceLeaf 🌿!",
+        mrkdwn: `Root hash: ${expectedTrees.poolRebalanceTree.tree.getHexRoot()}\nLeaf: ${leaf.leafId}`, // Just a placeholder
+      });
+    });
   }
 
   async executeRelayerRefundLeaves() {
-    // TODO:
+    const spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
```
nicholaspai, line 1010: You probably realize this but this function can easily be refactored to reuse
Methods from execute slow leaves

src/dataworker/index.ts
```diff
@@ -40,6 +40,13 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
 
       await dataworker.proposeRootBundle();
 
+      await dataworker.executePoolRebalanceLeaves();
```
nicholaspai, line 43: Should we execute before we propose? Would this give us a chance to execute and then propose in a single server less iteration ? This might actually be problematic with how bundle block ranges are used but curious your thoughts.

nicholaspai, line 43: Should we execute before we propose? Would this give us a chance to execute and then propose in a single server less iteration ? This might actually be problematic with how bundle block ranges are used but curious your thoughts.

nicholaspai (approved):
Just reviewed all the implementation logic but not the tests. I agree with this design, just have tips on refactoring mostly! Will review tests on follow up. 

Also consider rebasing off my outstanding PR which includes some useful utility functions you could use and will reduce any merge conflicgs

src/dataworker/Dataworker.ts
mrice32: In the other code, it seemed the pattern was sorting outside, but maybe I misunderstood! Either way is fine with me, although I could imagine users needed both directions of sorts (I feel like descending is kinda uncommon?).

Thoughts?

src/dataworker/Dataworker.ts
mrice32: This might have been an old implementation, but I thought this was possible if the end block was in the future by a little. Basically the implementation doesn't want to dispute, but there is no valid tree to produce for this block range yet? You may have eliminated this in your most recent code, though.

src/dataworker/index.ts
mrice32: We could! I don't think this will make a difference in the current implementation, though, because we can't take into account the execution _before_ it's performed on-chain, right?

Like the proposer will think that the roots aren't executed until _after_ the execution transaction is submitted. I think we'd need special logic to tell the proposer that the execution will happen just before it in the multicall, so it should assume that the execution has already happened.

mrice32: I think this is a duplicate comment (github has been doing this a lot lately :/)

src/clients/MultiCallerClient.ts
mrice32: Ah nice. Will remove.

src/dataworker/Dataworker.ts
mrice32: Do you think we should just merge this as is and then integrate the two implementations after they're both merged?

src/dataworker/Dataworker.ts
mrice32: I looked through them, but it wasn't super obvious to me what the assumptions were on them (for instance, there was one that implied it was going to get the "widest" block range). Open to using a different method if one does exactly what I'm doing here with no special, state-based logic.

src/dataworker/Dataworker.ts
mrice32: Yep, I think that makes sense! I think a lookback is good for automatic execution, but an alternative way to manually specify roots would be good for an external script. Is that how you were thinking about it? I could add a script to do that in a follow-up, and I could refactor this for that use case then. Does that work?

src/clients/SpokePoolClient.ts
mrice32: kk will do

src/dataworker/Dataworker.ts
mrice32: Oh? Which methods?

src/clients/SpokePoolClient.ts
mrice32: Done.

src/clients/MultiCallerClient.ts
mrice32: Done.

PR #57, opened 2022-05-23
feat: add slow relay logic

chrismaree:
The relayer needs the ability to send 0 sized fills if a relay is not totally fillable. This PR adds the logic to do this.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -93,12 +93,12 @@ export class SpokePoolClient {
     );
   }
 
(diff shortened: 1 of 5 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -109,7 +109,7 @@ export class SpokePoolClient {
         : 0
     );
     const lastFill = fillsOrderedByTotalFilledAmount[0];
```
nicholaspai, line 111: When there are 0 fills, won't `lastFill` be undefined?

nicholaspai (approved):
This LGTM, I'm going to add a test to the dataworker that zero fills are included in the slow root

PR #60, opened 2022-05-23
improve: Fix duplicate slow fill bug and add USD threshold to submit root bundles

nicholaspai:
This PR adds various `DataworkerConfig` variables that caller can use to toggle which functions are called in dataworker. Additionally adds a "usdThreshold" value that the caller can set to pause submitting root bundles unless their `netSendAmount + runningBalances` sum exceeds a value in $$.

Finally fixes a bug where the "excess" running balance from slow fills is counted even if the fill that completed the deposit was not in the bundle block range.

### DIFF
--- a/src/dataworker/DataworkerConfig.ts
+++ b/src/dataworker/DataworkerConfig.ts
@@ -7,6 +7,8 @@ export class DataworkerConfig extends CommonConfig {
   readonly tokenTransferThresholdOverride: { [l1TokenAddress: string]: BigNumber };
   readonly blockRangeEndBlockBuffer: { [chainId: number]: number };
   readonly rootBundleExecutionThreshold: BigNumber;
+  readonly disputerEnabled: boolean;
+  readonly proposerEnabled: boolean;
 
   constructor(env: ProcessEnv) {
     const {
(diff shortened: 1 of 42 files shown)


src/dataworker/DataworkerConfig.ts
```diff
@@ -30,9 +40,16 @@ export class DataworkerConfig extends CommonConfig {
     this.tokenTransferThresholdOverride = TOKEN_TRANSFER_THRESHOLD_OVERRIDE
       ? JSON.parse(TOKEN_TRANSFER_THRESHOLD_OVERRIDE)
       : {};
+    this.rootBundleExecutionThreshold = ROOT_BUNDLE_EXECUTION_THRESHOLD
+      ? toBNWei(ROOT_BUNDLE_EXECUTION_THRESHOLD)
+      : toBNWei("500000");
```
nicholaspai, line 45: currently set to $500,000, but can change obivously

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -141,14 +143,31 @@ export function addSlowFillsToRunningBalances(
...
 export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
   endBlockForMainnet: number,
   runningBalances: interfaces.RunningBalances,
   hubPoolClient: HubPoolClient,
   allValidFills: interfaces.FillWithBlock[],
+  allValidFillsInRange: interfaces.FillWithBlock[],
   chainIdListForBundleEvaluationBlockNumbers: number[]
 ) {
-  allValidFills.forEach((fill: interfaces.FillWithBlock) => {
+  allValidFillsInRange.forEach((fill: interfaces.FillWithBlock) => {
```
nicholaspai, line 170: This was the subtle bug: we were previously evaluating ALL fills even those outside of range for whether they created a slow fill excess amount, but we should only check fills in range. We still need `allValidFills` to see if a fill created excess, but we shouldn't proceed to use `allValidFills` unless the fill in question is in range

chrismaree (approved):

PR #62, opened 2022-05-23
fix: concurrency

chrismaree:
add better concurrency settings


nicholaspai (approved):
Hope this works! lgtm

PR #64, opened 2022-05-23
fix: improve serverless crash

chrismaree:


nicholaspai (approved):
Seems reasonable and likely to work as a temp solution

PR #65, opened 2022-05-24
improve: Refactor leaf executor implementation

nicholaspai:
I used this PR to execute leaves in prod (and propose a new bundle). It cuts down the number of spoke pool clients from 5 to 2

1 For the pending root bundle, used in validation and execution logic
1 For the next root bundle, used in propose logic

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -706,14 +706,7 @@ export class Dataworker {
   // TODO: this method and executeRelayerRefundLeaves have a lot of similarities, but they have some key differences
   // in both the events they search for and the comparisons they make. We should try to generalize this in the future,
   // but keeping them separate is probably the simplest for the initial implementation.
-  async executeSlowRelayLeaves() {
-    const spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
-      this.chainIdListForBundleEvaluationBlockNumbers,
-      this.clients,
-      this.logger,
-      this.clients.hubPoolClient.latestBlockNumber
-    );
-
+  async executeSlowRelayLeaves(spokePoolClients: { [chainId: number]: SpokePoolClient }) {
     Object.entries(spokePoolClients).forEach(([chainId, client]) => {
       let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays());
 
@@ -910,14 +903,7 @@ export class Dataworker {
     });
   }
 
-  async executeRelayerRefundLeaves() {
-    const spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
-      this.chainIdListForBundleEvaluationBlockNumbers,
-      this.clients,
-      this.logger,
-      this.clients.hubPoolClient.latestBlockNumber
-    );
-
+  async executeRelayerRefundLeaves(spokePoolClients: { [chainId: number]: SpokePoolClient }) {
     Object.entries(spokePoolClients).forEach(([chainId, client]) => {
       let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays());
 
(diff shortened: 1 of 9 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -724,16 +724,10 @@ export class Dataworker {
...
-            this.clients.hubPoolClient.latestBlockNumber;
-
-          const leaves = this.clients.hubPoolClient.getExecutedLeavesForRootBundle(bundle, followingBlockNumber);
-
-          // Only use this bundle if it had valid leaves returned (meaning it was at least partially executed).
-          return leaves.length > 0;
+          // TODO: Consider allowing dataworker to execute relayer refund leaves from partially executed root bundles,
+          // i.e. imagine the situation where the pool rebalance leaf with a specific chain ID was executed
+          // and we want to execute its relayer refund leaves without waiting for the other chains.
+          return this.clients.hubPoolClient.isRootBundleValid(bundle, this.clients.hubPoolClient.latestBlockNumber)
```
nicholaspai, line 730: The tradeoff with using this existing helper method is that L2 leaves can't be executed until ALL L1 pool rebalance leaves are executed

src/dataworker/Dataworker.ts
```diff
@@ -855,7 +849,7 @@ export class Dataworker {
     const widestPossibleExpectedBlockRange = await PoolRebalanceUtils.getWidestPossibleExpectedBlockRange(
       this.chainIdListForBundleEvaluationBlockNumbers,
       this.clients,
-      this.clients.hubPoolClient.latestBlockNumber
+      pendingRootBundle.proposalBlockNumber
```
nicholaspai, line 852: I think using `latestBlockNumber` isn't correct because it means that the dataworker will try to create the pool rebalance root including events that might have occurred after the `pendingRootBundle` was proposed. If so, then using the proposal block number makes more sense

src/dataworker/Dataworker.ts
```diff
@@ -706,14 +706,7 @@ export class Dataworker {
   // TODO: this method and executeRelayerRefundLeaves have a lot of similarities, but they have some key differences
   // in both the events they search for and the comparisons they make. We should try to generalize this in the future,
   // but keeping them separate is probably the simplest for the initial implementation.
-  async executeSlowRelayLeaves() {
-    const spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
```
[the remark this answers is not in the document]
nicholaspai: This will save web3 reqs

src/dataworker/Dataworker.ts
```diff
@@ -935,15 +906,10 @@ export class Dataworker {
...
-            this.clients.hubPoolClient.latestBlockNumber;
-
-          const leaves = this.clients.hubPoolClient.getExecutedLeavesForRootBundle(bundle, followingBlockNumber);
-
-          // Only use this bundle if it had valid leaves returned (meaning it was at least partially executed).
-          return leaves.length > 0;
+          // TODO: Consider allowing dataworker to execute relayer refund leaves from partially executed root bundles,
+          // i.e. imagine the situation where the pool rebalance leaf with a specific chain ID was executed
+          // and we want to execute its relayer refund leaves without waiting for the other chains.
+          return this.clients.hubPoolClient.isRootBundleValid(bundle, this.clients.hubPoolClient.latestBlockNumber);
```
mrice32: I feel like this removes functionality. Should we be making this change if it's making the data worker _less_ useful?

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
mrice32: 😍

PR #66, opened 2022-05-24
feat: add simple retry provider

mrice32:


src/utils/ProviderUtils.ts
```diff
@@ -1,6 +1,23 @@
 import { ethers, providers } from "ethers";
 
-const stallTimeout = 5000;
+const stallTimeout = 15 * 1000;
+
+function delay(s: number): Promise<void> {
```
chrismaree, line 5: import `delay` from `./` https://github.com/across-protocol/relayer-v2/blob/d392d051aedf197a221d17c3de0ffda0b89b0122/src/utils/index.ts#L6

chrismaree (approved):

PR #71, opened 2022-05-24
fix: formatting of percentages

chrismaree:
Percentages need to be scaled by 100.


src/relayer/Relayer.ts
```diff
@@ -61,6 +61,7 @@ export class Relayer {
       // Decrement tokens in token client used in the fill. This ensures that we dont try and fill more than we have.
       this.clients.tokenClient.decrementLocalBalance(deposit.destinationChainId, deposit.destinationToken, fillAmount);
     } catch (error) {
+      console.log("error", error);
```
nicholaspai, line 64: Leftover log

PR #75, opened 2022-05-24
feat: Polygon finalizer

nicholaspai:
This is still just a script that can be run with `ts-node ./src/finalizer/index.ts --wallet mnemonic` which is janky and manual for now but its confirmed to work with polygon and at least prints out transaction hashes for arbitrum

### DIFF
--- a/src/finalizer/utils/arbitrum.ts
+++ b/src/finalizer/utils/arbitrum.ts
@@ -1,4 +1,11 @@
-import { getProvider, groupObjectCountsByThreeProps, Wallet, winston, convertFromWei } from "../../utils";
+import {
+  getProvider,
+  groupObjectCountsByThreeProps,
+  Wallet,
+  winston,
+  convertFromWei,
+  groupObjectCountsByTwoProps,
+} from "../../utils";
 import { L2ToL1MessageWriter, L2ToL1MessageStatus, L2TransactionReceipt, getL2Network } from "@arbitrum/sdk";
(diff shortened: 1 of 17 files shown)


src/finalizer/utils/arbitrum.ts
```diff
@@ -0,0 +1,165 @@
...
+  winston,
+  convertFromWei,
+  groupObjectCountsByTwoProps,
+} from "../../utils";
+import { L2ToL1MessageWriter, L2ToL1MessageStatus, L2TransactionReceipt, getL2Network } from "@arbitrum/sdk";
+import { MessageBatchProofInfo } from "@arbitrum/sdk/dist/lib/message/L2ToL1Message";
+import { TokensBridged } from "../../interfaces";
+import { HubPoolClient } from "../../clients";
+
+export async function finalizeArbitrum(
```
mrice32, line 14: Unclear how much of this was copied from v1 (hopefully most of it!). Might be worth commenting in the PR what sections of code are new.

mrice32 (approved):
Would be really nice to see multicall and simulation mode added here, but those may be tricky given how these SDKs work.

This looks good, and I think it's relatively low risk code.

The biggest concern I have is hardening it to make it less noisy than the Across V1 finalizers.

LGTM!

src/finalizer/utils/arbitrum.ts
nicholaspai: Its all new actually, we don't use updated SDK's in new code

nicholaspai:
@mrice32 we can't use multicall easily since the contracts we're calling are all different. We could go through a DS proxy...

PR #84, opened 2022-05-27
improve: Better logging and reducing web3 requests

nicholaspai:


src/dataworker/Dataworker.ts
```diff
@@ -726,7 +731,13 @@ export class Dataworker {
 
     await Promise.all(
       Object.entries(spokePoolClients).map(async ([chainId, client]) => {
-        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays());
+        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter(
+          (rootBundle) => rootBundle.slowRelayRoot !== EMPTY_MERKLE_ROOT
```
nicholaspai, line 735: We should ignore entirely any slow relay roots that are empty

src/dataworker/Dataworker.ts
```diff
@@ -783,15 +795,29 @@ export class Dataworker {
             continue;
           }
 
-          const unexecutedLeaves = leaves.filter((leaf) => {
-            if (leaf.destinationChainId !== Number(chainId)) return false;
+          const leavesForChain = leaves.filter((leaf) => leaf.destinationChainId === Number(chainId));
+          if (leavesForChain.length === 0) {
```
nicholaspai, line 799: We should exit early if the constructed root has no leaves that could be executed on this chain

src/dataworker/Dataworker.ts
```diff
@@ -783,15 +795,29 @@ export class Dataworker {
...
+          }
+          const unexecutedLeaves = leavesForChain.filter((leaf) => {
             const executedLeaf = slowFillsForChain.find(
               (event) => event.originChainId === leaf.originChainId && event.depositId === leaf.depositId
             );
 
             // Only return true if no leaf was found in the list of executed leaves.
             return !executedLeaf;
           });
+          if (unexecutedLeaves.length === 0) {
```
nicholaspai, line 814: Should exit early if all leaves have been executed

src/dataworker/Dataworker.ts
```diff
@@ -810,6 +836,22 @@ export class Dataworker {
 
             return { ...leaf, fill };
           });
+          const unexecutableLeaves = leavesWithLatestFills.filter(
+            (leaf) => leaf.fill && leaf.fill.totalFilledAmount.eq(leaf.fill.amount)
+          );
+          if (unexecutableLeaves.length > 0) {
```
nicholaspai, line 842: Should exit early if slow leaves can't be executed due to following fills

src/dataworker/Dataworker.ts
```diff
@@ -1028,7 +1075,13 @@ export class Dataworker {
 
     await Promise.all(
       Object.entries(spokePoolClients).map(async ([chainId, client]) => {
-        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays());
+        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter(
+          (rootBundle) => rootBundle.relayerRefundRoot !== EMPTY_MERKLE_ROOT
```
nicholaspai, line 1079: Again, ignore empty refund roots.

src/dataworker/Dataworker.ts
```diff
@@ -1097,11 +1151,17 @@ export class Dataworker {
             this.tokenTransferThreshold
           );
 
-          const { tree, leaves } = this.buildRelayerRefundRoot(
-            blockNumberRanges,
-            spokePoolClients,
+          const maxRefundCount = this.maxRefundCountOverride
+            ? this.maxRefundCountOverride
+            : this.clients.configStoreClient.getMaxRefundCountForRelayerRefundLeafForBlock(endBlockForMainnet);
+          const { tree, leaves } = _buildRelayerRefundRoot(
```
nicholaspai, line 1157: We should be calling the internal _buildRelayerRefundRoot to remove a _loadData call

src/dataworker/Dataworker.ts
```diff
@@ -1116,19 +1176,33 @@ export class Dataworker {
...
 
-          const executableLeaves = leaves.filter((leaf) => {
-            if (leaf.chainId !== Number(chainId)) return false;
+          const leavesForChain = leaves.filter((leaf) => leaf.chainId === Number(chainId));
+          if (leavesForChain.length === 0) {
+            this.logger.debug({
+              at: "Dataworke#executeRelayerRefundLeaves",
+              message: `No refund leaves in bundle ${rootBundleRelay.rootBundleId} with chain ${chainId}`,
+            });
+            continue;
```
nicholaspai, line 1185: Exit early if no leaves for chain

src/dataworker/Dataworker.ts
```diff
@@ -1116,19 +1176,33 @@ export class Dataworker {
...
+          }
+          const unexecutedLeaves = leavesForChain.filter((leaf) => {
             const executedLeaf = executedLeavesForChain.find(
               (event) => event.rootBundleId === rootBundleRelay.rootBundleId && event.leafId === leaf.leafId
             );
             // Only return true if no leaf was found in the list of executed leaves.
             return !executedLeaf;
           });
+          if (unexecutedLeaves.length === 0) {
```
nicholaspai, line 1194: Exit early if all executed

src/finalizer/index.ts
```diff
@@ -78,17 +77,28 @@ export async function run(
...
+          message: "No Optimism finalizable messages",
         });
       for (const message of finalizable) {
+        logger.debug({
+          at: "OptimismFinalizer",
+          message: "Finalizing Optimism message",
+          l1Token: message.token,
+        });
         await crossChainMessenger.finalizeMessage(message.message);
+        logger.info({
```
nicholaspai, line 96: I need to clean up this entire file but for now I wanted to distinguish between the different finalizers in the log message

src/finalizer/index.ts
```diff
@@ -103,8 +113,6 @@ export async function runFinalizer(_logger: winston.Logger) {
   const relayerClients = await constructRelayerClients(logger, config);
 
   try {
-    await updateRelayerClients(relayerClients);
```
nicholaspai, line 106: This was an extra update for no reason

src/relayer/RelayerClientHelper.ts
```diff
@@ -75,9 +75,9 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
 
 export async function updateRelayerClients(clients: RelayerClients) {
   await updateClients(clients);
-  // Profit and SpokePoolClient client requires up to date HubPoolClient and rateModelClient.
+  // SpokePoolClient client requires up to date HubPoolClient and ConfigStore client.
+  await updateSpokePoolClients(clients.spokePoolClients)
   // Token client requires up to date spokePool clients to fetch token routes.
-  await Promise.all([clients.profitClient.update(), updateSpokePoolClients(clients.spokePoolClients)]);
```
nicholaspai, line 80: This was an extra profitClient.update

src/scripts/validateRootBundle.ts
```diff
@@ -66,22 +59,8 @@ export async function validate(_logger: winston.Logger) {
     clients,
     priceRequestBlock
   );
-  logger.debug({
```
nicholaspai, line 69: Useless logs

src/dataworker/Dataworker.ts
mrice32: Doesn't the current implementation do the same thing? The first line in the filter function returns false for leaves not for this chain?

src/dataworker/Dataworker.ts
mrice32: Similar to the above, this seems unnecessary. The loops below don't do anything if the length is 0, so we shouldn't need this extra if right?

src/dataworker/Dataworker.ts
mrice32: Same as the above comments. This seems unnecessary. We already filter for this below on line 861. Why do we need an extra filter statement for it?

src/dataworker/Dataworker.ts
mrice32: Just OOC, why do we want to avoid loadData calls? It doesn't query infura, right? Both are synchronous, right?

Note: I'm cool doing it a different way -- just seems like it shouldn't make too much difference in performance, right, since neither is doing any queries?

src/dataworker/Dataworker.ts
mrice32: Similar to others, why do we need to do this? Shouldn't the unexecutedLeaves array just be empty since we filter for this already?

src/dataworker/Dataworker.ts
mrice32: Same as above.

mrice32 (commented):
A few questions, as this seems to complicate the execution logic for what seems like the same result. Is the goal just to log the _reason_ for no executions?

src/dataworker/Dataworker.ts
nicholaspai: Yeah good point, the array will just be 0

src/dataworker/Dataworker.ts
nicholaspai: will remove like above

src/dataworker/Dataworker.ts
nicholaspai: yeah well its just a way to re-use code better. I build the internal methods for this reason.

src/dataworker/Dataworker.ts
nicholaspai: removed

src/dataworker/Dataworker.ts
nicholaspai: removed

mrice32 (approved):
LGTM! This is a really nice noise improvement!

PR #86, opened 2022-05-29
improve: Allow caller to override repayment chain for l1 token

nicholaspai:


src/relayer/Relayer.ts
```diff
@@ -5,9 +5,17 @@ import { RelayerClients } from "./RelayerClientHelper";
 import { Deposit } from "../interfaces/SpokePool";
 
 export class Relayer {
-  private repaymentChainId = 1; // Set to 1 for now. In future can be dynamically set to adjust bots capital allocation.
+  private defaultRepaymentChain = 1;
+  constructor(
+    readonly logger: winston.Logger,
+    readonly clients: RelayerClients,
+    readonly repaymentChainIdForToken: { [l1Token: string]: number } = {}
```
mrice32, line 12: Is there a way to default the repayment chain to the chain where the relay was sent? I feel like that would be the most "common sense" setting for us and other relayers.

nicholaspai: would it?

nicholaspai: Setting that default would be easy though, implementation-wise

nicholaspai: done, waiting to hear what @chrismaree thinks about the usefuless of this PR

src/relayer/RelayerConfig.ts
```diff
@@ -8,26 +8,33 @@ export class RelayerConfig extends CommonConfig {
...
     this.inventoryConfig = RELAYER_INVENTORY_CONFIG ? JSON.parse(RELAYER_INVENTORY_CONFIG) : {};
+    this.repaymentChainIdForToken = REPAYMENT_CHAIN_FOR_TOKEN ? JSON.parse(REPAYMENT_CHAIN_FOR_TOKEN) : {};
 
     if (Object.keys(this.inventoryConfig).length > 0) {
       this.inventoryConfig = replaceAddressCase(this.inventoryConfig); // Cast any non-address case addresses.
       this.inventoryConfig.wrapEtherThreshold = this.inventoryConfig.wrapEtherThreshold
         ? toBNWei(this.inventoryConfig.wrapEtherThreshold)
         : toBNWei(1); // default to keeping 2 Eth on the target chains and wrapping the rest to WETH.
 
-      this;
```
nicholaspai, line 25: was this a typo?

mrice32 (commented):
Where does this fit in with the inventory management work?

It seems like some of this would be possible via inventory management by setting a target of 100% on a particular chain for each token?

Would still like to see a super easy way to get repaid on the destination chain, but maybe all the other cases are implicitly covered by the inventory management system? @chrismaree thoughts?

src/relayer/Relayer.ts
```diff
@@ -44,21 +53,37 @@ export class Relayer {
...
       if (!Object.keys(this.clients.spokePoolClients).includes(deposit.destinationChainId.toString()))
         throw new Error("Fatal error! Repayment chain set to a chain that is not part of the defined sets of chains!");
 
-      this.logger.debug({ at: "Relayer", message: "Filling deposit", deposit, repaymentChain });
       // Add the fill transaction to the multiCallerClient so it will be executed with the next batch.
       this.clients.multiCallerClient.enqueueTransaction({
         contract: this.clients.spokePoolClients[deposit.destinationChainId].spokePool, // target contract
         chainId: deposit.destinationChainId,
         method: "fillRelay", // method called.
-        args: buildFillRelayProps(deposit, repaymentChain, fillAmount), // props sent with function call.
```
kevinuma, line 59: Does this undo the inventory management change?

PR #87, opened 2022-05-30
improve: Refactor finalizer

nicholaspai:
Follow up PR to add lifecyle/unit tests

### DIFF
--- a/scripts/sendTokens.ts
+++ b/scripts/sendTokens.ts
@@ -39,7 +39,7 @@ export async function run(): Promise<void> {
   else {
     const erc20 = new ethers.Contract(token, ERC20.abi, connectedSigner);
     const decimals = Number(await erc20.decimals());
-    const symbol = Number(await erc20.symbol());
+    const symbol = await erc20.symbol();
     const amountFromWei = ethers.utils.formatUnits(args.amount, decimals);
     // Check the user is ok with the info provided. else abort.
     console.log(`Send ${symbol} with amount ${amountFromWei} tokens to ${recipient} on chain ${args.chainId}`);
--- a/scripts/sendTokens.ts
+++ b/scripts/sendTokens.ts
@@ -31,11 +31,11 @@ export async function run(): Promise<void> {
     console.log(`Send ETH with amount ${amountFromWei} tokens to ${recipient} on chain ${args.chainId}`);
     if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) process.exit(0);
     console.log("sending...");
-    const tx = await connectedSigner.sendTransaction({ to: recipient, value: toBN(args.amount) })
+    const tx = await connectedSigner.sendTransaction({ to: recipient, value: toBN(args.amount) });
     const receipt = await tx.wait();
-    console.log("Transaction hash:", receipt.transactionHash);  
+    console.log("Transaction hash:", receipt.transactionHash);
   }
-  // Send ERC20 
+  // Send ERC20
   else {
     const erc20 = new ethers.Contract(token, ERC20.abi, connectedSigner);
     const decimals = Number(await erc20.decimals());
@@ -47,7 +47,7 @@ export async function run(): Promise<void> {
     console.log("sending...");
     const tx = await erc20.transfer(recipient, args.amount);
     const receipt = await tx.wait();
-    console.log("Transaction hash:", receipt.transactionHash);  
+    console.log("Transaction hash:", receipt.transactionHash);
   }
 }
 
--- a/scripts/sendTokens.ts
+++ b/scripts/sendTokens.ts
@@ -1,6 +1,17 @@
-import { delay, ethers, getSigner, getProvider, ERC20 } from "../src/utils";
+import { ethers, getSigner, getProvider, ERC20 } from "../src/utils";
 import { askYesNoQuestion } from "./utils";
-const args = require("minimist")(process.argv.slice(2));
+const args = require("minimist")(process.argv.slice(2), {
+  string: ["token", "to", "amount"],
+  number: ["chainId"],
+});
+
+// Example run:
+// ts-node ./scripts/sendTokens.ts
+// \ --token 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
+// \ --amount 350000000000 --to 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
+// \ --chainId 1
+// \ --wallet gckms
+// \ --keys bot1
 
 export async function run(): Promise<void> {
   console.log("Executing Token sender 💸");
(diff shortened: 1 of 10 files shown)


scripts/sendTokens.ts
```diff
@@ -1,6 +1,17 @@
-import { delay, ethers, getSigner, getProvider, ERC20 } from "../src/utils";
+import { ethers, getSigner, getProvider, ERC20, ZERO_ADDRESS, toBN } from "../src/utils";
```
mrice32, line 1: Happy to review this now, but in the interest of smaller PRs, I would suggest putting things like this script into separate PRs, since this doesn't seem to be related to the finalizer or simulation mode.

src/clients/ConfigStoreClient.ts
```diff
@@ -45,6 +45,9 @@ export class AcrossConfigStoreClient {
     // There is one deposit on optimism for DAI that is right before the DAI rate model was added.
     if (quoteBlock === 14830339) quoteBlock = 14830390;
 
+    // Test SNX deposit was before the rate model update for SNX.
+    if (quoteBlock === 14856066) quoteBlock = 14856211;
```
mrice32, line 49: nit: this also seems unrelated to the intention of the PR.

src/finalizer/utils/arbitrum.ts
```diff
@@ -106,13 +119,14 @@ export async function getAllMessageStatuses(
...
-      token: tokensBridged[i].l2TokenAddress,
-      info: tokensBridged[i],
-    };
-  });
+  )
+    .filter((result) => result.message !== undefined)
+    .map((result, i) => {
+      return {
+        ...result,
+        info: tokensBridged[i],
```
mrice32, line 127: I'm not sure this will work, since the filter may change the indices of the result array, making it not match the indices of tokensBridged.

I would suggest augmenting the returned result object with the token info before doing the !== undefined filter so you don't need to look it up later.

src/finalizer/utils/arbitrum.ts
```diff
@@ -128,52 +142,60 @@ export async function getMessageOutboxStatusAndProof(
...
-      proofInfo: undefined,
-      status: L2ToL1MessageStatus[L2ToL1MessageStatus.EXECUTED],
+      proofInfo,
+      status: L2ToL1MessageStatus[outboxMessageExecutionStatus],
     };
-  }
-  const outboxMessageExecutionStatus = await l2Message.status(proofInfo);
-  if (outboxMessageExecutionStatus !== L2ToL1MessageStatus.CONFIRMED) {
+  } catch (err) {
+    // Likely L1 message hasn't been included in an arbitrum batch yet, so ignore it for now.
```
mrice32, line 194: Maybe we should put a debug log here just so we can see the error in the GCP logs?

src/finalizer/utils/arbitrum.ts
```diff
@@ -128,52 +142,60 @@ export async function getMessageOutboxStatusAndProof(
...
-      txnHash: event.transactionHash,
-      error,
-      notificationPath: "across-error",
-    });
-    throw error;
-  }
+  try {
+    // Get L2-to-L1 message objects contained in transaction. In principle, a single transaction could trigger
+    // any number of outgoing messages; the common case will be there's only one. In the context of Across V2,
+    // there should only ever be one.
```
mrice32, line 148: Are you sure? Couldn't we batch execute a bunch of RelayerRefund leaves for different tokens/roots, triggering multiple L2->L1 messages in the same L2 txn hash?

src/finalizer/utils/arbitrum.ts
```diff
@@ -128,52 +142,60 @@ export async function getMessageOutboxStatusAndProof(
...
+    const l2ToL1Messages = await l2Receipt.getL2ToL1Messages(l1Signer, await getL2Network(l2Provider));
+    if (l2ToL1Messages.length === 0 || l2ToL1Messages.length - 1 < logIndex) {
```
mrice32, line 150: Can you explain what logIndex is representing here? Based on my naive understanding, this logic implies that we every log in the transaction to be an L2 -> L1 message log. Is that the case? If so, how? Don't we produce other logs in our contracts in the L2 -> L1 transactions?

src/finalizer/utils/polygon.ts
```diff
@@ -1,8 +1,27 @@
...
+  toBN,
+  Wallet,
+  winston,
+} from "../../utils";
 import { TokensBridged } from "../../interfaces";
-import { HubPoolClient } from "../../clients";
+import { HubPoolClient, MultiCallerClient } from "../../clients";
+
+const CHAIN_ID = 137;
+const POLYGON_MESSAGE_SENT_EVENT_SIG = "0x8c5261668696ce22758910d05bab8f186d6eb247ceac2af2e82c7dc17669b036";
```
mrice32, line 19: I'm assuming this client only works for PoS tokens. Might be worth adding a quick comment to that effect.

src/finalizer/utils/polygon.ts
```diff
@@ -28,35 +47,45 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
   tokensBridged: TokensBridged[],
   posClient: POSClient,
   hubPoolClient: HubPoolClient
 ) {
   const isCheckpointed = await Promise.all(
     tokensBridged.map((event) => posClient.exitUtil.isCheckPointed(event.transactionHash))
   );
-  const withdrawExitedOrIsNotCheckpointed = await Promise.all(
+  const exitStatus = await Promise.all(
     tokensBridged.map((event, i) => {
```
mrice32, line 59: nit: why not make this an async method? That would mean you could drop all these `new Promise` statements below and just return the values directly.

src/finalizer/utils/polygon.ts
```diff
@@ -28,35 +47,45 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
       const l1TokenCounterpart = hubPoolClient.getL1TokenCounterpartAtBlock(
-        "137",
+        CHAIN_ID.toString(),
         event.l2TokenAddress,
         hubPoolClient.latestBlockNumber
       );
-      return posClient.erc20(l1TokenCounterpart, true).isWithdrawExited(event.transactionHash);
+      return posClient.erc20(l1TokenCounterpart, true).isWithdrawExited(event.transactionHash)
+        ? new Promise((resolve) => resolve({ status: POLYGON_MESSAGE_STATUS.ALREADY_EXITED }))
+        : new Promise((resolve) => resolve({ status: POLYGON_MESSAGE_STATUS.CAN_EXIT }));
```
mrice32, line 69: I assume this is an async method. If so, I think this is probably what you want:
```suggestion
      return posClient.erc20(l1TokenCounterpart, true).isWithdrawExited(event.transactionHash).then((result) => result ?
        { status: POLYGON_MESSAGE_STATUS.ALREADY_EXITED } :
        { status: POLYGON_MESSAGE_STATUS.CAN_EXIT });
```

mrice32 (commented):
`src/finalizer/utils/utils.ts`

Empty file?

scripts/sendTokens.ts
nicholaspai: Will do

src/finalizer/utils/arbitrum.ts
nicholaspai: ah good catch I think I should just add the filter below the map

src/clients/ConfigStoreClient.ts
nicholaspai: Yeah good point, still need this to run against the canary deployments so worth including

src/finalizer/utils/arbitrum.ts
nicholaspai: Ah yeah this comment is outdated and why we use `logIndex`

src/finalizer/utils/arbitrum.ts
nicholaspai: `logIndex` represnets the arbitrum relay index for the log. This line pulls all the bridge logs from the L2 transaction hash:

`const l2ToL1Messages = await l2Receipt.getL2ToL1Messages(l1Signer, await getL2Network(l2Provider))`

This is an arbitrum SDK method and should not capture any non bridge logs

src/finalizer/utils/polygon.ts
nicholaspai: great point!

nicholaspai:
I had to remove all the multicaller code as I realized that none of the L2 outboxes on l1 have multicaller logic

src/finalizer/utils/arbitrum.ts
mrice32: That works too!

src/finalizer/utils/arbitrum.ts
mrice32: Ah, I see. I was just confused by the naming, since `logIndex` is also used in a transaction receipt to denote the index of that log _within the transaction_.

mrice32 (approved):
LGTM!

PR #88, opened 2022-05-30
feat: relayer inventory management

chrismaree:
**Adds ability for relayer to dynamically manage its inventory.**

This is a high risk pr! this PR adds the ability for the relayer to directly interact with canonical bridges. It is posible that this can delete money in the bridge, if configured wrong of if there is a bug in this code.

**This PR adds the following to the bot:**
1. Relayer can now send any configured L1 token to the associated L2 chain over the canonical bridge.
2. Relayer can work out if it currently has a transfer in the canonical bridge. this is used to inform where the bot should allocate its funds
3. Relayer can consider all current allocation over all chains and compare this to a target distribution it can then use this data to:
3.1 Choose to send funds to the L2
3.2 Influence which relayer repayment chain the bot chooses.
4. There is a LOT of new logging added to give clarity on the above and enable debugging if something goes wrong.

The most critical bit of code in this PR is the logic that connects directly to the canonical bridge. I have extensively tested the 4 main routes that this PR supports over the associated canonical bridges: USDC, DAI, WETH and WBTC to Optimism, Polygon, Boba and Arbitrum. I also added a few safety hatches to the code that show explode if there is a misparamewterization when using the bridge.


src/clients/ConfigStoreClient.ts
```diff
@@ -48,6 +48,7 @@ export class AcrossConfigStoreClient {
     const rateModel = this.getRateModelForBlockNumber(l1Token, quoteBlock);
 
     // There is one deposit on optimism that is right at the margin of when liquidity was first added.
+    if (quoteBlock === 14830339) quoteBlock = 14830390;
```
nicholaspai, line 51: sorry about this one...

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,142 @@
...
+        tx = await this.bobaAdapter.sendTokenToTargetChain(l1Token, this.l2TokenForL1Token(l1Token, 288), amount);
+        break;
+      case 42161:
+        tx = await this.arbitrumAdapter.sendTokenToTargetChain(l1Token, this.l2TokenForL1Token(l1Token, 42161), amount);
+        break;
+    }
+    return await tx.wait();
+  }
+
+  async wrapEthIfAboveThreshold(wrapThreshold: BigNumber) {
```
nicholaspai, line 59: This function is only used for optimism and boba because they require you to wrap ETH before bridging, right?

src/clients/HubPoolClient.ts
```diff
@@ -104,10 +104,14 @@ export class HubPoolClient {
     return l1Token;
   }
 
-  getDestinationTokenForL1TokenDestinationChainId(l1Token: string, destinationChainId: number) {
+  getDestinationTokenForL1Token(l1Token: string, destinationChainId: number) {
     return this.l1TokensToDestinationTokens[l1Token][destinationChainId];
```
nicholaspai, line 108: why not `return this.l1TokensToDestinationTokens[l1Token] ? this.l1TokensToDestinationTokens[l1Token][destinationChainId] : undefined` and then we can get rid of the try-catch in the Adapter and check for undefined? They're both pretty ugly but maybe its more conssitent

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,142 @@
...
+    return this.spokePoolClients[chainId].spokePool.signer;
+  }
+  getChainSearchConfig(chainId: number) {
+    return this.spokePoolClients[chainId].eventSearchConfig;
+  }
+
+  l2TokenForL1Token(l1Token: string, chainId: number): string {
+    // the try catch below is a safety hatch. If you try fetch an L2 token that is not within the hubPoolClient for a
+    // given L1Token and chainId combo then you are likely trying to send a token to a chain that does not support it.
+    try {
```
nicholaspai, line 94: one nit suggestion, feel free to ignore: https://github.com/across-protocol/relayer-v2/pull/88/files#r884872665

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0x6c411aD3E74De3E7Bd422b94A27770f5B86C623B", // WETH
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0x467194771dae2967aef3ecbedd3bf9a310c76c65", // DAI
+  "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599": "0x09e9222e96e7b4ae2a407b98d48e330053351eee", // WBTC
+};
+
+// TODO: replace these numbers using the arbitrum SDK. these are bad values that mean we will over pay but transactions
+// wont get stuck. These are the same params we are using in the smart contracts.
+const l2GasPrice = toBN(5e9);
+const l2GasLimit = toBN(2000000);
+// abi.encodeing of the maxL2Submission cost. of 0.01e18
```
nicholaspai, line 30:
```suggestion
// abi.encoding of the maxL2Submission cost. of 0.01e18
```

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+    //todo: swap this to pulling spokePoolClient.latestBlockNumber.
+    const [l1BlockNumber, l2BlockNumber] = await Promise.all([
+      this.getProvider(1).getBlockNumber(),
+      this.getProvider(this.chainId).getBlockNumber(),
+    ]);
+
+    this.l1SearchConfig.toBlock = l1BlockNumber;
+    this.l2SearchConfig.toBlock = l2BlockNumber;
+
+    this.l2SearchConfig.fromBlock = this.l2SearchConfig.toBlock - this.depositEvalTime / this.avgBlockTime();
```
nicholaspai, line 53: what's the logic behind this math?

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  async checkAndSendTokenApprovals(l1Tokens: string[], associatedL1Bridges: string[]) {
+    this.log("Checking and sending token approvals", { l1Tokens, associatedL1Bridges });
+    const tokensToApprove: { l1Token: any; targetContract: string }[] = [];
+    const l1TokenContracts = l1Tokens.map((l1Token) => new Contract(l1Token, ERC20.abi, this.getSigner(1)));
+    const allowances = await Promise.all(
```
nicholaspai, line 60: Can you not use the token client to set allowances? Or add this to the token client?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,184 @@
...
+  getBalanceOnChainForL1Token(chainId: number, l1Token: string): BigNumber {
+    const balance =
+      this.tokenClient.getBalance(chainId, this.hubPoolClient.getDestinationTokenForL1Token(l1Token, chainId)) ||
+      toBN(0); // If the chain does not have this token (EG BOBA on Optimism) then 0.
+
+    // Consider any L1->L2 transfers that are currently pending in the canonical bridge.
+    return balance.add(this.outstandingCrossChainTransfers?.[chainId]?.[l1Token] || toBN(0));
+  }
+
+  getChainDistribution(l1Token: string): { [chainId: number]: BigNumber } {
```
nicholaspai, line 40: Is there a way for the caller to set a default/target distribution? I think that could be useful

nicholaspai (commented):
Awesome first pass, left some comments. Will hold off on longer review until you finish PR but this was very helpful personally for me to understand the structure

src/clients/ConfigStoreClient.ts
chrismaree: true. reverted.

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,149 @@
+import { BigNumber, winston, toWei, toBN, createFormatFunction, etherscanLink } from "../../utils";
+import { SpokePoolClient, HubPoolClient } from "../";
+import { OptimismAdapter, ArbitrumAdapter, PolygonAdapter } from "./";
+export class AdapterManager {
+  private optimismAdapter: OptimismAdapter;
+  private bobaAdapter: OptimismAdapter;
+  private arbitrumAdapter: ArbitrumAdapter;
+  private polygonAdapter: PolygonAdapter;
+
+  constructor(
```
chrismaree, line 10: this manager abstracts away the chain specific adapters from the `InventoryClient` to expose simple interfaces.

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+      const events = result.map((event) => {
+        const eventSpread = spreadEventWithBlockNumber(event);
+        return { amount: eventSpread[index % 2 === 0 ? "_amount" : "amount"], blockNumber: eventSpread["blockNumber"] };
+      });
+      const storageName = index % 2 === 0 ? "l1DepositInitiatedEvents" : "l2DepositFinalizedEvents";
+      assign(this[storageName], [l1Token], events);
+    });
+
+    this.l1SearchConfig.fromBlock = this.l1SearchConfig.toBlock + 1;
+    this.l2SearchConfig.fromBlock = this.l2SearchConfig.toBlock + 1;
```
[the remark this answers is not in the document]
chrismaree: nope. it's not. take a look at the variables being set. we track `l1SearchConfig` and `l2SearchCOnfig`

src/clients/InventoryClient/BaseAdapter.ts
chrismaree: I'll add more comments on this, sorry. had not commented all this logic just yet.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+      : rebalance;
+  }
+
+  // Work out where a relay should be refunded to optimally manage the bots inventory. Use the following algorithm:
+  // a) Compute the virtual balance o the destination chain. This is current liquidity + any pending cross-chain transfers.
+  // b) Compute the virtual balance post relay on destination chain. i.e a - the size of the relay in question.
+  // c) Find the post relay virtual allocation by taking b and deviling it by the cumulative virtual balance. This number
+  // represents the share of the total liquidity if the current transfers conclude and the relay is filled (i.e forward looking)
+  // e) Decide on what chain to repay on: If this number of more than the target for the designation chain + the rebalance overshoot // then refund on L1. Else, the post fill amount is within the target, so refund on the destination chain.
+  determineRefundChainId(deposit: Deposit): number {
```
chrismaree, line 115: this method is important.

src/relayer/Relayer.ts
```diff
@@ -39,23 +37,23 @@ export class Relayer {
...
-
     // If during the execution run we had shortfalls or unprofitable fills then handel it by producing associated logs.
     if (this.clients.tokenClient.anyCapturedShortFallFills()) this.handleTokenShortfall();
     if (this.clients.profitClient.anyCapturedUnprofitableFills()) this.handleUnprofitableFill();
   }
 
   fillRelay(deposit: Deposit, fillAmount: BigNumber) {
-    this.logger.debug({ at: "Relayer", message: "Filling deposit", deposit, repaymentChain: this.repaymentChainId });
     try {
+      const repaymentChain = this.clients.inventoryClient.determineRefundChainId(deposit);
```
chrismaree, line 47: see here: this is where the bot asks the inventory client where it should get refunded

contracts/AtomicWethDepositor.sol
```diff
@@ -0,0 +1,58 @@
...
+        uint32 _l2Gas,
+        bytes calldata _data
+    ) external payable;
+}
+
+interface PolygonL1Bridge {
+    function depositEtherFor(address _to) external payable;
+}
+
+/**
```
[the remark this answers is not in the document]
nicholaspai: Super useful thanks. Should we make it multicall?

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+          ...this.l2DepositFinalizedEvents_DepositAdapter[l1Token].filter((event) => event.to === this.relayerAddress),
+        ].sort((a, b) => a.blockNumber - b.blockNumber);
+
+        
+        const newestFinalizedAmount = l2FinalizationSet[l2FinalizationSet.length - 1];
+      const newestDeposit = this.l1DepositInitiatedEvents[l1Token][this.l1DepositInitiatedEvents[l1Token].length - 1];
+      if (
+        !newestFinalizedAmount &&
+        newestDeposit.blockNumber < this.l1SearchConfig.toBlock - this.depositEvalTime / this.avgBlockTime(1)
+      ) {
```
mrice32, line 109: Can you add a comment here? I may be being dense, but I don't quite understand this check.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+    this.tokenClient.decrementLocalBalance(1, l1Token, rebalance);
+    if (!this.outstandingCrossChainTransfers[chainId]) this.outstandingCrossChainTransfers[chainId] = {};
+    const bal = this.outstandingCrossChainTransfers[chainId][l1Token];
+    this.outstandingCrossChainTransfers[chainId][l1Token] = toBN(bal).gt(toBN(0))
+      ? toBN(bal).sub(rebalance)
+      : rebalance;
+  }
+
+  // Work out where a relay should be refunded to optimally manage the bots inventory. Use the following algorithm:
+  // a) Compute the virtual balance o the destination chain. This is current liquidity + any pending cross-chain transfers.
```
mrice32, line 110:
```suggestion
  // a) Compute the virtual balance on the destination chain. This is current liquidity + any pending cross-chain transfers.
```

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+  // b) Compute the virtual balance post relay on destination chain. i.e a - the size of the relay in question.
```
mrice32, line 111:
```suggestion
  // b) Compute the virtual balance post relay on destination chain. i.e the size of the relay in question.
```

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+  // c) Find the post relay virtual allocation by taking b and deviling it by the cumulative virtual balance. This number
```
mrice32, line 112:
```suggestion
  // c) Find the post relay virtual allocation by taking b and dividing it by the cumulative virtual balance. This number
```

src/clients/InventoryClient/InventoryClient.ts
mrice32, line 110: I think this is missing some info -- what about relays that have been sent, but have not been refunded, yet? Are those considered as part of the virtual balance?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+    const allocationThreshold = this.inventoryConfig.targetL2PctOfTotal[deposit.destinationChainId].add(
+      this.inventoryConfig.rebalanceOvershoot
+    );
+    this.log("Evaluated refund Chain", {
+      expectedPostRelayAllocation,
+      allocationThreshold,
+      cumulativeVirtualBalancePostRelay,
+    });
+    if (expectedPostRelayAllocation.gt(allocationThreshold)) return 1;
+    else return deposit.destinationChainId;
```
mrice32, line 135: So right now, we never try to use this mechanism to move to a _different_ L2 if it's low. We basically just choose between mainnet and the destination L2.

I think this is a good assumption!

A slight improvement might be that we look for the chain that would be the least above its allocation threshold after accepting this refund. Regardless, I think what you have makes sense -- this is definitely something for a follow-up!

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+          if (!possibleRebalances[chainId]) possibleRebalances[chainId] = {};
+          possibleRebalances[chainId][l1Token] = requiredRebalance;
+
+          // Decrement token balance in client for this chain and increment cross chain counter.
+          this.trackCrossChainTransfer(l1Token, requiredRebalance, chainId);
+        }
+      }
+    }
+
+    // Extract unexcusable rebalances for logging.
```
mrice32, line 181:
```suggestion
    // Extract unexecutable rebalances for logging.
```

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+      for (const l1Token of Object.keys(rebalancesRequired[chainId])) {
+        if (!possibleRebalances[chainId] || !possibleRebalances[chainId][l1Token]) {
+          if (!unexecutedRebalances[chainId]) unexecutedRebalances[chainId] = {};
+          unexecutedRebalances[chainId][l1Token] = rebalancesRequired[chainId][l1Token];
+        }
+      }
+    }
+
+    // Finally, execute the rebalances.
+    // TODO: The logic below is slow as it waits for each transaction to be included before sending the next one. This
```
mrice32, line 193: +1 -- this bot doesn't need to be fast. Good call on putting this off.

mrice32 (approved):
LGTM!

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.logger.debug({ at: "AdapterManager", message: "Getting outstandingCrossChainTransfers", chainId, l1Tokens });
+    switch (chainId) {
+      case 10:
+        return await this.optimismAdapter.getOutstandingCrossChainTransfers(l1Tokens);
+      case 137:
+        return await this.polygonAdapter.getOutstandingCrossChainTransfers(l1Tokens);
+      case 288:
+        return await this.bobaAdapter.getOutstandingCrossChainTransfers(l1Tokens);
+      case 42161:
+        return await this.arbitrumAdapter.getOutstandingCrossChainTransfers(l1Tokens);
```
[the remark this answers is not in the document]
chrismaree: good point. It was like this before when each chain had more custom logic before I had abstracted them away more. have refactored.

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,149 @@
...
+      return l2TokenForL1Token;
+    } catch (error) {
+      this.logger.error({
+        at: "AdapterManager",
+        message: "Implementor attempted to get a l2 token address for an L1 token that does not exist in the routings!",
+        l1Token,
+        chainId,
+        error,
+      });
+      throw error;
```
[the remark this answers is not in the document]
chrismaree: yes. if this error occurs here then this is REALLY bad. If this section here did not throw and the thing it's catching got through you could delete money. I added it here to tripple check before telling the adapters to send that we are not about to delete money.

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  getAdapterConstructors(
+    l2ChainId: number
+  ): [
+    l1Provider: any,
+    l2Provider: any,
+    l1Signer: any,
+    l2Signer: any,
+    initialL1SearchConfig: any,
+    initialL2SearchConfig: any
```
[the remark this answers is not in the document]
chrismaree: fixed.

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+const firstL1BlockToSearch = 12640865;
+
+const l2Gateways = {
+  "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": "0x096760F208390250649E3e8763348E783AEF5562", // USDC
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0x6c411aD3E74De3E7Bd422b94A27770f5B86C623B", // WETH
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0x467194771dae2967aef3ecbedd3bf9a310c76c65", // DAI
+  "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599": "0x09e9222e96e7b4ae2a407b98d48e330053351eee", // WBTC
+};
+
+// TODO: replace these numbers using the arbitrum SDK. these are bad values that mean we will over pay but transactions
```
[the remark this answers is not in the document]
chrismaree: correct

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+      const searchFilter = [undefined, this.relayerAddress];
+      promises.push(
+        paginatedEventQuery(l1Bridge, l1Bridge.filters.DepositInitiated(...searchFilter), this.l1SearchConfig),
+        paginatedEventQuery(l2Bridge, l2Bridge.filters.DepositFinalized(...searchFilter), this.l2SearchConfig)
+      );
+    }
+
+    const results = await Promise.all(promises, { concurrency: 4 });
+    results.forEach((result, index) => {
+      const l1Token = l1Tokens[Math.floor(index / 2)];
```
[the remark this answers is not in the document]
chrismaree: that's true! that shortcut should _never_ be hit. it's more of a filter to ensure we dont try get transfers for something we dont know the address of. I've changed it to be a throw to blow up in this case.

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    this.l1SearchConfig.toBlock = l1BlockNumber;
+    this.l2SearchConfig.toBlock = l2BlockNumber;
+
+    this.l2SearchConfig.fromBlock = this.l2SearchConfig.toBlock - this.depositEvalTime / this.avgBlockTime();
+  }
+
+  async checkAndSendTokenApprovals(l1Tokens: string[], associatedL1Bridges: string[]) {
+    this.log("Checking and sending token approvals", { l1Tokens, associatedL1Bridges });
+    const tokensToApprove: { l1Token: any; targetContract: string }[] = [];
```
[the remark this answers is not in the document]
chrismaree: it's a contract. I've updated the type accordingly.

src/clients/InventoryClient/InventoryClient.ts
chrismaree: I've updated this. it was WIP when you read it the first time.

src/clients/InventoryClient/InventoryClient.ts
chrismaree: ye, totally. right now we only ever send funds from L1->L2 and then make decisions on where we want to get refunded based on balances.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,334 @@
...
+    // If the chain does not have this token (EG BOBA on Optimism) then 0.
+    const balance =
+      this.tokenClient.getBalance(chainId, this.getDestinationTokenForL1Token(l1Token, chainId)) || toBN(0);
+
+    // Consider any L1->L2 transfers that are currently pending in the canonical bridge.
+    return balance.add(this.getOutstandingCrossChainTransferAmount(chainId, l1Token));
+  }
+
+  // Get any funds currently in the canonical bridge.
+  getOutstandingCrossChainTransferAmount(chainId: number | string, l1Token: string) {
```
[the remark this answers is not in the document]
chrismaree: it does not. this is not really relevant for this. it cares more about the current balances.

src/clients/HubPoolClient.ts
chrismaree: the reason I do the try-catch there is that I need it to fail very loudly if there is a problem with this in that contex. other implementors of this are ok with undefined but the inventory client can delete money if there is an issue with this param returned.

src/clients/HubPoolClient.ts
```diff
@@ -104,10 +104,14 @@ export class HubPoolClient {
...
   }
 
+  l2TokenEnabledForL1Token(l1Token: string, destinationChainId: number) {
+    return this.l1TokensToDestinationTokens[l1Token][destinationChainId] != undefined;
```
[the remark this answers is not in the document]
chrismaree: I have not, no.

src/clients/HubPoolClient.ts
```diff
@@ -350,8 +354,11 @@ export class HubPoolClient {
 
     // For each enabled Lp token fetch the token symbol and decimals from the token contract. Note this logic will
     // only run iff a new token has been enabled. Will only append iff the info is not there already.
+    // Filter out any duplicate addresses. This might happen due to enabling, disabling and re-enabling a token.
```
chrismaree, line 357: this fixes a bug I found in the profit client where some tokens were listed multiple times

src/clients/InventoryClient/ArbitrumAdapter.ts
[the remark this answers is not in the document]
chrismaree: note that the case you described in the bug was not really an issue given the configs but still great catch.

src/relayer/Relayer.ts
```diff
@@ -72,14 +68,15 @@ export class Relayer {
   }
 
   zeroFillDeposit(deposit: Deposit) {
-    this.logger.debug({ at: "Relayer", message: "Zero filling", deposit, repaymentChain: this.repaymentChainId });
+    const repaymentChainId = 1; // Always refund zero fills on L1 to not send dust over the chain unnecessarily.
```
chrismaree, line 71: note slow relays always refunded on L1 to avoid dust sends to L2s.

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+  ) {
+    if (spokePoolClients) {
+      this.adapters[10] = new OptimismAdapter(logger, spokePoolClients, relayerAddress, true);
+      this.adapters[137] = new PolygonAdapter(logger, spokePoolClients, relayerAddress);
+      this.adapters[288] = new OptimismAdapter(logger, spokePoolClients, relayerAddress, false);
+      this.adapters[42161] = new ArbitrumAdapter(logger, spokePoolClients, relayerAddress);
+    }
+  }
+
+  async getOutstandingCrossChainTokenTransferAmount(
```
chrismaree, line 21: ^

contracts/AtomicWethDepositor.sol
```diff
@@ -0,0 +1,58 @@
...
+ * @notice Contract deployed on Ethereum helps relay bots atomically unwrap and bridge WETH over the canonical chain
+ * bridges for Optimism, Boba and Polygon. Needed as these chains only support bridging of ETH, not WETH.
+ */
+
+contract AtomicWethDepositor {
+    Weth weth = Weth(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
```
nicholaspai, line 32: should we declare these iimmutable?

scripts/sendTokens.ts
```diff
@@ -45,7 +46,7 @@ export async function run(): Promise<void> {
     console.log(`Send ${symbol} with amount ${amountFromWei} tokens to ${recipient} on chain ${args.chainId}`);
     if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) process.exit(0);
     console.log("sending...");
-    const tx = await erc20.transfer(recipient, args.amount);
+    const tx = await erc20.transfer(recipient, args.amount, { gasLimit: 100000 });
```
nicholaspai, line 49: has this failed for you before? I've never run into `gasLimit` issues

src/clients/InventoryClient/AdapterManager.ts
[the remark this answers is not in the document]
nicholaspai: I would just add a comment here that this error should only happen due to programmer error, or in really bad situations so its ok to bring down the entire bot here

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  async update() {}
+}
+
+// The most critical failure mode that can happen in this module is a mis-mapping between L1 token and the associated
+// L2 token. If this is wrong the bot WILL delete money. The mapping below is used to enforce that what the hubpool
+// thinks is the correct L2 token for a given L1 token is actually the correct L2 token. It is simply a sanity check:
+// if for whatever reason this does not line up the bot show fail loudly and stop execution as something is broken
+// and funds are not safe to be sent over the canonical L2 bridges.
+export const l2TokensToL1TokenValidation = {
```
nicholaspai, line 113: nit: move to Constants and rename to `l2ToL1TokenMapping`

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+  "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
+    10: "0x7F5c764cBc14f9669B88837ca1490cCa17c31607",
+    137: "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
+    288: "0x66a2A913e447d6b4BF33EFbec43aAeF87890FBbc",
+    42161: "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8",
+  }, // USDC
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": {
+    10: "0x4200000000000000000000000000000000000006",
+    137: "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619",
+    288: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
```
nicholaspai, line 123: this is so.....silly

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+    this.log("Checking and sending token approvals", { l1Tokens, associatedL1Bridges });
+    const tokensToApprove: { l1Token: Contract; targetContract: string }[] = [];
+    const l1TokenContracts = l1Tokens.map((l1Token) => new Contract(l1Token, ERC20.abi, this.getSigner(1)));
+    const allowances = await Promise.all(
+      l1TokenContracts.map((l1TokenContract, index) => {
+        // If there is not both a l1TokenContract and associatedL1Bridges[index] then return a number that wont send
+        // an approval transaction. For example not every chain has a bridge contract for every token. In this case
+        // we clearly dont want to send any approval transactions.
+        if (l1TokenContract && associatedL1Bridges[index])
+          return l1TokenContract.allowance(this.relayerAddress, associatedL1Bridges[index]);
```
nicholaspai, line 67: Is this going to add multiple web3 requests per relayer run? I thought we could re-use the hub client somehow

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+    for (const l1Token of l1Tokens) {
+      let l2FinalizationSet = this.l2DepositFinalizedEvents[l1Token];
+
+      if (this.isWeth(l1Token) && this.l2DepositFinalizedEvents_DepositAdapter?.[l1Token]?.length > 0)
+        // If this is WETH and there are atomic depositor events then consider the union as the ful set of finalization
+        // events. We do this as the output event on L2 will show the Atomic depositor as the sender, not the relayer.
+        l2FinalizationSet = [
+          ...l2FinalizationSet,
+          ...this.l2DepositFinalizedEvents_DepositAdapter[l1Token].filter((event) => event.to === this.relayerAddress),
+        ].sort((a, b) => a.blockNumber - b.blockNumber);
```
nicholaspai, line 105: nit: can probably use `sortEventsAscending` here along with `spreadEventWithBlockNumber`

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  isWeth(l1Token: string) {
+    return l1Token == "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+  }
+
+  // Define some very rough heuristics for the average block time per chain.
+  avgBlockTime(chainId: number = this.chainId) {
+    if (chainId == 1) return 14; // 1 block every 13.5 seconds.
+    if (chainId == 10) return 0.5; // 2 blocks every second.
```
nicholaspai, line 186: It seems like its safer to overestimate block time here, so I think for Optimism  and Arbitrum you can set these to 10 blocks per second, polygon seems stable at 2 but maybe overestimate to 1, mainnet can move down to 13, and boba seems fine

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+      }
+
+      // If, however, there is either a newestFinalizedDeposit or the most recent L1 deposit time is less than 1 day
+      // we can find how the current pending transactions! To do this we need to find the most recent L1 deposit event
+      // that overlaps with the most recent L2 finalization event. Not all bridges emit a deposit identifier or some
+      // unique way to map inputs to outputs. So the most general logic is to map the input and output amounts. Identify
+      // The associated L1 Deposit event index that maps to the newest finalization event. Note that if there is none
+      // this will return -1 i.e there has been no finalization event in the look back period but there is a deposit
+      // this would be the case if this was the first deposit in ~1day*l2LooBackSafetyMargin blocks on the L2.
+      // NOTE there is an edge case here where you have sent the same amount of funds over the bridge multiple times
```
nicholaspai, line 136: This edge case will be annoying but as you say not the end of the world, you think its worth deep copying the `l2FinalizationSet` list and deleting anyones that match in the following branch? Could also take advantage of class state variables to store already matched txns

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  // Work out where a relay should be refunded to optimally manage the bots inventory. If the inventory management logic
+  // not enabled then return funds on the chain the deposit was filled on Else, use the following algorithm:
+  // a) Find the chain virtual balance (current balance + pending relays) minus the current shortfall.
+  // b) find the cumulative virtual balance minus the current shortfall.
+  // c) consider the size of a and b post relay (i.e after the relay is paid and all current transfers are settled what
+  // will the balances be on the target chain and the overall cumulative balance).
+  // d) use c to compute what the post relay post current in-flight transactions allocation would be. Compare this
+  // number to the target threshold and:
+  //     If this number of more than the target for the designation chain + rebalance overshoot then refund on L1.
```
nicholaspai, line 117: What if there is a shortfall on the designation chain, but ALL of the rebalances would overshoot. This would mean that the shortfall is never covered right? I think this is OK just to keep things simple, but want to confirm my understanding

src/clients/InventoryClient/OptimismAdapter.ts
```diff
@@ -0,0 +1,144 @@
...
+import { BaseAdapter, weth9Abi, ovmL1BridgeInterface, ovmL2BridgeInterface, atomicDepositorInterface } from "./";
+
+const customL1OptimismBridgeAddresses = {
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0x10e6593cdda8c58a1d0f14c5164b376352a55f2f", // DAI
+};
+
+const l1StandardBridgeAddressOvm = "0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1";
+const l1StandardBridgeAddressBoba = "0xdc1664458d2f0B6090bEa60A8793A4E66c2F1c00";
+
+const firstL1BlockOvm = 13352477;
```
nicholaspai, line 13: Can't you grab this from SpokePool deployment block?

src/relayer/Relayer.ts
```diff
@@ -5,57 +5,56 @@ import { RelayerClients } from "./RelayerClientHelper";
 import { Deposit } from "../interfaces/SpokePool";
 
 export class Relayer {
-  private repaymentChainId = 1; // Set to 1 for now. In future can be dynamically set to adjust bots capital allocation.
-
   constructor(readonly logger: winston.Logger, readonly clients: RelayerClients) {}
-  async checkForUnfilledDepositsAndFill() {
+  async checkForUnfilledDepositsAndFill(sendSlowRelays: Boolean = true) {
```
nicholaspai, line 9: +1

src/relayer/Relayer.ts
nicholaspai: Do you think we need to check here that `repaymentChain` is set to one of the `this.spokePoolChains` just to be safe that we don't burn money with a bad chain id?

nicholaspai (approved):
This is a crazy amount of code, good work. Because the diff is so large its a bit hard to reason about but I recommend we add this to the relayer and run it in simulation mode until we have confidence to turn it on. Perhaps we can set the `RELAYER_INVENTORY_CONFIG. targetL2PctOfTotal` low enough to minimize the amount of funds automatically sent over the bridge.

src/clients/InventoryClient/AdapterManager.ts
chrismaree: done.

src/clients/InventoryClient/BaseAdapter.ts
chrismaree: the allowances are different, unfortunately. this is looking for L1 allowances against the L1 token bridge. the token hub client or token client does not worry about this.

src/clients/InventoryClient/BaseAdapter.ts
chrismaree: it could...I thought it better to leave it hear as this is re-used across the 4 clients and it's pretty custom to just the inventory logic. we could pull it out there and create a more common place allowance checking block but I think that should be left for a later PR as this would re-factor 3 diffrent bits of code to accommodate this.

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+  getName() {
+    return `${getNetworkName(this.chainId)}Adapter`;
+  }
+
+  isWeth(l1Token: string) {
+    return l1Token == "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+  }
+
+  // Define some very rough heuristics for the average block time per chain.
+  avgBlockTime(chainId: number = this.chainId) {
```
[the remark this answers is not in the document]
chrismaree: ye...we could. I think that should refactor that constant to be dependent on the block time constant. ideally they are both dependent on each other. I dont want to do that now though due to the size of this PR.

src/clients/InventoryClient/BaseAdapter.ts
chrismaree: done.

src/clients/InventoryClient/OptimismAdapter.ts
chrismaree: no, this is not the first deployment of the spoke pool. this is the first deployment of the _OVM_ standard bridge.

src/clients/InventoryClient/InventoryClient.ts
chrismaree: distribution means for a given set of tokens, what fraction is spread over what chains. what you are describing (target distribution) is configurable and is set in the `RelayerConfig`. (see `InventoryConfig`.)

src/clients/InventoryClient/InventoryClient.ts
chrismaree: so I think the situation you describe would happen if the target allocation is set to 0% and there is a shortfall. what will happen then is bot would send funds to the target chain to cover the shortfall + overshoot. then, any subsequent relay would be refunded on L1 if the balance post the relay was 0% plus overshoot. what this means is the chain will be left with overshoot balance on it, until all of it is used OR until additional funds are sent to that chain.

src/common/Config.ts
```diff
@@ -36,7 +36,7 @@ export class CommonConfig {
     else this.maxBlockLookBack = Constants.CHAIN_MAX_BLOCK_LOOKBACK;
     this.nodeQuorumThreshold = NODE_QUORUM_THRESHOLD ? Number(NODE_QUORUM_THRESHOLD) : 1;
     this.maxTxWait = MAX_TX_WAIT_DURATION ? Number(MAX_TX_WAIT_DURATION) : 180; // 3 minutes
-    this.relayerDiscount = RELAYER_DISCOUNT ? toBNWei(RELAYER_DISCOUNT) : toBNWei(0);
+    this.sendingTransactionsEnabled = SEND_TRANSACTIONS === "true";
```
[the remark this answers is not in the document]
chrismaree: this is not merged in yet, right? so assuming my code comes in first you can rebase off of this and change it accordingly.

src/relayer/Relayer.ts
chrismaree: done.

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0x6c411aD3E74De3E7Bd422b94A27770f5B86C623B", // WETH
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0x467194771dae2967aef3ecbedd3bf9a310c76c65", // DAI
+  "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599": "0x09e9222e96e7b4ae2a407b98d48e330053351eee", // WBTC
+};
+
+// TODO: replace these numbers using the arbitrum SDK. these are bad values that mean we will over pay but transactions
+// wont get stuck. These are the same params we are using in the smart contracts.
+
+export class ArbitrumAdapter extends BaseAdapter {
+  l2GasPrice = toBN(5e9);
```
kevinuma, line 29: Can these gas price and gas limit be done per-network in hardhat config instead?

src/clients/InventoryClient/BaseAdapter.ts
```diff
@@ -0,0 +1,191 @@
...
+      const receipt = await tx.wait();
+      mrkdwn +=
+        ` - Approved Canonical ${getNetworkName(this.chainId)} token bridge ${etherscanLink(targetContract, 1)} ` +
+        `to spend ${await l1Token.symbol()} ${etherscanLink(l1Token.address, 1)} on ${getNetworkName(1)}. ` +
+        `tx: ${etherscanLink(receipt.transactionHash, 1)}\n`;
+    }
+    this.log("Approved whitelisted tokens! 💰", { mrkdwn }, "info");
+  }
+
+  computeOutstandingCrossChainTransfers(l1Tokens: string[]): { [l1Token: string]: BigNumber } {
```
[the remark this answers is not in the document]
kevinuma: I actually was thinking of a simpler strategy than this:

1. Assume that the finalizer would act instantly the moment transfers can be finalized.
2. Look at the timestamp of transfer events from L2s and check if those timestamps have passed the 7-day period. If so, we can consider these transfers already finalized. Otherwise, they're pending.

This can cut down the complexity of checking transfers against fills, but at the potential cost of inaccuracy if the finalizer is delayed for some reason. What do you guys think?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,367 @@
...
+  //     Else, the post fill amount is within the target, so refund on the destination chain.
+  determineRefundChainId(deposit: Deposit): number {
+    if (!this.isfInventoryManagementEnabled()) return deposit.destinationChainId;
```
kevinuma, line 120: nit: I think there's an extra "f" after "is"

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,367 @@
...
+    });
+
+    this.log("Updated inventory information", { outstandingCrossChainTransfers: this.outstandingCrossChainTransfers });
+  }
+
+  isfInventoryManagementEnabled() {
+    if (this.inventoryConfig.managedL1Tokens) return true;
+    // Use logDisabledManagement to avoid spamming the logs on every check if this module is enabled.
+    else if ((this.logDisabledManagement = false)) this.log("Inventory Management Disabled");
+    this.logDisabledManagement = true;
```
kevinuma, line 360: Why change this to true when to true immediately after setting it to false in the else if branch?

test/InventoryClient.InventoryRebalance.ts
```diff
@@ -0,0 +1,231 @@
...
+    // The token shortfall should now no longer be an issue. This means we can fill the relay of 18 size now.
+    tokenClient.setTokenShortFallData(137, l2TokensForWeth[137], [6969], toBN(0));
+    tokenClient.decrementLocalBalance(137, l2TokensForWeth[137], shortfallAmount); // mock the relay actually filling.
+
+    await inventoryClient.update();
+    await inventoryClient.rebalanceInventoryIfNeeded();
+    expect(lastSpyLogIncludes(spy, `No rebalances required`)).to.be.true;
+    // We should see a log for chain 42161 that shows the actual balance after the relay concluded and the share.
+    // actual balance should be listed above at 945. share should be 945/(13500) =0.7 (initial total - withdrawAmount).
+    // expect(spyLogIncludes(spy, -2, `"42161":{"actualBalanceOnChain":"945.00"`)).to.be.true;
```
kevinuma, line 215: I wrote some code in https://github.com/across-protocol/relayer-v2/pull/108/files#diff-9434c6830a85f3e43e62881beb8cedff3bf7ccb82817bba40c17efee601c4820R64 to scan through the logs and assert there's at least one log with a matching message. Seems like that'd also be useful here?

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+        l1Token,
+        chainId,
+        error,
+      });
+      throw error;
+    }
+  }
+
+  async setL1TokenApprovals(l1Tokens: string[]) {
+    // Each of these calls must happen sequentially or we'll have collisions within the TransactionUtil. This should
```
kevinuma, line 91: Does this also send out approval transactions even if the tokens have already been approved before?

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  async sendTokenCrossChain(chainId: number | string, l1Token: string, amount: BigNumber) {
+    this.logger.debug({ at: "AdapterManager", message: "Sending token cross-chain", chainId, l1Token, amount });
+    const l2Token = this.l2TokenForL1Token(l1Token, Number(chainId));
+    const tx = await this.adapters[chainId].sendTokenToTargetChain(l1Token, l2Token, amount);
+    return await tx.wait();
+  }
+
+  async wrapEthIfAboveThreshold(wrapThreshold: BigNumber) {
+    const [optimismWrapTx, bobaWrapTx] = await Promise.all([
```
kevinuma, line 37: Is this something that's only necessary for Optimism and Boba but not Arbitrum + Polygon?

src/clients/InventoryClient/OptimismAdapter.ts
```diff
@@ -0,0 +1,144 @@
...
+const firstL1BlockBoba = 13012048;
+
+const ovmL2StandardBridgeAddress = "0x4200000000000000000000000000000000000010";
+const customOvmBridgeAddresses = {
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0x467194771dae2967aef3ecbedd3bf9a310c76c65", // DAI
+};
+
+const wethOptimismAddress = "0x4200000000000000000000000000000000000006";
```
kevinuma, line 21: These addresses seem redundant with the config in AdapterManager. Can we merge them somehow?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,367 @@
...
+      this.constructConsideringRebalanceDebugLog(tokenDistributionPerL1Token);
+
+      // First, compute the rebalances that we would do assuming we have sufficient tokens on L1.
+      for (const l1Token of Object.keys(tokenDistributionPerL1Token)) {
+        const cumulativeBalance = this.getCumulativeBalance(l1Token);
+        for (const chainId of this.getEnabledL2Chains()) {
+          const currentAllocationPct = this.getCurrentAllocationPctConsideringShortfall(l1Token, chainId);
+          const targetAllocationPct = toBN(this.inventoryConfig.targetL2PctOfTotal[chainId]);
+          if (currentAllocationPct.lt(targetAllocationPct)) {
+            if (!rebalancesRequired[chainId]) rebalancesRequired[chainId] = {};
```
kevinuma, line 176: nit: Can doing assign() on line 179 help getting rid of this line?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -0,0 +1,367 @@
...
+      }
+
+      // Next, evaluate if we have enough tokens on L1 to actually do these rebalances.
+      for (const chainId of Object.keys(rebalancesRequired)) {
+        for (const l1Token of Object.keys(rebalancesRequired[chainId])) {
+          const requiredRebalance = rebalancesRequired[chainId][l1Token];
+          // If the amount required in the rebalance is less than the total amount of this token on L1 then we can execute
+          // the rebalance to this particular chain. Note that if the sum of all rebalances required exceeds the l1
+          // balance then this logic ensures that we only fill the first n number of chains where we can.
+          if (requiredRebalance.lt(this.tokenClient.getBalance(1, l1Token))) {
```
kevinuma, line 196: Can this balance potentially be outdated as the token client update in relayer happens before deposits are filled? I might misunderstand things but wouldn't any relays on Ethereum would change the balance of tokens, and thus render the balances out of date after the fills and before rebalanceInventoryIfNeeded is called?

src/clients/InventoryClient/BaseAdapter.ts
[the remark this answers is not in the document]
chrismaree: this has nothing to do with the finalizer though. this is canonical bridge transfers (i.e sending over the Optimism/arbitrum/boba/polygon canonical bridge). what you describe will work fine for the finalization transfers but not the canonical bridge transfers.

src/clients/InventoryClient/AdapterManager.ts
```diff
@@ -0,0 +1,138 @@
...
+    return this.spokePoolClients[chainId].eventSearchConfig;
+  }
+
+  l2TokenForL1Token(l1Token: string, chainId: number): string {
+    // the try catch below is a safety hatch. If you try fetch an L2 token that is not within the hubPoolClient for a
+    // given L1Token and chainId combo then you are likely trying to send a token to a chain that does not support it.
+    try {
+      // No that the line below is critical. if the hubpool returns the wrong destination token for the L1 token then
+      // the bot can irrecoverably send the wrong token to the chain and loose money.
+      const l2TokenForL1Token = this.hubPoolClient.getDestinationTokenForL1Token(l1Token, chainId);
```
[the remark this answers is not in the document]
chrismaree: could do that! it's more just a sanity check to validate that _everything else_ is working correctly. you could have a situation where you use the `l2TokensToL1TokenValidation` but something else is misconfigured or broken and then would not catch it.

src/clients/InventoryClient/ArbitrumAdapter.ts
chrismaree: this is JUST an arbitrum specific bridge thing. it's not really a gas limit; it's a parameter put into the function call when sending tokens over the bridge. you'll see we do something similar in the on-chain version of this here: https://github.com/across-protocol/contracts-v2/blob/34cd1a34b28e63439e3e0c9fdacfd30a85e15808/contracts/chain-adapters/Arbitrum_Adapter.sol#L52

src/clients/InventoryClient/InventoryClient.ts
chrismaree: good catch! there was a typo in the line above. that should not be assignment but actually a check if it is set.

src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+    for (const l1Token of l1Tokens) {
+      if (l1Gateways[l1Token] === undefined || l2Gateways[l1Token] === null) continue;
+
+      const l1Bridge = this.getL1Bridge(l1Token);
+      const l2Bridge = this.getL2Bridge(l1Token);
+      const searchFilter = [undefined, this.relayerAddress];
+      promises.push(
+        paginatedEventQuery(l1Bridge, l1Bridge.filters.DepositInitiated(...searchFilter), this.l1SearchConfig),
+        paginatedEventQuery(l2Bridge, l2Bridge.filters.DepositFinalized(...searchFilter), this.l2SearchConfig)
+      );
```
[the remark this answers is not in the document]
chrismaree: ye, there is a reason for this: I need to collect all the promises first (for each of the async calls) and then in one go run all the promises in the final `const results = await Promise.all(promises, { concurrency: 4 });` call. if structured the way you describe then the construction of each promise would not be able to be batched like this.

if done using `await Promise.all(l1Tokens.map(` then I would need to construct all the custom logic for each specific L1 token within the map operation. within the arbitrum adapter this is not too bad and it would look fine but take a look at the complexity in the optimism adapter, for example, and you'll see why what you propose will not work super cleanly.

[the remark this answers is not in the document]
chrismaree: as an additional point depending on the chain we are looking at the props in the events change! the line:
`return { amount: eventSpread[index % 2 === 0 ? "_amount" : "amount"], blockNumber: eventSpread["blockNumber"] };` needs to leverage the index to work out what prop to look for. this is posible in the pattern you've describe but will be more convoluted.

src/utils/ProviderUtils.ts
```diff
@@ -1,6 +1,6 @@
 import { ethers, providers } from "ethers";
 
-const stallTimeout = 15 * 1000;
+const stallTimeout = 10 * 1000;
```
nicholaspai, line 3: shortening this seems dangerous for bots that run very complex calculations like the dataworker, or perhaps i'm misunderstanding this use

src/clients/InventoryClient/OptimismAdapter.ts
chrismaree: not really. these are seperate configs. This is ONLY relevent for Optimism. it's kinda bespoke; each chain has it's own custom logic and they sit within these adapters.

src/clients/InventoryClient/InventoryClient.ts
chrismaree: this logic is run after the relayer logic is run. therefore the token client is updated with the latest requirement. that client is updated syncroniously (it does not need to fetch data from the chain) and so it always has the most recent balance information. it runs first to fetch the data async then syncronously updates as it goes along.

src/clients/InventoryClient/AdapterManager.ts
chrismaree: yup! I added a comment to this effect.

PR #89, opened 2022-05-30
refactor: Move cancelled and dispute event fetching into HubPoolClient and remove the redundant RootBundleProcessor in monitoring code

kevinuma:
Related PR: https://github.com/across-protocol/relayer-v2/pull/91


src/interfaces/Common.ts
```diff
@@ -4,6 +4,7 @@ export interface SortableEvent {
   blockNumber: number;
   transactionIndex: number;
   logIndex: number;
+  transactionHash: string;
```
nicholaspai, line 7: This might break many tests but I do agree we should add these here!

nicholaspai (commented):
Initial pass looks great!

src/interfaces/Common.ts
kevinuma: It only broke one :)

src/dataworker/Dataworker.ts
```diff
@@ -151,7 +151,7 @@ export class Dataworker {
               return;
 
             // Now create a copy of fill with block data removed, and use its data to update the fills to refund obj.
-            const { blockNumber, transactionIndex, logIndex, ...fill } = fillWithBlock;
+            const { blockNumber, transactionIndex, transactionHash, logIndex, ...fill } = fillWithBlock;
```
nicholaspai, line 154: why is this changed?

nicholaspai (approved):
This looks good to me, lets see what @chrismaree says and we can merge after you resolve all outstanding comments!

PR #92, opened 2022-05-31
feat: optimize dataworker for looping mode

nicholaspai:
This PR removes repetitive spoke pool client construction and updates on each loop run. All dataworker functions can re-use the same spoke pool clients and therefore not have to update a bunch of events from block 0 to HEAD each loop. This does make the assumption that the spoke pools are no upgraded in the middle of a looping bot's run, but I think this is a reasonable responsibility of the bot runner to reset bots after a spoke pool upgrades.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -24,7 +24,8 @@ export class HubPoolClient {
   constructor(
     readonly logger: winston.Logger,
     readonly hubPool: Contract,
-    readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 }
+    readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 },
+    readonly endBlockBuffer: number = 0
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
   }
@@ -34,7 +35,13 @@ export class HubPoolClient {
   }
 
   hasPendingProposal() {
-    return this.pendingRootBundle !== undefined && this.pendingRootBundle.unclaimedPoolRebalanceLeafCount > 0;
+    return (
+      this.pendingRootBundle !== undefined &&
+      this.pendingRootBundle.unclaimedPoolRebalanceLeafCount > 0 &&
+      this.pendingRootBundle.bundleEvaluationBlockNumbers !== undefined
+      // This will fail if the other two conditions pass
+      // if the client has seen all events up to HEAD and the root bundle was recently proposed
+    );
   }
 
   getPendingRootBundleIfAvailable() {
@@ -261,7 +268,7 @@ export class HubPoolClient {
     this.latestBlockNumber = await this.hubPool.provider.getBlockNumber();
     const searchConfig = {
       fromBlock: this.firstBlockToSearch,
-      toBlock: this.eventSearchConfig.toBlock || this.latestBlockNumber,
+      toBlock: this.eventSearchConfig.toBlock || this.latestBlockNumber - this.endBlockBuffer,
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
     this.logger.debug({ at: "HubPoolClient", message: "Updating HubPool client", searchConfig });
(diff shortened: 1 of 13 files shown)


src/dataworker/index.ts
```diff
@@ -37,53 +36,36 @@ export async function createDataworker(_logger: winston.Logger) {
...
-        at: "Dataworker#index",
-        message: "Constructing spoke pool clients for next root bundle",
-      });
-      const latestSpokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
-        dataworker.chainIdListForBundleEvaluationBlockNumbers,
-        clients,
-        logger,
-        clients.hubPoolClient.latestBlockNumber
-      );
+      if (spokePoolClients === undefined)
```
kevinuma, line 45: Just a note: I think we should eventually make it possible to share the clients (both hubpool + spokepools) among all the bots running in the same process. Currently each of them (data worker, relayer, monitor) create their own versions, which is very inefficient and wasteful.

nicholaspai: agreed

src/clients/HubPoolClient.ts
```diff
@@ -144,7 +144,7 @@ export class HubPoolClient {
       rootBundle,
       nextRootBundle ? Math.min(nextRootBundle.blockNumber, latestMainnetBlock) : latestMainnetBlock
     );
-    return executedLeafCount.length === rootBundle.poolRebalanceLeafCount;
+    return rootBundle.poolRebalanceLeafCount > 0 && executedLeafCount.length === rootBundle.poolRebalanceLeafCount;
```
nicholaspai, line 147: Just an extra sanity check that `poolRebalanceLeafCount` isn't set to 0 for some reason

src/clients/HubPoolClient.ts
```diff
@@ -336,35 +333,27 @@ export class HubPoolClient {
...
-    if (this.pendingRootBundle.poolRebalanceRoot !== EMPTY_MERKLE_ROOT) {
-      // Throw an error if this client is configured in such a way that the most recent event does not match
-      // the pending root bundle. We should handle this case eventually, but for now loudly error so we don't dispute
-      // the wrong pending bundle.
+    // If the contract's current rootBundleProposal() value has an unclaimedPoolRebalanceLeafCount > 0, then
+    // it means that either the root bundle proposal is in the challenge period and can be disputed, or it has
+    // passed the challenge period and pool rebalance leaves can be executed. Once all leaves are executed, the
+    // unclaimed count will drop to 0 and at that point there is nothing more that we can do with this root bundle
+    // besides proposing another one.
+    if (pendingRootBundleProposal.unclaimedPoolRebalanceLeafCount > 0) {
```
nicholaspai, line 341: I think this check for whether a `rootBundleProposal()` return value is valid is much cleaner and guarantees that there is no potential clash from a pending root bundle's roots and its bundle block numbers. I was seeing strange issues locally where I was running the disputer and proposer (in looping mode) in parallel, proposed a root, and the disputer was loading the new root bundle correctly but not the bundle block evaluation numbers

src/interfaces/HubPool.ts
```diff
@@ -66,7 +66,7 @@ export interface SetPoolRebalanceRoot extends SortableEvent {
   destinationToken: string;
 }
 
-export interface RootBundle {
+export interface PendingRootBundle {
```
nicholaspai, line 69: This name makes it more obvious that this is different from the `ProposedRootBundle` interface used for the event, versus this one used for the contract state of `rootBundleProposal()`

mrice32 (approved):
LGTM!

src/clients/HubPoolClient.ts
```diff
@@ -29,20 +28,18 @@ export class HubPoolClient {
...
 
   hasPendingProposal() {
-    return this.pendingRootBundle !== undefined && this.pendingRootBundle.unclaimedPoolRebalanceLeafCount > 0;
+    return this.pendingRootBundle !== undefined;
   }
 
-  getPendingRootBundleIfAvailable() {
-    const hasPendingProposal = this.hasPendingProposal();
-    const pendingRootBundle = hasPendingProposal ? this.getPendingRootBundleProposal() : undefined;
+  getPendingRootBundle() {
```
mrice32, line 39: nit: this name may be a bit confusing. Why include hasPendingProposal instead of just returning the pending proposal since there's a separate method for hasPendingProposal?

PR #93, opened 2022-06-01
feat: Use spoke pool client lookback in dataworker

nicholaspai:
Certain functions in the dataworker like the leaf executor shouldn't have to keep looking back from the beginning of history. Instead, we can strategically set the `maxSpokeClientLookBack` per chain in combination with `spokeRootsLookbackCount` to limit the web3 requests per leaf executor run.

An example configuration would look back only ~48 hours of blocks and also only try to execute the most recent root bundle per chain.

```
MAX_SPOKE_CLIENT_LOOK_BACK='{ "1": 11500, "10": 400000, "137": 100000, "288": 5000, "42161": 150000 }'
SPOKE_ROOTS_LOOKBACK_COUNT=1
```

We could also strategically set the finalizer's (or even disputer's lookback) to a larger lookback but still very safe length of 7 days:

```
MAX_SPOKE_CLIENT_LOOK_BACK='{ "1": 46000, "10": 1600000, "137": 400000, "288": 20000, "42161": 600000 }'
```

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -1,15 +1,18 @@
 import winston from "winston";
 import { getProvider, getSigner, getDeployedContract, getDeploymentBlockNumber } from "../utils";
-import { Wallet, SpokePool, Contract } from "../utils";
+import { Wallet, Contract } from "../utils";
 import { HubPoolClient, MultiCallerClient, AcrossConfigStoreClient, SpokePoolClient, ProfitClient } from "../clients";
 import { CommonConfig } from "./Config";
-import { DataworkerClients } from "../dataworker/DataworkerClientHelper";
 
 export interface Clients {
   hubPoolClient: HubPoolClient;
   configStoreClient: AcrossConfigStoreClient;
   multiCallerClient: MultiCallerClient;
   profitClient: ProfitClient;
+}
+
+export interface SpokePoolClientsByChain {
+  [chainId: number]: SpokePoolClient;
 }
 
 export function getSpokePoolSigners(baseSigner: Wallet, config: CommonConfig): { [chainId: number]: Wallet } {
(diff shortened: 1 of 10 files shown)


src/common/ClientHelper.ts
```diff
@@ -20,31 +23,51 @@ export function getSpokePoolSigners(baseSigner: Wallet, config: CommonConfig): {
   );
 }
 
-export async function constructSpokePoolClientsForBlockAndUpdate(
-  chainIdListForBundleEvaluationBlockNumbers: number[],
-  clients: DataworkerClients,
+export async function constructSpokePoolClientsWithLookback(
```
nicholaspai, line 26: `constructSpokePoolClientsWithLookback` was just moved over from `RelayerClientHelper`

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -1,23 +1,44 @@
...
-import { getBlockRangeForChain } from "./DataworkerUtils";
 
 export interface DataworkerClients extends Clients {
   tokenClient: TokenClient;
   spokePoolSigners: { [chainId: number]: Wallet };
   spokePoolClientSearchSettings: { [chainId: number]: EventSearchConfig };
 }
 
+// TODO: Deprecate this function in favor of constructSpokePoolClientsWithLookback() in common/
+export async function constructSpokePoolClientsForBlockAndUpdate(
```
nicholaspai, line 14: Moved over from `common` and we should remove eventually but not this PR

src/dataworker/index.ts
```diff
@@ -34,23 +33,22 @@ export async function createDataworker(_logger: winston.Logger) {
...
       await updateDataworkerClients(clients);
       if (spokePoolClients === undefined)
-        spokePoolClients = await constructSpokePoolClientsForBlockAndUpdate(
-          dataworker.chainIdListForBundleEvaluationBlockNumbers,
-          clients,
-          logger,
-          clients.hubPoolClient.latestBlockNumber
-        );
-      else await updateSpokePoolClients(spokePoolClients);
+        spokePoolClients = await constructSpokePoolClientsWithLookback(logger, clients, config, baseSigner);
```
nicholaspai, line 50: This is the important change: create spoke pool clients with a `fromBlock` taking into account the lookback

PR #102, opened 2022-06-01
feat: add loadData cache for dataworker

mrice32:


src/dataworker/Dataworker.ts
```diff
@@ -82,6 +97,12 @@ export class Dataworker {
     allValidFills: FillWithBlock[];
     deposits: DepositWithBlock[];
   } {
+    const key = JSON.stringify(blockRangesForChains);
```
nicholaspai, line 100: What if `blockRangesForChains` is null for whatever reason? Should we care about this edge case?

mrice32: It can't be null, right?

PR #104, opened 2022-06-02
improve: exit early before checking balance allocator and refactor redis functions

nicholaspai:


src/dataworker/Dataworker.ts
```diff
@@ -1009,6 +1010,7 @@ export class Dataworker {
     const unexecutedLeaves = expectedTrees.poolRebalanceTree.leaves.filter((leaf) =>
       executedLeaves.every(({ leafId }) => leafId !== leaf.leafId)
     );
+    if (unexecutedLeaves.length === 0) return;
```
mrice32, line 1013: Do you think it's possible that it's the chainId call that's slowing this down?

I think the chainId call below is actually redundant, since we should just be able to use hubPoolChainId from above, right?

mrice32 (commented):
One minor comment. I'm curious if dropping the chainId call (which is correct since the call is a duplicate) would cause the same time savings.

src/dataworker/Dataworker.ts
nicholaspai: That duplicate only exists in the `executePoolRebalanceLeaves` function, and also when I was timing the functions I started the timer after that chainId load. I don't understand why the `Promise.all()` is spending any time at all on a zero length array

src/common/ClientHelper.ts
```diff
@@ -91,6 +91,11 @@ export async function constructClients(logger: winston.Logger, config: CommonCon
       url: config.redisUrl,
     });
     await redisClient.connect();
+    logger.debug({
+      at: "Dataworker#ClientHelper",
+      message: `Connected to redis server at ${config.redisUrl} successfully!`,
+      dbSize: await redisClient.dbSize(),
```
nicholaspai, line 97: @mrice32 i think this is a helpful log wdyt?

src/dataworker/Dataworker.ts
mrice32: Interesting.

mrice32 (approved):
LGTM!

PR #106, opened 2022-06-02
feat: Add ability for finalize to share clients with proposer and disputer, and add disputer simulation toggle

nicholaspai:


src/dataworker/Dataworker.ts
```diff
@@ -452,7 +452,10 @@ export class Dataworker {
     );
   }
 
-  async validatePendingRootBundle(spokePoolClients?: { [chainId: number]: SpokePoolClient }) {
+  async validatePendingRootBundle(
+    spokePoolClients?: { [chainId: number]: SpokePoolClient },
+    submitDisputes: boolean = true
```
nicholaspai, line 457: Useful toggle that we can use to allow `disputer` to share clients with leaf executor and finalizer but not send actual dispute txns but will still send pagerdity alerts

src/dataworker/index.ts
```diff
@@ -55,24 +56,41 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
...
+          balanceAllocator,
+          config.sendingExecutionsEnabled
+        );
       } else logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
 
-      await clients.multiCallerClient.executeTransactionQueue(!config.sendingTransactionsEnabled);
+      await clients.multiCallerClient.executeTransactionQueue();
+
+      if (config.finalizerEnabled)
+        await finalize(logger, clients.hubSigner, clients.hubPoolClient, spokePoolClients, config.finalizerChains);
```
nicholaspai, line 92: Removed finalizer's `index.ts` in favor of consoldating into this file so we can run all bots with same spoke client

src/dataworker/DataworkerConfig.ts
```diff
@@ -23,6 +35,11 @@ export class DataworkerConfig extends CommonConfig {
       PROPOSER_ENABLED,
       EXECUTOR_ENABLED,
       SPOKE_ROOTS_LOOKBACK_COUNT,
+      SEND_DISPUTES,
+      SEND_PROPOSALS,
+      SEND_EXECUTIONS,
+      FINALIZER_CHAINS,
+      FINALIZER_ENABLED,
```
mrice32, line 42: Not necessary to change, but is there a reason you chose not to have a simulation mode for the finalizer?

src/dataworker/index.ts
mrice32: +1

mrice32 (approved):
LGTM!

src/dataworker/DataworkerConfig.ts
nicholaspai: Yeah it's annoying bc there 4 different txns to send right now. I'll build out a dsproxy integration eventually

PR #108, opened 2022-06-02
feat: Add reporting to monitor bot

kevinuma:
Add a function to report relayer balances and pending refunds on different chains. We want to send this report to Slack (or console if run locally), so the relayer can manage their funds effectively.

Considerations:
1. There are two ways to get all tokens a SpokePool supports (so we can report balances): Via HubPool's Rebalance events (cached in HubPoolClient) and SpokePool's EnabledDepositRoute event (cached in SpokePoolClient). The former requires a much shorter lookback window (just long enough to see the last bundle executions) while the latter requires a lookback window as long as the SpokePool's age. Therefore, this PR went with the former approach for efficiency.
2. https://github.com/across-protocol/relayer-v2/pull/100 missed a call to update all SpokePoolClients. This PR corrects it.


chrismaree:
nit: please add a description on your PR! I know not all the other PRs in this repo have this always but it would be useful to help in reviewing.

kevinuma:
> nit: please add a description on your PR! I know not all the other PRs in this repo have this always but it would be useful to help in reviewing.

Added 🙏

src/monitor/Monitor.ts
```diff
@@ -61,43 +81,131 @@ export class Monitor {
...
-        `An unknown EOA ${etherscanLink(event.caller, 1)} has proposed a bundle on ${getNetworkName(1)}` +
-        `\ntx: ${etherscanLink(event.transactionHash, 1)}`;
-      this.logger.error({ at: "Monitor", message: "Unknown bundle proposer 🥷", mrkdwn });
+    // Send notifications if there are any bundle activities from an unknown data worker.
+    for (const event of proposedBundles) {
+      this.notifyIfUnknownCaller(event.proposer, "proposed", event.transactionHash);
+    }
+    for (const event of cancelledBundles) {
+      this.notifyIfUnknownCaller(event.disputer, "cancelled", event.transactionHash);
+    }
```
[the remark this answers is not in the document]
kevinuma: Updated.

src/clients/SpokePoolClient.ts
```diff
@@ -233,7 +239,12 @@ export class SpokePoolClient {
 
     for (const event of enableDepositsEvents) {
       const enableDeposit = spreadEvent(event);
-      assign(this.depositRoutes, [enableDeposit.originToken, enableDeposit.destinationChainId], enableDeposit.enabled);
+      const originToken = enableDeposit.originToken;
+      assign(this.depositRoutes, [originToken, enableDeposit.destinationChainId], enableDeposit.enabled);
+
+      const tokenContract = new Contract(originToken, ERC20.abi, this.spokePool.signer);
+      const [decimals, symbol] = await Promise.all([tokenContract.decimals(), tokenContract.symbol()]);
```
nicholaspai, line 246: This is going to add 2 web3 calls per spoke pool client update, not ideal. Can we cache decimals and symbol? They should really only be loaded once for an origin token address for this chain

src/interfaces/Common.ts
```diff
@@ -15,3 +15,8 @@ export interface TreeData<T> {
   tree: MerkleTree<T>;
   leaves: T[];
 }
+
+export interface TokenInfo {
```
nicholaspai, line 19: You should re-use `interfaces/HubPool/L1Token`

src/monitor/Monitor.ts
```diff
@@ -8,18 +12,33 @@ enum BundleAction {
   CANCELED = "canceled",
 }
 
+interface TokenBalance {
```
nicholaspai, line 15: This could extend `L1TokenInfo` and just add `balance`

src/monitor/Monitor.ts
```diff
@@ -104,6 +123,67 @@ export class Monitor {
     }
   }
 
+  // Report balances of all tokens on each supported chain for each relayer.
+  async reportRelayerBalances(): Promise<void> {
+    // Short-circuit if not enough time has passed since the last report.
+    const currentTimestamp = getCurrentTime();
+    if (currentTimestamp - this.lastReportTimestamp < this.monitorConfig.monitorReportInterval) {
```
nicholaspai, line 130: Are you sure that `monitorReportInterval` can't be undefined?

nicholaspai: Also, you should consider just deleting `monitorReportInterval` and assuming that you run the monitor on a 15 min cadence instead

src/monitor/MonitorConfig.ts
```diff
@@ -37,17 +44,18 @@ export class MonitorConfig extends CommonConfig {
       utilizationEnabled: UTILIZATION_ENABLED === "true",
       unknownRootBundleCallersEnabled: UNKNOWN_ROOT_BUNDLE_CALLERS_ENABLED === "true",
       unknownRelayerCallersEnabled: UNKNOWN_RELAYER_CALLERS_ENABLED === "true",
+      reportEnabled: MONITOR_REPORT_ENABLED === "true",
```
nicholaspai, line 47: Is this useful? Is there ever a case where you wouldn't want to send the report?

nicholaspai (commented):
Left some more comments!

src/clients/SpokePoolClient.ts
kevinuma: This got moved into Monitor. Even in this previous code, this would only do the calls once per EnabledDepositRoute which should only happen once per token per the Monitor bot instance's lifetime.

src/interfaces/Common.ts
kevinuma: Updated

src/monitor/Monitor.ts
kevinuma: I refactored the code so this is no longer needed and has been removed.

src/monitor/Monitor.ts
kevinuma: Removed

src/monitor/MonitorConfig.ts
kevinuma: Yes this can be useful later on when external users might be able to run these bots but don't want to have the report produced

src/clients/HubPoolClient.ts
```diff
@@ -109,6 +109,14 @@ export class HubPoolClient {
     return this.l1TokensToDestinationTokens[l1Token][destinationChainId];
   }
 
+  getDestinationTokensForChainId(chainId: number) {
+    let l2TokenAddresses = this.l1Tokens.map((l1Token) =>
+      this.getDestinationTokenForL1TokenDestinationChainId(l1Token.address, chainId)
+    );
+    l2TokenAddresses = l2TokenAddresses.filter((token) => token !== undefined);
```
nicholaspai, line 116: I think its just as readable and bit more concise to chain the `.filter` on the `.map`

src/clients/SpokePoolClient.ts
```diff
@@ -4,7 +4,6 @@ import { toBN, Event, ZERO_ADDRESS, winston, paginatedEventQuery, spreadEventWit
 import { AcrossConfigStoreClient } from "./ConfigStoreClient";
 import { Deposit, DepositWithBlock, Fill, SpeedUp, FillWithBlock, TokensBridged } from "../interfaces/SpokePool";
 import { RootBundleRelayWithBlock, RelayerRefundExecutionWithBlock } from "../interfaces/SpokePool";
-import { RootBundleExecutedEvent } from "@across-protocol/contracts-v2/dist/typechain/HubPool";
```
nicholaspai, line 7: +1 removing these unused imports

src/common/ClientHelper.ts
```diff
@@ -57,6 +58,56 @@ export async function constructSpokePoolClientsForBlockAndUpdate(
   return spokePoolClients;
 }
 
+export async function constructSpokePoolClientsWithLookback(
```
nicholaspai, line 61: Did you change anything in this function?

src/clients/HubPoolClient.ts
kevinuma: Updated

src/common/ClientHelper.ts
kevinuma: No, except for changing the type of the config argument to CommonConfig from RelayerConfig and a rename of the relayerLookback to initialLookbackOverride

src/relayer/RelayerClientHelper.ts
```diff
@@ -1,69 +1,28 @@
...
+import { Clients, constructClients, updateClients, updateSpokePoolClients } from "../common";
 import { SpokePoolClientsByChain } from "../interfaces";
+import { constructSpokePoolClientsWithLookback } from "../common";
 
 export interface RelayerClients extends Clients {
   spokePoolClients: SpokePoolClientsByChain;
   tokenClient: TokenClient;
 }
 
-export async function constructSpokePoolClientsWithLookback(
```
kevinuma, line 13: Note: This has been moved to [src/common/ClientHelper.ts](https://github.com/across-protocol/relayer-v2/pull/108/files#diff-164e17c689ae4c5c58d16bbc6aa11708876b38f189597ffce51e39a31a168162)

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+    // Refunds that will be processed in the current bundle still pending liveness (if any).
+    let pendingFillsToRefundPerChain: FillsToRefund | undefined = undefined;
+
+    let remainingLivenessMins = 0;
+    // If latest proposed bundle has not been executed yet, we have a bundle pending liveness.
+    if (!latestFullyExecutedBundle || latestBundle.blockNumber != latestFullyExecutedBundle.blockNumber) {
+      // Calculate the block ranges to calculate the current pending bundle's content.
+      const pendingBundleEvaluationBlockRanges: number[][] = [];
+      for (const [chainIndex, endingBlock] of latestBundle.bundleEvaluationBlockNumbers.entries()) {
+        const startingBlock = latestFullyExecutedBundle
```
nicholaspai, line 164: Why not use `HubPoolClient.getNextBundleStartBlockNumber` here? As an input for `latestMainnetBlock` you could use the mainnet block height for the `latestBundle .bundleEvalBlockNumbers`

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+      pendingFillsToRefundPerChain = fillsToRefund;
+
+      // Calculate remaining time in liveness.
+      remainingLivenessMins = Math.floor((latestBundle.challengePeriodEndTimestamp - getCurrentTime()) / 60);
+    }
+
+    // Calculate the block ranges to calculate the next bundle's content.
+    const futureBundleEvaluationBlockRanges: number[][] = [];
+    for (const endingBlock of latestBundle.bundleEvaluationBlockNumbers.values()) {
+      futureBundleEvaluationBlockRanges.push([endingBlock.toNumber() + 1, latestBlockNumber]);
```
nicholaspai, line 183: Again you could maybe more succinctly call `getNextBundleStartBlock` and passing in latest mainnet block as a param

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+    const latestBlockNumber = hubPoolClient.latestBlockNumber;
+    const latestBundle = hubPoolClient.getMostRecentProposedRootBundle(latestBlockNumber);
+    const latestFullyExecutedBundle = hubPoolClient.getLatestFullyExecutedRootBundle(latestBlockNumber);
+
+    // Refunds that will be processed in the current bundle still pending liveness (if any).
+    let pendingFillsToRefundPerChain: FillsToRefund | undefined = undefined;
+
+    let remainingLivenessMins = 0;
+    // If latest proposed bundle has not been executed yet, we have a bundle pending liveness.
+    if (!latestFullyExecutedBundle || latestBundle.blockNumber != latestFullyExecutedBundle.blockNumber) {
```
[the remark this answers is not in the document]
nicholaspai: This could replace `latestFullyExecutedBundle `

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+    let foundAnyRefunds = false;
+    for (const chainIdString of Object.keys(fillsToRefundPerChain)) {
+      mrkdwn += `*Chain ${chainIdString}*\n`;
+
+      let chainId = parseInt(chainIdString);
+      const fillsToRefund = fillsToRefundPerChain[chainId];
+      for (const tokenAddress of Object.keys(fillsToRefund)) {
+        const totalRefundAmount = fillsToRefund[tokenAddress].refunds[relayer];
+        const token = new Contract(tokenAddress, ERC20.abi, this.clients.spokePoolClients[chainId].spokePool.signer);
+        const tokenInfo = await this.getTokenInfo(chainId, token);
```
nicholaspai, line 239: You should probably cache `getTokenInfo` since it shouldn't change call to call.

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+        // Convert to number of tokens for readability.
+        const refund = convertFromWei(totalRefundAmount.toString(), tokenInfo.decimals);
+        mrkdwn += `${tokenInfo.symbol}: ${refund}\n`;
```
nicholaspai, line 242: This `mrkdwn` is going to become a very large list with potentially hundreds of fills per bundle. Can you possibly summairize them and use some strategic grouping by token/refund recipient, etc?

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+  }
+
+  private async notifyRelayerRefunds(
+    fillsToRefundPerChain: FillsToRefund,
+    relayer: string,
+    refundsMsg: string,
+    timeToRefunds: number,
+    noRefundMsg: string
+  ) {
+    let mrkdwn = `Refund ETA: ~${timeToRefunds} mins\n`;
```
nicholaspai, line 229: I don't think this log of Refund ETA is super useful. I would rather just print out a readable version of the bundle's challenge period end time instead instead of estimating ETA. This is motivated by wanting to mitigate the number of web3 calls we make. So I'd love to remove that `hubPool.livenss()` call. Also, realisticially the liveness amoutn isn't going to change. So we could just hardcode 2 hours in as the liveness amount and save the web3 call...

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+        await this.notifyRelayerRefunds(
+          pendingFillsToRefundPerChain,
+          relayer,
+          pendingRefundsMessage,
+          remainingLivenessMins,
+          noPendingRefundsMessage
+        );
+      }
+
+      const futureRefundsMessage = `Relayer (${relayer})'s refunds in the next bundle ⌛`;
```
nicholaspai, line 210:
```suggestion
      const futureRefundsMessage = `Relayer (${relayer})'s refunds to be included in the next bundle ⌛`;
```

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+    }
+    // Refunds that will be processed in the next bundle that will be proposed after the current pending bundle
+    // (if any) has been fully executed.
+    const { fillsToRefund: futureFillsToRefundPerChain } = this.bundleDataClient.loadData(
+      futureBundleEvaluationBlockRanges,
+      this.clients.spokePoolClients,
+      false
+    );
+    // Next bundle can only be executed once the current pending bundle (if any) passes liveness.
+    const timeToNextBundle = remainingLivenessMins + Math.floor(hubPoolClient.bundleLivenessSecs / 60);
```
nicholaspai, line 193: This is actually going to be a very inaccurate time estimate. I would just argue not even signaling when the next bundle will happen, just make a note which refunds will be included in next bundle (as you do now). This is because the next bundle will not be constructed until the total `netSendAmounts` (absolute value) sum to some $X threshold. Default to $500,000 and currently set to $1,000,000. So there is potentially a much longer time than 2 hours to wait for next bundle

src/monitor/Monitor.ts
```diff
@@ -104,6 +112,44 @@ export class Monitor {
...
+
+        this.logger.info({ at: "Monitor", message: `Token balances for chain ${chainId} 🚀`, mrkdwn });
+      }
+    }
+  }
+
+  private async getTokenInfo(chainId: number, token: Contract): Promise<L1Token> {
+    const address = token.address;
+    // Populate cache if we have not fetched the token info before.
+    if (!this.cachedTokenInfo[chainId] || !this.cachedTokenInfo[chainId][address]) {
```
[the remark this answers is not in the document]
nicholaspai: Bump here. Ideally can make this function not async

src/utils/FillUtils.ts
```diff
@@ -1,7 +1,22 @@
 import { HubPoolClient } from "../clients";
-import { Fill, FillsToRefund, FillWithBlock } from "../interfaces";
-import { assign, getRealizedLpFeeForFills, getRefundForFills, sortEventsDescending, toBN } from "../utils";
-import { getBlockRangeForChain } from "./DataworkerUtils";
+import { DepositWithBlock, Fill, FillsToRefund, FillWithBlock, UnfilledDeposit } from "../interfaces";
+import { assign, getRealizedLpFeeForFills, getRefundForFills, sortEventsDescending, toBN } from "./";
+import { groupObjectCountsByProp, groupObjectCountsByTwoProps, getDepositPath } from "./";
+
+export function getBlockRangeForChain(
```
nicholaspai, line 6: I don't think this makes sense as a `FillUtil`, as its used all across the dataworker methods. Why change it?

src/utils/FillUtils.ts
```diff
@@ -149,3 +164,48 @@ export function getFillsInRange(
     return fill.blockNumber <= blockRangeForChain[1] && fill.blockNumber >= blockRangeForChain[0];
   });
 }
+
+export function prettyPrintSpokePoolEvents(
```
nicholaspai, line 168: Again, this doesn't seem like a "Fill" util, why not add to the BundleClient?

nicholaspai: Or just leave it where it was

test/Monitor.ts
```diff
@@ -98,12 +113,21 @@ describe("Monitor", async function () {
...
   });
+
+  it("Monitor should report balances", async function () {
+    await monitorInstance.update();
+    await monitorInstance.reportRelayerBalances();
+
+    expect(lastSpyLogIncludes(spy, "Token balances for chain 1337 🚀")).to.be.true;
+    const balancesLog = spy.lastCall;
+    expect(balancesLog.lastArg.mrkdwn).to.equal(
+      "*USDC balance: 30,000.00*\n*USDC balance: 30,000.00*\n*USDC balance: 30,000.00*\n*USDC balance: 30,000.00*\n*USDC balance: 30,000.00*\n"
```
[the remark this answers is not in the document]
kevinuma: It's actually not easy to get the mrkdwn without replicating the logic from the tested code, which kinda defeats the purpose of testing the code I guess. I changed the test to just assert that the markdown contains the token balance amounts instead of checking the full content. I think this is okay for testing here as we don't care too much about the actual message anyway just that the content is correct.

src/monitor/Monitor.ts
```diff
@@ -13,24 +18,29 @@ export class Monitor {
...
 
   public async update() {
-    await this.clients.hubPoolClient.update();
+    await updateMonitorClients(this.clients);
     await this.computeHubPoolBlocks();
     await this.computeSpokePoolsBlocks();
   }
 
-  async checkUtilization(): Promise<void> {
+  async checkUtilization() {
```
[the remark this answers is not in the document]
kevinuma: I don't think this adds any value as the Typescript compiler can infer that there's no return + this is an async function => Promise<void>

src/monitor/Monitor.ts
```diff
@@ -104,6 +114,153 @@ export class Monitor {
...
+        // Use the provider from the spoke pool to send requests to the right provider.
+        const spokePoolClient = this.clients.spokePoolClients[chainId];
+        const tokens = tokenAddresses.map(
+          (tokenAddress) => new Contract(tokenAddress, ERC20.abi, spokePoolClient.spokePool.signer)
+        );
+
+        let mrkdwn = "";
+        // TODO: Potentially combine these into one single request to JSON RPC provider to save on bandwidth.
+        for (const token of tokens) {
+          const balanceBN = await token.balanceOf(relayer);
```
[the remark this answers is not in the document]
kevinuma: Isn't that what await does? My understanding is that async calls in Javascript are cooperative - any time there's a blocking call such as a network call, it switches context. Is this correct?

src/monitor/Monitor.ts
kevinuma: Updated to call hubPoolClient.getL1TokenInfoForL2Token which does that caching already

src/monitor/Monitor.ts
kevinuma: hubPoolClient.liveness is cached and the way I did it in hubPoolCLient.udpate(), it's only ever updated once. Let me ask Korpi and Kevin if the ETA would be useful at all, if not, I'll remove it

src/monitor/Monitor.ts
kevinuma: Updated.

src/monitor/Monitor.ts
kevinuma: Good point. Updated to remove the ETA for future bundles.

src/utils/FillUtils.ts
kevinuma: Same as the reasoning above. I think utils functions should never call anything not in common directories/packages.

src/monitor/Monitor.ts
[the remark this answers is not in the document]
nicholaspai: no await is synchronous. If you didn't have await, then they would be in parallel but you would not know when they end. Await means you won't go to next line until promise resolves

src/monitor/Monitor.ts
nicholaspai: Ah you're right good point

src/monitor/Monitor.ts
nicholaspai: but in a severless environment you're sending one web3 request per run. I don't think ETA vs end timestamp is a big UX difference so I'd prefer leaving this out

src/utils/FillUtils.ts
nicholaspai: No i mean, `DataworkerUtils` not `Dataworker`

src/utils/FillUtils.ts
nicholaspai: Can't you add to DataworkerUtils?

src/monitor/Monitor.ts
kevinuma: It wouldn't work here I think we because getNextBundleStartBlock uses the most recent fully executed bundle's ending block + 1 as the next starting block. Here we actually want the ending block of the potentially still pending bundle (can be different from the last fully executed block).

src/monitor/Monitor.ts
kevinuma: Good idea on using getNextBundleStartBlockNumber but we can actually use the latest block number here in the call because getNextBundleStartBlockNumber looks at the last fully executed bundle. Here in this conditional, we know there's a currently pending bundle, therefore, it would not get accidentally selected by getNextBundleStartBlockNumber

src/utils/FillUtils.ts
kevinuma: I put it back in DataworkerUtils

src/utils/FillUtils.ts
kevinuma: I put it back in DataworkerUtils

src/monitor/Monitor.ts
kevinuma: This ETA is actually not dependent on fetching liveness in HubPoolClient, it was the other ETA for the future bundle that needed it. This current pending bundle's ETA uses the latest proposed bundle's challengeTimestampEnd, which is already existing. I don't feel too strongly here but since it's free, we might as well use it :) If you still disagree, I'd be more than happy to remove.

nicholaspai: I just want to avoid the `hubPool.liveness()` async web3 request if possible, so as long as you can remove that I'm good.

src/dataworker/Dataworker.ts
```diff
@@ -88,175 +66,8 @@ export class Dataworker {
       });
   }
 
-  // This should be called whenever it's possible that the loadData information for a block range could have changed.
-  // For instance, if the spoke or hub clients have been updated, it probably makes sense to clear this to be safe.
-  clearCache() {
-    this.loadDataCache = {};
-    this.rootCache = {};
```
nicholaspai, line 95: Deleting this is a bug as `this.rootCache` is not updated in `_loadData`. I think you should keep Dataworker.clearCache() and just have it call `bundleClient.clearCache()` along with clearing the `rootCache`

src/dataworker/index.ts
```diff
@@ -44,7 +44,7 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
 
     for (;;) {
       // Clear cache so results can be updated with new data.
-      dataworker.clearCache();
+      dataworker.bundleDataClient.clearCache();
```
nicholaspai, line 47: As mentioned in https://github.com/across-protocol/relayer-v2/pull/108/files#r891376634 I would opt to keep this `clearCache` call

src/monitor/Monitor.ts
```diff
@@ -13,24 +27,29 @@ export class Monitor {
...
+  } = {};
+  private bundleDataClient: BundleDataClient;
 
   public constructor(
     readonly logger: winston.Logger,
     readonly monitorConfig: MonitorConfig,
     readonly clients: MonitorClients
   ) {
-    for (const chainId of Object.keys(clients.spokePoolClients)) {
+    this.bundleDataClient = new BundleDataClient(logger, clients, monitorConfig.spokePoolChains);
```
nicholaspai, line 40: Why break the `ClientHelper` pattern and not just add `bundleDataClient` to `MonitorClients`?

src/monitor/Monitor.ts
```diff
@@ -104,6 +123,169 @@ export class Monitor {
     }
   }
 
+  // Report balances of all tokens on each supported chain for each relayer.
+  async reportRelayerBalances() {
+    const relayers = this.monitorConfig.monitoredRelayers;
```
nicholaspai, line 128: I'd love to extract some of this logic that maps `relayer` addresses per chain to current balances. This way @chrismaree 's inventory management can take advantage of it + matt's upcoming PR that will examine how long it will take to fill a deposit of size X

src/monitor/Monitor.ts
```diff
@@ -104,6 +123,169 @@ export class Monitor {
...
+    for (const relayer of relayers) {
+      for (const chainId of this.monitorConfig.spokePoolChains) {
+        const tokenAddresses = this.clients.hubPoolClient.getDestinationTokensForChainId(chainId);
+
+        // Use the provider from the spoke pool to send requests to the right provider.
+        const spokePoolClient = this.clients.spokePoolClients[chainId];
+        const tokenBalances = await Promise.all(
```
nicholaspai, line 135: You should probably use the `TokenClient` here which has helful methods like `getBalance`

src/monitor/Monitor.ts
```diff
@@ -104,6 +123,169 @@ export class Monitor {
...
+    // Refunds that will be processed in the current bundle still pending liveness (if any).
+    let pendingFillsToRefundPerChain: FillsToRefund | undefined = undefined;
+
+    let remainingLivenessMins = 0;
+    // If latest proposed bundle has not been executed yet, we have a bundle pending liveness.
+    if (hubPoolClient.getExecutedLeavesForRootBundle(latestBundle, latestBlockNumber).length == 0) {
+      // Use the same block range as the current pending bundle to ensure we look at the right refunds.
+      const pendingBundleEvaluationBlockRanges: number[][] = [];
+      const latestBundleEvalNumbers = latestBundle.bundleEvaluationBlockNumbers;
+      for (let i = 0; i < latestBundleEvalNumbers.length; i++) {
```
nicholaspai, line 173: I kind of think its more readable to do something like:

```js
const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map((endBlock) => 
    [
        hubPoolClient.getNextBundleStartBlockNumber(chainIds, latestBlockNumber, chainIds[i]),
        endBlock.toNumber()
    ]
)
```

src/monitor/Monitor.ts
```diff
@@ -104,6 +123,169 @@ export class Monitor {
...
+        const startingBlock = hubPoolClient.getNextBundleStartBlockNumber(chainIds, latestBlockNumber, chainIds[i]);
+        pendingBundleEvaluationBlockRanges.push([startingBlock, latestBundleEvalNumbers[i].toNumber()]);
+      }
+      const { fillsToRefund } = this.bundleDataClient.loadData(
```
nicholaspai, line 177:
```suggestion
      const { fillsToRefund: pendingFillsToRefundPerChain } = this.bundleDataClient.loadData(
```

Thoughts? Would have to delete the following line of course

src/monitor/Monitor.ts
nicholaspai: Same comment as above about using `.map` to improve readability

src/utils/FillUtils.ts
```diff
@@ -149,3 +149,30 @@ export function getFillsInRange(
     return fill.blockNumber <= blockRangeForChain[1] && fill.blockNumber >= blockRangeForChain[0];
   });
 }
+
+// Returns all unfilled deposits over all spokePoolClients. Return values include the amount of the unfilled deposit.
+export function getUnfilledDeposits(
```
nicholaspai, line 154: Good refactor

nicholaspai (commented):
Left another round of reviews! I only found one bug, and the rest are stylistic comments

src/monitor/Monitor.ts
kevinuma: This is needed because pendingFillsToRefundPerChain is declared outside the if conditional. If we define it here inside the if, it'd shadow the outside variable and would not be available to look at later outside on line 196.

kevinuma: This is slightly more complex code that I hoped for but it reduces the amount of redundant code to account for 3 possible cases:

1. There's a pending bundle and a future bundle with refunds
2. There's a pending bundle but no future bundle with refunds
3. There's no pending bundle but there's a future bundle with refunds

src/monitor/Monitor.ts
kevinuma: I looked at it originally but decided not to use it as it has a lot of unnecessary caching logic, especially based on SpokePoolClients that's not needed here. To use it efficiently, we'd need to do some refactoring which might not be worth it just to get balances which can be done in 1 line here.

src/monitor/Monitor.ts
kevinuma: Can this refactoring be done in a separate PR? I need to take a look at Chris' PR and also understand what Matt wants to do. In general, I think reusing code is great but having more than one use cases first makes sense so the reusable code is written in a generic enough manner.

src/monitor/Monitor.ts
kevinuma: This is because the chain ids for bundle block number eval for data worker is the list of all chains (Constants.CHAIN_ID_LIST_INDICES) while for monitor, this list comes from CONFIGURED_NETWORKS, which is saved in CommonConfig. This somewhat makes sense to me because dataworker always needs to consider all chains while other bots such as relayer and monitor would worry about specific chains that the user wants to support. In general, we need a broader discussion on whether this patter makes sense, especially for external usages later on.

src/dataworker/Dataworker.ts
kevinuma: Thanks for catching this. My bad for somehow missing the other cache clearing.

src/dataworker/index.ts
kevinuma: Updated.

src/monitor/Monitor.ts
kevinuma: Great suggestion, thanks! Updated.

src/dataworker/Dataworker.ts
nicholaspai: We should actually just add a simple `clearCache` test for the Dataworker to make sure we don't accidentally make this error again in the future. Would you mind adding?

src/monitor/Monitor.ts
nicholaspai: So why not just add the `new BundleDataClient` construction to the `monitor/index.ts` file, add `BundleDataClient` to `MonitorClientHelper.ts/MonitorClients` and get rid of this construction of in this file? 

While I agree with you we should strongly consider changing this entire pattern, I still think while we have it its best to stick to the pattern we have because its easier to maintain and then refactor them all in one go

src/monitor/Monitor.ts
nicholaspai: Yes definitely can be done in separate PR

nicholaspai (commented):
Thanks for responding to my comments! Please resolve the remaining open comments and then LGTM!

src/monitor/Monitor.ts
kevinuma: Makes a lot of sense 👍 . I followed the existing pattern and moved BundleDataClient to MonitorClients and DataworkerClients.

src/dataworker/Dataworker.ts
kevinuma: Yes I added a test 👍

src/monitor/Monitor.ts
kevinuma: It's been removed 👍

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -40,8 +33,8 @@ export async function constructDataworkerClients(
       ];
     })
   );
-
-  return { ...commonClients, tokenClient, spokePoolSigners, spokePoolClientSearchSettings };
+  const bundleDataClient = new BundleDataClient(logger, commonClients, CHAIN_ID_LIST_INDICES);
```
nicholaspai, line 36: +1

nicholaspai (approved):
LGTM! thanks for following existing patterns even though I know its a pain. I think you should 100% make a PR to change the structure but I'd like to review that in a separate PR :D 

Excited to see this monitor work in prod

src/clients/HubPoolClient.ts
```diff
@@ -109,6 +109,17 @@ export class HubPoolClient {
     return this.l1TokensToDestinationTokens[l1Token][destinationChainId];
   }
 
+  getDestinationTokensToL1TokensForChainId(chainId: number) {
+    const destinationToL1Tokens: { [destinationToken: string]: L1Token } = {};
+    for (const l1Token of this.l1Tokens) {
+      const destinationToken = this.getDestinationTokenForL1TokenDestinationChainId(l1Token.address, chainId);
```
nicholaspai, line 115: I think you can make this function pretty succinct using `.map` and `Object.fromEntries`:

```js
return Object.fromEntries(this.l1Tokens.map((l1Token) => 
    [this.getDestinationTokenForL1TokenDestinationChainId(l1Token.address, chainId), l1Token]
).filter((entry) => entry[0] !== undefined))
```

src/common/Constants.ts
```diff
@@ -9,10 +10,19 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
...
+
 export const BUNDLE_END_BLOCK_BUFFERS = {
   1: 100, // At 15s/block, 100 blocks = 20 mins
   10: 3000, // At a conservative 10 TPS, 300 seconds = 3000 transactions. And 1 block per txn.
   137: 1500, // At 1s/block, 25 mins seconds = 1500 blocks
   288: 50, // At 30s/block, 50 blocks = 25 mins
   42161: 3000, // At a conservative 10 TPS, 300 seconds = 3000 transactions. And 1 block per txn.
 };
+
+export const CHAIN_NAMES = {
```
nicholaspai, line 22: Can remove this mapping in favor of using `NetworkUtils`: https://github.com/across-protocol/relayer-v2/blob/a95f98fbef56634a7a6d900c73459763514e5c9a/src/utils/NetworkUtils.ts#L3

nicholaspai (approved):
I left a couple suggestions, LGTM after resolving unless other questions come up!

.gitignore
```diff
@@ -9,3 +9,5 @@ typechain
 cache
 artifacts
 dist
+
+dump.rdb
```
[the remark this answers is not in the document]
kevinuma: Probably not. I think I just need to run redis outside of the repo.

src/common/Constants.ts
kevinuma: Done

src/clients/HubPoolClient.ts
kevinuma: Good point. Done

PR #113, opened 2022-06-03
improve: Cache pool rebalance roots for block ranges

nicholaspai:
_buildPoolRebalance root was anecdotally a slow function, so this caches the return of that for a given block range input﻿

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1,4 +1,4 @@
-import { winston, EMPTY_MERKLE_ROOT, sortEventsDescending, BigNumber, getRefund } from "../utils";
+import { winston, EMPTY_MERKLE_ROOT, sortEventsDescending, BigNumber, getRefund, MerkleTree } from "../utils";
 import {
   UnfilledDeposit,
   Deposit,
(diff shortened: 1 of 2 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1327,4 +1324,31 @@ export class Dataworker {
...
+  _getPoolRebalanceRoot(
+    blockRangesForChains: number[][],
+    endBlockForMainnet: number,
+    fillsToRefund: FillsToRefund,
+    deposits: DepositWithBlock[],
+    allValidFills: FillWithBlock[],
+    allValidFillsInRange: FillWithBlock[],
+    unfilledDeposits: UnfilledDeposit[]
+  ) {
+    const key = JSON.stringify(blockRangesForChains);
```
mrice32, line 1337: Do we need to cache any other variables in the input? Is there a concern that those others could change, but the block ranges would be identical?

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
nicholaspai: hm I guess its a lot safer to cache the entire set of inputs

mrice32 (approved):
LGTM!

PR #114, opened 2022-06-04
fix: Construct pool rebalance leaves from partially empty realizedLpFees/runningBalances

nicholaspai:
`constructPoolRebalanceLeaves()` does not currently handle case where 1 L1 token has realizedLpFees but the others are empty.

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -268,7 +268,7 @@ export function constructPoolRebalanceLeaves(
           else return toBN(0);
         });
         const leafRunningBalances = l1TokensToIncludeInThisLeaf.map((l1Token, index) => {
-          if (runningBalances[chainId] && runningBalances[chainId][l1Token])
+          if (runningBalances[chainId]?.[l1Token])
             return getRunningBalanceForL1Token(transferThresholds[index], runningBalances[chainId][l1Token]);
           else return toBN(0);
         });
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -259,7 +259,7 @@ export function constructPoolRebalanceLeaves(
         // Build leaves using running balances and realized lp fees data for l1Token + chain, or default to
         // zero if undefined.
         const leafBundleLpFees = l1TokensToIncludeInThisLeaf.map((l1Token) => {
-          if (realizedLpFees[chainId] && realizedLpFees[chainId][l1Token]) return realizedLpFees[chainId][l1Token];
+          if (realizedLpFees[chainId]?.[l1Token]) return realizedLpFees[chainId][l1Token];
           else return toBN(0);
         });
         const leafNetSendAmounts = l1TokensToIncludeInThisLeaf.map((l1Token, index) => {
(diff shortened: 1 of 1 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -256,21 +256,28 @@ export function constructPoolRebalanceLeaves(
...
+        const leafRunningBalances = l1TokensToIncludeInThisLeaf.map((l1Token, index) => {
+          if (runningBalances[chainId] && runningBalances[chainId][l1Token])
+            return getRunningBalanceForL1Token(transferThresholds[index], runningBalances[chainId][l1Token]);
+          else return toBN(0);
+        });
+
         leaves.push({
           chainId: Number(chainId),
-          bundleLpFees: realizedLpFees[chainId]
-            ? l1TokensToIncludeInThisLeaf.map((l1Token) => realizedLpFees[chainId][l1Token])
```
nicholaspai, line 262: The issue here is assuming that if `realizedLpFees[chainId]` exists, then there must be a value for each L1Token, but of course this is not true! We were getting away with it until now because all repayments for all tokens were set to L1, but once we sent fills with repayments for other chains (for one token), we saw this issue

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -256,21 +256,28 @@ export function constructPoolRebalanceLeaves(
             configStoreClient.getTokenTransferThresholdForBlock(l1Token, latestMainnetBlock)
         );
 
+        // Build leaves using running balances and realized lp fees data for l1Token + chain, or default to
+        // zero if undefined.
+        const leafBundleLpFees = l1TokensToIncludeInThisLeaf.map((l1Token) => {
+          if (realizedLpFees[chainId] && realizedLpFees[chainId][l1Token]) return realizedLpFees[chainId][l1Token];
```
mrice32, line 262:
```suggestion
          if (realizedLpFees[chainId]?.[l1Token]) return realizedLpFees[chainId][l1Token];
```

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -256,21 +256,28 @@ export function constructPoolRebalanceLeaves(
...
+          else return toBN(0);
+        });
+        const leafNetSendAmounts = l1TokensToIncludeInThisLeaf.map((l1Token, index) => {
+          if (runningBalances[chainId] && runningBalances[chainId][l1Token])
```
mrice32, line 266:
```suggestion
          if (runningBalances[chainId]?.[l1Token])
```

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -256,21 +256,28 @@ export function constructPoolRebalanceLeaves(
...
+            return getNetSendAmountForL1Token(transferThresholds[index], runningBalances[chainId][l1Token]);
+          else return toBN(0);
+        });
+        const leafRunningBalances = l1TokensToIncludeInThisLeaf.map((l1Token, index) => {
+          if (runningBalances[chainId] && runningBalances[chainId][l1Token])
```
mrice32, line 271:
```suggestion
          if (runningBalances[chainId]?.[l1Token])
```

mrice32 (approved):
LGTM!

PR #116, opened 2022-06-06
fix: Handle case where multicaller txn wasn't rejected but didn't produce transaction hash

nicholaspai:


src/clients/MultiCallerClient.ts
```diff
@@ -121,8 +121,11 @@ export class MultiCallerClient {
       const transactionHashes = [];
       Object.keys(groupedTransactions).forEach((chainId, chainIndex) => {
         mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
+
         if (transactionReceipts[chainIndex].status === "rejected") {
           mrkdwn += ` ⚠️ Transactions sent on ${getNetworkName(chainId)} failed to execute due to exceeding timeout.\n`;
+        } else if (!(transactionReceipts[chainIndex] as any).value) {
+          mrkdwn += ` ⚠️ Transactions sent on ${getNetworkName(chainId)} failed unexpectedly but not due to timeout.\n`;
```
mrice32, line 128: Do you think we should be dealing with this by catching this problem and emitting the error up here:
```ts
      // Construct multiCall transaction for each target chain.
        Object.keys(groupedTransactions).map((chainId) => this.buildMultiCallBundle(groupedTransactions[chainId]))
      );
      this.logger.debug({ at: "MultiCallerClient", message: "Waiting for bundle transaction inclusion" });
      const transactionReceipts = await Promise.allSettled(
        multiCallTransactionsResult.map((transaction) =>
          Promise.race([
            rejectAfterDelay(this.maxTxWait), // limit the maximum time to wait for a transaction receipt to mine.
            transaction ? (transaction as any)?.value?.wait() : null,
          ])
        )
      );
```

Right now it looks like we're just ignoring the case where the initial multiCallTransactionsResult is rejected, so we end up with no transaction hash.

Maybe we should try to preserve the error message so we can emit it down here rather than just logging that an unexpected error occurred?

src/clients/MultiCallerClient.ts
```diff
@@ -121,8 +121,11 @@ export class MultiCallerClient {
       const transactionHashes = [];
       Object.keys(groupedTransactions).forEach((chainId, chainIndex) => {
         mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
+
         if (transactionReceipts[chainIndex].status === "rejected") {
```
kevinuma, line 125: Just curious - is it possible to increase a retry here with some backoff (e.g. wait 10 seconds) to deal with timeout?

src/clients/MultiCallerClient.ts
```diff
@@ -157,12 +174,19 @@ export class MultiCallerClient {
         }),
         notificationPath: "across-error",
       });
-      return null; // If there is a problem in the targets in the bundle return null. This will be a noop.
+      return new Promise((_resolve, _reject) => _reject("some transactions in the bundle contain different targets"));
```
nicholaspai, line 177: This change was made so that `buildMultiCallBundle` either returns a successful `Transaction` object or throws an error.

src/utils/TransactionUtils.ts
```diff
@@ -10,8 +10,7 @@ export async function runTransaction(logger: winston.Logger, contract: Contract,
     return await contract[method](...args, gas);
   } catch (error) {
     logger.error({ at: "TxUtil", message: "Error executing tx", error, notificationPath: "across-error" });
-    console.log(error);
```
nicholaspai, line 13: is this log neccessary? I didnt think so I removed

src/clients/MultiCallerClient.ts
```diff
@@ -106,14 +107,19 @@ export class MultiCallerClient {
       const multiCallTransactionsResult = await Promise.allSettled(
         Object.keys(groupedTransactions).map((chainId) => this.buildMultiCallBundle(groupedTransactions[chainId]))
```
nicholaspai, line 108: Because of [this change](https://github.com/across-protocol/relayer-v2/pull/116/files#r891545106), `multiCallTransactionsResult` will either be `status.rejected` for any failed transactions, or be successful. It will never be `status.fulfilled` but `value == null` like it is currently

src/clients/MultiCallerClient.ts
```diff
@@ -122,12 +128,23 @@ export class MultiCallerClient {
       Object.keys(groupedTransactions).forEach((chainId, chainIndex) => {
         mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
         if (transactionReceipts[chainIndex].status === "rejected") {
-          mrkdwn += ` ⚠️ Transactions sent on ${getNetworkName(chainId)} failed to execute due to exceeding timeout.\n`;
+          const rejectionError = (transactionReceipts[chainIndex] as PromiseRejectedResult).reason;
+          mrkdwn += ` ⚠️ Transaction #${chainIndex} sent on ${getNetworkName(
+            chainId
+          )} failed or bot timed out waiting for transaction to mine, check logs for more details.\n`;
```
nicholaspai, line 134: I wasn't sure how to distinguish between `timeout` errors where the transaction actually mines but the bot timed out waiting for it to mine, versus other errors like nonce-collisions, or gas estimation issues or just general RPC request issues so I combined them all into anything that ends up here with `status ==rejected`

src/clients/MultiCallerClient.ts
nicholaspai: Good point, I refactored can you take another look and give thoughts?

src/clients/MultiCallerClient.ts
```diff
@@ -106,14 +107,19 @@ export class MultiCallerClient {
...
-            transaction ? (transaction as any)?.value?.wait() : null,
-          ])
-        )
+        multiCallTransactionsResult.map((transaction) => {
+          if (transaction.status !== "rejected") {
+            return Promise.race([
+              rejectAfterDelay(this.maxTxWait), // limit the maximum time to wait for a transaction receipt to mine.
+              (transaction as any).value.wait(),
+            ]);
+          } else return new Promise((_resolve, _reject) => _reject(transaction.reason));
```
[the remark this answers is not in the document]
mrice32:
```suggestion
          } else return Promise.reject(transaction.reason);
```

src/clients/MultiCallerClient.ts
mrice32, line 177:
```suggestion
      return Promise.reject("some transactions in the bundle contain different targets");
```

src/clients/MultiCallerClient.ts
mrice32: I think what you have here is totally fine! Logging-wise, the error should explain everything.

If you really wanted to distinguish, I think the cleanest way would be to filter out the rejections after the first loop and put them into a separate array that we could print out separately. But that seems like overkill IMO.

src/clients/MultiCallerClient.ts
```diff
@@ -106,14 +107,19 @@ export class MultiCallerClient {
...
-          Promise.race([
-            rejectAfterDelay(this.maxTxWait), // limit the maximum time to wait for a transaction receipt to mine.
-            transaction ? (transaction as any)?.value?.wait() : null,
-          ])
-        )
+        multiCallTransactionsResult.map((transaction) => {
+          if (transaction.status !== "rejected") {
+            return Promise.race([
+              rejectAfterDelay(this.maxTxWait), // limit the maximum time to wait for a transaction receipt to mine.
+              (transaction as any).value.wait(),
```
mrice32, line 119: Odd to me that this type isn't resolving correctly, but we can figure that out later.

mrice32 (approved):
LGTM!

PR #119, opened 2022-06-06
feat: add tsc -b to github ci

nicholaspai:
This PR is mainly to prevent cases where `test` action in CI succeeds but `build` fails in GCP because compiling fails in untested `ts` files like `scripts/*`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,5 +21,8 @@ jobs:
       - name: Install deps and build (with cache)
         uses: bahmutov/npm-install@v1
 
+      - name: Compile
+        run: tsc -b
+
       - name: Test
         run: yarn test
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -21,5 +21,8 @@ jobs:
       - name: Install deps and build (with cache)
         uses: bahmutov/npm-install@v1
 
+      - name: Compile
+        run: tsc -b
```
mrice32, line 25: Why `tsc -b`? Why not `yarn build` since that's what we typically run to build the repo?

mrice32 (approved):
Good idea!

.github/workflows/test.yml
nicholaspai: great point , fixed

PR #121, opened 2022-06-08
feat: add resilient, quorum provider

mrice32:


src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
 
 function delay(s: number): Promise<void> {
-  return new Promise<void>((resolve) => setTimeout(resolve, s * 1000));
+  return new Promise<void>((resolve) => setTimeout(resolve, Math.round(s * 1000)));
+}
+
+function isPromiseFulfulled<T>(
+  promiseSettledResult: PromiseSettledResult<T>
+): promiseSettledResult is PromiseFulfilledResult<T> {
+  return promiseSettledResult.status === "fulfilled";
```
nicholaspai, line 13: We should use this helper function in `MulticallerClient`!

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
+    chainId: number,
+    readonly nodeQuorumThreshold: number,
+    readonly retries: number,
+    readonly delay: number
+  ) {
+    // Initialize the super just with the chainId, which stops it from trying to immediately send out a .send before
+    // this derived class is initialized.
+    super(undefined, chainId);
+    this.providers = params.map((inputs) => new ethers.providers.JsonRpcProvider(...inputs));
+    if (this.nodeQuorumThreshold < 1 || !Number.isInteger(this.nodeQuorumThreshold))
```
nicholaspai, line 33: Do you think this sanity checking of config vars belongs in the `Config` files?

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
+  return new RetryProvider(constructorArgumentLists, chainId, nodeQuorumThreshold, retries, retryDelay);
 }
 
-// Returns a normal json RPC provider.
-export function getStandardProvider(networkId: number) {
-  const nodeUrl = process.env[`NODE_URL_${networkId}`];
-  if (!nodeUrl) throw new Error(`No NODE_URL_ for network ${networkId}`);
-  return new RetryProvider({ url: nodeUrl, timeout: stallTimeout });
+export function getNodeUrlList(chainId: number): string[] {
+  const retryConfigKey = `RETRY_CONFIG_${chainId}`;
```
nicholaspai, line 149: Maybe rename this to `RETRY_URLS_`

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
+      );
     }
 
+    const values = results.map((result) => result.value);
+
+    // Start at element 1 and begin comparing.
+    // If _all_ values are equal, we have hit quorum, so return.
+    if (values.slice(1).every(([, output]) => deepEqual(values[0][1], output))) return values[0][1];
+
+    // Iteratively add results until we have enough matching to return.
```
nicholaspai, line 92:
```suggestion
    // At this point, the required providers failed to reach quorum, so we need to use the fallback providers. 
    // Iteratively add results until we have enough matching to return.
```

nicholaspai: Are these comments accurate?

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
+    while (true) {
+      // Grab a new result from our fallback providers.
+      // If we run out of fallback providers, throw a useful error listing all of the successful and failed providers.
+      const [provider, newOutput] = await tryWithFallback(fallbackProviders.shift()).catch(() => {
```
nicholaspai, line 96: I'm a bit confused here. So you call `tryWithFallback` with one of the `fallbackProviders`. But, say `fallbackProviders.length == 0` so you're passing in an undefined object to `tryWithFallback`. It doesn't look like `tryWithFallback` will handle an `undefined` input correctly. Moreover, its going to recursively call itself with another `undefined` input.

nicholaspai (commented):
I love this PR, the high level flow makes sense. I have some questions on implementation

src/utils/ProviderUtils.ts
mrice32: You're totally right. That check needs to be added.

src/utils/ProviderUtils.ts
chrismaree: ye, it should be pulled out into another helper file called `PromiseUtils`, and export it from `index.ts` from `utils` to best re-use it.

src/utils/ProviderUtils.ts
chrismaree: I think it was done like this to be compatible with the existing variable naming done in protocol and other bot configs.

src/utils/ProviderUtils.ts
mrice32: I think it belongs here, since we could have other paths to initializing this object. If these checks fail, this class won't work properly, so I think this is the right layer for these checks to ensure there's no initialization path that causes strange behavior.

Thoughts?

package.json
```diff
@@ -21,6 +21,7 @@
     "@uma/contracts-node": "^0.3.1",
     "@uma/financial-templates-lib": "^2.27.4",
     "bluebird": "^3.7.2",
+    "deep-equal": "^2.0.5",
```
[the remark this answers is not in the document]
mrice32: Oh, good point. Will use lodash instead.

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
   );
+
+  // Default to 2 retries.
+  const retries = Number(process.env[`NODE_RETRIES_${chainId}`] || NODE_RETRIES || "2");
+
+  // Default to a delay of 1 second between retries.
+  const retryDelay = Number(process.env[`NODE_RETRY_DELAY_${chainId}`] || NODE_RETRY_DELAY || "1");
+
+  // Default to a node quorum of 1 node.
+  const nodeQuorumThreshold = Number(process.env[`NODE_QUORUM_${chainId}`] || NODE_QUORUM || "1");
```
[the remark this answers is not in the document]
mrice32: Maybe, but personally, I find it hard/annoying to stuff JSON into env strings. Thoughts?

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,166 @@
...
     }
 
+    const values = results.map((result) => result.value);
+
+    // Start at element 1 and begin comparing.
+    // If _all_ values are equal, we have hit quorum, so return.
+    if (values.slice(1).every(([, output]) => deepEqual(values[0][1], output))) return values[0][1];
+
+    // Iteratively add results until we have enough matching to return.
+    while (true) {
```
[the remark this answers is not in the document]
mrice32: 1) Good point. I will look into this.

2) We can do that instead. I think the code will be similar in complexity. My current approach was trying to minimize how many requests we send out. Do you think we should optimize for speed or request efficiency in the failure case. My assumption was request efficiency, but open to other opinions.

src/utils/ProviderUtils.ts
mrice32: Done. I named it TypeGuards.ts since I think this is just a simple custom type guard that has more in common with other type guards than other promise utilities.

src/utils/ProviderUtils.ts

src/utils/ProviderUtils.ts
nicholaspai: Fair points, can ignore!

src/utils/ProviderUtils.ts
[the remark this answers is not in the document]
nicholaspai: I agree its not worth solving.

On 2: I think we should optimize for speed but i'm not sure if my instinct is correct. I think requests in general are a rarer resource than speed

@chrismaree wdyt?

nicholaspai (approved):
LGTM after resolving outstanding issues, which are both nits:
- Possibly rename `RETRY_CONFIG` to `NODE_URLS`, or if the latter is too confusing and close with `NODE_URL` then use `FALLBACK_URLS`?
- Consider whether to try all fallback providers in parallel or not. This might cause more errors to be thrown longer term, so I probably lean towards keeping as is.

src/utils/ProviderUtils.ts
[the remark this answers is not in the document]
mrice32: Done, NODE_URLS_ it is.

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,215 @@
...
+      promise = promise.catch(() => delay(this.delay).then(() => super.send(method, params)));
+    }
     return promise;
   }
-}
 
-export function getProvider(networkId: number, nodeQuorumThreshold: number = 1) {
-  if (process.env[`RETRY_CONFIG_${networkId}`]) return getFallbackProvider(networkId, nodeQuorumThreshold);
-  else return getStandardProvider(networkId);
+  _getQuorum(method: string, params: Array<any>): number {
```
mrice32, line 159: Note: this is where we consider whether we should use the configured quorum or just use 1. We use 1 for most queries, as most queries depend on the sync of the node and probably aren't as sensitive for our bots.

mrice32:
> LGTM after resolving outstanding issues, which are both nits:
> 
> * Possibly rename `RETRY_CONFIG` to `NODE_URLS`, or if the latter is too confusing and close with `NODE_URL` then use `FALLBACK_URLS`?
> * Consider whether to try all fallback providers in parallel or not. This might cause more errors to be thrown longer term, so I probably lean towards keeping as is.

Done and done.

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,216 @@
...
+          // Otherwise, we can try to call another provider.
+          if (fallbackProviders.length === 0) {
+            throw new Error("Out of providers to fall back to.");
+          }
+
+          // This line does two things:
+          // 1. Removes a fallback provider from the array so it cannot be used as a fallback for another required
+          // provider.
+          // 2. Recursively calls this method with that provider so it goes through the same try logic as the previous one.
+          return tryWithFallback(fallbackProviders.shift());
```
kevinuma, line 70: If multiple required providers fail, would all the retry traffic then flow to the same (first) fallback provider? I think this can potential cause cascading timeouts or ratelimit-related failures.

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,216 @@
...
+        `Not enough providers succeeded. Errors:\n${errorTexts.join("\n")}\n` +
+          `Successful Providers:\n${successfulProviderUrls.join("\n")}`
+      );
     }
 
+    const values = results.map((result) => result.value);
+
+    // Start at element 1 and begin comparing.
+    // If _all_ values are equal, we have hit quorum, so return.
+    if (values.slice(1).every(([, output]) => lodash.isEqual(values[0][1], output))) return values[0][1];
```
kevinuma, line 90: What's in the value? Can there be timestamps that would lead to highly probable diffs?

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,216 @@
...
-}
 
-export function getProvider(networkId: number, nodeQuorumThreshold: number = 1) {
-  if (process.env[`RETRY_CONFIG_${networkId}`]) return getFallbackProvider(networkId, nodeQuorumThreshold);
-  else return getStandardProvider(networkId);
+  _getQuorum(method: string, params: Array<any>): number {
+    // Only use quorum if this is a historical query that doesn't depend on the current block number.
+
+    // All logs queries should use quorum.
+    if (method === "eth_getLogs") return this.nodeQuorumThreshold;
```
nicholaspai, line 164: 100%

src/utils/ProviderUtils.ts
nicholaspai: ah i see you address this potential variability in a request by request fashion here: https://github.com/across-protocol/relayer-v2/pull/121/files#r893901145

nicholaspai (commented):
Left some questions!

src/utils/ProviderUtils.ts
```diff
@@ -1,50 +1,216 @@
...
+
+    // getBlockByNumber should only use the quorum if it's not asking for the latest block.
+    if (method === "eth_getBlockByNumber" && params[0] !== "latest") return this.nodeQuorumThreshold;
```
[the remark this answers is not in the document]
mrice32: I think that's a little complex IMO. I think if we specify a block, we should probably be sure that all providers have that block -- this already causes issues in vanilla infura since one request might be load balanced to a node that has that block and the next may be load balanced to one that doesn't. So I think the caller should always be maintaining a little distance from HEAD when making block-based queries to avoid spurious errors. Thoughts?

src/utils/ProviderUtils.ts
mrice32: I think there are two ways to interpret your question:
1. For a particular request, let's say we have three providers: `Provider1`, `Provider2`, and `Provider3`. If both `Provider1` and `Provider2` fail, will `Provider3` be requested twice? The answer is no. That's why we call `.shift()`. We remove that fallback provider from the array to ensure it cannot be used again. Whichever provider hits this part of the code last will fail on the check just above this.
2. Across many requests, if some providers are consistently failing, will we _always_ fallback hit the same fallback provider? Meaning that if the first provider dies due to load, then this failure may cascade to the first fallback provider, and so on. The answer is yes. We do not randomize the order of fallbacks. This is intentional, but it does have tradeoffs. The advantage is that we have control over where our load is concentrated. We can send 99% of our load to provider 1 and then only have the failures fall back to another provider, thereby making our costs/load very predictable and concentrated. The disadvantage is that load-related failures are more likely to cascade than if the nodes were selected randomly and we load balanced. I think this is the right tradeoff given our (relatively) light load when compared with the scale of the providers we use.

src/utils/ProviderUtils.ts
mrice32: Yeah, we only have a > 1 quorum threshold for queries that should not have variability based on timestamps or latest block. Does that answer your question @kevinuma?

src/utils/ProviderUtils.ts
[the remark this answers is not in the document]
nicholaspai: Fair point

src/utils/ProviderUtils.ts
kevinuma: Yes my question is more about the second part but the tradeoff makes sense

kevinuma (approved):

PR #126, opened 2022-06-12
feat: Report categorized non-v2 transfers from relayers

kevinuma:
Currently the monitoring report only takes into account Across V2 relays/refunds. This leads to unexplained changes in relayer balances from to time due to:

1. Interactions with Across V1 (relays, refunds)
2. Automated and manual inventory management.
3. V2 bond payments/refunds.

This PR adds logic to report the above interactions.

Notes: This can create noise when the monitor bot runs in non-looping mode (POLLING_DELAY = 0) because if there are mismatching v1 relays/refunds for example in the 30m window (current monitoring bot interval), the net transfer balance within that window would not be 0 (we lost some v1 relays or refunds from before the window). This problem should go away once we switch to looping mode as the TokenTransferClient's transfer cache would then contain all transfers since the start of the bot run.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -68,7 +68,13 @@ export class SpokePoolClient {
   }
 
   getFillsWithBlockForOriginChain(originChainId: number): FillWithBlock[] {
-    return this.fillsWithBlockNumbers.filter((fill: Fill) => fill.originChainId === originChainId);
+    return this.fillsWithBlockNumbers.filter((fill: FillWithBlock) => fill.originChainId === originChainId);
+  }
+
+  getFillsWithBlockForDestinationChainAndRelayer(chainId: number, relayer: string): FillWithBlock[] {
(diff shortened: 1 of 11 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -68,7 +68,13 @@ export class SpokePoolClient {
   }
 
   getFillsWithBlockForOriginChain(originChainId: number): FillWithBlock[] {
-    return this.fillsWithBlockNumbers.filter((fill: Fill) => fill.originChainId === originChainId);
```
kevinuma, line 71: fillsWithBlockNumbers is an array of FillWithBlock's so technically casting to Fill seems strange.

nicholaspai: Its not really strange as `FillWithBlock` extends `Fill`, what's the issue here?

kevinuma: Not an issue. I'm just being OCD :)

PR #129, opened 2022-06-13
feat: Consider refunds from pending and next bundles in inventory management

kevinuma:
Context: When deciding which chain to get refunded on when filling a deposit, we need to consider the refunds from pending and next bundles for both the destination chain and L1.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -1,9 +1,10 @@
-import { winston } from "../utils";
+import { winston, BigNumber } from "../utils";
 import {
   Deposit,
   DepositWithBlock,
   FillsToRefund,
   FillWithBlock,
+  ProposedRootBundle,
   UnfilledDeposit,
   UnfilledDepositsForOriginChain,
 } from "../interfaces";
@@ -37,6 +38,7 @@ export class BundleDataClient {
   constructor(
     readonly logger: winston.Logger,
     readonly clients: Clients,
+    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly chainIdListForBundleEvaluationBlockNumbers: number[]
   ) {}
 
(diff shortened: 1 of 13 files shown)


src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -235,44 +249,41 @@ export class InventoryClient {
       // Construct logs on the cross-chain actions executed.
       let mrkdwn = "";
 
-      if (possibleRebalances != {}) {
```
kevinuma, line 238: Removed unnecessary if. It's also incorrect as javascript does reference comparison when comparing objects so != {} would always return true anyway.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -235,44 +249,41 @@ export class InventoryClient {
...
+            `${formatWei(toBN(targetPct).mul(100).toString())}% (trigger of ` +
+            `${formatWei(toBN(thresholdPct).mul(100).toString())}%) of the total ` +
+            `${formatter(
+              this.getCumulativeBalance(l1Token).toString()
+            )} ${symbol} over all chains (ignoring hubpool repayments). This chain has a shortfall of ` +
+            `${formatter(this.getTokenShortFall(l1Token, chainId).toString())} ${symbol} ` +
+            `tx: ${etherscanLink(executedTransactions[chainId][l1Token], 1)}\n`;
         }
       }
-      if (unexecutedRebalances != {}) {
```
kevinuma, line 259: Same as previous note.

src/clients/BundleDataClient.ts
```diff
@@ -46,6 +47,50 @@ export class BundleDataClient {
...
+    const latestBundle = hubPoolClient.getMostRecentProposedRootBundle(latestBlockNumber);
+    if (hubPoolClient.getExecutedLeavesForRootBundle(latestBundle, latestBlockNumber).length == 0) {
+      // Use the same block range as the current pending bundle to ensure we look at the right refunds.
+      const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map(
+        (endBlock, i) => [
+          hubPoolClient.getNextBundleStartBlockNumber(chainIds, latestBlockNumber, chainIds[i]),
+          endBlock.toNumber(),
+        ]
+      );
+      refunds = this.loadData(pendingBundleEvaluationBlockRanges, this.spokePoolClients, false).fillsToRefund;
```
nicholaspai, line 67:
```suggestion
      return this.loadData(pendingBundleEvaluationBlockRanges, this.spokePoolClients, false).fillsToRefund;
```

src/clients/BundleDataClient.ts
```diff
@@ -46,6 +47,50 @@ export class BundleDataClient {
...
+    }
```
nicholaspai, line 68:
```suggestion
    } else return {}
```

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -121,22 +123,34 @@ export class InventoryClient {
...
+
+    let l2Token = this.getDestinationTokenForL1Token(l1Token, destinationChainId);
+    // Consider any refunds from both current pending bundle (if any) and the next bundle.
+    let pendingBundleRefunds = this.bundleDataClient.getPendingBundleRefunds();
+    let nextBundleRefunds = this.bundleDataClient.getNextBundleRefunds();
+    chainVirtualBalanceWithShortfall = chainVirtualBalanceWithShortfall
+      .add(this.bundleDataClient.getRefundsFor(pendingBundleRefunds, this.relayer, destinationChainId, l2Token))
+      .add(this.bundleDataClient.getRefundsFor(nextBundleRefunds, this.relayer, destinationChainId, l2Token));
+    cumulativeVirtualBalanceWithShortfall = cumulativeVirtualBalanceWithShortfall
+      .add(this.bundleDataClient.getRefundsFor(pendingBundleRefunds, this.relayer, 1, l1Token))
```
nicholaspai, line 152: Why are you adding repayments for chainId 1?

src/clients/InventoryClient/InventoryClient.ts
nicholaspai: Yes better to do Object.keys(possibleRebalances).length == 0

nicholaspai (commented):
This all looks good and Ive left some minor suggestions!

src/clients/InventoryClient/InventoryClient.ts
kevinuma: It's actually not necessary since the loop goes through Object.keys anyway so if it's empty, the loop would not even execute.

src/clients/InventoryClient/InventoryClient.ts
kevinuma: Yeah this is not quite right. I need to add all refunds for equivalent of l1 token on all chains. This is needed to correctly compute the allocation % of destination chain post the relay, shortfall, and upcoming refunds.

src/clients/BundleDataClient.ts
kevinuma: Updated

src/clients/BundleDataClient.ts
kevinuma: Updated

src/clients/BundleDataClient.ts
```diff
@@ -46,6 +47,50 @@ export class BundleDataClient {
...
+
+    return refunds;
```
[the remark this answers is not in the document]
kevinuma: Updated

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -121,22 +123,44 @@ export class InventoryClient {
...
+    // Add upcoming refunds going to this destination chain.
+    const destinationChainRefunds = this.bundleDataClient
+      .getRefundsFor(pendingBundleRefunds, this.relayer, destinationChainId, l2Token)
+      .add(this.bundleDataClient.getRefundsFor(nextBundleRefunds, this.relayer, destinationChainId, l2Token));
+    chainVirtualBalanceWithShortfallPostRelay = chainVirtualBalanceWithShortfallPostRelay.add(destinationChainRefunds);
+    // To correctly compute the allocation % for this destination chain, we need to add all upcoming refunds for the
+    // equivalents of l1Token on all chains.
+    const cumulativeRefunds = this.getEnabledChains()
+      .map((chainId) => [this.getDestinationTokenForL1Token(l1Token, chainId), chainId])
+      .map(([tokenId, chainId]) =>
```
nicholaspai, line 157: Doesn't this double count the `destinationChainRefunds` for `l2Token` on destination chain?

nicholaspai (commented):
This looks great! See my comment on potential double counting some refunds

src/clients/InventoryClient/InventoryClient.ts
kevinuma: I think we want to include that here right? This is similar to cumulative balances where we do include the destination chain's balance as well. This ensures that the allocation % is calculated correctly - current balance + refunds on destination chain / all balances + refunds on all chains

nicholaspai: I feel like I'm missing something here. Why do you compute `destinationChainRefunds` above and then compute the same thing for all chains below? Why not just do the bottom calculation and then grab the entry for `destinationChainId` above and add it to `chainVirtualBalanceWithShortfallPostRelay`?

nicholaspai (approved):
I left one comment!

src/clients/InventoryClient/InventoryClient.ts
kevinuma: We can but it's not much less code I think. I would need to turn the following into an object mapping chain id => refunds then later use Object.values(...).reduce()

this.getEnabledChains()
      .map((chainId) => [this.getDestinationTokenForL1Token(l1Token, chainId), chainId])
      .map(([tokenId, chainId]) =>
        this.bundleDataClient
          .getRefundsFor(pendingBundleRefunds, this.relayer, Number(chainId), String(tokenId))
          .add(this.bundleDataClient.getRefundsFor(nextBundleRefunds, this.relayer, Number(chainId), String(tokenId)))
      )

nicholaspai: Hm I just don't like re-used logic. Its not always about the lines of code, but the fact that the exact same code is re-used which could mean in the future if we update this function, we might forget to update two instances of the same logic. So we should try to consolidate. Does that make sense?

kevinuma: Updated.

PR #132, opened 2022-06-13
fix: Fix pending refunds calculation in relayer report

kevinuma:
Current pending refunds calculation only checks if the bundle has been executed in HubPool (rebalance tree has been executed). This doesn't mean the actual refunds (leaves of relay refund tree) have been executed on destination chains.

### DIFF
--- a/src/interfaces/HubPool.ts
+++ b/src/interfaces/HubPool.ts
@@ -31,16 +31,6 @@ export interface ProposedRootBundle extends SortableEvent {
 }
 
 export interface CancelledRootBundle extends SortableEvent {
-  chainId: number;
-  bundleLpFees: BigNumber[];
-  netSendAmounts: BigNumber[];
-  runningBalances: BigNumber[];
-  leafId: number;
-  l1Tokens: string[];
-  proof: string[];
-}
-
-export interface CancelledRootBundle extends SortableEvent {
   disputer: string;
   requestTime: number;
 }
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -1,4 +1,4 @@
-import { FillsToRefund, TokenTransfer } from "../interfaces";
+import { FillsToRefund, ProposedRootBundle, TokenTransfer } from "../interfaces";
 import { BundleAction, BalanceType, L1Token, RelayerBalanceTable, RelayerBalanceReport } from "../interfaces";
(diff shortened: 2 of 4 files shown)


src/interfaces/HubPool.ts
```diff
@@ -30,16 +30,6 @@ export interface ProposedRootBundle extends SortableEvent {
   proposer: string;
 }
 
-export interface CancelledRootBundle extends SortableEvent {
```
kevinuma, line 33: Duplicate definition with that on line 43

src/monitor/Monitor.ts
```diff
@@ -473,15 +469,37 @@ export class Monitor {
     return reports;
   }
 
+  private deductExecutedRefunds(allRefunds: FillsToRefund, latestBundle: ProposedRootBundle) {
+    for (const chainIdStr of Object.keys(allRefunds)) {
+      const chainId = Number(chainIdStr);
+      const executedRefunds = this.clients.spokePoolClients[chainId].getExecutedRefunds(latestBundle.relayerRefundRoot);
+
+      for (const tokenAddress of Object.keys(allRefunds[chainId])) {
+        if (executedRefunds[tokenAddress] === undefined || allRefunds[chainId][tokenAddress].refunds === undefined)
```
nicholaspai, line 478:
```suggestion
        if (executedRefunds[tokenAddress]?.refunds === undefined)
```

i think this should work

src/monitor/Monitor.ts
```diff
@@ -473,15 +469,37 @@ export class Monitor {
...
+          continue;
+
+        const refunds = allRefunds[chainId][tokenAddress].refunds;
+        for (const relayer of Object.keys(refunds)) {
+          const executedAmount = executedRefunds[tokenAddress][relayer];
+          if (executedAmount === undefined) continue;
+
+          if (executedAmount.gt(refunds[relayer])) {
+            throw new Error(
```
nicholaspai, line 487: Is this error suggesting a programmer error?

src/monitor/Monitor.ts
kevinuma: This condition is checking two different maps though - executedRefunds[...] and allRefunds[...]

src/monitor/Monitor.ts
kevinuma: Likely. The way we extract these refund information is very complex. There might be bugs that affect some edge cases.

PR #135, opened 2022-06-14
feat: Truncate realized lp fee %'s after a hardcoded quote time

nicholaspai:
~This PR should not be merged in until [the UMIP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md)  is changed to enforce realized lp fee % truncation.~

EDIT: This [UMIP PR](https://github.com/UMAprotocol/UMIPs/pull/565) is ready to be reviewed now and once it is merged, this PR can be merged.

This PR once merged will allow the relayer to truncate fees after a certain hardcoded time. This PR could be merged safely today because the hardcoded time is set so far in the future that the truncation feature will never be enforced.


mrice32 (commented):
I'm not going to approve until we're ready to merge this, just to avoid a similar confusion to last time.

james-a-morris:
@mrice32 cc @nicholaspai Is this still a valid PR or should we move to close?

nicholaspai:
Woo V3 makes this moot

PR #137, opened 2022-06-14
feat: Add startingRunningBalance to TokenConfig

nicholaspai:
Allows admin to “bias” the running balances on a chain for a particular L1 token. This would allow admin to start all running balances for a particular SpokePool for a particular L1 token at a certain value. For example, we want WETH on Optimism to start with 100 WETH running balance. This would mean that if 50 WETH of deposits occurred during a root bundle period, then the HubPool would send 50 WETH (100 - 50 = +50) to the OptimismSpokePool. Currently, the system would return 50 WETH (0 - 50 = -50) WETH from the OptimismSpokePool to the HubPool.

The goal of this feature is to reduce gas costs for the dataworker (the one executing pool rebalance leaves) which would have to send fewer tokens from L1 --> L2 to pay refunds

### DIFF
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -9,13 +9,21 @@ import {
   EventSearchConfig,
   utf8ToHex,
   getCurrentTime,
+  toBN,
 } from "../utils";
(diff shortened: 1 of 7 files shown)


mrice32:
Closing in favor of https://github.com/across-protocol/relayer-v2/pull/301. @nicholaspai let me know if you disagree!

PR #138, opened 2022-06-14
feat: add a provider that rate limits the number of in-flight requests

mrice32:
This adds a provider that rate limits how many requests can be in-flight for any provider at any time. The default setting is 1000, which means that we shouldn't run out of ports unless we're concurrently trying to use 15 providers at max capacity. We can definitely tune this if we run into issues.

As with other provider settings, it can be adjusted per chain (`NODE_MAX_CONCURRENCY_[chainId]=X`) and a global default can be set (`NODE_MAX_CONCURRENCY=X`).

This provider sits _under_ the retry provider. This means that if the retry provider is concurrently hitting 5 different nodes, _each_ of those nodes could independently have the max concurrent number of requests in flight. This is important to consider since the retry provider often hits the required nodes concurrently (and if one fails, it hits the fallback nodes concurrently too).

Note: this has been tested locally to ensure that 1) the max concurrency setting actually works and 2) that this doesn't break anything.

I was able to use this to spin up the entire cache from scratch (taking a few minutes) without the relayer throwing any errors.

Even with this limited to 10, the relayer was able to complete a run in a minute and a half.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -5,8 +5,6 @@ import { AcrossConfigStoreClient } from "./ConfigStoreClient";
 import { Deposit, DepositWithBlock, Fill, SpeedUp, FillWithBlock, TokensBridged } from "../interfaces/SpokePool";
 import { RootBundleRelayWithBlock, RelayerRefundExecutionWithBlock } from "../interfaces/SpokePool";
 
-import bluebird from "bluebird";
-
 export class SpokePoolClient {
   private deposits: { [DestinationChainId: number]: Deposit[] } = {};
   private fills: Fill[] = [];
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -1,6 +1,56 @@
...
+  // Takes the same arguments as the JsonRpcProvider, but it has an additional maxConcurrency value at the beginning
+  // of the list.
+  constructor(
+    maxConcurrency: number,
+    ...jsonRpcConstructorParams: ConstructorParameters<typeof ethers.providers.JsonRpcProvider>
+  ) {
+    super(...jsonRpcConstructorParams);
+
+    // This sets up the queue. Each task is executed by calling the superclass's send method, which fires off the
+    // request. The maxConcurrency is configured here.
```
nicholaspai, line 33: Can you add a comment explaining that all tasks in queue are executed in parallel up to `maxConcurrency`? For some reason when I think of queue I think its a sequential data structure

src/utils/ProviderUtils.ts
```diff
@@ -192,7 +243,17 @@ export function getProvider(chainId: number) {
   // Default to a node quorum of 1 node.
   const nodeQuorumThreshold = Number(process.env[`NODE_QUORUM_${chainId}`] || NODE_QUORUM || "1");
 
-  return new RetryProvider(constructorArgumentLists, chainId, nodeQuorumThreshold, retries, retryDelay);
+  // Default to a max concurrency of 50 requests per node.
```
nicholaspai, line 246: 50? should this be 1000?

nicholaspai (approved):
LGTM, this async/queue class is pretty awesome

src/utils/ProviderUtils.ts
mrice32: Yes, sorry, I originally had it set to 50, but it was really slow.

src/utils/ProviderUtils.ts
mrice32: Can do!

src/utils/ProviderUtils.ts

src/utils/ProviderUtils.ts

nicholaspai (approved):

PR #140, opened 2022-06-15
feat: dataworker should send ETH to hub pool when executing arb transactions

mrice32:
This adds a way to include a value to a transaction sent to the multicaller client. Any value transactions are sent (in parallel) _before_ the multicall batches are sent.


nicholaspai (approved):
LGTM

PR #142, opened 2022-06-16
feat: Consider upcoming refunds when calculating relayer fund rebalancing

kevinuma:
Context: Currently inventory management can oversend funds to meet a shortfall. Consider the following scenario:

1. Polygon currently has a balance of 5 WETH with 20 WETH coming. The target allocation is 25 WETH so once refunds come in a few hours, the target allocation is still met.
2. There's a deposit of 10 WETH that needs to be filled on Polygon. This is a shortfall as we currently only have 5 WETH standing.
3. Currently inventory management would send 20 WETH total (triggered by the shortfall) to also meet the target alloc of 25 WETH. This means that after refunds, the final balance would be 45 WETH, which is almost double the target alloc.
4. This PR updates the logic so that it would only send 10 WETH over (exactly enough to cover the shortfall deposit). This will get refunded on L1 later as the repayment logic already knows of the 20 WETH coming. At the end, Polygon would only have 25 WETH remaining, which is perfectly at target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ typechain
 dump.rdb
 .idea
 
-#Hardhat files
+# Hardhat files
 cache
 artifacts
 dist
--- a/src/clients/InventoryClient/InventoryClient.ts
+++ b/src/clients/InventoryClient/InventoryClient.ts
@@ -64,19 +64,21 @@ export class InventoryClient {
     return distributionPerL1Token;
   }
 
-  // Get the allocation of a given token, considering the shortfall. This number will be negative if there is a shortfall!
-  // A shortfall, by definition, is the amount of tokens that are needed to to fill a given outstanding relay that the
-  // relayer could not fill. For example of the relayer has 10 WETH and needs to fill a relay of size 18 WETH then the
-  // shortfall will be 18. If there are 140 WETH across the whole ecosystem then this method will return (10-18)/140=-0.057
-  // This number is then used to inform how many tokens need to be sent to the chain to: a) cover the shortfall and
-  // b bring the balance back over the defined target.
-  getCurrentAllocationPctConsideringShortfall(l1Token: string, chainId: number): BigNumber {
-    const cumulativeBalance = this.getCumulativeBalance(l1Token); // If there is nothing over all chains, return early.
+  // Get the balance of a given token on a given chain, including shortfalls, any pending cross chain transfers and
+  // upcoming refunds.
+  getCurrentAllocationPct(l1Token: string, chainId: number): BigNumber {
+    const totalRefundsPerChain = this.getBundleRefunds(l1Token);
+    const cumulativeRefunds = Object.values(totalRefundsPerChain).reduce((acc, curr) => acc.add(curr), toBN(0));
+    // If there is nothing over all chains, return early.
+    const cumulativeBalance = this.getCumulativeBalance(l1Token).add(cumulativeRefunds);
     if (cumulativeBalance.eq(0)) return toBN(0);
-    const currentBalanceConsideringTransfers = this.getBalanceOnChainForL1Token(chainId, l1Token);
+
     const shortfall = this.getTokenShortFall(l1Token, chainId) || toBN(0);
-    const numerator = currentBalanceConsideringTransfers.sub(shortfall).mul(scalar);
-    return numerator.div(cumulativeBalance);
+    const currentBalance = this.getBalanceOnChainForL1Token(chainId, l1Token)
+      .add(totalRefundsPerChain[chainId])
+      .sub(shortfall);
+    // Multiply by scalar to avoid rounding errors.
+    return currentBalance.mul(scalar).div(cumulativeBalance);
   }
 
   // Find how short a given chain is for a desired L1Token.
@@ -113,35 +115,8 @@ export class InventoryClient {
       : rebalance;
   }
 
-  // Work out where a relay should be refunded to optimally manage the bots inventory. If the inventory management logic
-  // not enabled then return funds on the chain the deposit was filled on Else, use the following algorithm:
-  // a) Find the chain virtual balance (current balance + pending relays) minus the current shortfall.
-  // b) find the cumulative virtual balance minus the current shortfall.
-  // c) consider the size of a and b post relay (i.e after the relay is paid and all current transfers are settled what
-  // will the balances be on the target chain and the overall cumulative balance).
-  // d) use c to compute what the post relay post current in-flight transactions allocation would be. Compare this
-  // number to the target threshold and:
-  //     If this number of more than the target for the designation chain + rebalance overshoot then refund on L1.
-  //     Else, the post fill amount is within the target, so refund on the destination chain.
-  determineRefundChainId(deposit: Deposit): number {
-    const destinationChainId = deposit.destinationChainId;
-    if (!this.isInventoryManagementEnabled()) return destinationChainId;
-    if (destinationChainId === 1) return 1; // Always refund on L1 if the transfer is to L1.
-    const l1Token = this.hubPoolClient.getL1TokenForDeposit(deposit);
-
-    // If there is no inventory config for this token or this token and destination chain the return the destination chain.
-    if (
-      this.inventoryConfig.tokenConfig[l1Token] == undefined ||
-      this.inventoryConfig.tokenConfig?.[l1Token]?.[destinationChainId] == undefined
-    )
-      return destinationChainId;
-    const chainShortfall = this.getTokenShortFall(l1Token, destinationChainId);
-    const chainVirtualBalance = this.getBalanceOnChainForL1Token(destinationChainId, l1Token);
-    const chainVirtualBalanceWithShortfall = chainVirtualBalance.sub(chainShortfall);
-    let chainVirtualBalanceWithShortfallPostRelay = chainVirtualBalanceWithShortfall.sub(deposit.amount);
-    const cumulativeVirtualBalance = this.getCumulativeBalance(l1Token);
-    let cumulativeVirtualBalanceWithShortfall = cumulativeVirtualBalance.sub(chainShortfall);
-
+  // Return the upcoming refunds (in pending and next bundles) on each chain.
+  getBundleRefunds(l1Token: string): { [chainId: string]: BigNumber } {
     // Increase virtual balance by pending relayer refunds from the latest valid bundles.
     // Allow caller to set how many bundles to look back for refunds. The default is set to 2 which means
     // we'll look back only at the two latest valid bundle unless the caller overrides.
@@ -153,7 +128,7 @@ export class InventoryClient {
     const nextBundleRefunds = this.bundleDataClient.getNextBundleRefunds();
     refundsToConsider.push(nextBundleRefunds);
 
-    const totalRefundsPerChain = Object.fromEntries(
+    return Object.fromEntries(
       this.getEnabledChains().map((chainId) => {
         const destinationToken = this.getDestinationTokenForL1Token(l1Token, chainId);
         return [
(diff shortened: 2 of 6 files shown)


src/clients/TokenClient.ts
```diff
@@ -21,11 +21,6 @@ export class TokenClient {
     return this.tokenData;
   }
 
-  getDataForToken(chainId: number, token: string) {
```
kevinuma, line 24: This function is not used.

.gitignore
```diff
@@ -7,7 +7,7 @@ typechain
 dump.rdb
 .idea
 
-#Hardhat files
+# Hardhat files
```
chrismaree, line 10: :)

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -63,19 +63,19 @@ export class InventoryClient {
     return distributionPerL1Token;
   }
 
-  // Get the allocation of a given token, considering the shortfall. This number will be negative if there is a shortfall!
-  // A shortfall, by definition, is the amount of tokens that are needed to to fill a given outstanding relay that the
-  // relayer could not fill. For example of the relayer has 10 WETH and needs to fill a relay of size 18 WETH then the
-  // shortfall will be 18. If there are 140 WETH across the whole ecosystem then this method will return (10-18)/140=-0.057
-  // This number is then used to inform how many tokens need to be sent to the chain to: a) cover the shortfall and
-  // b bring the balance back over the defined target.
```
chrismaree, line 71: why remove this?

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -214,17 +227,27 @@ export class InventoryClient {
...
         if (cumulativeBalance.eq(0)) continue;
+
         for (const chainId of this.getEnabledL2Chains()) {
-          const currentAllocPct = this.getCurrentAllocationPctConsideringShortfall(l1Token, chainId);
-          const thresholdPct = toBN(this.inventoryConfig.tokenConfig[l1Token][chainId].thresholdPct);
-          if (currentAllocPct.lt(thresholdPct)) {
-            const deltaPct = toBN(this.inventoryConfig.tokenConfig[l1Token][chainId].targetPct).sub(currentAllocPct);
-            assign(rebalancesRequired, [chainId, l1Token], deltaPct.mul(cumulativeBalance).div(scalar));
+          const currentBalanceConsideringTransfers = this.getBalanceOnChainForL1Token(chainId, l1Token);
+          const shortfall = this.getTokenShortFall(l1Token, chainId) || toBN(0);
```
chrismaree, line 233: I see. I was confused where you were considering the shortfall but I see you do it here now. rather than factoring it into the pct allocation you rather target the _exact_ amount. that makes sence!

chrismaree (commented):
this looks good! this should make a meaningful impact. how have you tested this in prod?

nicholaspai:
This needs to be rebased off master due to merge conflicts FYI

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -214,17 +227,27 @@ export class InventoryClient {
...
+          // Rebalance if there's a shortfall (current balance + transfers cannot cover) or if allocation is below
+          // target.
+          if (currentBalanceConsideringTransfers.lt(shortfall)) {
```
mrice32, line 236: This makes sense! I don't think this should change anything in this PR, but I want to understand one case:
- Balance 399 USDC, Target 500 USDC, Threshold 400 USDC, Shortfall 201 USDC, 101 USDC transfer in progress.
- Is this state possible?
- If, so would we transfer 201 USDC?
- Once all transfers are completed, would we then fill the deposit, leaving 101 USDC (399 + 101 + 201 - 600), causing us to initiate a new transfer of 399 to be initiated?
- Does this mean that we should be sending shortfall + current balance rather than just shortfall so we can avoid this extra transfer?

I could be missing something so please feel free to point out any errors in my logic.

mrice32 (commented):
Looks great! A few minor comments and questions

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -63,19 +63,19 @@ export class InventoryClient {
...
-  // This number is then used to inform how many tokens need to be sent to the chain to: a) cover the shortfall and
-  // b bring the balance back over the defined target.
-  getCurrentAllocationPctConsideringShortfall(l1Token: string, chainId: number): BigNumber {
-    const cumulativeBalance = this.getCumulativeBalance(l1Token); // If there is nothing over all chains, return early.
+  // Get the balance of a given token on a given chain, including any pending cross chain transfers and upcoming refunds
+  getCurrentAllocationPctConsideringRefunds(l1Token: string, chainId: number): BigNumber {
+    const totalRefundsPerChain = this.getBundleRefunds(l1Token);
+    const cumulativeRefunds = Object.values(totalRefundsPerChain).reduce((acc, curr) => acc.add(curr), toBN(0));
+    // If there is nothing over all chains, return early.
+    const cumulativeBalance = this.getCumulativeBalance(l1Token).add(cumulativeRefunds);
```
nicholaspai, line 71: I think it'd be helpful for other logic like implementing partial fill logic to add into the `getCumulativeBalance(l1Token)` function the `getCumulativeBalance` logic. This way there is one easy function to ready the total "available" funds for an l1 token

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -112,13 +112,38 @@ export class InventoryClient {
...
+    // old. The downside to ignoring these potentially older unexecuted refunds is that the virtual balance will be
+    // lower than it should be, meaning that we'll send more refunds to the spoke pool. However, these unexecuted refunds
+    // should be rare in practice since leaves are usually executed immediately.
+    const latestBundleRefunds = this.bundleDataClient.getPendingRefundsFromLatestBundle();
+    const nextBundleRefunds = this.bundleDataClient.getNextBundleRefunds();
+
+    return Object.fromEntries(
+      this.getEnabledChains()
+        .map((chainId) => [this.getDestinationTokenForL1Token(l1Token, chainId), chainId])
+        .map(([tokenId, chainId]) => [
```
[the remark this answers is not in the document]
nicholaspai: Agreed but this bit has been changed and is code from an older rebase at this point

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -112,13 +112,38 @@ export class InventoryClient {
...
+          chainId,
+          this.bundleDataClient
+            .getRefundsFor(latestBundleRefunds, this.relayer, Number(chainId), String(tokenId))
+            .add(
+              this.bundleDataClient.getRefundsFor(nextBundleRefunds, this.relayer, Number(chainId), String(tokenId))
```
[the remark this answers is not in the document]
nicholaspai: agreed

src/clients/InventoryClient/InventoryClient.ts
kevinuma: I tried best to explain this in the PR's description and code's comment but let me know if/how it can be improved.

kevinuma:
> this looks good! this should make a meaningful impact. how have you tested this in prod?

Not yet my current strategy is just to run it locally (with no funds + in simulation mode) to cover different kind of circumstances. Is it worth doing what you did with running a separate relayer?

src/clients/InventoryClient/InventoryClient.ts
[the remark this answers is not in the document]
kevinuma: Yes it's gone after the merge

src/clients/InventoryClient/InventoryClient.ts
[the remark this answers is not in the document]
kevinuma: This was renamed to getTotalRefund after Nick's changes so getRefundsFor is now gone.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -142,26 +166,8 @@ export class InventoryClient {
...
-    const totalRefundsPerChain = Object.fromEntries(
-      this.getEnabledChains().map((chainId) => {
-        const destinationToken = this.getDestinationTokenForL1Token(l1Token, chainId);
-        return [
-          chainId,
-          this.bundleDataClient.getTotalRefund(refundsToConsider, this.relayer, Number(chainId), destinationToken),
-        ];
-      })
-    );
+    // Consider any refunds from both current pending bundle (if any) and the next bundle.
```
nicholaspai, line 169:
```suggestion
    // Consider any refunds from executed and to-be executed bundles
```

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -200,6 +206,13 @@ export class InventoryClient {
     return refundChainId;
   }
 
+  // A rebalance is triggered in the follow 2 cases:
+  // 1. The allocation is not below target when considering upcoming refunds, but there's a shortfall. We'll send the
```
nicholaspai, line 210: > The allocation is not below target when considering upcoming refunds

This function doesn't look like it implements the above logic exactly. You don't check this condition, instead you first check if there's a shortfall. If there is NOT a shortfall, then you check the above logic about whether the allocation is above or below target

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -64,19 +64,19 @@ export class InventoryClient {
     return distributionPerL1Token;
   }
 
-  // Get the allocation of a given token, considering the shortfall. This number will be negative if there is a shortfall!
-  // A shortfall, by definition, is the amount of tokens that are needed to to fill a given outstanding relay that the
-  // relayer could not fill. For example of the relayer has 10 WETH and needs to fill a relay of size 18 WETH then the
-  // shortfall will be 18. If there are 140 WETH across the whole ecosystem then this method will return (10-18)/140=-0.057
-  // This number is then used to inform how many tokens need to be sent to the chain to: a) cover the shortfall and
-  // b bring the balance back over the defined target.
-  getCurrentAllocationPctConsideringShortfall(l1Token: string, chainId: number): BigNumber {
```
nicholaspai, line 73: I'm not a huge fan of renaming this function and then changing the logic as drastically as you do. Why not just add in `refunds` logic into this function?

kevinuma: I think this is just github diffing issue. The new function is not the same as the old one at all - specifically this new one should not look at shortfalls. As explained in the code comment and in a Github comment further below, the checks happen in this exact order:

1. We check if there's a shortfall at all that's not already covered by pending cross chain transfers. If so, we rebalance to send the exact shorfall amount over.
2. If there's no shortfall, then we look at the current allocation (including refunds but not looking at shortfalls at all, which is what the previous function does). If this is below the threshold, we rebalance and send over the amount to cover the diff between current and target allocations.

src/clients/InventoryClient/InventoryClient.ts
kevinuma: Updated

src/clients/InventoryClient/InventoryClient.ts
nicholaspai: `if (currentBalanceConsideringTransfers.lt(shortfall))` i feel like i'm missing something, how does this implement: "The allocation is not below target when considering upcoming refunds, but there's a shortfall". You don't read the allocationPct until line 239

src/clients/InventoryClient/InventoryClient.ts
nicholaspai: > but not looking at shortfalls at all, which is what the previous function did

I thought the previous function `getCurrentAllocationPctConsideringShortfall` looks at shortfalls?

nicholaspai:
> > this looks good! this should make a meaningful impact. how have you tested this in prod?
> 
> Not yet my current strategy is just to run it locally (with no funds + in simulation mode) to cover different kind of circumstances. Is it worth doing what you did with running a separate relayer?

I think for various reasons you should run this in a separate relayer, and it can help with adding documentation details for external runners!

src/clients/InventoryClient/InventoryClient.ts
kevinuma: Sorry I think the comment is not super clear. I reworded it to be clearer that it looks at the shortfalls first. THEN current allocation (with refunds and without shortfalls).

src/clients/InventoryClient/InventoryClient.ts
kevinuma: It did. This new one does not. That's what that sentence is saying - the new one does not look at shortfalls while the old one does. Sorry I should stop writing long composite sentences haha.

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -214,17 +228,27 @@ export class InventoryClient {
...
-            const deltaPct = toBN(this.inventoryConfig.tokenConfig[l1Token][chainId].targetPct).sub(currentAllocPct);
-            assign(rebalancesRequired, [chainId, l1Token], deltaPct.mul(cumulativeBalance).div(scalar));
+          const currentBalanceConsideringTransfers = this.getBalanceOnChainForL1Token(chainId, l1Token);
+          const shortfall = this.getTokenShortFall(l1Token, chainId) || toBN(0);
+          // Rebalance if there's a shortfall (current balance + transfers cannot cover) or if allocation is below
+          // target.
+          if (currentBalanceConsideringTransfers.lt(shortfall)) {
+            assign(rebalancesRequired, [chainId, l1Token], shortfall);
+          } else {
+            const currentAllocPct = this.getCurrentAllocationPctConsideringRefunds(l1Token, chainId);
```
nicholaspai, line 240: Why can't you still consider shortfalls in this method? I think it's super useful to have a function that gives your total net current allocation inclusive of all virtual balance adjustments including cross chain transfers, shortfalls, and pending refunds. For example we should use this when computing partial fill decisions. Thoughts?

nicholaspai (commented):
One comment!

src/clients/InventoryClient/InventoryClient.ts
mrice32: Sorry, I probably didn't make this clear enough. I was saying a shortfall of 201 because of a transfer of 600. Does that change some of your logic above?

mrice32 (commented):
My only remaining comment: https://github.com/across-protocol/relayer-v2/pull/142#discussion_r903414175

src/clients/InventoryClient/InventoryClient.ts
kevinuma: So basically a shortfall of 600 (the full unfilled amount is registered as a shortfall not the diff). I think it'd play out like below:

1. When rebalanceInventoryIfNeeded is called, since there's a shortfall, rebalance is triggered and 600 USDC is sent over to L2.
2. Once the 600 USDC arrives and when filling that deposit, we see that after the deposit, the balance would be 399 + 600 - 600 + 101 (refunds) = 500, which is at target. The repayment chain is therefore mainnet for the 600 USDC fill. 

This is nice I think because we save the return trip of the rebalance via requesting refund on the right chain.

PR #143, opened 2022-06-16
fix: Look at the right starting block when calculating pending refunds

kevinuma:
Currently the logic looks at the starting block of the next bundle, which only works if the current pending bundle has not been executed in HubPool. Once it has been executed, the starting block would be in the future while the ending block is still in the past.


src/clients/BundleDataClient.ts
```diff
@@ -60,7 +60,7 @@ export class BundleDataClient {
     // Use the same block range as the current pending bundle to ensure we look at the right refunds.
     const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map(
       (endBlock, i) => [
-        hubPoolClient.getNextBundleStartBlockNumber(chainIds, latestBlockNumber, chainIds[i]),
```
nicholaspai, line 63: Jeez good catch, I should have caught this too

nicholaspai (approved):

PR #145, opened 2022-06-16
fix: Prevent overlapping when setting bundle's starting block for fetching pending refunds

kevinuma:
Problem: HubPoolClient#getNextBundleStartBlockNumber returns the starting block of the next bundle if the previous bundle is considered "valid", which is only the case if it has been fully executed by the time the next bundle is proposed. This creates a racing condition where HubPoolClient#getNextBundleStartBlockNumber can return 0 for the starting block number if the previous bundle has not fully been executed by the time the next bundle is proposed.

Impact: This leads to double counting of everything in the balance report.

Solution: We can either pass latestBundle.blockNumber in the getNextBundleStartBlockNumber or just look directly at the ending block of the last fully executed bundle. The latter is more readable/intuitive so this PR goes with it.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -50,19 +50,22 @@ export class BundleDataClient {
 
   // Return refunds from a pending bundle if any.
   getPendingBundleRefunds(): FillsToRefund {
-    const chainIds = Object.keys(this.spokePoolClients).map(Number);
-
     // If latest proposed bundle has not been executed yet, we have a bundle pending liveness.
     const hubPoolClient = this.clients.hubPoolClient;
     const latestBlockNumber = hubPoolClient.latestBlockNumber;
     const latestBundle = hubPoolClient.getMostRecentProposedRootBundle(latestBlockNumber);
+    // Look for the latest fully executed root bundle before the current last bundle.
+    // This ensures that we skip over any disputed (invalid) bundles.
+    const previousValidBundle = hubPoolClient.getLatestFullyExecutedRootBundle(latestBundle.blockNumber);
 
     // Use the same block range as the current pending bundle to ensure we look at the right refunds.
     const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map(
-      (endBlock, i) => [
-        hubPoolClient.getNextBundleStartBlockNumber(chainIds, endBlock.toNumber(), chainIds[i]),
-        endBlock.toNumber(),
-      ]
+      (endBlock, i) => {
+        const fromBlock = previousValidBundle?.bundleEvaluationBlockNumbers?.[i]
+          ? previousValidBundle.bundleEvaluationBlockNumbers[i].toNumber() + 1
+          : 0;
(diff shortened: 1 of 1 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -50,17 +50,16 @@ export class BundleDataClient {
 
   // Return refunds from a pending bundle if any.
   getPendingBundleRefunds(): FillsToRefund {
-    const chainIds = Object.keys(this.spokePoolClients).map(Number);
-
     // If latest proposed bundle has not been executed yet, we have a bundle pending liveness.
     const hubPoolClient = this.clients.hubPoolClient;
     const latestBlockNumber = hubPoolClient.latestBlockNumber;
     const latestBundle = hubPoolClient.getMostRecentProposedRootBundle(latestBlockNumber);
+    const previousBundle = hubPoolClient.getPreviousRootBundle(latestBundle);
```
nicholaspai, line 57: This is more readable + 1

nicholaspai: Perhaps we should refactor this logic and create a function `getBlockRangeForBundle(bundleProposalBlock: number): number[][]`

src/clients/BundleDataClient.ts
```diff
@@ -50,17 +50,16 @@ export class BundleDataClient {
...
 
     // Use the same block range as the current pending bundle to ensure we look at the right refunds.
     const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map(
       (endBlock, i) => [
-        hubPoolClient.getNextBundleStartBlockNumber(chainIds, endBlock.toNumber(), chainIds[i]),
```
nicholaspai, line 63: I think the reason why this is failing is because you're passing in `endBlock.toNumber()` instead of `latestBundle.blockNumber`. `endBlock.toNumber()` would mean that `getNextBundleStartBlockNumber` would return `endBlock + 1` in cases where the `latestBundle` has been fully executed

kevinuma: This makes 100% sense thank you! Updated.

kevinuma: Updated

src/clients/HubPoolClient.ts
```diff
@@ -226,6 +226,12 @@ export class HubPoolClient {
     ) as ProposedRootBundle;
   }
 
+  getPreviousRootBundle(currentRootBundle: ProposedRootBundle) {
+    return sortEventsDescending(this.proposedRootBundles).find(
+      (_rootBundle: ProposedRootBundle) => _rootBundle.blockNumber < currentRootBundle.blockNumber
```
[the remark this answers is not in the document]
kevinuma: This makes 100% sense thank you! Updated.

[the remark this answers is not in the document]
nicholaspai: Should we just delete this until we need it in future?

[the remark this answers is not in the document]
kevinuma: Deleted

nicholaspai (approved):

PR #147, opened 2022-06-17
fix: Deep clone data from cache before returning from loadData()

nicholaspai:
JS passes objects by reference, so when we changed `loadData` to return data from the cache (when possible), we sometimes run into bugs when we mutate that return value, causing corrupted data to be loaded from the cache on the next run.

This PR changes `loadData` to always return a deep cloned data from the cache

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -49,6 +49,12 @@ export class BundleDataClient {
     this.loadDataCache = {};
   }
 
+  loadDataFromCache(key: string) {
+    // Always return a deep cloned copy of object stored in cache. Since JS passes by reference instead of value, we
+    // want to minimize the risk that the programmer accidentally mutates data in the cache.
+    return _.cloneDeep(this.loadDataCache[key]);
+  }
+
   // Return refunds from latest bundle
   getPendingRefundsFromLatestBundle(): FillsToRefund {
     const hubPoolClient = this.clients.hubPoolClient;
@@ -148,9 +154,7 @@ export class BundleDataClient {
     const key = JSON.stringify(blockRangesForChains);
 
     if (this.loadDataCache[key]) {
-      // Always return a deep cloned copy of object stored in cache. Since JS passes by reference instead of value, we
-      // want to minimize the risk that the programmer accidentally mutates data in the cache.
-      return _.cloneDeep(this.loadDataCache[key]);
+      return this.loadDataFromCache(key);
     }
 
     if (!this.clients.hubPoolClient.isUpdated) throw new Error(`HubPoolClient not updated`);
@@ -292,8 +296,6 @@ export class BundleDataClient {
 
     this.loadDataCache[key] = { fillsToRefund, deposits, unfilledDeposits, allValidFills };
 
-    // Always return a deep cloned copy of object stored in cache. Since JS passes by reference instead of value, we
-    // want to minimize the risk that the programmer accidentally mutates data in the cache.
-    return _.cloneDeep(this.loadDataCache[key]);
+    return this.loadDataFromCache(key);
   }
 }
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -1,4 +1,5 @@
 import { winston, BigNumber } from "../utils";
+import * as _ from "lodash";
 import {
   Deposit,
   DepositWithBlock,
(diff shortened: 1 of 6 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -95,10 +94,13 @@ export class BundleDataClient {
     return this.loadData(futureBundleEvaluationBlockRanges, this.spokePoolClients, false).fillsToRefund;
   }
 
-  deductExecutedRefunds(allRefunds: FillsToRefund, latestBundle: ProposedRootBundle) {
+  deductExecutedRefunds(_allRefunds: FillsToRefund, bundleContainingRefunds: ProposedRootBundle): FillsToRefund {
+    const allRefunds = _.cloneDeep(_allRefunds);
```
kevinuma, line 98: Why do a deep clone here?

nicholaspai: This is actually the main bug fix I've realized. This is because `this.loadData` loads `fillsToRefund` from a cache. So passing in `fillsToRefund` directly into `deductExecutedRefunds` is inadvertently mutating the object in the cache. We don't want `fillsToRefund` to be different between different calls to `getPendingRefundsFromLatestBundle` so we should deep clone it before passing into `deductExecutedRefunds`

src/clients/BundleDataClient.ts
```diff
@@ -48,35 +49,33 @@ export class BundleDataClient {
...
+  getPendingRefundsFromLatestBundle(): FillsToRefund {
     const hubPoolClient = this.clients.hubPoolClient;
     const latestBlockNumber = hubPoolClient.latestBlockNumber;
     const latestBundle = hubPoolClient.getMostRecentProposedRootBundle(latestBlockNumber);
     // Look for the latest fully executed root bundle before the current last bundle.
     // This ensures that we skip over any disputed (invalid) bundles.
     const previousValidBundle = hubPoolClient.getLatestFullyExecutedRootBundle(latestBundle.blockNumber);
 
-    // Use the same block range as the current pending bundle to ensure we look at the right refunds.
-    const pendingBundleEvaluationBlockRanges: number[][] = latestBundle.bundleEvaluationBlockNumbers.map(
```
nicholaspai, line 62: I also did some renaming

src/clients/BundleDataClient.ts
```diff
@@ -48,35 +49,33 @@ export class BundleDataClient {
...
+      latestBundleEvaluationBlockRanges,
       this.spokePoolClients,
       false
     );
 
     // The latest proposed bundle's refund leaves might have already been partially or entirely executed.
     // We have to deduct the executed amounts from the total refund amounts.
-    this.deductExecutedRefunds(pendingBundleRefunds, latestBundle);
-    return pendingBundleRefunds;
+    return this.deductExecutedRefunds(latestFillsToRefund, latestBundle);
```
nicholaspai, line 78: I made `deductExecutedRefunds` a bit more functional and return the modified value

src/clients/BundleDataClient.ts
```diff
@@ -87,18 +86,24 @@ export class BundleDataClient {
     // The future bundle covers block range from the ending of the last proposed bundle (might still be pending liveness)
     // to the latest block on that chain.
     const chainIds = Object.keys(this.spokePoolClients).map(Number);
-    const futureBundleEvaluationBlockRanges: number[][] = latestProposedBundle.bundleEvaluationBlockNumbers.map(
-      (endingBlock, i) => [endingBlock.toNumber() + 1, this.spokePoolClients[chainIds[i]].latestBlockNumber]
-    );
+    const latestProposalEndBlocks = latestProposedBundle
```
nicholaspai, line 89: This is a small edge case fix to handle cae where there is no `latestProposedBundle` at the deployment of a hub pool

src/clients/InventoryClient/InventoryClient.ts
```diff
@@ -141,20 +141,26 @@ export class InventoryClient {
...
 
-    // Consider any refunds from both current pending bundle (if any) and the next bundle.
-    let pendingBundleRefunds = this.bundleDataClient.getPendingBundleRefunds();
-    let nextBundleRefunds = this.bundleDataClient.getNextBundleRefunds();
+    // Consider any refunds from both latest bundle (even if its pending liveness) and the next bundle.
+    // TODO: This will not count any refunds older than the latest bundle that are unexecuted on the spoke pools,
+    // so a future implementation should add in this feature as an option, to look back at refunds up to X bundles
+    // old. The downside to ignoring these potentially older unexecuted refunds is that the virtual balance will be
+    // lower than it should be, meaning that we'll send more refunds to the spoke pool. However, these unexecuted refunds
+    // should be rare in practice since leaves are usually executed immediately.
```
mrice32, line 149: What about this case:
1. Bundle gets past liveness, and all PoolRebalaceLeaves are executed.
2. New bundle is proposed.
3. 20-30 minutes later, the RelayerRefundRoots make it to the L2s and they are executed.

What happens between 2 and 3? Does the relayer's effective balance drop? In the comment, it wasn't obvious to me whether we're still considering the last _executed_ bundle or if we're just looking at the pending (in liveness bundle).

mrice32 (approved):
Only one comment. Looks great!

src/clients/InventoryClient/InventoryClient.ts
kevinuma: This is exactly what this comment is talking about IMO. I think the only ways we can really do this is to look at up to 2 or more bundles back every time. For now, this case is unlikely since we have a pretty high threshold of $1M before a bundle can be created. In the future, I think even if we start getting a lot more volume, we might increase this threshold as it can be costly to propose bundles too frequently right?

mrice32 (approved):
LGTM!

PR #148, opened 2022-06-17
improve: Deduplicate transactions when printing unknown transfers

kevinuma:


src/monitor/Monitor.ts
```diff
@@ -456,10 +456,8 @@ export class Monitor {
 
     const totalAmount = this.getTotalTransferAmount(transfers);
     let mrkdwn = `other: ${convertFromWei(totalAmount.toString(), decimals)}\n`;
-    mrkdwn += etherscanLinks(
```
nicholaspai, line 459: Why were there duplicates in the first place?

kevinuma: That's because the transfer data is organized by events, and there are transactions that contain multiple transfers/events. TokenTransferClient currently organizes these events by chain id + token and not by transaction hash.

nicholaspai (approved):

PR #149, opened 2022-06-17
improve: Consolidate bundle refund calculations and look up to N root bundles into past for pending refunds

nicholaspai:
There are now two types of possible pending refunds:
- Those contained in validated root bundles (i.e. the root bundle passed liveness and all root bundle leaves were executed)
- Those contained in unvalidated root bundles (i.e. the next potential root bundle to be proposed or contained in a pending liveness)

This makes it a bit easier to look up N validated root bundles into the past to find pending refunds. The main change with this is that refunds in the next potential root bundle are now grouped together with those in a pending bundle that has not passed liveness.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -61,10 +61,9 @@ export class BundleDataClient {
     let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
     for (let i = 0; i < bundleLookback; i++) {
       const bundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(latestBlock);
-      if (bundle) {
+      if (bundle === undefined) {
         latestBlock = bundle.blockNumber;
-        const bundleRefunds = this.getPendingRefundsFromBundle(bundle);
-        refunds.push(bundleRefunds);
+        refunds.push(this.getPendingRefundsFromBundle(bundle));
       } else break; // No more valid bundles in history!
     }
     return refunds;
@@ -91,7 +90,11 @@ export class BundleDataClient {
     return this.deductExecutedRefunds(fillsToRefund, bundle);
   }
 
-  // Return refunds from the next bundle.
+  // Return refunds from the next valid bundle. This will contain any refunds that have been sent but are not included
+  // in a valid bundle with all of its leaves executed. This contains refunds from:
+  // - Bundles that passed liveness but have not had all of their pool rebalance leaves executed.
+  // - Bundles that are pending liveness
+  // - Not yet proposed bundles
   getNextBundleRefunds(): FillsToRefund {
     const chainIds = Object.keys(this.spokePoolClients).map(Number);
     const futureBundleEvaluationBlockRanges: number[][] = chainIds.map((chainId, i) => [
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -61,7 +61,8 @@ export class BundleDataClient {
     let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
     for (let i = 0; i < bundleLookback; i++) {
       const bundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(latestBlock);
-      if (bundle === undefined) {
+      if (bundle !== undefined) {
+        // Update latest block so next iteration can get the next oldest bundle:
         latestBlock = bundle.blockNumber;
         refunds.push(this.getPendingRefundsFromBundle(bundle));
       } else break; // No more valid bundles in history!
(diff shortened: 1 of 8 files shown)


mrice32 (approved):
LGTM! This is awesome, should make the calculation a lot more accurate based on how the bots are run. I also think customizability isn't super important here unless we start running into a lot of situations where runningBalance is nonzero.

src/clients/BundleDataClient.ts
```diff
@@ -55,48 +55,51 @@ export class BundleDataClient {
...
 
-  // Return refunds from latest bundle
-  getPendingRefundsFromLatestBundle(): FillsToRefund {
+  getPendingRefundsFromValidBundles(bundleLookback: number): FillsToRefund[] {
+    const refunds = [];
+
+    let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
+    for (let i = 0; i < bundleLookback; i++) {
+      const bundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(latestBlock);
+      if (bundle) {
```
kevinuma, line 64: nit: I think we should make it explicit with bundle === undefined

src/clients/BundleDataClient.ts
```diff
@@ -55,48 +55,51 @@ export class BundleDataClient {
     return _.cloneDeep(this.loadDataCache[key]);
   }
 
-  // Return refunds from latest bundle
-  getPendingRefundsFromLatestBundle(): FillsToRefund {
+  getPendingRefundsFromValidBundles(bundleLookback: number): FillsToRefund[] {
```
kevinuma, line 58: I think bundleLookback is a bit of a magic number. I see that you're calling with 2 everywhere. In theory we could have an event where > 2 is required (let's say a chain is down for more than 2 hours in which case we would need to look at more than 2). This is definitely tricky. Let's at least make this a config param set via .env?

src/clients/BundleDataClient.ts
```diff
@@ -55,48 +55,51 @@ export class BundleDataClient {
...
+    const refunds = [];
+
+    let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
+    for (let i = 0; i < bundleLookback; i++) {
+      const bundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(latestBlock);
+      if (bundle) {
+        latestBlock = bundle.blockNumber;
+        const bundleRefunds = this.getPendingRefundsFromBundle(bundle);
+        refunds.push(bundleRefunds);
```
kevinuma, line 67: nit: You can do refunds.push(this.getPendingRefundsFromBundle(bundle))

src/clients/BundleDataClient.ts
```diff
@@ -55,48 +55,51 @@ export class BundleDataClient {
...
+      } else break; // No more valid bundles in history!
+    }
+    return refunds;
+  }
+
+  // Return refunds from input bundle.
+  getPendingRefundsFromBundle(bundle: ProposedRootBundle): FillsToRefund {
```
kevinuma, line 74: Nice!

src/clients/BundleDataClient.ts
```diff
@@ -55,48 +55,51 @@ export class BundleDataClient {
...
-    // The future bundle covers block range from the ending of the last proposed bundle (might still be pending liveness)
-    // to the latest block on that chain.
     const chainIds = Object.keys(this.spokePoolClients).map(Number);
-    const latestProposalEndBlocks = latestProposedBundle
-      ? latestProposedBundle.bundleEvaluationBlockNumbers.map((block) => block.toNumber() + 1)
-      : Array(chainIds.length).fill(0);
-    const futureBundleEvaluationBlockRanges: number[][] = latestProposalEndBlocks.map((endingBlock, i) => [
-      endingBlock,
+    const futureBundleEvaluationBlockRanges: number[][] = chainIds.map((chainId, i) => [
+      this.clients.hubPoolClient.getNextBundleStartBlockNumber(
```
kevinuma, line 98: I think we should document that this can return a block range that includes any current pending bundle (it's basically from the last fully executed block's end block + 1 on). We should also document getNextBundleRefunds and may be rename it to getUnprocessedRefunds or something because it includes both those from pending bundle and the bundle after that (so multiple bundles)

src/clients/BundleDataClient.ts
nicholaspai: yeah in my comment i just say make this configurable in another PR. See my TODO in the inventory client

src/clients/BundleDataClient.ts
nicholaspai: good callout

src/clients/BundleDataClient.ts
nicholaspai: Added this comment above:

```
  // Return refunds from the next bundle. This will contain any refunds that have been sent but are not included
  // in a valid bundle with all of its leaves executed. This contains refunds from:
  // - Bundles that passed liveness but have not had all of their pool rebalance leaves executed.
  // - Bundles that are pending liveness
  // - Not yet proposed bundles
```

kevinuma (approved):

PR #150, opened 2022-06-18
fix: Convert ExecutedRelayerRefundLeaf.l2TokenAddress from native to wrapped ETH on OVM events

nicholaspai:


src/clients/SpokePoolClient.ts
```diff
@@ -280,6 +285,17 @@ export class SpokePoolClient {
     this.log("debug", `SpokePool client for chain ${this.chainId} updated!`, searchConfig);
   }
 
+  static getExecutedRefundLeafL2Token(chainId: number, eventL2Token: string) {
+    // If execution of WETH refund leaf occurred on an OVM spoke pool, then we'll convert its l2Token from the native
+    // token address to the wrapped token address. This is because the OVM_SpokePool modifies the l2TokenAddress prop
+    // in _bridgeTokensToHubPool before emitting the ExecutedRelayerRefundLeaf event
```
kevinuma, line 291: I think we should document that this is due to https://github.com/across-protocol/contracts-v2/blob/954528a4620863d1c868e54a370fd8556d5ed05c/contracts/Ovm_SpokePool.sol#L142 for records. In case in the (hopefully near) future, someone else runs this bot and wonder why this has to be done.

kevinuma (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -99,7 +99,7 @@ export class SpokePoolClient {
     return this.relayerRefundExecutions;
   }
 
-  getExecutedRefunds(relayerRefundRoot: string) {
+  getExecutedRefunds(relayerRefundRoot: string, chainId: number) {
```
kevinuma, line 102: I think you need to change all references since it needs 2 args now.

src/clients/SpokePoolClient.ts
nicholaspai: added

src/clients/SpokePoolClient.ts
nicholaspai: oh shoot

nicholaspai: i removed this. Ended up not using this param since i push the modified l2 token address in update()

PR #151, opened 2022-06-19
fix: Correct chain id passed in to getExecutedRefundLeafL2Token

kevinuma:


src/clients/SpokePoolClient.ts
```diff
@@ -273,7 +273,7 @@ export class SpokePoolClient {
     for (const event of executedRelayerRefundRootEvents) {
       const executedRefund = spreadEventWithBlockNumber(event) as RelayerRefundExecutionWithBlock;
       executedRefund.l2TokenAddress = SpokePoolClient.getExecutedRefundLeafL2Token(
-        event.chainId,
```
kevinuma, line 276: event.chainId is not right because chainId is only added by spreadEvent. We need to pass executedRefund here

src/clients/SpokePoolClient.ts
```diff
@@ -294,7 +294,7 @@ export class SpokePoolClient {
     if (chainId === 10 && eventL2Token.toLowerCase() === "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000")
       return "0x4200000000000000000000000000000000000006";
     else if (chainId === 288 && eventL2Token.toLowerCase() === "0x4200000000000000000000000000000000000006")
-      return "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000";
+      return "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000";
```
kevinuma, line 297: Addresses returned from events are all in checksummed format  so data in our system all expects it. Therefore we should return it here.

nicholaspai (approved):

nicholaspai (approved):

nicholaspai (approved):

PR #152, opened 2022-06-19
feat: Simulation (Dry run) mode for all bots

kevinuma:


chrismaree (commented):
meta comment: this makes the code really messy and bloated. you have had to modify _every_ location where we send transactions with `return null`. would it not be cleaner to just change the `runTransaction` function from within the `TransactionUtil`? things like the changes in the Adapter clients seems way overly verbose. you take something that was 2 lines and turn it into 10 lines with a simple `return null`.

mrice32 (commented):
Is the intent of this change to replace the existing SEND_TRANSACTIONS? Should we remove other env vars so we don't have multiple that do the same thing?

nicholaspai:
> meta comment: this makes the code really messy and bloated. you have had to modify _every_ location where we send transactions with `return null`. would it not be cleaner to just change the `runTransaction` function from within the `TransactionUtil`? things like the changes in the Adapter clients seems way overly verbose. you take something that was 2 lines and turn it into 10 lines with a simple `return null`.

I agree with this this isn't the optimal solution IMO. I think like Chris says, we should have `runTransaction` have a `dry-run` mode.

On Matt's point about `SEND_TRANSACTIONS` I actually don't think this PR needs to touch that logic. The `SEND_TRANSACTIONS/SEND_DISPUTES/SEND_PROPOSALS/SEND_EXECUTIONS` toggles the `multicaller` client (which does NOT use `runTransactions` and I think I like having the bots be able to run some of the code in simulation mode and some in prod (for example, running a proposer in prod but a disputer in simulation mode).

kevinuma:
> > toggles the `multicaller` client (which does NOT use `runTransactions`
> 
> @nicholaspai I think it does use `runTransactions`, so if we modify that function, this _would_ affect anything sent through multicaller client, right?

I'll close this PR for now and look into the solution suggested here and potentially send it out as a separate PR later.

PR #157, opened 2022-06-22
fix: add cap to the number of multicall transactions sent in a single txn

mrice32:


src/clients/MultiCallerClient.ts
```diff
@@ -95,6 +98,10 @@ export class MultiCallerClient {
         assign(groupedTransactions, [transaction.chainId], [transaction]);
       }
 
+      const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
+        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 100)])
```
nicholaspai, line 102: Is 100 the limit to # of txns in multicall? A small comment would be great pls

nicholaspai (approved):

PR #158, opened 2022-06-22
improve: Remove .wait() for transactions since we already have hash to speedup bots

nicholaspai:
I believe that our signer sends back the `hash` right away when executing `.send()` because the signer is a `Wallet` and the refore we can derive the transaction has immediately, unlike a generic `Signer`. This is my understanding after reading the thread and comment [here](https://github.com/ethers-io/ethers.js/issues/511#issuecomment-491476049)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "node": ">=8.3.0"
   },
   "dependencies": {
-    "@across-protocol/contracts-v2": "^1.0.2",
+    "@across-protocol/contracts-v2": "^1.0.0",
     "@across-protocol/sdk-v2": "^0.1.10",
     "@arbitrum/sdk": "^1.1.4",
     "@defi-wonderland/smock": "^2.0.7",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "node": ">=8.3.0"
   },
   "dependencies": {
-    "@across-protocol/contracts-v2": "^1.0.0",
+    "@across-protocol/contracts-v2": "^1.0.2",
     "@across-protocol/sdk-v2": "^0.1.10",
     "@arbitrum/sdk": "^1.1.4",
     "@defi-wonderland/smock": "^2.0.7",
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -167,34 +167,23 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
   allValidFillsInRange: interfaces.FillWithBlock[],
   chainIdListForBundleEvaluationBlockNumbers: number[]
 ) {
-  // We need to subtract excess from any fills that might replaced a slow fill sent to the fill destination chain.
-  // This can only happen if the fill was the last fill for a deposit. Otherwise, its still possible that the slow fill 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "node": ">=8.3.0"
   },
   "dependencies": {
-    "@across-protocol/contracts-v2": "^1.0.0",
+    "@across-protocol/contracts-v2": "^1.0.2",
```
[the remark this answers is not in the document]
mrice32: Interesting

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -169,7 +169,7 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
   chainIdListForBundleEvaluationBlockNumbers: number[]
 ) {
   // We need to subtract excess from any fills that might replaced a slow fill sent to the fill destination chain.
-  // This can only happen if the fill was the last fill for a deposit. Otherwise, its still possible that the slow fill
```
kevinuma: lol. Intellij?

src/utils/FillUtils.ts
```diff
@@ -105,7 +113,7 @@ export function getFillDataForSlowFillFromPreviousRootBundle(
   chainIdListForBundleEvaluationBlockNumbers: number[]
 ) {
   // Find the first fill chronologically for matched deposit for the input fill.
-  const firstFillForSameDeposit = allValidFills.find(
+  const firstFillForSameDeposit = sortEventsAscending(allValidFills).find(
```
mrice32: Is this a fix or an optimization?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -167,25 +168,35 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
...
-      );
+  // We need to subtract excess from any fills that might replaced a slow fill sent to the fill destination chain.
+  // This can only happen if the fill was the last fill for a deposit. Otherwise, its still possible that the slow fill
+  // for the deposit can be executed, so we'll defer the excess calculation until the hypothetical slow fill executes.
+  // In addition to fills that are not the last fill for a deposit, we can ignore fills that completely fill a deposit
+  // as the first fill. These fills could never have triggered a deposit since there were no partial fills for it.
+  // This assumption depends on the rule that slow fills can only be sent after a partial fill for a non zero amount
+  // of the deposit. This is why "1 wei" fills are important, otherwise we'd never know which fills originally
+  // triggered a slow fill payment to be sent to the destination chain.
+  sortEventsAscending(allValidFillsInRange)
```
mrice32, line 178: ust so I'm understanding, this is effectively flipping `if (fill.totalFilledAmount.eq(fill.amount))` and `getFillDataForSlowFillFromPreviousRootBundle`, right?

Did this result in a speedup, because I find the new implementation a little harder to read, but if it's faster, I think that's fine.

mrice32 (approved):

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: pushed wrong branch

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: oysged wrong branch

PR #159, opened 2022-06-23
improve: optimize executor by flipping a few comparisons

mrice32:


src/utils/FillUtils.ts
```diff
@@ -77,11 +77,11 @@ export function isFirstFillForDeposit(fill: Fill): boolean {
...
 export function filledSameDeposit(fillA: Fill, fillB: Fill): boolean {
   return (
-    fillA.amount.eq(fillB.amount) &&
+    fillA.depositId === fillB.depositId &&
     fillA.originChainId === fillB.originChainId &&
+    fillA.amount.eq(fillB.amount) &&
     fillA.destinationChainId === fillB.destinationChainId &&
     fillA.relayerFeePct.eq(fillB.relayerFeePct) &&
-    fillA.depositId === fillB.depositId &&
     fillA.recipient === fillB.recipient &&
```
nicholaspai, line 85: We could probably move recipient above originChainId? Addresses are less likely to be repeated than chain ID?

src/utils/FillUtils.ts
```diff
@@ -106,7 +106,7 @@ export function getFillDataForSlowFillFromPreviousRootBundle(
 ) {
   // Find the first fill chronologically for matched deposit for the input fill.
   const firstFillForSameDeposit = allValidFills.find(
-    (_fill: FillWithBlock) => isFirstFillForDeposit(_fill as Fill) && filledSameDeposit(_fill, fill)
+    (_fill: FillWithBlock) => filledSameDeposit(_fill, fill) && isFirstFillForDeposit(_fill as Fill)
```
nicholaspai, line 109: Found this same one, +1000

nicholaspai (approved):
LGTM nice little finds

src/utils/FillUtils.ts
mrice32: I think we should put these two first:

1. I think numerical comparisons are the cheapest by far, so it's way better to try to short circuit before getting to the more expensive comparisons.
2. depositId + originChainId basically filters down on the deposit. Deposits usually have 1 fill, and occasionally have 2, so these two alone should be almost a perfect filter with no string or BN comparisons.

mrice32:
Going to go ahead and merge, just so the dataworker starts operating better in prod. We can tweak these later!

PR #160, opened 2022-06-23
improve: Dataworker should exit early from loops identifying slow fill replacements

nicholaspai:


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -167,15 +168,25 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
...
-      );
+  // We need to subtract excess from any fills that might replaced a slow fill sent to the fill destination chain.
+  // This can only happen if the fill was the last fill for a deposit. Otherwise, its still possible that the slow fill 
+  // for the deposit can be executed, so we'll defer the excess calculation until the hypothetical slow fill executes.
+  // In addition to fills that are not the last fill for a deposit, we can ignore fills that completely fill a deposit
+  // as the first fill. These fills could never have triggered a deposit since there were no partial fills for it.
+  // This assumption depends on the rule that slow fills can only be sent after a partial fill for a non zero amount
+  // of the deposit. This is why "1 wei" fills are important, otherwise we'd never know which fills originally
+  // triggered a slow fill payment to be sent to the destination chain.
+  sortEventsAscending(allValidFillsInRange)
```
nicholaspai, line 179: Sorting events in the same direction as `getFillDataForSlowFillFromPreviousRootBundle` seems like it would be an optimization since the above method runs `.find` on the list of allValidFills

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -167,15 +168,25 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
...
+    .filter((fill) => !fill.totalFilledAmount.eq(fill.amount) && !fill.fillAmount.eq(fill.amount))
```
nicholaspai, line 180: This filter is the key speedup

src/utils/FillUtils.ts
```diff
@@ -105,8 +113,8 @@ export function getFillDataForSlowFillFromPreviousRootBundle(
   chainIdListForBundleEvaluationBlockNumbers: number[]
 ) {
   // Find the first fill chronologically for matched deposit for the input fill.
-  const firstFillForSameDeposit = allValidFills.find(
-    (_fill: FillWithBlock) => filledSameDeposit(_fill, fill) && isFirstFillForDeposit(_fill as Fill)
+  const firstFillForSameDeposit = sortEventsAscending(allValidFills).find(
```
mrice32, line 116: Does the sorting not change behavior? It looks like we're specifically looking for the _first_ fill, right? So if it weren't already in ascending order, wouldn't that mean that this was previously wrong?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -196,16 +207,8 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
...
         chainIdListForBundleEvaluationBlockNumbers
       );
-
-      // Note: `rootBundleEndBlockContainingFirstFill` and `rootBundleEndBlockContainingFullFill` could both
-      // be undefined in the case that both fills are in this current bundle, which is a normal case.
       if (rootBundleEndBlockContainingFirstFill === rootBundleEndBlockContainingFullFill) return;
 
-      // If full fill and first fill are in different blocks, then we should always be able to find the last partial
-      // fill included in the root bundle that also included the slow fill refund.
-      if (!lastFillBeforeSlowFillIncludedInRoot)
```
mrice32, line 206: Where did this error case go? Why were we able to drop this?

src/utils/FillUtils.ts
nicholaspai: removed sorting

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: This error case goes because we exit now earlier if `lastFillbeforeSlowFillIncluded` is undefined. The reason why this error case was caught before is because if `lastFillbeforeSlowFillIncluded === undefined`, then we NEED to exit early because `rootBundleEndBlockContainingFullFill` should also be undefined. This means that the fill that complted the deposit is the current and only fill. If `rootBundleEndBlockContainingFullFill` is not undefined but `lastFillbeforeSlowFillIncluded` is undefined, it means that there is somehow a root bundle (not the current one!) that conatins the fulfill, but no root bundle that contained the first partial fill. This is an impossibility and indicates a programmer error

mrice32 (approved):
LGTM!

PR #161, opened 2022-06-23
improve: Cap # of TokensBridged events examined by Finalizer

nicholaspai:


src/finalizer/index.ts
```diff
@@ -12,21 +12,28 @@ import {
...
 import { HubPoolClient } from "../clients";
 
 export async function finalize(
   logger: winston.Logger,
   hubSigner: Wallet,
   hubPoolClient: HubPoolClient,
   spokePoolClients: SpokePoolClientsByChain,
-  configuredChainIds: number[]
+  configuredChainIds: number[],
+  bridgeEventsToLookup = 200
```
nicholaspai, line 24: After 30 days we're just about at 200 finalization events per chain so this will slow down eventually

PR #162, opened 2022-06-23
Store deposit hashes and map to fills in SpokeClient

nicholaspai:
This allows functions that match fills with deposits to work in O(1) time instead of O(N) which is a huge speedup inpractice because the dataworker and relayer need to run this matching logic on every single fill in a time period

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -147,10 +147,8 @@ export class SpokePoolClient {
 
   getValidUnfilledAmountForDeposit(deposit: Deposit): { unfilledAmount: BigNumber; fillCount: number } {
     const fillsForDeposit = this.depositHashesToFills[this.getDepositHash(deposit)];
-    if (fillsForDeposit === undefined) return { unfilledAmount: toBN(deposit.amount), fillCount: 0 }; // If no fills then the full amount is remaining.
+    if (fillsForDeposit === undefined || []) return { unfilledAmount: toBN(deposit.amount), fillCount: 0 }; // If no fills then the full amount is remaining.
     const fills = fillsForDeposit.filter((fill) => this.validateFillForDeposit(fill, deposit));
-
-    if (fills.length === 0) return { unfilledAmount: toBN(deposit.amount), fillCount: 0 }; // If no valid fills then the full amount is remaining.
 
     // Order fills by totalFilledAmount and then return the first fill's full deposit amount minus total filled amount.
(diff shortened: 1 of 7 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -171,6 +174,10 @@ export class SpokePoolClient {
     return isValid;
   }
 
+  getDepositHash(event: Deposit | Fill) {
+    return `${event.depositId}`
```
mrice32, line 178: If we're using this to map fills too, shouldn't we specify the origin chain and deposit id here to disambiguate?

src/clients/SpokePoolClient.ts
```diff
@@ -136,18 +138,19 @@ export class SpokePoolClient {
...
+      ? depositWithMatchingDepositId
+      : undefined;
   }
 
   getValidUnfilledAmountForDeposit(deposit: Deposit): { unfilledAmount: BigNumber; fillCount: number } {
-    const fills = this.getFillsForOriginChain(deposit.originChainId)
-      .filter((fill) => fill.depositId === deposit.depositId) // Only select the associated fill for the deposit.
-      .filter((fill) => this.validateFillForDeposit(fill, deposit)); // Validate that the fill was valid for the deposit.
+    const fillsForDeposit = this.depositHashesToFills[this.getDepositHash(deposit)];
+    if (fillsForDeposit === undefined) return { unfilledAmount: toBN(deposit.amount), fillCount: 0 }; // If no fills then the full amount is remaining.
```
mrice32, line 150: Nit you could probably simplify this by adding || [] and relying on the return below rather than checking twice.

mrice32 (approved):
LGTM -- very simple!

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -116,6 +116,7 @@ export function initializeRunningBalancesFromRelayerRepayments(
             [repaymentChainId, l1TokenCounterpart],
             fillsToRefund[repaymentChainId][l2TokenAddress].totalRefundAmount
           );
+        else assign(runningBalances, [repaymentChainId, l1TokenCounterpart], toBN(0));
```
kevinuma, line 119: Why not always assign instead of doing if (fillsToRefund[repaymentChainId][l2TokenAddress].totalRefundAmount) ? totalRefundAmount is a require field so it should be 0 if there's no refund

kevinuma (approved):

PR #166, opened 2022-06-23
fix: Add cushion to enable MAX_RELAYER_DEPOSIT_LOOK_BACK to work

nicholaspai:
MAX_RELAYER_DEPOSIT_LOOK_BACK sets how far we look events up from, but this is a fixed number of blocks per chain, not a fixed number of time per chain. So, assuming the block lookback !== time lookback, there will be some false positive unfilled deposits seen by the relayer. Imagine a deposit on a chain that looks back 2 hours is filled by a relay on a destination chain that only looks back 1 hour, the relayer will think that deposit is unfilled when in fact it is.

This PR adds a cushion to the relayer to throw away any unfilled deposits older than half the block length of the `MAX_RELAYER_DEPOSIT_LOOK_BACK` value. This is a bit hacky but it works empirically for now and will allow us to speed up the bot a lot by limiting its lookback

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -126,10 +126,10 @@ export class MultiCallerClient {
           mrkdwn += `*Transaction excluded from batches because it contained value:*\n`;
           mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ` + `${transaction.mrkdwn || "0 mrkdwn"}\n`;
         });
-        Object.keys(groupedTransactions).forEach((chainId) => {
+        Object.keys(chunkedTransactions).forEach((chainId) => {
           mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
-          groupedTransactions[chainId].forEach((transactionChunks, groupIndex) => {
-            transactionChunks.forEach((transaction, chunkTxIndex) => {
+          chunkedTransactions[chainId].forEach((chunk, groupIndex) => {
+            chunk.forEach((transaction, chunkTxIndex) => {
               mrkdwn +=
                 `  ${groupIndex + 1}-${chunkTxIndex + 1}. ${transaction.message || "0 message"}: ` +
                 `${transaction.mrkdwn || "0 mrkdwn"}\n`;
@@ -204,7 +204,7 @@ export class MultiCallerClient {
         }
       });
       let flatIndex = 0;
-      Object.keys(chunkedTransactions).forEach((chainId, chainIndex) => {
+      Object.keys(chunkedTransactions).forEach((chainId) => {
         mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
         chunkedTransactions[chainId].forEach((chunk, chunkIndex) => {
           const settledPromise = multiCallTransactionsResult[flatIndex++];
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -2,7 +2,7 @@ import { BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDep
 import { createFormatFunction, etherscanLink, toBN } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
 
-import { Deposit, DepositWithBlock } from "../interfaces/SpokePool";
+import { Deposit } from "../interfaces/SpokePool";
 
 export class Relayer {
   constructor(
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -121,20 +121,27 @@ export class Relayer {
           destinationChain,
           true
         );
+        console.log(`!!!Fetching unfilledDepositsForDestinationChain`)
         const unfilledDepositsForDestinationChain: {
           fillCount: number;
           unfilledAmount: BigNumber;
           deposit: DepositWithBlock;
-        }[] = depositsForDestinationChain.map((deposit) => {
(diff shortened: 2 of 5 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -126,10 +126,10 @@ export class MultiCallerClient {
           mrkdwn += `*Transaction excluded from batches because it contained value:*\n`;
           mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ` + `${transaction.mrkdwn || "0 mrkdwn"}\n`;
         });
-        Object.keys(groupedTransactions).forEach((chainId) => {
+        Object.keys(chunkedTransactions).forEach((chainId) => {
```
nicholaspai, line 129: This fixes a bug that only appears in simulation mode

src/relayer/Relayer.ts
```diff
@@ -99,31 +105,6 @@ export class Relayer {
     }
   }
 
-  // Returns all unfilled deposits over all spokePoolClients. Return values include the amount of the unfilled deposit.
-  getUnfilledDeposits(): { deposit: Deposit; unfilledAmount: BigNumber; fillCount: number }[] {
```
nicholaspai, line 103: Unused function, very confusing!

src/utils/FillUtils.ts
```diff
@@ -162,10 +163,21 @@ export function getUnfilledDeposits(
...
-      const depositsForDestinationChain = originClient.getDepositsForDestinationChain(destinationChain);
-      const unfilledDepositsForDestinationChain = depositsForDestinationChain.map((deposit) => {
-        return { ...destinationClient.getValidUnfilledAmountForDeposit(deposit), deposit };
-      });
+      const depositsForDestinationChain: DepositWithBlock[] = originClient.getDepositsForDestinationChain(
+        destinationChain,
+        true
+      );
+      const unfilledDepositsForDestinationChain = depositsForDestinationChain
+        .filter((deposit) => {
```
nicholaspai, line 171: Before getting unfilled amount for deposit, filter out deposits older than configured lookback window

src/relayer/RelayerConfig.ts
```diff
@@ -13,7 +14,19 @@ export class RelayerConfig extends CommonConfig {
...
+
+    // `maxRelayerLookBack` is how far we fetch events from, modifying the search config's 'fromBlock'
     this.maxRelayerLookBack = MAX_RELAYER_DEPOSIT_LOOK_BACK ? JSON.parse(MAX_RELAYER_DEPOSIT_LOOK_BACK) : {};
+    // `maxRelayerUnfilledDepositLookBack` informs relayer to ignore any unfilled deposits older than this amount of
+    // of blocks from latest. This allows us to ignore any false positive unfilled deposits that occur because of how
+    // `maxRelayerLookBack` is set. This can happen because block lookback per chain is not exactly equal to the same
+    // amount of time looking back on the chains, so you might produce some deposits that look like they weren't filled.
+    this.maxRelayerUnfilledDepositLookBack = { ...this.maxRelayerLookBack };
+    Object.keys(this.maxRelayerUnfilledDepositLookBack).forEach((chain) => {
+      this.maxRelayerUnfilledDepositLookBack[chain] = Number(this.maxRelayerLookBack[chain]) / 4; // TODO: Allow caller
```
kevinuma, line 26: Why 4?

kevinuma (approved):

src/relayer/Relayer.ts
```diff
@@ -2,16 +2,22 @@ import { BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDep
...
+
   async checkForUnfilledDepositsAndFill(sendSlowRelays: Boolean = true) {
     // Fetch all unfilled deposits, order by total earnable fee.
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
-    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients).sort((a, b) =>
-      a.unfilledAmount.mul(a.deposit.relayerFeePct).lt(b.unfilledAmount.mul(b.deposit.relayerFeePct)) ? 1 : -1
+    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack).sort(
+      (a, b) =>
+        a.unfilledAmount.mul(a.deposit.relayerFeePct).lt(b.unfilledAmount.mul(b.deposit.relayerFeePct)) ? 1 : -1
```
mrice32, line 20: Shouldn't we be using the lookback _at the query level_? Or do we already do that?

src/relayer/RelayerConfig.ts
mrice32, line 26: In the PR description, it looks like you specified 2.

mrice32 (approved):
I do agree this is a little hacky, but probably okay for now, just to bring the pages down.

The thing that was a little confusing to me was that I don't see where this change applies to the event queries, but maybe that code is already there?

src/relayer/Relayer.ts
nicholaspai: we already do that [here](https://github.com/across-protocol/relayer-v2/blob/master/src/relayer/RelayerClientHelper.ts#L26) via the MAX_RELAYER_DEPOSIT_LOOKBACK config, but we've never used it before due to issues that this PR resolves

src/relayer/RelayerConfig.ts
nicholaspai: I think 4 is safer

mrice32 (approved):

PR #167, opened 2022-06-24
Limit lookback

kevinuma:
## Context:

All bots currently always load all deposits/fills events since the beginning of time. This makes them really slow, especially during the first few runs when the redis cache has not caught up. Even when caching kicks in, the bots always have to process all events, which is very slow. There are two options:
1. We can allow configuring the number of blocks to look back, similar to MAX_RELAYER_LOOKBACK. However this can lead to problems in areas where bundle boundaries need to be respected. For example refund calculation where we compare the refunds in a bundle with executed leave refunds because we could have missed fills due to the arbitrary block cutoff.
2. Look at the nth to last fully executed bundles and only fetch deposits and fills since then. This is the cleanest approach as it respects bundle boundaries and would not lead to missing deposits/fills that can result in problems.

## Solution
This PR goes with approach (2). It reuses the bundle lookback already used to calculate refunds in the monitor and relayer bots to limit the lookback window of SpokePoolClients.

## Testing
1. Unit test coverage
2. Running in simulation mode in a VM instance. Empirically this has visibly sped up the relayer bot.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -253,6 +253,16 @@ export class HubPoolClient {
     });
   }
 
+  // Return the nth fully executed bundle from the last.
+  getNthToLastFullyExecutedRootBundle(index: number, latestMainnetBlock: number): ProposedRootBundle | undefined {
+    const fullyExecutedBundles = sortEventsDescending(this.proposedRootBundles).filter(
+      (rootBundle: ProposedRootBundle) => {
+        return this.isRootBundleValid(rootBundle, latestMainnetBlock);
+      }
+    );
+    return fullyExecutedBundles[index];
+  }
+
   getNextBundleStartBlockNumber(chainIdList: number[], latestMainnetBlock: number, chainId: number): number {
     const latestFullyExecutedPoolRebalanceRoot = this.getLatestFullyExecutedRootBundle(latestMainnetBlock);
 
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -14,7 +14,7 @@ export class CommonConfig {
   readonly maxTxWait: number;
   readonly sendingTransactionsEnabled: boolean;
   readonly redisUrl: string | undefined;
-  readonly bundleRefundLookback: number;
+  readonly bundleLookback: number;
 
   constructor(env: ProcessEnv) {
     const {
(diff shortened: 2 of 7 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -17,14 +17,9 @@ export async function constructRelayerClients(logger: winston.Logger, config: Re
...
 
   const commonClients = await constructClients(logger, config);
+  await commonClients.hubPoolClient.update();
 
-  const spokePoolClients = await constructSpokePoolClientsWithLookback(
-    logger,
-    commonClients.configStoreClient,
-    config,
-    baseSigner,
-    config.maxRelayerLookBack
```
nicholaspai, line 26: So how does this PR affect the max relayer lookback for the relayer? Does it just replace that?

src/clients/HubPoolClient.ts
```diff
@@ -253,6 +253,16 @@ export class HubPoolClient {
     });
   }
 
+  // Return the nth fully executed bundle from the last.
+  getNthToLastFullyExecutedRootBundle(index: number, latestMainnetBlock: number): ProposedRootBundle | undefined {
+    const fullyExecutedBundles = sortEventsDescending(this.proposedRootBundles).filter(
```
nicholaspai, line 258: This is a bit slow as it runs on every root bundle. Why not keep a counter of how many valid bundles you've seen and then exit early when "n" is hit?

src/monitor/MonitorClientHelper.ts
```diff
@@ -20,12 +20,8 @@ export interface MonitorClients extends Clients {
 export async function constructMonitorClients(config: MonitorConfig, logger: winston.Logger): Promise<MonitorClients> {
   const baseSigner = await getSigner(); // todo: add getVoidSigner
   const commonClients = await constructClients(logger, config);
-  const spokePoolClients = await constructSpokePoolClientsWithLookback(
-    logger,
-    commonClients.configStoreClient,
-    config,
-    baseSigner
-  );
+  await commonClients.hubPoolClient.update();
```
nicholaspai, line 23: This is not obvious from the callers perspective that you need to up date the hubpool client in the construct function. Could you just update in the construct function? Or error out if it's not updated?

nicholaspai (commented):
My first instinct is that this isn't very clean to force the hub pool client to be updated first. Now it needs to be updated twice per server less run. But this is OK I think because the second run is very fast?

I do agree though that using bundle evaluation numbers as a lookback instead of hard coded blocks is better

src/common/Config.ts
```diff
@@ -26,7 +26,7 @@ export class CommonConfig {
       MAX_TX_WAIT_DURATION,
       SEND_TRANSACTIONS,
       REDIS_URL,
-      BUNDLE_REFUND_LOOKBACK,
+      BUNDLE_LOOKBACK,
```
nicholaspai, line 29: Two questions:
- Is `BUNDLE_REFUND_LOOKBACK` only used in monitor right now?
- Should we also remove `MAX_RELAYER_DEPOSIT_LOOKBACK`?

PR #168, opened 2022-06-24
feat: Reduce number of events searched per bot

nicholaspai:
- Finalizer shouldn't have to load Fill/Deposit events for example, while Relayer and Dataworker shouldn't have to load TokensBridged events
- This PR allows bots to define which events to lookup per SpokePoolClient update.

This PR also limits number of events searched in Finalizer, which is very slow since there is an async request per TokensBridged event to check on its finalization status

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -188,7 +188,7 @@ export class SpokePoolClient {
     return `${event.depositId}-${event.originChainId}`;
   }
 
-  async update(eventsToQuery: string[] = []) {
+  async update(eventsToQuery?: string[]) {
     if (this.configStoreClient !== null && !this.configStoreClient.isUpdated) throw new Error("RateModel not updated");
 
     this.latestBlockNumber = await this.spokePool.provider.getBlockNumber();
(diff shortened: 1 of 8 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -200,92 +211,117 @@ export class SpokePoolClient {
...
+      .map((eventName) => {
+        return {
+          filter: this._queryableEventNames()[eventName],
+          searchConfig:
+            eventName === "EnabledDepositRoute" || eventName === "TokensBridged"
+              ? depositRouteSearchConfig
+              : searchConfig,
+        };
+      })
+      .filter((x) => x.filter !== undefined);
```
mrice32, line 227: Why would this ever be undefined?

Instead of checking this here, maybe it makes more sense to just throw an error if the caller specifies an event that's not in the _queryableEventNames() list?

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
```diff
@@ -177,7 +188,7 @@ export class SpokePoolClient {
     return `${event.depositId}-${event.originChainId}`;
   }
 
-  async update() {
+  async update(eventsToQuery: string[] = []) {
```
mrice32, line 191: nit: the way this is constructed, if a user passes in `[]`, they will end up querying _all_ events. That might be a little confusing. Instead, maybe we should consider making this:
```ts
async update(eventsToQuery?: string[])
```

and then checking if it's undefined instead of length === 0.

That way an empty array will be respected, but a user who doesn't pass anything in defaults to all events.

It's pretty minor, but I think it's a slightly more intuitive interface.

src/clients/SpokePoolClient.ts
nicholaspai: hm good point,

src/clients/SpokePoolClient.ts
nicholaspai: good point

PR #175, opened 2022-06-27
improve: bump hardhat to 2.9.9

nicholaspai:
Hoping to resolve error where "missing revert data in call exception" gets thrown when simulating Optimism contract transactions that end up reverting


package.json
```diff
@@ -62,6 +62,7 @@
     "eslint-plugin-promise": "^5.1.0",
     "ethereum-waffle": "^3.0.0",
     "ethers": "^5.6.9",
+    "hardhat-deploy": "^0.11.10",
```
nicholaspai, line 65: Had to also explicitly import `harhdat-deploy` because `hardhat.config.js` was throwing compile errors that some properties in `networks` like `saveDeployments` and `companionNetworks` weren't part of `HardhatNetworkConfig`. These are added as part of `hardhat-deploy`, which I guess wasn't a problem until we upgraded hardhat?

PR #183, opened 2022-06-29
improve: Increase arb price bids to send L1 to L2 messages

nicholaspai:


src/clients/InventoryClient/ArbitrumAdapter.ts
```diff
@@ -26,13 +26,13 @@ const l2Gateways = {
 // wont get stuck. These are the same params we are using in the smart contracts.
 
 export class ArbitrumAdapter extends BaseAdapter {
-  l2GasPrice: BigNumber = toBN(5e9);
-  l2GasLimit: BigNumber = toBN(2000000);
+  l2GasPrice: BigNumber = toBN(200e9);
+  l2GasLimit: BigNumber = toBN(150000);
   // abi.encoding of the maxL2Submission cost. of 0.01e18
   transactionSubmissionData =
```
chrismaree, line 32: This needs to change as well?

chrismaree, line 32: This needs to change as well?

PR #189, opened 2022-06-30
improve: Use configurable follow distance in SpokePoolClient

nicholaspai:
If follow distance exists, then look up events only until HEAD-follow. This can be used specifically in looping mode to avoid any issues where this client sees incorrect events near HEAD, but can't re-load them since the Dataworker is running in looping mode. This could ultimately cause the dataworker to propose an incorrect root bundle. Even worst, the same looping instance will never consider this root bundle invalid. So, to mitigate this, use a follow distance in spoke clients  specifically in the dataworker when loading spoke pool events.

The follow distance is set to replace the BUNDLE_BLOCK_RANGE_BUFFER. Both block buffers do the same thing, but this PR changes the SpokePoolClient to follow HEAD instead of the Dataworker endblocks to follow the SpokePoolClient.latestBlockNumbers. This is an important distinction especially for running the dataworker in looping mode since it will only ever look up events twice. This is not a concern if running in serverless mode, but running the dataworker currently in serverless mode is very slow and not really usable.

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -308,20 +308,11 @@ export function getRunningBalanceForL1Token(transferThreshold: BigNumber, runnin
 export async function getWidestPossibleExpectedBlockRange(
   chainIdListForBundleEvaluationBlockNumbers: number[],
   spokePoolClients: { [chainId: number]: SpokePoolClient },
-  endBlockBuffers: number[],
   clients: DataworkerClients,
   latestMainnetBlock: number
 ): Promise<number[][]> {
-  const latestBlockNumbers = await Promise.all(
-    chainIdListForBundleEvaluationBlockNumbers.map(
-      async (chainId: number, index) =>
-        Math.max(0, spokePoolClients[chainId].latestBlockNumber - endBlockBuffers[index])
-      // We subtract a buffer from the end blocks to reduce the chance that network providers
-      // for different bot runs produce different contract state because of variability near the HEAD of the network.
-      // Reducing the latest block that we query also gives partially filled deposits slightly more buffer for relayers
-      // to fully fill the deposit and reduces the chance that the data worker includes a slow fill payment that gets
-      // filled during the challenge period.
-    )
+  const latestBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
+    (chainId: number) => spokePoolClients[chainId].latestBlockNumber
   );
   return chainIdListForBundleEvaluationBlockNumbers.map((chainId: number, index) => [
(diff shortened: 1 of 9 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -307,14 +307,15 @@ export function getRunningBalanceForL1Token(transferThreshold: BigNumber, runnin
...
+  spokePoolClients: { [chainId: number]: SpokePoolClient },
   endBlockBuffers: number[],
   clients: DataworkerClients,
   latestMainnetBlock: number
 ): Promise<number[][]> {
   const latestBlockNumbers = await Promise.all(
     chainIdListForBundleEvaluationBlockNumbers.map(
       async (chainId: number, index) =>
-        Math.max(0, (await clients.spokePoolSigners[chainId].provider.getBlockNumber()) - endBlockBuffers[index])
+        Math.max(0, spokePoolClients[chainId].latestBlockNumber - endBlockBuffers[index])
```
nicholaspai, line 318: This is a subtle fix but I think it makes the bot a bit safer: It just means that the proposed root bundle end blocks will only start at the latest provider's block number *as of the start of the Dataworker.ts instance* minus a buffer, instead of the provider's block number as of entering into this function, minus the buffer. This should have always b een in the bots, but the BUNDLE_BLOCK_RANGE_END_BUFFERS mitigated any potential errors with this

nicholaspai:
I've realized that this PR makes the `blockRangeEndBuffer` redundant if you set the `spokePoolClientFollowDistance`, so I'd opt to keep the latter and remove the former. Thoughts? @kevinuma

mrice32 (approved):

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -307,20 +307,12 @@ export function getRunningBalanceForL1Token(transferThreshold: BigNumber, runnin
...
-    chainIdListForBundleEvaluationBlockNumbers.map(
-      async (chainId: number, index) =>
-        Math.max(0, (await clients.spokePoolSigners[chainId].provider.getBlockNumber()) - endBlockBuffers[index])
-      // We subtract a buffer from the end blocks to reduce the chance that network providers
-      // for different bot runs produce different contract state because of variability near the HEAD of the network.
-      // Reducing the latest block that we query also gives partially filled deposits slightly more buffer for relayers
-      // to fully fill the deposit and reduces the chance that the data worker includes a slow fill payment that gets
-      // filled during the challenge period.
-    )
+  const latestBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
```
nicholaspai, line 314: No longer need to subtract buffer from latest when creating the "widest" block ranges, since the buffer is built into `latestBlockNumber`

PR #192, opened 2022-07-02
WIP: Dataworker should cache older Deposit/Fill events

nicholaspai:
The idea behind this PR is that the Dataworker should be able to cache events older than the latest validated root bundle's evaluation end blocks

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -442,7 +442,7 @@ export class SpokePoolClient {
   private async setDepositDataInCache(newData: CachedDepositData) {
     if (!this.configStoreClient.redisClient) return;
     const key = `deposit_data_${this.chainId}`;
-    // await this.configStoreClient.redisClient.set(key, JSON.stringify(newData));
+    await this.configStoreClient.redisClient.set(key, JSON.stringify(newData));
   }
 
   // Neccessary to use this function because RedisDB can only store strings so we need to stringify the BN objects
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,10 +1,14 @@
-import { spreadEvent, assign, Contract, BigNumber, EventSearchConfig, Promise, EventFilter } from "../utils";
+import { spreadEvent, assign, Contract, BigNumber, EventSearchConfig, Promise, EventFilter, sortEventsDescending, assert } from "../utils";
 import { toBN, Event, ZERO_ADDRESS, winston, paginatedEventQuery, spreadEventWithBlockNumber } from "../utils";
 
 import { AcrossConfigStoreClient } from "./ConfigStoreClient";
-import { Deposit, DepositWithBlock, Fill, SpeedUp, FillWithBlock, TokensBridged } from "../interfaces/SpokePool";
+import { Deposit, DepositWithBlock, DepositWithBlockInCache, Fill, SpeedUp, FillWithBlock, TokensBridged } from "../interfaces/SpokePool";
 import { RootBundleRelayWithBlock, RelayerRefundExecutionWithBlock } from "../interfaces/SpokePool";
 
+interface CachedDepositData {
+  latestBlock: number;
+  deposits: { [destinationChainId: number]: DepositWithBlockInCache[] };
+}
 export class SpokePoolClient {
   private deposits: { [DestinationChainId: number]: Deposit[] } = {};
   private depositHashes: { [depositHash: string]: Deposit } = {};
@@ -188,7 +192,7 @@ export class SpokePoolClient {
     return `${event.depositId}-${event.originChainId}`;
   }
 
-  async update(eventsToQuery?: string[]) {
+  async update(eventsToQuery?: string[], endBlockForChainInLatestFullyExecutedBundle = 0) {
     if (this.configStoreClient !== null && !this.configStoreClient.isUpdated) throw new Error("RateModel not updated");
 
     this.latestBlockNumber = await this.spokePool.provider.getBlockNumber();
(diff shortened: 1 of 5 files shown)


scripts/sendTokens.ts
```diff
@@ -46,7 +46,7 @@ export async function run(): Promise<void> {
     console.log(`Send ${symbol} with amount ${amountFromWei} tokens to ${recipient} on chain ${args.chainId}`);
     if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) process.exit(0);
     console.log("sending...");
-    const tx = await erc20.transfer(recipient, args.amount, { gasLimit: 100000 });
+    const tx = await erc20.transfer(recipient, args.amount);
```
nicholaspai, line 49: Random change I made that should have been made a while ago, unrelated to PR. Sometimes causes weird gas estimation issues with rpc

scripts/sendTokens.ts
```diff
@@ -46,7 +46,7 @@ export async function run(): Promise<void> {
     console.log(`Send ${symbol} with amount ${amountFromWei} tokens to ${recipient} on chain ${args.chainId}`);
     if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) process.exit(0);
     console.log("sending...");
-    const tx = await erc20.transfer(recipient, args.amount, { gasLimit: 100000 });
```
kevinuma, line 49: Why remove this gas limit now?

src/clients/SpokePoolClient.ts
```diff
@@ -211,18 +229,74 @@ export class SpokePoolClient {
     });
     if (searchConfig.fromBlock > searchConfig.toBlock) return; // If the starting block is greater than the ending block return.
 
+    // Try to reduce the number of web3 requests sent to query FundsDeposited and FilledRelay events, of which there
+    // expected to be many. We will first try to load existing cached events if they exist and if they do, then
+    // we only need to search for new events after the last cached event searched. We will then save any events
+    // older than the bundle end block for the latest validated bundle for this chain. These events by definition were
+    // included in a validated bundle so we can feel confident about them being correct.
+
+    // Invariant: cachedData is undefined if we don't want to use any events from the cache.
```
kevinuma, line 238: This would only be useful when running dataworker in looping mode right?

src/clients/SpokePoolClient.ts
```diff
@@ -211,18 +229,74 @@ export class SpokePoolClient {
...
+    let cachedData: {
```
kevinuma, line 239: Just to confirm, did you check memory usage? I assume it'd not be a problem since the deposits and fills are may be some tens of bytes, so for a machine with a few GBs of memory, we'd only run into problems if we have hundreds of millions of them.

kevinuma: Alternatively why not store and load these from redis?

src/clients/SpokePoolClient.ts
```diff
@@ -263,6 +361,44 @@ export class SpokePoolClient {
           [{ ...deposit, blockNumber: dataForQuoteTime[index].quoteBlock, originBlockNumber: event.blockNumber }]
         );
       }
+
+      // Update cache with deposit events <= bundle end block in latest fully executed bundle for this chain. We'll
+      // update the cache even if we didn't load any events from the cache because there is a chance that the cached
+      // data is corrupted (e.g. has stored events that are too recent).
+      if (endBlockForChainInLatestFullyExecutedBundle > 0) {
+        const depositsToCache = Object.fromEntries(
```
kevinuma, line 369: Would there be a problem where if the JSON RPC provider misses a few events? We'd then be caching incorrect events.

scripts/sendTokens.ts
nicholaspai: it should just be automatically fetched from rpc rather than hardcoded, lways should have been this way and i run into errors a lot with it

src/clients/SpokePoolClient.ts
nicholaspai: I don't understand the question? This cache is all from Redis. The memory usage should be a lot lower (but I'll check empirically) because its replacing a lot of web3 requests with a single redis retrieval + set and JSON.parsing/stringifying that data, which should be a lot cheaper.

src/clients/SpokePoolClient.ts
nicholaspai: Yes 100%, but that's why I think its a reasonable tradeoff to only store events older than bundle end block for latest validated bundle. These events would be much older than the buffer end block we use today to propose new bundles and even if they are missed, they wouldn't affect pending bundle validation (i.e the disputer or proposer) and only would affect relayer leaf execution

src/clients/SpokePoolClient.ts
kevinuma: I think I misunderstood this change earlier - all makes sense now.

kevinuma (approved):

mrice32 (approved):
LGTM! Only a few high-level comments. Looks awesome! Great idea.

src/clients/SpokePoolClient.ts
```diff
@@ -263,6 +329,42 @@ export class SpokePoolClient {
...
+              depositsToCacheCount: depositsToCache[destChain].length,
+            };
+          }),
+        });
+
+        // Save new deposit cache for chain.
+        await this.setDepositDataInCache({
+          latestBlock: latestBlockToCache,
+          deposits: depositsToCache,
+        });
```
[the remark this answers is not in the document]
nicholaspai: This is a good point i'll set an earliestBlock as well. I've added comments please take another look at `getEventsFromCache`

src/clients/SpokePoolClient.ts
```diff
@@ -347,6 +488,98 @@ export class SpokePoolClient {
...
+      const [_cachedDepositData, _cachedFillData] = await Promise.all([
+        this.getDepositDataFromCache(),
+        this.getFillDataFromCache(),
+      ]);
+
+      // Arbitrarily fetch latestCachedBlock from cached deposit list. We make an assumption that either deposit and
+      // fill event data is cached, or neither are, and they are cached up until the same end block. If they do not
+      // match, then we won't fetch any data from the cache.
+      const latestCachedBlock = _cachedDepositData.latestBlock ? Number(_cachedDepositData.latestBlock) : undefined;
+      if (latestCachedBlock === undefined || _cachedFillData.latestBlock !== latestCachedBlock) return undefined;
```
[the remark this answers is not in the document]
nicholaspai: Oversight good catch!

src/clients/SpokePoolClient.ts
```diff
@@ -347,6 +488,98 @@ export class SpokePoolClient {
     return this.configStoreClient.hubPoolClient;
   }
 
+  private async getEventsFromCache(latestBlockToCache: number, searchConfig: EventSearchConfig) {
```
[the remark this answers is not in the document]
nicholaspai: Added comments! Its designed to be the bundle end block of the latest validated bundle. But from the SpokePoolClient's perspective it is just hte latest block to store in cache

src/clients/SpokePoolClient.ts
mrice32: We could also run this with redundant providers.

mrice32 (approved):
LGTM! Thanks for the quick response to those comments!

src/clients/SpokePoolClient.ts
```diff
@@ -284,12 +386,51 @@ export class SpokePoolClient {
...
+        this.log("debug", `Saved cached fills for chain ${this.chainId}`, {
+          latestBlockToCache,
+          cachedFills: fillsToCache.length,
+        });
+
+        // Save new fill cache for chain.
+        await this.setFillDataInCache({
+          latestBlock: latestBlockToCache,
+          fills: fillsToCache,
+        });
```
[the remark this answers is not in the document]
nicholaspai: I added earliestBlock

PR #193, opened 2022-07-04
refactor: Move InventoryClient out into clients/ while keeping bridge adapters in clients/adapters/

kevinuma:
Small refactoring since we're using the bridge adapters and CrossChainTransferClient for more and more use cases beyond inventory management such as monitoring outstanding HubPool => SpokePool cross chain transfers (rebalances).


src/clients/bridges/AdapterManager.ts
```diff
@@ -54,18 +53,10 @@ export class AdapterManager {
     }
   }
 
-  getProvider(chainId: number): Provider {
```
kevinuma, line 57: Unused function

src/clients/bridges/AdapterManager.ts
```diff
@@ -54,18 +53,10 @@ export class AdapterManager {
...
-    return this.spokePoolClients[chainId].spokePool.provider;
-  }
-
   getSigner(chainId: number): Signer {
     return this.spokePoolClients[chainId].spokePool.signer;
   }
 
-  getChainSearchConfig(chainId: number) {
```
kevinuma, line 65: Unused function

nicholaspai (approved):

PR #194, opened 2022-07-04
feat: Update adaptermanager and bridge adapters to support multiple addresses

kevinuma:
We'll need this support to monitor cross chain transfers from HubPool to SpokePools (rebalances).

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -47,7 +47,9 @@ export class InventoryClient {
       this.tokenClient.getBalance(chainId, this.getDestinationTokenForL1Token(l1Token, chainId)) || toBN(0);
 
     // Consider any L1->L2 transfers that are currently pending in the canonical bridge.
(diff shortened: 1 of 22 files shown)


nicholaspai (commented):
This LGTM even though its a lot of code, which I think is unavaoidable since each adapter is so different... What are the consequences of merging this for the relayer bot? If there is a subtle bug will it affect the relayer bot and how bad could the consequences be?

kevinuma:
> This LGTM even though its a lot of code, which I think is unavaoidable since each adapter is so different... What are the consequences of merging this for the relayer bot? If there is a subtle bug will it affect the relayer bot and how bad could the consequences be?

The worst case scenario is that inventory management might stop working, either (1) cross-chain rebalances would fail or (2) it might oversend tokens to one chain if for some reason it's not correctly detecting pending transfers. I've run both the relayer and reporter in simulation mode and have not noticed anything wrong yet, but it's hard to test 100% in simulation mode.

nicholaspai (approved):
LGTM

PR #196, opened 2022-07-06
feat: Polygon finalizer should be able to finalize L2 withdrawals that occurred in same L2 multicall txn

nicholaspai:
Upgrades maticjs sdk to version that supports finalizing multiple burn events in same L2 transaction: https://github.com/maticnetwork/matic.js/releases/tag/v3.4.0-beta.0

### DIFF
--- a/src/finalizer/utils/polygon.ts
+++ b/src/finalizer/utils/polygon.ts
@@ -23,9 +23,9 @@ import { HubPoolClient } from "../../clients";
 const CHAIN_ID = 137;
 enum POLYGON_MESSAGE_STATUS {
   NOT_CHECKPOINTED = "NOT_CHECKPOINTED",
-  ALREADY_EXITED = "ALREADY_EXITED",
-  CAN_EXIT = "CAN_EXIT",
+  CHECK_POINTED = "CHECK_POINTED",
 }
+const BURN_SIG = "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
 
 export async function getPosClient(mainnetSigner: Wallet) {
   // Following from https://maticnetwork.github.io/matic.js/docs/pos
@@ -53,26 +53,15 @@ export async function getPosClient(mainnetSigner: Wallet) {
 export async function getFinalizableTransactions(
   logger: winston.Logger,
   tokensBridged: TokensBridged[],
-  posClient: POSClient,
-  hubPoolClient: HubPoolClient
+  posClient: POSClient
 ) {
   const isCheckpointed = await Promise.all(
     tokensBridged.map((event) => posClient.exitUtil.isCheckPointed(event.transactionHash))
   );
   const exitStatus = await Promise.all(
-    tokensBridged.map(async (event, i) => {
+    tokensBridged.map((_, i) => {
       if (!isCheckpointed[i]) return { status: POLYGON_MESSAGE_STATUS.NOT_CHECKPOINTED };
-      const l1TokenCounterpart = hubPoolClient.getL1TokenCounterpartAtBlock(
-        CHAIN_ID.toString(),
-        event.l2TokenAddress,
-        hubPoolClient.latestBlockNumber
-      );
-      return posClient
-        .erc20(l1TokenCounterpart, true)
-        .isWithdrawExited(event.transactionHash)
-        .then((result) =>
-          result ? { status: POLYGON_MESSAGE_STATUS.ALREADY_EXITED } : { status: POLYGON_MESSAGE_STATUS.CAN_EXIT }
-        );
+      else return { status: POLYGON_MESSAGE_STATUS.CHECK_POINTED };
     })
   );
   logger.debug({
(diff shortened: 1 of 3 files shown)


src/finalizer/utils/polygon.ts
```diff
@@ -23,8 +23,13 @@ import { HubPoolClient } from "../../clients";
 const CHAIN_ID = 137;
 enum POLYGON_MESSAGE_STATUS {
   NOT_CHECKPOINTED = "NOT_CHECKPOINTED",
-  ALREADY_EXITED = "ALREADY_EXITED",
   CAN_EXIT = "CAN_EXIT",
+  EXIT_ALREADY_PROCESSED = "EXIT_ALREADY_PROCESSED",
+}
+const BURN_SIG = "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
```
kevinuma, line 29: What is this for?

nicholaspai: i'll add a comment

src/finalizer/utils/polygon.ts
```diff
@@ -53,40 +58,69 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
 ) {
   const isCheckpointed = await Promise.all(
     tokensBridged.map((event) => posClient.exitUtil.isCheckPointed(event.transactionHash))
   );
+
+  // For each token bridge event, store a unique log index for the event within the transaction hash.
+  // This is important for bridge transactions containing multiple events.
+  const logIndexesForMessage = {};
+  const tokensBridgedWithLogIndex = tokensBridged.map((_tokensBridged, index) => {
+    if (!isCheckpointed[index]) return undefined;
```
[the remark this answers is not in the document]
nicholaspai: Refactored this a bit. I was using `undefined` because of how I was constructing the `exitStatus` object but I figured out a cleaner way to keep track of which events were not checkpointed yet. Also I've added better comments

src/finalizer/utils/polygon.ts
```diff
@@ -53,71 +60,109 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
+        // This is just a temporary workaround because there is no method in the sdk like isExitProcessed(txn, index).
+        await (await posClient.rootChainManager.getContract()).method("exit", payload).estimateGas({});
+        finalizableMessages.push({
+          ...tokensBridged[i],
+          payload,
+        });
+        return { status: POLYGON_MESSAGE_STATUS.CAN_EXIT };
+      } catch (err) {
+        return { status: POLYGON_MESSAGE_STATUS.EXIT_ALREADY_PROCESSED };
+      }
```
mrice32, line 109: nit: do you think it's worth parsing the error to ensure that this is actually the error to avoid us ignoring other types of issues?

If we do find a different type of issue, do you think it would be worth an error or warning?

src/finalizer/utils/polygon.ts
```diff
@@ -53,71 +60,109 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
-      at: "PolygonFinalizer",
-      message: "Error creating exitTx",
-      error,
-      notificationPath: "across-error",
-    });
+    if (error.reason.includes("EXIT_ALREADY_PROCESSED"))
+      logger.debug({
+        at: "PolygonFinalizer",
+        message: "Exit hash already processed",
+      });
```
mrice32, line 158: Shouldn't we not ever make it here if the exit was already processed? Wouldn't this error be unexpected?

src/finalizer/utils/polygon.ts
```diff
@@ -53,71 +60,109 @@ export async function getPosClient(mainnetSigner: Wallet) {
...
+    else
+      logger.debug({
+        at: "PolygonFinalizer",
+        message: "Error creating exitTx",
+        error,
+        notificationPath: "across-error",
+      });
```
mrice32, line 165: Do you think this is warn or error worthy? (it used to be a warning). Or do you think this will be noisy?

The reason I ask is because if we have some other type of error here and consistently miss a withdrawal, that could mean that money gets silently stuck, but not sure how noisy this would be.

mrice32 (approved):
Looks great! A few minor comments, but code looks great! Nice fix.

kevinuma (approved):

src/finalizer/utils/polygon.ts
nicholaspai: ohh true i was meant to remove this

src/finalizer/utils/polygon.ts
nicholaspai: good point done

src/finalizer/utils/polygon.ts
nicholaspai: its empirically very noise i think bc of re-using a nonce

nicholaspai: actually it used to be warn so ill move to warn

PR #197, opened 2022-07-08
feat(monitor): Report stuck HubPool -> SpokePool rebalances

kevinuma:
This PR does 2 things:
1. Reduce adapters' event query by reusing max relayer lookback. Before they were always fetching from hardcoded block numbers which were weeks ago from the time of this PR.
2. Add support to the reporter to report stuck rebalances (HubPool -> SpokePool cross chain transfers). This uses a very simple approximation - if there are any pending cross chain transfers from HubPool -> SpokePool more than a certain amount of time (Grace period) from the last bundle execution, those funds must be stuck on the native bridge.

### DIFF
--- a/src/clients/bridges/ArbitrumAdapter.ts
+++ b/src/clients/bridges/ArbitrumAdapter.ts
@@ -13,8 +13,6 @@ const l1Gateways = {
 
 const l1GatewayRouter = "0x72ce9c846789fdb6fc1f34ac4ad25dd9ef7031ef";
 
-const firstL1BlockToSearch = 12640865;
-
 const l2Gateways = {
   "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": "0x096760F208390250649E3e8763348E783AEF5562", // USDC
   "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0x6c411aD3E74De3E7Bd422b94A27770f5B86C623B", // WETH
@@ -38,7 +36,7 @@ export class ArbitrumAdapter extends BaseAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly monitoredAddresses: string[]
   ) {
-    super(spokePoolClients, 42161, firstL1BlockToSearch);
+    super(spokePoolClients, 42161);
   }
 
   async getOutstandingCrossChainTransfers(l1Tokens: string[]) {
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -1,13 +1,13 @@
 import { Provider } from "@ethersproject/abstract-provider";
 import { Signer } from "@ethersproject/abstract-signer";
 import { SpokePoolClient } from "../../clients";
-import { toBN, MAX_SAFE_ALLOWANCE, Contract, ERC20, BigNumber } from "../../utils";
+import { toBN, MAX_SAFE_ALLOWANCE, Contract, ERC20, EventSearchConfig, BigNumber } from "../../utils";
 import { etherscanLink, getNetworkName, MAX_UINT_VAL, runTransaction } from "../../utils";
 
 export class BaseAdapter {
   chainId: number;
-  l1SearchConfig;
-  l2SearchConfig;
+  l1SearchConfig: EventSearchConfig;
+  l2SearchConfig: EventSearchConfig;
   monitoredAddresses: string[];
   logger;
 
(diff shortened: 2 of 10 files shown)


src/monitor/Monitor.ts
```diff
@@ -32,6 +33,7 @@ import {
 import { MonitorClients, updateMonitorClients } from "./MonitorClientHelper";
 import { MonitorConfig } from "./MonitorConfig";
 
+export const REBALANCE_FINALIZE_GRACE_PERIOD = 60 * 60 * 2; // 2 hours.
```
nicholaspai, line 36: can we do 4 hours? polygon bridge takes up to 4 hours

src/monitor/Monitor.ts
```diff
@@ -168,6 +170,47 @@ export class Monitor {
...
+      for (const l1Token of allL1Tokens) {
+        const transferBalance = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
+          spokePoolAddress,
+          chainId,
+          l1Token.address
+        );
+
+        if (transferBalance.gt(0)) {
+          const mrkdwn = `Rebalances of ${l1Token.symbol} to ${getNetworkName(chainId)} is stuck`;
+          this.logger.error({ at: "Monitor", message: "HubPool -> SpokePool rebalances stuck 🦴", mrkdwn });
```
nicholaspai, line 208: should this be a warn?

src/monitor/Monitor.ts
```diff
@@ -168,6 +170,47 @@ export class Monitor {
...
+      const spokePoolAddress = this.clients.spokePoolClients[chainId].spokePool.address;
+      for (const l1Token of allL1Tokens) {
+        const transferBalance = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
+          spokePoolAddress,
+          chainId,
+          l1Token.address
+        );
+
+        if (transferBalance.gt(0)) {
+          const mrkdwn = `Rebalances of ${l1Token.symbol} to ${getNetworkName(chainId)} is stuck`;
```
nicholaspai, line 207: Is there a way we can log the L1 transaction hash that sent funds from L1 to L2 so that we can speed it up once we see this log? Usually we'd need the L1 hash in order to speed anything up or report it to customer service

nicholaspai (commented):
Left some comments, I'm not approving yet because of the leftover .only

src/monitor/Monitor.ts
kevinuma: I thought of that but it's really tricky to surface it due to the complex nested processing done via CrossChainTransferClient -> AdapterManager -> individual adapters. It can be a bit of a refactoring. I can do it in a separate followup PR.

src/monitor/Monitor.ts
kevinuma: Done. If we ever need to change this again in the future, we should make it an env config.

src/monitor/Monitor.ts
kevinuma: I was thinking we should page on this but warning works as well.

test/Monitor.ts
```diff
@@ -274,6 +259,51 @@ describe("Monitor", async function () {
     ).to.be.equal(toBN(5));
   });
 
+  it.only("Monitor should report stuck rebalances", async function () {
```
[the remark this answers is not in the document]
kevinuma: Removed. I keep forgetting these. I think there should be a way for lint to catch this automatically.

src/monitor/Monitor.ts
nicholaspai: i think warn is fine

test/Monitor.ts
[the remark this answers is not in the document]
nicholaspai: Yeah we should add to linter

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -13,8 +13,6 @@ const l1Gateways = {
 
 const l1GatewayRouter = "0x72ce9c846789fdb6fc1f34ac4ad25dd9ef7031ef";
 
-const firstL1BlockToSearch = 12640865;
```
nicholaspai, line 16: Why remove this? @chrismaree can you comment on why you had this originally?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -38,7 +33,12 @@ export class BaseAdapter {
   }
 
   getSearchConfig(chainId: number) {
-    return this.spokePoolClients[chainId].eventSearchConfig;
+    const spokePoolConfig = this.spokePoolClients[chainId].eventSearchConfig;
+    return {
+      fromBlock: spokePoolConfig ? spokePoolConfig.fromBlock : 0,
```
nicholaspai, line 38: Instead of 0 you can use the deployment block of the SpokePool: `getDeploymentBlockNumber("SpokePool", networkId)`

src/utils/ExecutionUtils.ts
```diff
@@ -16,7 +16,7 @@ export async function processCrash(logger: winston.Logger, fileName: String, pol
   logger.error({
     at: `${fileName}#index`,
     message: `There was an execution error! ${pollingDelay != 0 ? "Re-running loop" : ""}`,
-    error,
+    error: JSON.stringify(error),
```
nicholaspai, line 19: This isn't necessary, we log the `error` object in other places in the repo

src/clients/bridges/ArbitrumAdapter.ts
nicholaspai: Just talked with chris, sounds like these were deprecated you're right

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -38,7 +36,7 @@ export class ArbitrumAdapter extends BaseAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly monitoredAddresses: string[]
   ) {
-    super(spokePoolClients, 42161, firstL1BlockToSearch);
```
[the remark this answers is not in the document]
nicholaspai: agreeed!

nicholaspai (approved):

mrice32 (approved):
Looks great!

PR #198, opened 2022-07-08
improve: better flow control around computing realized lp fee

daywiss:

## Motivation
Dataworker bot ran extremely slowly in docker ( as well as my local system), effectively never completing a run even after waiting 40 or so minutes. 

## Summary
After investigation, I determined that we were making tons of duplicate calls to `getBlock("latest")`. The cause of this was using `promise.all` on all deposit events in the spoke pool client which called `computeRealizedLpFeePct`.  This call is expensive because it requires a block search for a timestamp. Because we have so many deposit events, we would initiate a block search in parallel, which would then all hit the initialization branch of the block finder, executing a high number of `getBlock("latest")`. This caused downstream issues, as then the block finder cache would not be effectively updated, and successive calls to lookup blocks would almost always miss.  This would also cause a ton of bandwith usage, seeing usage of 1-2gb network bandwith every 1-2 minutes.  

The solution here is to control the parallelization around `computeRealizedLpFeePct`, using `bluebird.mapSeries`, or `bluebird.map` with a limited concurrency. In testing using a full redis cache and `mapSeries`, the time to complete a run went from unable to complete, to around 6 minutes. Bandwith usage is goes down a lot after this change, since the amount of redundant calls is reduced to 0, easily under 1gb for an entire run, vs 1gb per minute previously. 

Running an empty redis cache is much slower, with times to complete on the order of 1 hour, but successive runs would be much faster. This PR allows you to control the concurrency if you want to try to optimize a new run.


src/clients/SpokePoolClient.ts
```diff
@@ -226,8 +226,9 @@ export class SpokePoolClient {
       };
     });
 
-    const queryResults = await Promise.all(
-      eventSearchConfigs.map((config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig)),
+    const queryResults = await Promise.map(
```
daywiss, line 229: this was using promise.all before, which does not have a concurrency option

src/utils/EventUtils.ts
```diff
@@ -45,7 +45,9 @@ export async function paginatedEventQuery(contract: Contract, filter: EventFilte
   const promises = paginatedRanges.map(([fromBlock, toBlock]) => contract.queryFilter(filter, fromBlock, toBlock));
 
   try {
-    return (await Promise.all(promises, { concurrency: searchConfig.concurrency | defaultConcurrency })).flat(); // Default to 200 concurrent calls.
+    return (
+      await Promise.map(promises, (x) => x, { concurrency: searchConfig.concurrency || defaultConcurrency })
```
daywiss, line 49: this was previously using promise.all which does not have a concurrency option

src/utils/ProviderUtils.ts
```diff
@@ -28,28 +23,17 @@ class RateLimitedProvider extends ethers.providers.JsonRpcProvider {
...
     // This sets up the queue. Each task is executed by calling the superclass's send method, which fires off the
     // request. This queue sends out requests concurrently, but stops once the concurrency limit is reached. The
     // maxConcurrency is configured here.
-    this.queue = createQueue(async ({ sendArgs, resolve, reject }: RateLimitTask) => {
-      await super
-        .send(...sendArgs)
-        .then(resolve)
-        .catch(reject);
-    }, maxConcurrency);
+    this.queue = createQueue(async ({ sendArgs }: RateLimitTask) => super.send(...sendArgs), maxConcurrency);
```
daywiss, line 29: this simplifies the queue to use `pushAsync` which automatically resolves when task completes. wrapping this in a new promise is not necessary

mrice32:
> Signed-off-by: david [david@umaproject.org](mailto:david@umaproject.org)
> 
> ## Motivation
> Dataworker bot ran extremely slowly in docker ( as well as my local system), effectively never completing a run even after waiting 40 or so minutes.
> 
> ## Summary
> After investigation, I determined that we were making tons of duplicate calls to `getBlock("latest")`. The cause of this was using `promise.all` on all deposit events in the spoke pool client which called `computeRealizedLpFeePct`. This call is expensive because it requires a block search for a timestamp. Because we have so many deposit events, we would initiate a block search in parallel, which would then all hit the initialization branch of the block finder, executing a high number of `getBlock("latest")`. This caused downstream issues, as then the block finder cache would not be effectively updated, and successive calls to lookup blocks would almost always miss. This would also cause a ton of bandwith usage, seeing usage of 1-2gb network bandwith every 1-2 minutes.
> 
> The solution here is to control the parallelization around `computeRealizedLpFeePct`, using `bluebird.mapSeries`, or `bluebird.map` with a limited concurrency. In testing using a full redis cache and `mapSeries`, the time to complete went to around 6 minutes. Bandwith usage is goes down a lot after this change, since the amount of redundant calls is reduced to 0, easily under 1gb for an entire run, vs 1gb per minute previously. Previously the redis cache would not fill consistently due to so many network failures and timeouts due to the massive number of 'latest block" calls.
> 
> Running an empty redis cache is much slower, with times to complete on the order of 1 hour, but successive runs would be much faster. This PR allows you to control the concurrency if you want to try to optimize a new run.

I think this is a good find, but there are a few things I want to get clarity on.

1. 
> Dataworker bot ran extremely slowly in docker ( as well as my local system), effectively never completing a run even after waiting 40 or so minutes.

At the end you then say:

> Running an empty redis cache is much slower, with times to complete on the order of 1 hour, but successive runs would be much faster. This PR allows you to control the concurrency if you want to try to optimize a new run.

My naiive reading says that unless you had pre-populated the redis cache in your initial run that was slow on your local machine, the end state was no faster than your initial run (despite being lower bandwidth).

Did you run in docker to see if this resulted in a speedup/reliability increase there? The dataworker, when running in GCP is running with a full cache, but it is slow _despite that_. We think this is because of docker. But if we're not testing given those same constraints, it's not clear that this is solving that problem.

2. Hardcoding a concurrency of 2 seems pretty extreme. If this is the solution, I suspect we can set this number much higher (and make it tunable) as to not always degrade no-cache load time performance.

mrice32:
Just wanted to add some information related to my local testing:

I was unable to reproduce the "slow docker" observations that others (maybe @evaldofelipe) found. Somewhat unrelated to this PR, but I think if that's the issue we're trying to fix, we'll need a clear testing/reproduction methodology.

src/clients/SpokePoolClient.ts
mrice32: Makes sense.

One question though -- do we need to limit concurrency here at all given that the provider limits it per-chain? Since this just filters down to direct JSON-RPC calls, parameterizing at the provider level might be a better way to attack the problem.

Note: I know you're just correcting a bug, but my suggestion is to maybe just remove the concurrency limit completely for this case. Thoughts? Or did you notice speedups or other measurable changes from making this change that you were unable to create by limiting provider concurrency?

cc @nicholaspai

src/clients/SpokePoolClient.ts
```diff
@@ -246,8 +247,10 @@ export class SpokePoolClient {
         this.log("debug", `Fetching realizedLpFeePct for ${depositEvents.length} deposits on chain ${this.chainId}`, {
           numDeposits: depositEvents.length,
         });
-      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number }[] = await Promise.all(
-        depositEvents.map((event) => this.computeRealizedLpFeePct(event))
+      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number }[] = await Promise.map(
+        depositEvents,
+        (event) => this.computeRealizedLpFeePct(event),
+        { concurrency: 2 }
```
mrice32, line 253: Thoughts on bumping this to ~10 or so?

src/utils/ProviderUtils.ts
mrice32: Nice! Just want to double check that you've tested that this propagates return values and errors the same as before? (just asking because we have no unit tests for this class)

mrice32 (commented):
Looks awesome -- very nice find on the block finder. A few comments, mostly on the other fixes.

src/clients/SpokePoolClient.ts
daywiss: i can try it without limiting the concurrency at all, it might not be needed

src/clients/SpokePoolClient.ts
```diff
@@ -226,8 +226,9 @@ export class SpokePoolClient {
       };
     });
 
-    const queryResults = await Promise.all(
-      eventSearchConfigs.map((config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig)),
+    const queryResults = await Promise.map(
+      eventSearchConfigs,
+      (config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig),
       { concurrency: 2 }
```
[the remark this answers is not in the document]
daywiss: yea basically

src/clients/SpokePoolClient.ts
nicholaspai: Agreed with Matt this might be redundant. How did it work?

src/clients/SpokePoolClient.ts
nicholaspai: Can we try removing this like the one above since the provider will limit concurrency?

nicholaspai (commented):
Thanks for reviewing @mrice32 and thanks for responding @daywiss .

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
daywiss: im playing with this number, but during testing increasing this number had no performance impact when your redis is fully cached. i do think it will impact performance on non fully cached instance

daywiss:
> Did you run in docker to see if this resulted in a speedup/reliability increase there? The dataworker, when running in GCP is running with a full cache, but it is slow despite that. We think this is because of docker. But if we're not testing given those same constraints, it's not clear that this is solving that problem.

I just did a comparison of runtime between 2 docker containers with a fully cached redis instance.  With these code changes and concurrency set to 100 on the call to computeRealizedLpFee. It completed in around 7 minutes in a container, while running it with the original code did not complete in 30 minutes. both used the same cache.

i'm going to experiment some more with the concurrency as well as removing unecessary promise.maps.

daywiss:
@mrice32 I just a new tests both were run with full redis cache. The first run is the latest docker image, it did not complete in 30+ minutes still. I changed my local build with concurrency of 1000 (same as NODE_CONCURRENCY), removed all the other promise.maps to promise.alls. This version completed in about 7 minutes.  I think this basically shows the bottleneck is in the concurrent `computeRealizedLpFee` calls

src/utils/ProviderUtils.ts
daywiss: i reset this to the original implementation since i cant really test this

nicholaspai (approved):
This is awesome work and a small potentially high leverage PR! My only comment is whether we can use `NODE_MAX_CONCURRENCY` instead of hardcoding concurrency: 1000?

src/clients/ConfigStoreClient.ts
```diff
@@ -202,7 +202,7 @@ export class AcrossConfigStoreClient {
 
   // Avoid caching calls that are recent enough to be affected by things like reorgs.
   private shouldCache(eventTimestamp: number) {
-    // Current time must be > 5 minutes past the event timestamp for it to be stable enough to cache.
-    return getCurrentTime() - eventTimestamp > 300;
+    // Current time must be >= 5 minutes past the event timestamp for it to be stable enough to cache.
+    return getCurrentTime() - eventTimestamp >= 300;
```
mrice32, line 206: Not a problem, just curious why this was changed? Seems somewhat unrelated/arbitrary?

Totally fine to change if you prefer it this way -- just struck me as odd.

src/clients/SpokePoolClient.ts
```diff
@@ -246,8 +245,12 @@ export class SpokePoolClient {
...
           numDeposits: depositEvents.length,
         });
-      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number }[] = await Promise.all(
-        depositEvents.map((event) => this.computeRealizedLpFeePct(event))
+      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number }[] = await Promise.map(
+        depositEvents,
+        async (event) => {
+          return this.computeRealizedLpFeePct(event);
+        },
+        { concurrency: 1000 }
```
mrice32, line 253: Surprised you can get away with it being this high, but 🤷

mrice32 (approved):
LGTM!

src/clients/ConfigStoreClient.ts
daywiss: oddly was looking into why we werent caching certain requests, and a majority of them fell exactly on this 300 second difference. i dont really understand how this is possible, but wanted to make it more permissive to cache more blocks.

src/clients/SpokePoolClient.ts
daywiss: i was just matching the node_max_concurrency to see if that was reasonable, seems like it is

daywiss:
> This is awesome work and a small potentially high leverage PR! My only comment is whether we can use NODE_MAX_CONCURRENCY instead of hardcoding concurrency: 1000?

@nicholaspai I updated it to do this

PR #202, opened 2022-07-11
improve: Add eslint to lint-fix script

nicholaspai:
Motivation was to make sure we don't commit .only tests but realized this is caught by eslint which we don't use in CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "AGPL-3.0",
   "private": true,
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=12.9.0"
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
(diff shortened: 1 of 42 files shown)


kevinuma (approved):

package.json
```diff
@@ -7,25 +7,34 @@
...
-    "node": ">=8.3.0"
+    "node": ">=12.9.0"
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
     "@across-protocol/sdk-v2": "^0.1.10",
     "@arbitrum/sdk": "^1.1.4",
     "@defi-wonderland/smock": "^2.0.7",
     "@eth-optimism/sdk": "^1.1.5",
+    "@ethersproject/abstract-provider": "^5.6.1",
```
mrice32, line 18: Where did all of these new packages come from? Were they auto-added by the linter?

mrice32 (approved):
LGTM!

package.json
nicholaspai: THey weren't added before so i'm guessing they were imported as sub dependencies and were in node_Modules? eslint complained though suggesting it needed to be addeed to package.json. We import these directly

PR #203, opened 2022-07-11
improve(inventory): support UMA bridging to Polygon

kevinuma:
Currently UMA is not listed as one of the tokens that relayer's inventory management can bridge to Polygon


src/clients/bridges/PolygonAdapter.ts
```diff
@@ -37,6 +37,13 @@ const tokenToBridge = {
     l1AmountProp: "amount",
     l2AmountProp: "value",
   }, // WBTC
+  "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828": {
```
nicholaspai, line 40: Is this not neededon arbitrum/poly?

nicholaspai (approved):
1 question

src/clients/bridges/PolygonAdapter.ts
kevinuma: Added for Arbitrum. Optimism/Boba are not needed as it's all through the standard ERC 20 bridge

PR #204, opened 2022-07-12
fix(inventory): Add missing l1 -> l2 token mappings for UMA, BADGER, BOBA

kevinuma:
Inventory checks this map before sending rebalances.


src/common/Constants.ts
```diff
@@ -49,6 +49,15 @@ export const l2TokensToL1TokenValidation = {
...
     42161: "0x2f2a2543B76A4166549F7aaB2e75Bef0aefC5B0f",
   }, // WBTC
+  "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828": {
+    10: "0xE7798f023fC62146e8Aa1b36Da45fb70855a77Ea",
+    137: "0x3066818837c5e6eD6601bd5a91B0762877A6B731",
+    288: "0x780f33Ad21314d9A1Ffb6867Fe53d48a76Ec0D16",
+    42161: "0xd693Ec944A85eeca4247eC1c3b130DCa9B0C3b22",
+  }, // UMA
+  "0x42bBFa2e77757C645eeaAd1655E0911a7553Efbc": {
+    288: "0xa18bF3994C0Cc6E3b63ac420308E5383f53120D7",
```
kevinuma, line 59: No entries for BOBA on other chains are expected. This means if inventory ever incorrectly tries to send BOBA to those chains (it doesn't exist on those chains), the code would throw errors since it verifies with this map.

src/common/Constants.ts
```diff
@@ -49,6 +49,19 @@ export const l2TokensToL1TokenValidation = {
...
```
kevinuma, line 59: No entries for BOBA on other chains are expected. This means if inventory ever incorrectly tries to send BOBA to those chains (it doesn't exist on those chains), the code would throw errors since it verifies with this map.

nicholaspai (approved):

mrice32 (approved):
LGTM! This is way simpler than I thought it would be -- great work!

kevinuma:
> LGTM! This is way simpler than I thought it would be -- great work!

It took some time to get rid of the warnings due to missing entries for Boba and Badger on a few chains but yeah overall things are smooth I think. Need to test with small amounts of real rebalances though.

PR #206, opened 2022-07-12
improve: Update README

nicholaspai:


mrice32 (approved):

mrice32 (approved):
This is an awesome start! Thanks for adding this.

README.md
```diff
@@ -1 +1,21 @@
 # Across V2 Relayer
+
+This code implements [UMIP-157](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md) and interacts with these [smart contracts](https://github.com/across-protocol/contracts-v2). The contracts were audited [by OpenZeppelin](https://blog.openzeppelin.com/uma-across-v2-audit/).
+
+# Installation
+
+```sh
+# clone
+gh repo clone across-protocol/relayer-v2
```
mrice32, line 9: nit: prefer to use normal git clone syntax rather than the gh package, since git is more standard.

Also, since this is in the readme of this repository, I think we can assume they've already cloned the package, so we might be able to drop this step entirely. Thoughts?

PR #209, opened 2022-07-13
improve: stop notifying of unprofitable fills after a certain amount of time to avoid spamming

kevinuma:
Currently we always notify when there's an unprofitable fill for as long as the relayer's max lookback window (currently ~2 days). This is very spammy.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1,8 +1,10 @@
-import { BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDeposits } from "../utils";
+import {BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDeposits, getCurrentTime} from "../utils";
 import { createFormatFunction, etherscanLink, toBN } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
 
 import { Deposit } from "../interfaces/SpokePool";
+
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -1,9 +1,11 @@
-import { BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDeposits } from "../utils";
+import {BigNumber, winston, buildFillRelayProps, getNetworkName, getUnfilledDeposits, getCurrentTime} from "../utils";
 import { createFormatFunction, etherscanLink, toBN } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
 
 import { Deposit } from "../interfaces/SpokePool";
 
+const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
```
nicholaspai, line 7: Instead of hardcoding this number I think you should just use a fraction of the `MAX_RELAYER_DEPOSIT_LOOK_BACK`, or just use the `maxRelayerUnfilledDepositLookBack` config

kevinuma: Switched to a mix of hardcoded 1 hour and maxRelayerUnfilledDepositLookBack

nicholaspai (approved):

PR #210, opened 2022-07-14
improve(relayer): Add env config to allow specifying which tokens to relay

kevinuma:


src/relayer/Relayer.ts
```diff
@@ -11,16 +12,18 @@ export class Relayer {
     readonly relayerAddress: string,
     readonly logger: winston.Logger,
     readonly clients: RelayerClients,
-    readonly maxUnfilledDepositLookBack: { [chainId: number]: number } = {}
+    readonly config: RelayerConfig = undefined
```
nicholaspai, line 15: Why not default this to {} instead of undefined?

src/relayer/Relayer.ts
```diff
@@ -11,16 +12,18 @@ export class Relayer {
...
   async checkForUnfilledDepositsAndFill(sendSlowRelays = true) {
     // Fetch all unfilled deposits, order by total earnable fee.
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
-    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack).sort(
-      (a, b) =>
-        a.unfilledAmount.mul(a.deposit.relayerFeePct).lt(b.unfilledAmount.mul(b.deposit.relayerFeePct)) ? 1 : -1
+    const unfilledDeposits = getUnfilledDeposits(
+      this.clients.spokePoolClients,
+      this.config ? this.config.maxRelayerUnfilledDepositLookBack : {}
```
nicholaspai, line 24: I think we should probably load these config props like `maxRelayerUnfilledDepositLookBack` in the constructor

src/relayer/Relayer.ts
```diff
@@ -31,17 +34,25 @@ export class Relayer {
     // is has no other fills then send a 0 sized fill to initiate a slow relay. If unprofitable then add the
     // unprofitable tx to the unprofitable tx tracker to produce an appropriate log.
     for (const { deposit, unfilledAmount, fillCount } of unfilledDeposits) {
+      // Skip any L1 tokens that are not specified in the config.
+      // If relayerTokens is an empty list, we'll assume that all tokens are supported.
+      const l1Token = this.clients.hubPoolClient.getL1TokenInfoForL2Token(deposit.originToken, deposit.originChainId);
+      if (
+        this.config !== undefined &&
+        this.config.relayerTokens.length > 0 &&
```
nicholaspai, line 42: Same comment [here](https://github.com/across-protocol/relayer-v2/pull/210/files#r921349453) would make this if block a lot shorter

src/relayer/Relayer.ts
kevinuma: It's a bit redundant though but I don't feel too strongly so I can do that.

src/relayer/Relayer.ts

src/relayer/Relayer.ts
```diff
@@ -31,17 +33,22 @@ export class Relayer {
     // is has no other fills then send a 0 sized fill to initiate a slow relay. If unprofitable then add the
     // unprofitable tx to the unprofitable tx tracker to produce an appropriate log.
     for (const { deposit, unfilledAmount, fillCount } of unfilledDeposits) {
+      // Skip any L1 tokens that are not specified in the config.
+      // If relayerTokens is an empty list, we'll assume that all tokens are supported.
+      const l1Token = this.clients.hubPoolClient.getL1TokenInfoForL2Token(deposit.originToken, deposit.originChainId);
+      if (this.relayerTokens.length > 0 && !this.relayerTokens.includes(l1Token.address)) {
```
nicholaspai, line 39: Is checking .length > 0 redundant? doesn't .includes() just return false if array is empty?

src/relayer/Relayer.ts
```diff
@@ -11,6 +12,7 @@ export class Relayer {
     readonly relayerAddress: string,
     readonly logger: winston.Logger,
     readonly clients: RelayerClients,
+    readonly relayerTokens: string[] = [],
     readonly maxUnfilledDepositLookBack: { [chainId: number]: number } = {}
```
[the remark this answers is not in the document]
nicholaspai: This way we can comfortably merge this PR into master without worrying about breaking bots in prod

src/relayer/Relayer.ts
kevinuma: This is explained in the comment above:  If relayerTokens is an empty list, we'll assume that all tokens are supported. So we should not skip any deposits if it's empty.

src/relayer/Relayer.ts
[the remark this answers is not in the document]
kevinuma: Good point. Updated.

PR #211, opened 2022-07-15
feat(relayer): Allow relayer to configure which destination chains to fill deposits for

kevinuma:
Currently relayer can't specify that they only want to fill deposits with specific destination chains.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -41,14 +41,14 @@ export class Relayer {
       if (
         this.relayerTokens.length > 0 &&
         !this.relayerTokens.includes(l1Token.address) &&
-        this.relayerTokens.includes(l1Token.address.toLowerCase())
+        !this.relayerTokens.includes(l1Token.address.toLowerCase())
       ) {
         this.logger.debug({ at: "Relayer", message: "Skipping deposit for unwhitelisted token", deposit, l1Token });
         continue;
       }
 
       const destinationChainId = deposit.destinationChainId;
-      if (this.relayerDestinationChains.length > 0 && this.relayerDestinationChains.includes(destinationChainId)) {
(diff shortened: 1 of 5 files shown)


src/relayer/Relayer.ts
```diff
@@ -40,12 +41,23 @@ export class Relayer {
       if (
         this.relayerTokens.length > 0 &&
         !this.relayerTokens.includes(l1Token.address) &&
-        this.relayerTokens.includes(l1Token.address.toLowerCase())
+        !this.relayerTokens.includes(l1Token.address.toLowerCase())
```
nicholaspai, line 44: I should have caught this dang but we need to reduce these issues somehow

nicholaspai (commented):
I want to understand something first. How important is this feature? Isn't restricting tokens enough and choosing which networks to keep balance on already restricting which networks to fill on?

src/relayer/Relayer.ts
```diff
@@ -12,7 +12,8 @@ export class Relayer {
     readonly logger: winston.Logger,
     readonly clients: RelayerClients,
     readonly maxUnfilledDepositLookBack: { [chainId: number]: number } = {},
-    readonly relayerTokens: string[] = []
+    readonly relayerTokens: string[] = [],
+    readonly relayerDestinationChains: number[] = []
```
nicholaspai, line 16: Doesn't have to be in this PR but thoughts on inporting the full RelayerConfig instead of adding individual config vars?

nicholaspai (approved):
LGTM just one suggestion for a follow up PR

src/relayer/Relayer.ts
kevinuma: This is missing unit tests I think so definitely need to stricter with unit tests.

PR #212, opened 2022-07-15
feat(profitability): Remove hardcoded min revenue map and add logic to consider gas cost

kevinuma:


src/clients/MultiCallerClient.ts
```diff
@@ -26,7 +26,7 @@ export interface AugmentedTransaction {
 export class MultiCallerClient {
   private transactions: AugmentedTransaction[] = [];
   // eslint-disable-next-line no-useless-constructor
-  constructor(readonly logger: winston.Logger, readonly gasEstimator: any, readonly maxTxWait: number = 180) {}
```
kevinuma, line 29: Unused vars

src/clients/ProfitClient.ts
```diff
@@ -6,7 +6,7 @@ import { Coingecko } from "@uma/sdk";
 // Define the minimum revenue, in USD, that a relay must yield in order to be considered "profitable". This is a short
 // term solution to enable us to avoid DOS relays that yield negative profits. In the future this should be updated
 // to actual factor in the cost of sending transactions on the associated target chains.
-const chainIdToMinRevenue = {
+const MIN_REVENUE_BY_CHAIN_ID = {
```
kevinuma, line 9: nit: Constants should be ALL_CAPS

src/clients/ProfitClient.ts
```diff
@@ -22,6 +22,18 @@ const chainIdToMinRevenue = {
...
+// We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
+const WMATIC = "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270";
+const GAS_TOKEN_BY_CHAIN_ID = {
+  1: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  10: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  137: WMATIC,
+  288: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  42161: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+};
+// TODO: Make this dynamic once we support chains with gas tokens that have different decimals.
```
nicholaspai, line 34: I hope there is no world where a chain has a non-18 decimal gas token but sadly this will probably be the case

src/clients/ProfitClient.ts
```diff
@@ -41,7 +54,7 @@ export class ProfitClient {
 
   getPriceOfToken(token: string) {
     if (!this.tokenPrices[token]) {
-      this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
+      this.logger.debug({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
```
kevinuma, line 57: This can be noisy when coingecko apis keep failing

src/clients/ProfitClient.ts
```diff
@@ -69,10 +82,19 @@ export class ProfitClient {
     const tokenPriceInUsd = this.getPriceOfToken(l1Token);
     const fillRevenueInRelayedToken = toBN(deposit.relayerFeePct).mul(fillAmount).div(toBN(10).pow(decimals));
     const fillRevenueInUsd = fillRevenueInRelayedToken.mul(tokenPriceInUsd).div(toBNWei(1));
+
+    // Consider gas cost.
+    const gasCostInUsd = gasUsed
+      .mul(this.getPriceOfToken(GAS_TOKEN_BY_CHAIN_ID[deposit.destinationChainId]))
```
mrice32, line 88: Don't we need to compute the current gas price to take into account fluctuations in relayer costs? I think most chains have gas costs that can vary by up to 10x week-to-week, so I don't think hardcoding USD revenue will be sufficient to match the dynamic estimations being done by the frontend.

Note: maybe I'm missing the point of this PR! Feel free to clarify in the description. I'm just going off of the title.

src/clients/ProfitClient.ts
```diff
@@ -98,26 +123,48 @@ export class ProfitClient {
   }
 
   async update() {
-    const l1Tokens = this.hubPoolClient.getL1Tokens();
+    // Clone because we'll be adding WMATIC to the list for logging purposes later below.
+    const l1Tokens = _.cloneDeep(this.hubPoolClient.getL1Tokens());
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    const prices = await Promise.allSettled(l1Tokens.map((l1Token: L1Token) => this.coingeckoPrice(l1Token.address)));
+    const priceFetches = l1Tokens.map((l1Token: L1Token) => this.coingeckoPrice(l1Token.address));
```
mrice32, line 129: Have you considered trying to use the same SDK functionality as the FE to compute the expected relayer fee?

One failure mode that we saw in across v1 when we had separate implementations of the expected fee was that our frontend would produce transactions that our bot would then feel were underpriced or vice-versa. This isn't so bad if the bot's estimation is lower than the frontend, but when the frontend estimated lower than the bot, we would end up not relaying the transaction, creating a bad UX and support tickets where people ask when their money will arrive.

The strategy that we used was to use the same library to compute the estimated relayer fee in both places and get the FE to charge a configurable percentage more than the bot to ensure that small swings in gas prices between the FE estimation and the bot seeing the deposit wouldn't cause a deposit to get stuck in limbo.

Thoughts?

mrice32 (commented):
This is awesome work! A few high-level comments.

src/clients/ProfitClient.ts
kevinuma: gasUsed already includes the gas fee (gasUsed = gas cost * gas fee) as passed in from Relayer.

src/clients/ProfitClient.ts
kevinuma: My next followup is to update the SDK. I think the SDK would need a slightly different implementation for including gas costs in the relayer fee. We currently have a lot of small deposits with single digit in cents of profits. So the SDK seems to be underpricing some deposits. We'll need to add a gas buffer amount in SDK in case of gas spikes.

src/clients/ProfitClient.ts
```diff
@@ -2,36 +2,26 @@ import { BigNumber, winston, toBNWei, toBN, assign } from "../utils";
 import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
 import { Coingecko } from "@uma/sdk";
+import * as _ from "lodash";
 
-// Define the minimum revenue, in USD, that a relay must yield in order to be considered "profitable". This is a short
-// term solution to enable us to avoid DOS relays that yield negative profits. In the future this should be updated
-// to actual factor in the cost of sending transactions on the associated target chains.
-const chainIdToMinRevenue = {
```
kevinuma, line 9: Context: https://risklabs.slack.com/archives/D03GKFYCDJ7/p1657999254013629. tldr; is this hardcoded limit is filtering out a lot of small deposits with profits of < $1. When running with this previous logic, my relayer was skipping 9/10 deposits. We'll follow up with data analysis to understand profitability of small deposits.

src/clients/ProfitClient.ts
mrice32: Hmmm, maybe we're just using different language to describe the same thing.

I'm talking about the USD costs, which would amount to gasUsed * gasPrice * gasTokenPrice. I don't think this code takes `gasPrice` into account, so if the gas price swings very high, we could relay unprofitable deposits, and if it swings low, we could block profitable deposits.

I think if we want this code to correctly take into account the true cost of relaying a transaction, we need to estimate the gas cost that will be paid.

Do you agree with that assessment?

src/clients/ProfitClient.ts
mrice32: > I think the SDK would need a slightly different implementation for including gas costs in the relayer fee.

Why do you think the computation in the frontend should be different from the relayer? How do you think they should differ, specifically?

> We currently have a lot of small deposits with single digit in cents of profits.

Is there anything wrong with profiting a few cents on a small, cheap transaction? For instance, if the gas is $0.001 and the capital required is $0.50, a profit of $0.005 seems pretty reasonable, no?

> We'll need to add a gas buffer amount in SDK in case of gas spikes.

Agreed, a buffer percentage makes a lot of sense! Hopefully, this could just be a difference in parameterization of the same computation.

> My next followup is to update the SDK.

Rather than reinventing the wheel, why not start with integrating the [SDK implementation](https://github.com/across-protocol/sdk-v2/blob/7905929cd6971623a1b51143d66bbe814e28bf94/src/relayFeeCalculator/relayFeeCalculator.ts#L74-L82) that's running in prod and takes into account gas prices and configurable capital costs and try to fit that into this implementation, fixing any bugs or incompatibilities we find along the way?

This has the added advantage of ensuring that the results of both computations will match up every step along the way.

nicholaspai: > I think the SDK would need a slightly different implementation for including gas costs in the relayer fee.

Why is this?

src/clients/ProfitClient.ts
```diff
@@ -2,36 +2,26 @@ import { BigNumber, winston, toBNWei, toBN, assign } from "../utils";
...
-  // Rollups/L2s/sidechains & their testnets.
-  10: toBNWei(1),
-  69: toBNWei(1),
-  288: toBNWei(1),
-  28: toBNWei(1),
-  42161: toBNWei(1),
-  137: toBNWei(1),
-  80001: toBNWei(1),
+// We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
+export const WMATIC = "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270";
```
nicholaspai, line 8: Nit: you already hardcode WETH in multiple places so why not add this to the below hardcoded mapping and just leave a comment "// WMATIC"

src/utils/TransactionUtils.ts
```diff
@@ -63,6 +63,23 @@ export async function willSucceed(
...
 }
 
+export async function estimateGas(
+  transaction: AugmentedTransaction
+): Promise<{ transaction: AugmentedTransaction; succeed: boolean; reason: string; gasUsed: BigNumber }> {
+  try {
+    const args = transaction.value ? [...transaction.args, { value: transaction.value }] : transaction.args;
+    const [gasUsed, gasFee] = await Promise.all([
+      transaction.contract.estimateGas[transaction.method](...args),
+      transaction.contract.provider.getGasPrice(),
```
nicholaspai, line 73: Can you get this gas price once per relayer loop and then re use it for all invocations of estimateGas? It seems like this would be a low risk way to reduce requests to the RPC. 

Also, is provider.getGasPrice the same method that is used by runTransaction to get gas price? Do we not use a GasEstimatorClient? I only ask because we should be consistent with how we query gas price

nicholaspai (commented):
Nice PR left some comments/questions. Definitely think we should try to re use the sdk-v2 if possible!

src/clients/ProfitClient.ts
```diff
@@ -22,6 +22,18 @@ const chainIdToMinRevenue = {
   80001: toBNWei(1),
 };
 
+// We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
+const WMATIC = "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270";
```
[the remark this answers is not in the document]
nicholaspai: ohh sorry I missed this comment. Sounds good

src/clients/ProfitClient.ts
nicholaspai: I'm good with this, coingecko API seems to fail a lot. Any ideas for a fallback strategy?

src/clients/ProfitClient.ts
kevinuma: That's what I was talking about. That multiplication you're looking for is done in https://github.com/across-protocol/relayer-v2/pull/212/files#diff-562d70c5c8a2a3e8074947c1f599ce3c57b2abd4cdff462422b5bc9fa4257e9fR76

src/clients/ProfitClient.ts
kevinuma: > > Why do you think the computation in the frontend should be different from the relayer? How do you think they should differ, specifically?

I think the bot should potentially have a different gas buffer from the SDK. For example when computing relayer fee %, the SDK can use a buffer of 20 gwei on L1 for gas fee buffer (let's say current gas is 20-30 gwei). The relayer, on the other hand, might want to use 30 gwei instead when computing gas cost in profitability, because it wants to absolutely be sure it'd be profitable and not incur any loss due to gas spikes

> > Is there anything wrong with profiting a few cents on a small, cheap transaction? For instance, if the gas is $0.001 and the capital required is $0.50, a profit of $0.005 seems pretty reasonable, no?

yes but this thin margin means there's a decent chance of incurring a loss if gas turns against you.

> > My next followup is to update the SDK.

Makes sense. Ideally both SDK and relayer should be using the same formula/calculation but potentially with different gas buffer. I'll update this PR to use the SDK's code.

nicholaspai: > I think the bot should potentially have a different gas buffer from the SDK. For example when computing relayer fee %, the SDK can use a buffer of 20 gwei on L1 for gas fee buffer (let's say current gas is 20-30 gwei). The relayer, on the other hand, might want to use 30 gwei instead when computing gas cost in profitability, because it wants to absolutely be sure it'd be profitable and not incur any loss due to gas spikes

Couldn't the relayer instantiate the SDK with a different `gasBuffer` input in this case? I agree with your example .

src/utils/TransactionUtils.ts
nicholaspai: Also I just realized, why don't you use `utils/TransactionUtils/getGasPrice` here? I've never used provider.getGasPrice before so I'm not sure if it does what you think it does

src/clients/ProfitClient.ts
nicholaspai: Ah, i understand matt's confusion. Its that usually "gas used" in web3/ethers context refers the EVM gas assessed for a smart contract function. Also, `estimateGas` is a common function name that refers to the EVM gas cost of a function. Perhaps this would be reduce confusion if you renamed to `estimateGasCost` or `estimateTotalCost` wherever you call `estimateGas`. (I believe this is in the Relayer.ts file)

nicholaspai (commented):
Left some more comments

src/clients/ProfitClient.ts
mrice32: > > I think the bot should potentially have a different gas buffer from the SDK. For example when computing relayer fee %, the SDK can use a buffer of 20 gwei on L1 for gas fee buffer (let's say current gas is 20-30 gwei). The relayer, on the other hand, might want to use 30 gwei instead when computing gas cost in profitability, because it wants to absolutely be sure it'd be profitable and not incur any loss due to gas spikes
> 
> Couldn't the relayer instantiate the SDK with a different `gasBuffer` input in this case? I agree with your example .

@nicholaspai agree that this should be a parameterization of the SDK. Rather than it being a GWEI buffer, might be easier for us to make it percentage-based since gas prices work differently on different chains (for instance arbitrum and optimism have two types of gas), so a GWEI offset doesn't really work.

On your example @kevinuma, I think we'll ultimately want to make this configurable in both places as a GAS_MULTIPLIER, GAS_MARGIN, or GAS_DISCOUNT (same thing but inverted sign), so each caller can customize it. However, I think we probably want it to be the opposite of your example. In your example, my understanding is that the FE/API would _always_ be underpriced relative to the relayer, so we'd be blocking _most_ relays due to unprofitability.

I think we want the UI to be _more conservative_ than the relayer, because we want to avoid users who use our UI or API getting stuck (this is a bad product UX and creates support tickets). So I think, in general, we want GAS_MARGIN in the frontend/api to be greater than the default GAS_MARGIN in the relayer. Note: we can't control how 3rd parties set this, but we'd want them to use the default so they are willing to relay 99.9% of deposits sent through the FE/API.

Does that make sense?

mrice32: > > Is there anything wrong with profiting a few cents on a small, cheap transaction? For instance, if the gas is $0.001 and the capital required is $0.50, a profit of $0.005 seems pretty reasonable, no?
> 
> yes but this thin margin means there's a decent chance of incurring a loss if gas turns against you.

I'm splitting hairs here a bit, but just because the _absolute_ size and costs are low doesn't mean we have a higher chance of losing money I don't think. Likelihood of loss is dependent on gas price volatility. The _amount_ of money lost on a particular transaction is likely to be lower on a chain that consistently has lower gas prices. For instance, a bot that had no profitability module lost > $1k on a single transaction in Across v1 due to a mainnet gas spike. This size of loss is nearly impossible on a cheap chain, like polygon because gas prices just never get high enough.

Either way, I think we're aligned on how this system should work!

src/clients/ProfitClient.ts
mrice32: Yep, exactly. Sorry about that @kevinuma. You're completely right. Thanks for the clarification!

src/utils/TransactionUtils.ts
mrice32: It basically gives you the result of [eth_gasPrice](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gasprice) from the node. The rough algorithm used to compute this is [here](https://github.com/ethereum/pm/issues/328#issuecomment-853234014). I think this is pretty reasonable. I think the SDK does something similar for some chains.

The issue I see with this approach is that I'm not sure it will give reasonable results for all L2s. Optimism has a custom provider that has a special method to compute the estimated gas costs because optimism has _multiple types of gas_ (see the [SDK implementation](https://github.com/across-protocol/sdk-v2/blob/7905929cd6971623a1b51143d66bbe814e28bf94/src/relayFeeCalculator/chain-queries/optimism.ts#L39-L42)).

For the other chains, this approach should mostly work I think.

We mostly follow these rules in the SDK, [specific to each chain](https://github.com/across-protocol/sdk-v2/tree/master/src/relayFeeCalculator/chain-queries), but there are definitely areas to improve the estimation.

src/utils/TransactionUtils.ts
```diff
@@ -63,6 +63,23 @@ export async function willSucceed(
   }
 }
 
+export async function estimateGas(
+  transaction: AugmentedTransaction
+): Promise<{ transaction: AugmentedTransaction; succeed: boolean; reason: string; gasUsed: BigNumber }> {
+  try {
+    const args = transaction.value ? [...transaction.args, { value: transaction.value }] : transaction.args;
+    const [gasUsed, gasFee] = await Promise.all([
+      transaction.contract.estimateGas[transaction.method](...args),
```
mrice32, line 72: One thing we need to be careful about when estimating gas on this transaction is that it will revert if the estimation is sent from a wallet that either doesn't have enough of the currency or has not approved the contract to pull that token. So we need to make sure that this defaults to using a wallet that _will always_ work.

Is it possible to get here if the relayer doesn't have the balance to run this transaction?

Is the relayer address always used as the default `from` address in this simulated transaction?

src/utils/TransactionUtils.ts
```diff
@@ -63,6 +63,23 @@ export async function willSucceed(
   }
 }
 
+export async function estimateGas(
```
mrice32, line 66: nit: I think it might be clearer if we called this estimateGasCost, since estimateGas is the name of the method that is usually used to report the _amount of gas_ rather than the _amount of tokens_ used for gas.

mrice32 (commented):
This looks great! A few comments, but I think this mostly works as intended.

src/utils/TransactionUtils.ts
nicholaspai: Regardless of how `eth_gasPrice` works shouldn't we be consistent and use our own implementation of getGasPrice.ts?

src/utils/TransactionUtils.ts
nicholaspai: To be safe we can use the address hardcoded [here in the sdk-v2 repo](https://github.com/across-protocol/sdk-v2/blob/master/src/relayFeeCalculator/chain-queries/arbitrum.ts#L16)

nicholaspai:
Please add a PR description @kevinuma !

PR #213, opened 2022-07-18
improve(relayer): Track which deposits the relayer has tried to fill so it doesn't double fill (waste of gas)

kevinuma:
Currently when run in looping mode with POLLING_DELAY of 30s, the relayer bot sometimes double fill (second fill for the same deposit fails, which is a waste of gas). This happens very frequently for even smaller POLLING_DELAY such as 10s.

This solution keeps track of filled deposits in memory so that when run in looping mode, the relayer wouldn't try to do double fill when the first fill is still pending.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -96,12 +96,6 @@ export class Relayer {
       return;
     }
 
-    // Only track complete fills by the relayer.
-    // TODO: Revisit in the future when we implement partial fills.
-    if (toBN(deposit.amount).eq(fillAmount)) {
-      this.fullyFilledDeposits[fillKey] = true;
-    }
-
     try {
       // Fetch the repayment chain from the inventory client. Sanity check that it is one of the known chainIds.
       const repaymentChain = this.clients.inventoryClient.determineRefundChainId(deposit);
@@ -118,6 +112,9 @@ export class Relayer {
         message: "Relay instantly sent 🚀", // message sent to logger.
         mrkdwn: this.constructRelayFilledMrkdwn(deposit, repaymentChain, fillAmount), // message details mrkdwn
       });
+
+      // TODO: Revisit in the future when we implement partial fills.
+      this.fullyFilledDeposits[fillKey] = true;
 
       // Decrement tokens in token client used in the fill. This ensures that we dont try and fill more than we have.
       this.clients.tokenClient.decrementLocalBalance(deposit.destinationChainId, deposit.destinationToken, fillAmount);
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -3,12 +3,12 @@ import { createFormatFunction, etherscanLink, toBN } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
 
 import { Deposit } from "../interfaces";
-import { destinationChainId } from "@across-protocol/contracts-v2/dist/test/constants";
 
 const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
 
 export class Relayer {
-  private filledDepositsByChain: { [destinationChain: number]: { [depositId: number]: boolean } } = {};
+  // Track by originChainId since depositId is issued on the origin chain.
+  private filledDepositsByChain: { [originChainId: number]: { [depositId: number]: boolean } } = {};
 
   constructor(
     readonly relayerAddress: string,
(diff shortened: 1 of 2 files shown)


src/relayer/Relayer.ts
```diff
@@ -78,6 +81,18 @@ export class Relayer {
...
 
   fillRelay(deposit: Deposit, fillAmount: BigNumber) {
+    if (this.filledDepositsByChain[deposit.originChainId] === undefined) {
+      this.filledDepositsByChain[deposit.originChainId] = {};
+    }
+    // Skip deposits that this relayer has already filled to prevent double filling (which is a waste of gas as the
+    // second fill would fail).
+    // TODO: Handle the edge case scenario where the first fill failed due to transient errors and needs to be retried
+    if (this.filledDepositsByChain[deposit.originChainId][deposit.depositId]) {
+      this.logger.debug({ at: "Relayer", message: "Skipping deposits already fill by this same relayer", deposit });
```
nicholaspai, line 91:
```suggestion
      this.logger.debug({ at: "Relayer", message: "Skipping deposits already filled by this relayer", deposit });
```

src/relayer/Relayer.ts
```diff
@@ -78,6 +81,18 @@ export class Relayer {
...
+      return;
+    }
+    this.filledDepositsByChain[deposit.originChainId][deposit.depositId] = true;
```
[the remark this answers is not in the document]
kevinuma: Per Slack discussion, we'll only track complete fills by the same relayer and worry about partial fills in the future.

src/relayer/Relayer.ts
```diff
@@ -78,6 +82,26 @@ export class Relayer {
...
+        message: "Skipping deposits already filled by this relayer",
+        originChainId: deposit.originChainId,
+        depositId: deposit.depositId,
+      });
+      return;
+    }
+
+    // Only track complete fills by the relayer.
+    // TODO: Revisit in the future when we implement partial fills.
+    if (deposit.amount.eq(fillAmount)) {
```
nicholaspai, line 101: One frsutrating edge case is if the deposit was already partial filled. Perhaps we should pass in `totalFilledAmount` into this function, and just check that `deposit.amount.sub(totalFilledAmount).lte(fillAmount)`?

kevinuma: fillAmount is technically unfilled amount so it should always be exactly what's left after partial fills.

nicholaspai: ohhh gotcha. So this still will not correctly store the cases where there is an existing fill and the current relayer fills the remainder, right?

What if you passed in `unfiledAmount`, and then checked `deposit.amount.sub(unfilledAmount).eq(fillAmount)`?

nicholaspai: Ok fair, but I don't think you can just check `deposit.amount.eq(fillAmount)` because `deposit.amount` is the original deposit amount , so it won't be equal to `fillAmount` if there was a 1 wei fill. You know what I mean?

kevinuma: Yeah but otherwise it's pretty tricky since we'd need to handle partial fills completely. May be for now we keep simple so if the main relayer does a slow relay of 1 wei, the external relayers shouldn't touch those deposits? We can revisit this when we have proper support for partial fills?

nicholaspai: I agree that we should keep it simple but i'm still confused on this point:

> May be for now we keep simple so if the main relayer does a slow relay of 1 wei, the external relayers shouldn't touch those deposits?

If external relayers are running the same code as the "main" relayer in your construction, then the external relayer still can try to fully fill the relay since it hasn't seen it before, right?

kevinuma: the external relayers can disable slow relays with SEND_SLOW_RELAYS=false (default value). I don't think they would enable that anyway

src/relayer/Relayer.ts
```diff
@@ -78,6 +82,23 @@ export class Relayer {
...
+        at: "Relayer",
+        message: "Skipping deposits already filled by this relayer",
+        originChainId: deposit.originChainId,
+        depositId: deposit.depositId,
+      });
+      return;
+    }
+
+    // TODO: Revisit in the future when we implement partial fills.
+    this.fullyFilledDeposits[fillKey] = true;
```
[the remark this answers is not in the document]
kevinuma: This is a good point. Moved into the try block.

nicholaspai (approved):
LGTM!

PR #216, opened 2022-07-18
feat: New arg parsing and help/usage functions

pxrl:
This PR adds more consistent arg parsing and a help display:

- Use minimist for argument parsing. This allows more structured parsing, and supports some additional error checking.
- Implement usage() to document usage requirements.
- Implement support for -h|--help to give basic user help.
- Adjust call down to runX() functions to support future addition of top-level exception handling (address recursive calling issue).
- Per existing behaviour, the user is still permitted to provide (for example) --relayer and --dataworker. The command that applies is based on the ordering of an internal data structure.

For future
 - An update should modify usage() so that it can explain failures where a required argument was missing.
 - The --relayer(...) options should be converted to an operand, such that only one is provided (and we error out if more than one is present).


nicholaspai:
High level nit: use conventional commits when naming your PR's: https://www.conventionalcommits.org/en/v1.0.0/

index.ts
```diff
@@ -1,23 +1,100 @@
...
-export async function run(): Promise<void> {
+function usage(badInput: string = undefined) {
+  let usageStr = badInput ? `\nUnrecognized input: "${badInput}".\n\n` : "";
+  const userModes = "monitor|relayer";
+  const proModes = "dataworker|finalizer";
+  const walletOpts = "mnemonic|privateKey|gckms";
+
+  usageStr += `
+    Usage:
+    \tts-node ./index.ts --help
```
nicholaspai, line 18: what is the `tts-node` syntax?

index.ts
```diff
@@ -1,23 +1,100 @@
...
+    \tts-node ./index.ts [-h] <--monitor|--relayer>      --wallet <${walletOpts}>
+    \tts-node ./index.tx [-h] <--dataworker|--finalizer> --wallet <${walletOpts}>
+  `.slice(1);
```
nicholaspai, line 21: why are you slicing 1?

index.ts
```diff
@@ -1,23 +1,100 @@
...
+
+  console.log(usageStr);
+
+  // eslint-disable-next-line no-process-exit
+  process.exit(badInput === undefined ? 0 : -1);
```
nicholaspai, line 26: why -1? Isn't 1 the error code? https://nodejs.org/api/process.html#process_exit_codes

index.ts
```diff
@@ -1,23 +1,100 @@
...
+}
+
+function help() {
```
nicholaspai, line 29: instead of placing `help` and `usage` in `index.ts`, I would add a new file `help.ts` or something. Thoughts? Or we could add to `src/`?

index.ts
```diff
@@ -1,23 +1,100 @@
...
+    Across v2 Bot
+
+    Description:
+    \tThis application performs off-chain duties in support of the Across v2 bridge.
+
+    \tBasic sub-functions include:
+    \t  Monitor:    Monitor all relay and bundle events.
+    \t  Relayer:    Perform transaction relays based on eligible SpokePool deposits.
+
+    \tAdvanced sub-functions include:
```
nicholaspai, line 42: Should we notify user that these "advanced" functions will earn the users zero* incentives and cost gas, but will enable the user to secure the network and possibly earn dispute rewards (but these will be rare and are not much, like 0.4 ETH per dispute).

index.ts
```diff
@@ -1,23 +1,100 @@
...
+
+export async function run(args: { [k: string]: boolean | string }): Promise<void> {
   logger = Logger;
-  if (process.argv.includes("--relayer")) await runRelayer(Logger);
-  else if (process.argv.includes("--dataworker")) await runDataworker(Logger);
-  else if (process.argv.includes("--finalizer")) await runFinalizer(Logger);
-  else if (process.argv.includes("--monitor")) await runMonitor(Logger);
-  else console.log("Select either relayer, dataworker, or monitor to run!");
+
+  const cmds = {
```
nicholaspai, line 58: +1 for this structure, well organized

index.ts
```diff
@@ -1,23 +1,100 @@
...
+    help: help,
+    monitor: runMonitor,
+    relayer: runRelayer,
+  };
+
+  /* XXX Make the mode of operation an operand, rather than an option.
+   * i.e. ts-node ./index.ts [options] <relayer|...>
+   */
+  let cmd;
+  for (const cmdName of Object.keys(cmds)) {
```
nicholaspai, line 70: I agree with this change but this is a breaking change so we'll need to remember to update the bot configs in GCP once we merge this in

index.ts
```diff
@@ -1,23 +1,100 @@
...
+  await cmds[cmd](logger);
 }
 
 if (require.main === module) {
-  run()
+  const opts = {
+    boolean: ["dataworker", "finalizer", "help", "monitor", "relayer"],
+    string: ["wallet"],
+    alias: { h: "help" },
+    unknown: usage,
```
nicholaspai, line 93: why is `usage` listed here as an `opt`? I thought this dictionary would only contain potential command line keywords?

nicholaspai (commented):
This is really succinct code, I've left some preliminary questions and just some warnings on where we need to be careful!

index.ts
pxrl: Each sentence is prepended by a tab (\t). It produces the following output:

`Usage:
        ts-node ./index.ts --help
        ts-node ./index.ts [-h] <--monitor|--relayer>      --wallet <mnemonic|privateKey|gckms>
        ts-node ./index.tx [-h] <--dataworker|--finalizer> --wallet <mnemonic|privateKey|gckms>`

index.ts
pxrl: This is a bit of a bodge to allow usage() to be called directly from run(), or also via help(). The way the string is built, the first character is a newline. i.e.:

`let myStr = \`
  ^^ This string is built literally, so its char is a newline.
\``

In building the string I prioritised formatting it closely to how it will appear on screen. There's probably a better way to do that - happy to hear if you have any tips.

index.ts
pxrl: Ah - I wasn't aware that node specified such a detailed set of return codes - that's helpful!

If I understand that text correctly, it's describing the actual return behaviour of node itself - not necessarily the code that node is executing. Specifically for the return code of 1:

```
1 Uncaught Fatal Exception: There was an uncaught exception, and it was not handled by a domain or an ['uncaughtException'](https://nodejs.org/api/process.html#event-uncaughtexception) event handler.
```

I interpret this to mean that our code actually crashed hard and node had to take over and bail out on our behalf. In that case it's not our code that determines the error code, but node itself. 

In the scenario where bad input is provided by the user, we are actually catching that and handling it gracefully, so I don't think it's correct to return 1 for that. If we follow in the footsteps of node, the most appropriate would seem to be 9:

```
9 Invalid Argument: Either an unknown option was specified, or an option requiring a value was provided without a value.
```

This is actually how node behaves:

```
$ node --known-bad-argument
node: bad option: --known-bad-argument
$ echo $?
9
```

So...we could choose to "inherit" this behaviour, in which case we'd also return 9 here. It would however mean that the caller is not able to programmatically determine whether it was node, or the bot, that rejected the input arguments - this is not a problem for the nominal usage case, but might be annoying for any user running with a particular set of options to node. Alternatively, we could define our own set of return codes that fall somewhere in the range of (13, 127], since this seems to be unused by node.

Also worth mentioning, this is all predicated on the assumption that users of the bot actually need to differentiate between different failure causes. It's only actually useful if something is monitoring the return codes and logging or taking some other decision based on the result. We could implement some tests that call the bot with different sets of bad input in order to verify that it fails correctly, and that might help us to catch some future bugs before they are merged. For the average user, I think their main concern is "was the return code zero or non-zero?"

What do you think?

index.ts
pxrl: Yeah - agree on this. These two functions are disproportionately large in index.ts and they steal attention away from what is really happening.

Happy to take guidance on the best place to put it - are there norms/conventions for this?

index.ts
pxrl: Yep - will see if I can come up with some text for this.

index.ts
pxrl: Yeah - I don't have the overview of all the places this would need to be adjusted, so decided to mark it as a comment. Nice to hear that you think it's a reasonable suggestion!

index.ts
```diff
@@ -1,23 +1,100 @@
...
+
+  const cmds = {
+    dataworker: runDataworker,
+    finalizer: runFinalizer,
+    help: help,
+    monitor: runMonitor,
+    relayer: runRelayer,
+  };
+
+  /* XXX Make the mode of operation an operand, rather than an option.
```
mrice32, line 66: nit: what does XXX mean? Is that like a TODO?

index.ts
mrice32: I would put in `src/utils/help.ts`, maybe? Not strongly opinionated, though. I do agree it should go somewhere in src, but we don't have a strong convention for where to put something like this.

If we imagine there will be a lot of other CLI related files/utilities, we could create a src/cli folder, but that's probably overkill if this is the only one we foresee having.

index.ts
```diff
@@ -1,23 +1,100 @@
...
+   * i.e. ts-node ./index.ts [options] <relayer|...>
+   */
+  let cmd;
+  for (const cmdName of Object.keys(cmds)) {
+    if (args[cmdName]) {
+      cmd = cmdName;
+      break;
+    }
+  }
```
mrice32, line 75: nit: this is a more idiomatic way to write this code. See find docs [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).

```suggestion
  const cmd = Object.keys(cmds).find(cmdName => !!args[cmdName]);
```

index.ts
```diff
@@ -1,23 +1,100 @@
...
+
+  if (cmd === "help") {
+    cmds[cmd](); // no return
+  }
+
+  if (cmd === undefined || typeof args["wallet"] !== "string" || args["wallet"].length === 0) {
+    usage(""); // no return
```
mrice32, line 82: Do we have a way of telling the user what they did wrong?

index.ts
```diff
@@ -1,23 +1,100 @@
...
+
+  // eslint-disable-next-line no-process-exit
+  process.exit(badInput === undefined ? 0 : -1);
+}
+
+function help() {
+  const botRepoUrl = "https://github.com/across-protocol/relayer-v2";
+  const relayerDocsUrl = "https://docs.across.to/v2/developers/running-a-relayer";
+  const helpStr = `
+    Across v2 Bot
```
mrice32, line 33: This is very nicely formatted. Big +1 to using this type of string literal.

mrice32 (commented):
This is awesome! A few minor comments, but this is really cool!

mrice32 (commented):
We've gotten sort of bad about doing this well as an organization, but I would suggest a more detailed PR description that explains what the change does. This sort of thing helps the reviewer understand what was intended and the known shortfalls.

An _alternative_ to doing this is to make comments on your own code, giving the reviewer context for certain bits of code.

In this case, here's an example description that might help a reviewer:

```
This PR adds more consistent arg parsing and a help display.

Details:
- The argument parsing detects common issues, like a missing wallet or command argument (--dataworker).
- The new help output is triggered if the user provides --help instead of another command.
- This PR does _not_ prevent users from passing in multiple commands. In this case (as was true before), the first command argument will be used and the rest ignored.
```

index.ts
pxrl: It will explain if incorrect input is provided, but in the scenario where something is missing it just prints the generic usage without any specific feedback. Should probably improve on that.

index.ts
pxrl: XXX => todo so it fits in with the existing code.

index.ts
pxrl: Have made it return 9 for invalid runtime args in the latest change.

index.ts
pxrl: Have added to src/utils/usage.ts in the latest. Usage seemed right at the time...now it feels like help is a better name. Naming things is hard.

index.ts
pxrl: I've split this out in the latest change, so we check cmd and wallet separately. This paves the way for providing feedback that some input was missing. I'd prefer to have that as a follow-up improvement if it's OK.

index.ts
nicholaspai: Can you just add a comment explaining this im sure i won't be the only one wondering why you slice(1). Your explanation above makes sense!

index.ts
nicholaspai: agreed 9 seems fine!

index.ts
nicholaspai: yeah i could imagine cli becoming a larger directory

index.ts
nicholaspai: nothing else besides --network --wallet and --monitor|relayer|finalizer|dataworker are used rn

pxrl: Hm --network is new to me. I've searched for it and didn't see anything obvious. Are you able to provide a reference?

index.ts
pxrl: Fixed in the latest commit.

index.ts
nicholaspai: hmm you're right you don't need --network, my bad!

pxrl:
@nicholaspai Have permitted --keys as a runtime argument, so this should be OK with the gckms usage now. I believe this is ready to go.

nicholaspai (approved):
LGTM!

index.ts
mrice32: SGTM!

index.ts
```diff
@@ -1,23 +1,48 @@
...
+    dataworker: runDataworker,
+    finalizer: runFinalizer,
+    help: help,
+    monitor: runMonitor,
+    relayer: runRelayer,
+  };
+
+  /* todo Make the mode of operation an operand, rather than an option.
+   * i.e. ts-node ./index.ts [options] <relayer|...>
+   */
```
mrice32, line 23: nit: use `//` comment style in all cases except when adding specially-formatted documentation for functions, etc (https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/).

mrice32 (approved):
LGTM!

PR #217, opened 2022-07-19
improve: Set defaults for relayer deposit lookback

pxrl:
Enables the user to avoid manually setting MAX_RELAYER_DEPOSIT_LOOK_BACK
in their environment, thereby making the overall configuration simpler. The default
values match the configuration recommended in the Across GitBook.

Ref: ACX-73

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -3,12 +3,12 @@
 export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 
 export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
-    1: 11500,
-    10: 350000,
-    137: 70000,
-    288: 6000,
-    42161: 35000,
-}
+  1: 11500,
+  10: 350000,
+  137: 70000,
+  288: 6000,
+  42161: 35000,
+};
 
 // Optimism, ethereum can do infinity lookbacks. boba and Arbitrum limited to 100000 on infura.
 export const CHAIN_MAX_BLOCK_LOOKBACK = {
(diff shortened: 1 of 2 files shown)


src/common/Constants.ts
```diff
@@ -2,6 +2,14 @@
 // in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
 export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 
+export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
```
kevinuma, line 5: Can we add a comment to explain these numbers target ~2 days?

kevinuma (approved):

src/common/Constants.ts
nicholaspai: Yeah something like this:

```
// ~2 days on each chain: { 1: 15s/block, 10/42161: 0.5s/block, 137: 2.5s/block, 288: 30s/block }
```

src/common/Constants.ts
```diff
@@ -2,6 +2,14 @@
...
+  1: 11500,
+  10: 350000,
+  137: 70000,
+  288: 6000,
+  42161: 35000,
+};
+
```
nicholaspai, line 12: This is missing a zero right? Sanity check is that block lookbacks for Arbitrum and Optimism should roughly be the same since they both process ~2 blocks/second
```suggestion
  42161: 350000,
};

```

pxrl: Well spotted - thanks.

nicholaspai (approved):
LGTM!

PR #218, opened 2022-07-19
improve: Align .env.example with GitBook

pxrl:
Stretch goal...updating .env.example to make it easier for people to bootstrap a new config.


pxrl:
Hm..had expected it would realise the older commits already exist on master, but I guess that's not how it works...

pxrl:
This one got polluted with the previous PR. Will fix that and get it back up.

PR #219, opened 2022-07-19
improve: Align .env.example with Across relayer docs

pxrl:
I'd propose to relocate the configuration examples from the Across
relayer docs to this .env.sample file, and simply reference it from the
docs. This will hopefully make it easier for users to get started, and
will make it easier for us to keep the sample configuration current.


.env.example
```diff
@@ -1,3 +1,161 @@
...
+#        "targetPct": 0.2, "thresholdPct": 0.1,
+#        "unwrapWethThreshold": 0.65, "unwrapWethTarget": 1,
+#      },
+#      "42161": {
+#        "targetPct": 8, "thresholdPct": 5,
+#        "unwrapWethThreshold": 0.75, "unwrapWethTarget": 1.5,
+#      },
+#    },
+#  },
+#}'
```
pxrl, line 145: The dotenv package claims it's capable of handling multi-line variables formatted this way. I haven't actually verified it yet.

.env.example
```diff
@@ -1,3 +1,161 @@
...
+#NODE_URLS_137=[<Polygon-Node-URL>,<Polygon-Node-URL-2>]
+#NODE_URLS_288=[<Boba-Node-URL>,<Boba-Node-URL-2>]
+#NODE_URLS_42161=[<Arbitrum-Node-URL>,<Arbitrum-Node-URL-2>]
+
+# Defaults to 1, must be set >= 1. If set > 1, then for any RPC request, this
+# requires that at least this count of RPCs specified in the NODE_URLS_[ID] list
+# list return the exact same response. We recommend setting this to 2 so that
+# the relayer does not accidentally relay a deposit whose properties two RPCs
+# disagree about. This edge case is rare but could lead to loss of funds.
+#NODE_QUORUM_THRESHOLD=2
```
pxrl, line 161: It strikes me that we recommend people run with at least 2 RPCs per chain, and then bury it at the bottom of the "advanced" section of the config.

.env.example
```diff
@@ -1,3 +1,161 @@
...
+
+# Configure whether the bot will actively try to send relays or not. Note that
+# the bot will currently still attempt token approvals if they are missing -
+# even when relaying is disabled - but will not attempt to perform any relays
+# thereafter.
+#
+# It's strongly recommended to initially test this bot with SEND_RELAYS=false
+# in order to validate the configuration and behaviour of the bot. Set
+# SEND_RELAYS=true later when comfortable that the behaviour as intended.
+#SEND_RELAYS=false
```
kevinuma, line 36: The problem with this though is currently even if you set this to false, the relayer would still attempt to send token approvals. Not sure if we need to document that. We have a task outstanding to skip approvals when in simulation.

.env.example
```diff
@@ -1,3 +1,161 @@
...
+# A Redis in-memory DB can drastically speed up the performance of the bot.
+# This is technically not required, but can reduce the instance of repeated
+# RPC queries and therefore reduces the time and network bandwidth required
+# for successful relay bot operation.
+#
+# Install redis and then ensure that redis-server is started:
+#     https://redis.io/docs/getting-started/
+# Under the hood, the relayer will cache JSON-rpc request data from requests
+# like `eth_getBlock` in the Redis DB.
+REDIS_URL="redis://127.0.0.1:6379"
```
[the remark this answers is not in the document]
kevinuma: If people start redis as-is without overriding the port, it's always 6379 by default. So I think we can leave this not commented.

.env.example
```diff
@@ -1,3 +1,161 @@
...
+SEND_SLOW_RELAYS=false
+
+
+# How many blocks to look back per request for a chain. Relayer code will bundle
+# requests, so a lower number implies that there will be more requests to the
+# RPC endpoint. This implies more request overhead and therefore impacts the
+# speed at which the bot can operate. Some RPCs, like Infura, limit lookback to
+# fixed number of blocks. The example configuration below is also the default
+# behaviour of the bot and is known to work with Infura.
+#MAX_BLOCK_LOOK_BACK='{
```
kevinuma, line 99: I think after we have a default here we can remove this config?

.env.example
```diff
@@ -1,3 +1,161 @@
...
+#}'
+
+
+# How many blocks to lookback for Deposit events to be filled. The below ranges
+# The below lookback ranges contain approximately 2 days of blocks for each
+# network. This can be shortened to send fewer requests to the RPC endpoint and
+# generally speed up the bot, at the risk of overlooking events that fall
+# outside of the specified window. The example configuration below is also the
+# default behaviour of the bot.
+#MAX_RELAYER_DEPOSIT_LOOK_BACK='{
```
kevinuma, line 110: ditto earlier comment

.env.example
kevinuma: I think may be we should recommend just using 1?

.env.example
pxrl: Ah - I actually created an issue on this as well: https://linear.app/uma/issue/ACX-77/consider-skipping-token-approvals-when-relaying-is-disabled-via-send

Do you have a reference for the other one? I had a quick look but couldn't see it.

I agree it's counter-intuitive behaviour for approval txns to be attempted in this case. Whether or not we add it to .env.example doesn't really change that though, I guess.

kevinuma: It's here https://linear.app/uma/issue/ACX-64/add-simulation-mode-for-token-allowances

.env.example
pxrl, line 99: Note to self - this doesn't work. It seems the only way to get a multi-line variable is to use a double quote. Need to investigate further...but might also be redundant based on Kevin's comment above.

.env.example
```diff
@@ -1,3 +1,161 @@
...
+
+# By default set to false. If false, the relayer will ignore any deposits that
+# it cannot fully fill. A feature for partial fills is not yet implemented but
+# it's possible to do so at the contract level. If this is set to true, then the
+# relayer will fill the minimum possible amount of the deposited amount (i.e. 1
+# wei for ETH), which will trigger a slow fill payment to be sent from the mainnet
+# HubPool to the destination SpokePool. There is no benefit for a user to do this,
+# and the system will function correctly as long as at least one relayer is
+# handling slow relays. We recommend that you leave this value set to "false".
+SEND_SLOW_RELAYS=false
```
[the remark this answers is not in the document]
pxrl: Confirmed that this is the default configuration.

james-a-morris:
@pxrl  cc @nicholaspai Is this still a valid PR or should we move to close?

pxrl:
> @pxrl cc @nicholaspai Is this still a valid PR or should we move to close?

The PR needs an overhaul but I'll make sure to update it early next week and re-propose. We have made a few improvements to the relayer configuration over the past 6 months that should make it easier to fix now.

james-a-morris (approved):
@pxrl this is amazing

PR #224, opened 2022-07-20
fix: Dataworker should filter out all deposits younger than bundle end block

nicholaspai:
Same motivation and logic as https://github.com/across-protocol/relayer-v2/pull/221 but this filter on fillWithBlock is set higher in the function.

I don't know if this fixes the underlying bug, but the current situation is that sometimes proposer and disputer see the exact same deposits and fills in the range but construct different `netSendAmounts/runningBalances`, with the disputer consistently seeing lower values.

My guess is that there is some bug in `dataworker/PoolRebanceUtils.ts/subtractExcessFromPreviousSlowFillsFromRunningBalances()` which could only lower the net send amounts.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -222,51 +222,55 @@ export class BundleDataClient {
         );
 
         // Find all valid fills matching a deposit on the origin chain and sent on the destination chain.
-        destinationClient.getFillsWithBlockForOriginChain(Number(originChainId)).forEach((fillWithBlock) => {
-          // If fill matches with a deposit, then its a valid fill.
-          const matchedDeposit: Deposit = originClient.getDepositForFill(fillWithBlock);
-          if (matchedDeposit) {
-            // Fill was validated. Save it under all validated fills list with the block number so we can sort it by
-            // time. Note that its important we don't skip fills earlier than the block range at this step because
-            // we use allValidFills to find the first fill in the entire history associated with a fill in the block
-            // range, in order to determine if we already sent a slow fill for it.
-            if (fillWithBlock.blockNumber <= blockRangeForChain[1]) allValidFills.push(fillWithBlock);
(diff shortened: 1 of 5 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -222,51 +222,55 @@ export class BundleDataClient {
...
         // Find all valid fills matching a deposit on the origin chain and sent on the destination chain.
-        destinationClient.getFillsWithBlockForOriginChain(Number(originChainId)).forEach((fillWithBlock) => {
-          // If fill matches with a deposit, then its a valid fill.
-          const matchedDeposit: Deposit = originClient.getDepositForFill(fillWithBlock);
-          if (matchedDeposit) {
-            // Fill was validated. Save it under all validated fills list with the block number so we can sort it by
-            // time. Note that its important we don't skip fills earlier than the block range at this step because
-            // we use allValidFills to find the first fill in the entire history associated with a fill in the block
-            // range, in order to determine if we already sent a slow fill for it.
-            if (fillWithBlock.blockNumber <= blockRangeForChain[1]) allValidFills.push(fillWithBlock);
```
nicholaspai, line 233: The change here is that this `(fillWithBlock.blockNumber <= blockRangeForChain[1])` check was moved to a `.filter()` on what was line 225

src/scripts/validateRootBundle.ts
```diff
@@ -72,7 +79,8 @@ export async function validate(_logger: winston.Logger) {
       "EnabledDepositRoute",
       "RelayedRootBundle",
       "ExecutedRelayerRefundRoot",
-    ]
+    ],
+    bundleEndBlockMapping
```
mrice32, line 83: What does this change?

mrice32 (approved):
Looks great! Great idea to log!

src/scripts/validateRootBundle.ts
nicholaspai: Ah this allows the script to use the cache

PR #225, opened 2022-07-20
improve: More detailed logging for subtracting excesses from slow fill

nicholaspai:


src/dataworker/Dataworker.ts
```diff
@@ -1213,7 +1217,8 @@ export class Dataworker {
     deposits: DepositWithBlock[],
     allValidFills: FillWithBlock[],
     allValidFillsInRange: FillWithBlock[],
-    unfilledDeposits: UnfilledDeposit[]
+    unfilledDeposits: UnfilledDeposit[],
+    logSlowFillExcessData = false
```
nicholaspai, line 1221: I just don't want to log when executing historical relayer refund leaves or slow relay leaves

mrice32 (approved):
LGTM!

PR #227, opened 2022-07-21
fix: Temporary stopgap to delay proposing bundles until 20 mins after latest Ethereum relayer refund leaf execution

nicholaspai:
https://linear.app/uma/issue/ACX-96/fix-bug-related-to-proposing-bundles-right-after-poolrebalanceleaves

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ coverage
 coverage.json
 typechain
 .DS_Store
-dump.rdb
+dump.rdb*
 .idea
 
 # Hardhat files
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -61,26 +61,6 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
       );
 
       if (spokePoolClients === undefined) {
-        // This is a temporary fix: update spoke clients to repopulate cache twice. This guarantees that the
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@ coverage
 coverage.json
 typechain
 .DS_Store
-dump.rdb
+dump.rdb*
```
nicholaspai, line 7: I keep dump.rdb backups to test my redisdb state at points in time

src/dataworker/index.ts
```diff
@@ -61,26 +61,6 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
       );
 
       if (spokePoolClients === undefined) {
-        // This is a temporary fix: update spoke clients to repopulate cache twice. This guarantees that the
-        // RedisDB cache always contains all events from block 0 to the latest bundle end block.
-        // Without this second update, the spoke client could have populated data from an "incomplete" cache that
-        // only contains events from block 0 to the N-1 bundle end block. The client should then fetch events
-        // from N-1 bundle end block until latest bundle end block, but this doesn't seem to be working right now.
-        await constructSpokePoolClientsForBlockAndUpdate(
```
nicholaspai, line 69: This turned out not to be a fix and just costs extra time+networking

PR #228, opened 2022-07-21
fix: shouldWaitToPropose should use most recently validated proposal

nicholaspai:
Bug: uses most recent proposal instead of validated proposal.

This would cause problems if a proposal were to be disputed, then the next proposer would never propose again since there would never be a relayer refund root relayed to EthereumSpokePool with the matching root

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -113,7 +113,7 @@ export class MultiCallerClient {
       }
 
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
-        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 75)])
+        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 100)])
       );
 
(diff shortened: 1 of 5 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -113,7 +113,7 @@ export class MultiCallerClient {
       }
 
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
-        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 75)])
+        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 100)])
```
mrice32, line 116: Was this intentional? I thought we set this to 50?

mrice32 (approved):
LGTM! Nice fix

src/clients/MultiCallerClient.ts
nicholaspai: I actually bumped this back up, see this issue: https://linear.app/uma/issue/ACX-98/relayer-occassionally-runs-into-nonce-has-already-been-used-errors

I think. me lowering it last time was a false bug fix

PR #229, opened 2022-07-23
improve: execute leaves from oldest to newest

nicholaspai:
This mitigates any timing issues with stuck leaves, oldest ones should be unstuck first


mrice32 (approved):
LGTM good idea!

PR #230, opened 2022-07-25
fix: return unfilled amount = 0 if all fills found for a deposit are invalid

kevinuma:
Currently if the relayer finds only invalid fills for a deposit, the code throws an error as it expects at least one of the fills to be valid. This PR adds a check to return unfilledAmount = 0 if all fills are invalid.

error: TypeError: Cannot read property 'amount' of undefined
   at SpokePoolClient.getValidUnfilledAmountForDeposit (/across-relayer/dist/src/clients/SpokePoolClient.js:147:61)
   at /across-relayer/dist/src/utils/FillUtils.js:115:47
   at Array.map (<anonymous>)
   at getUnfilledDeposits (/across-relayer/dist/src/utils/FillUtils.js:112:18)
   at Relayer.checkForUnfilledDepositsAndFill (/across-relayer/dist/src/relayer/Relayer.js:29:66)
   at runRelayer (/across-relayer/dist/src/relayer/index.js:21:27)
   at runMicrotasks (<anonymous>)
   at processTicksAndRejections (internal/process/task_queues.js:95:5)
   at async run (/across-relayer/dist/index.js:13:9)

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -173,9 +173,15 @@ export class SpokePoolClient {
 
   getValidUnfilledAmountForDeposit(deposit: Deposit): { unfilledAmount: BigNumber; fillCount: number } {
     const fillsForDeposit = this.depositHashesToFills[this.getDepositHash(deposit)];
-    if (fillsForDeposit === undefined || fillsForDeposit === [])
-      return { unfilledAmount: toBN(deposit.amount), fillCount: 0 }; // If no fills then the full amount is remaining.
+    // If no fills then the full amount is remaining.
+    if (fillsForDeposit === undefined || fillsForDeposit.length === 0) {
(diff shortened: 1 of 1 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -185,6 +187,11 @@ export class SpokePoolClient {
         ? -1
         : 0
     );
+    // If all fills are invalid we can consider this unfilled.
+    if (fillsOrderedByTotalFilledAmount.length === 0) {
```
nicholaspai, line 191: nit: do you think this check should be for `fills.length` and could it be made higher on line 181?

nicholaspai (approved):
One nit!

src/clients/SpokePoolClient.ts
kevinuma: Good call 👍 Updated.

PR #231, opened 2022-07-27
improve(finalizer): Don't filter out older token bridge events

nicholaspai:
L2 block times are hard to predict so we shouldn't filter out events that are "too old", but we can continue to filter out TokensBridged events that are "too young" because eventually they'll be old enough


mrice32 (approved):
LGTM!

kevinuma (approved):

PR #232, opened 2022-07-27
improve: bot exception handling

pxrl:
This PR addresses two main observations:

1. The existing exception handling recursively calls into the same function that experienced the exception. With a non-zero POLLING_DELAY, this increase stack usage and will leak memory.

2. For some unknown reason, Redis client instances don't appear to be GC'ed after they have gone out of scope. This means that Redis connections are maintained indefinitely - even though they are not used. When an exception is caught by the top-level handler, this results in a new connection being established. At some point, the server will stop accepting new connections. I have seen the bot hang in this scenario.

The fix for (1) is easy enough - exceptions that are otherwise unhandled will now ripple all the way back to the top-level index.ts, which avoids recursive re-entry.

A workaround/bodge is proposed to address (2). There's some underlying issue here that needs deeper investigation; the proposed workaround simply manually shuts down the Redis connection before re-throwing whatever underlying exception occurred.

As an observation from this work, there is some "common" functionality that is repeated across each supported sub-function (dataworker, finalizer, monitor, relayer). This can be seen in that the redis teardown had to be inserted in 4 separate places, all of which were very similar (but not identical). A future task could be to factor some of this common functionality out into the top-level index.ts.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,6 @@
 import minimist from "minimist";
-import { delay, help, Logger, usage, winston } from "./src/utils";
+import { CommonConfig } from "./src/common";
+import { config, delay, help, Logger, processCrash, usage, winston } from "./src/utils";
 import { runRelayer } from "./src/relayer";
 import { runDataworker } from "./src/dataworker";
 import { runMonitor } from "./src/monitor";
@@ -9,6 +10,8 @@ let logger: winston.Logger;
 
 export async function run(args: { [k: string]: boolean | string }): Promise<void> {
   logger = Logger;
+
+  const config = new CommonConfig(process.env);
 
   const cmds = {
     dataworker: runDataworker,
@@ -29,11 +32,20 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
     // todo: Update usage() to provide a hint that wallet is missing/malformed.
     usage(""); // no return
   } else {
-    await cmds[cmd](logger);
+    do {
+      try {
+        await cmds[cmd](logger);
+      } catch (error) {
+        // eslint-disable-next-line no-process-exit
+        if (await processCrash(logger, cmd, config.pollingDelay, error)) process.exit(1);
+      }
+    } while (config.pollingDelay !== 0);
   }
 }
 
 if (require.main === module) {
+  config();
+
   const opts = {
     boolean: ["dataworker", "finalizer", "help", "monitor", "relayer"],
     string: ["wallet", "keys"],
--- a/index.ts
+++ b/index.ts
@@ -21,9 +21,8 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
     relayer: runRelayer,
   };
 
-  /* todo Make the mode of operation an operand, rather than an option.
-   * i.e. ts-node ./index.ts [options] <relayer|...>
-   */
+  // todo Make the mode of operation an operand, rather than an option.
+  // i.e. ts-node ./index.ts [options] <relayer|...>
   const cmd = Object.keys(cmds).find((_cmd) => !!args[_cmd]);
 
   if (cmd === "help") cmds[cmd](); // no return
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -134,6 +134,6 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
     logger.debug("Disconnecting from redis server.");
     clients.configStoreClient.redisClient.disconnect();
 
-    throw error
+    throw error;
   }
 }
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -127,9 +127,6 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
       if (await processEndPollingLoop(logger, "Dataworker", config.pollingDelay)) break;
     }
   } catch (error) {
-    // eslint-disable-next-line no-process-exit
-    if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
-
     if (clients.configStoreClient.redisClient !== undefined) {
       // todo understand why redisClient isn't GCed automagically.
       logger.debug("Disconnecting from redis server.");
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -130,9 +130,11 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
 
-    // todo understand why redisClient isn't GCed automagically.
-    logger.debug("Disconnecting from redis server.");
-    clients.configStoreClient.redisClient.disconnect();
+    if (clients.configStoreClient.redisClient !== undefined) {
+      // todo understand why redisClient isn't GCed automagically.
+      logger.debug("Disconnecting from redis server.");
+      clients.configStoreClient.redisClient.disconnect();
+    }
 
     throw error;
   }
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -134,6 +134,6 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
     logger.debug("Disconnecting from redis server.");
     clients.configStoreClient.redisClient.disconnect();
 
-    await runDataworker(logger);
+    throw error
   }
 }
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -129,6 +129,11 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
   } catch (error) {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
+
+    // todo understand why redisClient isn't GCed automagically.
+    logger.debug("Disconnecting from redis server.");
+    clients.configStoreClient.redisClient.disconnect();
+
     await runDataworker(logger);
   }
 }
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -157,9 +157,6 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
       if (await processEndPollingLoop(logger, "Dataworker", config.pollingDelay)) break;
     }
   } catch (error) {
-    // eslint-disable-next-line no-process-exit
-    if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
-
     if (commonClients.configStoreClient.redisClient !== undefined) {
       // If this throws an exception, it will mask the underlying error.
       logger.debug("Disconnecting from redis server.");
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -159,6 +159,10 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
   } catch (error) {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
+
+    // If this throws an exception, it will mask the underlying error.
+    commonClients.configStoreClient.redisClient.disconnect();
+
     await runFinalizer(logger);
   }
 }
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -162,7 +162,6 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
 
     // If this throws an exception, it will mask the underlying error.
     commonClients.configStoreClient.redisClient.disconnect();
-
-    await runFinalizer(logger);
+    throw error
   }
 }
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -160,8 +160,11 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
 
-    // If this throws an exception, it will mask the underlying error.
-    commonClients.configStoreClient.redisClient.disconnect();
+    if (commonClients.configStoreClient.redisClient !== undefined) {
+      // If this throws an exception, it will mask the underlying error.
+      logger.debug("Disconnecting from redis server.");
+      commonClients.configStoreClient.redisClient.disconnect();
+    }
     throw error;
   }
 }
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -162,6 +162,6 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
 
     // If this throws an exception, it will mask the underlying error.
     commonClients.configStoreClient.redisClient.disconnect();
-    throw error
+    throw error;
   }
 }
--- a/src/monitor/index.ts
+++ b/src/monitor/index.ts
@@ -49,7 +49,6 @@ export async function runMonitor(_logger: winston.Logger) {
       logger.debug("Disconnecting from redis server.");
       clients.configStoreClient.redisClient.disconnect();
     }
-
-    await runMonitor(logger);
+    throw error
   }
 }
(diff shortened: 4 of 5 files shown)


src/dataworker/index.ts
```diff
@@ -129,6 +129,11 @@ export async function runDataworker(_logger: winston.Logger): Promise<void> {
   } catch (error) {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
-    await runDataworker(logger);
+
+    // todo understand why redisClient isn't GCed automagically.
+    logger.debug("Disconnecting from redis server.");
```
nicholaspai, line 134: How does this ever get here, doesn't `await processCrash(logger, "Dataworker", config.pollingDelay, error)` return true?

src/monitor/index.ts
```diff
@@ -41,6 +43,12 @@ export async function runMonitor(_logger: winston.Logger) {
   } catch (error) {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Monitor", config.pollingDelay, error)) process.exit(1);
-    await runMonitor(logger);
+
+    if (clients !== undefined) {
```
nicholaspai, line 47: Is it possible that `redisClient` is undefined if REDIS_URL isn't set?

index.ts
```diff
@@ -29,11 +31,20 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
     // todo: Update usage() to provide a hint that wallet is missing/malformed.
     usage(""); // no return
   } else {
-    await cmds[cmd](logger);
+    do {
+      try {
```
nicholaspai, line 35: I'm not sure I understand what you're doing here by adding the do-while loop. Can you explain?

nicholaspai (commented):
This redis-client `disconnect()` workaround makes sense for now, and I understand the diagnosis of the recursive issue.

nicholaspai:
requesting @daywiss as an optional reviewer because he's got a knack for designing good error processing architecture

src/monitor/index.ts
pxrl: Yeah, redisClient is undefined if the REDIS_URL env var isn't set.

This is an arguably not-so-nice edge case here, since trying to disconnect() on an undefined redisClient throws another exception. This second exception is immediately escalated to the upper level, and it could give the incorrect impression that the underlying cause was something to do with the redisClient instance, rather than whatever landed us in an exception handler in the first place.

I can't really see any side-effects from this, but it's a bit of a red herring and adds unnecessary noise to the logs - i.e. it might mislead someone who is trying to troubleshoot. In practice I still think that's unlikely, since we anticipate that every user will run with Redis anyway. However, a compromise might be to make the check a bit more robust:

`
if (clients !== undefined && clients.configStoreClient.redisClient !== undefined)  {
  // do teardown here
}
`

Any thoughts on this approach?

src/dataworker/index.ts
pxrl: processCrash only returns true for pollingDelay === 0 (i.e. serverless). In this case there's not much point in tearing connections down gracefully because the whole process will terminate anyway.

With a non-zero POLLING_DELAY configuration, processCrash will return false and allows us to proceed with a graceful teardown because we expect to set everything up again on the next loop.

index.ts
pxrl: The do..while loop ensures at least one run through of the requested command. In the case where we have a zero POLLING_DELAY, we will never repeat the loop afterwards - no matter the result. With a non-zero POLLING_DELAY, we will unconditionally loop.

The intention is just to match the existing behaviour as is implemented today - there shouldn't be any new behaviour here.

pxrl:
Also FWIW, I think there are more opportunities for simplifying the error-paths that can be traversed here. For instance, I think it would be an improvement if we could avoid calling processCrash() in multiple different places. Additionally, if the common clients setup was handled at the top level index.ts (i.e. in the fun() function) then we could also handle the redisClient disconnect()ion in one place, instead of 4.

The reason I have proposed the current change is that is addresses two issues without requiring a lot of refactoring, and it otherwise intends to maintain the existing behaviour around POLLING_DELAY. So I see it as an inexpensive (in terms of code and time) change to make to remove some sharp edges - not as a complete solution.

index.ts
nicholaspai: I didn't realize that `cmds` was a looping function, how does `cmds` work?

src/dataworker/index.ts
nicholaspai: Gotcha. So I see this `logger.debug(...); clients.redisClient.disconnect(); throw error` pattern being repeated multiple times in this PR. Do you think it would make sense to turn it into a function that can be reused?

src/monitor/index.ts
nicholaspai: > In practice I still think that's unlikely, since we anticipate that every user will run with Redis anyway. 
This is true but even I forget to start my `redis-server` before I run the bot sometimes so this could happen more than we expect

nicholaspai: I would go with making the check more robust in exchange for extra verbosity

index.ts
```diff
@@ -29,11 +31,20 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
...
+        await cmds[cmd](logger);
+      } catch (error) {
+        // eslint-disable-next-line no-process-exit
+        if (await processCrash(logger, cmd, config.pollingDelay, error)) process.exit(1);
```
daywiss, line 39: you can just re-throw the error rather than forcibly exiting the process
EDIT: i see now that this pattern was already set by previous code

index.ts
daywiss: i think this is like a UX question. if you are specifying by the command line a particular bot to run, do you want it to respect the `pollingDelay` config, or just run a single time ( which was the previous implementation)

daywiss (commented):
very subtle changes, its hard to know as the reviewer if things are working as intended. maybe you could provide how you verified the changes work? otherwise the code seems fine, though my gut says the existing architecture looks fragile.

daywiss:
> requesting @daywiss as an optional reviewer because he's got a knack for designing good error processing architecture

Thanks @nicholaspai, im just not sure how much refactoring you have an appetite for. Paul is basically doing some isolated improvements but i think changing the philosophy of the logging/error handling/ resource management for the better would require a much bigger refactor and may not be worth the trouble.

nicholaspai:
> For instance, I think it would be an improvement if we could avoid calling processCrash() in multiple different places. 

This is a great idea.

> Additionally, if the common clients setup was handled at the top level index.ts (i.e. in the run() function) then we could also handle the redisClient disconnect()ion in one place, instead of 4.

This is also a great idea @pxrl . Can you work on these in a follow up larger refactor?

@daywiss I do think we should do a larger refactor, do you have any other suggestions in addition to what Paul recommended above?

index.ts
nicholaspai: I would have thought that POLLING_DELAY is always respected

index.ts
nicholaspai: Do you think we should just throw the error instead of process.exit-ing? Seems like we should avoid the low level process.X calls if possible

nicholaspai (approved):
LGTM after resolving all open threads. As a follow up, @pxrl and I should create issues on Linear for refactoring the whole exception handling architecture, as suggested by these threads:
- https://github.com/across-protocol/relayer-v2/pull/232#issuecomment-1201262398
- https://github.com/across-protocol/relayer-v2/pull/232#issuecomment-1198250400

index.ts
mrice32: I think the process.exit call is important at the top level to avoid "unhandled promise rejection" printouts, but I think the calling function (see line 56) does that for us. I would think rethrowing in this case would be sufficient here, right?

src/finalizer/index.ts
```diff
@@ -159,6 +159,9 @@ export async function runFinalizer(_logger: winston.Logger): Promise<void> {
   } catch (error) {
     // eslint-disable-next-line no-process-exit
     if (await processCrash(logger, "Dataworker", config.pollingDelay, error)) process.exit(1);
-    await runFinalizer(logger);
+
+    // If this throws an exception, it will mask the underlying error.
+    commonClients.configStoreClient.redisClient.disconnect();
```
mrice32, line 164: My guess is that the redis client isn't disconnected and garbage collected because it or its connection is referenced by a global variable somewhere (either in our code or in the redis code), which would prevent js from garbage collecting it.

Your solution does tear down the connection, but it's unclear if it allows the entire redis client to be GC'd. However, doing this disconnect in so many places seems error-prone since we'd need to keep doing this in all new teardown code.

Have you considered embracing the long-lived redis client and keeping a global reference to the redis client? For instance, you could do this in a global/static class object that maps from url to redis client. That way, whenever starting up, the code could check if the redis client has already been created/initialized, and if so, it will just be reused?

I only suggest this because that change might be a bit more maintainable long term.

Thoughts?

mrice32 (commented):
Looks good! A few minor comments

index.ts
pxrl: Per this PR, POLLING_DELAY should continue to be used in the same was as before - i.e. it wasn't intended to introduce any change in behaviour, and I'm not aware of one. Happy to be corrected on that with an example, though :)

src/monitor/index.ts
pxrl: > > In practice I still think that's unlikely, since we anticipate that every user will run with Redis anyway.
> 
> This is true but even I forget to start my `redis-server` before I run the bot sometimes so this could happen more than we expect

Have just verified the behaviour when the redis server isn't running. If the REDIS_URL env var is set but the server isn't contactable, then the bot actually fails during initialisation and we never even get the clients variable, so the change as it was proposed does handle that OK, and the user sees the following error:

```
2022-08-02 10:29:20 [error]: {
  "at": "Relayer#index",
  "message": "There was an execution error! Re-running loop",
  "notificationPath": "across-error",
  "bot-identifier": "NO_BOT_ID"
}
```

As far as I can see, checking for the existence of the redisClient variable is only needed in the case where the user has not defined REDIS_URL in their environment, and an exception occurs. In that case it would a red-herring exception to be thrown.

It's definitely more correct to check for the existence of the redisClient variable before attempting to disconnect(), and I have added that in the latest update. It's just something that should be exceedingly rare in practice, since we don't anticipate a user running without Redis.

src/dataworker/index.ts
pxrl: > Gotcha. So I see this `logger.debug(...); clients.redisClient.disconnect(); throw error` pattern being repeated multiple times in this PR. Do you think it would make sense to turn it into a function that can be reused?

Generally yes, but I think we can alternatively move the "common clients" setup to the top-level index.ts and just handle it there. Then we don't need 4 x disconnect() calls, since one will suffice.

That change requires a few more modifications, so I'd probably handle it separately so it's easy for us to reason that the behaviour of the bot isn't meaningfully modified.

index.ts
pxrl: > very subtle changes, its hard to know as the reviewer if things are working as intended. maybe you could provide how you verified the changes work? otherwise the code seems fine, though my gut says the existing architecture looks fragile.

Just to give feedback on the question about how it's been tested - I've forced exceptions by inserting a 'throw "XXX";' statement into the for(;;) loops of each "command" (i.e. runRelayer()).

pxrl: > Do you think we should just throw the error instead of process.exit-ing? Seems like we should avoid the low level process.X calls if possible

> I think the process.exit call is important at the top level to avoid "unhandled promise rejection" printouts, but I think the calling function (see line 56) does that for us. I would think rethrowing in this case would be sufficient here, right?

Per the latest change, I've removed the 2nd-tier "processCrash()" calls. So it's now only handled in the top-level index.ts. I think this is cleaner and makes the behaviour more obvious.

src/finalizer/index.ts
pxrl: > My guess is that the redis client isn't disconnected and garbage collected because it or its connection is referenced by a global variable somewhere (either in our code or in the redis code), which would prevent js from garbage collecting it.
> 
> Your solution does tear down the connection, but it's unclear if it allows the entire redis client to be GC'd. However, doing this disconnect in so many places seems error-prone since we'd need to keep doing this in all new teardown code.
> 
> Have you considered embracing the long-lived redis client and keeping a global reference to the redis client? For instance, you could do this in a global/static class object that maps from url to redis client. That way, whenever starting up, the code could check if the redis client has already been created/initialized, and if so, it will just be reused?
> 
> I only suggest this because that change might be a bit more maintainable long term.
> 
> Thoughts?

I agree that it's not nice to have _n_ x disconnect() calls. My hope would be we can cut that down to 1 by migrating the "common" client init/setup calls to the top level, so we're not repeating "common" stuff at the 2nd tier.

It does however leave the potential issue of shared global state not being cleaned up properly, and I have no evidence that I am handling that correctly...so wrapping it more deliberately and introducing some basic state management would make sense. I think it's a good suggestion!

daywiss:
> @daywiss I do think we should do a larger refactor, do you have any other suggestions in addition to what Paul recommended above?

@nicholaspai Incremental improvements are fine, imo but if you want to refactor some high level things id suggest:

- move towards dependency injection, for example instantiating redis and other tools at the top level and passing them into code that needs it.  This makes it easy to see all the configuration needed for the app at an obvious place in the file.
- I'd probably also want to create a standard interface around redis or other state management tools, so your code isnt tied to one specific store, and then theres a natural place to have the connection management and other tools specific code inside each wrapper class. This would make it work more like a plugin that you could then replace the underlying tool with anything else.  [example in sdk](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/stores/index.ts)
- My philosophy around error handling is to prefer bubbling errors up to the top of the app to crash the app, or "fail fast".  If failure is not an option then you can have a single handler manage all the errors. this simplifies logging as well as the lower level code, as there are less try catches. 
- parse process.env in a single file (index, app or constants) into individual named variables, and pass those variables around to configure the rest of the app, rather than accessing process.env in real time.  This just makes it easier to see all the envs, validate them in one place, set defaults, and document them.

could be more things if i looked closer at the code but these should help general maintainence of the app

pxrl:
> > @daywiss I do think we should do a larger refactor, do you have any other suggestions in addition to what Paul recommended above?
> 
> @nicholaspai Incremental improvements are fine, imo but if you want to refactor some high level things id suggest:
> 
>     * move towards dependency injection, for example instantiating redis and other tools at the top level and passing them into code that needs it.  This makes it easy to see all the configuration needed for the app at an obvious place in the file.
> 
>     * I'd probably also want to create a standard interface around redis or other state management tools, so your code isnt tied to one specific store, and then theres a natural place to have the connection management and other tools specific code inside each wrapper class. This would make it work more like a plugin that you could then replace the underlying tool with anything else.  [example in sdk](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/stores/index.ts)
> 
>     * My philosophy around error handling is to prefer bubbling errors up to the top of the app to crash the app, or "fail fast".  If failure is not an option then you can have a single handler manage all the errors. this simplifies logging as well as the lower level code, as there are less try catches.
> 
>     * parse process.env in a single file (index, app or constants) into individual named variables, and pass those variables around to configure the rest of the app, rather than accessing process.env in real time.  This just makes it easier to see all the envs, validate them in one place, set defaults, and document them.
> 
> 
> could be more things if i looked closer at the code but these should help general maintainence of the app

I agree on each of these points; would love to spend some time to take it in that direction.

src/dataworker/index.ts
nicholaspai: > That change requires a few more modifications, so I'd probably handle it separately so it's easy for us to reason that the behaviour of the bot isn't meaningfully modified.

Makes sense

nicholaspai (approved):
LGTM, and as a follow up @pxrl can you create an issue to track [david's comment](https://github.com/across-protocol/relayer-v2/pull/232#issuecomment-1203897763) and then implement ? I think its a great set of suggestions! Excited to see this top level infra getting more robust

src/dataworker/index.ts
pxrl: > LGTM after resolving all open threads. As a follow up, @pxrl and I should create issues on Linear for refactoring the whole exception handling architecture, as suggested by these threads:
> 
>     * [#232 (comment)](https://github.com/across-protocol/relayer-v2/pull/232#issuecomment-1201262398)
> 
>     * [#232 (comment)](https://github.com/across-protocol/relayer-v2/pull/232#issuecomment-1198250400)

I've now migrated these goals into ACX-122.

src/finalizer/index.ts
pxrl: Follow-up opportunities to move in this direction have been identified in ACX-122.

PR #233, opened 2022-07-28
fix(dataworker): Check for executed pool rebalance leaves until proposal block, not bundle end block

nicholaspai:
This is a bug but AFAICT did not lead to any issues in production. The following situation highlights the bug:

- There was some partial fill in bundle 1.
- Bundle 1 sends slow relay funds.
- Bundle 1 is executed.
- We’re now trying to build bundle 2.
- Our mainnet end block for bundle 2 is before the execution block for bundle 1 because of our buffer.

I think in this case, we consider bundle 1 as not valid yet (reasonable, since this bundle hasn’t really “seen” those blocks yet). This means [this check](https://github.com/across-protocol/relayer-v2/blob/587b9039d27cab2fd0775e08934dd78c8628413f/src/clients/HubPoolClient.ts#L189) returns false for the bundle that would have contained the first fill. This means [this call](https://github.com/across-protocol/relayer-v2/blob/587b9039d27cab2fd0775e08934dd78c8628413f/src/utils/FillUtils.ts#L114) returns undefined, causing both values returned [here](https://github.com/across-protocol/relayer-v2/blob/587b9039d27cab2fd0775e08934dd78c8628413f/src/utils/FillUtils.ts#L129-L132) to be undefined. This means we hit [this return](https://github.com/across-protocol/relayer-v2/blob/587b9039d27cab2fd0775e08934dd78c8628413f/src/dataworker/PoolRebalanceUtils.ts#L199), meaning we ignore this excess. The end result is that a bundle constructed by current code might not return enough excess to mainnet.

The reason why I think this hasn't been an issue in prod yet is TODO.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -131,6 +131,7 @@ export class Dataworker {
 
     return this._getPoolRebalanceRoot(
       blockRangesForChains,
+      spokePoolClients[1].latestBlockNumber,
       endBlockForMainnet,
       fillsToRefund,
       deposits,
@@ -195,6 +196,7 @@ export class Dataworker {
     this.logger.debug({ at: "Dataworker", message: "Building pool rebalance root", blockRangesForProposal });
     const poolRebalanceRoot = this._getPoolRebalanceRoot(
       blockRangesForProposal,
+      spokePoolClients[1].latestBlockNumber,
       endBlockForMainnet,
       fillsToRefund,
       deposits,
@@ -548,6 +550,7 @@ export class Dataworker {
     );
     const expectedPoolRebalanceRoot = this._getPoolRebalanceRoot(
       blockRangesImpliedByBundleEndBlocks,
+      spokePoolClients[1].latestBlockNumber,
       endBlockForMainnet,
       fillsToRefund,
       deposits,
@@ -1115,6 +1118,7 @@ export class Dataworker {
           );
           const expectedPoolRebalanceRoot = this._getPoolRebalanceRoot(
             blockNumberRanges,
+            spokePoolClients[1].latestBlockNumber,
             endBlockForMainnet,
             fillsToRefund,
             deposits,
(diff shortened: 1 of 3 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -118,7 +118,7 @@ export class Dataworker {
       spokePoolClients
     );
 
-    const endBlockForMainnet = getBlockRangeForChain(
+    const mainnetBundleEndBlock = getBlockRangeForChain(
```
nicholaspai, line 121: Renamed this to be more clear its the bundle end block

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -203,7 +203,7 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
       // first fill for this deposit. If it is the same as the ProposeRootBundle event containing the
       // current fill, then the first fill is in the current bundle and we can exit early.
       const rootBundleEndBlockContainingFullFill = hubPoolClient.getRootBundleEvalBlockNumberContainingBlock(
-        endBlockForMainnet,
+        hubPoolClient.latestBlockNumber,
```
nicholaspai, line 206: This is the bugfix: `getRootBundleEvalBlockNumberContainingBlock` should always default to the HEAD mainnet block, not the bundle end block, in the event that a root bundle its searching a fill for is `undefined`.

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -182,7 +182,7 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
     .forEach((fill: interfaces.FillWithBlock) => {
       const { lastFillBeforeSlowFillIncludedInRoot, rootBundleEndBlockContainingFirstFill } =
         getFillDataForSlowFillFromPreviousRootBundle(
-          endBlockForMainnet,
+          hubPoolClient.latestBlockNumber,
```
nicholaspai, line 185: This function calls `getRootBundleEvalBlockNumberContainingBlock` and passes in this arg, so we should use the HEAD mainnet block

mrice32 (approved):
+1 This is awesome!

PR #235, opened 2022-07-29
improve: Correctly use concurrency in paginated Event query

nicholaspai:
Inspired by #198, this fixes an incorrect usage of `concurrency` in `Promise.all` which does not take such a property so no concurrency constraint is being applied. This could cause issues with machines that have limited network bandwidth.

### DIFF
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -58,9 +58,9 @@ export async function paginatedEventQuery(
       )
     ).flat();
   } catch (error) {
-    if (retryCount++ < maxRetries) {
+    if (retryCount < maxRetries) {
       await delay(retrySleepTime);
-      return await paginatedEventQuery(contract, filter, searchConfig, retryCount);
+      return await paginatedEventQuery(contract, filter, searchConfig, retryCount+1);
     } else throw error;
   }
 }
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -32,8 +32,12 @@ export interface EventSearchConfig {
   concurrency?: number | null;
 }
 
-export async function paginatedEventQuery(contract: Contract, filter: EventFilter, searchConfig: EventSearchConfig) {
-  let retryCounter = 0;
+export async function paginatedEventQuery(
+  contract: Contract,
+  filter: EventFilter,
+  searchConfig: EventSearchConfig,
+  retryCount = 0
+) {
   // If the max block look back is set to 0 then we dont need to do any pagination and can query over the whole range.
   if (searchConfig.maxBlockLookBack === 0)
     return await contract.queryFilter(filter, searchConfig.fromBlock, searchConfig.toBlock);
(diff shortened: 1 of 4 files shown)


src/utils/EventUtils.ts
```diff
@@ -42,14 +46,21 @@ export async function paginatedEventQuery(contract: Contract, filter: EventFilte
...
+        paginatedRanges,
+        async ([fromBlock, toBlock]) => {
+          return contract.queryFilter(filter, fromBlock, toBlock);
+        },
+        { concurrency: searchConfig.concurrency | defaultConcurrency }
+      )
+    ).flat();
   } catch (error) {
-    if (retryCounter++ < maxRetries) {
+    if (retryCount++ < maxRetries) {
```
nicholaspai, line 61: I'm not sure how this `retryCounter` was working before since it was always getting re-initialized at 0 on each recursive call. Am I missing something?

src/utils/ProviderUtils.ts
```diff
@@ -53,7 +53,7 @@ class RateLimitedProvider extends ethers.providers.JsonRpcProvider {
   }
 }
 
-const defaultTimeout = 15 * 1000;
+const defaultTimeout = 60 * 1000;
```
nicholaspai, line 56: So I'm on the fence about including this change and leaning towards leaving things as is. Locally when trying to pull events I'd occasionally run into this 15 sec timeout, (usually with Arbitrum-alchemy nodes but I've seen it before with other nodes). Is there a downside to increasing this timeout? Perhaps we should parameterize it?

src/utils/EventUtils.ts
daywiss: yea i think you are right!

src/utils/ProviderUtils.ts
daywiss: i thought there was an environment parameter that overrides this?

nicholaspai: I don't see it anywhere?

daywiss: https://github.com/across-protocol/relayer-v2/blob/master/src/utils/ProviderUtils.ts#L238

mrice32: I think defaulting to 60 makes sense!

mrice32 (approved):
LGTM!

daywiss (approved):

kevinuma (approved):

src/utils/EventUtils.ts
kevinuma: nit: It's strange to update a local copy. How about retryCount < maxRetries and then pass retryCount + 1 in the call on line 63?

src/utils/EventUtils.ts
```diff
@@ -42,14 +46,21 @@ export async function paginatedEventQuery(contract: Contract, filter: EventFilte
...
+      await Promise.map(
+        paginatedRanges,
+        async ([fromBlock, toBlock]) => {
+          return contract.queryFilter(filter, fromBlock, toBlock);
+        },
+        { concurrency: searchConfig.concurrency | defaultConcurrency }
+      )
+    ).flat();
   } catch (error) {
-    if (retryCounter++ < maxRetries) {
```
kevinuma, line 50: So currently, it'd infinitely retry as retryCounter is always at most 1?

src/utils/EventUtils.ts
nicholaspai: I like that, its more readable

src/utils/EventUtils.ts
nicholaspai: yeah....

kevinuma: This explains the hanging behavior when things timeout - it'd just keep retrying forever and unnecessarily flood the provider with a lot of requests. Plus, the increased timeout definitely would help.

PR #238, opened 2022-08-04
improve(dataworker): Buffer between proposals should start at block of oldest executed relayer refund leaf, not newest

nicholaspai:
The dataworker currently waits until after the most recently proposed bundle has been executed, and that means it not only waits for the PoolRebalanceLeaves to be executed but also for the subsequent
RelayerRefundLeaf to be executed on Mainnet. In serverless mode, this  means the dataworker will wait at least 2 iterations between proposing new bundles. In practice, the buffer is ~2 hours after
the newest RelayerRefundLeaf on Mainnet is executed, however this buffer can be a lot longer than expected if the leaf executions are delayed for some reason. It's reasonable to start this buffer
after the oldest RelayerRefundLeaf assuming the buffer is set to something reasonable like >30 mins.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -252,15 +252,15 @@ export class Dataworker {
             value: true,
             mostRecentValidatedBundle: mostRecentValidatedBundle.blockNumber,
             mostRecentEthereumRootBundle: mostRecentEthereumRootBundle.rootBundleId,
-            latestExecutedLeaf: undefined,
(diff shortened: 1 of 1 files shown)


kevinuma:
Nice! Do we have a test for this?

kevinuma (approved):

nicholaspai:
> Nice! Do we have a test for this?

No we don't yet

mrice32 (approved):
LGTM!

PR #239, opened 2022-08-04
improve: More robustness in CoinGecko price retrieval

pxrl:
Today we had many incidents where various bots were unable to retrieve
prices from CoinGecko. In each incident, *all* requested token prices
were unavailable. This indicates either a network/transport issue, or
some error over at CoinGecko. Unfortunately I wasn't able to see any
logging of the underlying cause (maybe I wasn't looking in the right
place).

In any case, this change introduces the following modifications:

 - Bundle token price requests, such that we perform one HTTP request
   in total, instead of 1 per token (currently 7).

 - Support up to 2 retries on a failed HTTP request. Generate events
   for retries in the debug log, but only generate warnings if we
   totally fail to retrieve a price.

 - When exceptions occur during retrieval and parsing, ensure that the
   underlying cause is logged.

Ref: ACX-125

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -40,11 +40,12 @@ export class ProfitClient {
   }
 
   getPriceOfToken(token: string) {
-    if (!this.tokenPrices[token]) {
+    const _token: string = token.toLowerCase();
+    if (!this.tokenPrices[_token]) {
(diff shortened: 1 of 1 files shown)


pxrl:
...Fat-fingered that one...it was supposed to be a draft PR. This works OK on my system in the sunshine scenario, but I am lacking a test case for it so I hadn't intended to propose it yet.

PR #240, opened 2022-08-04
improve: Reduce overheads in CoinGecko price retrieval

pxrl:
Today we had many incidents where various bots were unable to retrieve
prices from CoinGecko. In each incident, *all* requested token prices
were unavailable. This indicates either a network/transport issue, or
some error over at CoinGecko. Unfortunately I wasn't able to see any
logging of the underlying cause (maybe I wasn't looking in the right
place).

In any case, this change introduces the following modifications:

 - Bundle token price requests, such that we perform one HTTP request
   in total, instead of 1 per token (currently 7).

 - When exceptions occur during retrieval and parsing, ensure that the
   underlying cause is logged.
   
An additional part of this PR was to support retries on the CoinGecko HTTP
request (nominally 2), but that was stripped out after feedback from
David.

Ref: ACX-125

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -3,6 +3,13 @@ import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
 import { Coingecko } from "@uma/sdk";
 
+// Copied from @uma/sdk/coingecko. Propose to export export it upstream in the sdk.
+type CoinGeckoPrice = {
+  address: string;
+  timestamp: number;
+  price: number;
+};
+
 // Define the minimum revenue, in USD, that a relay must yield in order to be considered "profitable". This is a short
 // term solution to enable us to avoid DOS relays that yield negative profits. In the future this should be updated
 // to actual factor in the cost of sending transactions on the associated target chains.
@@ -40,12 +47,11 @@ export class ProfitClient {
   }
 
   getPriceOfToken(token: string) {
-    const _token: string = token.toLowerCase();
-    if (!this.tokenPrices[_token]) {
+    if (!this.tokenPrices[token]) {
       this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
       return toBN(0);
     }
-    return this.tokenPrices[_token];
+    return this.tokenPrices[token];
   }
 
   getUnprofitableFills() {
@@ -99,44 +105,39 @@ export class ProfitClient {
   }
 
   async update() {
-    const l1Tokens = Object.fromEntries(
-      this.hubPoolClient.getL1Tokens().map((token) => [token["address"].toLowerCase(), { symbol: token["symbol"] }])
+    const l1Tokens: { [k: string]: L1Token } = Object.fromEntries(
+      this.hubPoolClient.getL1Tokens().map((token) => [token["address"], token])
     );
 
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    let cgRetries = 2;
-    let errors: Array<{ [k: string]: string }>;
-    do {
-      errors = [];
-      try {
-        const cgPrices: Array<any> = await this.coingeckoPrices(Object.keys(l1Tokens));
-        cgPrices.forEach((priceFeed) => {
-          const address: string = priceFeed["address"].toLowerCase();
-          const price: number = priceFeed["price"];
+    let cgPrices: Array<CoinGeckoPrice> = [];
+    try {
+      cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
+    } catch (err) {
+      this.logger.warn({
+        message: "Failed to retrieve prices.",
+        err,
+        l1Tokens,
+      });
+    }
 
-          // todo: Any additional validation to perform?
-          //       Ensure that timestamps are always moving forwards?
-          if (typeof price !== "number") {
-            this.logger.debug({
-              message: "Unexpected CoinGecko price response.",
-              priceFeed,
-            });
-            errors.push({ address: address, symbol: priceFeed["symbol"] });
-          } else this.tokenPrices[address] = toBN(price);
-        });
-      } catch (err) {
-        this.logger.debug({
-          message: `Failed to retrieve CoinGecko prices (${cgRetries} tries left).`,
-          err,
-          l1Tokens,
-        });
-      }
-    } while (errors.length === 0 && cgRetries--);
+    const errors: Array<{ [k: string]: string }> = [];
+    for (const address of Object.keys(l1Tokens)) {
+      const tokenPrice: CoinGeckoPrice = cgPrices.find((price) => address.toLowerCase() === price.address);
+
+      // todo: Any additional validation to do? Ensure that timestamps are always moving forwards?
+      if (tokenPrice === undefined || typeof tokenPrice.price !== "number") {
+        const symbol = l1Tokens[address].symbol;
+        const errmsg = tokenPrice ? "Unexpected price response." : `Missing price for ${symbol}.`;
+        this.logger.warn({ message: errmsg, tokenPrice });
+        errors.push({ address: address, symbol: symbol });
+      } else this.tokenPrices[address] = toBNWei(tokenPrice.price);
+    }
 
     if (errors.length > 0) {
       let mrkdwn = "The following L1 token prices could not be fetched:\n";
-      errors.forEach((token: { [k: string]: string }) => {
-        mrkdwn += `- ${token["symbol"]} Not found. Using last known price of ${this.getPriceOfToken(token.address)}.\n`;
+      errors.forEach((token) => {
+        mrkdwn += `- ${token.symbol} Not found. Using last known price of ${this.getPriceOfToken(token.address)}.\n`;
       });
       this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
     }
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -40,11 +40,12 @@ export class ProfitClient {
   }
 
   getPriceOfToken(token: string) {
-    if (!this.tokenPrices[token]) {
+    const _token: string = token.toLowerCase();
+    if (!this.tokenPrices[_token]) {
       this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
       return toBN(0);
     }
-    return this.tokenPrices[token];
+    return this.tokenPrices[_token];
   }
 
   getUnprofitableFills() {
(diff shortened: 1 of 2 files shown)


pxrl:
FWIW this works in the sunshine scenario, but it still needs a test case.

src/clients/ProfitClient.ts
```diff
@@ -98,26 +99,51 @@ export class ProfitClient {
...
+          } else this.tokenPrices[address] = toBN(price);
+        });
+      } catch (err) {
+        this.logger.debug({
+          message: `Failed to retrieve CoinGecko prices (${cgRetries} tries left).`,
+          err,
+          l1Tokens,
+        });
+      }
+    } while (errors.length === 0 && cgRetries--);
```
daywiss, line 134: this logic seems wrong to me, it seems like you would want to keep looping if you run into errors

src/clients/ProfitClient.ts
```diff
@@ -98,26 +99,51 @@ export class ProfitClient {
...
+      errors = [];
+      try {
+        const cgPrices: Array<any> = await this.coingeckoPrices(Object.keys(l1Tokens));
+        cgPrices.forEach((priceFeed) => {
+          const address: string = priceFeed["address"].toLowerCase();
+          const price: number = priceFeed["price"];
+
+          // todo: Any additional validation to perform?
+          //       Ensure that timestamps are always moving forwards?
+          if (typeof price !== "number") {
```
daywiss, line 119: generalize your error handler so you can just throw errors as a guard to the code, this will reduce nesting

```js
  assert(typeof price === "number", "Unexpected Coingecko Price response")
  // resume code
  // handle logging the error message in the catch block
```

```

**Review by daywiss — COMMENTED** [review]
personally i think id just experiment with a single improvement first ( batching coingecko price requests), and if that still causes issues add in the more complex and hard to test retry logic

File: src/clients/ProfitClient.ts [hunk 3]
```diff
@@ -98,26 +99,51 @@ export class ProfitClient {
...
+        mrkdwn += `- ${token["symbol"]} Not found. Using last known price of ${this.getPriceOfToken(token.address)}.\n`;
       });
       this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
     }
     this.logger.debug({ at: "ProfitClient", message: "Updated Profit client", tokenPrices: this.tokenPrices });
   }
 
-  private async coingeckoPrice(token: string) {
-    return await this.coingecko.getCurrentPriceByContract(token, "usd");
+  private async coingeckoPrices(tokens: Array<string>) {
```
daywiss, line 146: dont modify the old call, add a new one to maintain backward compatibility

src/clients/ProfitClient.ts
```diff
@@ -98,26 +99,51 @@ export class ProfitClient {
...
+    );
+
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    const prices = await Promise.allSettled(l1Tokens.map((l1Token: L1Token) => this.coingeckoPrice(l1Token.address)));
+    let cgRetries = 2;
+    let errors: Array<{ [k: string]: string }>;
+    do {
+      errors = [];
+      try {
+        const cgPrices: Array<any> = await this.coingeckoPrices(Object.keys(l1Tokens));
```
daywiss, line 112: this is a really good candidate for some type validation. we have precedent using superstruct in other code.  not strictly required but if you are worried about being safe, that will help.  [example here](https://github.com/across-protocol/sdk-v2/blob/master/src/contracts/acrossConfigStore.ts#L6)

src/clients/ProfitClient.ts
pxrl: Have removed the loop, per another of your comments. I'm not entirely convinced that our woes will disappear, but it's super easy to add in the looped request later.

src/clients/ProfitClient.ts
pxrl: This has been updated to expect an array of CoinGeckoPrice objects. That's actually defined in uma/sdk/CoinGecko, so I've temporarily implemented it in ProfitClient. I would like to hear your thoughts on exporting it directly from the UMA SDK though.

src/clients/ProfitClient.ts
pxrl: The code has now been split into a try .. catch wrapping the CoinGecko retrieval, and there is a subsequent loop that validates all the input. The assert() therefore wouldn't make sense in the new context.

I was also a bit curious - is it standard to assert() to catch runtime errors that may forseeably happen? Seems a little "violent" - even if it is caught and mitigated by a handler.

src/clients/ProfitClient.ts
```diff
@@ -3,6 +3,13 @@ import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
 import { Coingecko } from "@uma/sdk";
 
+// Copied from @uma/sdk/coingecko. Propose to export export it upstream in the sdk.
+type CoinGeckoPrice = {
+  address: string;
+  timestamp: number;
+  price: number;
+};
```
pxrl, line 11: Just to be explicit - this should only be here temporarily, pending a conclusion to whether we can export it from the UMA/sdk package.

test/ProfitClient.PriceRetrieval.ts
```diff
@@ -19,17 +27,15 @@ describe("ProfitClient: Price Retrieval", async function () {
...
-    // The client should have fetched the prices for both tokens.
-    expect(Object.keys(profitClient.getAllPrices())).to.deep.equal([mainnetWeth, mainnetUsdc]);
-    Object.values(profitClient.getAllPrices()).forEach(
-      (price: BigNumber) => expect(toBN(price).gt(toBN(0))).to.be.true
-    );
-    Object.keys(profitClient.getAllPrices()).forEach(
+    const tokenPrices: { [k: string]: BigNumber } = profitClient.getAllPrices();
+
+    // The client should have fetched prices for all requested tokens.
+    expect(Object.keys(tokenPrices)).to.have.deep.members(mainnetTokens.map((token) => token["address"]));
```
pxrl, line 36: Thanks to @daywiss for whacking me with the cluebat on this one!

src/clients/ProfitClient.ts
```diff
@@ -98,26 +105,46 @@ export class ProfitClient {
...
+        const symbol = l1Tokens[address].symbol;
+        const errmsg = tokenPrice ? "Unexpected price response." : `Missing price for ${symbol}.`;
+        this.logger.warn({ message: errmsg, tokenPrice });
+        errors.push({ address: address, symbol: symbol });
+      } else this.tokenPrices[address] = toBNWei(tokenPrice.price);
     }
+
     if (errors.length > 0) {
       let mrkdwn = "The following L1 token prices could not be fetched:\n";
-      errors.forEach((token: L1Token) => {
```
pxrl, line 112: Seems like an error to remove the type information from token.

src/clients/ProfitClient.ts
daywiss: i think its just up to the circumstances, since this is only accessed internally then yea its probably ok to remove.

src/clients/ProfitClient.ts
daywiss: yea it would be nice to re-use and add functionality to that coingecko class in the sdk, but we dont have any strict guidelines around this. i usually will just copy stuff out like you did if i only need a very small peice, rather than including the entire package

src/clients/ProfitClient.ts
daywiss: yes those should probably all be exported in the sdk. feel free to submit a change to do that

src/clients/ProfitClient.ts
daywiss: nothing is standard! but typically i will always try and throw errors because that will get you a proper stack trace.  i like `assert` because i find it nicer than the alternative `if(something) throw new Error('blah')`.  WIth some of your logs you only add data and no stack information on where its originating, which just clues you in a bit when debugging.

daywiss (approved):

pxrl:
I propose to get this in for now, and I will follow up with a bump to the UMA SDK version once I get the CoinGeckoPrice update submitted over there.

src/clients/ProfitClient.ts
pxrl: Point taken - and I have observed that the logging-only approach can be tricky when there are duplicate log messages in the code.

FWIW on a previous project I did take the approach of always assert()ing on all errors - runtime errors and exceptions - but in the end I had the feeling that I was misusing the language by not using exceptions. On the definition of "runtime error" and "exception", I like this one from "[C Interfaces & Implementations](https://dl.acm.org/doi/book/10.5555/235284)":

> Interfaces often specify _unchecked runtime errors_, _checked runtime errors_ and _exceptions_. Unchecked and checked runtime errors are not expected user errors, such as failing to open a file. Runtime errors are breaches of the contract between clients and implementations, and are program bugs from which there is no recovery. Exceptions are conditions that, while possible, rarely occur. Programs may be able to recover from exceptions. Running out of memory is an example. ... An unchecked runtime error is a breach of contract that implementations do not guarantee to detect. If an unchecked runtime error occurs, execution might continue, but with unpredictable and perhaps unrepeatable results. Good interfaces avoid unchecked runtime errors when possible, but must specify those that can occur.

In the context of the above quote, I guess it feels natural for me to assert on runtime errors, and otherwise throw an exception.

pxrl:
The original version of this code supported retrying the call to coingecko. I've just found that there's a local copy of the coingecko class, with retry support added.

https://github.com/across-protocol/sdk-v2/blob/e73c0b8a591cbdb446c0d318bef76b2246f12731/src/coingecko/Coingecko.ts

So if retries are necessary, I might switch over to the Across sdk version. Might also be worth considering whether we want to re-merge these two implementations somewhere common.

src/clients/ProfitClient.ts
```diff
@@ -98,26 +105,47 @@ export class ProfitClient {
...
+
+    const errors: Array<{ [k: string]: string }> = [];
+    for (const address of Object.keys(l1Tokens)) {
+      const tokenPrice: CoinGeckoPrice = cgPrices.find((price) => address.toLowerCase() === price.address);
 
-    const errors = [];
-    for (const [index, priceResponse] of prices.entries()) {
-      if (priceResponse.status === "rejected") errors.push(l1Tokens[index]);
-      else this.tokenPrices[l1Tokens[index].address] = toBNWei(priceResponse.value[1]);
+      // todo: Any additional validation to do? Ensure that timestamps are always moving forwards?
```
nicholaspai, line 128: What do you mean by timestamps are always moving forwards?

src/clients/ProfitClient.ts
```diff
@@ -98,26 +105,47 @@ export class ProfitClient {
...
-    const l1Tokens = this.hubPoolClient.getL1Tokens();
+    const l1Tokens: { [k: string]: L1Token } = Object.fromEntries(
+      this.hubPoolClient.getL1Tokens().map((token) => [token["address"], token])
+    );
+
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    const prices = await Promise.allSettled(l1Tokens.map((l1Token: L1Token) => this.coingeckoPrice(l1Token.address)));
+    let cgPrices: Array<CoinGeckoPrice> = [];
+    try {
+      cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
```
nicholaspai, line 115: How does this consolidate HTTP requests? Feell like i'm missing something obvious

nicholaspai (commented):
Left some questions!

src/clients/ProfitClient.ts
pxrl: It's not really obvious here, but this.coingeckoPrice() has become this.coinGeckoPrices() (i.e. plural). Instead of submitting n * HTTP requests containing a single token address in parallel, it submits a single HTTP request with a comma-separated list of token addresses.

src/clients/ProfitClient.ts
pxrl: It was just an idea, but in looping mode, we would validate that the timestamp for the price that we have just retrieved is more recent than the one we previously got. The scenarios where this validation would be triggered are probably mostly theoretical though, to be fair.

src/clients/ProfitClient.ts
```diff
@@ -3,6 +3,13 @@ import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
 import { Coingecko } from "@uma/sdk";
 
+// Copied from @uma/sdk/coingecko. Propose to export export it upstream in the sdk.
```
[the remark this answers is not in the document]
pxrl: I only found that we had a local variant of this in the Across SDK after I did the work, but I generally prefer keeping the dependencies "internal" to Across rather than relying on the UMA SDK.

It does make me wonder though - should we try to harmonise the two implementations and put them somewhere else? I haven't dug into the CoinGecko Pro to understand whether there are any API differences, but I'd imagine we could allow the user of the CoinGecko class to specify a custom CoinGecko endpoint, and otherwise fall back to the public one.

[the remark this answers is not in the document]
nicholaspai: the Pro and Basic API interfaces are the same. I think we wanted to hold off on using the Pro API for now in all places except the frontend so lets just leave as is for now

src/clients/ProfitClient.ts
nicholaspai: Ah gotcha, maybe just clarify this comment but its a good one

src/clients/ProfitClient.ts
nicholaspai: Oh nice!

test/ProfitClient.PriceRetrieval.ts
```diff
@@ -19,17 +27,15 @@ describe("ProfitClient: Price Retrieval", async function () {
...
+    Object.values(tokenPrices).forEach((price: BigNumber) => expect(toBN(price).gt(toBN(0))).to.be.true);
```
nicholaspai, line 37: Does this test against the real API?

nicholaspai (commented):
One more question about how much confidence we should take from the unit test, otherwise LGTM!

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/240#discussion_r943763957

Just one comment and then LGTM

mrice32:
https://github.com/across-protocol/relayer-v2/pull/240#discussion_r943855807

Sorry had some queued review comments that I never posted.

src/clients/ProfitClient.ts
```diff
@@ -98,26 +105,47 @@ export class ProfitClient {
...
+      this.logger.warn({
+        message: "Failed to retrieve prices.",
+        err,
+        l1Tokens,
+      });
+    }
+
+    const errors: Array<{ [k: string]: string }> = [];
+    for (const address of Object.keys(l1Tokens)) {
+      const tokenPrice: CoinGeckoPrice = cgPrices.find((price) => address.toLowerCase() === price.address);
```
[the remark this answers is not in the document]
pxrl: Yeah, I can't say it couldn't happen. I'll adjust it to convert their response to lower case as well.

pxrl:
> One more question about how much confidence we should take from the unit test, otherwise LGTM!

I'm working on unit test improvements now.

src/clients/ProfitClient.ts
pxrl: Will adjust it in a follow-up change.

PR #241, opened 2022-08-05
fix(reporter): Query balances at the same block number as events to avoid double counting

kevinuma:
Context: https://risklabs.slack.com/archives/C03JLFNQNDV/p1659046599069139

Problem: Around the time of relayer refund leaf executions we often see double counting in the balances report where it shows updated balances (including refunds) but there are still pending refunds. This likely indicates that the reporter bot didn't see the refund leaf executions yet but the balances it got are still updated. One possible root cause of this is the spokepoolclients are updated at the beginning of the bot run so they might not see the execution events yet. However when balances are queried later on, the provider node might see now see the executions => updated balances.

Solution: Query balances with the same latest block number the SpokePoolClients see. This ensures both the execution events and balances are in sync (either both outdated or both updated).

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -686,6 +686,7 @@ export class Monitor {
     };
   }
 
+  // Returns balances from cache or from provider if there's a cache miss.
   private async _getBalances(
     balanceRequests: { chainId: number; token: string; account: string }[]
   ): Promise<BigNumber[]> {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
This is a great solution! I agree with your problem diagnosis where balance is queried using later block number than client. 

I think it's better to have the reporter slightly lagging HEAD than double count

PR #244, opened 2022-08-09
improve: Consolidate use of getSigner() during startup

pxrl:
At very least this will leave fewer copies of the private key in memory.
For gckms configurations it should also save a few redundant connections
to Google on each run of the bot.

Ref: ACX-40
Ref: ACX-122


nicholaspai (approved):
LGTM

PR #245, opened 2022-08-10
improve: Drop excess per-chain multicall chunks

pxrl:
This change modifies the bot's transaction handling when a chain has > x
relay fills to be submitted in a multicall transaction. The approach is to
basically give it a lobotomy, so it just truncates the list of transactions to
ensure that we will always fit within the configured chunk size limit.

The current implementation doesn't work as intended due to the fact that
successive chunks are submitted in parallel, so they all tend to get the
same nonce. This means that a maximum of one multicall txn can succeed
per chain. On top of this, the successful multicall seems to often be
the smallest, which means that a large number of fills are "blocked"
until demand slows down and the bot has a chance at submitting the same
multicall txn without any new Deposit candidates arriving.

The change itself is pretty lightweight, and deliberately retains compatibility
with the surrounding code - i.e. by pretending that we might actually still
see multiple multicall chunks per chain. This makes it easier to re-add that
capability once we have a test environment that will enable proper
verification of that behaviour.
 
As part of this change, also add support for per-chain multicall chunk size
configuration. It seems likely that this flexibility might be needed at some
point in the future, so that we don't have to spend time finding the
goldilocks chunk size. Finally, while here, it addressed a small nit in some
markdown formatting in the same area.

Ref: ACX-98

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -11,8 +11,6 @@ import {
   TransactionResponse,
 } from "../utils";
 
-import lodash from "lodash";
-
 export interface AugmentedTransaction {
   contract: Contract;
   chainId: number;
@@ -112,8 +110,21 @@ export class MultiCallerClient {
         assign(groupedTransactions, [transaction.chainId], [transaction]);
       }
 
+      const chunkSize = 100;
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
-        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 100)])
+        Object.entries(groupedTransactions).map(([chainId, transactions]) => {
+          if (transactions.length > chunkSize) {
+            const dropped: Array<{ [k: string]: string }> = transactions.slice(chunkSize).map((txn) => {
+              return { address: txn.contract.address, method: txn.method, args: txn.args };
+            });
+            this.logger.info({
+              message: `Dropping ${dropped.length} transactions on chain ${chainId}.`,
+              dropped,
+            });
+          }
+          // Multi-chunks not attempted due to nonce reuse complications, but may return in future.
+          return [chainId, [transactions.slice(0, chunkSize)]];
+        })
       );
 
       if (simulationModeOn) {
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -183,7 +183,7 @@ export class MultiCallerClient {
       this.logger.debug({
         at: "MultiCallerClient",
         message: "Executing transactions grouped by target chain",
-        txs: Object.keys(groupedTransactions).map((chainId) => ({ chainId, num: groupedTransactions[chainId].length })),
+        txs: Object.entries(chunkedTransactions).map(([chainId, txns]) => ({ chainId, num: txns.length })),
       });
 
       // Construct multiCall transaction for each target chain.
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -1,4 +1,4 @@
-import { CHAIN_MULTICALL_MAX_CHUNK_SIZE } from "../common";
+import { DEFAULT_MULTICALL_CHUNK_SIZE, CHAIN_MULTICALL_CHUNK_SIZE } from "../common";
 import {
   winston,
   getNetworkName,
@@ -113,7 +113,7 @@ export class MultiCallerClient {
 
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
         Object.entries(groupedTransactions).map(([chainId, transactions]) => {
-          const chunkSize: number = CHAIN_MULTICALL_MAX_CHUNK_SIZE[chainId];
+          const chunkSize: number = CHAIN_MULTICALL_CHUNK_SIZE[chainId] || DEFAULT_MULTICALL_CHUNK_SIZE;
           if (transactions.length > chunkSize) {
             const dropped: Array<{ [k: string]: string }> = transactions.slice(chunkSize).map((txn) => {
               return { address: txn.contract.address, method: txn.method, args: txn.args };
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -136,7 +136,7 @@ export class MultiCallerClient {
         let mrkdwn = "";
         valueTransactions.forEach((transaction, i) => {
           mrkdwn += "*Transaction excluded from batches because it contained value:*\n";
-          mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ` + `${transaction.mrkdwn || "0 mrkdwn"}\n`;
+          mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ${transaction.mrkdwn || "0 mrkdwn"}\n`;
         });
         Object.keys(chunkedTransactions).forEach((chainId) => {
           mrkdwn += `*Transactions sent in batch on ${getNetworkName(chainId)}:*\n`;
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -1,3 +1,4 @@
+import { CHAIN_MULTICALL_MAX_CHUNK_SIZE } from "../common";
 import {
   winston,
   getNetworkName,
@@ -110,9 +111,9 @@ export class MultiCallerClient {
         assign(groupedTransactions, [transaction.chainId], [transaction]);
       }
 
-      const chunkSize = 100;
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
         Object.entries(groupedTransactions).map(([chainId, transactions]) => {
+          const chunkSize: number = CHAIN_MULTICALL_MAX_CHUNK_SIZE[chainId];
           if (transactions.length > chunkSize) {
             const dropped: Array<{ [k: string]: string }> = transactions.slice(chunkSize).map((txn) => {
               return { address: txn.contract.address, method: txn.method, args: txn.args };
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -28,6 +28,14 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   42161: 3000, // At a conservative 10 TPS, 300 seconds = 3000 transactions. And 1 block per txn.
 };
 
+export const CHAIN_MULTICALL_MAX_CHUNK_SIZE: { [chainId: number]: number } = {
+  1: 100,
+  10: 100,
+  137: 100,
+  288: 100,
+  42161: 100,
+};
+
 // The most critical failure mode that can happen in the inventory management module is a miss-mapping between L1 token
 //  and the associated L2 token. If this is wrong the bot WILL delete money. The mapping below is used to enforce that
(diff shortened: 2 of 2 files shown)


pxrl:
Here's an example of the "Dropping..." log message:

`2022-08-10 14:34:32 [ESC[32minfoESC[39m]: {
  "message": "Dropping 3 transactions on chain 137.",
  "dropped": [
    {
      "address": "0x69B5c72837769eF1e7C164Abc6515DcFf217F920",
      "method": "fillRelay",
      "args": [
        "0xa74FB7544d54a1d8f8178B6aE3f2700ebd3A65ba",
        "0xa74FB7544d54a1d8f8178B6aE3f2700ebd3A65ba",
        "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
        "30000000",
        "30000000",
        137,
        10,
        "754528997619040",
        "479337833333333",
        43610
      ]
    },
    {
      "address": "0x69B5c72837769eF1e7C164Abc6515DcFf217F920",
      "method": "fillRelay",
      "args": [
        "0xA7f7e4fE8E4cdDCD9969Bd3fbcFF67000CD7DE47",
        "0xB4d00788F75F27b85285b3aF21Bdc16b3336D91E",
        "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
        "120000",
        "120000",
        137,
        10,
        "754528996165321",
        "27550000018000000",
        43584
      ]
    },
    {
      "address": "0x69B5c72837769eF1e7C164Abc6515DcFf217F920",
      "method": "fillRelay",
      "args": [
        "0x00899ADda769602c9526043158585099bEC8829A",
        "0x00899ADda769602c9526043158585099bEC8829A",
        "0x1BFD67037B42Cf73acF2047067bd4F2C47D9BfD6",
        "109690",
        "109690",
        137,
        10,
        "938712703112303",
        "637434908247788",
        43619
      ]
    }
  ],
  "bot-identifier": "xxxxx"
}`

src/clients/MultiCallerClient.ts
```diff
@@ -124,7 +136,7 @@ export class MultiCallerClient {
         let mrkdwn = "";
         valueTransactions.forEach((transaction, i) => {
           mrkdwn += "*Transaction excluded from batches because it contained value:*\n";
-          mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ` + `${transaction.mrkdwn || "0 mrkdwn"}\n`;
+          mrkdwn += `  ${i + 1}. ${transaction.message || "0 message"}: ${transaction.mrkdwn || "0 mrkdwn"}\n`;
```
pxrl, line 139: This is just a driveby fix while I was in the neighbourhood - not related to this PR, but also not really worth an entire separate PR.

src/common/Constants.ts
```diff
@@ -28,6 +28,14 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   42161: 3000, // At a conservative 10 TPS, 300 seconds = 3000 transactions. And 1 block per txn.
 };
 
+export const CHAIN_MULTICALL_MAX_CHUNK_SIZE: { [chainId: number]: number } = {
+  1: 100,
+  10: 100,
+  137: 100,
+  288: 100,
```
pxrl, line 35: This is intended to be static - i.e. not user-configurable. Therefore I have avoided bundling it into the common config implementation. A bit unsure as to whether that's desirable or not.

src/common/Constants.ts
```diff
@@ -28,6 +28,14 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
...
+  42161: 100,
+};
```
pxrl, line 37: This is intended to be static - i.e. not user-configurable. Therefore I have avoided bundling it into the common config implementation. A bit unsure as to whether that's desirable or not.

src/clients/MultiCallerClient.ts
```diff
@@ -113,7 +112,20 @@ export class MultiCallerClient {
       }
 
       const chunkedTransactions: { [networkId: number]: AugmentedTransaction[][] } = Object.fromEntries(
-        Object.entries(groupedTransactions).map(([chainId, transactions]) => [chainId, lodash.chunk(transactions, 100)])
+        Object.entries(groupedTransactions).map(([chainId, transactions]) => {
+          const chunkSize: number = CHAIN_MULTICALL_MAX_CHUNK_SIZE[chainId];
```
pxrl, line 116: Probably should have a safe fallback in case this is undefined.

nicholaspai: Yes, 100 seems like a fair fallback?

src/clients/MultiCallerClient.ts
```diff
@@ -113,7 +112,20 @@ export class MultiCallerClient {
...
+          if (transactions.length > chunkSize) {
+            const dropped: Array<{ [k: string]: string }> = transactions.slice(chunkSize).map((txn) => {
+              return { address: txn.contract.address, method: txn.method, args: txn.args };
+            });
+            this.logger.info({
```
nicholaspai, line 121: So the implication of this is that droppped txns will just get included in the next iteration right?

nicholaspai (commented):
Nice changes, i'm holding off on approving to let others think of new failure cases introduced

src/clients/MultiCallerClient.ts
pxrl: Yep. Per this change, those excess txns are simply not included in the multicall, so they should be collected on the next run of the bot. In cases where the overflow was only a small handful of transactions, this at least offers some minor "gas efficiency" gain because it allows extra time for additional deposit transactions to be placed. With a more decentralised relayer network, it might alternatively be that another relayer handles it in the mean time.

pxrl:
> Can you add some unit tests that test the chunking logic?

I've been considering how to do this, but so far it seems to be disproportionately more work to test this via unit test than it was to implement the actual change. This is mostly because the chunking functionality is embedded in the middle of the executeTransactionQueue() method, and is surrounded by lots of transforming of transaction groups according to various critiera, as well as RPC submission at the end. All of this needs to be mocked in order to test a few lines of code in the middle, which feels like the wrong approach to me.

I think the entire class would be better supported by unit test if a lot of the on-the-fly sorting of transactions was handled directly in the enqueue() method. This would probably reduce the length of executeTransactionQueue() quite a bit and make it faster and easier to focus on the more core functionalities like chunking and RPC submission.

Of course...I probably haven't considered all the angles here - hit me with the cluebat if you have another idea :)

FWIW, my actual method of testing this change was to hack together the following patch, which handles the following:
 - Don't filter out transactions for deposits that already got filled or are considered to be too old.
 - Don't filter out transactions that would otherwise fail.
 - Reduce the chunk size to 1.
 - Throw an exception as late as possible, but before multicall transaction submission (i.e. as a safeguard).

Here's the patch:
`
From 73d14bc5a18ed9513030b130445b9a6781fe6834 Mon Sep 17 00:00:00 2001
From: Paul <108695806+pxrl@users.noreply.github.com>
Date: Wed, 10 Aug 2022 14:38:12 +0000
Subject: [PATCH] xxx test hacks

---
 src/clients/MultiCallerClient.ts | 5 +++--
 src/common/Constants.ts          | 2 +-
 src/utils/FillUtils.ts           | 6 ------
 src/utils/TransactionUtils.ts    | 1 +
 4 files changed, 5 insertions(+), 9 deletions(-)

diff --git a/src/clients/MultiCallerClient.ts b/src/clients/MultiCallerClient.ts
index fb33a02..1659b4e 100644
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -53,7 +53,9 @@ export class MultiCallerClient {
 
       // Simulate the transaction execution for the whole queue.
       const _transactionsSucceed = await Promise.all(
-        this.transactions.map((transaction: AugmentedTransaction) => willSucceed(transaction))
+        this.transactions.map((transaction: AugmentedTransaction) => {
+          return { transaction, succeed: true, reason: null };
+        })
       );
 
       // Filter out transactions that revert for expected reasons. For example, the "relay filled" error
@@ -90,7 +92,6 @@ export class MultiCallerClient {
           notificationPath: "across-error",
         });
       const validTransactions: AugmentedTransaction[] = transactionsSucceed
-        .filter((transaction) => transaction.succeed)
         .map((transaction) => transaction.transaction);
 
       if (validTransactions.length === 0) {
diff --git a/src/common/Constants.ts b/src/common/Constants.ts
index a1eec09..6fc3383 100644
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -28,7 +28,7 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   42161: 3000, // At a conservative 10 TPS, 300 seconds = 3000 transactions. And 1 block per txn.
 };
 
-export const DEFAULT_MULTICALL_CHUNK_SIZE = 100;
+export const DEFAULT_MULTICALL_CHUNK_SIZE = 1;
 export const CHAIN_MULTICALL_CHUNK_SIZE: { [chainId: number]: number } = {
   1: DEFAULT_MULTICALL_CHUNK_SIZE,
   10: DEFAULT_MULTICALL_CHUNK_SIZE,
diff --git a/src/utils/FillUtils.ts b/src/utils/FillUtils.ts
index cacb453..c537780 100644
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -168,12 +168,6 @@ export function getUnfilledDeposits(
         true
       );
       const unfilledDepositsForDestinationChain = depositsForDestinationChain
-        .filter((deposit) => {
-          // If deposit is older than unfilled deposit lookback, ignore it
-          const lookback = maxUnfilledDepositLookBack[deposit.originChainId];
-          const latestBlockForOriginChain = originClient.latestBlockNumber;
-          return lookback === undefined || deposit.originBlockNumber >= latestBlockForOriginChain - lookback;
-        })
         .map((deposit) => {
           // Remove block number that we don't need anymore and because function expects to return a Deposit type.
           const { blockNumber, originBlockNumber, ...depositData } = deposit;
diff --git a/src/utils/TransactionUtils.ts b/src/utils/TransactionUtils.ts
index 052704b..49d3a5e 100644
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -13,6 +13,7 @@ export async function runTransaction(
   gasLimit: BigNumber | null = null
   try {
+    throw "XXX test";
     const gas = await getGasPrice(contract.provider);
     logger.debug({ at: "TxUtil", message: "Send tx", target: getTarget(contract.address), method, args, value, gas });
     // TX config has gas (from gasPrice function), value (how much eth to send) and an optional gasLimit. The reduce
-- 
2.30.2
`

src/clients/MultiCallerClient.ts
pxrl: Set at 100 in the latest update.

nicholaspai (approved):
> I think the entire class would be better supported by unit test if a lot of the on-the-fly sorting of transactions was handled directly in the enqueue() method. This would probably reduce the length of executeTransactionQueue() quite a bit and make it faster and easier to focus on the more core functionalities like chunking and RPC submission. I'm happy to use some time on that if we think it makes sense.

Let's create a follow up issue and implement this, I agree with @kevinuma that this client desperataely needs some unit tests since its so critical!

pxrl:
> > I think the entire class would be better supported by unit test if a lot of the on-the-fly sorting of transactions was handled directly in the enqueue() method. This would probably reduce the length of executeTransactionQueue() quite a bit and make it faster and easier to focus on the more core functionalities like chunking and RPC submission. I'm happy to use some time on that if we think it makes sense.
> 
> Let's create a follow up issue and implement this, I agree with @kevinuma that this client desperataely needs some unit tests since its so critical!

Have recorded the work to be done in ACX-156.

PR #246, opened 2022-08-12
improve(bridge adapters): Use the same event search config as SpokePoolClients

kevinuma:
Context: https://linear.app/uma/issue/ACX-14/monitor-should-track-l1-l2-bridged-funds-from-hubpool-to-spokepool

Currently the adapters define their own search range based on hardcoded block production rates of different chains. This is very brittle and makes it very hard to extend them to identify any stuck transfers. Next steps:

1. Update adapter logic to search for deposit events without matching finalization events instead of automatically throwing away all deposits older than the latest finalization event.
2. Add monitoring for stuck HubPool -> SpokePool transfers.

### DIFF
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -15,17 +15,10 @@ export class BaseAdapter {
   l2DepositFinalizedEvents: { [address: string]: { [l1Token: string]: any[] } } = {};
   l2DepositFinalizedEvents_DepositAdapter: { [address: string]: { [l1Token: string]: any[] } } = {};
 
-  depositEvalTime = 24 * 60 * 60; // Consider all deposit finalization events over the last 1 day.
-  l2LookBackSafetyMargin = 5; // The pending transfers util can accommodate up to this multiplier
-
-  constructor(
-    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
-    _chainId: number,
-    readonly l1FromBlock: number
-  ) {
+  constructor(readonly spokePoolClients: { [chainId: number]: SpokePoolClient }, _chainId: number) {
     this.chainId = _chainId;
-    this.l1SearchConfig = { ...this.getSearchConfig(1), fromBlock: l1FromBlock };
-    this.l2SearchConfig = { ...this.getSearchConfig(this.chainId), fromBlock: 0 };
+    this.l1SearchConfig = { ...this.getSearchConfig(1) };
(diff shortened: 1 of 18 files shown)


src/clients/bridges/BaseAdapter.ts
```diff
@@ -36,23 +29,16 @@ export class BaseAdapter {
...
-    const [l1BlockNumber, l2BlockNumber] = await Promise.all([
-      this.getProvider(1).getBlockNumber(),
-      this.getProvider(this.chainId).getBlockNumber(),
-    ]);
-
-    this.l1SearchConfig.toBlock = l1BlockNumber;
-    this.l2SearchConfig.toBlock = l2BlockNumber;
-
-    this.l2SearchConfig.fromBlock = Math.floor(
-      this.l2SearchConfig.toBlock - (this.depositEvalTime / this.avgBlockTime()) * this.l2LookBackSafetyMargin
```
nicholaspai, line 54: So removing this, what are the consequences?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -116,30 +103,6 @@ export class BaseAdapter {
           ].sort((a, b) => a.blockNumber - b.blockNumber);
         }
 
-        // Find the most recent L2 finalization event and most recent deposit event.
-        const newestFinalizedDeposit = l2FinalizationSet[l2FinalizationSet.length - 1];
```
nicholaspai, line 120: Happy to get rid of this convoluted logic!

src/clients/bridges/BaseAdapter.ts
```diff
@@ -198,13 +153,4 @@ export class BaseAdapter {
   isWeth(l1Token: string) {
     return l1Token.toLowerCase() === "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
   }
-
-  // Define some very rough heuristics for the average block time per chain.
-  avgBlockTime(chainId: number = this.chainId) {
-    if (chainId == 1) return 14; // 1 block every 13.5 seconds.
```
nicholaspai, line 204: Oh thank god for removing this

nicholaspai (approved):
Nice change!

src/clients/bridges/BaseAdapter.ts
kevinuma: Risk here is if the search range from SpokePoolClients doesn't work right, we might missing pending cross chain transfers and thus inventory management might send a cross chain transfer twice.

PR #247, opened 2022-08-12
feat: add incremental event caching class and update spokepookclient …

daywiss:
…to use

## Motivation
We are currently only caching event data if we have a successful run in the dataworker. This is problematic because the lack of events in cache can typically cause timeout/request errors that prevent a successful run in the first place. We want to incrementally save events as we get them, so that in the event of a failure we can resume with less probability of erroring out.

## Summary
This change adds a new paginated query function that uses a new `EventCache` class which can incrementally store events for each successfully paged event query. This requires a redis connection.  The main structural changes are to the spoke pool class, which now has its caching mechanism completely removed in favor of this new implementation, and the event data structures are slightly changed. Now the spoke pool must be able to use a deserialzied event object from redis, rather than previously using the raw event data from the blockchain.  Most changes to the spoke pool are to handle these cases.

## Testing
This is a very high risk change, but I tried to be thorough testing this by running the dataworker as if it was in production and validating the redis caching and retrieval. Still theres a possibility of issues if other parts of the system were relying on the spoke pools cache being filled, as this has been disabled.  Also if there are still parts of the spoke pool that require raw event data, those places will have to be changed to use the deserialzied event data instead.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import {
   spreadEvent,
   assign,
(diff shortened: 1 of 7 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -277,103 +305,61 @@ export class SpokePoolClient {
...
-          dataToCache.deposits,
-          [deposit.destinationChainId],
-          [{ ...deposit, blockNumber: dataForQuoteTime[index].quoteBlock, originBlockNumber: event.blockNumber }]
+        const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number }[] = await Promise.map(
+          depositEvents,
+          async (event: DepositWithBlock) => this.computeRealizedLpFeePct(event),
+          { concurrency: nodeMaxConcurrency }
         );
+        // start assigning things in order
+        depositEvents.forEach((event, index) => {
```
nicholaspai, line 341: so this logic basically reverts the logic back to what it was pre-caching logic that I implemented. From the client's view it just loads all events and saves them, right? The caching magic now happens behind the scenes

daywiss: thats right, its more generalized and uses the paged event query as its interface to do the caching and retrieval. the main thing to test is if retrieving and using events from cache breaks other things.

daywiss:
closing because this may be redudant at this point

PR #248, opened 2022-08-12
improve(relayer inventory management): Match deposits and finalizations by amount

kevinuma:
Details: https://linear.app/uma/issue/ACX-14/monitor-should-track-l1-l2-bridged-funds-from-hubpool-to-spokepool

### DIFF
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -105,36 +105,13 @@ export class BaseAdapter {
           ].sort((a, b) => a.blockNumber - b.blockNumber);
         }
 
-        // If, however, there is either a newestFinalizedDeposit or the most recent L1 deposit time is less than 1 day
-        // we can find how the current pending transactions! To do this we need to find the most recent L1 deposit event
-        // that overlaps with the most recent L2 finalization event. Not all bridges emit a deposit identifier or some
-        // unique way to map inputs to outputs. So the most general logic is to map the input and output amounts. Identify
-        // The associated L1 Deposit event index that maps to the newest finalization event. Note that if there is none
-        // this will return -1 i.e there has been no finalization event in the look back period but there is a deposit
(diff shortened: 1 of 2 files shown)


src/clients/bridges/BaseAdapter.ts
```diff
@@ -105,36 +105,13 @@ export class BaseAdapter {
...
-        // with the most recent event last. The event set contains all deposit events that have ever happened on L1 for
-        // the relayer. We know the L1 event index where which matches to the most recent L2 finalization event
-        // (associatedL1DepositIndex). If we take a slice of the array from this index to the end then we have only events
-        // that have occurred on L1 and have no matching event on L2 (and have happened in the last day due to the
-        // previous filter). These events are deposits that must have been made on L1 but not are not yet finalized on L2.
-        const l1EventsToConsider = this.l1DepositInitiatedEvents[monitoredAddress][l1Token].slice(
-          associatedL1DepositIndex + 1
+        // Match deposits and finalizations by amount. We're only doing a limited lookback of events so collisions
+        // should be unlikely.
+        const finalizedAmounts = new Set(l2FinalizationSet.map((finalization) => finalization.amount.toString()));
```
nicholaspai, line 110: Are there no other `finalization` props we can use to reduce the chance of collisions more? Also, what are the implications of collisions? Does that mean that the monitor will think a deposit is finalized when it might actually be stuck? When would we ever know about the stuck situation in this case?

kevinuma: Unfortunately no. If there's a collision, we would mistakenly think a past transfer to have already been finalized while a new transfer for the amount to not be finalized yet. So if we set a delay before sending alerts about stuck transfers, we'd miss it if the second deposit is relatively new. But this is fine imo, as we should have caught the first deposit  (as stuck) before the second deposit happened.

In summary, from the perspective of only using this for inventory management and monitoring stuck transfers, there should be no problems.

kevinuma:
Added tests

nicholaspai (approved):

PR #249, opened 2022-08-15
improve(monitoring): Add monitoring of stuck HubPool -> SpokePools cross chain transfers

kevinuma:
Originally reviewed in https://github.com/across-protocol/relayer-v2/pull/197

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -371,7 +371,12 @@ export class Monitor {
 
         if (transferBalance.gt(0)) {
           const mrkdwn = `Rebalances of ${l1Token.symbol} to ${getNetworkName(chainId)} is stuck`;
-          this.logger.warn({ at: "Monitor", message: "HubPool -> SpokePool rebalances stuck 🦴", mrkdwn });
+          this.logger.warn({
+            at: "Monitor",
+            message: "HubPool -> SpokePool rebalances stuck 🦴",
+            mrkdwn,
+            transferBalance: transferBalance.toString(),
+          });
         }
       }
     }
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -368,6 +368,14 @@ export class Monitor {
           chainId,
           l1Token.address
         );
+        const outstandingDepositTxs = etherscanLinks(
+          this.clients.crossChainTransferClient.getOutstandingCrossChainTransferTxs(
+            spokePoolAddress,
+            chainId,
+            l1Token.address
+          ),
+          1
+        );
 
         if (transferBalance.gt(0)) {
           const mrkdwn = `Rebalances of ${l1Token.symbol} to ${getNetworkName(chainId)} is stuck`;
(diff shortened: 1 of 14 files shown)


src/monitor/Monitor.ts
```diff
@@ -335,6 +336,47 @@ export class Monitor {
...
+    const hubPoolClient = this.clients.hubPoolClient;
+    const lastFullyExecutedBundle = hubPoolClient.getLatestFullyExecutedRootBundle(hubPoolClient.latestBlockNumber);
+    // This case shouldn't happen outside of tests as Across V2 has already launched.
+    if (lastFullyExecutedBundle === undefined) {
+      return;
+    }
+    // If we're still within the grace period, skip looking for any stuck rebalances.
+    // Again, this would give false negatives for transfers that have been stuck for longer than one bundle if the
+    // current time is within the grace period of last executed bundle. But this is a good trade off for simpler code.
+    const lastFullyExecutedBundleTime = lastFullyExecutedBundle.challengePeriodEndTimestamp;
```
nicholaspai, line 354: For transfers that  "have been stuck for longer than one bundle if the current time is within the grace period of last executed bundle", will they eventually be caught by this monitor once they are past the grace period?

src/monitor/Monitor.ts
```diff
@@ -335,6 +336,47 @@ export class Monitor {
...
+    for (const chainId of this.monitorConfig.spokePoolChains) {
+      const spokePoolAddress = this.clients.spokePoolClients[chainId].spokePool.address;
+      for (const l1Token of allL1Tokens) {
+        const transferBalance = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
+          spokePoolAddress,
+          chainId,
+          l1Token.address
+        );
+
+        if (transferBalance.gt(0)) {
```
nicholaspai, line 372: Would be kind of weird to see a `transferBalance == 0`, so why check this case?

test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
```diff
@@ -32,7 +31,7 @@ class TestAdapter extends BaseAdapter {
 }
 
 let adapter: TestAdapter;
-describe.only("AdapterManager: Send tokens cross-chain", async function () {
```
nicholaspai, line 35: Wow that's bad this got committed I thought we linted these tests [here](https://github.com/across-protocol/relayer-v2/blob/master/.eslintrc.js#L30), unless this was included a while ago?

nicholaspai (approved):
This looks great, much easier to review since you broke this feature into multiple PR's!

test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
kevinuma: This wasn't committed to main but yeah, I forgot to remove .only from a pending PR's commit that this is diff based off and that didn't fail lint for some reason.

src/monitor/Monitor.ts
kevinuma: This is so that we don't emit anything if there's nothing stuck (outstanding balance = 0)

src/monitor/Monitor.ts
kevinuma: Yes once they're past the grace period.

nicholaspai (approved):
LGTM

PR #250, opened 2022-08-15
fix: Correct broken CoinGecko commit

pxrl:


src/clients/ProfitClient.ts
```diff
@@ -123,13 +123,15 @@ export class ProfitClient {
 
     const errors: Array<{ [k: string]: string }> = [];
     for (const address of Object.keys(l1Tokens)) {
-      const tokenPrice: CoinGeckoPrice = cgPrices.find((price) => address.toLowerCase() === price.address);
+      const tokenPrice: CoinGeckoPrice = cgPrices.find(
+        (price) => address.toLowerCase() === price.address.toLowerCase()
```
chrismaree, line 127: nit: never use `toLowerCase`. use `toChecksumAddress` from ethers.

chrismaree (approved):

src/clients/ProfitClient.ts
pxrl: This one went in before I realised there was a comment, but there's a follow-up change to the CoinGecko class coming so I will bundle in this comment there.

PR #251, opened 2022-08-15
feat: Enable BAL in relayer

nicholaspai:
Adds hardcoded values to support BAL in:
- clients/bridges/*
- constants.ts


kevinuma (approved):
Some of the addresses are not in checksummed format. Can you change them to be consistent?

kevinuma (approved):

mrice32 (approved):
LGTM! Addresses all checked out

PR #252, opened 2022-08-15
fix: Incorrect log message format + reduce noise

pxrl:
The current construction can produce a *lot* of warning logs when things
aren't going so well, and this can in turn generate a lot of monitoring
events.

Also correct the following:
 - A log message was incorrectly attributed to the TokenClient.
 - A log message was incorrectly formatted and caused downstream
   reporting warnings (missing at: keyword).

Ref: ACX-125

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -96,7 +96,7 @@ export class ProfitClient {
   }
 
   captureUnprofitableFill(deposit: Deposit, fillAmount: BigNumber) {
-    this.logger.debug({ at: "TokenClient", message: "Handling unprofitable fill", deposit, fillAmount });
+    this.logger.debug({ at: "ProfitClient", message: "Handling unprofitable fill", deposit, fillAmount });
     assign(this.unprofitableFills, [deposit.originChainId], [{ deposit, fillAmount }]);
   }
 
@@ -114,11 +114,7 @@ export class ProfitClient {
     try {
       cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
     } catch (err) {
-      this.logger.warn({
-        message: "Failed to retrieve prices.",
-        err,
-        l1Tokens,
-      });
+      this.logger.warn({ at: "ProfitClient", message: "Failed to retrieve prices.", err, l1Tokens });
     }
 
     const errors: Array<{ [k: string]: string }> = [];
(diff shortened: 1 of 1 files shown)


src/clients/ProfitClient.ts
```diff
@@ -129,17 +125,18 @@ export class ProfitClient {
...
       // todo: Any additional validation to do? Ensure that timestamps are always moving forwards?
       if (tokenPrice === undefined || typeof tokenPrice.price !== "number") {
-        const symbol = l1Tokens[address].symbol;
-        const errmsg = tokenPrice ? "Unexpected price response." : `Missing price for ${symbol}.`;
-        this.logger.warn({ at: "ProfitClient", message: errmsg, tokenPrice });
-        errors.push({ address: address, symbol: symbol });
+        errors.push({
+          address: address,
+          symbol: l1Tokens[address].symbol,
+          cause: tokenPrice ? "Unexpected price response" : "Missing price",
```
nicholaspai, line 131: If tokenPrice is 0 even though it's not undefined it will log "Unexpected price", is that expected?

nicholaspai: If not, then we should change the condition to tokenPrice === undefined

nicholaspai (approved):
LGTM. Just one question

src/clients/ProfitClient.ts
```diff
@@ -129,17 +125,18 @@ export class ProfitClient {
 
       // todo: Any additional validation to do? Ensure that timestamps are always moving forwards?
       if (tokenPrice === undefined || typeof tokenPrice.price !== "number") {
-        const symbol = l1Tokens[address].symbol;
-        const errmsg = tokenPrice ? "Unexpected price response." : `Missing price for ${symbol}.`;
-        this.logger.warn({ at: "ProfitClient", message: errmsg, tokenPrice });
```
pxrl, line 134: FWIW this line was causing 2x the PD events to be generated, because we also warn about the same up at line 51 (later when the client requests a specific price). There's no reason to generate a warning per token here, so I bundled the cause into the errors array so it can be included in the markdown instead.

src/clients/ProfitClient.ts
pxrl: Good catch; that was unintended. I'll make the check more explicit.

pxrl: @nicholaspai Actually tokenPrice is of type CoinGeckoPrice - it's an object with the following definition: 

```
export type CoinGeckoPrice = {
  address: string;
  timestamp: number;
  price: number;
};
```

If we receive a 0 then something has gone completely wrong - i.e. it's not an actual price and we should drop it.

nicholaspai: ah you're right

PR #253, opened 2022-08-16
improve(relayer): Filter all deposits with relayer fee % < 0.03%

kevinuma:
This is a temporary solution until we fully implement a capital fee requirement to be consistent with the API.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -37,7 +37,8 @@ export class ProfitClient {
   constructor(
     readonly logger: winston.Logger,
     readonly hubPoolClient: HubPoolClient,
-    readonly relayerDiscount: BigNumber = toBNWei(0)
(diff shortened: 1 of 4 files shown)


nicholaspai (approved):

src/clients/ProfitClient.ts
```diff
@@ -63,6 +64,11 @@ export class ProfitClient {
   }
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
+    if (toBN(deposit.relayerFeePct).lt(toBNWei(this.minRelayerFeePct))) {
+      this.logger.debug({ at: "ProfitClient", message: "Relayer fee % < 0.03%" });
```
nicholaspai, line 68: You should change this to say
```suggestion
      this.logger.debug({ at: "ProfitClient", message: "Relayer fee % < minimum relayer fee %",  minRelayerFeePct: this. minRelayerFeePct });
```

PR #255, opened 2022-08-16
improve: Use Across' internal Coingecko class

pxrl:
This one implements a singleton pattern so it required a change in how
it is instantiated.

Ref: ACX-125

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
-    "@across-protocol/sdk-v2": "^0.1.10",
+    "@across-protocol/sdk-v2": "^0.1.19",
     "@arbitrum/sdk": "^2.0.9",
     "@defi-wonderland/smock": "^2.0.7",
     "@eth-optimism/sdk": "^1.1.5",
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,14 +1,7 @@
+import { coingecko } from "@across-protocol/sdk-v2";
 import { BigNumber, winston, toBNWei, toBN, assign } from "../utils";
 import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
-import { Coingecko } from "@uma/sdk";
-
-// Copied from @uma/sdk/coingecko. Propose to export export it upstream in the sdk.
-type CoinGeckoPrice = {
-  address: string;
-  timestamp: number;
-  price: number;
-};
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
-    "@across-protocol/sdk-v2": "^0.1.10",
+    "@across-protocol/sdk-v2": "^0.1.19",
```
pxrl, line 14: Placeholder for the real version number once it's bumped.

src/clients/ProfitClient.ts
```diff
@@ -1,14 +1,7 @@
+import { coingecko } from "@across-protocol/sdk-v2";
```
pxrl, line 1: Got irritated here...intuitively I want to write:

```
import { Coingecko, CoinGeckoPrice } from "@across-protocol/sdk-v2/coingecko";
```

But I can't figure out what I need to do to make that work. Haven't been successful in finding many descriptions on how to import from a specific module within a package. Probably did something wrong over in https://github.com/across-protocol/sdk-v2/pull/78.

src/clients/ProfitClient.ts
```diff
@@ -1,14 +1,7 @@
...
 import { BigNumber, winston, toBNWei, toBN, assign } from "../utils";
 import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
-import { Coingecko } from "@uma/sdk";
```
pxrl, line 4: Is there any standard for the ordering of import statements? My preference is always to list external things at the top (as I've done here), but I've also previously had discussions with people who preferred it to be the other way. In any case, our own style guide doesn't seem to have an opinion on it.

src/clients/ProfitClient.ts
```diff
@@ -40,7 +33,7 @@ export class ProfitClient {
     readonly relayerDiscount: BigNumber = toBNWei(0),
     readonly minRelayerFeePct: string = "0"
   ) {
-    this.coingecko = new Coingecko();
+    this.coingecko = coingecko.Coingecko.get(logger);
```
pxrl, line 36: One thing to note is that the sdk-v2 Coingecko instance has 3 inbuilt retries on a 1-second interval. It will be interesting to see whether this makes the problems better or worse...

pxrl:
This PR isn't really applicable anymore. We weren't ready for it last time because of other changes that were made to the Coingecko class in sdk-v2. It will be integrated before too long, though.

PR #256, opened 2022-08-16
improve(logging): Add % when displaying relayerFeePct as %

kevinuma:


nicholaspai (approved):
This is a good refactor. It highlights however the inconsistency in how we denominate RELAYER_FEE_PCT and MIN_RELAYER_FEE_PCT as %'s while we denominate the transfer/trigger thresholds as 100x %'s. So "3" as a transfer threshold pct = 0.03, while 3 as a RELAYER_FEE_PCT is 300%

PR #257, opened 2022-08-16
improve(ProfitClient): Followup & logging improvements

pxrl:
The work to consolidate the CoinGecko price requests unfortunately      
hasn't eliminated rate-limiting, so we're still seeing quite some noise 
in log monitoring. A single instance of rate-limiting generates 7 - 9   
events (i.e. one per warning message), so cut down on those by reducing 
duplication in the log messages.                                        
                                                                        
Additionally, address some feedback that didn't make it into the         
previous review, and modify the test case so that it supplies various
address formats as input to the Profit Client.

Ref: ACX-125


src/clients/ProfitClient.ts
```diff
@@ -130,7 +142,8 @@ export class ProfitClient {
         (price) => address.toLowerCase() === price.address.toLowerCase()
```
pxrl, line 142: @chrismaree You had previously commented that ethers' getAddress() should be used instead of toLowerCase().

I tested with this, but getAddress() has side-effects in that it can throw an exception if the address is not already in an expected format (i.e. it doesn't like upper-case addresses). Using getAddress() would therefore force ProfitClient to be picky about the format of both addresses that it's handling in this comparison, so I've left it as-is to avoid that scenario and permit it to be tolerant of a variety of inputs.

Using compareAddresses() from utils.ts is another possibility, but my guess is that it's probably faster to perform a string comparison than a numeric comparison because the numbers are so huge.

```
// Throws if a checksummed address is provided, but a
// letter is the wrong case
// ------------v (should be lower-case)
getAddress("0x8Ba1f109551bD432803012645Ac136ddd64DBA72")
// [Error: bad address checksum] {
//   argument: 'address',
//   code: 'INVALID_ARGUMENT',
//   reason: 'bad address checksum',
//   value: '0x8Ba1f109551bD432803012645Ac136ddd64DBA72'
// }
```
https://docs.ethers.io/v5/api/utils/address/#utils-getAddress

Happy to consider it further if you feel strongly about it.

src/clients/ProfitClient.ts
```diff
@@ -49,7 +49,13 @@ export class ProfitClient {
 
   getPriceOfToken(token: string) {
     if (!this.tokenPrices[token]) {
-      this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
+      // A warning is already emitted for token price retrieval failures, so only warn if we are missing specific prices.
+      const errmsg = `Token ${token} not found in state. Using 0.`;
+      if (Object.keys(this.tokenPrices).length === 0) {
+        this.logger.info({ at: "ProfitClient", message: errmsg });
+      } else {
+        this.logger.warn({ at: "ProfitClient", message: errmsg });
```
kevinuma, line 57: How often does this case happen compared to when all prices are missing? I'm under the impression that it's relatively rare. Most of the warnings I've seen in the past couple of days are for all prices being missing.

src/clients/ProfitClient.ts
```diff
@@ -121,7 +127,13 @@ export class ProfitClient {
     try {
       cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
     } catch (err) {
-      this.logger.warn({ at: "ProfitClient", message: "Failed to retrieve prices.", err, l1Tokens });
+      const addresses: Array<string> = Object.keys(l1Tokens);
```
kevinuma, line 130: nit: : Array\<string\> doesn't seem necessary?

kevinuma (approved):

src/clients/ProfitClient.ts
pxrl: I have also only seen the former case (no prices retrieved due to rate-limiting), but we're running up against a third-party service so it's hard to rule out what they might do from time to time. In practice I expect that we'll almost always be logging at level info, though.

src/clients/ProfitClient.ts
pxrl: Hm, what's the norm on this? The linter didn't suggest to remove it, which is something it has done in some other cases where the type is 100% inferrable from the initial assignment.

nicholaspai: It's implied by Object.keys so I agree with removing the type. I think this is 100% inferrable from the assignment

src/clients/ProfitClient.ts
```diff
@@ -121,7 +127,13 @@ export class ProfitClient {
...
+      this.logger.warn({
+        at: "ProfitClient",
+        message: `Failed to retrieve token prices (${err}).`,
```
nicholaspai, line 133: I would just make `err` a key in the log like so
```suggestion
        message: `Failed to retrieve token prices.`,
        err,
```

src/clients/ProfitClient.ts
nicholaspai: I'm good with this

nicholaspai (approved):
Once nits are resolved LGTM

src/clients/ProfitClient.ts
pxrl: FWIW I've just removed this logging entirely in the latest update. Now I propose that we only log on failure to retrieve the prices. This should really cut down on noise.

src/clients/ProfitClient.ts
pxrl: I adjusted it so that the list of addresses is produced directly in the logging message. This seems cleaner to me anyway.

src/clients/ProfitClient.ts
pxrl: This is what I had originally, but the contents of err never gets included in the log message - it only prints "err". I suspect that's because it's a string and winston just skips it - does that sound plausible?

pxrl: > This is what I had originally, but the contents of err never gets included in the log message - it only prints "err". I suspect that's because it's a string and winston just skips it - does that sound plausible?

So it's not "just" a string..it seems a bit maleable, depending on where it's used. Including it in the log directly as "error: err" blows up and dumps the corresponding stack trace _after_ the end of the log:

```
2022-08-17 19:52:19 [warn]: {                                                                                                                                                                             [285/1799]
  "at": "ProfitClient",                                                                                                                                                                                             
  "message": "Failed to retrieve token prices.",                                                                                                                                                                    
  "tokens": [                                                                                                                                                                                                       
    {                                                                                                                                                                                                               
      "address": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",                                                                                                                                                      
      "symbol": "WETH",                                                                                                                                                                                             
      "decimals": 18                                                                                                                                                                                                
    },                                                                                                                                                                                                              
    {                                                                                                                                                                                                               
      "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",                                                                                                                                                      
      "symbol": "USDC",                                                                                                                                                                                             
      "decimals": 6                                                                                                                                                                                                 
    },                                                                                                                                                                                                              
    {                                                                                                                                                                                                               
      "address": "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",                                            
      "symbol": "WBTC",                       
      "decimals": 8                                                                         
    },                                        
    {                                                                                       
      "address": "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828",                                            
      "symbol": "UMA",                        
      "decimals": 18                                                                        
    },                                                                                                    
    {                                                                                       
      "address": "0x6B175474E89094C44Da98b954EedeAC495271d0F",                                            
      "symbol": "DAI",                                                                                    
      "decimals": 18                          
    },                                                                                      
    {                                                                                       
      "address": "0x42bBFa2e77757C645eeaAd1655E0911a7553Efbc",                                            
      "symbol": "BOBA",                                                                     
      "decimals": 18                          
    },                                                                                      
    {                                                                                       
      "address": "0x3472A5A71965499acd81997a54BBA8D852C6E53d",                              
      "symbol": "BADGER",                                                                   
      "decimals": 18                                                                                      
    },                                        
    {                                                                                       
      "address": "0xba100000625a3754423978a60c9317c58a424e3D",                                            
      "symbol": "BAL",                                                                      
      "decimals": 18                                                                                      
    }                                                                                       
  ],                                          
  "bot-identifier": "XXXX"                                                              
}                                                                                                         
Error: Unknown Coingecko Error                
    at Coingecko.call (/home/pxrl/dev/across-protocol/relayer-v2/node_modules/@uma/sdk/dist/src/coingecko/coingecko.ts:97:13)                                                                                       
    at runMicrotasks (<anonymous>)                                                          
    at processTicksAndRejections (node:internal/process/task_queues:96:5)                                 
    at async Coingecko.getContractPrices (/home/pxrl/dev/across-protocol/relayer-v2/node_modules/@uma/sdk/dist/src/coingecko/coingecko.ts:75:28)                                                                    
    at async ProfitClient.coingeckoPrices (/home/pxrl/dev/across-protocol/relayer-v2/src/clients/ProfitClient.ts:174:12)                                                                                            
    at async ProfitClient.update (/home/pxrl/dev/across-protocol/relayer-v2/src/clients/ProfitClient.ts:134:18)                                                                                                     
    at async updateClients (/home/pxrl/dev/across-protocol/relayer-v2/src/common/ClientHelper.ts:176:3)   
    at async updateRelayerClients (/home/pxrl/dev/across-protocol/relayer-v2/src/relayer/RelayerClientHelper.ts:64:3)                                                                                               
    at async Object.runRelayer [as relayer] (/home/pxrl/dev/across-protocol/relayer-v2/src/relayer/index.ts:29:7)                                                                                                   
    at async run **(/home/pxrl/dev/across-protocol/relayer-v2/index.ts:38:9)**
```

src/clients/ProfitClient.ts
```diff
@@ -147,7 +165,7 @@ export class ProfitClient {
     if (errors.length > 0) {
       let mrkdwn = "The following L1 token prices could not be fetched:\n";
       errors.forEach((token: { [k: string]: string }) => {
-        mrkdwn += `- ${token["symbol"]} not found (${token["cause"]}).`;
+        mrkdwn += `- ${token["symbol"]} not found (${token["cause"]}).\n`;
```
pxrl, line 168: Ergh...this shouldn't be here.

src/clients/ProfitClient.ts
mrice32: @pxrl That looks more like some print that's happening _outside of_ your log since it's outside of the json.

There's a lot of formatting that's happening [behind the scenes](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/logger) with the logger. One tip that may magically solve this is if you rename err in your object to error (there's some special handling in the logger for the field error).

Basically just this:
```ts
this.logger.warn({
...
  error: err,
...
});
```

src/clients/ProfitClient.ts
```diff
@@ -116,16 +113,36 @@ export class ProfitClient {
...
 
   async update() {
+    // Generate list of tokens to retrieve and pre-populate any new addresses
+    const newTokens: Array<{ [k: string]: string }> = [];
     const l1Tokens: { [k: string]: L1Token } = Object.fromEntries(
-      this.hubPoolClient.getL1Tokens().map((token) => [token["address"], token])
+      this.hubPoolClient.getL1Tokens().map((token) => {
+        const { address, symbol } = token;
+
+        if (typeof this.tokenPrices[address] === "undefined") {
```
mrice32, line 122: nit: why typeof? Why not just compare the value to undefined?

```ts
if (this.tokenPrices[address] === undefined)
```

src/clients/ProfitClient.ts
```diff
@@ -116,16 +113,36 @@ export class ProfitClient {
   }
 
   async update() {
+    // Generate list of tokens to retrieve and pre-populate any new addresses
+    const newTokens: Array<{ [k: string]: string }> = [];
```
mrice32, line 117: nit: it's good practice to make types as specific as possible:

```ts
const newTokens: Array<{ address: string; symbol: string }> = [];
```

This one is just stylistic and won't affect anything (Array and [] do the same thing), but it's a bit more concise. Totally up to you, though.
```ts
const newTokens: { address: string; symbol: string }[] = [];

File: src/clients/ProfitClient.ts [hunk 8]
```diff
@@ -48,10 +48,7 @@ export class ProfitClient {
   }
 
   getPriceOfToken(token: string) {
-    if (!this.tokenPrices[token]) {
-      this.logger.warn({ at: "ProfitClient", message: `Token ${token} not found in state. Using 0` });
-      return toBN(0);
-    }
+    assert(typeof this.tokenPrices[token] !== "undefined", `Token ${token} not in price list.`);
```
mrice32, line 51: Same comment as elsewhere, typeof isn't really an idiomatic way to check for undefined. More concise and simpler to just check against `undefined`

test/fixtures/Dataworker.Fixture.ts
```diff
@@ -262,6 +262,7 @@ export async function setupDataworker(
     updateAllClients: async () => {
       await hubPoolClient.update();
       await configStoreClient.update();
+      await profitClient.update();
```
mrice32, line 265: Kinda surprised this didn't cause any issues before.

mrice32 (commented):

src/clients/ProfitClient.ts
nicholaspai: Good catch matt

test/fixtures/Dataworker.Fixture.ts
nicholaspai: I think because we don't really use the profit client in the dataworker

src/clients/ProfitClient.ts
pxrl: Resolved in DM - this is a known quirk in the logging implementation. Will proceed with embedding err directly in the message - this behaves fine in my testing.

https://github.com/UMAprotocol/protocol/blob/8c47a3bccb302d46ad22ee31ccc93090dcc79164/packages/financial-templates-lib/src/logger/ConsoleTransport.ts#L20-L25

```
2022-08-19 20:19:14 [warn]: {                        
  "at": "ProfitClient",                              
  "message": "Failed to retrieve token prices (Error: Unknown Coingecko Error).",                         
  "tokens": [                                        
    "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",                                                         
    "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",                                                         
    "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",                                                         
    "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828",                                                         
    "0x6B175474E89094C44Da98b954EedeAC495271d0F",                                                         
    "0x42bBFa2e77757C645eeaAd1655E0911a7553Efbc",                                                         
    "0x3472A5A71965499acd81997a54BBA8D852C6E53d",                                                         
    "0xba100000625a3754423978a60c9317c58a424e3D"                                                          
  ],                                                 
  "bot-identifier": "XXXX"                       
}
```

src/clients/ProfitClient.ts
pxrl: Understood - happy to keep things more concise, and have updated the type specifiers where I have seen them.

src/clients/ProfitClient.ts
pxrl: Surprised that the use of typeof isn't idiomatic - but can't argue against it! Now I know what to do in future.

test/fixtures/Dataworker.Fixture.ts
pxrl: The pre-existing behaviour of the ProfitClient just dishes out 0 for any price that it didn't have knowledge of in the past. Per this change, we assert on any unknown tokens, since it's (IMO...) an error to request a price for a token that never got loaded. Therefore, without this change, the test suddenly fails.

I will concede that there is perhaps a counter-argument that we are loading tokens from the hubPoolClient, not from the caller directly, so perhaps it isn't so unreasonable. Happy to take a discussion on it if needed.

src/clients/ProfitClient.ts
pxrl: Typeof just seemed more intuitive to me, but if checking directly for undefined is better then I'll just do that from now on. Fixed in the latest update!

pxrl:
Should be OK to go now. I plan to submit it drop it in on Monday.

test/fixtures/Dataworker.Fixture.ts
```diff
@@ -175,7 +175,7 @@ export async function setupDataworker(
     [repaymentChainId]: spokePoolClient_3,
     1: spokePoolClient_4,
   };
-  const profitClient = new clients.ProfitClient(spyLogger, hubPoolClient, spokePoolClients, false, []);
+  const profitClient = new clients.ProfitClient(spyLogger, hubPoolClient, spokePoolClients, false, [], true);
```
pxrl, line 178: @kevinuma I expect you'll be fine with this change (ignoreTokenPriceFailures => true in test). This test seems to lookup prices for test tokens that CoinGecko doesn't know about, so price failures are inevitable.

pxrl: OK after other recent changes - everything runs with ignoreTokenPriceFailures now anyway.

PR #260, opened 2022-08-17
fix: Lint on SpokePoolClient SpeedUp test

pxrl:
Error introduced on previous PR.


nicholaspai (approved):

pxrl:
Looks like this is already handled by Nick's PR https://github.com/across-protocol/relayer-v2/pull/259, but I'll park it here anyway so it doesn't block anything else.

PR #261, opened 2022-08-17
feat(relayer): Consider gas cost when calculating profitability

kevinuma:
Originally reviewed in https://github.com/across-protocol/relayer-v2/pull/212

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
-    "@across-protocol/sdk-v2": "^0.1.10",
+    "@across-protocol/sdk-v2": "^0.1.18",
     "@arbitrum/sdk": "^2.0.9",
     "@defi-wonderland/smock": "^2.0.7",
     "@eth-optimism/sdk": "^1.1.5",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
-    "@across-protocol/sdk-v2": "^0.1.18",
+    "@across-protocol/sdk-v2": "^0.1.21",
     "@arbitrum/sdk": "^2.0.9",
     "@defi-wonderland/smock": "^2.0.7",
     "@eth-optimism/sdk": "^1.1.5",
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -25,6 +25,14 @@ const GAS_TOKEN_BY_CHAIN_ID = {
 // TODO: Make this dynamic once we support chains with gas tokens that have different decimals.
 const GAS_TOKEN_DECIMALS = 18;
 
+const QUERY_HANDLERS = {
+  1: relayFeeCalculator.EthereumQueries,
+  10: relayFeeCalculator.OptimismQueries,
+  137: relayFeeCalculator.PolygonQueries,
+  288: relayFeeCalculator.BobaQueries,
+  42161: relayFeeCalculator.ArbitrumQueries,
+};
+
 export class ProfitClient {
   private readonly coingecko;
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
@@ -242,64 +250,15 @@ export class ProfitClient {
     const coingeckoProApiKey = undefined;
     // TODO: Set this once we figure out gas markup on the API side.
     const gasMarkup = 0;
-    switch (chainId) {
-      case 1:
-        return new relayFeeCalculator.EthereumQueries(
-          provider,
-          undefined,
-          undefined,
-          undefined,
-          undefined,
-          coingeckoProApiKey,
-          this.logger,
-          gasMarkup
-        );
-      case 10:
-        return new relayFeeCalculator.OptimismQueries(
-          provider,
-          undefined,
-          undefined,
-          undefined,
-          undefined,
-          coingeckoProApiKey,
-          this.logger,
-          gasMarkup
-        );
-      case 137:
-        return new relayFeeCalculator.PolygonQueries(
-          provider,
-          undefined,
-          undefined,
-          undefined,
-          undefined,
-          coingeckoProApiKey,
-          this.logger,
-          gasMarkup
-        );
-      case 288:
-        return new relayFeeCalculator.BobaQueries(
-          provider,
-          undefined,
-          undefined,
-          undefined,
-          undefined,
-          coingeckoProApiKey,
-          this.logger,
-          gasMarkup
-        );
-      case 42161:
-        return new relayFeeCalculator.ArbitrumQueries(
-          provider,
-          undefined,
-          undefined,
-          undefined,
-          undefined,
-          coingeckoProApiKey,
-          this.logger,
-          gasMarkup
-        );
-      default:
-        throw new Error(`Unexpected chain ${chainId}`);
-    }
+    return new QUERY_HANDLERS[chainId](
+      provider,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      coingeckoProApiKey,
+      this.logger,
+      gasMarkup
+    );
   }
 }
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -21,6 +21,10 @@ const GAS_TOKEN_BY_CHAIN_ID = {
   137: WMATIC,
   288: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
   42161: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+
+  // For tests.
+  666: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+  1337: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
 };
 // TODO: Make this dynamic once we support chains with gas tokens that have different decimals.
 const GAS_TOKEN_DECIMALS = 18;
@@ -106,12 +110,13 @@ export class ProfitClient {
     // Consider gas cost.
     const totalGasCostWei = this.totalGasCosts[deposit.destinationChainId];
     if (!totalGasCostWei) {
-      const errorMsg = "Missing total gas cost. This likely indicate some gas cost requests to provider failed";
+      const chainId = deposit.destinationChainId;
+      const errorMsg = `Missing total gas cost for ${chainId}. This likely indicate some gas cost request failed`;
       this.logger.error({
         at: "ProfitClient",
         message: errorMsg,
         allGasCostsFetched: this.totalGasCosts,
-        chainId: deposit.destinationChainId,
+        chainId,
       });
       throw new Error(errorMsg);
     }
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -161,21 +161,21 @@ export class ProfitClient {
     );
 
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    const cgPrices: CoinGeckoPrice[] = [];
+    let cgPrices: CoinGeckoPrice[] = [];
     try {
       const [maticTokenPrice, otherTokenPrices] = await Promise.all([
         // Add WMATIC for gas cost calculations.
         this.coingeckoPrices([WMATIC], "polygon-pos"),
         this.coingeckoPrices(Object.keys(l1Tokens)),
       ]);
-      cgPrices.concat(maticTokenPrice);
-      cgPrices.concat(otherTokenPrices);
+      cgPrices = cgPrices.concat(maticTokenPrice);
+      cgPrices = cgPrices.concat(otherTokenPrices);
     } catch (err) {
       this.logger.warn({ at: "ProfitClient", message: "Failed to retrieve prices.", err, l1Tokens });
     }
 
     // Add to l1Tokens after the fetches, so prices and l1Tokens have the same entries, for any error logging later.
-    l1Tokens["WMATIC"] = {
+    l1Tokens[WMATIC] = {
       address: WMATIC,
       symbol: "WMATIC",
       decimals: 18,
@@ -188,8 +188,7 @@ export class ProfitClient {
       );
 
       // TODO: Any additional validation to do? Ensure that timestamps are always moving forwards?
-      const priceFetchSuccessful = tokenPrice !== undefined && typeof tokenPrice.price === "number";
-      if (priceFetchSuccessful || this.ignoreTokenPriceFailures) {
+      if (tokenPrice !== undefined && typeof tokenPrice.price === "number") {
         this.tokenPrices[address] = toBNWei(tokenPrice.price);
       } else {
         errors.push({
@@ -206,7 +205,12 @@ export class ProfitClient {
         mrkdwn += `- ${token["symbol"]} not found (${token["cause"]}).`;
         mrkdwn += ` Using last known price of ${this.getPriceOfToken(token["address"])}.\n`;
       });
-      this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+      if (this.ignoreTokenPriceFailures) {
+        this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+      } else {
+        this.logger.error({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+        throw new Error(mrkdwn);
+      }
     }
 
     // Pre-fetch total gas costs for relays on enabled chains.
@@ -229,7 +233,7 @@ export class ProfitClient {
     this.logger.debug({ at: "ProfitClient", message: "Updated Profit client", tokenPrices: this.tokenPrices });
   }
 
-  private async coingeckoPrices(tokens: Array<string>, platformId?: string) {
+  protected async coingeckoPrices(tokens: string[], platformId?: string) {
     return await this.coingecko.getContractPrices(tokens, "usd", platformId);
   }
 
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -2,6 +2,7 @@ import { BigNumber, formatFeePct, winston, toBNWei, toBN, assign } from "../util
 import { HubPoolClient } from ".";
 import { Deposit, L1Token } from "../interfaces";
 import { Coingecko } from "@uma/sdk";
+import { relayFeeCalculator } from "@across-protocol/sdk-v2";
 
 // Copied from @uma/sdk/coingecko. Propose to export export it upstream in the sdk.
 type CoinGeckoPrice = {
@@ -10,37 +11,43 @@ type CoinGeckoPrice = {
   price: number;
 };
 
-// Define the minimum revenue, in USD, that a relay must yield in order to be considered "profitable". This is a short
-// term solution to enable us to avoid DOS relays that yield negative profits. In the future this should be updated
-// to actual factor in the cost of sending transactions on the associated target chains.
-const chainIdToMinRevenue = {
-  // Mainnet and L1 testnets.
-  1: toBNWei(10),
-  4: toBNWei(10),
-  5: toBNWei(10),
-  42: toBNWei(10),
-  // Rollups/L2s/sidechains & their testnets.
-  10: toBNWei(1),
-  69: toBNWei(1),
-  288: toBNWei(1),
-  28: toBNWei(1),
-  42161: toBNWei(1),
-  137: toBNWei(1),
-  80001: toBNWei(1),
+// We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
+export const WMATIC = "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270";
+
+const GAS_TOKEN_BY_CHAIN_ID = {
+  1: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  10: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  137: WMATIC,
+  288: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  42161: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
 };
+// TODO: Make this dynamic once we support chains with gas tokens that have different decimals.
+const GAS_TOKEN_DECIMALS = 18;
 
 export class ProfitClient {
   private readonly coingecko;
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
   private unprofitableFills: { [chainId: number]: { deposit: Deposit; fillAmount: BigNumber }[] } = {};
 
+  // Track total gas costs of a relay on each chain.
+  private totalGasCosts: { [chainId: number]: BigNumber } = {};
+
+  private relayerFeeQueries: { [chainId: number]: relayFeeCalculator.QueryInterface } = {};
+
   constructor(
     readonly logger: winston.Logger,
     readonly hubPoolClient: HubPoolClient,
-    readonly relayerDiscount: BigNumber = toBN(0),
+    readonly enableProfitability: boolean,
+    readonly enabledChainIds: number[],
+    // Default to throwing errors if fetching token prices fails.
+    readonly ignoreTokenPriceFailures: boolean = false,
     readonly minRelayerFeePct: BigNumber = toBN(0)
   ) {
     this.coingecko = new Coingecko();
+
+    for (const chainId of enabledChainIds) {
+      this.relayerFeeQueries[chainId] = this.constructRelayerFeeQuery(chainId);
+    }
   }
 
   getAllPrices() {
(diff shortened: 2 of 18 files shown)


src/clients/ProfitClient.ts
```diff
@@ -73,24 +86,42 @@ export class ProfitClient {
...
+    const gasCostInUsd = totalGasCostWei
+      .mul(this.getPriceOfToken(GAS_TOKEN_BY_CHAIN_ID[deposit.destinationChainId]))
+      .div(toBN(10).pow(GAS_TOKEN_DECIMALS));
+
     // How much minimumAcceptableRevenue is scaled. If relayer discount is 0 then need minimumAcceptableRevenue at min.
-    const revenueScalar = toBNWei(1).sub(this.relayerDiscount);
-    const minimumAcceptableRevenue = chainIdToMinRevenue[deposit.destinationChainId].mul(revenueScalar).div(toBNWei(1));
-    const fillProfitable = fillRevenueInUsd.gte(minimumAcceptableRevenue);
+    const fillProfitInUsd = fillRevenueInUsd.sub(gasCostInUsd);
+    const fillProfitable = fillProfitInUsd.gte(toBN(0));
```
nicholaspai, line 124: Is the gas cost buffer logic in the sdk being included in this calculation somehow?

src/clients/ProfitClient.ts
```diff
@@ -152,10 +203,95 @@ export class ProfitClient {
       });
       this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
     }
+
+    // Pre-fetch total gas costs for relays on enabled chains.
+    const getGasCosts = [];
+    for (const chainId of this.enabledChainIds) {
+      getGasCosts.push(this.relayerFeeQueries[chainId].getGasCosts());
```
nicholaspai, line 210: Ah so this is where the sdk gas buffer would come in

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -1,20 +1,42 @@
...
+import {AcrossConfigStoreClient, HubPoolClient, MultiCallerClient, SpokePoolClient} from "../src/clients";
 
 let hubPoolClient: MockHubPoolClient, spyLogger: winston.Logger, profitClient: MockProfitClient;
 
 const mainnetWeth = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
 const mainnetUsdc = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
 
-describe("ProfitClient: Consider relay profit", async function () {
-  beforeEach(() => {
+describe.only("ProfitClient: Consider relay profit", async function () {
```
nicholaspai, line 20: Leftover .only. Eslint should have caught this

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -1,20 +1,42 @@
...
+      spyLogger,
+      spokePool_2.connect(owner),
+      null,
+      destinationChainId
+    );
+    const spokePoolClients = { [originChainId]: spokePoolClient_1, [destinationChainId]: spokePoolClient_2 };
+    profitClient = new MockProfitClient(spyLogger, hubPoolClient, spokePoolClients, true, [], false, toBN(0));
+    profitClient.setTokenPrices({ [mainnetWeth]: toBNWei(3000), [mainnetUsdc]: toBNWei(1) });
+    profitClient.setGasCosts({ [mainnetWeth]: toBNWei(1), [mainnetUsdc]: toBNWei(1) });
   });
```
nicholaspai, line 40: Can you add a specific test to this file that fills are only considered if they have positive profit? I'm not sure how tests are currently set up but fills need to be over a min revenue, which might not include the gas costs.

nicholaspai (commented):
Left some comments

test/ProfitClient.ConsiderProfitability.ts
kevinuma: Updated

test/ProfitClient.ConsiderProfitability.ts
kevinuma: Updated

src/clients/ProfitClient.ts
kevinuma: Yep

src/clients/ProfitClient.ts
kevinuma: All gas costs are fetched in update()

src/clients/ProfitClient.ts
```diff
@@ -116,32 +153,50 @@ export class ProfitClient {
...
     );
 
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    let cgPrices: Array<CoinGeckoPrice> = [];
+    let cgPrices: CoinGeckoPrice[] = [];
     try {
-      cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
+      const [maticTokenPrice, otherTokenPrices] = await Promise.all([
+        // Add WMATIC for gas cost calculations.
+        this.coingeckoPrices([WMATIC], "polygon-pos"),
```
pxrl, line 168: Should this be conditional upon RELAYER_DESTINATION_CHAINS including Polygon?

package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^1.0.2",
-    "@across-protocol/sdk-v2": "^0.1.10",
+    "@across-protocol/sdk-v2": "^0.1.18",
```
nicholaspai, line 14: The latest version is higher now

src/clients/ProfitClient.ts
```diff
@@ -10,37 +12,52 @@ type CoinGeckoPrice = {
...
+
+const GAS_TOKEN_BY_CHAIN_ID = {
+  1: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  10: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  137: WMATIC,
+  288: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  42161: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+
+  // For tests.
+  666: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
```
nicholaspai, line 26: I don't like adding test chain ID's to this client, I think you should consider injecting `GAS_TOKEN_BY_CHAIN_ID` in the constructor and setting it to a reasonable default. This way in tests you can inject different tokens for test chain IDs

src/clients/ProfitClient.ts
```diff
@@ -10,37 +12,52 @@ type CoinGeckoPrice = {
...
+  1337: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
 };
+// TODO: Make this dynamic once we support chains with gas tokens that have different decimals.
+const GAS_TOKEN_DECIMALS = 18;
```
nicholaspai, line 30: Might as well also inject this decimals

src/clients/ProfitClient.ts
```diff
@@ -116,32 +153,50 @@ export class ProfitClient {
...
-    let cgPrices: Array<CoinGeckoPrice> = [];
+    let cgPrices: CoinGeckoPrice[] = [];
     try {
-      cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
+      const [maticTokenPrice, otherTokenPrices] = await Promise.all([
+        // Add WMATIC for gas cost calculations.
+        this.coingeckoPrices([WMATIC], "polygon-pos"),
+        this.coingeckoPrices(Object.keys(l1Tokens)),
+      ]);
+      cgPrices = cgPrices.concat(maticTokenPrice);
```
nicholaspai, line 171: why not just `concat(maticTokenPrice, otherTokenPrices)`?

src/clients/ProfitClient.ts
```diff
@@ -116,32 +153,50 @@ export class ProfitClient {
...
+      cgPrices = cgPrices.concat(otherTokenPrices);
     } catch (err) {
       this.logger.warn({ at: "ProfitClient", message: "Failed to retrieve prices.", err, l1Tokens });
     }
 
+    // Add to l1Tokens after the fetches, so prices and l1Tokens have the same entries, for any error logging later.
+    l1Tokens[WMATIC] = {
```
nicholaspai, line 178: nit: update `l1Tokens` as close as possible to declaring it, so on line 162 perhaps

src/clients/ProfitClient.ts
```diff
@@ -116,32 +153,50 @@ export class ProfitClient {
...
     const errors: Array<{ [k: string]: string }> = [];
     for (const address of Object.keys(l1Tokens)) {
       const tokenPrice: CoinGeckoPrice = cgPrices.find(
         (price) => address.toLowerCase() === price.address.toLowerCase()
       );
 
-      // todo: Any additional validation to do? Ensure that timestamps are always moving forwards?
-      if (tokenPrice === undefined || typeof tokenPrice.price !== "number") {
+      // TODO: Any additional validation to do? Ensure that timestamps are always moving forwards?
+      if (tokenPrice !== undefined && typeof tokenPrice.price === "number") {
```
nicholaspai, line 191:
```suggestion
      if (tokenPrice !== undefined && !isNan(tokenPrice.price)) {
```

nit: not your change but these `typeof`s that paul left in we should try to remove. I don't think it matters that much if the `.price` is a string or number, as long as it can be coerced into a number by `toBNWei(string | number)`

src/clients/ProfitClient.ts
kevinuma: Nope this is only to fetch WMATIC on Polygon for gas price. Plus it doesn't exist on other chains anyway

src/clients/ProfitClient.ts
```diff
@@ -10,37 +12,52 @@ type CoinGeckoPrice = {
...
+
   constructor(
     readonly logger: winston.Logger,
     readonly hubPoolClient: HubPoolClient,
-    readonly relayerDiscount: BigNumber = toBN(0),
+    spokePoolClients: SpokePoolClientsByChain,
+    readonly enableProfitability: boolean,
+    readonly enabledChainIds: number[],
+    // Default to throwing errors if fetching token prices fails.
+    readonly ignoreTokenPriceFailures: boolean = false,
```
nicholaspai, line 50: Just curious, do you think the relayer of last resort (i.e. RL) should set `ignoreTokenPriceFailures=true` or not? What should we set in prod?

src/clients/ProfitClient.ts
```diff
@@ -150,12 +205,102 @@ export class ProfitClient {
...
 
-  private async coingeckoPrices(tokens: Array<string>) {
-    return await this.coingecko.getContractPrices(tokens, "usd");
+  protected async coingeckoPrices(tokens: string[], platformId?: string) {
+    return await this.coingecko.getContractPrices(tokens, "usd", platformId);
+  }
+
+  private constructRelayerFeeQuery(chainId: number, provider: Provider): relayFeeCalculator.QueryInterface {
+    // Fallback to Coingecko's free API for now.
+    // TODO: Add support for Coingecko Pro.
```
nicholaspai, line 242: eventually we can just use the sdk-v2's `CoingeckoClient` and get the fallback logic for free

src/clients/ProfitClient.ts
kevinuma: I'd propose leaving this in here for now. When we mixed decimals, we'd need a constant map, something similar to TokenSymbol map, which is a constant and doesn't need to be passed in.

src/clients/ProfitClient.ts
kevinuma: That would likely break because it'd try to fetch WMATC in this call this.coingeckoPrices(Object.keys(l1Tokens)). WMATIC doesn't exist on Ethereum.

src/common/ClientHelper.ts
```diff
@@ -161,10 +161,10 @@ export async function constructClients(
     redisClient
   );
 
-  // const gasEstimator = new GasEstimator() // todo when this is implemented in the SDK.
-  const multiCallerClient = new MultiCallerClient(logger, null, config.maxTxWait);
+  const multiCallerClient = new MultiCallerClient(logger);
 
-  const profitClient = new ProfitClient(logger, hubPoolClient);
+  // Disable profitability by default as only the relayer needs it. Relayer has its own client initializations.
```
nicholaspai, line 166: +1

src/clients/ProfitClient.ts
kevinuma: I think we should have ignoreTokenPriceFailures=false so we respect gas profitability and only change it to true when CG is down or when there's some other bug leading to CG failures. This is similar to enableProfitability which should be true by default but can be set to false if there's an incident.

src/relayer/RelayerClientHelper.ts
```diff
@@ -31,10 +31,16 @@ export async function constructRelayerClients(
 
   const tokenClient = new TokenClient(logger, baseSigner.address, spokePoolClients, commonClients.hubPoolClient);
 
+  // Default to all chains.
+  const enabledChainIds =
+    config.relayerDestinationChains.length > 0 ? config.relayerDestinationChains : CHAIN_ID_LIST_INDICES;
```
nicholaspai, line 36: Shouldn't `enabledChainIds` always match 1:1 with the `relayerDestinationChains`? What would happen if there was a chain ID that was NOT enabled in the ProfitClient but was a relayer destination chain?

package.json
kevinuma: Updated to 0.1.21

src/clients/ProfitClient.ts
kevinuma: Removed. We no longer need them as the tests can control which chains the deposits are on.

src/clients/ProfitClient.ts
kevinuma: SG. We can bump the sdk-v2 version when that's ready

src/relayer/RelayerClientHelper.ts
kevinuma: This is a feature where if relayerDestinationChains is not set or is set to [], it defaults to all chain ids.

src/clients/ProfitClient.ts
nicholaspai: I'm ok with that

src/clients/ProfitClient.ts
nicholaspai: fair

src/clients/ProfitClient.ts
nicholaspai: It's ready right now but I want to isolate only the Vercel API to using CG pro to reduce pro requests for time being, basic seems to work for the most part for relayer

src/relayer/RelayerClientHelper.ts
```diff
@@ -31,10 +31,16 @@ export async function constructRelayerClients(
 
   const tokenClient = new TokenClient(logger, baseSigner.address, spokePoolClients, commonClients.hubPoolClient);
 
+  // Default to all chains.
```
nicholaspai, line 34:
```suggestion
  // If `relayerDestinationChains` is a non-empty array, then copy its value, otherwise default to all chains.
```

src/relayer/RelayerClientHelper.ts
nicholaspai: Ah i see that was confusing. Left a comment change suggestion

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -1,74 +1,94 @@
...
 
 let hubPoolClient: MockHubPoolClient, spyLogger: winston.Logger, profitClient: MockProfitClient;
 
-const mainnetWeth = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
+const mainnetWeth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
 const mainnetUsdc = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
 
-describe("ProfitClient: Consider relay profit", async function () {
-  beforeEach(() => {
+describeg("ProfitClient: Consider relay profit", async function () {
```
nicholaspai, line 26:
```suggestion
describe("ProfitClient: Consider relay profit", async function () {
```

nicholaspai: Also, usually `describe` is not async, why are you changing all the arrow functions to `async function/function` and generally changing the test syntax?

nicholaspai (commented):
OOC, why are you changing the test syntax and modifying arrow function syntax and async/sync? Is something not working locally?

src/clients/ProfitClient.ts
nicholaspai: Gotcha

test/ProfitClient.ConsiderProfitability.ts
kevinuma: async function () { is very verbose and can be shortened to () =>, which is better JS syntax IMO. I'll update all other tests to do this in a separate PR.

src/relayer/RelayerClientHelper.ts
kevinuma: Updated

test/ProfitClient.ConsiderProfitability.ts
nicholaspai: wait but async function () {} and () => are NOT the same thing, do you mean async () => ?

Just want to. make sure you're replacing arrow functions correctly

nicholaspai: I also don't think you should be changing test syntax in this PR, just do it all in one big PR

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -1,78 +1,98 @@
...
 
 let hubPoolClient: MockHubPoolClient, spyLogger: winston.Logger, profitClient: MockProfitClient;
 
-const mainnetWeth = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
+const mainnetWeth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
 const mainnetUsdc = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
 
-describe("ProfitClient: Consider relay profit", async function () {
-  beforeEach(() => {
+describe("ProfitClient: Consider relay profit", () => {
```
nicholaspai, line 26: +1

nicholaspai (approved):
LGTM, i'm gonna request from @mrice32 too since this is a sensitive PR!

test/ProfitClient.ConsiderProfitability.ts
kevinuma: async () is not needed. () would handle the async calls fine :) Otherwise things would fail because the await calls.

kevinuma: I reverted these changes as suggested though

src/clients/ProfitClient.ts
```diff
@@ -150,12 +204,102 @@ export class ProfitClient {
...
+        return new relayFeeCalculator.EthereumQueries(
+          provider,
+          undefined,
+          undefined,
+          undefined,
+          undefined,
+          coingeckoProApiKey,
+          this.logger,
+          gasMarkup
+        );
```
pxrl, line 256: This is just a style comment, but is there any possibility for a construction something like the following?

```
const queryHandlers: { [chainId: number]: queryFunction } = {
    1: relayFeeCalculator.EthereumQueries,
    10: relayFeeCalculator.OptimismQueries,
    137: relayFeeCalculator.PolygonQueries,
    288: relayFeeCalculator.BobaQueries,
    42161: relayFeeCalculator.ArbitrumQueries,
}

let queryFn: queryFunction = queryHandlers[chainId];
if (queryFn === undefined) {
    throw new Error(`Unexpected chain ${chainId}`);
}

return new queryFn(
    provider,
    undefined,
    undefined,
    undefined,
    undefined,
    coingeckoProApiKey,
    this.logger,
    gasMarkup
);
```

Advantage would be that it's a 1-line change to add a new chain, and the overhead of changing the structure of the function arguments is quite a bit lower. It's also arguably easier to audit and ensure that all chains are supplying the same arguments in the same order.

I've used it a lot in C and Python, but I'm not sure how common it is in TS/JS though. Does it make sense here?

pxrl (commented):
Driveby comments

src/clients/ProfitClient.ts
kevinuma: Amazing suggestion! It's much cleaner and more elegant. I updated the code.

src/clients/ProfitClient.ts
kevinuma: Hmm that is true. It does make the code a bit more complex to have WMATIC being optional when the relayer's enabledChainIds doesn't include Polygon though. For now I think an extra request to get gas price for Polygon might not be a deal breaker. We can address this in a followup PR if needed.

pxrl: Yeah I can see it's a bit clunky to do in the middle. Happy to consider it as an optimisation later on, once things settle down.

src/clients/ProfitClient.ts
```diff
@@ -10,37 +12,56 @@ type CoinGeckoPrice = {
...
-  28: toBNWei(1),
-  42161: toBNWei(1),
-  137: toBNWei(1),
-  80001: toBNWei(1),
+// We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
+export const WMATIC = "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270";
+
+const GAS_TOKEN_BY_CHAIN_ID = {
+  1: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
+  10: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", // WETH
```
mrice32, line 20: nit: why not make WETH a constant, like WMATIC?

src/clients/ProfitClient.ts
```diff
@@ -73,24 +98,43 @@ export class ProfitClient {
...
     const { decimals, address: l1Token } = this.hubPoolClient.getTokenInfoForDeposit(deposit);
     const tokenPriceInUsd = this.getPriceOfToken(l1Token);
     const fillRevenueInRelayedToken = toBN(deposit.relayerFeePct).mul(fillAmount).div(toBN(10).pow(decimals));
     const fillRevenueInUsd = fillRevenueInRelayedToken.mul(tokenPriceInUsd).div(toBNWei(1));
+
+    // Consider gas cost.
+    const totalGasCostWei = this.getTotalGasCost(deposit.destinationChainId);
+    if (!totalGasCostWei) {
+      const chainId = deposit.destinationChainId;
+      const errorMsg = `Missing total gas cost for ${chainId}. This likely indicate some gas cost request failed`;
```
mrice32, line 122:
```suggestion
      const errorMsg = `Missing total gas cost for ${chainId}. This likely indicates some gas cost request failed`;
```

src/clients/ProfitClient.ts
```diff
@@ -73,24 +98,43 @@ export class ProfitClient {
...
+    }
+
     const { decimals, address: l1Token } = this.hubPoolClient.getTokenInfoForDeposit(deposit);
     const tokenPriceInUsd = this.getPriceOfToken(l1Token);
     const fillRevenueInRelayedToken = toBN(deposit.relayerFeePct).mul(fillAmount).div(toBN(10).pow(decimals));
     const fillRevenueInUsd = fillRevenueInRelayedToken.mul(tokenPriceInUsd).div(toBNWei(1));
+
+    // Consider gas cost.
+    const totalGasCostWei = this.getTotalGasCost(deposit.destinationChainId);
+    if (!totalGasCostWei) {
```
mrice32, line 120: Could this ever be falsy? It looks like the method defaults to returning a BigNumber of 0 if there is no price stored, and I don't think a BigNumber object can be falsy (please correct me if I'm wrong!).

src/clients/ProfitClient.ts
```diff
@@ -73,24 +98,43 @@ export class ProfitClient {
...
+      const chainId = deposit.destinationChainId;
+      const errorMsg = `Missing total gas cost for ${chainId}. This likely indicate some gas cost request failed`;
+      this.logger.error({
+        at: "ProfitClient",
+        message: errorMsg,
+        allGasCostsFetched: this.totalGasCosts,
+        chainId,
+      });
+      throw new Error(errorMsg);
```
mrice32, line 129: Should we log an error _and_ throw an error? Not necessarily wrong, but I think this will end up generating two error alerts. One from the crash of the process and one from the error that you explicitly send here. Also, a crash will generally be retried, so if the issue is sporadic, the retries will mean that no human gets alerted, but this error log will immediately alert someone.

src/clients/ProfitClient.ts
```diff
@@ -73,24 +98,43 @@ export class ProfitClient {
...
-      this.logger.debug({ at: "ProfitClient", message: "Relayer discount set to 100%. Accepting relay" });
-      return true;
-    }
-
     if (toBN(deposit.relayerFeePct).eq(toBN(0))) {
       this.logger.debug({ at: "ProfitClient", message: "Deposit set 0 relayerFeePct. Rejecting relay" });
       return false;
     }
 
+    // This should happen before the previous checks as we don't want to turn them off when profitability is disabled.
```
mrice32, line 106: nit: I'm a little confused by this comment. Should this check happen before or after the checks above? If it should happen before them, then shouldn't we move it?

src/clients/ProfitClient.ts
```diff
@@ -116,32 +161,49 @@ export class ProfitClient {
...
     );
 
     this.logger.debug({ at: "ProfitClient", message: "Updating Profit client", l1Tokens });
-    let cgPrices: Array<CoinGeckoPrice> = [];
+    let cgPrices: CoinGeckoPrice[] = [];
     try {
-      cgPrices = await this.coingeckoPrices(Object.keys(l1Tokens));
+      const [maticTokenPrice, otherTokenPrices] = await Promise.all([
+        // Add WMATIC for gas cost calculations.
+        this.coingeckoPrices([WMATIC], "polygon-pos"),
```
mrice32, line 176: Can we use the ETH mainnet matic token [here](https://etherscan.io/token/0x7d1afa7b718fb893db30a3abc0cfc608aacfebb0) instead, so we don't need to do this special casing for matic? We could then do all of our queries with ethereum as the platform, I think, right?

I think that might also simplify some of the logic below.

src/clients/ProfitClient.ts
```diff
@@ -150,12 +212,53 @@ export class ProfitClient {
         mrkdwn += `- ${token["symbol"]} not found (${token["cause"]}).`;
         mrkdwn += ` Using last known price of ${this.getPriceOfToken(token["address"])}.\n`;
       });
-      this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+      if (this.ignoreTokenPriceFailures) {
+        this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+      } else {
+        this.logger.error({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+        throw new Error(mrkdwn);
```
mrice32, line 219: Same comment as above -- are we sure we want to alert _and_ crash?

src/clients/ProfitClient.ts
```diff
@@ -150,12 +212,53 @@ export class ProfitClient {
...
+      }
+    }
+
+    // Pre-fetch total gas costs for relays on enabled chains.
+    const getGasCosts = [];
+    for (const chainId of this.enabledChainIds) {
+      getGasCosts.push(this.relayerFeeQueries[chainId].getGasCosts());
```
mrice32, line 226: nit: Is there some reason we've spread the looping and Promise.all out? Could we just do it in a simple Promise.all wrapping a map?

```ts
const gasCosts = await Promise.all(this.enabledChainIds.map(
  (chainId) => this.relayerFeeQueries[chainId].getGasCosts()
));
```

mrice32 (commented):
All of this logic looks great! Really excited about rolling this out. A few very minor comments and nits.

src/clients/ProfitClient.ts
kevinuma: I can change the logging to warn and still leave throw new Error so the relayer would fail.

src/clients/ProfitClient.ts
kevinuma: So sorry I had a brain fart. It should be "after"

src/clients/ProfitClient.ts
kevinuma: Changed the logging to warn

src/clients/ProfitClient.ts
kevinuma: Good find! I updated to fetch MATIC price on ethereum

src/clients/ProfitClient.ts
kevinuma: Good suggestion! Updated

src/clients/ProfitClient.ts
kevinuma: Good call! Updated to be more explicit and check against undefined + toBN(0)

src/clients/ProfitClient.ts
kevinuma: Updated

src/clients/ProfitClient.ts
kevinuma: Updated

PR #264, opened 2022-08-18
fix(arbitrum adapter): Correctly filter events by l1Token when multiple share the same gateway contract

kevinuma:
Detailed explanation of the bug here: https://risklabs.slack.com/archives/C01PEM1S8E4/p1660837233794319?thread_ts=1660832199.916539&cid=C01PEM1S8E4


src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -8,8 +16,8 @@ import { arbitrumL2Erc20GatewayInterface, arbitrumL1Erc20GatewayInterface } from
 const l1Gateways = {
   "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": "0xcEe284F754E854890e311e3280b767F80797180d", // USDC
   "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0xd92023E9d9911199a6711321D1277285e6d4e2db", // WETH
-  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0xd3b5b60020504bc3489d6949d545893982ba3011", // DAI
```
kevinuma, line 11: This is not related to the cause of the issue but has inconsistent formatting. So might as well fix these here.

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -91,7 +103,10 @@ export class ArbitrumAdapter extends BaseAdapter {
       // l2DepositFinalizedEvents state from the BaseAdapter.
       eventsToProcess.forEach((result, index) => {
         const l1Token = validTokens[Math.floor(index / 2)];
-        const events = result.map((event) => {
+        // l1Token is not an indexed field on Aribtrum gateway's deposit events so we need to filter unrelated deposits
```
kevinuma, line 106: This is the root cause of the issue that leads to attributing BAL token transfers as WBTC.

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -91,7 +105,10 @@ export class ArbitrumAdapter extends BaseAdapter {
       // l2DepositFinalizedEvents state from the BaseAdapter.
       eventsToProcess.forEach((result, index) => {
         const l1Token = validTokens[Math.floor(index / 2)];
-        const events = result.map((event) => {
+        // l1Token is not an indexed field on Aribtrum gateway's deposit events, so these events are for all tokens.
```
kevinuma, line 108: This is the root cause of the issue that leads to attributing BAL token transfers as WBTC.

src/clients/bridges/ArbitrumAdapter.ts
mrice32: Good call. Would be nice to have a linter that checks for non-checksummed address constants.

mrice32 (approved):
LGTM!

PR #265, opened 2022-08-19
improve(relayer): Add a per-network following distance to spoke pool's event search config

kevinuma:
Forks can happen on chains such as Polygon. We need to add support for following distances to make sure we only fill deposits that have been finalized (and not subject to forks being undone).

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -43,10 +43,7 @@ export class SpokePoolClient {
     readonly configStoreClient: AcrossConfigStoreClient | null,
     readonly chainId: number,
     readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0,
-    // Number of blocks to subtract from latest block the provider returns. This helps avoid fork reorgs that can lead
-    // to double filling of the same deposit.
-    readonly blockFollowingDistance: number = 0
+    readonly spokePoolDeploymentBlock: number = 0
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
   }
@@ -226,7 +223,8 @@ export class SpokePoolClient {
     );
 
     // Use a following distance to ensure the data we see are final.
-    this.latestBlockNumber = (await this.spokePool.provider.getBlockNumber()) - this.blockFollowingDistance;
+    const followingDistance = this.eventSearchConfig.followingDistance ?? 0;
+    this.latestBlockNumber = (await this.spokePool.provider.getBlockNumber()) - followingDistance;
     const searchConfig = {
       fromBlock: this.firstBlockToSearch,
       toBlock: this.eventSearchConfig.toBlock || this.latestBlockNumber,
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -39,10 +39,14 @@ export class SpokePoolClient {
   constructor(
     readonly logger: winston.Logger,
     readonly spokePool: Contract,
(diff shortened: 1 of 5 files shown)


src/common/ClientHelper.ts
```diff
@@ -80,22 +80,23 @@ export async function constructSpokePoolClientsWithLookback(
   });
 
   // Create client for each spoke pool.
-  spokePools.forEach((obj: { networkId: number; contract: Contract }) => {
```
kevinuma, line 83: Better Typescript syntax to deconstruct function args

src/clients/SpokePoolClient.ts
```diff
@@ -39,10 +39,14 @@ export class SpokePoolClient {
...
-    readonly configStoreClient: AcrossConfigStoreClient | null, // Can be excluded. This disables some deposit validation.
+    // Can be excluded. This disables some deposit validation.
+    readonly configStoreClient: AcrossConfigStoreClient | null,
     readonly chainId: number,
     readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0
+    readonly spokePoolDeploymentBlock: number = 0,
+    // Number of blocks to subtract from latest block the provider returns. This helps avoid fork reorgs that can lead
+    // to double filling of the same deposit.
+    readonly blockFollowingDistance: number = 0
```
nicholaspai, line 49: It seems like you could just override the `eventSearchConfig` in `common/ClientHelper/constructSpokePoolClientWithLookback` that the `RelayerClientHelper` uses to construct a new spoke pool client

nicholaspai: Instead of adding a new constructor param

nicholaspai (commented):
I don't think we need that extra constructor param?

src/clients/SpokePoolClient.ts
kevinuma: That wouldn't work as constructSpokePoolClientWithLookback doesn't yet know what the right value for toBlock is (provider.latestBlockNumber might not have been fetched). Plus in cases where update() gets called multiple times, fixing toBlock would affect all subsequent updates. I.e. only the first update() call would do anything.

nicholaspai: hmmm i see. What if we added the follow distance to the `eventSearchConfig` the same way we set the `maxBlockLookback`?

kevinuma: Great idea. Updated.

src/clients/SpokePoolClient.ts
```diff
@@ -221,7 +222,9 @@ export class SpokePoolClient {
       process.env[`NODE_MAX_CONCURRENCY_${this.chainId}`] || NODE_MAX_CONCURRENCY || "1000"
     );
 
-    this.latestBlockNumber = await this.spokePool.provider.getBlockNumber();
+    // Use a following distance to ensure the data we see are final.
+    const followingDistance = this.eventSearchConfig.followingDistance ?? 0;
```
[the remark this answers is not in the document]
kevinuma: I'd propose we stay consistent in the search even config because if we have a different config for each event, it can lead to some strange errors that would be very hard to debug. Especially when we take forks and reorgs into account.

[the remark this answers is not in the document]
mrice32: Seems reasonable. I assume we'd keep this value undefined for everything except for the relayer, right?

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
kevinuma: Yep unless we see the same problem in the dataworker as well

PR #266, opened 2022-08-21
fix(profit client): Allow fetching CG prices when relay gas profitability is disabled

kevinuma:
**Breaking change:** https://github.com/across-protocol/relayer-v2/pull/261 introduced an enableProfitability flag that, when set to false, skip all prices and gas cost updates in ProfitClient#update(). This can be useful for the relayer when we want to disable profitability check.

**Problem:** However, the dataworker also needs to use the ProfitClient to calculate the current volume of deposits when deciding if the volume is big enough to propose a bundle. Since the dataworker has enableProfitability = false, this stops ProfitClient from fetching prices and thus breaks the dataowker's volume calculation (it's always 0).

**Solution:** Allow ProfitClient to fetch prices in update() regardless of the enableProfitability flag.
**Testing:** An extra unit test is added to ensure prices are still fetched when enableProfitability = false. I also ran the dataworker manually and verified that it gets the prices from Coingecko and still performs volume calc correctly.

### DIFF
--- a/test/ProfitClient.PriceRetrieval.ts
+++ b/test/ProfitClient.PriceRetrieval.ts
@@ -2,12 +2,12 @@ import { L1Token } from "../src/interfaces";
 import { expect, createSpyLogger, winston, BigNumber, toBN } from "./utils";
 
 import { MockHubPoolClient } from "./mocks";
-import { ProfitClient, MATIC } from "../src/clients"; // Tested
+import { ProfitClient, MATIC, WETH } from "../src/clients"; // Tested
 
 let hubPoolClient: MockHubPoolClient, spy: sinon.SinonSpy, spyLogger: winston.Logger;
 
 const mainnetTokens: Array<L1Token> = [
-  { symbol: "WETH", address: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2", decimals: 18 },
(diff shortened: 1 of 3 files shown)


test/ProfitClient.PriceRetrieval.ts
```diff
@@ -31,20 +31,30 @@ describe("ProfitClient: Price Retrieval", async function () {
...
   });
 });
+
+const verifyTokenPrices = () => {
+  const tokenPrices: { [k: string]: BigNumber } = profitClient.getAllPrices();
+
+  // The client should have fetched prices for all requested tokens.
+  expect(Object.keys(tokenPrices)).to.have.deep.members(mainnetTokens.map((token) => token["address"]));
+  Object.values(tokenPrices).forEach((price: BigNumber) => expect(toBN(price).gt(toBN(0))).to.be.true);
+  Object.keys(tokenPrices).forEach((token) => expect(toBN(profitClient.getPriceOfToken(token)).gt(toBN(0))).to.be.true);
```
pxrl, line 59: FWIW I have another pending commit that mocks out CoinGecko and hardcodes the response.

pxrl (approved):
LGTM!

mrice32 (approved):
LGTM!

PR #267, opened 2022-08-21
improve(bundle logging): Fix fee formatting and explicitly say when there are no slow relay leaves

kevinuma:
1. Fee formatting is incorrect:
SlowRelay
    root:0xc8a05438...
        leaves:
            0: {"depositor":"0xD275E5cb","recipient":"0xD275E5cb","amount":"15,000,000.00","depositId":50372,"originChain":10,"destinationChain":1,"destToken":"USDC","realizedLpFee":"754,528,997.61%","relayerFee":"727,688,639.33%"}

2. Slack message for new bundle currently prints out when there are no slow fills

SlowRelay
    root:0x00000000...
        leaves:

when there are no slow relay leaves which is confusing as it looks like an error. We should explicitly say "No slow relay leaves"

### DIFF
--- a/src/utils/FormattingUtils.ts
+++ b/src/utils/FormattingUtils.ts
@@ -16,7 +16,8 @@ export const toBN = (num: string | number | BigNumber) => {
 };
 
 export const formatFeePct = (relayerFeePct: BigNumber): string => {
-  return createFormatFunction(2, 4, false, 18)(toBN(relayerFeePct).mul(100).toString());
+  // 1e18 = 100% so 1e16 = 1%.
+  return createFormatFunction(2, 4, false, 16)(toBN(relayerFeePct).toString());
 };
 
 export { createFormatFunction } from "@uma/common";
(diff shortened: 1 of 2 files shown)


pxrl (approved):
LGTM

src/utils/FormattingUtils.ts
```diff
@@ -16,7 +16,8 @@ export const toBN = (num: string | number | BigNumber) => {
 };
 
 export const formatFeePct = (relayerFeePct: BigNumber): string => {
-  return createFormatFunction(2, 4, false, 18)(toBN(relayerFeePct).mul(100).toString());
```
kevinuma, line 19: No need to multiply by 100 when we can change decimals to 16, which makes sense because 1e16 = 1% (1e18 = 100%)

mrice32: +1 nice simplification

mrice32 (approved):
LGTM!

PR #271, opened 2022-08-25
fix: SpokePoolClient incorrectly overrides deposit.relayerFeePct when a RequestedSpeedUp event is emitted

nicholaspai:
By overriding `deposit.relayerFeePct`, the `SpokePoolClient` inadvertently leads to the `relayer` sending invalid fills with an incorrect `relayerFeePct`. This could lead to a double spend attack where a depositor receives their full deposit, requests a speedup on their deposit, and then receives another full deposit. In the second case, the fill is invalid, so this is a relayer bug.

The solution is:
- in `SpokePoolClient.appendMaxSpeedUpSignatureToDeposit` add a new field to deposit instead of overriding the `relayerFeePct`, name it `newRelayerFeePct`
- In `Relayer.fillRelay` if `fill.modifiedRelayerFeePct !== undefined` then call `fillRelayWithUpdatedFee()` instead of `fillRelay` on the `SpokePool` contract and pass in the `newRelayerFeePct` and `depositorSignature` params.

In the case of slow fills, sent via `Relayer.zeroFill()`, we can ignore the `newRelayerFeePct` since [any slow leaves that get executed on SpokePools will result in an effective 0% `relayerFeePct`](https://github.com/across-protocol/contracts-v2/blob/master/contracts/SpokePool.sol#L629), so it doesn't do anything to 1-wei fill the deposit with a higher relayer fee pct since the fees on a 1-wei fill will be negligible.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -92,7 +92,8 @@ export class ProfitClient {
   }
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
-    if (toBN(deposit.relayerFeePct).lt(this.minRelayerFeePct)) {
+    const relayerFeePct = toBN(deposit.newRelayerFeePct ?? deposit.relayerFeePct);
+    if (relayerFeePct.lt(this.minRelayerFeePct)) {
       this.logger.debug({
         at: "ProfitClient",
         message: "Relayer fee % < minimum relayer fee %",
@@ -101,7 +102,7 @@ export class ProfitClient {
       return false;
     }
 
-    if (toBN(deposit.relayerFeePct).eq(toBN(0))) {
+    if (relayerFeePct.eq(toBN(0))) {
       this.logger.debug({ at: "ProfitClient", message: "Deposit set 0 relayerFeePct. Rejecting relay" });
       return false;
     }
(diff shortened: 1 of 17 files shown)


src/clients/ProfitClient.ts
```diff
@@ -92,7 +92,8 @@ export class ProfitClient {
   }
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
-    if (toBN(deposit.relayerFeePct).lt(this.minRelayerFeePct)) {
+    const relayerFeePct = toBN(deposit.newRelayerFeePct ?? deposit.relayerFeePct);
```
nicholaspai, line 95: You should use the max fee % including the newRelayerFeePct if it exists. If newRelayerFeePct is lower than the original relayer fee pct it shouldn't be used.

nicholaspai (commented):
One comment but also probably need to merge master to reduce code diff. Seeing a lot of unrelated code to this PR touching the profit client stuff

src/clients/ProfitClient.ts
kevinuma: > One comment but also probably need to merge master to reduce code diff. Seeing a lot of unrelated code to this PR touching the profit client stuff

Strange may be you're looking at diff of a single commit because you clicked on the link from the email? Looking at the diff in the entire PR doesn't show any unrelated merge changes.

mrice32 (approved):
Core logic looks great! I will do another pass when out of draft, but essentially LGTM!

src/clients/ProfitClient.ts
kevinuma: > You should use the max fee % including the newRelayerFeePct if it exists. If newRelayerFeePct is lower than the original relayer fee pct it shouldn't be used.

Good call. I updated the logic to use the max between the 2. Also added some unit tests in ProfitClient.test.ts

src/clients/ProfitClient.ts
```diff
@@ -92,7 +92,14 @@ export class ProfitClient {
   }
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
-    if (toBN(deposit.relayerFeePct).lt(this.minRelayerFeePct)) {
+    const newRelayerFeePct = toBN(deposit.newRelayerFeePct ?? 0);
+    let relayerFeePct = toBN(deposit.relayerFeePct);
+    // Use the maximum between the original newRelayerFeePct and any updated fee from speedups.
```
kevinuma, line 97: @mrice32 Slight change in the logic

kevinuma, line 97: @mrice32 Slight change in the logic

mrice32 (approved):
LGTM!

PR #272, opened 2022-08-25
improve(config): Safer defaults for min deposit age

pxrl:
Following on from a recent Polygon reorg, increase the default "minimum
age" for deposit transactions before filling them. This will make the relayer
safer for third-party operators.

Ref: ACX-211


src/common/Config.ts
```diff
@@ -62,6 +62,6 @@ export class CommonConfig {
     this.sendingTransactionsEnabled = SEND_TRANSACTIONS === "true";
     this.redisUrl = REDIS_URL;
     this.bundleRefundLookback = BUNDLE_REFUND_LOOKBACK ? Number(BUNDLE_REFUND_LOOKBACK) : 2;
-    this.followingDistances = FOLLOWING_DISTANCES ? JSON.parse(FOLLOWING_DISTANCES) : {};
+    this.followingDistances = FOLLOWING_DISTANCES ? JSON.parse(FOLLOWING_DISTANCES) : Constants.CHAIN_DEPOSIT_MIN_AGE;
```
pxrl, line 65: Note that the name "CHAIN_DEPOSIT_MIN_AGE" is obviously different from the pre-existing "FOLLOWING_DISTANCES". I'd like to hear thoughts on renaming "FOLLOWING_DISTANCES" to this, because it's a bit more descriptive and hopefully more intuitive for any third-parties to intuitively understand.

src/common/Constants.ts
```diff
@@ -11,6 +11,14 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
   42161: 350000,
 };
 
+export const CHAIN_DEPOSIT_MIN_AGE: { [chainId: number]: number } = {
+  1: 2,
+  10: 1,
+  137: 100,
+  288: 1,
+  42161: 1,
```
pxrl, line 19: These numbers are semi plucked out of thin air.

100 for Polygon obviously matches the existing relayer configuration, and 2 for Ethereum was discussed informally and considered OK (implied ~2.5 - ~3 minute delay for deposits that are made within ~30 seconds of the relayer running?).

In practice, Arbitrum and Optimism/Boba apparently don't need a "min age" given their current operation, but they also move so quickly that there's practically no difference between 0 and 1 anyway.

pxrl:
Verified that the bot reads the new defaults correctly:

```
    "followingDistances": {                                                                               
      "1": 2,                                        
      "10": 1,                                                                                            
      "137": 100,                                    
      "288": 1,                                      
      "42161": 1                                     
    },
```

src/common/Constants.ts
mrice32: Why set 1 instead of 0 for them? Seems clearer if we set 0 to imply that there is no need for a follow distance.

I only ask because it made me a little confused when I first saw it :)

src/common/Constants.ts
```diff
@@ -11,6 +11,14 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
   42161: 350000,
 };
 
+export const CHAIN_DEPOSIT_MIN_AGE: { [chainId: number]: number } = {
```
mrice32, line 14: Could you add a comment explaining where these numbers are coming from?

Why 100 on polygon?
Why 2 on mainnet?
Why 0 or 1 on 10, 288, and 42161?

mrice32 (commented):
This is awesome! Just a few minor comments/suggestions

src/common/Constants.ts
kevinuma: 100 is the current value we set for our relayer for Polygon as well. This is based off https://polygonscan.com/blocks_forked where the biggest number of blocks in a reorg seems to be ~60-70 so I thought 100 would be safe.

2 on mainnet is based on Nick's suggestion from his intuition regarding Ethereum I think.
0-1 on other chains seem a bit arbitrary IMO. We should either leave them as 0 or do some quick check on what would be reasonable following distances to use. I'm sure many bridges, especially messaging ones, do these numbers on their side. I remember talking to a couple about following distances before joining us :)

pxrl: I have just updated with new defaults. All Optimistic L2s are set to 0, and there's a comment explaining why. I also found that the blocks_forked page is also available for Etherscan, and a quick scan back through a number of pages showed that reorgs are quite regular, but I didn't see any of depth > 1 block.

src/common/Config.ts
mrice32: I like CHAIN_DEPOSIT_MIN_AGE more for sure. I could also see something like `REQUIRED_DEPOSIT_CONFIRMATIONS` making sense.

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
```diff
@@ -223,8 +223,8 @@ export class SpokePoolClient {
     );
 
     // Use a following distance to ensure the data we see are final.
-    const followingDistance = this.eventSearchConfig.followingDistance ?? 0;
-    this.latestBlockNumber = (await this.spokePool.provider.getBlockNumber()) - followingDistance;
+    const minDepositConfirmations = this.eventSearchConfig.minDepositConfirmations ?? 0;
```
pxrl, line 226: Driveby comment on this - is 0 a desirable default in the event that no minimum was specified for a particular chain? The scenario I am thinking about is when a user has manually specified a config, and then pulls in an update of the bot where a new chain has been supported. That could have unintended consequences.

We could consider asserting on the existence of this config item. This will only blow up if the user has manually set this field, but has failed to specify a specific value for a source chain.

mrice32: I think it's fine to blow up since there should either be a default or a provided config.

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
nicholaspai: i agree with this variable name change but lets try to avoid such breaking changes

src/common/Config.ts
```diff
@@ -18,7 +18,7 @@ export class CommonConfig {
   readonly maxRelayerLookBack: { [chainId: number]: number };
   readonly maxRelayerUnfilledDepositLookBack: { [chainId: number]: number };
   // Following distances in blocks to guarantee finality on each chain.
```
nicholaspai, line 20: Should change this comment

src/common/Constants.ts
```diff
@@ -11,6 +11,18 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
   42161: 350000,
 };
 
+// Reorgs are anticipated on Ethereum and Polygon.
+// Ethereum: https://etherscan.io/blocks_forked
+// Polygon: https://polygonscan.com/blocks_forked
+// Optimistic Rollups are currently centrally serialized and are not expected to reorg.
```
nicholaspai, line 17: +1 nice comment

nicholaspai (approved):
Looks good, lgtm after resolving comments

src/common/Config.ts
pxrl: Making as resolved after moving to MIN_DEPOSIT_CONFIRMATIONS.

src/common/Constants.ts
pxrl: Marking as resolved since we seem to have settled on a sensible config.

src/common/Config.ts
pxrl: Good catch - fixed in the latest update.

src/common/Config.ts
```diff
@@ -62,6 +62,17 @@ export class CommonConfig {
...
     this.redisUrl = REDIS_URL;
     this.bundleRefundLookback = BUNDLE_REFUND_LOOKBACK ? Number(BUNDLE_REFUND_LOOKBACK) : 2;
-    this.followingDistances = FOLLOWING_DISTANCES ? JSON.parse(FOLLOWING_DISTANCES) : {};
+    this.minDepositConfirmations = MIN_DEPOSIT_CONFIRMATIONS
+      ? JSON.parse(MIN_DEPOSIT_CONFIRMATIONS)
+      : Constants.MIN_DEPOSIT_CONFIRMATIONS;
+
+    this.spokePoolChains.forEach((chainId) => {
+      const nBlocks: number = this.minDepositConfirmations[chainId];
+      console.log(`Got nBlocks ${nBlocks} for chainID ${chainId}.`);
```
pxrl, line 71: Oops, that's not meant to be there...

pxrl: Fixed in update.

nicholaspai (approved):

PR #273, opened 2022-08-29
temp(profit client): Temporarily downgrade logs for CG failures to debug

kevinuma:
CG fetches currently fail a lot due to rate limit. Thus, issuing warning logs on these failures currently makes oncall very noisy as it generates a lot of low-urgency alerts.

This PR temporarily lowers the logs to debug. We can change them back to warn after we address underlying CG failure issues.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -208,7 +208,7 @@ export class ProfitClient {
       if (!this.ignoreTokenPriceFailures) {
         throw new Error(errMsg);
       }
-      this.logger.debug({ at: "ProfitClient", message: errMsg, tokens: tokens });
+      this.logger.info({ at: "ProfitClient", message: errMsg, tokens: tokens });
       return;
     }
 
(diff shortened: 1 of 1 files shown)


mrice32 (commented):
Why not info? That would cause them to still show up in slack, but no longer hit PD.

I forget, are we using these prices for anything right now? If we are, I'd be a little uncomfortable burying these alerts since consistent failures could mean the bot is effectively down.

nicholaspai (approved):

src/clients/ProfitClient.ts
```diff
@@ -237,7 +237,7 @@ export class ProfitClient {
         mrkdwn += `- ${token["symbol"]} not found (${token["cause"]}).`;
         mrkdwn += ` Using last known price of ${this.getPriceOfToken(token["address"])}.\n`;
       });
-      this.logger.warn({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
+      this.logger.debug({ at: "ProfitClient", message: "Could not fetch all token prices 💳", mrkdwn });
```
pxrl, line 240: This change isn't needed in order to reduce the noise coming from rate limiting. It would only arise if we receive an incomplete or malformed response from CoinGecko (which would actually be a noteworthy event).

kevinuma:
> Why not info? That would cause them to still show up in slack, but no longer hit PD.
> 
> I forget, are we using these prices for anything right now? If we are, I'd be a little uncomfortable burying these alerts since consistent failures could mean the bot is effectively down.

Updated to be info. Yes we do use these price infos:
1. Dataworker uses the prices to decide when there's enough volume to propose. If the failures sustain, no bundles will be proposed.
2. Relayer uses the prices to decide profitability. But profitability is not yet turned on so the failures don't affect the dataworker.
3. Monitor and disputer use the prices for

PR #274, opened 2022-08-29
fix(profit client): Only instantiate/update profit client for dataworker + relayer

kevinuma:
I reexamined current uses of ProfitClient as it's automatically instantiated and updated as part of the common clients. This means all bots use it although only 2 really need it:
1. Dataworker needs ProfitClient to decide whether there's enough volume to propose a bundle
2. Relayer uses ProfitClient to decide whether a fill is profitable. But profitability is currently disabled so it doesn't need those prices really.

Since the other bots don't need ProfitClient, we can separate it from CommonClients and only specifically instantiate/call update() for the dataworker + relayer. This will significantly reduce the price fetch volume and subsequently alert noise.


pxrl (commented):
Looks OK to me. Really pleased that you spotted this...fingers crossed that it makes a difference.

nicholaspai (approved):
Excellent refactor! I agree this should reduce CG requests originating from monitors

PR #275, opened 2022-08-29
improve(ProfitClient): Make profitability opt-out

pxrl:
The vast majority of relayer operators will want to run with
profitability rules being enforced. The documentation does not currently
inform them of the need to explicitly set ENABLE_PROFITABILITY=true in
their env, and it seems like unnecessary additional configuration when
RL can just set ENABLE_PROFITABILITY=false.


src/relayer/RelayerConfig.ts
```diff
@@ -63,7 +63,7 @@ export class RelayerConfig extends CommonConfig {
         });
       });
     }
-    this.enableProfitability = ENABLE_PROFITABILITY === "true";
+    this.enableProfitability = ENABLE_PROFITABILITY !== "false";
```
pxrl, line 66: An alternative version to this, which I probably prefer, involves changing "enableProfitability" => "ignoreProfitability". This implies inverting the logic, but seems more logical and self-explanatory to me.

Not sure whether we care enough to make that change, since it's mostly cosmetic, but I think it makes the configuration more intuitive/self-explanatory.

pxrl:
Note that a prerequisite to merging this would be that the RL bot configuration should set ENABLE_PROFITABILITY=false, since it currently relies on the default of ENABLE_PROFITABILITY being unset. This implicitly means that any third-party relayer operators who updated to master within the past ~1 - 2 weeks may have been making unprofitable relays without realising it.

mrice32 (approved):
LGTM! Have we changed our internal config?

src/relayer/RelayerConfig.ts
pxrl: @kevinuma Just wanted to check on the naming - do you have any thoughts about it? Reason for bringing it up is that I'm trying to think about how a third-party user might perceive the naming, and am trying to make it ++simple to understand.

kevinuma: We can name it DISABLE_PROFITABILITY if needed. I don't have strong opinions here.

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -121,15 +105,7 @@ describe("ProfitClient: Consider relay profit", async function () {
...
-    const profitClientWithMinFee = new MockProfitClient(
-      spyLogger,
-      hubPoolClient,
-      {},
-      false,
-      [],
-      false,
-      toBNWei("0.03")
-    );
+    const profitClientWithMinFee = new MockProfitClient(spyLogger, hubPoolClient, {}, true, [], false, toBNWei("0.03"));
```
pxrl, line 108: FWIW this was just lint blowing up after a single character was removed :)

nicholaspai (approved):
LGTM. Just want to check: if we merge this, will this change the behavior existing bots?

pxrl:
> LGTM! Have we changed our internal config?

Internal config is now updated in anticipation of this change.

pxrl:
> LGTM. Just want to check: if we merge this, will this change the behavior existing bots?

Nope, no change. The internal configuration has been pre-emptively updated to ignore profitability. So we should be good to go on this.

PR #277, opened 2022-08-29
fix(dataworker): Change missed logging for leave execution failure to warn

kevinuma:
https://github.com/across-protocol/relayer-v2/commit/2f3ee963eb6399b48fca4a0ecbfbf59d51b4df64 only changed the logging for slow relay leaves but not relayer refund leaves


mrice32 (approved):
LGTM!

PR #279, opened 2022-08-31
fix(relayer): Remember to update the profit client

pxrl:
Wasn't working after the recent profitClient dedup. Only detectable when
running with ENABLE_PROFITABILITY=true.

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -68,6 +68,7 @@ export async function constructRelayerClients(
 
 export async function updateRelayerClients(clients: RelayerClients) {
   await updateClients(clients);
+  await clients.profitClient.update();
   // SpokePoolClient client requires up to date HubPoolClient and ConfigStore client.
 
   // TODO: the code below can be refined by grouping with promise.all. however you need to consider the inter


src/relayer/RelayerClientHelper.ts
```diff
@@ -68,6 +68,7 @@ export async function constructRelayerClients(
 
 export async function updateRelayerClients(clients: RelayerClients) {
   await updateClients(clients);
+  await clients.profitClient.update();
```
pxrl, line 71: Note: This could _probably_ be done concurrently with the spoke pool client updates. I haven't done it at this point because this code will probably get refactored as part of the error handling work that's on the backlog.

kevinuma:
Good catch!

src/relayer/RelayerClientHelper.ts
mrice32: How was this working before?

pxrl: > How was this working before?

The profitClient instance was separately instantiated by ClientHelper/constructClients(), and the update() call was done in ClientHelper/updateClients(). This was removed in Kev's recent commit, because the Profit Client is not needed for all instances of the bot.

https://github.com/across-protocol/relayer-v2/commit/379218b61b7ebecd19dafbdd2e254a89d57109ec

It hasn't been erroring because we have profitability disabled in out bots, but I noticed that it was causing problems when profitability is enabled on a local instance.

mrice32 (approved):

src/relayer/RelayerClientHelper.ts
mrice32: Makes sense!

PR #280, opened 2022-08-31
improve(ProfitClient): Safe default for minRelayerFeePct

pxrl:
Third-party relayers should not need to configure this to avoid filling
unprofitable relays.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -2,6 +2,8 @@
 // in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
 export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 
+export const RELAYER_MIN_FEE_PCT = 0.0003;
+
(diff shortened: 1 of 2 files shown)


src/common/Constants.ts
```diff
@@ -2,6 +2,8 @@
 // in the HubPool's proposeRootBundle method should be: Mainnet, Optimism, Polygon, Boba, Arbitrum
 export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 
+export const RELAYER_MIN_FEE_PCT = 0.0003;
```
pxrl, line 5: Have retained the existing name (*_PCT) for this new constant, but we're not actually specifying a percentage here. Should we change it so the user configures an actual percentage (i.e. .03) and then handle the division by 100 internally?

mrice32: Personally, I don't really see anything wrong with labeling something a percentage even if 100% is represented by 1. However, maybe the more accurate phrasing would be "fraction".

Either way, this LGTM

mrice32 (approved):
LGTM!

src/common/Constants.ts
pxrl: It's not contentious, so I'll leave it as-is. Might revisit later if I develop strong feelings either way.

PR #285, opened 2022-09-06
feat(relayer fills): Log invalid fills (gated by a config)

kevinuma:
Currently the relayer silently ignores invalid fills. We want to have the ability to more explicitly monitor these in case the invalid fills are sent by the same relayer wallet.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -43,7 +43,8 @@ export class SpokePoolClient {
     readonly configStoreClient: AcrossConfigStoreClient | null,
     readonly chainId: number,
     readonly eventSearchConfig: EventSearchConfig = { fromBlock: 0, toBlock: null, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0
+    readonly spokePoolDeploymentBlock: number = 0,
+    readonly logInvalidFills: boolean = false
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
   }
@@ -208,6 +209,18 @@ export class SpokePoolClient {
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) isValid = false;
     });
+
+    // Log any invalid deposits with same deposit id but different params.
+    if (fill.depositId === deposit.depositId && !isValid && this.logInvalidFills) {
+      this.logger.info({
+        at: "SpokePoolClient",
+        chainId: this.chainId,
+        message: "Invalid fill found",
+        deposit,
+        fill,
+      });
+    }
+
     return isValid;
   }
 
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -56,7 +56,8 @@ export async function constructSpokePoolClientsWithLookback(
   configStoreClient: AcrossConfigStoreClient,
   config: CommonConfig,
   baseSigner: Wallet,
-  initialLookBackOverride: { [chainId: number]: number } = {}
+  initialLookBackOverride: { [chainId: number]: number } = {},
+  logInvalidFills = false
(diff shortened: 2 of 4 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -208,6 +209,18 @@ export class SpokePoolClient {
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) isValid = false;
     });
+
+    // Log any invalid deposits with same deposit id but different params.
+    if (fill.depositId === deposit.depositId && !isValid && this.logInvalidFills) {
+      this.logger.info({
```
nicholaspai, line 215: Should this be a warn?

src/clients/SpokePoolClient.ts
```diff
@@ -208,6 +209,18 @@ export class SpokePoolClient {
     Object.keys(deposit).forEach((key) => {
       if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) isValid = false;
     });
+
+    // Log any invalid deposits with same deposit id but different params.
+    if (fill.depositId === deposit.depositId && !isValid && this.logInvalidFills) {
```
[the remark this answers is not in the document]
kevinuma: It should because for that issue, all previous fills for the same deposits are considered invalid as the speed up changes the deposit's relayerFeePct, which would lead to a mismatch here.

src/clients/SpokePoolClient.ts
kevinuma: I was thinking of starting out with info to be less noisy. But we can start out with warn

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
nicholaspai: As discussed offline, we should treat the case special if any of these invalid fills were sent by the relayer. Should we return a list of invalid fills that the relayer can use?

nicholaspai (approved):
As a follow up we should enable the relayer to use invalid fill information to block itself from ~double filling~ triple fillling (it can still send the bad double fill but it won't send the third?)

src/clients/SpokePoolClient.ts
kevinuma: Updated to warn

src/clients/SpokePoolClient.ts
```diff
@@ -182,7 +183,30 @@ export class SpokePoolClient {
     if (fillsForDeposit === undefined || fillsForDeposit.length === 0) {
       return { unfilledAmount: toBN(deposit.amount), fillCount: 0 };
     }
-    const fills = fillsForDeposit.filter((fill) => this.validateFillForDeposit(fill, deposit));
+    const fills = fillsForDeposit.filter((fill) => {
+      const isValid = this.validateFillForDeposit(fill, deposit);
```
mrice32, line 187: This is kind of subtle, but will this _always_ catch invalid or double fills?

Won't this miss reorg style cases where the fill's hash doesn't match any deposit hash because the order flipped?

mrice32 (commented):
Just one high-level comment!

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
kevinuma: We can but unfortunately this is going to be a non-trivial change due to the nesting of these calls. We'd need to refactor FillUtils#getUnfilledDeposits and getValidUnfilledAmountForDeposit to bubble up the invalid fills.

src/clients/SpokePoolClient.ts
kevinuma: Yes this would not catch reorg cases where the depositId changes because SpokePoolClient mains a map of (origin chain id, deposit id) => fills. Those cases are much trickier we can't really tell if it's the same deposit anymore if depositid changes

PR #286, opened 2022-09-07
fix(lint): Make sure no console log rule works

kevinuma:


scripts/constructEmergencyRoot.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-process-exit */
+/* eslint-disable no-console */
```
kevinuma, line 2: Script so okay to leave console.logs here.

scripts/sendTokens.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-process-exit */
+/* eslint-disable no-console */
```
kevinuma, line 2: ditto

scripts/unwrapWeth.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-process-exit */
+/* eslint-disable no-console */
```
kevinuma, line 2: ditto

src/utils/GckmsUtils.ts
```diff
@@ -35,6 +35,7 @@ export function getGckmsConfig(keys) {
     try {
       if (fs.existsSync(`${__dirname}/${overrideFname}`)) configOverride = require(`./${overrideFname}`);
     } catch (err) {
+      // eslint-disable-next-line no-console
```
kevinuma, line 38: Not sure what this is used for yet + no access to logger here so I'll leave this console.log here for now

src/utils/Help.ts
```diff
@@ -11,6 +11,7 @@ export function usage(badInput: string = undefined) {
     \tts-node ./index.tx [-h] <--dataworker|--finalizer> --wallet <${walletOpts}>
   `.slice(1); // Skip leading newline
 
+  // eslint-disable-next-line no-console
```
kevinuma, line 14: Not removed for the same reason as above

src/utils/Help.ts
```diff
@@ -56,6 +57,7 @@ export function help() {
     \tRelayer Instructions: ${relayerDocsUrl}
   `.slice(0, -1); // Skip trailing newline
 
+  // eslint-disable-next-line no-console
```
kevinuma, line 60: Not removed for the same reason as above

kevinuma:
> You should look into overriding `eslint` rules for particular paths [here](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns) rather than ignore on a file by file basis.

I originally didn't want to skip all rules for specific dirs and it's not easy to only skip specific rules for certain dirs without creating more eslint config files in local dirs.

nicholaspai (approved):
Nice LGTM! The `overrides` rule will be very useful in `/scripts` and I agree with not touching the `utils/` files as they are potentially used all in unknown ways

PR #287, opened 2022-09-07
feat(relayer): Skip deposits with invalid fills from the same relayer

kevinuma:
Currently the relayer could fill a deposit multiple times if their previous fills were considered invalid. This PR adds protection against potentially making multiple invalid fills for the same deposit.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -177,24 +177,17 @@ export class SpokePoolClient {
       : undefined;
   }
 
-  getValidUnfilledAmountForDeposit(deposit: Deposit): { unfilledAmount: BigNumber; fillCount: number } {
+  getValidUnfilledAmountForDeposit(deposit: Deposit): { unfilledAmount: BigNumber; fillCount: number; invalidFills: Fill[] } {
+    const invalidFills: Fill[] = [];
     const fillsForDeposit = this.depositHashesToFills[this.getDepositHash(deposit)];
     // If no fills then the full amount is remaining.
     if (fillsForDeposit === undefined || fillsForDeposit.length === 0) {
-      return { unfilledAmount: toBN(deposit.amount), fillCount: 0 };
+      return { unfilledAmount: toBN(deposit.amount), fillCount: 0, invalidFills };
     }
-    const fills = fillsForDeposit.filter((fill) => {
-      const isValid = this.validateFillForDeposit(fill, deposit);
-      console.log(
-        "isValid",
-        this.logInvalidFills,
-        isValid,
-        fill.depositId,
-        deposit.depositId,
-        fill.relayerFeePct,
-        deposit.relayerFeePct
-      );
 
+    const validFills: Fill[] = [];
+    for (const fill of fillsForDeposit) {
+      const isValid = this.validateFillForDeposit(fill, deposit);
       // Log any invalid deposits with same deposit id but different params.
       if (this.logInvalidFills && !isValid && fill.depositId === deposit.depositId) {
         this.logger.warn({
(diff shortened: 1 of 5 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -177,24 +177,17 @@ export class SpokePoolClient {
...
-        this.logInvalidFills,
-        isValid,
-        fill.depositId,
-        deposit.depositId,
-        fill.relayerFeePct,
-        deposit.relayerFeePct
-      );
 
+    const validFills: Fill[] = [];
+    for (const fill of fillsForDeposit) {
```
nicholaspai, line 189: this is a bit of a nit but it will make the code a but more succinct if you use a `.reduce()`. I'm not super opinionated but could look nice

```ts
fillsForDeposit.reduce(
    (groupedFills: { validFills: Fill[], invalidFills: Fill[]} , fill: Fill) => { 
        // do grouping logic here based on whether fill is valid 
    },
    { [], [] }
)
```

src/relayer/Relayer.ts
```diff
@@ -77,6 +77,19 @@ export class Relayer {
         continue;
       }
 
+      // Skip deposits that contain invalid fills from the same relayer. This prevents potential corrupted data from
+      // making the same relayer fill a deposit multiple times.
+      if (invalidFills.filter((fill) => fill.relayer === this.relayerAddress).length > 0) {
```
nicholaspai, line 82: Could use a `.some()` instead of `.filter().length > 0`

src/relayer/Relayer.ts
```diff
@@ -77,6 +77,19 @@ export class Relayer {
...
+        this.logger.error({
```
nicholaspai, line 83: +1 on making this `error` level, we definitely want to be aware of these

src/relayer/Relayer.ts
```diff
@@ -77,6 +77,19 @@ export class Relayer {
...
+          at: "Relayer",
+          message: "Skipping deposit with invalid fills from the same relayer",
```
nicholaspai, line 85: Let's add some emojis :)

nicholaspai (commented):
The logic looks sound, I left some stylistic suggestions and would like to see a unit test!

src/clients/SpokePoolClient.ts
kevinuma: Updated

src/relayer/Relayer.ts
kevinuma: Updated

src/relayer/Relayer.ts

test/Relayer.BasicFill.ts
```diff
@@ -49,7 +49,7 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
     hubPoolClient = new HubPoolClient(spyLogger, hubPool);
     configStoreClient = new AcrossConfigStoreClient(spyLogger, configStore, hubPoolClient);
 
-    multiCallerClient = new MultiCallerClient(spyLogger); // leave out the gasEstimator for now.
```
kevinuma, line 52: Outdated comment

src/clients/SpokePoolClient.ts
```diff
@@ -196,15 +220,18 @@ export class SpokePoolClient {
           fill,
         });
       }
-      return isValid;
-    });
+
+      if (isValid) validFills.push(fill);
```
nicholaspai, line 224: Why do you still touch `validFills` and `invalidFills` after implementing the `.reduce()`?

nicholaspai (approved):
LGTM!

src/clients/SpokePoolClient.ts
kevinuma: Merge artifact. Removed now and updated tests to make sure they fail if there are duplicates.

PR #288, opened 2022-09-07
improve(relayer): Default to reasonable following distances

nicholaspai:
Relayers should be able to fill deposits with minimal risk of filling orphaned deposits without having to set the `FOLLOWING_DISTANCES` environment variable. This PR is in accordance with our philosophy that bot runners should have to set minimal environment variables and in most cases only "Pro/professional" bot runners should modify the environment variables.


pxrl (commented):
Ah - we've had a collision :) I have a similar PR for this. They're functionally the same, but mine modifies the name FOLLOWING_DISTANCES to something (hopefully) a bit more self-explanatory.

https://github.com/across-protocol/relayer-v2/pull/272

nicholaspai:
@pxrl gonna close this in favor of #272 :)

PR #289, opened 2022-09-07
fix(relayer): Correctly return 0 for total gas cost when it's not set

kevinuma:
We would still error out if the total gas cost failed to fetch for any chain, but the error message would be more meaningful rather than a read property of undefined error.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -81,7 +81,7 @@ export class ProfitClient {
 
   getTotalGasCost(chainId: number): BigNumber {
     // TODO: Figure out where the mysterious BigNumber -> string conversion happens.
-    return toBN(this.totalGasCosts[chainId]) ?? toBN(0);
+    return this.totalGasCosts[chainId] ? toBN(this.totalGasCosts[chainId]) : toBN(0);
   }
 
   getUnprofitableFills() {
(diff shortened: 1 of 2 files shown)


src/clients/ProfitClient.ts
```diff
@@ -81,7 +81,7 @@ export class ProfitClient {
 
   getTotalGasCost(chainId: number): BigNumber {
     // TODO: Figure out where the mysterious BigNumber -> string conversion happens.
```
nicholaspai, line 83: What is this TODO referencing?

nicholaspai (approved):
Nice LGTM, have a question about a TODO though. prob unrelated to this PR

src/clients/ProfitClient.ts
kevinuma: This refers to a strange problem that without the explicit toBN(), an error along the line of "gt function does not exist on ....". The gas cost is set to a BigNumber here https://github.com/across-protocol/relayer-v2/blob/master/src/clients/ProfitClient.ts#L265 but for some reason, when it's read again, it becomes a string. See https://github.com/across-protocol/relayer-v2/pull/282 for more details

mrice32 (approved):

nicholaspai (approved):

PR #290, opened 2022-09-07
improve(SpokePoolClient): simplify and optimize spoke pool client

mrice32:
Our bots have changed a lot since we originally built them. This PR intends to do some cleanup and optimization based on the usage patterns that are obvious now, but weren't clear when they were originally being written.

List of updates being made to simplify the spoke pool client:
- We shouldn't need to store any event types without the block information. Right now, we store and return some with and some without. We should just add a block number to everything.
- The way we store events isn't optimized for how they're typically queried in the bots. For instance, we do a lot of lookups and filters by chain, but we rarely need a full list of all fills/deposits (especially not over and over). So we should store our data with these queries in mind.
- It doesn't seem to be guaranteed that events are returned by ethers in a particular order. This PR adds an in-place sort to the events before letting them be used, so that all stored events are in a guaranteed order.
- There were quite a few methods that were only used in tests.
- The way we validated fill and deposit structs in the SpokePoolClient seemed a bit unsafe. We relied on the structs coming in with the right keys. For such a sensitive function, it seemed safer to have a hardcoded list of fields to compare.
- In the process of implementing this change, I noticed that the bluebird types were not being importing, causing much of the bluebird-related code to fall back to any, meaning we were losing a good amount of type checking. This revealed a few minor bugs and types that were insufficient, which were fixed.


src/clients/SpokePoolClient.ts
```diff
@@ -33,8 +44,9 @@ export class SpokePoolClient {
   public isUpdated = false;
   public firstBlockToSearch: number;
   public latestBlockNumber: number;
-  public fillsWithBlockNumbers: FillWithBlock[] = [];
-  public depositsWithBlockNumbers: { [DestinationChainId: number]: DepositWithBlock[] } = {};
+  // public fills: FillWithBlock[] = [];
```
nicholaspai, line 47:
```suggestion
```

src/clients/SpokePoolClient.ts
```diff
@@ -593,13 +588,13 @@ export class SpokePoolClient {
     };
   }
 
-  private async getDepositDataFromCache(): Promise<CachedDepositData> {
+  private async getDepositDataFromCache(): Promise<CachedDepositData | Record<string, never>> {
```
nicholaspai, line 591: Can you explain this change? I'm not familiar with `Record` syntax

nicholaspai (approved):
I love this PR, thank you for doing it!
+1 on using bluebird where possible
+1 on removing and consolidating SpokePoolClient functions, this is a much needed facelift

Are there any followup refactors?

src/clients/SpokePoolClient.ts
```diff
@@ -13,18 +13,29 @@ import {
...
+  "amount",
+  "originChainId",
+  "relayerFeePct",
+  "realizedLpFeePct",
+  "depositId",
+  "depositor",
+  "recipient",
+  "destinationChainId",
+  "destinationToken",
+] as const;
```
kevinuma, line 34: Why do we need "as const"?

src/clients/SpokePoolClient.ts
```diff
@@ -234,12 +226,12 @@ export class SpokePoolClient {
...
   // by the depositor as well as the realizedLpFeePct and the destinationToken, which are pulled from other clients.
   validateFillForDeposit(fill: Fill, deposit: Deposit) {
-    let isValid = true;
-    Object.keys(deposit).forEach((key) => {
-      if (fill[key] !== undefined && deposit[key].toString() !== fill[key].toString()) isValid = false;
+    // Note: this short circuits when a key is found where the comparison doesn't match.
+    // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
+    // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
+    return FILL_DEPOSIT_COMPARISON_KEYS.every((key) => {
+      return fill[key] !== undefined && fill[key].toString() === deposit[key]?.toString();
```
kevinuma, line 233: Nice!!!

kevinuma (approved):

src/clients/SpokePoolClient.ts
mrice32: There's a bit of a debate around this online, but this basically just means that no property is allowed to exist, aka {} is the only thing that can satisfy it.

The type {} cannot be used (and you get a warning) because any non-nullish value works for it, which is not what we want.

In the future, we should probably return null or undefined rather than returning {}, since it's hard and controversial to have a type that _only_ {} works for.

src/clients/SpokePoolClient.ts
mrice32: This means typescript will see the array as the narrowest possible type. For instance, it sees this as `("amount" | "originChainId" | ...)[]` rather than `string[]`. This doesn't matter too much now, but once we turn on strict in typescript, we will get an error if one of these fields _doesn't exist_ on the fill or deposit types because we use elements of this array to access properties in those objects.

src/utils/ArrayUtils.ts
```diff
@@ -0,0 +1,16 @@
+import { SortableEvent } from "../interfaces";
+import lodash from "lodash";
+
+export function sliceBlockRangeSorted<T extends SortableEvent>(
```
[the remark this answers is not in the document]
mrice32: I built this a few weeks ago and forgot to actually put it into action. In general, I think this will make some of our filter calls more efficient. I will remove it and save it for a follow-up PR.

mrice32:
> I love this PR, thank you for doing it! +1 on using bluebird where possible +1 on removing and consolidating SpokePoolClient functions, this is a much needed facelift
> 
> Are there any followup refactors?

Yes, the first would be to turn on strict typing in typescript. The second might be to put in faster block range filtering on sorted arrays (but I'd like to profile it first).

src/clients/SpokePoolClient.ts
nicholaspai: Ah ok

nicholaspai: Its kind of nice to expect an object even ifits empty because we usually iterate the object on its keys which could be empty

PR #291, opened 2022-09-08
chore: Ignore bundles on chain 1 with USDT payments

nicholaspai:
Admin bundle 357 replaces bundles 322 and 321 that cannot be executed on current Ethereum_SpokePool. 357 is executed but won't be able to be reproduced by dataworker since its construction bypasses dataworker logic. When Ethereum_SpokePool is upgraded so that 322 and 321 could be executed, we should NOT execute these or it will result in a deficit of USDT and excess payment to the relayer


mrice32 (approved):
LGTM!

PR #293, opened 2022-09-09
chore: bump arbitrum sdk to 2.0.17

kevinuma:
Arbitrum sdk's latest version is now 2.0.17 and was from before the Nitro upgrade. Their tutorials are all using 2.0.17 now so we should bump our version as well.


nicholaspai (approved):

nicholaspai:
@kevinuma please start filling in PR descriptions even if they are a single line

PR #294, opened 2022-09-10
fix(relayer): Make decision to skip deposits with invalid fills from same relayer configurable by environment variable

nicholaspai:
There is a possibility that a Relayer has incorrectly submitted a fill, and then its code is patched, and then it is attempting to submit a valid fill, but the current logic would have the relayer skip this situation. This behavior should be able to be overridden manually by a bot runner.

Additionally, this PR fixes a false positive test that was succeeding for an unexpected reason

### DIFF
--- a/src/relayer/RelayerConfig.ts
+++ b/src/relayer/RelayerConfig.ts
@@ -16,6 +16,7 @@ export class RelayerConfig extends CommonConfig {
   readonly relayerDestinationChains: number[];
   readonly minRelayerFeePct: BigNumber;
   readonly logInvalidFills: boolean;
+  readonly skipInvalidFills: boolean;
 
   constructor(env: ProcessEnv) {
     const {
@@ -28,6 +29,7 @@ export class RelayerConfig extends CommonConfig {
       SEND_SLOW_RELAYS,
       MIN_RELAYER_FEE_PCT,
       LOG_INVALID_FILLS,
+      SKIP_INVALID_FILLS,
     } = env;
     super(env);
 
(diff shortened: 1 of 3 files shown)


test/Relayer.UnfilledDeposits.ts
```diff
@@ -256,13 +258,16 @@ describe("Relayer: Unfilled Deposits", async function () {
     await relayerInstance.checkForUnfilledDepositsAndFill();
```
nicholaspai, line 258: This test was working for the wrong reason. It was not filling anything because the bot "had insufficient balance" because the `tokenClient` was not included in `updateAllClients`.

mrice32 (approved):
LGTM!

test/Relayer.UnfilledDeposits.ts
kevinuma: Nice thanks for the catch!

src/relayer/RelayerConfig.ts
```diff
@@ -71,5 +73,6 @@ export class RelayerConfig extends CommonConfig {
     this.sendingRelaysEnabled = SEND_RELAYS === "true";
     this.sendingSlowRelaysEnabled = SEND_SLOW_RELAYS === "true";
     this.logInvalidFills = LOG_INVALID_FILLS === "true";
+    this.skipInvalidFills = SKIP_INVALID_FILLS === "true";
```
kevinuma, line 76: Should we set this to true by default?

nicholaspai: yep done, changed the name to `acceptInvalidFills` so if env is not set then relayer will skip the invalid fills

PR #296, opened 2022-09-13
improve(tsconfig): Add watch defaults

pxrl:
Opens up a nice productivity boost when using tsc [...] --watch.

Mostly informed by the typescript docs:
https://www.typescriptlang.org/docs/handbook/configuring-watch.html


daywiss (commented):
is there a command to run this in watch mode?

pxrl:
> is there a command to run this in watch mode?

`tsc --build --incremental --watch` works for me.

daywiss:
> tsc --build --incremental --watch works for me.

would be good to add this to the package.json file as like `yarn watch`

mrice32 (approved):
Could you add a quick line in the readme about how this can be used?

pxrl:
> Nice!! Could you add a quick line in the readme about how this can be used?

Sure, np.

pxrl:
This response applies to both comments above:

> > tsc --build --incremental --watch works for me.
> 
> would be good to add this to the package.json file as like `yarn watch`

> Nice!! Could you add a quick line in the readme about how this can be used?

There seem to be a few different options for using it, but my current thinking is that `yarn build --incremental --watch` is probably the example that we might want to provide. Then we could potentially piggyback on the existing `yarn build` command (pending the following consideration...). I am however curious: our `yarn build` script target is as follows:

`tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./dist/`

This is a pattern that appears in a couple of across-protocol repos, but seems less common in UMAproject repos. Is there a reason for needing the subsequent call to rsync? I've read the docs for tsconfig and I _think_  the rsync part could be handled automatically by additional entries in tsconfig.json. If so, I'd probably advocate to add that new config to tsconfig.json, and then adjust `yarn build` to be the following:

`tsc --build`

With this, it's possible to append arbitrary arguments to `yarn build` and we don't need to specify additional script targets.

daywiss:
would be great if we could avoid using rsync

pxrl:
> would be great if we could avoid using rsync

So, I _think_ we're fine to just run with `tsc --build`. I've tested this by nuking my ./dist and ./node_modules directories, and running `yarn` and `tsc --build`. After that it seems like I can run the bot without any issues.

```
$ rm -r ./node_modules ./dist
$ yarn                                                                                                    
yarn install v1.22.19                                
[1/5] Validating package.json...                     
[2/5] Resolving packages...                                                                                                                                                                                         
[3/5] Fetching packages...                           
[4/5] Linking dependencies...
<....>
gyp ERR! not ok 
Done in 159.36s.
$ npx tsc --build
$ node ./dist/index.js 
    Usage:
        ts-node ./index.ts --help
        ts-node ./index.ts [-h] <--monitor|--relayer>      --wallet <mnemonic|privateKey|gckms>
        ts-node ./index.tx [-h] <--dataworker|--finalizer> --wallet <mnemonic|privateKey|gckms>
```

package.json
```diff
@@ -47,7 +47,8 @@
     "prettier": "prettier .",
     "eslint": "eslint .",
     "test": "hardhat test",
-    "build": "tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./dist/"
```
pxrl, line 50: The only thing I can think of is that rsync might be faster or more efficient to populate ./dist than whatever tsc is doing.

package.json
```diff
@@ -47,7 +47,8 @@
...
+    "build": "tsc --build",
+    "watch": "tsc --build --incremental --watch"
```
pxrl, line 51: Need to document this in README.md.

pxrl:
Have confirmed in a new environment that `tsc --build && node ./dist/index.js <...>` works. `rsync` may offer some marginal performance increase, but that's probably offset by the use of the `--incremental` flag when iteratively building during development.

daywiss (approved):
nice

package.json
pxrl: Just to wrap this one up, I built relayer-v2 against master and against this branch, then saved the output if `find dist > dist.files.<branch-name>`. Both output files have identical SHA256 sums. This is enough for me to conclude that we're OK to proceed.

pxrl:
> Have confirmed in a new environment that `tsc --build && node ./dist/index.js <...>` works. `rsync` may offer some marginal performance increase, but that's probably offset by the use of the `--incremental` flag when iteratively building during development.

See also conclusions here: https://github.com/across-protocol/relayer-v2/pull/296#discussion_r979698219

PR #297, opened 2022-09-13
improve: Update validation script

nicholaspai:
- Script is missing instructions about using cache﻿
- Script should not set allowances on hub pool for wallet


mrice32 (approved):
LGTM!

PR #299, opened 2022-09-13
improve: Remove less useful Relayer logs

nicholaspai:
We have a lot of debug level logs that aren't that useful in practice and add to noise in production. We can selectively add these back in future if neccessary


kevinuma (approved):

mrice32 (approved):
LGTM!

PR #300, opened 2022-09-13
chore(constants): Align minimum default confirmations

pxrl:
As the merge approaches, be a bit more conservative and
deliberate when selecting deposits to be filled.

FWIW these are ++easy to do, but it feels a bit cumbersome.
Still, if we've encouraged users to use the default values, and
if we have adjusted our own internal defaults, then it seems
like the right thing to do.


pxrl:
Assuming this goes in, I'd like to announce to all with the Relayer role in the Discord that they should update their bots, and prepare to take their bots offline before the merge.

mrice32 (approved):
LGTM!

PR #301, opened 2022-09-14
feat: add target balances

mrice32:
This is a draft of how the target balance functionality might work. This is meant as an implementation proposal roughly following [this spec](https://docs.google.com/document/d/1kpu4r9FGU8CmAD6oUHVWX6V7DGkHww-hBdcdU7bQMLE/edit?usp=sharing).

The associated UMIP change is here: https://github.com/UMAprotocol/UMIPs/pull/556.

Note: the functionality is intended to be 100% backwards compatible, because a default threshold and target of 0 should result in the current behavior.

### DIFF
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -186,7 +186,6 @@ export class AcrossConfigStoreClient {
           this.cumulativeSpokeTargetBalanceUpdates.push({ ...args, spokeTargetBalances: {}, l1Token });
         }
       } catch (err) {
-        console.log(err);
         continue;
       }
     }
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -186,6 +186,7 @@ export class AcrossConfigStoreClient {
           this.cumulativeSpokeTargetBalanceUpdates.push({ ...args, spokeTargetBalances: {}, l1Token });
         }
       } catch (err) {
+        console.log(err);
         continue;
       }
     }
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -9,8 +9,17 @@ import {
   EventSearchConfig,
   utf8ToHex,
   getCurrentTime,
+  toBN,
 } from "../utils";
-import { L1TokenTransferThreshold, Deposit, TokenConfig, GlobalConfigUpdate } from "../interfaces";
+import {
+  L1TokenTransferThreshold,
+  Deposit,
+  TokenConfig,
+  GlobalConfigUpdate,
+  ParsedTokenConfig,
+  SpokeTargetBalanceUpdate,
+  SpokePoolTargetBalance,
+} from "../interfaces";
 
 import { lpFeeCalculator } from "@across-protocol/sdk-v2";
 import { BlockFinder, across } from "@uma/sdk";
@@ -29,6 +38,7 @@ export class AcrossConfigStoreClient {
   public cumulativeTokenTransferUpdates: L1TokenTransferThreshold[] = [];
   public cumulativeMaxRefundCountUpdates: GlobalConfigUpdate[] = [];
   public cumulativeMaxL1TokenCountUpdates: GlobalConfigUpdate[] = [];
+  public cumulativeSpokeTargetBalanceUpdates: SpokeTargetBalanceUpdate[] = [];
 
   private rateModelDictionary: across.rateModel.RateModelDictionary;
   public firstBlockToSearch: number;
@@ -85,6 +95,17 @@ export class AcrossConfigStoreClient {
     return config.transferThreshold;
   }
 
+  getSpokeTargetBalancesForBlock(
+    l1Token: string,
+    blockNumber: number = Number.MAX_SAFE_INTEGER
+  ): { [chainId: number]: SpokePoolTargetBalance } {
+    const config = (sortEventsDescending(this.cumulativeTokenTransferUpdates) as SpokeTargetBalanceUpdate[]).find(
+      (config) => config.blockNumber <= blockNumber && config.l1Token === l1Token
+    );
+    if (!config?.spokeTargetBalances) return {};
+    return config.spokeTargetBalances;
+  }
+
   getMaxRefundCountForRelayerRefundLeafForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number {
     const config = (sortEventsDescending(this.cumulativeMaxRefundCountUpdates) as GlobalConfigUpdate[]).find(
       (config) => config.blockNumber <= blockNumber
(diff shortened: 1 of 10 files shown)


src/clients/ConfigStoreClient.ts
```diff
@@ -85,6 +95,17 @@ export class AcrossConfigStoreClient {
     return config.transferThreshold;
   }
 
+  getSpokeTargetBalancesForBlock(
+    l1Token: string,
+    blockNumber: number = Number.MAX_SAFE_INTEGER
+  ): { [chainId: number]: SpokePoolTargetBalance } {
+    const config = (sortEventsDescending(this.cumulativeTokenTransferUpdates) as SpokeTargetBalanceUpdate[]).find(
+      (config) => config.blockNumber <= blockNumber && config.l1Token === l1Token
```
nicholaspai, line 103:
```suggestion
      (config) => config.l1Token === l1Token && config.blockNumber <= blockNumber
```

tiny nit but small optimization to filter on l1 token first. I doubt there will be 1000s of this event but just good practice to filter on rarer conditions first

src/clients/ConfigStoreClient.ts
```diff
@@ -141,7 +163,22 @@ export class AcrossConfigStoreClient {
...
+          transferThreshold: toBN(transferThresholdForToken),
+          l1Token,
+        });
+
+        if (parsedValue?.spokeTargetBalances) {
+          // Note: cast is required because fromEntries always produces string keys, despite the function returning a
+          // numerical key.
+          const targetBalances = Object.fromEntries(
+            Object.entries(parsedValue.spokeTargetBalances).map(([chainId, targetBalance]) => {
+              return [chainId, { target: toBN(targetBalance.target), threshold: toBN(targetBalance.threshold) }];
```
nicholaspai, line 177: So targets will be in wei? I think this is consistent with the other parameter units like `tokenTransferthreshold`

src/dataworker/DataworkerUtils.ts
```diff
@@ -151,9 +151,15 @@ export function _buildRelayerRefundRoot(
         tokenTransferThresholdOverrides[l1TokenCounterpart] ||
         clients.configStoreClient.getTokenTransferThresholdForBlock(l1TokenCounterpart, endBlockForMainnet);
 
+      const spokePoolTargetBalance = clients.configStoreClient.getSpokeTargetBalancesForBlock(
+        l1TokenCounterpart,
+        endBlockForMainnet
+      )[Number(repaymentChainId)] || { target: toBN(0), threshold: toBN(0) };
```
nicholaspai, line 157: Should we do the defaulting (`|| { target: toBN(0), threshold: toBN(0) }`) in the ConfigStoreClient? Might be cleaner to abstract the defaulting logic away from the Dataworker

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -285,6 +286,14 @@ export function constructPoolRebalanceLeaves(
             configStoreClient.getTokenTransferThresholdForBlock(l1Token, latestMainnetBlock)
         );
 
+        const spokeTargetBalances = l1TokensToIncludeInThisLeaf.map(
+          (l1Token) =>
+            configStoreClient.getSpokeTargetBalancesForBlock(l1Token, latestMainnetBlock)[Number(chainId)] || {
+              target: toBN(0),
```
nicholaspai, line 292: Same comment as above. I'm seeing a lot of this redundant defaulting code that could be removed

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -316,16 +333,52 @@ export function constructPoolRebalanceLeaves(
...
+  // Transfer is always desired if hub owes spoke.
+  if (runningBalance.gte(0)) return runningBalance;
+
+  // Running balance is negative, but its absolute value is less than the spoke pool target balance threshold.
+  // In this case, we transfer nothing.
+  if (runningBalance.abs().lt(spokePoolTargetBalance.threshold)) return toBN(0);
+
+  // We are left with the case where the spoke pool is beyond the threshold.
+  // A transfer needs to be initiated to bring it down to the target.
+  const transferSize = runningBalance.abs().sub(spokePoolTargetBalance.target);
```
nicholaspai, line 351: This makes me think: should we enforce somewhere that the `spokePoolTargetBalance.target` is always positive?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -316,16 +333,52 @@ export function constructPoolRebalanceLeaves(
   return leaves;
 }
 
+// Note: this function computes the intended transfer amount before considering the transfer threshold.
+// A positive number indicates a transfer from hub to spoke.
+export function computeDesiredTransferAmount(
```
nicholaspai, line 338: Should we rename this function to `...ToSpoke` or `...toHub`? I'm a bit confused reading this in isolation to determine which direction the transfer amount should be interpreted as

nicholaspai (commented):
This is really nice and overall very easy to understand, great work. Should we also open a draft UMIP PR reflecting these changes so they can be merged together?

src/clients/ConfigStoreClient.ts
mrice32: Good point!

src/clients/ConfigStoreClient.ts
mrice32: Yep, I think that makes the most sense and is consistent.

src/dataworker/DataworkerUtils.ts
mrice32: Hmmm, yeah, I think that could make sense. Let me try to rework it a little.

src/dataworker/PoolRebalanceUtils.ts
mrice32: Good idea!

src/dataworker/DataworkerUtils.ts
mrice32: Done.

src/dataworker/PoolRebalanceUtils.ts
mrice32: Done.

src/dataworker/PoolRebalanceUtils.ts
mrice32: We have some choices:
- Ignore the specific target balance object with negative values.
- Ignore the entire updates that contain any negative values.
- Treat negative values as 0.

I've gone with the last option, but open to the other two as well -- they just might be a bit more complex. This has been added to the implementation. PTAL!

src/dataworker/PoolRebalanceUtils.ts
mrice32: Done!

mrice32:
> This is really nice and overall very easy to understand, great work. Should we also open a draft UMIP PR reflecting these changes so they can be merged together?

Yes, I will do this.

mrice32:
@nicholaspai see the description for a link to the UMIP PR.

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: negative options as 0 makes sense!

nicholaspai (approved):
LGTM since this seems like it is backwards compatible and will not impact bots in prod without a config change on the ConfigStore contract

PR #302, opened 2022-09-14
improve(relayer): Move follow distance logic to Relayer

nicholaspai:
This PR was motivated by a finding that the Relayer was submitting duplicate partial fills because it could not see its own fills due to the follow distance applied to ALL events in the SpokePoolClient.

This PR (1) applies the follow distance only to Deposit events and (2) isolates the logic to the Relayer bot so other bots are not affected

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -35,10 +35,19 @@ export class Relayer {
     // Fetch all unfilled deposits, order by total earnable fee.
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
-    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack).sort(
-      (a, b) =>
+
+    // Require that all fillable deposits meet the minimum specified number of confirmations.
+    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack)
+      .filter((x) => {
+        return (
+          x.deposit.originBlockNumber <=
+            this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
+              this.config.minDepositConfirmations[x.deposit.originChainId] ?? 0
+        );
+      })
+      .sort((a, b) =>
         a.unfilledAmount.mul(a.deposit.relayerFeePct).lt(b.unfilledAmount.mul(b.deposit.relayerFeePct)) ? 1 : -1
-    );
+      );
     if (unfilledDeposits.length > 0) {
       this.logger.debug({ at: "Relayer", message: "Unfilled deposits found", number: unfilledDeposits.length });
     } else {
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -42,7 +42,7 @@ export class Relayer {
         return (
           x.deposit.originBlockNumber <=
           this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
-            (this.config.minDepositConfirmations[x.deposit.originChainId] ?? 0)
+            this.config.minDepositConfirmations[x.deposit.originChainId]
         );
       })
       .sort((a, b) =>
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -41,8 +41,8 @@ export class Relayer {
       .filter((x) => {
         return (
           x.deposit.originBlockNumber <=
-            this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
-              this.config.minDepositConfirmations[x.deposit.originChainId] ?? 0
+          this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
+            (this.config.minDepositConfirmations[x.deposit.originChainId] ?? 0)
         );
       })
       .sort((a, b) =>
--- a/test/utils/utils.ts
+++ b/test/utils/utils.ts
@@ -18,6 +18,8 @@ import { buildRelayerRefundTree, toBN, toBNWei, utf8ToHex } from "../../src/util
 import winston from "winston";
 import sinon from "sinon";
 import chai from "chai";
+import chaiExclude from "chai-exclude";
+chai.use(chaiExclude);
 export { winston, sinon };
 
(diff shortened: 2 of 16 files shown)


src/relayer/Relayer.ts
```diff
@@ -35,10 +35,19 @@ export class Relayer {
...
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
-    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack).sort(
-      (a, b) =>
+
+    // Require that all fillable deposits meet the minimum specified number of confirmations.
+    const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack)
+      .filter((x) => {
+        return (
+          x.deposit.originBlockNumber <=
```
kevinuma, line 43: This doesn't take into account speedups right? I wonder if there's any edge case there.

nicholaspai: so the spokepool client will see the speedup and apply it to the deposit, right?

kevinuma: Yes I was thinking of the case of forks where the speed up event gets undone. This would basically proceed with a relay using the sped up data, which means higher fee. This is fine IMO but just wanted to double check.

test/utils/utils.ts
```diff
@@ -18,6 +18,8 @@ import { buildRelayerRefundTree, toBN, toBNWei, utf8ToHex } from "../../src/util
 import winston from "winston";
 import sinon from "sinon";
 import chai from "chai";
+import chaiExclude from "chai-exclude";
+chai.use(chaiExclude);
```
kevinuma, line 22: What is this for?

nicholaspai: I had to change the type of the `Deposit ==> DepositWithBlock` to be able to synchronously read a deposit's `blockNumber` and `originBlockNumber` but the tests don't gracefully handle these new props. I am using `chaiExclude` to easily ignore those keys in the `DepositWithBlock` comparison, see the test changes in `Relayer.UnfilledDeposits.ts`

src/relayer/Relayer.ts
```diff
@@ -35,10 +35,19 @@ export class Relayer {
...
+          this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
+            (this.config.minDepositConfirmations[x.deposit.originChainId] ?? 0)
```
pxrl, line 45: Just to confirm, is it correct that the fallback to 0 is only required to make the tests work?

pxrl (commented):
Other than one small question, I think there might be some additional code to clean up in test/Monitor.ts. I think it should be as simple as removing lines 73 and 88 in the file linked below. It's not required at all for that to happen as part of this PR; we should just make sure not to leave it hanging around.

https://github.com/across-protocol/relayer-v2/commit/67ce2ddf3f6bfbc59f94ca4eddc33330589fed14#diff-181c41b36b4d5e7f7dbc26851f1dd1f79ef1695cd83bf6bc41776498bf1a8de6

src/relayer/Relayer.ts
nicholaspai: well if the `minDepositConfirmations` for a chain isn't set it might fail since subtracting `undefined` could be an issue?

nicholaspai:
> Other than one small question, I think there might be some additional code to clean up in test/Monitor.ts. I think it should be as simple as removing lines 73 and 88 in the file linked below. It's not required at all for that to happen as part of this PR; we should just make sure not to leave it hanging around.
> 
> [67ce2dd#diff-181c41b36b4d5e7f7dbc26851f1dd1f79ef1695cd83bf6bc41776498bf1a8de6](https://github.com/across-protocol/relayer-v2/commit/67ce2ddf3f6bfbc59f94ca4eddc33330589fed14#diff-181c41b36b4d5e7f7dbc26851f1dd1f79ef1695cd83bf6bc41776498bf1a8de6)

Great catch paul! removed

src/relayer/Relayer.ts
pxrl: I _think_ we should be protected against an undefined minDepositConfirmations in normal scenarios, because we assert() on their existence when reading in the configuration: https://github.com/across-protocol/relayer-v2/pull/302/files#diff-cfd7a6f4fb96bcb633a6a603efa2cc1ac9fe6879bfafdef1c3a935be188e27db

The justification for this is that a user might define an alternative configuration for themselves, and then we add a new chain, and their environment suddenly silently doesn't handle that. I tried to remove this fallback to 0 previously, but found that it created issues in Dataworker tests (...and I remember being confused as to why this should impact the Dataworker!).

I think it'd be nice to remove the implicit fallback here (or at least revert to a default value from Constants). It's not necessarily a part of this change, though :)

kevinuma (approved):

src/relayer/Relayer.ts
mrice32: We could also verify this when ingesting the env variable rather than doing it here.

mrice32 (approved):
LGTM once @pxrl's comment thread is resolved!

src/relayer/Relayer.ts
nicholaspai: I think the fallback to 0 is useful though in this exact situation, right?

> a user might define an alternative configuration for themselves, and then we add a new chain, and their environment suddenly silently doesn't handle that.

pxrl: > I think the fallback to 0 is useful though in this exact situation, right?

My preference is to just fail hard and fast if we detect this situation, and that's how it currently works. If we decided to allow for a silent fallback, I'd opt for falling back to our own internal default for the specific chain, rather than 0.

So, this is a bit of an academic discussion, since we _should_ never encounter this behaviour in practice given the current strictness of the config validation. I just wanted to take the opportunity to discuss potentially removing/adjusting this fallback as well, which would imply that we need to introduce minDepositConfirmations support in our tests. I looked at this briefly but was seeing test chain IDs and never found where they were defined.

pxrl (approved):
Ship it!

src/relayer/Relayer.ts
nicholaspai: I've removed the fallback and fixed tests :)

PR #303, opened 2022-09-14
improve(dataworker): Output token symbol in dataworker's leaf execution logs

kevinuma:
Current log requires more work to understand which token is involved:

1-4. Executed RelayerRefundLeaf :herb:!: rootBundleId: 384
relayerRefundRoot: 0x1c4d6a3e3bc7fc3fc1f9e7d71333333a7673cb41892f098d24b7aa647bd4eeb4
Leaf: 16
chainId: 42161
token: 0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8
amount: 908438221
tx: [0x9dc...edff4e](https://arbiscan.io/tx/0x9dc7bdb229c9907ff4f749dac789c41b52cae4a527a6114da85a12734dedff4e)

This PR would make it easier by printing out the token symbol instead of address.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -7,6 +7,7 @@ import {
   MerkleTree,
   toBN,
   sortEventsAscending,
+  getNetworkName,
 } from "../utils";
 import { toBNWei, getFillsInRange, ZERO_ADDRESS } from "../utils";
 import { DepositWithBlock, FillsToRefund, FillWithBlock, UnfilledDeposit } from "../interfaces";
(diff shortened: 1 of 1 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1194,10 +1198,13 @@ export class Dataworker {
...
 
           fundedLeaves.forEach((leaf) => {
+            const l1TokenInfo = this.clients.hubPoolClient.getL1TokenInfoForL2Token(leaf.l2TokenAddress, chainId);
+            const networkName = getNetworkName(chainId);
+
             const mrkdwn = `rootBundleId: ${rootBundleRelay.rootBundleId}\nrelayerRefundRoot: ${
               rootBundleRelay.relayerRefundRoot
-            }\nLeaf: ${leaf.leafId}\nchainId: ${chainId}\ntoken: ${
-              leaf.l2TokenAddress
+            }\nLeaf: ${leaf.leafId}\nnetwork: ${networkName}(${chainId})\ntoken: ${
```
nicholaspai, line 1206: nit: do we need both network name and chainId? I think chain ID is pretty easy to understand, right? Network name adds unneccessary verbosity

nicholaspai (approved):
LGTM but I don't agree with adding chain name, i think everyone knows which chain IDs correspnd to which networks

src/dataworker/Dataworker.ts
kevinuma: I undid the network name change. It's only doing token symbol now.

mrice32 (approved):
+1!

PR #304, opened 2022-09-16
feat(relayer): Improve log for invalid fills and always log them

nicholaspai:
The motivation for this is:

1) We should reduce number of configuration variables where possible
2) The specific "invalid fills found" log can be summarized by grouping invalid fills per relayer rather than logging once per invalid fill found
3) I believe we should always log invalid fills found. If this gets super noisy as number of external relayers goes up we can look into changing this logic, but for now we should be aware of invalid fills sent

I think 3 is most contentious point, open to suggestions!


kevinuma (approved):

mrice32 (approved):
I agree! I think invalid fills should be almost non existent if everything is working properly, so having a notification about it, even if noisy is good.

PR #305, opened 2022-09-16
feat: Ignore "Already filled" revert reason

nicholaspai:
We already ignore expected revert reasons like "relay filled" because the Relayer is expected to run very frequently. Similarly, some dataworker reverts can happen if the dataworker runs quickly relative to its total run time.

The "already claimed" error will occur if the Dataworker executor is running at a
practical cadence of ~10 mins per run, because it takes ~5-7 mins per run, these runs can overlap and execution collisions can occur. These are non critical errors we can ignore to filter out the noise.

This PR also refactors the Multicaller logic a bit. A follow up PR should add tests, see [this issue](https://linear.app/uma/issue/ACX-156/implement-unit-test-for-multicallerclient-class)

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -66,12 +66,12 @@ export class MultiCallerClient {
       // Note: Check for exact revert reason instead of using .includes() to partially match reason string in order
       // to not ignore errors thrown by non-contract reverts. For example, a NodeJS error might result in a reason
       // string that includes more than just the contract revert reason.
-      const canIgnoreRevertReason = (reason: string) => reason === "relay filled" || reason === "Already claimed";
+      const canIgnoreRevertReasons = new Set(["relay filled", "Already claimed"])
       const transactionRevertsToIgnore = _transactionsSucceed.filter(
-        (txn) => !txn.succeed && canIgnoreRevertReason(txn.reason)
+        (txn) => !txn.succeed && canIgnoreRevertReasons.has(txn.reason)
       );
       const transactionRevertsToLog = _transactionsSucceed.filter(
-        (txn) => !txn.succeed && !canIgnoreRevertReason(txn.reason)
+        (txn) => !txn.succeed && !canIgnoreRevertReasons.has(txn.reason)
       );
       if (transactionRevertsToIgnore.length > 0)
         this.logger.debug({
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -63,12 +63,12 @@ export class MultiCallerClient {
       // execution collisions can occur. These are non critical errors we can ignore to filter out the noise.
       // TODO: Figure out less hacky way to reduce these errors rather than ignoring them.
       const canIgnoreRevertReason = (reason: string) => reason === "relay filled" || reason === "Already claimed";
-      const transactionsToIgnore = _transactionsSucceed.filter(
+      const transactionRevertsToIgnore = _transactionsSucceed.filter(
         (txn) => !txn.succeed && canIgnoreRevertReason(txn.reason)
       );
       this.logger.debug({
         at: "MultiCallerClient",
-        message: `Filtering out ${transactionsToIgnore.length} relay transactions that will fail because the relay has already been filled`,
+        message: `Filtering out ${transactionRevertsToIgnore.length} relay transactions that will fail because the relay has already been filled`,
         totalTransactions: _transactionsSucceed.length,
       });
       const transactionsSucceed = _transactionsSucceed.filter(
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -62,41 +62,45 @@ export class MultiCallerClient {
       // practical cadence of ~10 mins per run, because it takes ~5-7 mins per run, these runs can overlap and
       // execution collisions can occur. These are non critical errors we can ignore to filter out the noise.
       // TODO: Figure out less hacky way to reduce these errors rather than ignoring them.
+
+      // Note: Check for exact revert reason instead of using .includes() to partially match reason string in order
+      // to not ignore errors thrown by non-contract reverts. For example, a NodeJS error might result in a reason
+      // string that includes more than just the contract revert reason.
       const canIgnoreRevertReason = (reason: string) => reason === "relay filled" || reason === "Already claimed";
       const transactionRevertsToIgnore = _transactionsSucceed.filter(
         (txn) => !txn.succeed && canIgnoreRevertReason(txn.reason)
       );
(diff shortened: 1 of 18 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -58,17 +58,21 @@ export class MultiCallerClient {
...
-        message: `Filtering out ${
-          _transactionsSucceed.filter((txn) => !txn.succeed && txn.reason === "relay filled").length
-        } relay transactions that will fail because the relay has already been filled`,
+        message: `Filtering out ${transactionRevertsToIgnore.length} relay transactions that will fail because the relay has already been filled`,
         totalTransactions: _transactionsSucceed.length,
-        relayFilledReverts: _transactionsSucceed.filter((txn) => !txn.succeed && txn.reason === "relay filled").length,
       });
       const transactionsSucceed = _transactionsSucceed.filter(
-        (transaction) => transaction.succeed || transaction.reason !== "relay filled"
+        (transaction) => transaction.succeed || !canIgnoreRevertReason(transaction.reason)
```
nicholaspai, line 75: Can someone double check that this negation is equivalent to previous?

src/clients/MultiCallerClient.ts
```diff
@@ -58,17 +58,21 @@ export class MultiCallerClient {
 
       // Filter out transactions that revert for expected reasons. For example, the "relay filled" error
       // will occur frequently if there are multiple relayers running at the same time because only one relay
-      // can go through. This is a non critical error we can ignore to filter out the noise.
+      // can go through. Similarly, the "already claimed" error will occur if the dataworker executor is running at a
+      // practical cadence of ~10 mins per run, because it takes ~5-7 mins per run, these runs can overlap and
+      // execution collisions can occur. These are non critical errors we can ignore to filter out the noise.
+      // TODO: Figure out less hacky way to reduce these errors rather than ignoring them.
+      const canIgnoreRevertReason = (reason: string) => reason === "relay filled" || reason === "Already claimed";
```
daywiss, line 65: might be a bit more robust to use `reason.includes`

src/clients/MultiCallerClient.ts
daywiss: i think you need && or if you want to be really sure `!(!txn.succeed && canIgnoreRevertReason(txn.reason))`

src/clients/MultiCallerClient.ts
nicholaspai: So I might disagree here because the `reason` is a smart contract revert string, so there is a small chance of that revert reason changing, and if the reason isn't exactly the full revert string, then there's a different issue that might not be caught by `.includes()`. For example, m aybe there's some sort of other provider issue where the revert reason is included but the error is not coming from just calling a smart contract function that will revert

daywiss: ok sounds good

daywiss (approved):

src/clients/MultiCallerClient.ts
nicholaspai: Added a note on this for future dev

src/clients/MultiCallerClient.ts
```diff
@@ -58,41 +58,49 @@ export class MultiCallerClient {
...
-      // can go through. This is a non critical error we can ignore to filter out the noise.
-      this.logger.debug({
-        at: "MultiCallerClient",
-        message: `Filtering out ${
-          _transactionsSucceed.filter((txn) => !txn.succeed && txn.reason === "relay filled").length
-        } relay transactions that will fail because the relay has already been filled`,
-        totalTransactions: _transactionsSucceed.length,
-        relayFilledReverts: _transactionsSucceed.filter((txn) => !txn.succeed && txn.reason === "relay filled").length,
-      });
-      const transactionsSucceed = _transactionsSucceed.filter(
```
nicholaspai, line 70: I refactored this whole code block a bit. Currently, we later filter the `transactionsSucceed` list but I think we can be clearer up front and group transactions that will revert versus those that will succeed before we enter the logger functions

src/clients/MultiCallerClient.ts
nicholaspai: I've actually refactored this whole chunk of code to be more readable PTAL

src/clients/MultiCallerClient.ts
```diff
@@ -58,41 +58,49 @@ export class MultiCallerClient {
...
-        (transaction) => transaction.succeed || transaction.reason !== "relay filled"
+      // can go through. Similarly, the "already claimed" error will occur if the dataworker executor is running at a
+      // practical cadence of ~10 mins per run, because it takes ~5-7 mins per run, these runs can overlap and
+      // execution collisions can occur. These are non critical errors we can ignore to filter out the noise.
+      // TODO: Figure out less hacky way to reduce these errors rather than ignoring them.
+
+      // Note: Check for exact revert reason instead of using .includes() to partially match reason string in order
+      // to not ignore errors thrown by non-contract reverts. For example, a NodeJS error might result in a reason
+      // string that includes more than just the contract revert reason.
+      const canIgnoreRevertReason = (reason: string) => reason === "relay filled" || reason === "Already claimed";
```
mrice32, line 69: nit: if we want this to be easy and clear to add to, we might consider creating a constant Set that we check for inclusion rather than using an if chain.

mrice32 (approved):
LGTM!

src/clients/MultiCallerClient.ts
nicholaspai: great point, implemented. ptal as I've never used a Set before

src/clients/MultiCallerClient.ts
```diff
@@ -58,41 +58,49 @@ export class MultiCallerClient {
...
-        (transaction) => transaction.succeed || transaction.reason !== "relay filled"
+      // can go through. Similarly, the "already claimed" error will occur if the dataworker executor is running at a
+      // practical cadence of ~10 mins per run, because it takes ~5-7 mins per run, these runs can overlap and
+      // execution collisions can occur. These are non critical errors we can ignore to filter out the noise.
+      // TODO: Figure out less hacky way to reduce these errors rather than ignoring them.
+
+      // Note: Check for exact revert reason instead of using .includes() to partially match reason string in order
+      // to not ignore errors thrown by non-contract reverts. For example, a NodeJS error might result in a reason
+      // string that includes more than just the contract revert reason.
+      const canIgnoreRevertReasons = new Set(["relay filled", "Already claimed"]);
```
mrice32, line 69: Usage all looks great! Since it's just a set of constants, I would move it out of the function so it only gets created once.

nicholaspai: done

mrice32 (approved):
LGTM!

PR #307, opened 2022-09-21
improve: Ignore "missing revert data..." reason

nicholaspai:
This PR ignores the following revert reason:

> "missing revert data in call exception; Transaction reverted without a reason string"

Ignoring this reason potentially includes false positives of reverts that we should be alerted on, however there is something likely broken in how the provider is interpreting contract reverts. Currently, there are a lot of duplicate transaction sends that are reverting with this reason, for example, sending a transaction to execute a relayer refund leaf takes a while to execute and ends up reverting because a duplicate transaction mines before it. This situation leads to this revert reason which is spamming the Logger currently.

So, this PR reduces alert noise at the cost of introducing false positive ignored errors

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -27,7 +27,16 @@ export interface AugmentedTransaction {
 // .includes() to partially match reason string in order to not ignore errors thrown by non-contract reverts.
 // For example, a NodeJS error might result in a reason string that includes more than just the contract r
 // evert reason.
-const canIgnoreRevertReasons = new Set(["relay filled", "Already claimed"]);
+const canIgnoreRevertReasons = new Set([
+  "relay filled",
+  "Already claimed",
+  // The following reason potentially includes false positives of reverts that we should be alerted on, however
+  // there is something likely broken in how the provider is interpreting contract reverts. Currently, there are
+  // a lot of duplicate transaction sends that are reverting with this reason, for example, sending a transaction
+  // to execute a relayer refund leaf takes a while to execute and ends up reverting because a duplicate transaction
+  // mines before it. This situation leads to this revert reason which is spamming the Logger currently.
+  "missing revert data in call exception; Transaction reverted without a reason string",
+]);
 
 export class MultiCallerClient {
   private transactions: AugmentedTransaction[] = [];
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -43,16 +43,19 @@ const unknownRevertReasonMethodsToIgnore = new Set([
   "executeRootBundle",
 ]);
 
+// Ignore the general unknown revert reason for specific methods or uniformly ignore specific revert reasons
+// for any contract method.
 const canIgnoreRevertReasons = (obj: {
   succeed: boolean;
   reason: string;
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -27,7 +27,16 @@ export interface AugmentedTransaction {
...
-const canIgnoreRevertReasons = new Set(["relay filled", "Already claimed"]);
+const canIgnoreRevertReasons = new Set([
+  "relay filled",
+  "Already claimed",
+  // The following reason potentially includes false positives of reverts that we should be alerted on, however
+  // there is something likely broken in how the provider is interpreting contract reverts. Currently, there are
+  // a lot of duplicate transaction sends that are reverting with this reason, for example, sending a transaction
+  // to execute a relayer refund leaf takes a while to execute and ends up reverting because a duplicate transaction
+  // mines before it. This situation leads to this revert reason which is spamming the Logger currently.
+  "missing revert data in call exception; Transaction reverted without a reason string",
```
nicholaspai, line 38: Originally I wanted to read the transaction's `message` to filter out Dataworker execution logic speciifcally but I noticed its also happening with the relayer so I think its best to blanket ignore for now until we fix the provider

chrismaree (approved):

src/clients/MultiCallerClient.ts
chrismaree: could this not introduce other cases where the bot incorrectly ignores valid errors?

daywiss (approved):
fyi when we switch to the pd v2 api we have the option of filtering this in PD which is nice since it wont need a pr

src/clients/MultiCallerClient.ts
mrice32: I have mixed feelings on this. @nicholaspai to @chrismaree's point, do you think we should specifically try to single out relay fills and dataworker executions, so that we still get alerts on failed disputes and proposals, for instance?

mrice32 (commented):
Thank you for working on this @nicholaspai -- this is awesome!

src/clients/MultiCallerClient.ts
nicholaspai: Hm fair point. I need to refactor this a bit

src/clients/MultiCallerClient.ts
```diff
@@ -27,7 +27,37 @@ export interface AugmentedTransaction {
...
+]);
+
+// Ignore the general unknown revert reason for specific methods or uniformly ignore specific revert reasons
+// for any contract method.
+const canIgnoreRevertReasons = (obj: {
+  succeed: boolean;
+  reason: string;
+  transaction: AugmentedTransaction;
+}): boolean => {
+  // prettier-ignore
```
nicholaspai, line 53: I need to prettier ignore this because prettier doesn't seem to understand boolean algebra and keeps wanting to remove the parentheses after the && on line 55

src/clients/MultiCallerClient.ts
nicholaspai: I added some logic to only ignore this revert reason for specific contract methods

mrice32 (approved):
LGTM!!

PR #308, opened 2022-09-21
improve(ProviderUtils): Avoid redundant eth_chainId rpc calls

pxrl:
Since ethers v5, all RPC requests have been preceed by an eth_chainId lookup. This is described by ethers as a safety feature to mitigate being "RPC rugged" - i.e. where a wallet silently changes RPC without notifying the provider.

Inspecting the Across logs, eth_chainId is the second-most popular RPC call, after only eth_getLogs. Over the past 30 days, Infura usage has been:

```
   eth_getLogs                264M
   eth_chainId                 30M
   eth_call                    24M
   eth_getTransactionReceipt   11M
   eth_blockNumber              3M
 
   Total                      336M
```

Given that the Across bots maintain a 1:1 relationship between provider instance and back-end RPC provider, there's no obvious way for the chainId to ever change. It's therefore worth considering using StaticJsonRpcProvider instead, since it's intended for exactly this scenario.

It should be noted that the 30 day figures might be lower than normal due to downtime for both Arbitrum Nitro and the merge. In any case, migrating to StaticJsonRpcProvider would reduce total requests by about 9% based on these figures, and would predominantly help reduce latency in the bots.

See also: https://github.com/ethers-io/ethers.js/issues/901

Ref: ACX-67

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -15,9 +15,13 @@ interface RateLimitTask {
   reject: (err: any) => void;
 }
 
-// This provider is a very small addition to the JsonRpcProvider that ensures that no more than `maxConcurrency`
+// StaticJsonRpcProvider is used in place of JsonRpcProvider to avoid redundant eth_chainId queries prior to each
+// request. This is safe to use when the back-end provider is guaranteed not to change.
+// See https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#StaticJsonRpcProvider
+
+// This provider is a very small addition to the StaticJsonRpcProvider that ensures that no more than `maxConcurrency`
 // requests are ever in flight. It uses the async/queue library to manage this.
-class RateLimitedProvider extends ethers.providers.JsonRpcProvider {
+class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
This change makes a lot of sense, `StaticJsonRpcProvider` is awesome! How do you plan to track network usage post merging this PR to determine its efficacy?

pxrl:
> This change makes a lot of sense, `StaticJsonRpcProvider` is awesome! How do you plan to track network usage post merging this PR to determine its efficacy?

We have reasonably good reporting via Infura so I was planning to keep an eye on the usage dashboard. Alchemy's reporting seems even better and more fine-grained, but we're using them less.

If you have any other suggestions for where else to look then I'm happy to follow up on that too!

nicholaspai:
> > This change makes a lot of sense, `StaticJsonRpcProvider` is awesome! How do you plan to track network usage post merging this PR to determine its efficacy?
> 
> We have reasonably good reporting via Infura so I was planning to keep an eye on the usage dashboard. Alchemy's reporting seems even better and more fine-grained, but we're using them less.
> 
> If you have any other suggestions for where else to look then I'm happy to follow up on that too!

Ah yes good point on watching Infura usage, LGTM!

PR #312, opened 2022-09-23
script: Add a script to reconstruct bundles and find mismatches

kevinuma:
To run the script:

1. Set the desired number of bundles to examine in .env (First n bundles starting from the start of Across). For example:
DATAWORKER_FAST_LOOKBACK_COUNT=100
DATAWORKER_FAST_START_BUNDLE=100
2. Run ts-node src/scripts/investigate.ts --wallet mnemonic
The mnemonic doesn't matter as the script doesn't send any txs.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,7 @@ module.exports = {
     mocha: true,
     node: true,
   },
-  plugins: ["node", "prettier", "@typescript-eslint", "mocha"],
+  plugins: ["node", "prettier", "@typescript-eslint", "mocha", "chai-expect"],
   extends: [
     "plugin:prettier/recommended",
     "eslint:recommended",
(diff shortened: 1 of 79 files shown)


.eslintrc.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
   },
   overrides: [
     {
-      files: ["scripts/*.ts", "tasks/*.ts"],
+      files: ["src/scripts/*.ts", "scripts/*.ts", "tasks/*.ts"],
```
kevinuma, line 43: To skip lint errors on console.logs in scripts

mrice32 (commented):
This looks great on first pass -- I don't see any high-level issues with the logic flow. The code is pretty dense (which is understandable for this sort of work!), so I will try to take another pass to see if I can identify any operations that seem error prone.

nicholaspai:
I'm not sure we want to actually merge this into `master` because its going to get out of date quickly as Dataworker logic changes and we don't want to have to keep updating it. There's also a lot of hardcoded variables in here. I think its worth reviewing to get one same page about logic but not sure about how to proceed from here.

I also don't think we came away with any action items out of this besides "we don't know what we don't know" and we decided to go with an empirical observation of what the spoke pool deficits were:

1. We see leaf is stuck
2. We see how much token leaf is trying to withdraw from spokepool
3. Compare above amount to current spoke pool balance
4. This is an approximation of the deficit
5. Repeat steps 1-4 after every bundle is validated and pool leaves are executed, and see if deficits are stable

Currently we saw there was a WBTC deficit of 0.03 and 0.06 on Arbitrum and Optimism, so we donated those amounts to the spoke pools, and there hasn't been a leaf stuck since then

PR #315, opened 2022-09-23
improve: improve typing in src/clients/bridges/

mrice32:
This is the first PR in what will hopefully be a series of PRs to address type issues that are blocking us from enabling the `strict` setting in our tsconfig.

The idea behind this change is to make our bots safer and more error-proof by expanding the sort of type-checking we're getting from typescript.

I've created this by enabling strict locally and working through the errors manually. Since there are hundreds of errors, I've decided to start working through them folder-by-folder and doing PRs as I go so these can be carefully reviewed.

### DIFF
--- a/src/clients/bridges/PolygonAdapter.ts
+++ b/src/clients/bridges/PolygonAdapter.ts
@@ -123,7 +123,7 @@ export class PolygonAdapter extends BaseAdapter {
 
         const l2Token = this.getL2Token(l1Token);
 
-        if (l2Token == null) continue;
+        if (l2Token === null) continue;
         if (!this.isSupportedToken(l1Token)) continue;
 
         const l1Method = tokenToBridge[l1Token].l1Method;
--- a/src/clients/bridges/PolygonAdapter.ts
+++ b/src/clients/bridges/PolygonAdapter.ts
@@ -1,8 +1,19 @@
-import { runTransaction, assign, Contract, BigNumber, bnToHex, winston } from "../../utils";
+import {
+  runTransaction,
+  assign,
+  Contract,
+  BigNumber,
+  bnToHex,
+  winston,
+  Event,
+  isDefined,
+  BigNumberish,
+} from "../../utils";
 import { ZERO_ADDRESS, spreadEventWithBlockNumber, paginatedEventQuery, Promise } from "../../utils";
 import { SpokePoolClient } from "../../clients";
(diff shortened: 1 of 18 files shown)


src/clients/bridges/PolygonAdapter.ts
```diff
@@ -110,14 +123,17 @@ export class PolygonAdapter extends BaseAdapter {
 
         const l2Token = this.getL2Token(l1Token);
 
+        if (l2Token == null) continue;
```
kevinuma, line 126: nit: l2Token === null

kevinuma (approved):

src/clients/bridges/PolygonAdapter.ts
mrice32: Nice catch!

mrice32: Fixed!

kevinuma (approved):

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -182,46 +201,52 @@ export class PolygonAdapter extends BaseAdapter {
...
-        this.log("Could not construct l1Bridge. Likely misconfiguration", { l1Token, error, tokenToBridge }, "error");
-        return null;
-      }
+    else return new Contract(l1RootChainManager, polygonL1RootChainManagerInterface, this.getSigner(1));
   }
 
   // Note that on polygon we dont query events on the L2 bridge. rather, we look for mint events on the L2 token.
   getL2Token(l1Token: string): Contract | null {
-    try {
+    if (this.isSupportedToken(l1Token))
```
nicholaspai, line 233: Doesn't converting from a try-catch to an if-else change behavior?

src/utils/TypeGuards.ts
```diff
@@ -9,3 +9,7 @@ export function isPromiseRejected<T>(
 ): promiseSettledResult is PromiseRejectedResult {
   return promiseSettledResult.status === "rejected";
 }
+
+export function isDefined<T>(input: T | null | undefined): input is T {
```
nicholaspai, line 13: +1

nicholaspai (approved):
LGTM

src/clients/bridges/PolygonAdapter.ts
mrice32: Yep, I think this is a behavior change.

Based on the error message, I assumed all we were catching was an unsupported token, though? Or are there other cases we were trying to catch here?

PR #317, opened 2022-09-24
refactor: Speed up Dataworker by optmistically using fixed lookback window

nicholaspai:


src/dataworker/Dataworker.ts
```diff
@@ -655,16 +655,17 @@ export class Dataworker {
 
     await Promise.all(
       Object.entries(spokePoolClients).map(async ([chainId, client]) => {
-        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter(
-          (rootBundle) =>
-            rootBundle.slowRelayRoot !== EMPTY_MERKLE_ROOT &&
-            (IGNORED_SPOKE_BUNDLES[chainId] ? !IGNORED_SPOKE_BUNDLES[chainId].includes(rootBundle.rootBundleId) : true)
+        let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter((rootBundle) =>
```
nicholaspai, line 658: This is actually a separate refactor thats important: The point of `spokeRootsLookbackCount` is to look at only latest few proposed bundles, however the ordering of this `filter` is producing unexpected behavior. Currently we filter out all bundles with empty slow roots before we `.slice()` which means we end up looking at much older root bundles if all of the recent slow roots were 0x0. We should first `.slice()` and then `filter` out empty roots

src/common/ClientHelper.ts
```diff
@@ -21,36 +21,6 @@ export function getSpokePoolSigners(baseSigner: Wallet, config: CommonConfig): {
   );
 }
 
-export async function constructSpokePoolClientsForBlockAndUpdate(
```
nicholaspai, line 24: Legacy function that we don't need anymore, deprecating in favor of the much more understandable `constructSpokePoolClientsWithLookback`

src/dataworker/Dataworker.ts
```diff
@@ -491,14 +524,15 @@ export class Dataworker {
...
           at: "Dataworker#validate",
-          message: "A bundle end block is > latest block but within buffer, skipping",
+          message: "Cannot validate because a bundle end block is > latest block but within buffer",
           expectedEndBlocks: widestPossibleExpectedBlockRange.map((range) => range[1]),
           pendingEndBlocks: rootBundle.bundleEvaluationBlockNumbers,
           endBlockBuffers,
         });
       }
       return {
-        valid: true,
```
nicholaspai, line 501: I think its more accurate to make this return `false` since we should assume all proposals are invalid unless proven otherwise but we can use the above `invalidReasonsToSkipDispute` list to filter on this invalid reason and not submit an actual dispute

src/dataworker/Dataworker.ts
```diff
@@ -389,21 +402,41 @@ export class Dataworker {
       pendingRootBundle,
       spokePoolClients
     );
+    const invalidReasonsToSkipDispute = new Set(["insufficient-dataworker-lookback", "bundle-end-block-buffer"]);
```
mrice32, line 405: nit: can we move this outside of the function since it will never change?

src/dataworker/Dataworker.ts
mrice32: Agreed, that's probably more clear.

src/dataworker/Dataworker.ts
```diff
@@ -389,21 +402,41 @@ export class Dataworker {
...
+              "Cannot validate bundle with some chain's startBlock < client start block. Set a larger DATAWORKER_FAST_LOOKBACK",
+            clientStartBlocks: Object.values(spokePoolClients).map((client) => client.eventSearchConfig.fromBlock),
+          });
+        }
+        // Catch-all debug level alert for known invalid reasons.
+        this.logger.debug({
+          at: "Dataworker",
+          message: "Skipping dispute 😪",
+          mrkdwn: reason,
+        });
```
mrice32, line 423: This seems a little dangerous. If we were to add something to that list but we forgot to add a special if statement above it would be ignored.

Could we have 2 sets: one that errors and one that is ignored. So the code might be structured more like this:
```ts
if (ignoreDisputeReasons.has(reason)) {
  // just debug log
  logger.debug....
} else if (errorDisputeReasons.has(reason)) {
  // send error
  logger.error...
} else {
  // send dispute
}
```

Thoughts?

src/dataworker/Dataworker.ts
mrice32: Great point!

src/dataworker/Dataworker.ts
```diff
@@ -714,6 +762,24 @@ export class Dataworker {
...
           });
 
+          if (
+            blockRangesAreInvalidForSpokeClients(
+              spokePoolClients,
+              blockNumberRanges,
+              this.chainIdListForBundleEvaluationBlockNumbers
+            )
+          ) {
+            this.logger.debug({
```
mrice32, line 772: In this case, we basically don't try to execute because we don't have the data, right? Do you think this is worth a warning? Do we consider it a misconfig if our root lookback is greater than our spoke block lookback?

src/dataworker/Dataworker.ts
```diff
@@ -515,29 +549,40 @@ export class Dataworker {
...
+      this.logger.debug({
+        at: "Dataworker#validate",
+        message:
+          "Cannot validate bundle with some chain's startBlock < client start block. Set a larger DATAWORKER_FAST_LOOKBACK",
+        rootBundleRanges: blockRangesImpliedByBundleEndBlocks,
+        clientStartBlocks: Object.values(spokePoolClients).map((client) => client.eventSearchConfig.fromBlock),
+      });
+      return {
+        valid: false,
+        reason: "insufficient-dataworker-lookback",
```
mrice32, line 570: I think this is a great call -- the disputer just blows up rather than actually trying to iteratively look back.

src/utils/FillUtils.ts
```diff
@@ -188,3 +188,20 @@ export function getUnfilledDeposits(
...
   return unfilledDeposits;
 }
+
+export function allFillsHaveMatchingDeposits(spokePoolClients: SpokePoolClientsByChain): boolean {
+  for (const originChainId of Object.keys(spokePoolClients)) {
+    for (const destChainId of Object.keys(spokePoolClients)) {
+      if (originChainId === destChainId) continue;
+      const foundUnmatchedFill = spokePoolClients[destChainId]
+        .getFillsForOriginChain(Number(originChainId))
+        .filter((fill) => fill.totalFilledAmount.eq(fill.amount) && !fill.fillAmount.eq(fill.amount))
```
mrice32, line 198: The method name implies that we're finding _all fills_, but this looks like it's only looking for fills that complete a partial fill? Can you explain (and comment) what this method is searching for, precisely?

I also don't quite understand why we don't care about typical single-shot fills. Why aren't they important here?

mrice32 (commented):
Looks great! A few comments

src/dataworker/Dataworker.ts
nicholaspai: yes done

src/dataworker/Dataworker.ts
nicholaspai: so I'm kind of torn here. I can see this getting noisy especially if block times begin to vary a lot, but its probably better to get warnings in this case and expand the lookback. I'm going to change to warn +1

src/dataworker/Dataworker.ts
nicholaspai: yeah its really important we don't send erroneous disputes

src/utils/FillUtils.ts
nicholaspai: Hm yeah I think you're right, we should just care about all fills. This is going to make the lookback algo a bit tougher though since there's probably always going to be some fill near the `fromBlock` without a deposit on the other chain

src/common/Constants.ts
```diff
@@ -4,13 +4,22 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
...
 
-// Target ~2 days per chain. Avg. block times: { 1: 15s, 10/42161: 0.5s, 137: 2.5s, 288: 30s }
+// Target ~2 days per chain. Avg. block times: { 1: 12s, 10/42161: 0.5s, 137: 2.5s, 288: 30s }
 export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
-  1: 11500,
-  10: 350000,
-  137: 70000,
-  288: 6000,
-  42161: 350000,
+  1: 14400,
```
kevinuma, line 9: Why changed these numbers?

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -75,3 +83,37 @@ export function spokePoolClientsToProviders(spokePoolClients: { [chainId: number
...
+      "RelayedRootBundle",
+      "ExecutedRelayerRefundRoot",
+    ]
+    // Don't use the cache for the quick lookup so we don't load and parse unneccessary events from Redis DB
+    // that we'll throw away if the below checks succeed.
+  );
+
+  // Validate data loaded from clients. Every partial fill that completed a deposit in the event search window must
+  // match with a deposit in the search window.
+  if (allFillsHaveMatchingDeposits(spokePoolClients)) return spokePoolClients;
```
kevinuma, line 117: nit: Why not return allFillsHaveMatchingDeposits(spokePoolClients) ? spokePoolClients : undefined;

Also is undefined the right value to return for the empty case instead of {} ?

kevinuma (approved):

src/utils/FillUtils.ts
nicholaspai: I replaced this function with a new algorithm, will explain below

nicholaspai:
I have updated the algorithm a bit as I think its a lot better to control the lookback. Previously the user would set the lookback as a block count on a per-chain basis. This is hard to control and keep block lookbacks in sync across chains. The new algorithm uses a validated root bundle lookback count and the dataworker infers the lookback from the bundle evaluation blocks from that Nth latest root bundle.

Secondly, the decision about whether to skip validating or proposing a bundle was convoluted. Previously it would skip a root bundle if the start block for that bundle was earlier than the spoke client's earliest event search from block. Now, it skips the bundle if the bundle's start block for a chain is earlier than the earliest fill event on the chain's spoke client that can be matched with a deposit event.

This allowed me to remove the (retry + increase lookback) logic. The new behavior is that the dataworker in "fast" mode has a limited lookback and will refuse to validate bundles if a bundle start block is earlier than the "earliest matched fill" block for the same chain. This does mean that the dataworker needs to loudly alert in these cases so the dataworker user knows that they have misconfigured the bot.

src/utils/FillUtils.ts
```diff
@@ -188,3 +196,26 @@ export function getUnfilledDeposits(
 
   return unfilledDeposits;
 }
+
+// Returns set of block numbers, keyed by chain ID, that represent the earliest block number for
+// that chain where a fill event matches a deposit event on its origin chain. This dictionary can be used
+// by the dataworker to conveniently skip root bundles it won't be able to construct roots for.
+export function getEarliestMatchedFillBlocks(spokePoolClients: SpokePoolClientsByChain): { [chainId: number]: number } {
```
mrice32, line 203: I could be misunderstanding, so please feel free to correct me.

I'm a little confused as to why the earliest matched fill should inform whether we can support a block number for the chain _with that fill_. Wouldn't we want the _latest umatched_ fill? Otherwise, couldn't you have a fill in block 1 that is matched, but fill in block 2 that is unmatched and still be willing to validate a bundle that includes blocks 1 and 2, despite not having enough information to validate the fill in block 2.

As a secondary note: let's say there's a partial fill in block 1 and the fill is completed by a relayer in block 100. Let's say the the partial fill triggered a slow relay. The deposit that triggered these fills is in range, but the initial partial fill is not in range. Are we still able to validate the "pull back" of capital that's supposed to happen in the new bundle without being able to see the initial partial fill? I think we are, but I wanted to double check.

mrice32 (commented):
Looks awesome! I really like this strategy. One high-level comment.

src/utils/FillUtils.ts
```diff
@@ -188,3 +197,31 @@ export function getUnfilledDeposits(
...
+      // block is manipulated too high.
+      const latestUnmatchedFill = _.findLast(
+        spokePoolClients[destChainId].getFillsForOriginChain(Number(originChainId)),
+        (fill) => fill.depositId < spokePoolClients[originChainId].earliestDepositId
+      );
+
+      // If all fills can be matched for this destination chain, then the latest unmatched fill block should be
+      // 0 so that the downstream Dataworker can validate all roots.
+      if (latestUnmatchedFill !== undefined) blocks[destChainId] = latestUnmatchedFill.blockNumber;
+      else blocks[destChainId] = 0;
```
nicholaspai, line 223: I just realized that this is a bug. It should default to the earliest fills block number instead of 0. Imagine the lookback is very short but also all fills in it are matched with deposits. We should count any blocks greater than the earliest fill as validatable but not older than the earliest fill block.

nicholaspai (commented):
Found one bad bug and writing this as a reminder to self

src/utils/FillUtils.ts
nicholaspai: This is a good point, after discussing offline we agreed that there could be a big difference between the `earliestMatchedFill` and the `latestUnmatchedFill`. I've implemented the latestUnmatchedFill logic and added a note that this introduces a potential DOS vector that a malicious relayer could execute by sending invalid fills with low deposit ID's

src/common/ClientHelper.ts
```diff
@@ -101,6 +85,30 @@ export async function constructSpokePoolClientsWithLookback(
...
+    return { networkId, contract: getDeployedContract("SpokePool", networkId, spokePoolSigners[networkId]) };
+  });
+
+  // If no lookback is set, fromBlock will be set to spoke pool's deployment block.
+  const fromBlocks = {};
+  spokePools.forEach((obj: { networkId: number; contract: Contract }) => {
+    if (startBlockOverride[obj.networkId]) {
+      fromBlocks[obj.networkId] = startBlockOverride[obj.networkId];
+    }
+  });
```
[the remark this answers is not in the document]
nicholaspai: done

src/clients/HubPoolClient.ts
```diff
@@ -258,6 +258,16 @@ export class HubPoolClient {
     });
   }
 
+  getNthFullyExecutedRootBundle(n: number): ProposedRootBundle | undefined {
```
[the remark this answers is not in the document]
nicholaspai: done, i've blocked n == 0

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -75,3 +84,87 @@ export function spokePoolClientsToProviders(spokePoolClients: { [chainId: number
...
+  config: DataworkerConfig,
+  clients: DataworkerClients,
+  dataworker: Dataworker
+) {
+  const toBundle =
+    config.dataworkerFastStartBundle === "latest"
+      ? undefined
+      : clients.hubPoolClient.getNthFullyExecutedRootBundle(Number(config.dataworkerFastStartBundle));
+  const fromBundle =
+    config.dataworkerFastLookbackCount === config.dataworkerFastStartBundle
```
mrice32, line 129: This comparison is a bit surprising? Why would the fast lookback count ever be equal to the start bundle...? I may be missing something.

mrice32 (commented):
Looks great! Just a few questions. I'm super excited about this PR!

src/dataworker/DataworkerUtils.ts
```diff
@@ -45,6 +59,55 @@ export function getBlockRangeForChain(
...
+      Number(chainId),
+      chainIdListForBundleEvaluationBlockNumbers
+    );
+    const clientLastBlockQueried =
+      spokePoolClients[chainId].eventSearchConfig.toBlock ?? spokePoolClients[chainId].latestBlockNumber;
+    const bundleRangeToBlock = blockRangeForChain[1];
+
+    // If `latestInvalidBundleStartBlock` is not defined for chain, then block range is valid for chain by default
+    // unless the spoke client hasn't queried the end block yet.
+    if (latestInvalidBundleStartBlock[chainId] === undefined) return bundleRangeToBlock > clientLastBlockQueried;
```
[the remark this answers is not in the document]
nicholaspai: Good point I'll remove. I was being overly general with this function

src/scripts/validateRootBundle.ts
```diff
@@ -60,30 +72,51 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet) {
...
-      "RelayedRootBundle",
-      "ExecutedRelayerRefundRoot",
-    ],
-    bundleEndBlockMapping
+    clients.configStoreClient,
+    config,
+    baseSigner,
+    fromBlocks,
+    toBlocks
+    // Don't use the cache for shorter lookback windows because loading and storing into RedisDB adds
```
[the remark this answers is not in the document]
nicholaspai: so I'm a bit conflicted. With the expected production configuration of a 16 root bundle lookback, the dataworker (proposer + disputer + executor) runs in <1 min without the cache and <5 mins with the cache. The slow down is loading the growing JSON Blob from the redis and parsing the blob into fill/deposit events. Unfortunately this operation will increase over time as the event history grows. 

So, I think this PR can be a first step towards deprecating the redis cache for use in the Dataworker (its still useful in the Relayer). Or, we should change what we store into the redis cache and expire values after a while.

src/dataworker/DataworkerClientHelper.ts
nicholaspai: The way it works now: imagine you set the start bundle equal to 10, and the lookback equal to 10. This would mean query all blocks, so I catch this case and set fromBlocks == undefined

src/scripts/validateRootBundle.ts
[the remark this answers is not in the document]
mrice32: That sounds great to me! +1 to not needing the redis cache. Just wanted to understand. Does any other bot use this part of the redis cache? (I know the relayer uses the individual keys for deposit fees)

There could be a world where a per-deposit cache could be a solution to a the low deposit id issue. For instance, if we notice a deposit id that we need to get info about, we could first check the cache, and then binary search event ranges until we find that deposit id (caching the deposits we find along the way).

Regardless, I think what you have here is great and we shouldn't change anything!

src/clients/HubPoolClient.ts
```diff
@@ -252,12 +255,52 @@ export class HubPoolClient {
...
+    return _.findLast(this.proposedRootBundles, (rootBundle: ProposedRootBundle) => {
       if (rootBundle.blockNumber > latestMainnetBlock) return false;
       return this.isRootBundleValid(rootBundle, latestMainnetBlock);
     });
   }
 
+  getEarliestFullyExecutedRootBundle(latestMainnetBlock: number, startBlock = 0): ProposedRootBundle | undefined {
+    return this.proposedRootBundles.find((rootBundle: ProposedRootBundle) => {
+      if (rootBundle.blockNumber > latestMainnetBlock) return false;
+      if (rootBundle.blockNumber < startBlock) return false;
```
pxrl, line 267: Minor uncertainty from my side: does returning boolean (false) comply with the (ProposedRootBundle | undefined) type annotation?

src/clients/HubPoolClient.ts
```diff
@@ -252,12 +255,52 @@ export class HubPoolClient {
...
+  getNthFullyExecutedRootBundle(n: number, startBlock?: number): ProposedRootBundle | undefined {
+    if (n === 0) throw new Error("n cannot be 0");
+
+    let bundleToReturn: ProposedRootBundle | undefined;
+
+    // If n is negative, then return the Nth latest executed bundle, otherwise return the Nth earliest
+    // executed bundle.
+    if (n < 0) {
+      let nextLatestMainnetBlock = startBlock ?? this.latestBlockNumber;
+      for (let i = 0; i < -n; i++) {
```
pxrl, line 285: Consider `i < Math.abs(n)` instead of `i < -n` ?

src/dataworker/DataworkerConfig.ts
```diff
@@ -83,8 +90,22 @@ export class DataworkerConfig extends CommonConfig {
...
+    // Set to 0 to load all events, but be careful as this will cause the Dataworker to take 30+ minutes to complete.
+    // The average bundle frequency is 4 bundles per day so 16 bundles is a reasonable default
+    // to lookback 4 days.
+    this.dataworkerFastLookbackCount = DATAWORKER_FAST_LOOKBACK_COUNT ? Number(DATAWORKER_FAST_LOOKBACK_COUNT) : 16;
+    this.dataworkerFastStartBundle = DATAWORKER_FAST_START_BUNDLE ? Number(DATAWORKER_FAST_START_BUNDLE) : "latest";
+    if (typeof this.dataworkerFastStartBundle === "number") {
+      assert(this.dataworkerFastStartBundle > 0, "dataworkerFastStartBundle > 0");
+      assert(
+        this.dataworkerFastStartBundle >= this.dataworkerFastLookbackCount,
+        "dataworkerFastStartBundle >= dataworkerFastLookbackCount"
```
pxrl, line 107: Minor QoL for anyone who ever encounters this - it might be nice to include the value that failed these assertions in the accompanying message.

src/dataworker/DataworkerClientHelper.ts
mrice32: Then should this be config.dataworkerFastLookbackCount >= config.dataworkerFastStartBundle? I assume it would be possible to also set this value to greater than the start bundle, no?

mrice32 (approved):
Looks good! This is really cool -- I'm really excited about this bringing down the resources required to run the dataworker. This will be really impactful.

mrice32:
Only remaining comment: https://github.com/across-protocol/relayer-v2/pull/317#discussion_r987373945

mrice32 (approved):
LGTM!

PR #318, opened 2022-09-26
chore: bump financial temp logging version

daywiss:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16
```
daywiss, line 1: this needs to be node 16 for superstruct dependency to work right ( this was already done in protocol)

nicholaspai (approved):

PR #319, opened 2022-09-28
improve: improve typing in src/clients/

mrice32:
This builds on #315 and makes the src/clients directory compatible with typescript strict.

As a part of this change, a few of the loops and data structures were refactored to make type safety easier to achieve and easier to reason about.

Note: this builds on #315, so its changes are also reflected here. Once that is merged, that part of the diff should disappear.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -1,7 +1,6 @@
 import { winston, BigNumber, toBN } from "../utils";
 import * as _ from "lodash";
 import {
-  Deposit,
   DepositWithBlock,
   FillsToRefund,
   FillWithBlock,
@@ -119,10 +118,9 @@ export class BundleDataClient {
       );
 
       for (const tokenAddress of Object.keys(allRefunds[chainId])) {
-        if (executedRefunds[tokenAddress] === undefined || allRefunds[chainId][tokenAddress].refunds === undefined)
-          continue;
-
         const refunds = allRefunds[chainId][tokenAddress].refunds;
+        if (executedRefunds[tokenAddress] === undefined || refunds === undefined) continue;
+
         for (const relayer of Object.keys(refunds)) {
           const executedAmount = executedRefunds[tokenAddress][relayer];
           if (executedAmount === undefined) continue;
@@ -229,7 +227,7 @@ export class BundleDataClient {
           .filter((fillWithBlock) => fillWithBlock.blockNumber <= blockRangeForChain[1])
           .forEach((fillWithBlock) => {
             // If fill matches with a deposit, then its a valid fill.
-            const matchedDeposit: Deposit = originClient.getDepositForFill(fillWithBlock);
+            const matchedDeposit = originClient.getDepositForFill(fillWithBlock);
             if (matchedDeposit) {
               // Fill was validated. Save it under all validated fills list with the block number so we can sort it by
               // time. Note that its important we don't skip fills earlier than the block range at this step because
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -56,7 +56,7 @@ export class BundleDataClient {
 
   getPendingRefundsFromValidBundles(bundleLookback: number): FillsToRefund[] {
     const refunds = [];
-    if (this.clients.hubPoolClient.latestBlockNumber === undefined)
+    if (!this.clients.hubPoolClient.isUpdated || this.clients.hubPoolClient.latestBlockNumber === undefined)
       throw new Error("BundleDataClient::getPendingRefundsFromValidBundles HubPoolClient not updated.");
 
     let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -56,6 +56,8 @@ export class BundleDataClient {
 
   getPendingRefundsFromValidBundles(bundleLookback: number): FillsToRefund[] {
     const refunds = [];
+    if (this.clients.hubPoolClient.latestBlockNumber === undefined)
+      throw new Error("BundleDataClient::getPendingRefundsFromValidBundles HubPoolClient not updated.");
 
     let latestBlock = this.clients.hubPoolClient.latestBlockNumber;
     for (let i = 0; i < bundleLookback; i++) {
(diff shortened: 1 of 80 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -420,23 +430,9 @@ export class SpokePoolClient {
...
-              originBlockNumber,
-            };
-            if (dataToCache?.deposits?.[destinationChainId])
-              this.deposits[destinationChainId][index] = {
-                ...speedUpDeposit,
-                blockNumber,
-                originBlockNumber,
-              };
-          }
+      for (const deposits of Object.values(this.deposits))
```
nicholaspai, line 433: This seems like a breaking change. First of all is `deposits[index] = x` equivalent to `this.deposits[destinationChainId][index] = x`? 

Also, you removed the dataToCache setting which I implemented correctly (I think) here: https://github.com/across-protocol/relayer-v2/pull/328

nicholaspai (commented):
This is a lot of great work and looks like errors will be caught earlier with unexpected `undefined` values. I wish this PR was smaller though its kind of hard to review

src/clients/SpokePoolClient.ts
mrice32: Yep, I removed it because it was a no-op! Once that's merged, I will merge in that update!

mrice32: > First of all is deposits[index] = x equivalent to this.deposits[destinationChainId][index] = x?

Yes, both of these reference the same array, so either way, you're modifying the same data structure. The only operation that wouldn't work would be replacing the array entirely.

src/clients/ConfigStoreClient.ts
```diff
@@ -46,22 +50,20 @@ export class AcrossConfigStoreClient {
...
+    readonly redisClient?: RedisClient
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
     this.blockFinder = new BlockFinder(this.configStore.provider.getBlock.bind(this.configStore.provider));
     this.rateModelDictionary = new across.rateModel.RateModelDictionary();
   }
 
   async computeRealizedLpFeePct(
-    deposit: Deposit,
+    deposit: { quoteTimestamp: number; amount: BigNumber },
```
pxrl, line 66: The `deposit` container looks a bit superfluous here...Worth breaking `quoteTimestamp` and `amount` out as discrete variables?

mrice32:
> This is a lot of great work and looks like errors will be caught earlier with unexpected `undefined` values. I wish this PR was smaller though its kind of hard to review

@nicholaspai 

My current plan was to split folder-by-folder -- this just covers src/clients.  But since this folder is big, would you prefer that I split this one in half or thirds (i.e. take half or a third of these files)? Happy to do so if that would help.

src/clients/HubPoolClient.ts
```diff
@@ -64,7 +73,7 @@ export class HubPoolClient {
     else return mostRecentSpokePoolUpdatebeforeBlock.spokePool;
   }
 
-  getDestinationTokenForDeposit(deposit: Deposit) {
+  getDestinationTokenForDeposit(deposit: { originChainId: number; originToken: string; destinationChainId: number }) {
```
pxrl, line 76: Similar comment to the previous about the `deposit` container.

I am curious though - is there an advantage to not using the `Deposit` annotation?

src/clients/MultiCallerClient.ts
```diff
@@ -198,17 +204,17 @@ export class MultiCallerClient {
...
+      );
 
       this.logger.debug({
         at: "MultiCallerClient",
         message: "Executing transactions with msg.value excluded from batches by target chain",
-        txs: Object.keys(groupedValueTransactions).map((chainId) => ({
-          chainId,
-          num: groupedValueTransactions[chainId].length,
+        txs: Object.entries(groupedValueTransactions).map(([chainId, transactions]) => ({
+          chainId: Number(chainId),
```
pxrl, line 216: Is the conversion to a number important here?

nicholaspai (approved):

src/clients/HubPoolClient.ts
mrice32: I'll explain why I did this. Let me know if you think this makes sense!

Previously, the `Deposit` type had two optional fields. These two methods depended on that being optional because these methods _compute_ those unknown fields for the purpose of populating them. Put differently, we wouldn't know those values until calling this method.

However, those fields being optional meant that outside of the update() method of the SpokePoolClient, we had to continuously check if those fields were undefined despite knowing that we _never_ store deposits before those fields are populated. So, I decided to make those fields required and then have these methods just explicitly specify the _parts_ of the Deposit type that they depend on, which allows this incomplete deposit type to be passed to them. This means that everywhere _else_ we don't have to check for undefined, but these methods feel a little weirder.

The alternative would be to build an incomplete deposit type that doesn't contain those two fields (or makes them optional) and use that in these methods rather than specifying the fields directly. Or, we could pass the fields individually as you suggested.

Thoughts?

src/clients/MultiCallerClient.ts
mrice32: Probably not! I'm trying to be consistent about _always_ representing chainIds as Numbers, but in this case, we're just logging it, so you're right that it isn't necessary. Would you like me to drop the Number conversion?

pxrl:
> This is a lot of great work and looks like errors will be caught earlier with unexpected `undefined` values.

+1 !!

> I wish this PR was smaller though its kind of hard to review

I also found that it was a pretty heavy to check. Are there many inter-dependencies between the files? I feel it might be less risky be interested to see if it could be achieved in a smaller batches.

pxrl (commented):
Just publishing these outstanding comments now; will do a proper check tomorrow.

src/clients/BundleDataClient.ts
```diff
@@ -57,6 +56,8 @@ export class BundleDataClient {
 
   getPendingRefundsFromValidBundles(bundleLookback: number): FillsToRefund[] {
     const refunds = [];
+    if (this.clients.hubPoolClient.latestBlockNumber === undefined)
```
pxrl, line 59: `hubPoolClient.latestBlockNumber` seems to be set at the beginning of the HubPool client `update()` method, after which we actually commence updating (including some blocking tasks that will yield). That might indicate there's a theoretical window of opportunity to sneak past this check, if the methods are called concurrently (or something like that).

Admittedly, that shouldn't _ever_ occur in practice because of the current structure of the code. Is it worth considering, though? A possible mitigation might be to set `this.latestBlockNumber = undefined` at the start of `update()`, and to update it at the successful conclusion of the method. It would probably require some refactoring, if it calls other methods that rely on `this.latestBlockNumber`, though.

src/clients/TokenClient.ts
```diff
@@ -97,14 +102,11 @@ export class TokenClient {
...
-          this.tokenData[chainId][token].allowance.lt(toBN(MAX_SAFE_ALLOWANCE))
-        )
-          tokensToApprove.push({ chainId, token });
+    const tokensToApprove: { chainId: number; token: string }[] = [];
+    Object.entries(this.tokenData).forEach(([_chainId, tokenMap]) => {
+      const chainId = Number(_chainId);
+      Object.entries(tokenMap).forEach(([token, { balance, allowance }]) => {
+        if (balance.gt(toBN(0)) && allowance.lt(toBN(MAX_SAFE_ALLOWANCE))) tokensToApprove.push({ chainId, token });
       });
     });
```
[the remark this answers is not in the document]
mrice32: Makes sense. I think this suggestion might accidentally drop the `tokensToApprove` declaration. Will push an update, though.

src/clients/BundleDataClient.ts
mrice32: I can also check for hubPoolClient.isUpdated, which is set at the end. Thoughts?

mrice32: Updated!

src/clients/TokenClient.ts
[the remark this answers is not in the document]
pxrl: Oops, yeah - you're right. My suggestion was also wrong because MAX_SAFE_ALLOWANCE is a string, so it doesn't need to be converted to a BN.

PR #320, opened 2022-09-29
chore(FormattingUtils): Use sdk-v2 implementations

pxrl:
Remove the opportunity for errors to creep in over time due to duplicate implementations. The sdk-v2 implementation of toBN() allows optional decimal scaling, which is nice to have.

Without going on too much of a yak-shaving odyssey, we should also probably migrate some of the extra implementations from here over into the SDK.

For reference, the SDK implementations are available here: https://github.com/across-protocol/sdk-v2/blob/f77190132d3fb8cb88fb34f3141d397942816bdb/src/utils.ts


pxrl:
Closing for now since this needs adjustments in some tests.

PR #321, opened 2022-09-29
improve: Store sped-up relayer fee % in cache

nicholaspai:
When updating deposits following speed ups, the Dataworker cached deposit does not get updated. There is no inherent risk. It does just mean when the Spokepool client updates, and loads deposits from cache, it won't be the most up-to-date version containing the new relayerFeePct.

In practice, deposits are only loaded from the cache if they are older than the most recently validated root bundle, and most deposits are fully filled well before then, but this PR will enhance UX for the edge cases where deposits only get partially filled and are not fully filled for a long time.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -424,11 +424,12 @@ export class SpokePoolClient {
               blockNumber,
               originBlockNumber,
             };
-            dataToCache.deposits[destinationChainId][index] = {
-              ...speedUpDeposit,
-              blockNumber,
-              originBlockNumber,
-            };
+            if (dataToCache?.deposits && dataToCache?.deposits[destinationChainId])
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -425,6 +424,12 @@ export class SpokePoolClient {
               blockNumber,
               originBlockNumber,
             };
+            if (dataToCache?.deposits && dataToCache?.deposits[destinationChainId])
```
mrice32, line 427:
```suggestion
            if (dataToCache?.deposits?.[destinationChainId])
```

mrice32 (approved):
LGTM! I'm trying to think about edge cases, but I guess this is low risk, since if we miss a speedup, nothing breaks.

kevinuma (approved):

PR #326, opened 2022-10-06
improve(ProfitClient): Add new profitability logic

pxrl:
This commit introduces new profitability logic to the ProfitClient.
    The change currently has no functional impact, but prepares the
    ProfitClient to support user-configurable profit margins, rather than
    gross fee margins, as is currently the case.
    
    It also increases the reporting of profitability information by logging,
    but also by exposing the various outputs of the new profitability logic.
    
    A follow-up commit will activate this new logic in the production code.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -108,8 +108,7 @@ export class ProfitClient {
     gasCostUsd: BigNumber;
   } {
     const nativeGasCost: BigNumber = this.getTotalGasCost(chainId); // gas cost in native token
-    const gasToken: string = GAS_TOKEN_BY_CHAIN_ID[chainId];
-    const gasPriceUsd: BigNumber = this.getPriceOfToken(gasToken);
+    const gasPriceUsd: BigNumber = this.getPriceOfToken(GAS_TOKEN_BY_CHAIN_ID[chainId]);
     const gasCostUsd: BigNumber = nativeGasCost.mul(gasPriceUsd).div(toBN(10).pow(GAS_TOKEN_DECIMALS));
 
     return {
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -19,7 +19,21 @@ export const MATIC = "0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0";
 export const USDC = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
 export const WETH = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
 
-const GAS_TOKEN_BY_CHAIN_ID = {
+export type FillProfit = {
+  grossRelayerFeePct: BigNumber;
+  l1TokenPriceUsd: BigNumber;
+  fillAmountUsd: BigNumber;
+  grossRelayerFeeUsd: BigNumber;
+  nativeGasCost: BigNumber;
+  gasPriceUsd: BigNumber;
+  gasCostUsd: BigNumber;
+  relayerCapitalOutlayUsd: BigNumber;
+  netRelayerFeePct: BigNumber;
+  netRelayerFeeUsd: BigNumber;
+  fillProfitable: boolean;
+};
+
+export const GAS_TOKEN_BY_CHAIN_ID = {
   1: WETH,
   10: WETH,
   137: MATIC,
@@ -87,12 +101,90 @@ export class ProfitClient {
     return this.totalGasCosts[chainId] ? toBN(this.totalGasCosts[chainId]) : toBN(0);
   }
 
+  // Estimate the gas cost of filling this relay.
+  calculateFillCost(chainId: number): {
+    nativeGasCost: BigNumber;
+    gasPriceUsd: BigNumber;
+    gasCostUsd: BigNumber;
+  } {
+    const nativeGasCost: BigNumber = this.getTotalGasCost(chainId); // gas cost in native token
+    const gasToken: string = GAS_TOKEN_BY_CHAIN_ID[chainId];
+    const gasPriceUsd: BigNumber = this.getPriceOfToken(gasToken);
+    const gasCostUsd: BigNumber = nativeGasCost.mul(gasPriceUsd).div(toBN(10).pow(GAS_TOKEN_DECIMALS));
+
+    return {
+      nativeGasCost,
+      gasPriceUsd,
+      gasCostUsd,
+    };
+  }
+
   getUnprofitableFills() {
     return this.unprofitableFills;
   }
 
   clearUnprofitableFills() {
     this.unprofitableFills = {};
+  }
+
+  calculateFillProfitability(deposit: Deposit, fillAmount: BigNumber, l1Token?: L1Token): FillProfit {
+    assert(fillAmount.gt(0), `Unexpected fillAmount: ${fillAmount}`);
+    assert(
+      Object.keys(GAS_TOKEN_BY_CHAIN_ID).includes(deposit.destinationChainId.toString()),
+      `Unsupported destination chain ID: ${deposit.destinationChainId}`
+    );
+
+    l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
+    assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
+    const l1TokenPriceUsd: BigNumber = this.getPriceOfToken(l1Token.address);
+
+    // Scale l1Token to 18 decimals.
+    assert(l1Token.decimals > 0 && l1Token.decimals <= 18, `Unsupported token: ${JSON.stringify(l1Token)}`);
+    if (l1Token.decimals < 18) {
+      const _fillAmount = fillAmount;
+      fillAmount = toBN(fillAmount).mul(toBNWei(1, 18 - l1Token.decimals));
+    }
+
+    // Use the maximum available relayerFeePct (max of Deposit and SpeedUps).
+    // nb. This grossRelayerFeePct defaults to 0 if deposit.relayerFeePct < 0
+    const grossRelayerFeePct: BigNumber = deposit.relayerFeePct.gte(deposit.newRelayerFeePct ?? 0)
+      ? deposit.relayerFeePct
+      : deposit.newRelayerFeePct || toBN(0);
+
+    // Calculate relayer fee and capital outlay in relay token terms.
+    const grossRelayerFee: BigNumber = grossRelayerFeePct.mul(fillAmount).div(toBNWei(1));
+    const relayerCapitalOutlay: BigNumber = fillAmount.sub(grossRelayerFee);
+    assert(relayerCapitalOutlay.add(grossRelayerFee).eq(fillAmount), "Relayer fee miscalculation");
+
+    // Normalise to USD terms.
+    const fillAmountUsd: BigNumber = fillAmount.mul(l1TokenPriceUsd).div(toBNWei(1));
+    const grossRelayerFeeUsd: BigNumber = grossRelayerFee.mul(l1TokenPriceUsd).div(toBNWei(1));
+    const relayerCapitalOutlayUsd: BigNumber = relayerCapitalOutlay.mul(l1TokenPriceUsd).div(toBNWei(1));
+    assert(fillAmountUsd.eq(relayerCapitalOutlayUsd.add(grossRelayerFeeUsd)), "Relayer fee miscalculation");
+
+    // Estimate the gas cost of filling this relay.
+    const { nativeGasCost, gasPriceUsd, gasCostUsd } = this.calculateFillCost(deposit.destinationChainId);
+
+    // Subtract gas cost from revenue.
+    const netRelayerFeeUsd: BigNumber = grossRelayerFeeUsd.sub(gasCostUsd);
+    const netRelayerFeePct: BigNumber = netRelayerFeeUsd.mul(toBNWei(1)).div(relayerCapitalOutlayUsd);
+
+    // If the gas cost is unknown, assume the relay is unprofitable.
+    const fillProfitable = gasCostUsd.gt(0) && netRelayerFeePct.gte(this.minRelayerFeePct);
+
+    return {
+      grossRelayerFeePct,
+      l1TokenPriceUsd,
+      fillAmountUsd,
+      grossRelayerFeeUsd,
+      nativeGasCost,
+      gasPriceUsd,
+      gasCostUsd,
+      relayerCapitalOutlayUsd,
+      netRelayerFeePct,
+      netRelayerFeeUsd,
+      fillProfitable,
+    } as FillProfit;
   }
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber) {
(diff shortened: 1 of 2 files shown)


src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
     this.unprofitableFills = {};
   }
 
+  calculateFillProfitability(deposit: Deposit, fillAmount: BigNumber, l1Token?: L1Token): FillProfit {
```
pxrl, line 129: nb. there is a fairly comprehensive test for this method that is introduced in a follow-up commit. It wasn't added as part of this change, since it relies on testing via isFillProfitable().

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+      Object.keys(GAS_TOKEN_BY_CHAIN_ID).includes(deposit.destinationChainId.toString()),
+      `Unsupported destination chain ID: ${deposit.destinationChainId}`
+    );
+
+    l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
+    assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
+    const l1TokenPriceUsd = this.getPriceOfToken(l1Token.address);
+
+    // Scale l1Token to 18 decimals.
+    assert(l1Token.decimals > 0 && l1Token.decimals <= 18, `Unsupported token: ${JSON.stringify(l1Token)}`);
```
nicholaspai, line 141: Does getL1Token not have this decimals assertion already? If not we should add it there so all downstream callers can have more confidence the decimals are correct

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+    // Calculate relayer fee and capital outlay in relay token terms.
+    const grossRelayerFee = grossRelayerFeePct.mul(scaledFillAmount).div(toBNWei(1));
+    const relayerCapitalOutlay: BigNumber = scaledFillAmount.sub(grossRelayerFee);
+    assert(relayerCapitalOutlay.add(grossRelayerFee).eq(scaledFillAmount), "Relayer fee miscalculation");
+
+    // Normalise to USD terms.
+    const fillAmountUsd = scaledFillAmount.mul(l1TokenPriceUsd).div(toBNWei(1));
+    const grossRelayerFeeUsd = grossRelayerFee.mul(l1TokenPriceUsd).div(toBNWei(1));
+    const relayerCapitalOutlayUsd = relayerCapitalOutlay.mul(l1TokenPriceUsd).div(toBNWei(1));
+    assert(fillAmountUsd.eq(relayerCapitalOutlayUsd.add(grossRelayerFeeUsd)), "Relayer fee miscalculation");
```
nicholaspai, line 160: Same comment as above about unnecessary check

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+    // Estimate the gas cost of filling this relay.
+    const { nativeGasCost, gasPriceUsd, gasCostUsd } = this.calculateFillCost(deposit.destinationChainId);
+
+    // Subtract gas cost from revenue.
+    const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
+    const netRelayerFeePct = netRelayerFeeUsd.mul(toBNWei(1)).div(relayerCapitalOutlayUsd);
+
+    // If the gas cost is unknown, assume the relay is unprofitable.
+    const fillProfitable = gasCostUsd.gt(0) && netRelayerFeePct.gte(this.minRelayerFeePct);
+
```
nicholaspai, line 171: Should we log something if gas cost is unknown?

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -102,7 +140,7 @@ describe("ProfitClient: Consider relay profit", async function () {
     expect(profitClientWithMinFee.isFillProfitable(profitableWethL1Relay, toBNWei(1))).to.be.false;
   });
 
-  it("Considers deposits with newRelayerFeePct", async function () {
+  it("Considers deposits with newRelayerFeePct (old)", async function () {
```
nicholaspai, line 143: What does old mean? Can we include a comment explaining it?

nicholaspai (approved):
This is really great work. I've left a few questions but structurally this looks solid

src/clients/ProfitClient.ts
pxrl: It hadn't occurred to me to check, but I have now checked and it seems not to check the number of decimals. I can easily propose that in a follow-up PR.

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+    // Use the maximum available relayerFeePct (max of Deposit and SpeedUps).
+    // nb. This grossRelayerFeePct defaults to 0 if deposit.relayerFeePct < 0
+    const grossRelayerFeePct = deposit.relayerFeePct.gte(deposit.newRelayerFeePct ?? 0)
+      ? deposit.relayerFeePct
+      : deposit.newRelayerFeePct || toBN(0);
+
+    // Calculate relayer fee and capital outlay in relay token terms.
+    const grossRelayerFee = grossRelayerFeePct.mul(scaledFillAmount).div(toBNWei(1));
+    const relayerCapitalOutlay: BigNumber = scaledFillAmount.sub(grossRelayerFee);
+    assert(relayerCapitalOutlay.add(grossRelayerFee).eq(scaledFillAmount), "Relayer fee miscalculation");
```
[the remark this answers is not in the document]
pxrl: Fair call; I had a few safeguards in whilst I was testing various formulations but I think they're safe to pull out now.

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+    const grossRelayerFeeUsd = grossRelayerFee.mul(l1TokenPriceUsd).div(toBNWei(1));
+    const relayerCapitalOutlayUsd = relayerCapitalOutlay.mul(l1TokenPriceUsd).div(toBNWei(1));
+    assert(fillAmountUsd.eq(relayerCapitalOutlayUsd.add(grossRelayerFeeUsd)), "Relayer fee miscalculation");
+
+    // Estimate the gas cost of filling this relay.
+    const { nativeGasCost, gasPriceUsd, gasCostUsd } = this.calculateFillCost(deposit.destinationChainId);
+
+    // Subtract gas cost from revenue.
+    const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
+    const netRelayerFeePct = netRelayerFeeUsd.mul(toBNWei(1)).div(relayerCapitalOutlayUsd);
```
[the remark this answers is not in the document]
pxrl: If we fail to resolve a token price (or if a token actually _does_ go to zero...) then this can also be zero, and then this will catch fire. Nice catch; I totally overlooked that possibility.

test/ProfitClient.ConsiderProfitability.ts
pxrl: Ah - this one actually gets removed in the follow-up commit. It's not relevant with the new profitability logic, but I thought it'd be relevant to retain the existing test since this change does not actually activate the new logic yet.

src/clients/ProfitClient.ts
pxrl: This one worked fine with my nice, conformant test vectors...but it is subject to single-digit Wei rounding errors and I have seen it blow up when evaluating real deposits. In any case, it should be fine to remove now, as you suggest.

src/clients/ProfitClient.ts
pxrl: See below for the logging that is added to the end of isFillProfitable() by my follow-up change.

I'll propose to change it as follows:
`if (this.debugProfitability) {` => `if (!fill.fillProfitable || this.debugProfitability) {`

Then we'll only see logs being generated when we specifically want them, or when something is unprofitable (which we probably want to look more closely at anyway). Hopefully this includes enough information for us to quickly intuit what the issue is.

```
    if (this.debugProfitability) {
      // @todo: Remove cloneDeep() once the logging implementation is updated.
       this.logger.debug({
        at: "ProfitClient#isFillProfitable",
        message: `${l1Token.symbol} fill is ${fill.fillProfitable ? "" : "un"}profitable`,
        deposit: _.cloneDeep(deposit),
        l1Token: _.cloneDeep(l1Token),
        fillAmount,
        fillAmountUsd: fill.fillAmountUsd,
        grossRelayerFeePct: `${formatFeePct(fill.grossRelayerFeePct)}%`,
        nativeGasCost: fill.nativeGasCost,
        gasPriceUsd: fill.gasPriceUsd,
        relayerCapitalOutlayUsd: `${fill.relayerCapitalOutlayUsd}`,
        grossRelayerFeeUsd: fill.grossRelayerFeeUsd,
        gasCostUsd: fill.gasCostUsd,
        netRelayerFeeUsd: `${fill.netRelayerFeeUsd}`,
        netRelayerFeePct: `${formatFeePct(fill.netRelayerFeePct)}%`,
        minRelayerFeePct: `${formatFeePct(this.minRelayerFeePct)}%`,
        fillProfitable: fill.fillProfitable,
     });
```

src/clients/ProfitClient.ts
nicholaspai: why not add the fix there in this PR? Seems low lift and more beneficial repo-wide

src/clients/ProfitClient.ts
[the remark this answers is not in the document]
nicholaspai: yeah lets just assume that BN functions work

src/clients/ProfitClient.ts
nicholaspai: I empathize with your test vectors, i feel like we can still test the math by testing the ultimate return values from this function

src/clients/ProfitClient.ts
nicholaspai: should we only emit this when the not profitable trigger is hit?

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+    const grossRelayerFeePct = deposit.relayerFeePct.gte(deposit.newRelayerFeePct ?? 0)
+      ? deposit.relayerFeePct
+      : deposit.newRelayerFeePct || toBN(0);
+
+    // Calculate relayer fee and capital outlay in relay token terms.
+    const grossRelayerFee = grossRelayerFeePct.mul(scaledFillAmount).div(toBNWei(1));
+    const relayerCapitalOutlay = scaledFillAmount.sub(grossRelayerFee);
+
+    // Normalise to USD terms.
+    // nb. Price may be zero.
```
mrice32, line 155: When would the price be zero? Is it useful to run this profitability logic when the price is zero?

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+      fillAmountUsd,
+      grossRelayerFeeUsd,
+      nativeGasCost,
+      gasPriceUsd,
+      gasCostUsd,
+      relayerCapitalOutlayUsd,
+      netRelayerFeePct,
+      netRelayerFeeUsd,
+      fillProfitable,
+    };
```
mrice32, line 184: As a point of documentation, it's probably a good idea to document (either here or elsewhere) these return values.

Some are obvious, but some are less obvious to the casual observer, for instance relayerCapitalOutlayUsd. Additionally, documenting the scaling of these variables might be useful!

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -26,6 +28,16 @@ const tokenPrices: { [symbol: string]: BigNumber } = {
   WETH: toBNWei(3000),
 };
 
+// Quirk: Use the chainId as the gas price in Gwei. This gives a range of
```
mrice32, line 31: Ha!

mrice32 (commented):
Looks great A few minor comments

src/clients/ProfitClient.ts
pxrl: Since we're discussing it [here](https://github.com/across-protocol/relayer-v2/pull/329), I'll mark this resolved :+1:.

src/clients/ProfitClient.ts
pxrl: @nicholaspai I've made it configurable by the DEBUG_PROFITABILITY env var, so we can toggle it without needing a code change. Does that sound OK?

test/ProfitClient.ConsiderProfitability.ts
pxrl: Assume this is OK to resolve now.

src/clients/ProfitClient.ts
pxrl: > When would the price be zero? Is it useful to run this profitability logic when the price is zero?

The price can be zero if we fail on a price lookup - this is existing behaviour. We can absolutely infer `fillProfitable = false` in that instance, so it's definitely valid to ask why we bother with the calculation.

I landed on a strategy to avoid branching on these "sometimes fails" conditionals that are dependent on external factors (i.e. price lookups), and instead tried to make sure that the profitability logic is robust to a lookup failure and will still produce a correct result. That at least addresses one of my own observations in the existing code - i.e. that we can make it half way through `isFillProfitable()` and then bail out if a price lookup failed, but only dependent on the setting of an env var. I'm not enthusiastic about that behaviour because it makes it difficult for us to reason about the behaviour for RL and non-RL bots.

Thinking about it a bit more...a small refactor here would make it possible to skip the price normalisation (i.e. conversion to USD base) in the event of a price lookup failure.

src/clients/ProfitClient.ts
```diff
@@ -95,6 +126,64 @@ export class ProfitClient {
...
+
+    // Normalise to 18 decimals.
+    const scaledFillAmount =
+      l1Token.decimals === 18 ? fillAmount : toBN(fillAmount).mul(toBNWei(1, 18 - l1Token.decimals));
+
+    // Use the maximum available relayerFeePct (max of Deposit and SpeedUps).
+    // nb. This grossRelayerFeePct defaults to 0 if deposit.relayerFeePct < 0.
+    const grossRelayerFeePct = deposit.relayerFeePct.gte(deposit.newRelayerFeePct ?? 0)
+      ? deposit.relayerFeePct
+      : deposit.newRelayerFeePct || toBN(0);
```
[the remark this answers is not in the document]
pxrl: Done :)

https://github.com/across-protocol/relayer-v2/pull/326/commits/2ddfd7086c7f2d1292b66ffd3afcb54261599264

src/clients/ProfitClient.ts
pxrl: Simplified it here: https://github.com/across-protocol/relayer-v2/pull/326/commits/9e989169cfaaa37b6442692c20b2bdd51c984296

Now we throw if we can't resolve a price or cost. I came to see that it made the most sense, since we otherwise can't produce any meaningful value for netRelayerFeePct in the absence of the costs.

src/clients/ProfitClient.ts
pxrl: Have tried to document some of these. The naming is not great (especially `relayerCapitalOutlayUsd`). Not entirely satisfied, but I feel like there should be enough here (+ w/ comments in `calculateFillProfitability()`) for someone to figure it out.

src/clients/ProfitClient.ts
nicholaspai: That sounds good to me

mrice32 (approved):
LGTM!

src/clients/ProfitClient.ts
mrice32: Sounds great!

PR #327, opened 2022-10-06
fix(logger): fix logger bug that converts BNs to strings

mrice32:
This pulls in an update to the logger that stops it from converting BNs to strings in objects that are passed to it by reference.


pxrl (approved):
Nice!!

PR #328, opened 2022-10-07
fix: Correctly update deposits in cache with sped up relayer fee pct 

nicholaspai:
Implements intended change that should have been included in https://github.com/across-protocol/relayer-v2/pull/321


mrice32 (approved):
LGTM!

PR #329, opened 2022-10-07
improve(HubPoolClient): Warn on unexepcted token decimals

pxrl:
Should _never_ happen in practice, but if it did then it would break a lot of assumptions in the bot. This therefore saves consumers of these L1Tokens from having to explicitly check.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -426,13 +426,7 @@ export class HubPoolClient {
     for (const info of tokenInfo) {
       if (!this.l1Tokens.find((token) => token.symbol === info.symbol)) {
         if (info.decimals > 0 && info.decimals <= 18) this.l1Tokens.push(info);
-        else {
-          this.logger.warn({
-            at: "HubPoolClient#update",
-            message: "Unsupported HubPool token.",
-            token: info,
-          });
-        }
+        else throw new Error(`Unsupported HubPool token: ${JSON.stringify(info)}`);
       }
     }
 
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -424,7 +424,16 @@ export class HubPoolClient {
       Promise.all(uniqueL1Tokens.map(async (l1Token: string) => await this.hubPool.pooledTokens(l1Token))),
     ]);
     for (const info of tokenInfo) {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

src/clients/HubPoolClient.ts
```diff
@@ -424,7 +424,16 @@ export class HubPoolClient {
       Promise.all(uniqueL1Tokens.map(async (l1Token: string) => await this.hubPool.pooledTokens(l1Token))),
     ]);
     for (const info of tokenInfo) {
-      if (!this.l1Tokens.find((token) => token.symbol === info.symbol)) this.l1Tokens.push(info);
+      if (!this.l1Tokens.find((token) => token.symbol === info.symbol)) {
+        if (info.decimals > 0 && info.decimals <= 18) this.l1Tokens.push(info);
+        else {
+          this.logger.warn({
```
nicholaspai, line 430: Should we throw here instead of just failing somewhat silently?

pxrl: Normally I would say no, since this is data that's received over the network, and standard practice is not to allow external actors to trigger side-effects like assertions. In this specific case, I'm actually not sure, because I still have some uncertainties about how we view on the trust boundaries in our system.

On the one hand, warning will create alerts, so we will be notified. But I don't know what the impact is of excluding a token from our set. What would that do for any deposit or fill events that might somehow include this token? I can't really comprehend the impact that it might have (..probably we crash in some weird unforeseen way).

It's definitely at least a _little_ bit academic, since the contract logic and contract update procedures _should_ prevent weird tokens from being served here. But we're also receiving RPC data via third-parties and are trusting that the data is correct...so, idk?

mrice32: This is a trusted on-chain action, so I would say we should either throw or at least trigger an error log since this would be really unexpected.

nicholaspai: agreed with matt, across governance shouldn't be approving any such tokens

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/329#discussion_r990525836

src/clients/HubPoolClient.ts
pxrl: Makes sense - have updated to throw an exception instead.

PR #331, opened 2022-10-14
improve(Dataworker): Constrain ProfitClient update

pxrl:
Only update the ProfitClient if the proposer is enabled. Other dataworker functions don't need to know about token prices.

nb. This is a simple, low-impact change to work around logging of failed price request lookups.

### DIFF
--- a/src/dataworker/DataworkerClientHelper.ts
+++ b/src/dataworker/DataworkerClientHelper.ts
@@ -14,11 +14,11 @@ import { getBlockForChain } from "./DataworkerUtils";
 import { Dataworker } from "./Dataworker";
 
 export interface DataworkerClients extends Clients {
-  profitClient: ProfitClient;
   tokenClient: TokenClient;
   spokePoolSigners: { [chainId: number]: Wallet };
   spokePoolClientSearchSettings: { [chainId: number]: EventSearchConfig };
   bundleDataClient: BundleDataClient;
+  profitClient?: ProfitClient;
 }
 
 export async function constructDataworkerClients(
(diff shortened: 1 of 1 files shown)


pxrl:
nb. This PR is deliberately lightweight...we could probably do a bit more to clean up the setup/init for the various dataworker sub-functions, but I don't think this specific change should be held back by that.

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -50,15 +50,17 @@ export async function constructDataworkerClients(
 
   // Disable profitability by default as only the relayer needs it.
   // The dataworker only needs price updates from ProfitClient to calculate bundle volume.
-  const profitClient = new ProfitClient(logger, commonClients.hubPoolClient, {}, false, [], true);
+  const profitClient = config.proposerEnabled
+    ? new ProfitClient(logger, commonClients.hubPoolClient, {}, false, [], true)
+    : undefined;
```
mrice32, line 55: When we start using the price client instead, we won't need this switching logic anymore, right? Because the price client only requests prices _when asked for them_, no? So if the proposer isn't enabled, the code will just never ask for the prices.

Is my understanding correct?

mrice32 (approved):
LGTM!

src/dataworker/DataworkerClientHelper.ts
pxrl: Yeah, that sounds pretty much correct to me.

I guess there's a consideration about whether we should bother initialising/instantiating something that we know in advance won't be used. There are arguments for and against that - i.e. it's arguably simpler, albeit wasteful.

PR #332, opened 2022-10-17
improve(ProfitClient): Simplify update

pxrl:
- Factor out token price update and gas cost update into separate
   methods.
 - Update token prices and gas cost update in parallel.

Motivation: This makes the the follow-up diff to support the PriceClient
smaller and cleaner.


nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #333, opened 2022-10-18
improve: Document and modify default deposit confirmation wait times

nicholaspai:


mrice32 (approved):

pxrl (approved):
Looks good. I wasn't aware that 128 blocks was considered the norm for Polygon - that's interesting to know. The mort important part here is that we set defaults that are safe for third parties.

PR #334, opened 2022-10-18
improve(ProfitClient): Use new PriceClient

pxrl:
This change simplifies the ProfitClient by outsourcing price lookups to
the new PriceClient. This also has the advantage of supporting multiple
price feeds, thereby improving robustness against a single point of
failure.

There's further cleanup that can be done here once other "price
consumers" switch over to using the PriceClient directly.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -19,6 +19,7 @@ export type FillProfit = {
   fillAmountUsd: BigNumber; // Amount of the bridged token being filled.
   grossRelayerFeeUsd: BigNumber; // USD value of the relay fee paid by the user.
   nativeGasCost: BigNumber; // Cost of completing the fill in the native gas token.
+  gasMultiplier: number; // Multiplier to apply to nativeGasCost as padding or discount
   gasPriceUsd: BigNumber; // Price paid per unit of gas in USD.
   gasCostUsd: BigNumber; // Estimated cost of completing the fill in USD.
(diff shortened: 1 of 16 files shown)


src/clients/ProfitClient.ts
```diff
@@ -81,7 +77,10 @@ export class ProfitClient {
     readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT),
     readonly debugProfitability: boolean = false
   ) {
-    this.coingecko = new Coingecko();
+    this.priceClient = new PriceClient(logger, [
+      new acrossApi.PriceFeed("Across API", {}),
+      new coingecko.PriceFeed("CoinGecko Free", {}),
```
mrice32, line 82: nit: why can't we just not pass in the second arg if we want to just keep the defaults? Might be a good thing to add to these classes in the SDK in a follow-up (not necessary for this PR).

mrice32 (approved):
LGTM! Only minor comments

src/clients/ProfitClient.ts
pxrl: Yeah; I agree. I fiddled with it briefly in the SDK but got stuck on some syntactical issues, and dropped it. I'm sure it can be done and I just need a little bit more patience.

It should also be possible to drop the name - it'll be relatively rare for the caller to want a different name.

pxrl: Had a crack at it here: https://github.com/across-protocol/sdk-v2/pull/114

PR #336, opened 2022-10-18
improve(ProfitClient): Activate profitability logic

pxrl:
This commit enables the new profitability logic that has been introduced in recent ProfitClient work.

The env var MIN_RELAYER_FEE_PCT is now more strict, and enforces a minimum fee as a percentage of the amount to be relayers _after_ taking the gas cost into account.

Additionally, more logging/diagnostic information is now available via the DEBUG_PROFITABILITY env var.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -151,7 +151,7 @@ export class ProfitClient {
     l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
     assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
     const tokenPriceUsd = this.getPriceOfToken(l1Token.address);
-    if (tokenPriceUsd.lte(0)) throw new Error(`Unable to determine ${l1Token.symbol}) L1 token price`);
+    if (tokenPriceUsd.lte(0)) throw new Error(`Unable to determine ${l1Token.symbol} L1 token price`);
 
     // Normalise to 18 decimals.
     const scaledFillAmount =
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -141,6 +141,11 @@ export class ProfitClient {
     this.unprofitableFills = {};
   }
 
+  appliedRelayerFeePct(deposit: Deposit): BigNumber {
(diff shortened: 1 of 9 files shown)


src/clients/ProfitClient.ts
```diff
@@ -191,80 +194,35 @@ export class ProfitClient {
...
+        gasCostUsd: fill.gasCostUsd,
+        netRelayerFeeUsd: `${fill.netRelayerFeeUsd}`,
+        netRelayerFeePct: `${formatFeePct(fill.netRelayerFeePct)}%`,
         minRelayerFeePct: `${formatFeePct(this.minRelayerFeePct)}%`,
+        fillProfitable: fill.fillProfitable,
       });
-      return false;
     }
 
-    if (relayerFeePct.eq(toBN(0))) {
```
nicholaspai, line 211: Just curious, was it an oversight that we didn't remove this logical block from this function in the previous PR?

nicholaspai (commented):
This LGTM

src/clients/ProfitClient.ts
pxrl: My previous PR deliberately ignored this function because I wanted to add the new logic without impacting any of our current behaviour...but that code does look redundant so we've missed it at some point.

src/clients/ProfitClient.ts
```diff
@@ -191,80 +194,35 @@ export class ProfitClient {
...
-      fillRevenueInUsd,
-      totalGasCostWei,
-      gasCostInUsd,
-      fillProfitInUsd,
-      fillProfitable,
-    });
-    return fillProfitable;
+    // If profitability is disabled, ensure _at least_ that the relayerFeePct >= minRelayerFeePct.
+    // This is a temporary measure and can hopefully be removed (together with ignoreProfitability) in future.
+    return fill.fillProfitable || (this.ignoreProfitability && fill.grossRelayerFeePct.gte(this.minRelayerFeePct));
```
mrice32, line 225: nit: we don't need to change this here, but I think these variables are a bit confusing. If we're ignoring profitability, why do we care about the minRelayerFeePct?

It seems like what we're really saying is: do we care about _gas costs, capital "costs", both, or neither_.

I would argue that what we're really doing when ignoring "profitability" is just ignoring gas costs.

mrice32 (approved):
LGTM!

PR #337, opened 2022-10-19
feat: Allow custom minimum deposit confirmation numbers based on deposited USD amount

nicholaspai:
This will reduce the average time to respond to a deposit assuming the large majority of deposits are < SMALL_TRANSFER_USD_THRESHOLD and LARGE_TRANSFER_USD_THRESHOLD

This PR should not be merged without first changing the bot-configs for live bots because the MIN_DEPOSIT_CONFIRMATIONS env variable shape has changed

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -288,13 +288,7 @@ export class ProfitClient {
   }
 
   async update(): Promise<void> {
-    const updates = [this.updateTokenPrices()];
-
-    // Skip gas cost lookup if profitability is disabled. We still
-    // need to fetch token prices but don't care about gas costs.
-    if (!this.ignoreProfitability) updates.push(this.updateGasCosts());
-
-    await Promise.all(updates);
+    await Promise.all([this.updateTokenPrices(), this.updateGasCosts()]);
   }
 
   protected async updateTokenPrices(): Promise<void> {
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -17,6 +17,7 @@ type CoinGeckoPrice = {
 // @todo: These don't belong in the ProfitClient; they should be relocated.
 export const MATIC = "0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0";
 export const USDC = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
+export const WBTC = "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599";
 export const WETH = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
 
 // note: All FillProfit BigNumbers are scaled to 18 decimals unless specified otherwise.
@@ -68,6 +69,7 @@ export class ProfitClient {
   // Queries needed to fetch relay gas costs.
   private relayerFeeQueries: { [chainId: number]: relayFeeCalculator.QueryInterface } = {};
 
+  // @todo: Consolidate this set of args before it grows legs and runs away from us.
   constructor(
     readonly logger: winston.Logger,
     readonly hubPoolClient: HubPoolClient,
@@ -76,7 +78,8 @@ export class ProfitClient {
     readonly enabledChainIds: number[],
     // Default to throwing errors if fetching token prices fails.
     readonly ignoreTokenPriceFailures: boolean = false,
-    readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT)
+    readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT),
+    readonly debugProfitability: boolean = false
   ) {
     this.coingecko = new Coingecko();
 
@@ -138,6 +141,11 @@ export class ProfitClient {
     this.unprofitableFills = {};
   }
 
+  appliedRelayerFeePct(deposit: Deposit): BigNumber {
+    // Return the maximum available relayerFeePct (max of Deposit and any SpeedUp).
+    return max(deposit.relayerFeePct, deposit.newRelayerFeePct ?? toBN(0));
+  }
+
   calculateFillProfitability(deposit: Deposit, fillAmount: BigNumber, l1Token?: L1Token): FillProfit {
     assert(fillAmount.gt(0), `Unexpected fillAmount: ${fillAmount}`);
     assert(
@@ -148,14 +156,13 @@ export class ProfitClient {
     l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
     assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
     const tokenPriceUsd = this.getPriceOfToken(l1Token.address);
-    if (tokenPriceUsd.lte(0)) throw new Error(`Unable to determine ${l1Token.symbol}) L1 token price`);
+    if (tokenPriceUsd.lte(0)) throw new Error(`Unable to determine ${l1Token.symbol} L1 token price`);
 
     // Normalise to 18 decimals.
     const scaledFillAmount =
       l1Token.decimals === 18 ? fillAmount : toBN(fillAmount).mul(toBNWei(1, 18 - l1Token.decimals));
 
-    // Use the maximum available relayerFeePct (max of Deposit and SpeedUps).
-    const grossRelayerFeePct = max(deposit.relayerFeePct, deposit.newRelayerFeePct ?? toBN(0));
+    const grossRelayerFeePct = this.appliedRelayerFeePct(deposit);
 
     // Calculate relayer fee and capital outlay in relay token terms.
     const grossRelayerFee = grossRelayerFeePct.mul(scaledFillAmount).div(toBNWei(1));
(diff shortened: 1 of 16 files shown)


src/common/Constants.ts
```diff
@@ -22,22 +22,34 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
-// finality for any distance between 0 and 1 week.
+// Reorgs are anticipated on Ethereum and Polygon. We use different following distances when processing deposit
+// events based on the USD amount of the deposit.
+// Ethereum: https://etherscan.io/blocks_forked (not working since the merge)
+// Polygon: https://polygonscan.com/blocks_forked
+// Optimistic Rollups are currently centrally serialized and are not expected to reorg.
+export const MIN_DEPOSIT_CONFIRMATIONS_SMALL_TRANSFERS: { [chainId: number]: number } = {
+  1: 4,
+  10: 0,
+  137: 32,
```
pxrl, line 33: I think this will probably cause third-party relayers to lose funds. Should we set it more conservatively and just override in our local config?

src/common/Constants.ts
```diff
@@ -22,22 +22,34 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
+  288: 0,
+  42161: 0,
+};
 export const MIN_DEPOSIT_CONFIRMATIONS: { [chainId: number]: number } = {
-  1: 32, // Eth post-merge finality time
+  1: 16,
+  10: 0,
+  137: 64,
```
pxrl, line 40: Same as above.

src/common/Constants.ts
pxrl, line 33: What if we just had a single MIN_DEPOSIT_CONFIRMATIONS that was divided by threshold? Then we'd just need to modify the code to use the lowest usdVolume key in the object.

It would give us a lot of flexibility in the configuration and doesn't shoehorn us into "small" and "large". For a user who doesn't want to make the distinction, they can just set the limit to 1M (or 0).

```suggestion
export const MIN_DEPOSIT_CONFIRMATIONS: { [usdVolume: number]: { [chainId: number]: number } = {
  5_000: { 1: 4, /* ... */ },
  10_000: { 1: 16, /* ... */ },
  50_000: { 1: 24, /* ... */ },
  1_000_000: { 1: 32, /* ... */ },
}; 
```

nicholaspai: what do you think are reasonable settings for small, medium, large transfers?

nicholaspai: hmm thats interesting, i'll take a look

nicholaspai: I've implemented the suggested structure, nice suggestion!

cc7768: I agree that default settings should probably be more conservative for non-RL users -- We don't want relayers to double fill without actively choosing to set more aggressive bounds themselves (and thus understanding they're taking on reorg risk).

I'd lean towards

```
{
  1: 12,
  10: 0,
  137: 80,
  288: 0,
  42161: 0
}
```

as the default for all MDCs.

cc7768:
I mentioned this in a comment above but @nicholaspai asked for my thoughts on what good defaults are for 3rd party relayers:

I think we should set safe bounds for third party relayers (maybe even our original bounds that were extra safe) and then document the feature clearly. If they want to loosen the safety rails on their own, then we want them to be able to but I think the default should be extra safe to minimize the risk of loss of funds.

We could even leave the bounds that we implement commented out below the default values.

src/common/Constants.ts
nicholaspai: Good point I've changed

src/common/Constants.ts
```diff
@@ -22,21 +22,47 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
+// 137, so it would use a follow distance of 64 blocks.
+
+// To see the latest block reorg events go to:
+// - Ethereum: https://etherscan.io/blocks_forked
+// - Polygon: https://polygonscan.com/blocks_forked
+
+// Optimistic Rollups are currently centrally serialized and are not expected to reorg. Technically a block on an
+// ORU will not be finalized until after 7 days, so there is little difference in following behind 0 blocks versus
+// anything under 7 days.
+export const MIN_DEPOSIT_CONFIRMATIONS: { [threshold: number]: { [chainId: number]: number } } = {
```
cc7768, line 44: These updated values make a lot of sense to me.

cc7768 (approved):

src/common/Constants.ts
```diff
@@ -22,21 +22,47 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
+    42161: 0,
+  },
+  5_000: {
+    1: 32, // Justified block
+    10: 0,
+    137: 100, // Probabilistically safe level based on historic Polygon reorgs
+    288: 0,
+    42161: 0,
+  },
+  0: {
```
mrice32, line 59: What happens if someone doesn't specify a value for 0?

src/relayer/Relayer.ts
```diff
@@ -36,13 +36,22 @@ export class Relayer {
...
+    // Sort thresholds in descending order.
+    const minimumDepositConfirmationThresholds = Object.keys(this.config.minDepositConfirmations).sort(
+      (x, y) => Number(y) - Number(x)
+    );
+
     // Require that all fillable deposits meet the minimum specified number of confirmations.
     const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack)
       .filter((x) => {
+        const unfilledAmountUsd = this.clients.profitClient.getFillAmountInUsd(x.deposit, x.unfilledAmount);
+        const usdThresholdToUse = minimumDepositConfirmationThresholds.find((usdThreshold) => {
```
mrice32, line 48: Related to my comment above, how would an undefined response be handled here?

Note: this is the sort of thing that typescript strict will help us catch!

mrice32 (commented):
Looks good! One minor comment about handling misconfigurations/missing lower-end values.

src/common/Constants.ts
pxrl: I like the new format! Still trying to get my head around the limit structure.

src/common/Constants.ts
nicholaspai: Good point, i'll add a check to RelayerConfig to force one key to be 0

nicholaspai:
```ts
assert(
      Object.keys(this.minDepositConfirmations).some((threshold) => threshold === "0"),
      "Unspecified 0 threshold in minDepositConfirmations"
    );
```

src/relayer/Relayer.ts
nicholaspai: I added a comment but I think [this check](https://github.com/across-protocol/relayer-v2/pull/337#discussion_r1000528271) enforces that `undefined` can never be returned since the fill amount in USD is always >= 0

src/common/Constants.ts
```diff
@@ -22,21 +22,47 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
+// events based on the USD amount of the deposit. This protects the relayer from the worst case situation where it fills
+// a large deposit (i.e. with an amount equal to a large amount of $$$) but the deposit is included in a re-orged
+// block. This would cause the relayer to unintentionally send an invalid fill and not refunded. The tradeoff is that
+// the larger the follow distance, the slower the relayer will be to fulfill deposits. Therefore, the following
+// configuration allows the user to set higher follow distances for higher deposit amounts.
+// The Key of the following dictionary is used as the USD threshold to determine the MDC:
+// - Searching from highest USD threshold to lowest
+// - if the deposited USD amount is <= the key value, then the MDC associated with the key for the origin chain
+// - For example, a deposit on Polygon worth $4,000 would use the MDC associated with the 5_000 key and chain
+// 137, so it would use a follow distance of 64 blocks.
```
pxrl, line 35: I'm a little confused by this comment in relation to the 0 key. The comment seems to indicate that any value from 0 -> 5,000 would be handled by the 5,000 object. What then is the purpose of the 0 key? It specifies min deposit confirmations that are even more aggressive than for the 5,000 key, so it doesn't seem like a fallback option.

nicholaspai: I think the top comment is key: "Searching from highest USD threshold to lowest", this guarantees that 0 key is always used as a fallback

nicholaspai: I've updated the example though which I think was wrong. $4000 should use 0 key not $5000 key

pxrl: I'm still struggling to wrap my head around it. There are the following comments:

`if the deposited USD amount is <= the key value,`
^^ If I understand correctly, this comment can never hold true for the 0 threshold.

`For example, a deposit on Polygon worth $4,000 would use the MDC associated with the 5_000 key ...`
^^ Additionally, this one still identifies that the 5_000 threshold applies to a fill amount of $4,000 USD.

There's a non-zero chance that I'm just misunderstanding the text...but at this point, I _think_ the most logical thing (for me) would be:
 - 0 is explicitly identified as a fallback option.
 - Each threshold identifies the _maximum_ USD value for the corresponding MDCs to be used (i.e. 5_000 applies for 0 < fillAmountUsd < 5_000.

Alternatively (and maybe this is better?):
 - Each threshold identifies the maximum USD value (as above).
 - The fallback value is not configurable; we instead hardcode "absolute final" block numbers for any value that falls is higher than the highest configured threshold.

The reason I'm zooming in on this particular item is that I foresee it might be difficult for us to easily explain it to third-parties, and I'd really like to avoid a situation where someone misconfigures it because they misunderstood how it worked.

pxrl:
@nicholaspai Following up from our sync, it seems more likely that an overall volume limit is pretty important. Consider a sophisticated user who wants to transfer $1M for across-chain arb...If they know that the bot will happily accept smaller deposits earlier, they might be motivated to split their transfer into smaller chunks in order to have their relays completed faster. If this coincides with a reorg we'd have lots of deposits and a high volume, which could be pretty damaging.

src/common/Constants.ts
nicholaspai: > if the deposited USD amount is <= the key value,
^^ If I understand correctly, this comment can never hold true for the 0 threshold.

How is this untrue for the 0 threshold?

nicholaspai: I've also added this rule into `RelayerConfig` to guarantee that the fallback is set:

```
    assert(
      Object.keys(this.minDepositConfirmations).some((threshold) => threshold === "0"),
      "Unspecified 0 threshold in minDepositConfirmations"
    );
```

nicholaspai:
> @nicholaspai Following up from our sync, it seems more likely that an overall volume limit is pretty important. Consider a sophisticated user who wants to transfer $1M for across-chain arb...If they know that the bot will happily accept smaller deposits earlier, they might be motivated to split their transfer into smaller chunks in order to have their relays completed faster. If this coincides with a reorg we'd have lots of deposits and a high volume, which could be pretty damaging.

Great point, i've implemented this can you PTAL!

src/clients/ProfitClient.ts
```diff
@@ -143,7 +143,7 @@ export class ProfitClient {
 
   appliedRelayerFeePct(deposit: Deposit): BigNumber {
     // Return the maximum available relayerFeePct (max of Deposit and any SpeedUp).
-    return max(deposit.relayerFeePct, deposit.newRelayerFeePct ?? toBN(0));
+    return max(toBN(deposit.relayerFeePct), deposit.newRelayerFeePct ? toBN(deposit.newRelayerFeePct) : toBN(0));
```
pxrl, line 146: I think the root cause of the need for this lies in a test fixture in the contracts-v2 repository: https://github.com/across-protocol/contracts-v2/blob/a0748ef25ef322b4e91002872c7687d6c7be489d/test/fixtures/SpokePool.Fixture.ts#L213

How's our enthusiasm about correcting that at the source? The change that's been applied to this function seems _only_ needed to pass certain tests; in prod we should never see anything other than BNs here.

src/common/Constants.ts
pxrl: Marking this resolved since we aligned in DM :+1:

src/relayer/Relayer.ts
```diff
@@ -36,13 +36,53 @@ export class Relayer {
...
+        ];
+      })
+    );
+    this.logger.debug({
+      at: "Relayer",
+      message: "Setting minimum deposit confirmation based on origin chain aggregate deposit amount",
+      unfilledDepositAmountsPerChain,
+      mdcPerChain,
+      minDepositConfirmations: this.config.minDepositConfirmations,
+    });
```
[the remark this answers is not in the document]
nicholaspai: they're all logged as keys in `unfilledDepositAmountsPerChain`, `mdcPerChain`, and `minDepositConfirmations`

src/clients/ProfitClient.ts
nicholaspai: We should just not import `deposit` from contracts-v2, as its a typescript helper function. We should implement in this repo only

src/common/Constants.ts
```diff
@@ -22,21 +22,49 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
...
+    42161: 0,
+  },
+  5_000: {
+    1: 16, // Mainnet reorgs are rarely > 4 blocks in depth so this is a very safe buffer
+    10: 0,
+    137: 80,
+    288: 0,
+    42161: 0,
+  },
 };
```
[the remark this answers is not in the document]
nicholaspai: I'm with you on forcing default to be in there, i'll make that change. I don't want to remove the 5k and 10k levels though as leaving them in there gives confidence (and reasonable defaults) that we can add those thresholds without breaking the bot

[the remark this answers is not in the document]
nicholaspai: Why don't we lower the thresholds for now to 50 to 1000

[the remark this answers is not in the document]
nicholaspai: or 100 and 1000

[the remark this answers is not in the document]
pxrl: Sure, I think that's a good proposal!

pxrl (approved):
Nice feature!!

src/clients/ProfitClient.ts
pxrl: Have created a Linear task for that: https://linear.app/uma/issue/ACX-413/relayer-v2-deposit-implementation-for-test

src/relayer/Relayer.ts
```diff
@@ -36,13 +36,53 @@ export class Relayer {
...
+    const minimumDepositConfirmationThresholds = Object.keys(this.config.minDepositConfirmations)
+      .filter((x) => x !== "default")
+      .sort((x, y) => Number(x) - Number(y));
+
+    // Set the MDC for each origin chain equal to lowest threshold greater than the unfilled USD deposit amount.
+    // If we can't find a threshold greater than the USD amount, then use the default.
+    const mdcPerChain = Object.fromEntries(
+      Object.keys(unfilledDepositAmountsPerChain).map((chainId) => {
+        const usdThreshold = minimumDepositConfirmationThresholds.find((_usdThreshold) => {
+          return toBNWei(_usdThreshold).gte(unfilledDepositAmountsPerChain[chainId]);
```
mrice32, line 60: Nit: up to you, but I find using entries to be simpler than re-accessing the struct. I think it's also a bit easier to guarantee that the assessed value actually exists (i.e. isn't undefined).

```suggestion
    const mdcPerChain = Object.fromEntries(
      Object.entries(unfilledDepositAmountsPerChain).map(([chainId, unfilledAmount]) => {
        const usdThreshold = minimumDepositConfirmationThresholds.find((_usdThreshold) => {
          return toBNWei(_usdThreshold).gte(unfilledAmount);
```

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
nicholaspai: Yeah its cleaner good call out

PR #338, opened 2022-10-19
improve(ProfitClient): Always update gas costs

pxrl:
This relieves the the new profitability logic from needing to know whether gas costs were retrieved.


pxrl:
> This won't make the dataworker disputer or executor update this client right?

Correct - the ProfitClient is only instantiated for the dataworker when the proposer function is enabled, so the disputer and executor won't be impacted here.

The proposer will ultimately avoid it too, once it switches from ProfitClient to PriceClient.

PR #339, opened 2022-10-21
chore: Bump sdk-v2 ^0.1.21 -> ^0.1.30

pxrl:
Gives us access to some new toys.


pxrl:
FWIW have tested both relayer and dataworker (PROPOSER_ENABLED=true, DISPUTER_ENABLED=true). No regressions or side-effects observed, but will await resolution to the existing deployment issue before merging this one.

mrice32 (approved):

PR #340, opened 2022-10-24
fix(relayer): Ignore deposits with quote time in future

nicholaspai:
Adds a deposit quote timestamp buffer to protect relayer from edge case where a quote time is > HEAD's latest block. Currently, the relayer is exposed to the risk that HubPool utilization changes between `now` and the eventual block mined at that timestamp, since the ConfigStoreClient.computeRealizedLpFee returns the current lpFee % for quote times >HEAD

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -36,11 +36,19 @@ export class Relayer {
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
 
+    // Remove deposits whose deposit quote timestamp is > HubPool's current time, because there is a risk that
+    // the ConfigStoreClient's computed realized lp fee % is incorrect for quote times in the future. The client
+    // would use the current utilization as an input to compute this fee %, but if the utilization is different for the
+    // actual block that is mined at the deposit quote time, then the fee % would be different.
+    const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
+
     // Sum the total unfilled deposit amount per origin chain and set a MDC for that chain.
     const unfilledDepositAmountsPerChain: { [chainId: number]: BigNumber } = getUnfilledDeposits(
       this.clients.spokePoolClients,
       this.maxUnfilledDepositLookBack
-    ).reduce((agg, curr) => {
+    )
+    .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
+    .reduce((agg, curr) => {
       const unfilledAmountUsd = this.clients.profitClient.getFillAmountInUsd(curr.deposit, curr.unfilledAmount);
       if (!agg[curr.deposit.originChainId]) agg[curr.deposit.originChainId] = toBN(0);
       agg[curr.deposit.originChainId] = agg[curr.deposit.originChainId].add(unfilledAmountUsd);
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -86,10 +86,9 @@ export class Relayer {
 
     // Require that all fillable deposits meet the minimum specified number of confirmations.
     const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack)
-      .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
       .filter((x) => {
         return (
-          x.deposit.quoteTimestamp <= latestHubPoolTime &&
+          x.deposit.quoteTimestamp + this.config.quoteTimeBuffer <= latestHubPoolTime &&
           x.deposit.originBlockNumber <=
             this.clients.spokePoolClients[x.deposit.originChainId].latestBlockNumber -
               mdcPerChain[x.deposit.originChainId]
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -47,13 +47,13 @@ export class Relayer {
       this.clients.spokePoolClients,
       this.maxUnfilledDepositLookBack
     )
-    .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
-    .reduce((agg, curr) => {
-      const unfilledAmountUsd = this.clients.profitClient.getFillAmountInUsd(curr.deposit, curr.unfilledAmount);
-      if (!agg[curr.deposit.originChainId]) agg[curr.deposit.originChainId] = toBN(0);
-      agg[curr.deposit.originChainId] = agg[curr.deposit.originChainId].add(unfilledAmountUsd);
-      return agg;
-    }, {});
+      .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
+      .reduce((agg, curr) => {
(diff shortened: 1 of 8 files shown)


src/relayer/Relayer.ts
```diff
@@ -36,16 +36,26 @@ export class Relayer {
     // TODO: Note this does not consider the price of the token which will be added once the profitability module is
     // added to this bot.
 
+    // Remove deposits whose deposit quote timestamp is > HubPool's current time, because there is a risk that
+    // the ConfigStoreClient's computed realized lp fee % is incorrect for quote times in the future. The client
+    // would use the current utilization as an input to compute this fee %, but if the utilization is different for the
+    // actual block that is mined at the deposit quote time, then the fee % would be different. This should not
+    // impact the bridge users' UX in the normal path because deposit UI's have no reason to set quote times in the
+    // future.
+    const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
```
mrice32, line 45: I think this is great since currentTime comes directly from mainnet HEAD, which means that we have block data for this timestamp or earlier.

I think as a follow-up, we should try to use a block buffer similar to what we do for deposits so that mainnet reorgs don't cause problems. Thoughts?

src/relayer/Relayer.ts
```diff
@@ -36,16 +36,26 @@ export class Relayer {
...
       this.clients.spokePoolClients,
       this.maxUnfilledDepositLookBack
-    ).reduce((agg, curr) => {
-      const unfilledAmountUsd = this.clients.profitClient.getFillAmountInUsd(curr.deposit, curr.unfilledAmount);
-      if (!agg[curr.deposit.originChainId]) agg[curr.deposit.originChainId] = toBN(0);
-      agg[curr.deposit.originChainId] = agg[curr.deposit.originChainId].add(unfilledAmountUsd);
-      return agg;
-    }, {});
+    )
+      .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
```
mrice32, line 52: Doesn't this just filter the _amounts_? Don't we need to add this filter on line 91 so these deposits don't end up in `unfilledDeposits`?

mrice32 (commented):
This is great! Thank you for adding this. A few comments

src/relayer/Relayer.ts
nicholaspai: i'm curious why we need a block buffer at all, because once a timestamp has been mined, its utilization and corresponding realized lp fee % is locked in right? Unless you're suggesting that we should account for the chance that the block returned by `getBlockForTimestamp` is not actually mined yet or incorrect.

src/relayer/Relayer.ts
nicholaspai: good point done

src/relayer/Relayer.ts
mrice32: Note: I think this PR is good _as is_. The following explains why we might want to consider using buffers _in the future_. No need to do any of this here.

I think the issue is similar to deposit reorgs.

Let's say that initially the HEAD block is block A with timestamp t, where t = quoteTime (so it would pass the check).

We then use block A's state to compute the fee.

Then a reorg happens causing A to be replaced with B. B has the same timestamp, but has a different state, causing the fee to be different. We initially relay with the fee compute from A, but later, other relayers/dataworkers will see B and say that the fee we computed was wrong and our relay was invalid.

A buffer would prevent this case because A would be far enough into the past to be somewhat finalized and have a very low risk of reorg.

Note: the above situation is unlikely, probably more unlikely than the reorgs that cause deposits to swap deposit ids. It's unlikely because you would need three things to happen at once:
1. A hub pool utilization change (via some mainnet transaction).
2. A mainnet reorg that actually moves that transaction between blocks (not just changes the ordering within a block).
3. A spoke deposit.

However, I think it's still possible, so we should probably use similar finality buffers as we do elsewhere for mainnet, which can be counteracted by users using quote times that are delayed by the same amount or more.

src/relayer/Relayer.ts
```diff
@@ -76,13 +76,23 @@ export class Relayer {
...
+    // impact the bridge users' UX in the normal path because deposit UI's have no reason to set quote times in the
+    // future.
+    const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
+
     // Require that all fillable deposits meet the minimum specified number of confirmations.
     const unfilledDeposits = getUnfilledDeposits(this.clients.spokePoolClients, this.maxUnfilledDepositLookBack)
+      .filter((x) => x.deposit.quoteTimestamp <= latestHubPoolTime)
       .filter((x) => {
         return (
+          x.deposit.quoteTimestamp <= latestHubPoolTime &&
```
mrice32, line 92: Looks like we're filtering for the same thing in both of these statements? Should we remove one?

mrice32 (commented):
One remaining comment!

src/relayer/Relayer.ts
```diff
@@ -76,13 +76,23 @@ export class Relayer {
       minDepositConfirmations: this.config.minDepositConfirmations,
     });
 
+    // Remove deposits whose deposit quote timestamp is > HubPool's current time, because there is a risk that
+    // the ConfigStoreClient's computed realized lp fee % is incorrect for quote times in the future. The client
+    // would use the current utilization as an input to compute this fee %, but if the utilization is different for the
+    // actual block that is mined at the deposit quote time, then the fee % would be different. This should not
+    // impact the bridge users' UX in the normal path because deposit UI's have no reason to set quote times in the
+    // future.
+    const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
```
kevinuma, line 85: Is this the right timestamp to use? We have a following distance here so the actual timestamp of the deposits being reviewed is < latest hubpool time - the following distance.

src/relayer/Relayer.ts
nicholaspai: This a really interesting edge case I hadn't thought about, so do you think a buffer of say 2 blocks (~24 seconds in the quote time) could be combined with this [change in the front end](https://github.com/across-protocol/frontend-v2/pull/383/files) to make the relayers safer whilst not impacting bridge user's experienced bridge time?

src/relayer/Relayer.ts
nicholaspai: yes oops my mistake!

src/relayer/Relayer.ts
kevinuma: There's none but we have a following distance on the SpokePoolEvent. This could lead to an inconsistency. May be not a critical one but just feels odd to me when thinking about it:
1. A deposit happens but sets the quoteTimestamp to 5m into the future
2. A few mins later, after the following distance is satisfied, the relayer sees the deposit.
3. However the quote timestamp is still ahead of what the timestamp of the events the relayer is seeing.

nicholaspai: the issue stems specifically where quoteTime is >= hubPool.latestTime so as long as we can force relayer to ignore deposit quote times in the future we should be safe, does that answer your question?

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
mrice32: Yeah, I agree that a matching time buffer is probably good!

PR #341, opened 2022-10-24
logging(relayer): Update logging of invalid fills from warn to error

kevinuma:
So these errors are more noticeable


mrice32 (approved):

nicholaspai:
I'm not sure these should be errors, how is this going to affect our current PD alerts?

kevinuma:
> I'm not sure these should be errors, how is this going to affect our current PD alerts?

I don't think it would affect anything actually since we don't have PD set up for relayer. I'll close this PR for now and we can revisit in the future

PR #342, opened 2022-10-24
improve(ProfitClient): Add relayer gas multiplier

pxrl:
Default to 1.2, to allow some headroom for gas price changes between
when the profitability is calculated and when the txn is submitted.

This won't impact the behaviour of the RL bot, but will make any third-
party bots more conservative.

One thing to note is that this change introduces a single, global multiplier.
The SDK has a per-chain configurable multiplier...so there is some difference
with the SDK, which might be relevant, since it typically computes the fees.


nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #343, opened 2022-10-25
feat(readme): Improve README

nicholaspai:
In preparation for first release, introduces proposed process for publishing future releases and a more specific branching model.


README.md
```diff
@@ -33,22 +60,70 @@ When running the bot, be sure to set the following environment variable
...
+Here's the official Across [bug bounty program](https://docs.across.to/v2/miscellaneous/bug-bounty).
+
+# Branching Model
+
+## Active Branches
+
+| Branch          | Status                                                                           |
+| --------------- | -------------------------------------------------------------------------------- |
+| [master](https://github.com/across-protocol/relayer-v2/tree/master)                   | Accepts PRs from `develop` when we intend to deploy to mainnet.                                      |
+| [develop](https://github.com/across-protocol/relayer-v2/tree/develop)                 | Accepts PRs that are compatible with `master` OR from `release/X.X.X` branches.                    |
```
mrice32, line 86: What's the purpose of develop and master _when we have github releases_? Don't they kind of do the same thing?

The develop/master flow:
1. Create PR to merge develop into master.
2. Create release x.y.z. Tag x.y.z is created that always points to the commit in 1.
3. Do some more work on develop.
4. Go to 1.

The flow without develop:
1. Do work on master.
2. Create release x.y.z when ready to "release" this work. This release will create tag x.y.z which will always point to this commit.

The latter is simpler, so I would lean towards that unless we have a reason to prefer having a develop branch that is distinct from master.

README.md
```diff
@@ -33,22 +60,70 @@ When running the bot, be sure to set the following environment variable
...
+| release/X.X.X                                                                          | Accepts PRs for all changes, particularly those not backwards compatible with `develop` and `master`. |
+
+## Overview
+
+We generally follow [this Git branching model](https://nvie.com/posts/a-successful-git-branching-model/).
+Please read the linked post if you're planning to make frequent PRs into this repository (e.g., people working at/with UMA).
```
mrice32, line 92:
```suggestion
Please read the linked post if you're planning to make frequent PRs into this repository (e.g., people working at/with Risk Labs).
```

README.md
nicholaspai: Yeah I think we should just do work on master, i'll reduce this language a bit

README.md
nicholaspai: removed this line

README.md
```diff
@@ -33,22 +61,58 @@ When running the bot, be sure to set the following environment variable
...
+
+We publish pre-releases to signal to users about potential releases that are risky to use in production setups.
+
+## Documenting changes
+
+Every merged PR into `master` should be following the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format, as documented in [CONTRIBUTING.md](https://github.com/UMAprotocol/protocol/blob/master/CONTRIBUTING.md).
+
+# License
+
+All files within this repository are licensed under the [TODO](TODO) unless stated otherwise.
```
pxrl, line 118: This contradicts our package.json:

`"license": "AGPL-3.0",`

README.md
```diff
@@ -14,14 +23,33 @@ yarn build
 
 # run test suite
 yarn test
+
+# apply stylistic changes (e.g. eslint and prettier)
+yarn lint-fix
+
+# (for developers) continuously watch for changes and rebuild TS files as required
+yarn watch
```
pxrl, line 31: This isn't required for installation, and any users who are not familiar with the npm package workflow will probably think they need to run it.

What do you think about relocating it under a "Developers" section? Or attaching it to the Contributing section?

README.md
```diff
@@ -14,14 +23,33 @@ yarn build
...
 ```
 
-# How to run a relayer using this code
+# Integration tests
+
+You can conveniently run the dataworker, relayer, and finalizer functions via the hardhat task `integration-tests` which sets safe configurations like `PROPOSER_ENABLED=false` and prevents the user from accidentally sending an on-chain transaction. The test will complete successfully if no functions throw an error, which can be used as an additional source of confidence (in addition to the unit tests) that code has not broken the production bots.
```
pxrl, line 36: Should we add a brief section (or refer off somewhere else) describing each of the dataworker, relayer and finalizer functions?

README.md
```diff
@@ -14,14 +23,33 @@ yarn build
...
+
+# apply stylistic changes (e.g. eslint and prettier)
+yarn lint-fix
+
+# (for developers) continuously watch for changes and rebuild TS files as required
+yarn watch
 ```
 
-# How to run a relayer using this code
+# Integration tests
```
pxrl, line 34: Presume that this is intended for developers/contributors, rather than users?

README.md
```diff
@@ -14,14 +23,33 @@ yarn build
...
+
+You can conveniently run the dataworker, relayer, and finalizer functions via the hardhat task `integration-tests` which sets safe configurations like `PROPOSER_ENABLED=false` and prevents the user from accidentally sending an on-chain transaction. The test will complete successfully if no functions throw an error, which can be used as an additional source of confidence (in addition to the unit tests) that code has not broken the production bots.
+
+```sh
+# Run with gckms keys
+LOG_IN_TEST=true yarn hardhat integration-tests --wallet gckms --keys bot2
```
pxrl, line 40: Consider removing the part about gckms until we have a proper description about how to use it?

README.md
nicholaspai: Added a link to the `LICENSE` file actually, forgot that was there

README.md
nicholaspai: Good point let's just remove it and move to contributing

README.md
nicholaspai: Good point, linked to docsite

README.md
nicholaspai: Good point, i've reorganized a bit and now have the following sections:

- How to run a relayer
- Community
- License
- Developers

README.md
nicholaspai: Great idea

pxrl (approved):
This reads well and doesn't leave me with any questions. Nice work!

tasks/integration-tests.ts
```diff
@@ -32,6 +32,7 @@ task("integration-tests", "Run integration tests against key Dataworker function
 
     // Override environment variables that we want to control for, for example we want to prevent the runner of this
     // script from sending any on-chain transactions.
+    process.env.POLLING_DELAY = "0";
```
kevinuma, line 35: Why override this to 0 now?

kevinuma (approved):

tasks/integration-tests.ts
nicholaspai: Because the integration tests will never finish if the polling delay isn't set to 0 (i.e. it will keep looping) i'll add a comment

mrice32 (approved):

PR #344, opened 2022-10-25
feat: add publish ci workflow

mrice32:
This should mean that a new release will trigger a new npm package publish.


nicholaspai (approved):

pxrl:
Great! Do we have any chance of supporting pre-releases?

mrice32:
> Great! Do we have any chance of supporting pre-releases?
> 
> Also, do we potentially lack a section in our package.json? This is included in sdk-v2/package.json:
> 
> ```
>   "publishConfig": {
>     "registry": "https://registry.npmjs.com/",
>     "access": "public"
>   },
> ```

Good call -- added.

No it will not support pre-releases. We will need new functionality for that.

pxrl (approved):
LGTM. I'm learning about Actions and I have a rough idea of how to adapt this to support pre-releases, but let's just get this in as-is.

PR #345, opened 2022-10-25
improve(relayer): Skip approvals in test

pxrl:
When sending relays is disabled, don't bother checking for approvals.

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -98,9 +98,8 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
   ]);
 
   // Update the token client after the inventory client has done its wrapping of L2 ETH to ensure latest WETH ballance.
+  // The token client needs route data, so wait for update before checking approvals.
   await clients.tokenClient.update();
-  if (config.sendingRelaysEnabled) {
-    // Run approval check after updating token clients as needs route data.
+  if (config.sendingRelaysEnabled)
     await clients.tokenClient.setOriginTokenApprovals();
-  }
 }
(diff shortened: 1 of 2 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -99,5 +99,8 @@ export async function updateRelayerClients(clients: RelayerClients) {
 
   // Update the token client after the inventory client has done its wrapping of L2 ETH to ensure latest WETH ballance.
   await clients.tokenClient.update();
-  await clients.tokenClient.setOriginTokenApprovals(); // Run approval check  after updating token clients as needs route data.
+  if (config.sendingRelaysEnabled) {
```
nicholaspai, line 102: nit: can make this one line and remove the brackets. Should still keep the comment though

nicholaspai (approved):
Major upgrade incoming 🚨

Really we needed this a long time ago

src/relayer/RelayerClientHelper.ts
pxrl: I couldn't find a way to word the comment and keep it all under 120 chars, which is why the braces were there. I've since relocated the comment.

pxrl:
This _should_ eliminate the weird situation where the relayer complains about missing token approvals, even though SEND_RELAYS != true. The motivation for this is the discussion [here](https://github.com/across-protocol/relayer-v2/pull/219#discussion_r924995182). Hopefully this change makes it easier for users to simulate relayer behaviour without needing to mess with approvals.

nb I haven't touched the token approvals that are handled by the InventoryClient, because that's guarded/contingent upon Inventory Management being configured.

mrice32 (approved):

PR #346, opened 2022-10-25
fix(ProfitClient): Avoid ethers underflow

pxrl:
Straight to the wall of shame...

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -80,7 +80,7 @@ export class ProfitClient {
   ) {
     // Require .01e18 <= gasMultiplier <= 4e18
     assert(
-      this.gasMultiplier.div(toBNWei(".01")).gte(1) && this.gasMultiplier.lte(toBNWei(4)),
+      this.gasMultiplier.gte(toBNWei(".01")) && this.gasMultiplier.lte(toBNWei(4)),
       `Gas multiplier out of range (${this.gasMultiplier})`
     );
 
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -78,9 +78,9 @@ export class ProfitClient {
     readonly debugProfitability: boolean = false,
     protected gasMultiplier: BigNumber = toBNWei(1)
   ) {
-    // Require .01e18 <= gasMultiplier <= 4e18
+    // Require 0% < gasMultiplier <= 400%
     assert(
-      this.gasMultiplier.gte(toBNWei(".01")) && this.gasMultiplier.lte(toBNWei(4)),
+      this.gasMultiplier.gt(0) && this.gasMultiplier.lte(toBNWei(4)),
       `Gas multiplier out of range (${this.gasMultiplier})`
     );
 
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -19,7 +19,7 @@ export type FillProfit = {
   fillAmountUsd: BigNumber; // Amount of the bridged token being filled.
   grossRelayerFeeUsd: BigNumber; // USD value of the relay fee paid by the user.
   nativeGasCost: BigNumber; // Cost of completing the fill in the native gas token.
-  gasMultiplier: number; // Multiplier to apply to nativeGasCost as padding or discount
+  gasMultiplier: BigNumber; // Multiplier to apply to nativeGasCost as padding or discount
   gasPriceUsd: BigNumber; // Price paid per unit of gas in USD.
   gasCostUsd: BigNumber; // Estimated cost of completing the fill in USD.
   relayerCapitalUsd: BigNumber; // Amount to be sent by the relayer in USD.
@@ -58,7 +58,6 @@ export class ProfitClient {
   private readonly priceClient;
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
   private unprofitableFills: { [chainId: number]: { deposit: Deposit; fillAmount: BigNumber }[] } = {};
-  private _gasMultiplier: BigNumber;
 
   // Track total gas costs of a relay on each chain.
   protected totalGasCosts: { [chainId: number]: BigNumber } = {};
@@ -75,11 +74,16 @@ export class ProfitClient {
     readonly enabledChainIds: number[],
     // Default to throwing errors if fetching token prices fails.
     readonly ignoreTokenPriceFailures: boolean = false,
-    readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT),
+    readonly minRelayerFeePct: BigNumber = toBNWei(constants.RELAYER_MIN_FEE_PCT),
     readonly debugProfitability: boolean = false,
-    gasMultiplier = 1.0
+    protected gasMultiplier: BigNumber = toBNWei(1)
   ) {
-    this.gasMultiplier = gasMultiplier;
+    // Require .01e18 <= gasMultiplier <= 4e18
+    assert(
+      this.gasMultiplier.div(toBNWei(".01")).gte(1) && this.gasMultiplier.lte(toBNWei(4)),
+      `Gas multiplier out of range (${this.gasMultiplier})`
+    );
+
     this.priceClient = new PriceClient(logger, [
       new acrossApi.PriceFeed("Across API", {}),
       new coingecko.PriceFeed("CoinGecko Free", {}),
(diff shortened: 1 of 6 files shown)


src/clients/ProfitClient.ts
```diff
@@ -125,7 +125,10 @@ export class ProfitClient {
       throw new Error(`Unable to compute gas cost (${err} unknown)`);
     }
 
-    const gasCostUsd = nativeGasCost.mul(this.gasMultiplier).mul(gasPriceUsd).div(toBN(10).pow(GAS_TOKEN_DECIMALS));
+    const gasCostUsd = nativeGasCost
+      .mul(this.gasMultiplier.toString())
```
pxrl, line 129: FWIW I think it's a bit dumb to be doing a string conversion every single time, but I also don't really like storing gasMultiplier as a string in the ProfitClient class.

What's the lesser evil?

pxrl:
nb There's also an issue with the tests here, so I'll need to fix that as well.

src/clients/ProfitClient.ts
nicholaspai: Third door: I kind of think you should be storing it as a BN...

nicholaspai (approved):
One suggestion

src/clients/ProfitClient.ts
```diff
@@ -92,6 +94,16 @@ export class ProfitClient {
...
+  set gasMultiplier(gasMultiplier: number) {
+    assert(gasMultiplier > 0.0 && gasMultiplier <= 2.0, `Gas multiplier out of range: ${gasMultiplier}`);
+    this._gasMultiplier = toBNWei(gasMultiplier);
+  }
+
+  get gasMultiplier(): number {
+    // Preserve at most 2 decimal places
+    return this._gasMultiplier.div(toBNWei("0.01")).toNumber() / 100;
+  }
+
```
mrice32, line 106: Love the use of setters and getters. This is really nice to keep the external and internal view of the variable different.

mrice32 (approved):
LGTM!

src/clients/ProfitClient.ts
```diff
@@ -19,6 +19,7 @@ export type FillProfit = {
   fillAmountUsd: BigNumber; // Amount of the bridged token being filled.
   grossRelayerFeeUsd: BigNumber; // USD value of the relay fee paid by the user.
   nativeGasCost: BigNumber; // Cost of completing the fill in the native gas token.
+  gasMultiplier: number; // Multiplier to apply to nativeGasCost as padding or discount
```
nicholaspai, line 22: I think this should be `BigNumber` type

src/clients/ProfitClient.ts
```diff
@@ -57,6 +58,7 @@ export class ProfitClient {
   private readonly priceClient;
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
   private unprofitableFills: { [chainId: number]: { deposit: Deposit; fillAmount: BigNumber }[] } = {};
+  private _gasMultiplier: BigNumber;
```
nicholaspai, line 61: I think we can remove this

src/clients/ProfitClient.ts
```diff
@@ -74,8 +76,10 @@ export class ProfitClient {
     // Default to throwing errors if fetching token prices fails.
     readonly ignoreTokenPriceFailures: boolean = false,
     readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT),
-    readonly debugProfitability: boolean = false
+    readonly debugProfitability: boolean = false,
+    gasMultiplier = 1.0
```
nicholaspai, line 80: We could make this `readonly` and make user pass in BN

nicholaspai (commented):
Left some comments on what type we store in the client

src/clients/ProfitClient.ts
```diff
@@ -92,6 +94,16 @@ export class ProfitClient {
     }
   }
 
+  set gasMultiplier(gasMultiplier: number) {
+    assert(gasMultiplier > 0.0 && gasMultiplier <= 2.0, `Gas multiplier out of range: ${gasMultiplier}`);
```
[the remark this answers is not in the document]
nicholaspai: in `RelayerConfig` you allow this to be equal to 0. I think you should require > 0.0 and assert that only in this client, which should throw an error upon the Relayer booting up if the config is set to 0.0 right?

src/clients/ProfitClient.ts
nicholaspai, line 80: As mentioned offline, I'm curious why we can't make the type and usage of `gasMultiplier` similar to `minRelayerFeePct`, since they are both intuitively percentages?

pxrl: I'd prefer it if we didn't force ProfitClient users to know about the inner details/constraints/quirks of the ProfitClient. I'm not really sure that this was considered for the existing minRelayerFeePct, but given that it's already here, I'm fine with adjusting this implementation to be more consistent with that. If it was being implemented from scratch I'd probably be more invested in achieving a simpler interface, though.

src/clients/ProfitClient.ts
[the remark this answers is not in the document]
pxrl: One quirk of moving to a BigNumber as input is that it seems harder to enforce sane inputs, because the caller might supply toBN(1) (i.e. 000000000000000001). I've instead enforced it to be .01e18 <= gasMultiplier <= 4e18, and have applied that in the ProfitClient class. I think that should cover our foreseeable use cases.

I would normally like to give the fastest feedback possible on this, and that would mean implementing the check in relayerConfig, but the ProfitClient should complain loudly if a caller supplies a really dumb gasMultiplier (i.e. negative), and we probably don't want to duplicate the checks. We do also instantiate the ProfitClient almost immediately on startup, so we will anyway fail _extremely_ quickly in the ProfitClient.

src/clients/ProfitClient.ts
```diff
@@ -73,9 +74,16 @@ export class ProfitClient {
     readonly enabledChainIds: number[],
     // Default to throwing errors if fetching token prices fails.
     readonly ignoreTokenPriceFailures: boolean = false,
-    readonly minRelayerFeePct: BigNumber = toBN(constants.RELAYER_MIN_FEE_PCT),
-    readonly debugProfitability: boolean = false
+    readonly minRelayerFeePct: BigNumber = toBNWei(constants.RELAYER_MIN_FEE_PCT),
```
pxrl, line 77: FWIW, the existing toBN() was an error, so I fixed it whilst I was in the neighbourhood.

src/clients/ProfitClient.ts
```diff
@@ -73,9 +74,16 @@ export class ProfitClient {
...
+    readonly debugProfitability: boolean = false,
+    protected gasMultiplier: BigNumber = toBNWei(1)
   ) {
+    // Require .01e18 <= gasMultiplier <= 4e18
+    assert(
+      this.gasMultiplier.gte(toBNWei(".01")) && this.gasMultiplier.lte(toBNWei(4)),
```
nicholaspai, line 83: why not `.gt(0)`?

src/clients/ProfitClient.ts
nicholaspai: How did you know this was an error?

src/clients/ProfitClient.ts
nicholaspai: Agreed but since we're adding a constructor var in this PR lets be consistent a separate PR should redo the full interface

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -125,10 +125,35 @@ describe("ProfitClient: Consider relay profit", async function () {
       const gasPriceUsd = tokenPrices[gasToken.symbol];
       expect(gasPriceUsd.eq(tokenPrices[gasToken.symbol])).to.be.true;
 
-      const estimate: { [key: string]: BigNumber } = profitClient.calculateFillCost(chainId);
+      const estimate: { [key: string]: BigNumber } = profitClient.estimateFillCost(chainId);
       expect(estimate.nativeGasCost.eq(gasCost[chainId])).to.be.true;
       expect(estimate.gasPriceUsd.eq(tokenPrices[gasToken.symbol])).to.be.true;
-      expect(estimate.gasCostUsd.eq(gasPriceUsd.mul(nativeGasCost).div(toBN(10).pow(gasToken.decimals)))).to.be.true;
+      expect(estimate.gasCostUsd.eq(gasPriceUsd.mul(nativeGasCost).div(toBNWei(1)))).to.be.true;
```
nicholaspai, line 131: Why make this change? This only works because `decimals = 18` right?

src/clients/ProfitClient.ts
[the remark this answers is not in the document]
nicholaspai: I can see users who'd want to set a sub 100%. If they want to ignore gas costs entirely, then they can set IGNORE_PROFITABILITY=true. I think since 400% is arbitrary, we can set a lower bound of 25% arbitrarily.

src/clients/ProfitClient.ts
pxrl: See the corresponding line in RelayerConfig.ts: https://github.com/across-protocol/relayer-v2/blob/77c91776597f0d70c9588bf53282ec96cf58061e/src/relayer/RelayerConfig.ts#L54

test/ProfitClient.ConsiderProfitability.ts
pxrl: That's an error - thanks for catching it. Have a PR to undo it here: https://github.com/across-protocol/relayer-v2/pull/352.

PR #347, opened 2022-10-26
fix(ProfitClient): Fix tests

pxrl:
expect() was being used incorrectly, with the potential to suppress errors. One suppressed error was found, but the test code was found to be at fault - not the implementation.

This error was uncovered via https://github.com/across-protocol/relayer-v2/pull/346. The relevant gas multiplier test will be corrected as part of that PR.

### DIFF
--- a/test/ProfitClient.ConsiderProfitability.ts
+++ b/test/ProfitClient.ConsiderProfitability.ts
@@ -116,19 +116,19 @@ describe("ProfitClient: Consider relay profit", async function () {
 
       const nativeGasCost = profitClient.getTotalGasCost(chainId);
       expect(nativeGasCost.eq(0)).to.be.false;
-      expect(nativeGasCost.eq(gasCost[chainId]));
+      expect(nativeGasCost).to.deep.equal(gasCost[chainId]);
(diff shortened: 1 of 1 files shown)


nicholaspai (commented):
Are you sure this is the fix? I have never had to use deepEqual with BNs before. Have you tried adding .to.equal(true) after the expect() statement?

pxrl:
> Are you sure this is the fix? I have never had to use deepEqual with BNs before. Have you tried adding .to.equal(true) after the expect() statement?

There are definitely multiple ways to do this, and `.to.deep.equal();` just one option. I think the alternative would be `expect(a.eq(b)).to.be.true;`. I shied away from that because it's easy to forget to append `.to.be.true;` (i.e. which is what I did), and that suppresses the error.

nicholaspai (approved):

nicholaspai:
I think this PR can be closed and replaced with #349

nicholaspai (changes requested):
https://github.com/across-protocol/relayer-v2/pull/347#issuecomment-1291310224

PR #349, opened 2022-10-26
improve: Add es-lint rules to prevent chai "expect" statements without assertions

nicholaspai:
Prevents the case where an `expect(...)` without an assertion (e.g. `to.be.true`) suppresses an error or contains something `false` in the inner expression. This is because `expect(...)` returns true if the inside statement is truthy which includes Errors and `false`


pxrl (approved):
LGTM; glad we caught this before it blew up!

PR #351, opened 2022-10-26
improve: typescript strict in src/dataworker/Dataworker.ts

mrice32:
This builds on #350 to add typescript strict to the Dataworker.ts file. I only updated that file since there were quite a few issues there and I figured it would be enough for a single PR.


src/dataworker/Dataworker.ts
```diff
@@ -377,19 +380,24 @@ export class Dataworker {
     submitDisputes = true,
     latestInvalidBundleStartBlock: { [chainId: number]: number } = {}
   ) {
-    if (!this.clients.hubPoolClient.isUpdated) throw new Error("HubPoolClient not updated");
+    if (
+      !this.clients.hubPoolClient.isUpdated ||
+      this.clients.hubPoolClient.currentTime === undefined ||
+      this.clients.hubPoolClient.latestBlockNumber === undefined
+    )
+      throw new Error("HubPoolClient not updated");
```
nicholaspai, line 388: Could be useful to have `isUpdated` be a class level function on `hubPoolClient`

src/dataworker/Dataworker.ts
```diff
@@ -452,15 +460,28 @@ export class Dataworker {
...
-    valid: boolean;
-    reason?: string;
-    expectedTrees?: {
-      poolRebalanceTree: TreeData<PoolRebalanceLeaf>;
-      relayerRefundTree: TreeData<RelayerRefundLeaf>;
-      slowRelayTree: TreeData<RelayData>;
-    };
-  }> {
+  ): Promise<
+    // If valid is false, we get a reason and we might get expected trees.
```
nicholaspai, line 464: This is cool didn't know you could have conditional types

src/utils/TypeGuards.ts
```diff
@@ -13,3 +13,15 @@ export function isPromiseRejected<T>(
...
+// This function allows you to test for the key type in an object literal.
+// For instance, this would compile in typescript strict:
+//   const myObj = { a: 1, b: 2, c: 3 } as const;
+//   const d: string = "a";
+//   const myNumber = isKeyOf(d, myObj) ? myObj[d] : 4;
+export function isKeyOf<T extends V, V extends number | string | symbol>(
+  input: V,
+  obj: Record<T, unknown>
+): input is T {
+  return input in obj;
```
nicholaspai, line 26: This is a nice use of TS

nicholaspai (approved):
Left some comments/suggestions but nothing blocking LGTM

src/dataworker/Dataworker.ts
```diff
@@ -377,19 +380,24 @@ export class Dataworker {
...
+      !this.clients.hubPoolClient.isUpdated ||
+      this.clients.hubPoolClient.currentTime === undefined ||
+      this.clients.hubPoolClient.latestBlockNumber === undefined
+    )
+      throw new Error("HubPoolClient not updated");
     const hubPoolChainId = (await this.clients.hubPoolClient.hubPool.provider.getNetwork()).chainId;
 
     // Exit early if a bundle is not pending.
-    if (this.clients.hubPoolClient.hasPendingProposal() === false) {
+    const pendingRootBundle = this.clients.hubPoolClient.getPendingRootBundle();
```
pxrl, line 392: Nice catch.

src/dataworker/Dataworker.ts
```diff
@@ -940,19 +965,24 @@ export class Dataworker {
       message: "Executing pool rebalance leaves",
     });
 
-    if (!this.clients.hubPoolClient.isUpdated) throw new Error("HubPoolClient not updated");
+    if (
+      !this.clients.hubPoolClient.isUpdated ||
+      this.clients.hubPoolClient.currentTime === undefined ||
+      this.clients.hubPoolClient.latestBlockNumber === undefined
```
pxrl, line 971: Consideration for follow-up - this seems to be a repeating pattern for HubPoolClient, so I wonder if it could be make sense to incorporate it directly into HubPoolClient to make it easier for callers to evalate.

src/utils/TypeGuards.ts
pxrl: Agree. I have wondered if it's more at home in the sdk, though, since we could also get some benefit from it there.

src/dataworker/Dataworker.ts
pxrl: Ah looks like @nicholaspai had the same thought here: https://github.com/across-protocol/relayer-v2/pull/351#discussion_r1006175685

Related discussion from a previous PR here: https://github.com/across-protocol/relayer-v2/pull/319#discussion_r992425862

src/dataworker/Dataworker.ts
mrice32: Yes! Unions like this are messy to write, but it makes life easier for the caller so typescript can infer a bunch of things once we check whether it's valid or not.

src/dataworker/Dataworker.ts
mrice32: @nicholaspai and @pxrl these are great comments!

This is where typescript makes life difficult. These extra assertions are meant to validate that _these specific fields_ are not undefined. If I don't validate that, typescript complains that these values "may be undefined". If we put it into a function in the Hub, this function won't be able to infer that they aren't undefined.

I would like to have some sort of way to help typescript understand that if isUpdated is true, then the others are never undefined, but I'm not sure how to do that. Do y'all have ideas?

src/dataworker/Dataworker.ts
mrice32: Reply is [here](https://github.com/across-protocol/relayer-v2/pull/351#discussion_r1007123092)

src/dataworker/Dataworker.ts
nicholaspai: Ah i understand now, nothing to do for now then. Perhaps we'd have to define custom types and have every client be known to typescript as "updated" or not

src/utils/TypeGuards.ts
nicholaspai: yeah I sort of agree except that updating utility functions from outside this repo is very annoying. Think about some of the test utils like `buildDeposit` that we import from `contracts-v2` that are a nightmare to deal with now

src/dataworker/Dataworker.ts
pxrl: I don't have anything to add to the above; this might be something to chew on over a longer period of time.

src/utils/TypeGuards.ts
pxrl: > yeah I sort of agree except that updating utility functions from outside this repo is very annoying.

Yeah; it can get a bit tricky - especially if it involves changing things that others depend on.

pxrl (approved):

src/dataworker/Dataworker.ts
mrice32: > Perhaps we'd have to define custom types and have every client be known to typescript as "updated" or not

Yeah, this is super interesting. I wish ts had some obvious pattern to do stuff like this with classes (like a tagged union).

PR #353, opened 2022-10-27
improve: Log package version on startup

pxrl:
Could alternatively be bundled into the global config within src/common/Config.ts, so it would be automagically logged by the respective "{Dataworker,Relayer} started" log messages.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -5,12 +5,14 @@ import { runRelayer } from "./src/relayer";
 import { runDataworker } from "./src/dataworker";
 import { runMonitor } from "./src/monitor";
 import { runFinalizer } from "./src/finalizer";
+import { version } from "./package.json";
 
 let logger: winston.Logger;
 
 export async function run(args: { [k: string]: boolean | string }): Promise<void> {
   logger = Logger;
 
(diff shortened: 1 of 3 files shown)


nicholaspai (commented):
I like logging this in the global config since I'd want to know the version for the dataworker run as well!

pxrl:
> I like logging this in the global config since I'd want to know the version for the dataworker run as well!

This logs unconditionally - so it is always available, no matter how the bot is started. It generates an additional log message, though.

I'll move it to global config.

index.ts
```diff
@@ -46,6 +47,8 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
 }
 
 if (require.main === module) {
+  // Inject version into process.env so CommonConfig and all subclasses inherit it.
+  process.env.VERSION = version;
```
pxrl, line 51: Hm, VERSION might be a bit too generic. Should probably prefix by something Across-specific.

pxrl:
FWIW one additional thing that I think we need is to make the git commit hash available to the bot, so it can be logged as well. That's a little bit more complex to implement though.

nicholaspai (approved):

PR #354, opened 2022-10-27
chore: Require dotenv 10.0.0 -> 16.0.x

pxrl:
This change formalises our dependency on dotenv 15.0.0, which is needed because our .env configuration syntax requires multi-line parsing for some configuration items, i.e.:
 - MIN_DEPOSIT_CONFIRMATIONS
 - RELAYER_INVENTORY_CONFIG

In practice, we seem to already have this version installed via some other dependencies, as this upgraded did not result in any change to yarn.lock.

Additionally, specify dotenv as a raw dependency, rather than a dev dependency, since non-developers are reliant on it.


pxrl:
One alternative I have been thinking about is to migrate configuration to a raw json file. Node can read those directly without any dependencies, and the syntax is less ugly for multi-line configurations.

There might be some specific considerations around the PRIVATE_KEY and MNEMONIC env vars, but we don't actually need dotenv for those (i.e. if they're specified directly in the environment, and not in .env, then they wouldn't be specified in any JSON configuration file either).

Thoughts?

nicholaspai (approved):
I would love to see configuration change to JSON, its a huge pain to specify nested JSON objects on a single string

PR #356, opened 2022-10-28
improve: Update Optimism finalizer SDK to latest minor version 1.6.X

nicholaspai:


src/finalizer/utils/optimism.ts
```diff
@@ -94,16 +85,16 @@ export async function getOptimismFinalizableMessages(
   );
 }
 
-export function getL1TokenInfoForOptimismToken(hubPoolClient: HubPoolClient, l2Token: string) {
-  return hubPoolClient.getL1TokenInfoForL2Token(SpokePoolClient.getExecutedRefundLeafL2Token(10, l2Token), CHAIN_ID);
+export function getL1TokenInfoForOptimismToken(hubPoolClient: HubPoolClient, l2Token: string): L1Token {
+  return hubPoolClient.getL1TokenInfoForL2Token(SpokePoolClient.getExecutedRefundLeafL2Token(10, l2Token), 10);
```
pxrl, line 89: Just curious - any reason to delete CHAIN_ID here? Seems like we could actually have used it more than we have -i.e. this call uses the chain ID twice.

Optimism currently only have one chain, but if this was Arbitrum One/Nova then we'd probably find a global chainId useful.

pxrl: Can also see that we have multiple uses of the chainId above - could be a candidate for a single global variable?

yarn.lock
```diff
@@ -2623,6 +2677,204 @@
...
+
+"@nomicfoundation/solidity-analyzer-win32-arm64-msvc@0.1.0":
+  version "0.1.0"
+  resolved "https://registry.yarnpkg.com/@nomicfoundation/solidity-analyzer-win32-arm64-msvc/-/solidity-analyzer-win32-arm64-msvc-0.1.0.tgz#9d4bca1cc9a1333fde985675083b0b7d165f6076"
+  integrity sha512-7x5SXZ9R9H4SluJZZP8XPN+ju7Mx+XeUMWZw7ZAqkdhP5mK19I4vz3x0zIWygmfE8RT7uQ5xMap0/9NPsO+ykw==
+
+"@nomicfoundation/solidity-analyzer-win32-ia32-msvc@0.1.0":
+  version "0.1.0"
+  resolved "https://registry.yarnpkg.com/@nomicfoundation/solidity-analyzer-win32-ia32-msvc/-/solidity-analyzer-win32-ia32-msvc-0.1.0.tgz#0db5bfc6aa952bea4098d8d2c8947b4e5c4337ee"
+  integrity sha512-m7w3xf+hnE774YRXu+2mGV7RiF3QJtUoiYU61FascCkQhX3QMQavh7saH/vzb2jN5D24nT/jwvaHYX/MAM9zUw==
```
pxrl, line 2855: Driveby comment...wow...didn't expect to see platform-specific dependencies popping up at this level. That's pretty ordinary ;( Probably not much we can do, though.

pxrl (approved):
Looks OK. There seem to be a lot of yarn.lock updates, but we're also doing a pretty big jump here. Hopefully Optimism have good control over their dependencies...

src/finalizer/utils/optimism.ts
nicholaspai: I think its useful if we want to use this finalizer easily between mainnet and test nets but there are many other setup changes we'd need to make so I don't see a reason to do this

yarn.lock
nicholaspai: hmmm pretty strange

mrice32 (approved):

src/finalizer/utils/optimism.ts
pxrl: I understand it's not a change to make all at once; would be nice if we could incrementally edge closer to that over time :+1:.

kevinuma (approved):

PR #357, opened 2022-10-28
feat(finalizer): Add 14 day default lookback to TokensBridged events

nicholaspai:
This will speed up finalizer from 10mins per run to 30s, and reduce web3 queries. It also means the performance will no longer be O(n) but O(1) where n = # of TokensBridged events


src/finalizer/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Wallet, config, startupLogLevel, processEndPollingLoop, processCrash } from "../utils";
```
pxrl, line 1: Was this a lint warning or something?

pxrl (commented):
No objections from me! (...FWIW)

src/clients/SpokePoolClient.ts
```diff
@@ -312,9 +312,8 @@ export class SpokePoolClient {
       if (!Object.keys(this._queryableEventNames()).includes(eventName))
         throw new Error("Unknown event to query in SpokePoolClient");
       let searchConfigToUse = searchConfig;
-      if (eventName === "EnabledDepositRoute" || eventName === "TokensBridged")
-        searchConfigToUse = depositRouteSearchConfig;
-      else if (eventName === "FundsDeposited" || eventName === "FilledRelay")
+      if (eventName === "EnabledDepositRoute") searchConfigToUse = depositRouteSearchConfig;
+      else if (eventName === "FundsDeposited" || eventName === "FilledRelay" || eventName === "TokensBridged")
```
mrice32, line 316: What's the change here? Can you add a quick comment as to why we're using each config for each set of events?

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
kevinuma: It looks like Nick just moved TokensBridged from using depositRouteSearchConfig to depositEventSearchConfig. Most likely because the route config looks through entirety of history which is much longer and thus heavily affect the finalizer. Switching to depositEventSearchConfig = shorter lookback for finalizer => Speedup

kevinuma (approved):

src/finalizer/index.ts
nicholaspai: no i just saw they were unused

src/clients/SpokePoolClient.ts
nicholaspai: I've added comments but this specific change I actually revertted, it wasn't necessary to make!

PR #358, opened 2022-10-31
improve: Add yarn script for starting relayer

pxrl:
There are a couple of benefits to this:
 - Make it simpler for others to start the relayer bot.
 - Avoid running tsc/ts-node each time.


pxrl:
> There are a couple of benefits to this:
> 
> * Make it simpler for others to start the relayer bot.
> * Avoid running tsc/ts-node each time.

Usage would now be:

`yarn relay --wallet <mnemonic|privateKey>`

nicholaspai:
Does this require a pre-step now to compile ts --> js?

pxrl:
> Does this require a pre-step now to compile ts --> js?

Implicitly it does, yeah. I believe running `yarn` should do this, though I haven't verified that yet. That is per our instructions in the docs, so if `yarn` behaves as anticipated then I'd argue it should be OK.

If the user is actively modifying the code then I'd expect that they should know they need to run tsc over it again before running it again.

```
# Clone relayer code with Github CLI or git clone.
gh repo clone across-protocol/relayer-v2

# Establish environment file and restrict filesystem permissions
touch ./relayer-v2/.env
chmod -R 0600 ./relayer-v2

# Install dependencies and build relayer code.
# Nodejs and yarn are required.
cd relayer-v2
yarn

# Run unit tests.
yarn test
```

mrice32 (approved):

mrice32 (commented):
I think a compilation step is 👍 as long as it's clearly documented!

pxrl:
Have confirmed that `yarn [install]` and `yarn build` are both needed. Documentation has been updated to make that explicit.

PR #359, opened 2022-10-31
fix: Use typecasting to fix optimism finalizer

nicholaspai:
New optimism sdk 1.6.x has a different Signer type than that exported by ethers.Signer that we use as dependency. I was able to run this code locally and it works in run time so the main issue seemed to be compile time. For example I finalized this optimism L2 message using this code: https://etherscan.io/tx/0x22e80cd98823c47f6949e08429f97cf6e0e95ee0ae793d7a931acb4ef435a333

### DIFF
--- a/src/finalizer/utils/optimism.ts
+++ b/src/finalizer/utils/optimism.ts
@@ -1,14 +1,27 @@
 import * as optimismSDK from "@eth-optimism/sdk";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
 import { L1Token, TokensBridged } from "../../interfaces";
-import { convertFromWei, delay, etherscanLink, getNodeUrlList, groupObjectCountsByProp, winston } from "../../utils";
+import {
+  convertFromWei,
+  delay,
+  ethers,
+  etherscanLink,
+  getNodeUrlList,
+  groupObjectCountsByProp,
+  Wallet,
+  winston,
+} from "../../utils";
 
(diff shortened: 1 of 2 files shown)


src/finalizer/utils/optimism.ts
```diff
@@ -1,14 +1,32 @@
...
   return new optimismSDK.CrossChainMessenger({
     l1ChainId: 1,
     l2ChainId: 10,
-    l1SignerOrProvider: optimismSDK.toSignerOrProvider(getNodeUrlList(1)[0]),
-    l2SignerOrProvider: optimismSDK.toSignerOrProvider(getNodeUrlList(10)[0]),
+    // @dev: The optimismSDK's version of Signer and Provider is a bit different than the one imported by
+    // this repo, specifically it claims:
+    // - Type 'Wallet' is not assignable to type 'Signer'.
+    // - The types returned by 'provider.getFeeData()' are incompatible between these types.
+    // - So we cast to unknown to ignore this incompatibility which doesn't seem important.
```
pxrl, line 23: I think this is symptomatic of an ethers mismatch. v5.7.x introduced a new member to the FeeData object.

In general we do want to jump to 5.7.x, so this might be a good trigger for doing that.

pxrl (approved):
Seems OK, though I think we should revisit this once we bump ethers to 5.7.x as a direct relayer-v2 dependency, because some of these issues might disappear.

PR #360, opened 2022-10-31
chore: Bump ethers versionto 5.7.x

nicholaspai:
ethers.Signer has breaking changes in 5.7.x so we should update to stay up to date with other deps that import ethers


package.json
```diff
@@ -77,7 +77,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "eslint-plugin-promise": "^5.1.0",
     "ethereum-waffle": "^3.0.0",
-    "ethers": "^5.6.9",
+    "ethers": "^5.7.0",
```
pxrl, line 80: Should we move this directly to dependencies?

nicholaspai: yeah probably

kevinuma (approved):

PR #366, opened 2022-11-07
chore: Bump sdk-v2

pxrl:
PriceClient becomes a bit smarter (can now handle partial price feed responses), as well as a little bit easier to use. As a side-effect, we now automagically use CoinGecko Pro if if the COINGECKO_PRO_API_KEY env var is defined, and otherwise gracefully fall back to the free tier.


nicholaspai (approved):
Excited for this!

PR #368, opened 2022-11-10
chore(finalizer): Increase lookback for Optimism

pxrl:
The existing config isn't long enough to catch withdrawals on Optimism. This is
a band-aid to allow withdrawals to resume, to be addressed properly once we
understand more.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -4,22 +4,22 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
 
 export const RELAYER_MIN_FEE_PCT = 0.0003;
 
-// Target ~2 days per chain. Avg. block times: { 1: 12s, 10/42161: 0.5s, 137: 2.5s, 288: 30s }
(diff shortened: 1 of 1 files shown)


src/common/Constants.ts
```diff
@@ -26,10 +26,10 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
 // rollup challenge period seems safe.
 export const FINALIZER_TOKENBRIDGE_LOOKBACK: { [chainId: number]: number } = {
   1: 100800,
-  10: 2419200,
+  10: 4838400,
```
nicholaspai, line 29: On line 7 can you change comment to show 0.25s per block for 10/42161 and double the lookbacks on `MAX_RELAYER_DEPOSIT_LOOK_BACK` and `DATAWORKER_FAST_LOOKBACK` to be safe

nicholaspai (approved):
One comment

nicholaspai (approved):

PR #370, opened 2022-11-18
fix(balance-allocator): Count WETH and ETH for OVM spoke chains

nicholaspai:
Ovm_SpokePool will deposit ETH into WETH before executing any leaves, so both the ETH and WETH balances should be counted together when deciding whether a leaf can be executed. Without this logic, the dataworker mistakenly skips executing leaves that it does have balance for but the balance is split between eth and weth

### DIFF
--- a/src/clients/BalanceAllocator.ts
+++ b/src/clients/BalanceAllocator.ts
@@ -40,7 +40,6 @@ export class BalanceAllocator {
           )
         );
 
-        // Replace `tokens` property with one of the interchangeable tokens.
         const returnedRequest = {
           ...request,
           balances,
--- a/src/clients/BalanceAllocator.ts
+++ b/src/clients/BalanceAllocator.ts
@@ -1,8 +1,5 @@
 import { BigNumber, ERC20, ethers, ZERO_ADDRESS } from "../utils";
 
-const ovmEthAddresses = ["0x4200000000000000000000000000000000000006", "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000"];
-const ovmChainIds = [10, 288];
-
 // This type is used to map used and current balances of different users.
 interface BalanceMap {
   [chainId: number]: {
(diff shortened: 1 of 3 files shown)


src/clients/BalanceAllocator.ts
```diff
@@ -49,10 +52,18 @@ export class BalanceAllocator {
 
   async getBalance(chainId: number, token: string, holder: string) {
     if (!this.balances?.[chainId]?.[token]?.[holder]) {
-      const balance =
-        token === ZERO_ADDRESS
-          ? await this.providers[chainId].getBalance(holder)
-          : await ERC20.connect(token, this.providers[chainId]).balanceOf(holder);
+      let balance: BigNumber;
```
mrice32, line 55: Would it be possible for the caller to deal with this special case? This seems like it's only needed in a particular callsite, and it breaks the abstraction of the client.

mrice32 (commented):
One question!

src/clients/BalanceAllocator.ts
nicholaspai: One way we could do this is to allow the caller to set "interchangeable tokens" per chain, then the BalanceAllocator could combine balances for tokens in this list

nicholaspai: I've implemented this: the call to `requestBalanceAllocation` now takes in `tokens: string[]` instead of `token: string` and treats all `tokens` as having fungible balance

src/clients/BalanceAllocator.ts
```diff
@@ -16,35 +17,68 @@ export class BalanceAllocator {
...
+            request.tokens.map(
+              async (token): Promise<[string, BigNumber]> => [
+                token,
+                await this.getBalance(request.chainId, token, request.holder),
+              ]
+            )
+          )
+        );
+
+        // Replace `tokens` property with one of the interchangeable tokens.
```
nicholaspai, line 43:
```suggestion
```

This is no longer true right?

PR #375, opened 2022-11-24
refactor: Remove Fill/Deposit cache from SpokePool client

nicholaspai:
## Motivation
We originally added the FIll/Deposit cache to help the Dataworker look with very long lookbacks in order to evaluate historical bundles.

We ended up not using this logic by fixing the dataworker's lookback, which made the time to query fresh Fill/Deposit events very small, rendering the cache not as useful.

Additionally it is dangerous to cache potentially bad data from an RPC, especially if all of the dataworker functions share the same cache.

The ideal scenario for loading events for a particular bundle would be that the Dataworker knows how to evaluate one root bundle at a time, and has the right logic to dynamically identify the necessary block ranges based on the two root bundles.

## Summary
This PR removes a lot of code!

## Benefits for future improvement
The additional bonus with this PR is that it will make it easier to refactor how the SpokePoolClient loads data. Today it must load data a fromBlock to a toBlock, and on subsequent runs, must add only future blocks > toBlock to the end of the array.

A desirable feature would be able to be able to load earlier blocks thann the fromBlock to the beginning of the array. This would make it possible to extend the spoke pool client's lookback backwards dynamically without repeating queries.


src/dataworker/index.ts
```diff
@@ -72,108 +68,64 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
...
       );
 
-      // Construct spoke pool clients.
-      let spokePoolClients: SpokePoolClientsByChain;
       // Record the latest blocks for each spoke client that we can use to construct root bundles. This is dependent
       // on the fills for the client having matching deposit events on a different client. If a fill cannot be matched
       // with a deposit then we can't construct a root bundle for it.
       let latestInvalidBundleStartBlocks: { [chainId: number]: number } = {};
 
-      if (config.dataworkerFastLookbackCount > 0) {
```
nicholaspai, line 82: This large diff is a removal of the fallback case where `dataworkerFastLookbackCount=0`, in which case we try to load events from ALL time. This is impractically slow now and is not an option I envision ever being used in by any bot runner, so I think we can remove this code. 

Additionally, we now need to enforce that `dataworkerFastLookbackCount > 0`

mrice32 (approved):
LGTM! This reduces complexity a lot. +1

PR #376, opened 2022-11-25
improve: Proposer should wait after roots are relayed to propose

nicholaspai:
Currently the proposer waits `bufferToPropose` blocks after the first Ethereum RelayerRefundLeaf is executed to propose a bundle. This rule was originally implemented to prevent self disputes occurring where a proposal happened right after the previous root bundle's PoolRebalanceLeaves were executed. For some unknown reason, the proposer was producing an invalid bundle in such a situation, but did not do so after waiting some time to propose.

The "first executed RelayerRefundLeaf" rule was an approximate way to measure an amount of time before proposing the bundle.

I want to try using the rule because I think its (1) more correct and (2) easier to test hypotheses about the self-dispute bug with. The time it takes for all root bundles to propogate to the L2s will always be longer than the time to execute the first EthereumRelayerRefundLeaf, because of the latency to send messages from L1 to L2. In practice, the Eth --> Polygon route has always been the bottleneck, taking up to 30 mins sometimes. Additionally, I want to try setting the `bufferToPropose` to 0 after this PR and see if the roots being relayed, rather than the leaves being executed, is the cause of the self-dispute bug.

This PR does NOT solve the underlying self-dispute issue


src/dataworker/Dataworker.ts
```diff
@@ -268,9 +267,9 @@ export class Dataworker {
         });
     }
 
-    // This is a temporary fix: wait some time until we propose another root bundle. Currently, there appears to be a bug where proposing a
-    // root bundle immediately after executing all PoolRebalanceLeaves proposes an invalid bundle, but waiting
-    // a bit, possibly for cache to populate, works.
+    // This is a temporary fix: Currently, there appears to be a bug where proposing a root bundle before any
```
nicholaspai, line 270: This comment is key to understanding the motivation for this PR, curious if you agree/disagree with it

nicholaspai:
Hm i'm rethinking this PR, because ther eis benefit to waiting a fixed amount of time even after the first relayer refund leaf is executed to reduce slow fill count as the refunds process

src/dataworker/Dataworker.ts
```diff
@@ -150,6 +156,60 @@ export class Dataworker {
...
+    spokePoolClients: { [chainId: number]: SpokePoolClient },
+    bufferToPropose: number = this.bufferToPropose
+  ) {
+    const mostRecentValidatedBundle = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(
+      await this.clients.hubPoolClient.hubPool.provider.getBlockNumber()
+    );
+    // If there has never been a validated root bundle, then we can always propose a new one:
+    if (mostRecentValidatedBundle === undefined)
+      return {
+        value: false,
```
kevinuma, line 173: nit: How about naming this shouldWait to be clearer instead of value?

src/dataworker/Dataworker.ts
```diff
@@ -150,6 +156,60 @@ export class Dataworker {
...
+    // We should wait to propose until all root bundles have been relayed to spoke pools.
+    if (Object.entries(relayedRootBundles).length !== expectedRootBundles) {
+      return {
+        value: true,
+        mostRecentValidatedBundle: mostRecentValidatedBundle?.blockNumber,
+        expectedRootBundles,
+        relayedRootBundles,
+      };
+    } else {
+      const latestRelayedRootBundle = sortEventsAscending(Object.values(relayedRootBundles))[0];
```
kevinuma, line 200: Unless I misunderstood something, shouldn't this be sortEventsDescending? Sorting events ascending means first event would be the smallest or earliest while we want the latest.

nicholaspai: ah shoot this is a bad variable name it should be `earliestRelayedRootBundle`

kevinuma: Yep that sounds correct.

kevinuma (approved):

src/dataworker/Dataworker.ts
nicholaspai: good point

src/utils/FillUtils.ts
```diff
@@ -94,7 +94,7 @@ export function getLastMatchingFillBeforeBlock(
   lastBlock: number
 ): FillWithBlock {
   return sortEventsDescending(allFills).find(
-    (fill: FillWithBlock) => !fill.isSlowRelay && filledSameDeposit(fillToMatch, fill) && lastBlock >= fill.blockNumber
```
nicholaspai, line 97: We shouldn't check `!fill.isSlowRelay` since it doesn't make this function as generalized as its name implies. It should just return the latest matching fill before the block supplied.

PR #377, opened 2022-11-27
feat: Relayer should query /limits endpoint to not overfill deposits 

nicholaspai:
Protects relayer against delayed refunds

This PR will slow down relayer a bit since it has to make additional API requests.

I considered adding this functionality to a client but chose not to for simplicity sake.


src/relayer/Relayer.ts
```diff
@@ -148,6 +160,28 @@ export class Relayer {
         continue;
       }
 
+      // Cache limits to speed up run for next deposit for same path.
+      if (!limits[deposit.originToken]) limits[deposit.originToken] = {};
+      if (!limits[deposit.originToken][deposit.destinationChainId]) {
+        const _limits = await this.callLimits(deposit.originToken, deposit.destinationChainId);
+        limits[deposit.originToken][deposit.destinationChainId] = _limits;
```
mrice32, line 167: nit: why not just set this in one line?

src/relayer/Relayer.ts
```diff
@@ -148,6 +160,28 @@ export class Relayer {
...
+      }
+
+      if (
+        limits[deposit.originToken][deposit.destinationChainId] &&
```
mrice32, line 171: Is this ever undefined in this case? Shouldn't it always be defined by this point?

mrice32 (approved):
Looks great!

src/relayer/Relayer.ts
nicholaspai: true

src/relayer/Relayer.ts
nicholaspai: ah yes will remove, leftover code

src/relayer/Relayer.ts
```diff
@@ -148,6 +161,22 @@ export class Relayer {
         continue;
       }
 
+      // Cache limits to speed up run for next deposit for same path.
+      // Enforcing a timeout of 1s means that this external call will only add up to 1s per l1 token in this batch
+      // which doesn't add significant latency to the relayer run compared to other bottlenecks.
+      if (!limits[l1Token.address]) limits[l1Token.address] = await this.callLimits(l1Token.address, 1000);
```
kevinuma, line 167: I wonder if this can be moved into a separate client so that we can fetch the limit at the same time as event queries, which would help reduce any extra latency.

src/relayer/Relayer.ts
```diff
@@ -148,6 +161,22 @@ export class Relayer {
...
+
+      if (unfilledAmount.gt(limits[l1Token.address].maxDeposit)) {
+        this.logger.warn({
```
kevinuma, line 170: This will keep emitting the warning message for the entire lookback of the relayer and be spammy. Should we subject it to a shorter alert period (e.g. 1 hour)?

src/relayer/Relayer.ts
```diff
@@ -353,4 +382,43 @@ export class Relayer {
...
   }
+
+  private async callLimits(l1Token: string, timeout = 1000): Promise<DepositLimits> {
+    const url = "https://across.to/api/limits";
+    // Need to look up a valid dest chain for an l1 token since not all dest chains are supported, for example
+    // BOBA can only bridge to 288
+    const validDestinationChainForL1Token = l2TokensToL1TokenValidation[l1Token]
+      ? Object.keys(l2TokensToL1TokenValidation[l1Token])[0]
+      : 10;
+    const params = { token: l1Token, destinationChainId: validDestinationChainForL1Token };
```
kevinuma, line 393: Can we pass originChainId as well? https://github.com/across-protocol/frontend-v2/blob/master/api/limits.ts#L25

API endpoints will derive originChainId based on the L2 token address if not passed, and this currently is broken for DAI since it has the same address on Optimism and Arbitrum

src/relayer/Relayer.ts
```diff
@@ -353,4 +382,43 @@ export class Relayer {
...
+        url,
+        params,
+        data,
+      });
+      return {
+        maxDeposit: BigNumber.from(data.maxDeposit),
+      };
+    } catch (err) {
+      const msg = get(err, "response.data", get(err, "response.statusText", (err as AxiosError).message));
+      this.logger.debug({
```
kevinuma, line 412: Should this be warn?

src/relayer/Relayer.ts
nicholaspai: Yeah that's a good call, i was a bit lazy to do this but i think the parallel calls to the external API is worth including

src/relayer/Relayer.ts
nicholaspai: but what if the maxDeposit amount changes during the lookback? i.e. funds get added to hub pool

kevinuma: That's true. We'll play it by ear then

src/relayer/Relayer.ts
nicholaspai: done

src/relayer/Relayer.ts
nicholaspai: hmm good point

kevinuma: The error only manifested in one direction. Did you try having the destinationChainId set to 10 and 42161?

src/relayer/Relayer.ts
nicholaspai: yeah i guess so

src/relayer/Relayer.ts
nicholaspai: yeah but i use the `l1Token` address so the origin chain should always be 1 right? I'll hardcode it to be safe

kevinuma: Yes if it's always l1 token address. It should work as all token addresses are distinct on L1.

kevinuma (approved):

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,80 @@
...
+export class AcrossApiClient {
+  private endpoint = "https://across.to/api/";
+
+  private limits: { [token: string]: BigNumber } = {};
+
+  constructor(readonly logger: winston.Logger, readonly tokensQuery: string[] = [], readonly timeout: number = 5000) {
+    if (Object.keys(tokensQuery).length === 0) this.tokensQuery = Object.keys(l2TokensToL1TokenValidation);
+  }
+
+  async update(): Promise<void> {
```
pxrl, line 19: update() => updateLimits() ?

Just thinking ahead in case we want to bundle in support for querying historical suggested fees, which has been discussed. In that case the update() method would probably be a bit misleading.

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,80 @@
...
+      this.tokensQuery.map((l1Token) => {
+        const validDestinationChainForL1Token = l2TokensToL1TokenValidation[l1Token]
+          ? Number(Object.keys(l2TokensToL1TokenValidation[l1Token])[0])
+          : 10;
+        return this.callLimits(l1Token, validDestinationChainForL1Token);
+      })
+    );
+    for (let i = 0; i < this.tokensQuery.length; i++) {
+      const l1Token = this.tokensQuery[i];
+      this.limits[l1Token] = data[i]?.maxDeposit ? data[i].maxDeposit : BigNumber.from(MAX_UINT_VAL);
```
pxrl, line 42: Curious about two points here:
 - Do we want to fail open or fail closed on a failed lookup?
 - If we want to fail open, MAX_UINT_VAL risks being greater than LP deposits.

On this basis, I'd be inclined to fail closed and default to 0. This means we'd be unable to make fills in the event of an API outage, but protects us against accepting deposits that are too much for us to really handle.

If we fail open, then we might instead want to default to a configurable value from the environment (...although we should be wary of adding even more configuration items, because they become hard to manage...).

wdyt?

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    this.logger.debug({
+      at: "AcrossAPIClient",
+      message: "🏁 Fetched max deposit limits",
+      limits: this.limits,
+    });
+
+    this.logger.debug({ at: "AcrossAPIClient", message: "AcrossAPI client updated!" });
```
pxrl, line 50: Could these two be merged?

src/clients/AcrossAPIClient.ts
kevinuma: Failing close is risky though. For example yesterday the API was hammered hard and was overloaded. This means the reliability of the bots would be affected by the performance of the API, which increases risks of things going down together like dominoes. As long as we log the errors as at least warnings, we can be alerted and react to this situation.

src/clients/AcrossAPIClient.ts
nicholaspai: I was thinking we'd add in the suggested fees query into the `update()` to be consistent with other bots?

src/clients/AcrossAPIClient.ts
nicholaspai: hmmm this is a great point @pxrl. I'm inclined also to fail close and default to 0, thus blocking fills. We should `warn` when the API fails, as @kevinuma suggests, and then what if we add an easy flag like `useLimits` that we can use to easily bypass the API incase its down for a known reason

src/clients/AcrossAPIClient.ts
nicholaspai: Ill delete the second one its not useful

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  getLimit(l1Token: string): BigNumber {
+    return this.limits[l1Token] ? this.limits[l1Token] : BigNumber.from(MAX_UINT_VAL);
```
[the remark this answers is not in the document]
nicholaspai: your suggestion seems less readable?

src/clients/AcrossAPIClient.ts
pxrl: Suggested fees depends on deposit-specific things though, so I'm not sure it makes sense to batch it all together.

src/clients/AcrossAPIClient.ts
pxrl: I've thought about it some more, and I think it's preferable to fail closed, rather than open.

In the scenario where the API is down and we can't query limits, the FE and aggregators are going to be down too, because they probably can't query suggested-fees or limits anyway. This could even be dangerous if they mask the errors and proceed with default values that we might never have seen in actual deposits.

At the same time, if API endpoints are down then we should expect that the vast majority of new deposits reduces dramatically or entirely. Therefore, I think taking marginal risks to ensure the bots keep operating seems to have a marginal benefit at best. In the worst case, we would be able to make local changes to the bot to flush and outstanding deposits.

A possible compromise would be to disable slow fills in the event that we can't query /limits. This puts a pretty low ceiling on the amount of possible damage that could be done.

src/clients/AcrossAPIClient.ts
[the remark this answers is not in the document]
pxrl: Just to make sure I have described it properly, I propose to change the line to 

`return this.limits[l1Token] ?? BigNumber.from(XXX)`

It is arguably more or less readable...it relies on a feature of js that's not in any other language I've ever used (afaik), but one advantage I do like is that it makes future edits less risky because there's no duplication of `this.limits[l1Token]`.

Anyway, just a suggestion :)

mrice32 (approved):

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,85 @@
...
+    return this.limits[l1Token];
+  }
+
+  private async callLimits(
+    l1Token: string,
+    destinationChainId: number,
+    timeout = this.timeout
+  ): Promise<DepositLimits> {
+    const path = "limits";
+    const url = `${this.endpoint}${path}`;
```
pxrl, line 67: Very minor nit: I'd pull the '/' out of this.endpoint and put it here instead, so it's clear that there is a '/' delimiter between the two variables. Otherwise it's not super clear how the two variables relate to each other.

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,85 @@
...
+  private limits: { [token: string]: BigNumber } = {};
+
+  public updatedLimits = false;
+
+  constructor(readonly logger: winston.Logger, readonly tokensQuery: string[] = [], readonly timeout: number = 5000) {
+    if (Object.keys(tokensQuery).length === 0) this.tokensQuery = Object.keys(l2TokensToL1TokenValidation);
+  }
+
+  async update(ignoreLimits: boolean): Promise<void> {
+    if (!ignoreLimits) {
```
pxrl, line 22: Maybe just reverse this, so we log and return immediately if ignoreLimits is true? It would remove an entire level of indentation for all the other code in this method.

```
  if (ignoreLimits) {
    this.logger.debug({ ..., message: "Skipping /limits query", ... });
    return;
  }

 / ... proceed with update.
```

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      // /limits
+      // - Store the max deposit limit for each L1 token. DestinationChainId doesn't matter since HubPool
+      // liquidity is shared for all tokens and affects maxDeposit. We don't care about maxDepositInstant
+      // when deciding whether a relay will be refunded.
+      const data = await Promise.all(
+        this.tokensQuery.map((l1Token) => {
+          const validDestinationChainForL1Token = l2TokensToL1TokenValidation[l1Token]
+            ? Number(Object.keys(l2TokensToL1TokenValidation[l1Token])[0])
+            : 10;
```
pxrl, line 39: Nice; this is an elegant way of handling this.

src/clients/AcrossAPIClient.ts
nicholaspai: I agree with failing closed and I've implemented that

src/clients/AcrossAPIClient.ts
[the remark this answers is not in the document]
nicholaspai: oh ok that makes sense yes i was confused by your "=>" which suggested some strange arrow function

src/clients/AcrossAPIClient.ts
```diff
@@ -0,0 +1,85 @@
...
+  async update(ignoreLimits: boolean): Promise<void> {
+    if (!ignoreLimits) {
+      this.logger.debug({
+        at: "AcrossAPIClient",
+        message: "Querying /limits",
+        timeout: this.timeout,
+        tokensQuery: this.tokensQuery,
+        endpoint: this.endpoint,
+      });
+
```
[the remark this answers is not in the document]
nicholaspai: It defaults to false so you think every time we enter `update` it should also set `updatedLimits = false`?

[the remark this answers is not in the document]
pxrl: My expectation is that we want to ensure that we update our limits on each loop, right? Otherwise, a bot might be running in looping mode for several days, and the API could have been down for half that time. As long as it was updated once during that time then it would end up using stale pool data.

PR #378, opened 2022-11-30
feat(finalizer): Add support for lookback in seconds (instead of blocks)

kevinuma:
Context:
Currently the finalizer sets the lookback for fetching transfer/finalization events using number of blocks. However, as the block production rate varies, the finalizer can sometimes fall short (lookback is not enough to cover all deposits). It's also easier for operators to set the config if it's duration instead of number of blocks.


src/finalizer/index.ts
```diff
@@ -90,13 +92,30 @@ export async function finalize(
...
 
-  const spokePoolClients = await constructSpokePoolClientsWithLookback(
-    logger,
-    commonClients.configStoreClient,
-    config,
-    baseSigner,
-    config.maxFinalizerLookback
-  );
+  let spokePoolClients;
+  if (config.maxFinalizerLookbackSecs > 0) {
```
kevinuma, line 96: I'm using this as a feature flag so we can enable/disable this new timestamp-based lookup easily

src/common/ClientHelper.ts
```diff
@@ -48,6 +49,30 @@ export async function constructSpokePoolClientsWithLookback(
...
+  logger: winston.Logger,
+  configStoreClient: AcrossConfigStoreClient,
+  config: CommonConfig,
+  baseSigner: Wallet,
+  lookbackSecs: { [chainId: number]: number } = {}
+): Promise<SpokePoolClientsByChain> {
+  // Calculate fromBlocks based on the desired lookback duration (in secs).
+  const fromBlocks: { [chainId: number]: number } = {};
+  const blockFutures: Promise<number>[] = [];
+  for (const strChainId of Object.keys(lookbackSecs)) {
```
mrice32, line 63: nit: prefer .map() when you're just trying to return an array of the same length as the thing you're looping over.

nit: prefer Object.entries() when using both the key and value, since it returns a list of pairs of keys and values.

src/common/ClientHelper.ts
```diff
@@ -48,6 +49,30 @@ export async function constructSpokePoolClientsWithLookback(
...
+    const chainId = Number(strChainId);
+    blockFutures.push(utils.findBlockAtOrOlder(getProvider(chainId), lookbackSecs[chainId]));
+  }
+  const fromBlocksList = await Promise.all(blockFutures);
+  for (const strChainId of Object.keys(lookbackSecs)) {
+    const chainId = Number(strChainId);
+    fromBlocks[chainId] = fromBlocksList.shift();
+  }
```
mrice32, line 71: nit: prefer Object.fromEntries when creating an object from an array where each element corresponds to one key/value pair.

mrice32 (approved):
Looks great! Only minor nits.

src/common/ClientHelper.ts
```diff
@@ -48,6 +49,30 @@ export async function constructSpokePoolClientsWithLookback(
   return getSpokePoolClientsForContract(logger, configStoreClient, config, spokePools, fromBlocks);
 }
 
+// Construct SpokePoolClients for all chains based on durations of time to fetch events for.
+export async function constructSpokePoolClientsWithLookbackSecs(
+  logger: winston.Logger,
+  configStoreClient: AcrossConfigStoreClient,
+  config: CommonConfig,
+  baseSigner: Wallet,
+  lookbackSecs: { [chainId: number]: number } = {}
```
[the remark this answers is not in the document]
nicholaspai: This is a good point I think we want the same lookback for all chains actually, that's the main benefit of this PR: condense a dictionary of lookback blocks into a single lookback time

src/common/Constants.ts
```diff
@@ -32,6 +32,10 @@ export const FINALIZER_TOKENBRIDGE_LOOKBACK: { [chainId: number]: number } = {
   42161: 4838400,
 };
```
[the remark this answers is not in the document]
pxrl: I see - will be removed later.

nicholaspai (changes requested):
requesting changes to matt's review

nicholaspai:
Replacing with #395

PR #379, opened 2022-11-30
improve: Refine logic in dynamically extending dataworker lookback

nicholaspai:
I forgot to commit some code that I tested locally as part of PR #371


src/dataworker/index.ts
```diff
@@ -53,17 +53,15 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
...
-      const latestFullyExecutedBundleEndBlocks = clients.hubPoolClient.getLatestFullyExecutedRootBundle(
-        clients.hubPoolClient.latestBlockNumber
-      ).bundleEvaluationBlockNumbers;
-      const bundleEndBlockMapping = Object.fromEntries(
+      // Get end blocks from second to latest fully executed root bundle. We'll use this end block to determine
+      // if we need to load more events (by expanding the dataworker's lookback) in order to evaluate the pending
+      // bundle. In order to do so, we need to load events as oldas the pending bundle's fromBlock, which is the
+      // the second to latest fully executed bundle's end blocks + 1
+      const nthFullyExecutedBundleEndBlocks =
+        clients.hubPoolClient.getNthFullyExecutedRootBundle(-2).bundleEvaluationBlockNumbers;
```
nicholaspai, line 61: This is critical to add as it reduces chance of dynamically extending the lookback by using the latest executed bundle's FROM BLOCKS rather than its TO BLOCKS when determining whether to extend a lookback. To evaluate a bundle, you need to look up events older than its fromBlock so this is an oversight

mrice32 (approved):

src/dataworker/index.ts
```diff
@@ -53,17 +53,15 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
...
-      // to the bundle end blocks for the chains. This will let the dataworker's proposal and validation logic
-      // to continue to fetch fresh events while allowing the leaf execution methods to use cached logic and overall
-      // reduce the number of web3 requests sent.
-      const latestFullyExecutedBundleEndBlocks = clients.hubPoolClient.getLatestFullyExecutedRootBundle(
-        clients.hubPoolClient.latestBlockNumber
-      ).bundleEvaluationBlockNumbers;
-      const bundleEndBlockMapping = Object.fromEntries(
+      // Get end blocks from second to latest fully executed root bundle. We'll use this end block to determine
+      // if we need to load more events (by expanding the dataworker's lookback) in order to evaluate the pending
+      // bundle. In order to do so, we need to load events as oldas the pending bundle's fromBlock, which is the
```
kevinuma, line 58: nit: oldas => old as

src/dataworker/index.ts
```diff
@@ -53,17 +53,15 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
...
+      // the second to latest fully executed bundle's end blocks + 1
+      const nthFullyExecutedBundleEndBlocks =
```
kevinuma, line 60: nit: Why not name this secondMostRecentBundleEndBlocks?

src/dataworker/index.ts
kevinuma: Makes sense

kevinuma (approved):

PR #381, opened 2022-12-01
feat: Use Multicall to batch Finalizer transactions

nicholaspai:
Currently only batches Polygon and Optimism finalizations because Arbitrum's calldata is a bit harder to extract. I will work on this as a follow up to this PR once we are happy with the structure.

Example txn I sent with this:
- https://etherscan.io/tx/0x26ad50ff8bd6a81e36251838c222067da39bb0f0156deb152d31de0c2f8cf29d (this one I had a bug in the code that didn't correctly include the `PolygonTokenBridger.retrieve()` txns but the second txn includes this bug fix and correctly calls `retrieve()`)
- https://etherscan.io/tx/0xc852ab8dedcdc8909e84377e1266cdf9c9f413199daa97330a8e9e74d746b425

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -64,16 +64,16 @@ export async function finalize(
     Multicall2Ethers__factory.abi,
     hubPoolClient.hubPool.signer
   );
-
+  const finalizationsToBatch: { callData: Multicall2Call[]; withdrawals: Withdrawal[] } = {
+    callData: [],
+    withdrawals: [],
+  };
   // For each chain, look up any TokensBridged events emitted by SpokePool client that we'll attempt to finalize
   // on L1.
   for (const chainId of configuredChainIds) {
     const client = spokePoolClients[chainId];
     const tokensBridged = client.getTokensBridged();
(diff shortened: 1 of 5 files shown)


src/finalizer/index.ts
```diff
@@ -62,27 +90,53 @@ export async function finalize(
...
+            withdrawal.l1TokenSymbol
+          } 🪃`,
+          transactionHash: etherscanLink(txn.transactionHash, 1),
+        });
+      });
+    } catch (error) {
+      logger.warn({
+        at: "Finalizer",
+        message: "Error creating aggregateTx",
+        error,
```
pxrl, line 137: Were you able to provoke this? I recall that it doesn't work as expected to drop an exception object directly as a key in the logger. My experience was that the text would fall completely outside of the log message.

pxrl (approved):
Nice improvement!

src/finalizer/index.ts
nicholaspai: Yeah same experience I'll log out the Error object and then try to read the reason

PR #382, opened 2022-12-01
improve: Add method to look up historical deposit for fill older than SpokePoolClient's starting lookback

nicholaspai:
This PR will allow us to shorten the dataworker default search range because it no longer needs to account for older deposits that get filled very late.

Blocked by https://github.com/across-protocol/relayer-v2/pull/436

What this PR does:
- See description in https://github.com/across-protocol/relayer-v2/pull/436

What this PR does not do:
- Allows us to dynamically expand the full SpokePoolClient lookback for each chain from `[x, fromBlock]`. This doesn't seem particularly useful provided we set `DATAWORKER_FAST_BUNDLE_LOOKBACK_COUNT` reasonably (~ 2 days of blocks) and bundle cadence doesn't get a lot slower. Even in that case we can increase `DATAWORKER_FAST_BUNDLE_LOOKBACK_COUNT`.
- Does not help when a slow fill completes an old deposit and we need to compute the excess. This case is very rare but should be addressed in a [follow up PR](https://github.com/across-protocol/relayer-v2/tree/npai/query-range) since this is a large one

Resolves ACX-529


nicholaspai:
Resolves ACX-529

src/clients/BundleDataClient.ts
```diff
@@ -193,6 +200,46 @@ export class BundleDataClient {
     const allValidFills: FillWithBlock[] = [];
     const allInvalidFills: FillWithBlock[] = [];
 
+    const addRefundForValidFill = (
```
nicholaspai, line 203: This is a pure refactor of logic already existing in `loadData` since we now have multiple places where we want to record validated fills we need to refund

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +279,63 @@ export class BundleDataClient {
...
+            // Note: If the fill's origin chain is set incorrectly (e.g. equal to the destination chain, or
+            // set to some unexpected chain), then we won't enter this branch because we wouldn't
+            // have been able to grab it from the destinationClient.getFillsForOriginChain call.
+
+            // Matched deposit for fill was not found in spoke client. This situation should be rare so let's
+            // send some extra RPC requests to find it if it exists.
+            if (this.historicalDepositCache[originChainId] === undefined)
+              this.historicalDepositCache[originChainId] = {};
+            if (this.historicalDepositCache[originChainId][fillWithBlock.depositId] === undefined) {
+              return originClient.queryHistoricalDepositForFill(fillWithBlock);
```
nicholaspai, line 300: This part is new. Whenever there is a fill we haven't matched with a deposit, before immediately slotting it into `invalidFills`, we try to lookup its deposit via `queryHistoricalDepositForFill`

src/clients/SpokePoolClient.ts
```diff
@@ -60,6 +63,11 @@ export class SpokePoolClient {
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
     readonly spokePoolDeploymentBlock: number = 0
   ) {
+    try {
+      this.spokePoolDeploymentBlock = getDeploymentBlockNumber("SpokePool", chainId);
```
nicholaspai, line 67: Unrelated fix: We should be loading spoke pool deployment from the `deployments.json` file whenever we can

src/clients/SpokePoolClient.ts
```diff
@@ -341,7 +425,8 @@ export class SpokePoolClient {
         });
       for (const [index, event] of depositEvents.entries()) {
         // Append the realizedLpFeePct.
-        const processedEvent: Omit<Deposit, "destinationToken" | "realizedLpFeePct"> = spreadEvent(event);
+        const processedEvent: Omit<DepositWithBlock, "destinationToken" | "realizedLpFeePct"> =
```
nicholaspai, line 428: Unrelated fix: We store `this.deposits` as type `DepositWithBlock` so we should be casting the event to that type via `spreadEventWithBlockNumber` instead of `spreadEvent`. We had set up tests with false positives that ignored the extra `WithBlock` fields so I fixed those too

src/dataworker/index.ts
```diff
@@ -78,129 +61,43 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
...
+        config,
+        baseSigner,
+        fromBlocks,
+        toBlocks
       );
 
-      // Each loop iteration constructs spoke clients using a lookback and sees if we can query old enough events
-      // to cover the target bundle range determined above (i.e. the bundle after the nthFullyExecutedBundle).
-      const customConfig = { ...config };
-      for (let i = 0; i <= config.dataworkerFastLookbackRetryCount; i++) {
```
nicholaspai, line 106: Removing this for loop is nice because we can now validate any filll even if its deposits came before the spoke pool client's first block searched. However, this PR does not currently query fills before that first block searched.

src/utils/FillUtils.ts
```diff
@@ -96,27 +96,42 @@ export function filledSameDeposit(fillA: Fill, fillB: Fill): boolean {
...
+  let firstFillForSameDeposit = allMatchingFills.find((_fill) => isFirstFillForDeposit(_fill));
+
+  // If `allValidFills` doesn't contain the first fill for this deposit then we have to perform a historical query to
+  // find it. This is inefficient, but should be rare.
+  if (!firstFillForSameDeposit) {
+    // Note: allMatchingFills might have duplicate fills if there are multiple fills in the
+    // `firstFillForSameDeposit.blockNumber` but we don't care since we only need to find the first fill, of which
+    // there should be exactly 1, and the last fill in the same bundle.
+    allMatchingFills.push(
+      ...(await spokePoolClientsByChain[fill.originChainId].queryHistoricalMatchingFills(
```
nicholaspai, line 128: This change means that we no longer need to worry if we execute a slow fill for a deposit that happened a LONG time ago

test/Dataworker.loadData.ts
```diff
@@ -656,14 +657,70 @@ describe("Dataworker: Load data used in all functions", async function () {
...
+      .excludingEvery(["logIndex", "transactionHash", "transactionIndex"])
+      .to.deep.equal([{ ...deposit1, blockNumber: realizedLpFeePctData.quoteBlock, originBlockNumber: originBlock }]);
 
     // If block range does not cover deposits, then they are not included
     expect(
-      dataworkerInstance.clients.bundleDataClient.loadData(IMPOSSIBLE_BLOCK_RANGE, spokePoolClients).deposits
+      (await dataworkerInstance.clients.bundleDataClient.loadData(IMPOSSIBLE_BLOCK_RANGE, spokePoolClients)).deposits
     ).to.deep.equal([]);
   });
+  it("Can fetch historical deposits older than SpokePoolClient's search range", async function () {
```
nicholaspai, line 670: This is the new unit tests to review carefully

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +281,71 @@ export class BundleDataClient {
...
-              updateUnfilledDepositsWithMatchedDeposit(fill, matchedDeposit, unfilledDepositsForOriginChain);
-
-              // Update total refund counter for convenience when constructing relayer refund leaves
-              updateTotalRefundAmount(fillsToRefund, fill, chainToSendRefundTo, repaymentToken);
-            } else {
-              // Note: If the fill's origin chain is set incorrectly (e.g. equal to the destination chain, or
-              // set to some unexpected chain), then it won't be added to `allInvalidFills` because we wouldn't
-              // have been able to grab it from the destinationClient.getFillsForOriginChain call.
+          .filter((fillWithBlock) => fillWithBlock.blockNumber <= blockRangeForChain[1]);
+        const historicalDepositPromises = fillsForOriginChain.map((fillWithBlock) => {
```
mrice32, line 287: nit: if you make this function async and wrap it in a Promise.all, you can avoid the second loop below. Without this, the return type is a bit convoluted (with some being promises and some not). Something like this I think would be a bit simpler:

```ts
const historicalDeposits = Promise.all(fillsForOriginChain.map(async (fillWithBlock) => {
  ...
}))
```

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +281,71 @@ export class BundleDataClient {
...
+        matchedHistoricalDeposits.forEach(([deposit, fillWithBlock]) => {
+          this.historicalDepositCache[this.getUniqueDepositKey(fillWithBlock)] = deposit;
+          addRefundForValidFill(fillWithBlock, deposit, blockRangeForChain);
+        });
+
+        // Mark fills without matching deposits, even after additional queries for deposit, as invalid.
+        const unmatchedHistoricalDeposits = historicalDepositQueries.filter((x) => x[0] === undefined);
+        unmatchedHistoricalDeposits.forEach(([, fill]) => {
+          allInvalidFills.push(fill);
+        });
```
mrice32, line 341: This logic feels similar, in concept, to the other matching/validation logic inside the map above. Thoughts on having an async function called findDepositForFill that does all of this logic:
- Tries to find in the origin client.
- If not found, fall back to checking the historical deposit cache.
- If not found, make RPC queries to find and store in cache.
- Validates the fill against the deposit (if necessary). If invalid, return undefined.

Then your logic could essentially be:
```ts
const depositFillPairs = await Promise.all(fillsForOriginChain.map(async (fill) => {
  const validatedDeposit = await findDepositForFill(fill, .....);
  return [validatedDeposit, fill];
}));

const invalidFills = depositFillPairs.filter(([deposit]) => deposit === undefined).map(([, fill]) => fill);

const validFills = depositFillPairs.filter(([deposit]) => deposit !== undefined).map(([, fill]) => fill);
```

The only reason I suggest this structure is that it would help break this method up and hopefully make the high-level logic a bit clearer, while leaving all the query/caching details to the inner findDepositForFill method. Thoughts?

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +245,63 @@ export class SpokePoolClient {
...
+  // where `numberOfDeposits == targetDepositId`.
+  async binarySearchForBlockContainingDepositId(targetDepositId: number): Promise<number> {
+    let high = this.firstBlockToSearch;
+    let low = this.spokePoolDeploymentBlock;
+    do {
+      const mid = Math.floor((high + low) / 2);
+      const searchedDepositId = await this.spokePool.numberOfDeposits({ blockTag: mid });
+      if (targetDepositId > searchedDepositId) low = mid + 1;
+      else if (targetDepositId < searchedDepositId) high = mid - 1;
+      else return mid;
```
mrice32, line 260: Won't this just find _any_ block where the numberOfDeposits equals the target? Couldn't that be many many blocks (not just the block where it incremented)?

I think the right search criteria is to find the block where numberOfDeposits equals the target _and_ the block before it equals the target - 1, which means, by definition, this block is the one with the event, no?

Edit: I see, you binary search for any block with the deposit below and above. That makes sense! Might want to do some edge case analysis around what happens if the deposit we're searching for is in the future, 0, or what would happen if the spoke pool starts at a nonzero id.

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +245,63 @@ export class SpokePoolClient {
...
+    } while (low <= high);
```
mrice32, line 261: Should we error if we get past the while? Theoretically, it shouldn't happen, but it seems like a failure mode/bug we'd want to catch.

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +245,63 @@ export class SpokePoolClient {
...
+    ]);
+
+    const query = await paginatedEventQuery(this.spokePool, this._queryableEventNames().FundsDeposited, {
+      fromBlock: blockBeforeDeposit,
+      toBlock: blockAfterDeposit,
+      maxBlockLookBack: 0,
+    });
+    const event: FundsDepositedEvent = (query as FundsDepositedEvent[]).find(
+      (deposit) => deposit.args.depositId === fill.depositId
+    );
```
mrice32, line 285: If you use a topic filter, the node should do the filtering for us.

```suggestion
    const [event] = await paginatedEventQuery(this.spokePool, this.spokePool.filters.FundsDeposited(null, null, null, null, fill.depositId), {
      fromBlock: blockBeforeDeposit,
      toBlock: blockAfterDeposit,
      maxBlockLookBack: 0,
    });
```

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +281,71 @@ export class BundleDataClient {
...
+              if (depositPromise === undefined) return;
+              return [await depositPromise, fillsForOriginChain[i]];
+            })
+          )
+        ).filter((x) => x !== undefined) as [DepositWithBlock | undefined, FillWithBlock][];
+
+        // Mark fills with matching deposits as valid and add them to the list of fills to refund.
+        const matchedHistoricalDeposits = historicalDepositQueries.filter((x) => x[0] !== undefined);
+        matchedHistoricalDeposits.forEach(([deposit, fillWithBlock]) => {
+          this.historicalDepositCache[this.getUniqueDepositKey(fillWithBlock)] = deposit;
```
mrice32, line 333: Is the point of this cache just in case we see a lot of the same/similar fills so we can avoid repeatedly looking it up?

mrice32 (commented):
Looks awesome! Left a few comments on structure, but overall looks good!

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +281,71 @@ export class BundleDataClient {
...
+          .filter((fillWithBlock) => fillWithBlock.blockNumber <= blockRangeForChain[1]);
+        const historicalDepositPromises = fillsForOriginChain.map((fillWithBlock) => {
+          // If fill matches with a deposit, then its a valid fill.
+          const matchedDeposit = originClient.getDepositForFill(fillWithBlock);
+          if (matchedDeposit) {
+            addRefundForValidFill(fillWithBlock, matchedDeposit, blockRangeForChain);
+          } else {
+            // Note: If the fill's origin chain is set incorrectly (e.g. equal to the destination chain, or
+            // set to some unexpected chain), then we won't enter this branch because we wouldn't
+            // have been able to grab it from the destinationClient.getFillsForOriginChain call.
```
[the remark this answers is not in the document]
nicholaspai: This is probably a legacy comment at this point, for us to hit this branch we need to have the origin chain correctly set in the fill

src/clients/BundleDataClient.ts
```diff
@@ -232,52 +281,71 @@ export class BundleDataClient {
...
+              return originClient.queryHistoricalDepositForFill(fillWithBlock);
+            }
+            // Older deposit for this deposit ID has already been matched with a fill so we know this fill must be
+            // valid if it can be validated against that deposit. We still need to validate the fill because we don't
+            // know if the other parts of the deposit hash match besides the depositID.
+            else if (
+              originClient.validateFillForDeposit(
+                fillWithBlock,
+                this.historicalDepositCache[this.getUniqueDepositKey(fillWithBlock)]
+              ) === undefined
```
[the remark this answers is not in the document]
nicholaspai: Ah good point

src/clients/BundleDataClient.ts
nicholaspai: yeah, so every time we `loadData` with the same `fromBlock` we'll repeat requests. This can happen in the dataworker which is always looking up to a fixed `fromBlock`, however since each {propose,dispute,execute} run goes separately, the cache in-memory isn't actually shared. So perhaps I can remove the caching logic in this PR to reduce code, add a comment about benefits of caching in redis, and adding in a follow up

nicholaspai: Actually, it does save a query when you try to execute slow and relayerRefund leaves in the same run, so I'm going to keep it in.

src/clients/SpokePoolClient.ts
nicholaspai: > Might want to do some edge case analysis around what happens if the deposit we're searching for is in the future, 0, or what would happen if the spoke pool starts at a nonzero id.

Great points. This [test](https://github.com/across-protocol/relayer-v2/pull/382/files#diff-e73c5f246f260ad0f49e8990b9602cfb17f18b3e5f3c1627ec20e48db048c284R670) should already handle the case deposit ID is 0 and we need to historically search for it. This [commit](https://github.com/across-protocol/relayer-v2/pull/382/commits/4ec382a07d2dd3f8ba22ac45ef20d0596cb859a2) adds test cases where deposit ID is greater than the max deposit ID queried and when deposit ID is less than the spoke pool's deposit ID:

src/clients/BundleDataClient.ts
nicholaspai: Great eye! i'll implement this

src/clients/SpokePoolClient.ts
nicholaspai: agreed, we'd still error later but it'd be less clear so explicitly erroring is better

src/clients/SpokePoolClient.ts
nicholaspai: Done

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +250,88 @@ export class SpokePoolClient {
...
+  // deposit ID. This can be used by the Dataworker to determine whether to give a relayer a refund for a fill
+  // of a deposit older than its fixed lookback.
+  async queryHistoricalDepositForFill(fill: Fill): Promise<DepositWithBlock | undefined> {
+    if (fill.originChainId !== this.chainId) throw new Error("fill.originChainId !== this.chainid");
+    if (fill.depositId < this.firstDepositIdForSpokePool) return undefined;
+    if (fill.depositId >= this.earliestDepositIdQueried) return this.getDepositForFill(fill);
+
+    let deposit: DepositWithBlock, cachedDeposit: Deposit | undefined;
+    if (this.configStoreClient.redisClient) {
+      cachedDeposit = await getDeposit(getRedisDepositKey(fill), this.configStoreClient.redisClient);
```
nicholaspai, line 282: @mrice32 i cache this event query. It might have to change based on how you implement general event caching following https://github.com/across-protocol/relayer-v2/pull/429

test/ConfigStoreClient.ts
```diff
@@ -279,33 +279,24 @@ describe("AcrossConfigStoreClient", async function () {
...
-      // Now this is true since latest version in contract is 0, same as default.
-      expect(configStoreClient.hasLatestConfigStoreVersion).to.be.true;
-
       // Can't set update to non-integer:
       await configStore.updateGlobalConfig(utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.VERSION), "1.6");
-      await updateAllClients();
-      expect(configStoreClient.cumulativeConfigStoreVersionUpdates.length).to.equal(0);
-
+      // Set config store version to 6, making client think it doesn't have latest version, which is 0 in SDK.
+      await configStore.updateGlobalConfig(utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.VERSION), "6");
```
nicholaspai, line 285: I'm not sure why but this test started failing because the `configStoreClient.update()` was happening too quickly and the `toBlock = provider.getBlockNumber()` kept fetching blocks < searchConfig.fromBlock, so I reduced the number of `updateAllClients` calls to 1.

This led me to realize that in other unit tests we may have false positives because `spokepool/hubpool/configstoreclient.update()` returns early because `fromBlock > toBlock`

src/clients/BundleDataClient.ts
nicholaspai: See a redis caching feature implemented here https://github.com/across-protocol/relayer-v2/pull/382#discussion_r1083542182

pxrl (commented):
Not sure how helpful this comment is, but I'm pretty late to this PR (my own fault) and the size of it is pretty intimidating. It's challenging to separate the changes that are just rearranging the furniture, from the changes that are actually implementing new functionality. Are there any isolated/low-risk parts that could be carved out early and merged now, so that the actual impactful changes are remaining and easier to focus on?

nicholaspai:
> Not sure how helpful this comment is, but I'm pretty late to this PR (my own fault) and the size of it is pretty intimidating. It's challenging to separate the changes that are just rearranging the furniture, from the changes that are actually implementing new functionality. Are there any isolated/low-risk parts that could be carved out early and merged now, so that the actual impactful changes are remaining and easier to focus on?

Great idea. I pulled out the SpokePoolClient's helper function into a [new PR](https://github.com/across-protocol/relayer-v2/pull/436) which should reduce the diff in this PR a lot.

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
```diff
@@ -339,11 +339,11 @@ export class SpokePoolClient {
       };
       this.logger.debug({
         at: "SpokePoolClient",
-        message: "Queried RPC for deposit older than SpokePoolClient's lookback",
+        message: "Queried RPC for deposit outside SpokePoolClient's search range",
         deposit,
       });
       if (this.configStoreClient.redisClient)
-        await setDeposit(deposit, this.configStoreClient.redisClient, 30 * 24 * 60 * 60);
+        await setDeposit(deposit, this.configStoreClient.redisClient, 24 * 60 * 60);
```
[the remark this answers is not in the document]
nicholaspai: So this is an interesting point. Every time the dataworker bundle advances, the historical deposit requests we'll send will be different because the spoke pool client lookback will change. I wanted to err on the side of preserving less data and 24 hours gives plenty of time for the bundle to advance at which point the dataworker would no longer try to load these cached deposits.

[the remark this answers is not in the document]
mrice32: Seems reasonable! How many bundles do we typically look back by default?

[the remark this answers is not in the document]
mrice32: Agreed. This makes sense! 1 day seems totally fine!

mrice32 (approved):
LGTM!

PR #387, opened 2022-12-13
improve(multicaller): Env-configurable chunk sizes

pxrl:
Identified as a requirement when we detected that Optimism RPCs were
not accepting the default chunk size of 100 txns. At that point it would
have been really convenient to override via the environment rather than
forcing a new bot revision...but this wasn't supported by the bots.

Fixes ACX-437.

### DIFF
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -61,7 +61,7 @@ export class CommonConfig {
             ?? DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE[chainId]
             ?? DEFAULT_MULTICALL_CHUNK_SIZE
         );
-        assert(chunkSize > 0, `Chain ${chainId} multicall chunk size must be greater than 0.`);
+        assert(chunkSize > 0, `Chain ${chainId} multicall chunk size (${chunkSize}) must be greater than 0.`);
         return [chainId, chunkSize];
       })
     );
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -61,7 +61,7 @@ export class CommonConfig {
             ?? DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE[chainId]
             ?? DEFAULT_MULTICALL_CHUNK_SIZE
         );
-        assert(chunkSize > 0, `Chain ${chainId} multicall chunk size (${chunkSize}) must be greater than 0.`);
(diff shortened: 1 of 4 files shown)


src/common/Config.ts
```diff
@@ -62,5 +65,12 @@ export class CommonConfig {
     this.sendingTransactionsEnabled = SEND_TRANSACTIONS === "true";
     this.redisUrl = REDIS_URL;
     this.bundleRefundLookback = BUNDLE_REFUND_LOOKBACK ? Number(BUNDLE_REFUND_LOOKBACK) : 2;
+
+    this.multiCallChunkSize = CHAIN_MULTICALL_CHUNK_SIZE
+      ? JSON.parse(CHAIN_MULTICALL_CHUNK_SIZE)
+      : DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE;
```
mrice32, line 71: Note: this is a weakly held opinion, but I always flinch a little at JSON envs just because they are hard to encode for the user.

Have you considered something more like:
```sh
DEFAULT_MULTICALL_CHUNK_SIZE=123
MULTICALL_CHUNK_SIZE_[chainId]=456
```

Basically we create one env var that allows the user to override the default, and then users can provide individual envs for any network where they want to override that default on that network only.

Thoughts on that structure?

mrice32 (commented):
Looks great! A small comment on the structure of the envs.

src/common/Config.ts
nicholaspai: I'm guilty of JSON envs... +1 on this Matt I think it works well for NODE_URL_X for example

nicholaspai (approved):
LGTM after integrating matt's suggestion

src/common/Config.ts
mrice32: I think this is good in theory, but it's tricky in environments, like serverless ones, where you don't really have direct access to a disk.

I also prefer files in general, but I just don't see how to make a file-based approach work across many different execution environments with different levels of infrastructure abstraction.

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/387#discussion_r1050255069

pxrl:
Have updated this PR after the feedback; we now examine a single env var per chunk size variable that is overridden by the environment: https://github.com/across-protocol/relayer-v2/pull/387/commits/b048ac38bfe51c6cc133badfbfbd164552e0b8c3.

mrice32 (approved):
LGTM!

PR #388, opened 2022-12-14
improve(ProfitClient): Remove ignoreTokenPriceFailures option

pxrl:
This variable seems pretty redundant now. Price lookup failures are very
rare these days because we have fallback from the Across API to
CoinGecko Free/Pro, and DefiLlama exists as a third option if we choose
to use it. Additionally, the ignoreProfitability option is another good
proxy for whether we want to be strict about price lookup failures.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -72,8 +72,6 @@ export class ProfitClient {
     spokePoolClients: SpokePoolClientsByChain,
     readonly ignoreProfitability: boolean,
     readonly enabledChainIds: number[],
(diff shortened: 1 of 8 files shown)


pxrl:
nb. A side-benefit of this is that it reduces the complexity of using the ProfitClient. The argument list required to instantiate this this class is currently long, and it's easy to be confused by the arguments. Ideally I'd like to cut it down further over time.

nicholaspai (approved):
Nice cleanup! Love to see removing code thanks for doing this Paul

mrice32 (approved):
LGTM!

PR #389, opened 2022-12-14
improve(MultiCallerClient): Reduce try/catch scope

pxrl:
Very minor tweak to reduce the scope of the try/catch clause in the executeTransactionQueue() method. This has no functional impact and is done in isolation to avoid future a megadiff against this file.


nicholaspai (approved):
Great call thanks for this

PR #390, opened 2022-12-14
improve(MultiCallerClient): Factor out txn simulation

pxrl:
Reduce the size/complexity of the executeTransactionQueue() method by
factoring out the transaction simulation component. There is no
functional impact as part of this change - the relocated code performs
the same function as before.
    
Follow-up commits might make tweaks to improve transaction simulation
now that it is confined to a separate private method.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -80,15 +80,15 @@ export class MultiCallerClient {
   }
 
   async executeTransactionQueue(simulationModeOn = false) {
-    try {
-      if (this.transactions.length === 0) return;
-      this.logger.debug({
-        at: "MultiCallerClient",
-        message: "Executing tx bundle",
-        number: this.transactions.length,
-        simulationModeOn,
-      });
+    if (this.transactions.length === 0) return;
+    this.logger.debug({
+      at: "MultiCallerClient",
+      message: "Executing tx bundle",
+      number: this.transactions.length,
+      simulationModeOn,
+    });
 
+    try {
       // Simulate the transaction execution for the whole queue.
       const _transactionsSucceed = await Promise.all(
         this.transactions.map((transaction: AugmentedTransaction) => willSucceed(transaction))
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -89,65 +89,14 @@ export class MultiCallerClient {
...
-
-      const validTransactions = _transactionsSucceed
-        .filter((txn) => txn.succeed)
-        .map((transaction) => transaction.transaction);
-      if (validTransactions.length === 0) {
-        this.logger.debug({ at: "MultiCallerClient", message: "No valid transactions in the queue" });
+      const transactions = await this.simulateTransactionQueue(this.transactions);
+      if (transactions.length === 0) {
+        this.logger.debug({ at: "MultiCallerClient", message: "No valid transactions in the queue." });
         return;
```
pxrl, line 95: Note for follow-up change: we probably want to flush the transaction queue here (as per line 152).

nicholaspai (approved):
+1

src/clients/MultiCallerClient.ts
```diff
@@ -89,65 +89,14 @@ export class MultiCallerClient {
...
       }
 
-      const valueTransactions = validTransactions.filter((transaction) => transaction.value && transaction.value.gt(0));
-      const nonValueTransactions = validTransactions.filter(
-        (transaction) => !transaction.value || transaction.value.eq(0)
-      );
+      const valueTransactions = transactions.filter((transaction) => transaction.value && transaction.value.gt(0));
+      const nonValueTransactions = transactions.filter((transaction) => !transaction.value || transaction.value.eq(0));
```
mrice32, line 99: What changed here? Did they get moved in or out of a block? Sorry, just hard to tell from the diff.

mrice32 (approved):
LGTM!

src/clients/MultiCallerClient.ts
pxrl: It's a slight naming change, because the former validTransactions object no longer exists, so validTransactions.filter(...) => transactions.filter(...).

PR #391, opened 2022-12-14
feat: Enable route-specific rate model in ConfigStoreClient

nicholaspai:
Supports the ability for Across to charge different LP fees based on route.

Follow up PR will add feature to `sdk-v2` and the API in `frontend-v2`.

There is no risk to merging this now since `AcrossConfigStore` does not currently contain any `routeRateModels`

### DIFF
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -75,10 +75,12 @@ export class AcrossConfigStoreClient {
     // Test SNX deposit was before the rate model update for SNX.
     if (quoteBlock === 14856066) quoteBlock = 14856211;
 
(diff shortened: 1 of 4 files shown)


pxrl:
We have parallel discussion over [here](https://github.com/across-protocol/sdk-v2/pull/134#discussion_r1050223670), so i just wanted to note that in this PR :)

nicholaspai:
> We have parallel discussion over [here](https://github.com/across-protocol/sdk-v2/pull/134#discussion_r1050223670), so i just wanted to note that in this PR :)

Yes great reminder! Have responded [here](https://github.com/across-protocol/sdk-v2/pull/134#discussion_r1050364933) and will deprecate that PR in favor of this one

mrice32 (approved):
This looks awesome!

PR #393, opened 2022-12-17
improve: Construct AcrossAPIClient with only L1 tokens enabeld in HubPool

nicholaspai:
AcrossAPI fails to return data for l1 tokens not enabled in hub pool, so relayer should construct the APIClient only with enabled L1 tokens


pxrl (approved):
LGTM! Super quick turnaround on this!

PR #395, opened 2022-12-20
improve: Use lookback based on time for all chains in Finalizer, Relayer and Monitor

nicholaspai:
These functions currently require caller to pass in lookback per chain in terms of blocks. Can now specify in seconds which willl be same for all chains.

Dataworker currently sets lookback based on bundle end blocks, so that implicitly standardizes time across chains.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -52,7 +52,6 @@ export async function constructSpokePoolClientsWithLookback(
     )
   );
   );
-  const fromBlocks = Object.fromEntries(spokePools.map((obj, index) => [obj.chainId, _fromBlocks[index]]));
 
   return getSpokePoolClientsForContract(logger, configStoreClient, config, spokePools, fromBlocks);
 }
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -51,7 +51,6 @@ export async function constructSpokePoolClientsWithLookback(
         : spokePools.map(async (obj) => [obj.chainId, await getDeployedBlockNumber("SpokePool", obj.chainId)])
     )
   );
-  );
 
   return getSpokePoolClientsForContract(logger, configStoreClient, config, spokePools, fromBlocks);
 }
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -1,5 +1,12 @@
 import winston from "winston";
-import { getProvider, getDeployedContract, getDeploymentBlockNumber, Wallet, Contract } from "../utils";
+import {
+  getProvider,
+  getDeployedContract,
+  getDeploymentBlockNumber,
+  Wallet,
+  Contract,
+  getDeployedBlockNumber,
+} from "../utils";
 import { HubPoolClient, MultiCallerClient, AcrossConfigStoreClient, SpokePoolClient } from "../clients";
 import { CommonConfig } from "./Config";
 import { createClient } from "redis4";
@@ -41,7 +48,7 @@ export async function constructSpokePoolClientsWithLookback(
       ? spokePools.map((obj: { chainId: number; contract: Contract }) =>
           utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride)
         )
-      : spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
+      : spokePools.map((obj: { chainId: number }) => getDeployedBlockNumber("SpokePool", obj.chainId))
   );
   const fromBlocks = Object.fromEntries(
     spokePools.map((obj: { chainId: number }, i: number) => {
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -55,7 +55,7 @@ export async function constructSpokePoolClientsWithLookback(
   return getSpokePoolClientsForContract(logger, configStoreClient, config, spokePools, fromBlocks);
 }
 
-function getSpokePoolClientsForContract(
+export function getSpokePoolClientsForContract(
   logger: winston.Logger,
   configStoreClient: AcrossConfigStoreClient,
   config: CommonConfig,
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -43,17 +43,15 @@ export async function constructSpokePoolClientsWithLookback(
 
   // If initial lookback override is non-zero, then look up the l2 block number at that time, otherwise
   // use latest L2 block.
-  const _fromBlocks = await Promise.all(
-    initialLookBackOverride > 0
-      ? spokePools.map((obj: { chainId: number; contract: Contract }) =>
-          utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride)
-        )
-      : spokePools.map((obj: { chainId: number }) => getDeployedBlockNumber("SpokePool", obj.chainId))
-  );
   const fromBlocks = Object.fromEntries(
-    spokePools.map((obj: { chainId: number }, i: number) => {
-      return [obj.chainId, _fromBlocks[i]];
-    })
+    await Promise.all(
+      initialLookBackOverride > 0
+        ? spokePools.map((obj) => [
+            obj.chainId,
+            utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride),
+          ])
+        : spokePools.map((obj) => [obj.chainId, getDeployedBlockNumber("SpokePool", obj.chainId)])
+    )
   );
 
   return getSpokePoolClientsForContract(logger, configStoreClient, config, spokePools, fromBlocks);
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -4,6 +4,7 @@ import { HubPoolClient, MultiCallerClient, AcrossConfigStoreClient, SpokePoolCli
 import { CommonConfig } from "./Config";
 import { createClient } from "redis4";
 import { SpokePoolClientsByChain } from "../interfaces";
+import { utils } from "@across-protocol/sdk-v2"
 
 export interface Clients {
   hubPoolClient: HubPoolClient;
@@ -25,7 +26,7 @@ export async function constructSpokePoolClientsWithLookback(
   configStoreClient: AcrossConfigStoreClient,
   config: CommonConfig,
   baseSigner: Wallet,
-  initialLookBackOverride: { [chainId: number]: number } = {}
+  initialLookBackOverride = 0
 ): Promise<SpokePoolClientsByChain> {
   // Set up Spoke signers and connect them to spoke pool contract objects:
   const spokePoolSigners = getSpokePoolSigners(baseSigner, config);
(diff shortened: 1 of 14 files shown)


src/common/ClientHelper.ts
```diff
@@ -25,25 +26,28 @@ export async function constructSpokePoolClientsWithLookback(
...
-  const l2BlockNumbers = await Promise.all(
-    spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
+  // If initial lookback override is non-zero, then look up the l2 block number at that time, otherwise
+  // use latest L2 block.
+  const _fromBlocks = await Promise.all(
+    initialLookBackOverride > 0
+      ? spokePools.map((obj: { chainId: number; contract: Contract }) =>
+          utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride)
+        )
+      : spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
```
pxrl, line 44: This will mean that we only look at 1 block, right? Do we know of any scenarios where we'd actually want to do that?

src/common/Config.ts
```diff
@@ -37,19 +37,15 @@ export class CommonConfig {
 
     // `maxRelayerLookBack` is how far we fetch events from, modifying the search config's 'fromBlock'
     this.maxRelayerLookBack = MAX_RELAYER_DEPOSIT_LOOK_BACK
```
pxrl, line 39: Minor: _Might_ take up less space as:

`this.maxRelayerLookBack = Number(MAX_RELAYER_DEPOSIT_LOOK_BACK ?? Constants.MAX_RELAYER_DEPOSIT_LOOK_BACK);`

src/common/Constants.ts
```diff
@@ -4,14 +4,8 @@ export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
...
-// Target ~2 days per chain. Avg. block times: { 1: 12s, 10/42161: ~0.25s, 137: 2.5s, 288: 30s }
-export const MAX_RELAYER_DEPOSIT_LOOK_BACK: { [chainId: number]: number } = {
-  1: 14400,
-  10: 691200,
-  137: 69120,
-  288: 5760,
-  42161: 691200,
-};
+// Target ~2 days per chain.
+export const MAX_RELAYER_DEPOSIT_LOOK_BACK = 2 * 24 * 60 * 60;
```
pxrl, line 8: Should we consider reducing this default for third parties? 2 days is really RPC-intensive, so I don't think the average user would want to run with that. We can easily override it in our own env. Plucking a number out of thin air...3 hours might be a more reasonable default?

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -87,6 +89,31 @@ export function spokePoolClientsToProviders(spokePoolClients: { [chainId: number
...
+  startBlockOverride: { [chainId: number]: number } = {},
+  toBlockOverride: { [chainId: number]: number } = {}
+): Promise<SpokePoolClientsByChain> {
+  // Set up Spoke signers and connect them to spoke pool contract objects:
+  const spokePoolSigners = getSpokePoolSigners(baseSigner, config);
+  const spokePools = config.spokePoolChains.map((chainId) => {
+    return { chainId, contract: getDeployedContract("SpokePool", chainId, spokePoolSigners[chainId]) };
+  });
+
+  // If no lookback is set, fromBlock will be set to spoke pool's deployment block.
```
pxrl, line 106: Does this work in practice with our default CHAIN_MAX_BLOCK_LOOK_BACK config?

I've got a vague idea that we might hit some RPC issues by requesting so far back whilst also requesting eth_getLogs() over such a large range of blocks.

src/utils/FillUtils.ts
```diff
@@ -1,6 +1,14 @@
...
 import { DepositWithBlock, Fill, FillsToRefund, FillWithBlock, SpokePoolClientsByChain } from "../interfaces";
-import { BigNumber, assign, getRealizedLpFeeForFills, getRefundForFills, sortEventsDescending, toBN } from "./";
+import {
+  BigNumber,
+  assign,
+  getRealizedLpFeeForFills,
+  getRefundForFills,
+  sortEventsDescending,
+  toBN,
+  MAX_UINT_VAL,
```
pxrl, line 10: Was this for lint or something?

pxrl (commented):
Have left a few comments, but the overriding takeaway is that it's _really_ nice to simplify the lookback config.

src/common/ClientHelper.ts
nicholaspai: ah good point I should replace getBlockNumber() with the deployment block

src/common/Config.ts
nicholaspai: typeof Number(undefined) == "number" (its weirdly the special number "NaN" (not a number)) so idk if this would work

src/common/Constants.ts
nicholaspai: hmmm yeah lets go with 4 hours i like powers of 2 :P

src/dataworker/DataworkerClientHelper.ts
nicholaspai: this is just a copy paste of a function that existed in common/ClientHelper but its only used for the Dataworker so I moved it to this file

nicholaspai: so it has been working in practice

src/utils/FillUtils.ts
nicholaspai: yep

nicholaspai:
> Have left a few comments, but the overriding takeaway is that it's _really_ nice to simplify the lookback config.

Yeah once we verify this works on these functions we can add it to all other block based lookups through out app, there are others!

src/common/ClientHelper.ts
```diff
@@ -25,25 +33,28 @@ export async function constructSpokePoolClientsWithLookback(
...
-  // For each spoke chain, look up its latest block and adjust by lookback configuration to determine
-  // fromBlock. If no lookback is set, fromBlock will be set to spoke pool's deployment block.
-  const fromBlocks: { [chainId: number]: number } = {};
-  const l2BlockNumbers = await Promise.all(
-    spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
+  // If initial lookback override is non-zero, then look up the l2 block number at that time, otherwise
+  // use latest L2 block.
+  const _fromBlocks = await Promise.all(
+    initialLookBackOverride > 0
+      ? spokePools.map((obj: { chainId: number; contract: Contract }) =>
```
mrice32, line 48: Was it not able to infer the type of obj? If not, no problem, I just figured it would do that for you.

src/common/ClientHelper.ts
```diff
@@ -25,25 +33,28 @@ export async function constructSpokePoolClientsWithLookback(
...
+          utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride)
+        )
+      : spokePools.map((obj: { chainId: number }) => getDeployedBlockNumber("SpokePool", obj.chainId))
```
mrice32, line 51: nit: if you return pairs from these loops, you can avoid the second loop entirely (since you will already have chainId-blockNumber pairs to pass into Object.fromEntries):
```ts
spokePools.map((obj: { chainId: number }) => [chainId, getDeployedBlockNumber("SpokePool", obj.chainId)])
```

mrice32 (approved):
This looks great!

src/utils/FillUtils.ts
pxrl: Just to double-check, I am referring specifically to the addition of MAX_UINT_VAL. Maybe I am missing something, but I don't see it used anywhere else in this file :thinking:

test/Relayer.UnfilledDeposits.ts
```diff
@@ -124,7 +125,7 @@ describe("Relayer: Unfilled Deposits", async function () {
     const deposit1Complete = await buildDepositStruct(deposit1, hubPoolClient, configStoreClient, l1Token);
     const deposit2Complete = await buildDepositStruct(deposit2, hubPoolClient, configStoreClient, l1Token);
 
-    expect(getUnfilledDeposits(relayerInstance.clients.spokePoolClients))
+    expect(getUnfilledDeposits(relayerInstance.clients.spokePoolClients, 1 * 24 * 60 * 60))
```
pxrl, line 128: Minor maintainability nit: Consider defining this globally in this file so that it's easier to update in future?

src/dataworker/DataworkerClientHelper.ts
pxrl: Do you know whether we have tested it with full default settings, though? We usually override a lot here, so the fact that it works for us _might_ not be indicative of the default case.

src/common/Config.ts
pxrl: I'm pretty sure it should work fine. Demonstration below, just for consideration. It relies on the variable TEST_VAR being declared, but I'm pretty sure that we can rely on that.

```
> let TEST_VAR = undefined
undefined

> Number(TEST_VAR ?? 456)
456
```

nicholaspai: oh I missed your parentheses you're right this should work

src/dataworker/DataworkerClientHelper.ts
nicholaspai: what kind of issues do you forsee? The default lookbacks are:

```
export const CHAIN_MAX_BLOCK_LOOKBACK = {
  1: 0, // Note: 0 gets defaulted to infinity lookback
  10: 0,
  137: 3490,
  288: 4990,
  42161: 99990,
};
```

Which should handle every RPC regardless of lookback right? Well, Actually, we might ened to change the 42161 and 10 lookbacks to be 10000 to work with QuickNode

src/utils/FillUtils.ts
nicholaspai: oh let me check

nicholaspai: you're right! removed

src/common/Config.ts
nicholaspai: Have fixed this and other instances in this file as well!

src/common/ClientHelper.ts
```diff
@@ -25,25 +33,22 @@ export async function constructSpokePoolClientsWithLookback(
...
-  // fromBlock. If no lookback is set, fromBlock will be set to spoke pool's deployment block.
-  const fromBlocks: { [chainId: number]: number } = {};
-  const l2BlockNumbers = await Promise.all(
-    spokePools.map((obj: { contract: Contract }) => obj.contract.provider.getBlockNumber())
+  // If initial lookback override is non-zero, then look up the l2 block number at that time, otherwise
+  // use latest L2 block.
+  const _fromBlocks = await Promise.all(
+    initialLookBackOverride > 0
+      ? spokePools.map((obj) => utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride))
+      : spokePools.map((obj) => getDeployedBlockNumber("SpokePool", obj.chainId))
```
mrice32, line 49: The way you had it before your fix commit was good, I just think we needed some extra async/await keywords.

I think this should work to avoid two loops:
```suggestion
  const fromBlocks = Object.fromEntries(
    await Promise.all(
      initialLookBackOverride > 0
        ? spokePools.map(async (obj) => [
            obj.chainId,
            await utils.findBlockAtOrOlder(obj.contract.provider, initialLookBackOverride),
          ])
        : spokePools.map(async (obj) => [obj.chainId, await getDeployedBlockNumber("SpokePool", obj.chainId)])
    )
  );
```

src/common/ClientHelper.ts
```diff
@@ -25,25 +33,22 @@ export async function constructSpokePoolClientsWithLookback(
...
   );
-  spokePools.forEach((obj: { chainId: number; contract: Contract }, index) => {
-    if (initialLookBackOverride[obj.chainId]) {
-      fromBlocks[obj.chainId] = l2BlockNumbers[index] - initialLookBackOverride[obj.chainId];
-    }
-  });
+  const fromBlocks = Object.fromEntries(spokePools.map((obj, index) => [obj.chainId, _fromBlocks[index]]));
```
mrice32, line 51: I couldn't suggest across a blank line, so I think this line isn't needed if the above suggestion is taken.

```suggestion
```

mrice32 (approved):
LGTM!

PR #396, opened 2022-12-23
fix: Use BlockFinder instead of sdk.findBlockAtOrOlder

nicholaspai:
findBlockAtOrOlder doesn't always work causing some timeouts bc lookback defaults to full range﻿

Added comment about buggy code in sdk-v2 [here](https://github.com/across-protocol/sdk-v2/pull/137)

### DIFF
--- a/src/clients/ConfigStoreClient.ts
+++ b/src/clients/ConfigStoreClient.ts
@@ -12,6 +12,8 @@ import {
   MakeOptional,
   toBN,
   max,
+  getBlockForTimestamp,
+  shouldCache,
 } from "../utils";
 
 import {
@@ -259,16 +261,7 @@ export class AcrossConfigStoreClient {
   }
 
   private async getBlockNumber(timestamp: number) {
-    if (!this.redisClient) return (await this.blockFinder.getBlockForTimestamp(timestamp)).number;
-    const key = `block_number_${timestamp}`;
-    const result = await this.redisClient.get(key);
-    if (result === null) {
-      const blockNumber = (await this.blockFinder.getBlockForTimestamp(timestamp)).number;
-      if (this.shouldCache(timestamp)) await this.redisClient.set(key, blockNumber.toString());
-      return blockNumber;
-    } else {
-      return parseInt(result);
-    }
+    return await getBlockForTimestamp(1, timestamp, getCurrentTime(), this.blockFinder, this.redisClient);
   }
 
   private async getUtilization(l1Token: string, blockNumber: number, amount: BigNumber, timestamp: number) {
@@ -277,17 +270,12 @@ export class AcrossConfigStoreClient {
     const result = await this.redisClient.get(key);
     if (result === null) {
       const { current, post } = await this.hubPoolClient.getPostRelayPoolUtilization(l1Token, blockNumber, amount);
-      if (this.shouldCache(timestamp)) await this.redisClient.set(key, `${current.toString()},${post.toString()}`);
+      if (shouldCache(getCurrentTime(), timestamp))
+        await this.redisClient.set(key, `${current.toString()},${post.toString()}`);
       return { current, post };
     } else {
       const [current, post] = result.split(",").map(BigNumber.from);
       return { current, post };
     }
-  }
-
-  // Avoid caching calls that are recent enough to be affected by things like reorgs.
-  private shouldCache(eventTimestamp: number) {
-    // Current time must be >= 5 minutes past the event timestamp for it to be stable enough to cache.
-    return getCurrentTime() - eventTimestamp >= 300;
   }
 }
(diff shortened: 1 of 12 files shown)


pxrl (approved):
LGTM

src/utils/RedisUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+// Get the block number for a given timestamp fresh from on-chain data if not found in redis cache.
+export async function getBlockForTimestamp(
+  chainId: number,
+  timestamp: number,
+  currentChainTime: number,
+  blockFinder: BlockFinder<Block>,
+  redisClient?: RedisClient
+): Promise<number> {
+  if (!redisClient) return (await blockFinder.getBlockForTimestamp(timestamp)).number;
+  const key = chainId === 1 ? `block_number_${timestamp}` : `${chainId}_block_number_${timestamp}`;
```
mrice32, line 20: Very cool! Nice idea to take advantage of the cache here.

src/utils/RedisUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// Get the block number for a given timestamp fresh from on-chain data if not found in redis cache.
+export async function getBlockForTimestamp(
+  chainId: number,
+  timestamp: number,
+  currentChainTime: number,
+  blockFinder: BlockFinder<Block>,
+  redisClient?: RedisClient
+): Promise<number> {
+  if (!redisClient) return (await blockFinder.getBlockForTimestamp(timestamp)).number;
```
mrice32, line 19: The blockfinder is pretty efficient at finding mainnet blocks because it uses an average block time to guess where the timestamp will roughly be.

Before we put it into action here, I think we should make sure we can seed it with a decent average block time for each L2 (probably better to underestimate the time so it shoot farther back than it should by default).

mrice32 (commented):
Looks great! Only comment is that we should give the blockfinder a more realistic block time for each L2. I think we should be able to do this by allowing the caller to override the way the block finder estimates the location of a block: https://github.com/UMAprotocol/protocol/blob/25d22a9088a64da8f6783130c65d717a9104015d/packages/sdk/src/blockFinder.ts#L34.

src/clients/ConfigStoreClient.ts
```diff
@@ -259,16 +261,7 @@ export class AcrossConfigStoreClient {
...
-    const key = `block_number_${timestamp}`;
-    const result = await this.redisClient.get(key);
-    if (result === null) {
-      const blockNumber = (await this.blockFinder.getBlockForTimestamp(timestamp)).number;
-      if (this.shouldCache(timestamp)) await this.redisClient.set(key, blockNumber.toString());
-      return blockNumber;
-    } else {
-      return parseInt(result);
-    }
+    return await getBlockForTimestamp(1, timestamp, getCurrentTime(), this.blockFinder, this.redisClient);
```
mrice32, line 264: You don't have to add this now, but if you share a blockfinder between this and the spoke pool client for mainnet, you can eek out a bit more efficiency since the BlockFinder is faster the more searches it performs.

Not super important or anything, though.

nicholaspai: So something like this is what you hand in mind? https://github.com/UMAprotocol/protocol/pull/4351/files

src/utils/RedisUtils.ts
nicholaspai: Updated to the updated blockfinder available in @uma/financial-templates-lib:2.32.1 that has good average block time estimates for all chains now

nicholaspai:
@mrice32 this PR now uses the new blockfinder available in @uma/financial-tempaltes-lib:2.32.1 with better average block time estimates for all chains

src/clients/ConfigStoreClient.ts
mrice32: +1

src/utils/RedisUtils.ts
mrice32: This all looks great!

src/clients/ConfigStoreClient.ts
mrice32: Hmmmm, I don't think so. I was basically just saying it would be more efficient if we only had one instance of a BlockFinder for mainnet across both the ConfigStoreClient and the SpokePoolClient. Right now, we create a separate instance for each. This isn't a big deal, it just means that the cache for one isn't improving the efficiency of the other. Not a big deal, though.

mrice32 (approved):
LGTM!

I think a good tech debt item for the backlog is to create a redis mock that we can use so we can have unit and integration tests for changes like this since they are hard to get right and can break prod.

src/clients/ConfigStoreClient.ts
nicholaspai: Ah I see what you're saying I actually responded to the wrong comment with mine then

nicholaspai: cool I implemented this https://github.com/across-protocol/relayer-v2/pull/396/commits/09bbe9095c6dcf4f449d5aa473ef40d89a887046

PR #397, opened 2022-12-26
fix(validation-script): Handle case where dispute is atomically bundled with new proposal

nicholaspai:
This PR makes the validation script more robust when handling atomic disputes (i.e. disputes sent in same block as subsequent proposals) by querying the PriceRequestAdded event emitted by the disputed root bundle event. The script now throws if the
found proposal and filtering out events that don't precede the dispute---by block number, transaction index, or log index.

Additionally, the hub pool client's getMostRecentProposedRootBundle now uses a < comparison instead of a <= to find a root bundle earlier than a block number. This 
function is only used in one place by the Dataworker where it does not change behavior.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -236,10 +236,8 @@ export class HubPoolClient {
     );
   }
 
-  getMostRecentProposedRootBundle(latestBlockToSearch: number) {
-    return sortEventsDescending(this.proposedRootBundles).find(
-      (proposedRootBundle: ProposedRootBundle) => proposedRootBundle.blockNumber < latestBlockToSearch
-    ) as ProposedRootBundle;
+  getLatestProposedRootBundle() {
+    return sortEventsDescending(this.proposedRootBundles)[0] as ProposedRootBundle;
   }
 
   getFollowingRootBundle(currentRootBundle: ProposedRootBundle) {
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -238,7 +238,7 @@ export class HubPoolClient {
 
   getMostRecentProposedRootBundle(latestBlockToSearch: number) {
     return sortEventsDescending(this.proposedRootBundles).find(
-      (proposedRootBundle: ProposedRootBundle) => proposedRootBundle.blockNumber <= latestBlockToSearch
+      (proposedRootBundle: ProposedRootBundle) => proposedRootBundle.blockNumber < latestBlockToSearch
     ) as ProposedRootBundle;
   }
 
--- a/src/scripts/validateRootBundle.ts
+++ b/src/scripts/validateRootBundle.ts
@@ -18,9 +18,8 @@ import {
   startupLogLevel,
   Logger,
   getLatestInvalidBundleStartBlocks,
-  Contract,
-  isEventOlder,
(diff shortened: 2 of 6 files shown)


src/scripts/validateRootBundle.ts
```diff
@@ -56,6 +59,23 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet) {
 
   await updateDataworkerClients(clients, false);
   const precedingProposeRootBundleEvent = clients.hubPoolClient.getMostRecentProposedRootBundle(priceRequestBlock);
+
+  // Find dispute transaction so we can gain additional confidence that the preceding root bundle is older than the
```
mrice32, line 63: Interesting check! IIUC, this would fail only in some weird situation with two proposals and a dispute in the same txn or something like that.

mrice32 (approved):
LGTM!

src/clients/HubPoolClient.ts
```diff
@@ -238,7 +238,7 @@ export class HubPoolClient {
 
   getMostRecentProposedRootBundle(latestBlockToSearch: number) {
     return sortEventsDescending(this.proposedRootBundles).find(
-      (proposedRootBundle: ProposedRootBundle) => proposedRootBundle.blockNumber <= latestBlockToSearch
+      (proposedRootBundle: ProposedRootBundle) => proposedRootBundle.blockNumber < latestBlockToSearch
```
mrice32, line 241: Would this have any effect on the dispute or proposal logic in the dataworker? Just want to make sure there wasn't some deeper reason for the <= here.

src/scripts/validateRootBundle.ts
nicholaspai: yeah exactly its a really strange edge case

src/clients/HubPoolClient.ts
nicholaspai: It doesn't but I've changed the HubPoolCLient interface a bit with a new method `getLatestProposedRootBundle` to make it more clear this has no effect

src/scripts/validateRootBundle.ts
```diff
@@ -54,8 +58,27 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet) {
...
+  const dispute = disputes.find((e) => e.args.time.toString() === priceRequestTime.toString());
+  if (!dispute) throw new Error("Could not find PriceRequestAdded event on DVM matching price request time");
+
   await updateDataworkerClients(clients, false);
-  const precedingProposeRootBundleEvent = clients.hubPoolClient.getMostRecentProposedRootBundle(priceRequestBlock);
+  const precedingProposeRootBundleEvent = sortEventsDescending(clients.hubPoolClient.getProposedRootBundles()).find(
+    (e) => isEventOlder(e as SortableEvent, dispute as SortableEvent)
+  );
+
+  if (!precedingProposeRootBundleEvent) throw new Error("Failed to identify a root bundle preceding dispute");
```
pxrl, line 80: Consider factoring this out to a separate function? That would make it easier to verify this specific functionality, and would avoid adding to the length of this function.

nicholaspai: good idea will add a utlity method `getDisputedProposal(disputeRequestTimestamp: number): ProposedRootBundle`

PR #398, opened 2022-12-27
feat: Support version key in ConfigStoreClient

nicholaspai:
Implements [this feature](https://github.com/UMAprotocol/UMIPs/pull/567) in the UMIP.

This PR is backwards compatible meaning that it will not break the currently running relayers and dataworkers because the default config store version is 0 and there have been no "VERSION" updates as of  December 29 2022 in the `ConfigStore` contract.

This PR adds a new in-memory list to the ConfigStoreClient that saves the history of "VERSION" updates in the `ConfigStoreClient`. The relayer uses this by checking the required version for each deposit's quote timestamp before relaying. If it doesn't have the latest version then it emits a `warn` log and skips that deposit. The dataworker refuses to propose if it doesn't have the latest/highest version and emits a `warn` log, and it refuses to dispute and emits an `error` level log. If the dataworker is trying to execute spoke leaves, then it will similarly exit early and emit a `warn` log if it doesn't have the highest config store version as of the bundle's mainnet end block.


test/Monitor.ts
```diff
@@ -86,6 +86,9 @@ describe("Monitor", async function () {
       CONFIGURED_NETWORKS: JSON.stringify(configuredNetworks),
     });
 
+    // Set the config store version to 0 to match the default version in the ConfigStoreClient.
+    process.env.CONFIG_STORE_VERSION = "0";
```
pxrl, line 90: Also RE my comment to the UMIP update, I'm a bit unclear if this is a config store version, or something broader. The naming here indicates it is a version of the config store, in which case we'd probably want to bump that version each time we push a new config. If so, I think semver would probably make more sense, so we can distinguish major (breaking) changes, minor changes, and configuration tweaks with no impact.

src/utils/ConfigStoreUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { AcrossConfigStoreClient } from "../clients";
+import { CONFIG_STORE_VERSION } from "../common";
+require("dotenv").config();
+
+export function checkConfigStoreVersion(client: AcrossConfigStoreClient): void {
+  const localVersion = process.env.CONFIG_STORE_VERSION ?? CONFIG_STORE_VERSION;
```
pxrl, line 6: Is it useful for people to override this via configuration? Intuitively I feel we should leave that for developers to do - it seems dangerous to expose that to users.

test/ConfigStoreClient.ts
```diff
@@ -270,6 +270,12 @@ describe("AcrossConfigStoreClient", async function () {
     });
   });
   describe("GlobalConfig", function () {
+    it("Gets latest version", async function () {
+      await configStore.updateGlobalConfig(utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.VERSION), "6");
```
pxrl, line 274: During the dev support call we discussed whether this should be made available via the sdk instead. I think that makes sense, because third-party implementations are expected to import the SDK to access some of the key functionalities (relayerFeeCalculator, etc...).

If we bundled this function into the SDK then we could also hardcode the compatible/supported version there as well. It seems like a cleaner approach to me. wdyt?

src/utils/ConfigStoreUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { AcrossConfigStoreClient } from "../clients";
+import { CONFIG_STORE_VERSION } from "../common";
+require("dotenv").config();
+
+export function checkConfigStoreVersion(client: AcrossConfigStoreClient): void {
```
pxrl, line 5: During the dev support call we discussed whether this should be made available via the sdk instead. I think that makes sense, because third-party implementations are expected to import the sdk to access some of the key functionalities (relayerFeeCalculator, etc...).

If we bundled this function into the SDK then we could also hardcode the compatible/supported version there as well. It seems like a cleaner approach to me. wdyt?

test/Monitor.ts
nicholaspai: >  we'd probably want to bump that version each time we push a new config. I don't think that's the intention

This was actually my intention, what's wrong with that? I don't think semver as the first version is something we should consider, as it stands every single update is a "major" breaking one

src/utils/ConfigStoreUtils.ts
nicholaspai: hmm this seems reasonable but same question as https://github.com/across-protocol/relayer-v2/pull/398#discussion_r1058689508

src/utils/ConfigStoreUtils.ts
nicholaspai: Ah I figured actually using a MockConfigStoreClient with an overridden function `validateVersion` would work

src/utils/ConfigStoreUtils.ts
nicholaspai: https://github.com/across-protocol/sdk-v2/pull/138

src/clients/ConfigStoreClient.ts
```diff
@@ -245,6 +249,8 @@ export class AcrossConfigStoreClient {
         if (!isNaN(args.value)) this.cumulativeMaxRefundCountUpdates.push(args);
       } else if (args.key === utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.MAX_POOL_REBALANCE_LEAF_SIZE)) {
         if (!isNaN(args.value)) this.cumulativeMaxL1TokenCountUpdates.push(args);
+      } else if (args.key === utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.VERSION)) {
+        if (!isNaN(args.value)) this.configStoreVersion = args.value;
```
mrice32, line 253: Do we want to store only the most up-to-date one? Or do we want to track the config store versions over time so we know which version a relay or bundle was constructed with?

How should we handle this being changed? Do we assume all changes are backwards compatible, so a future version can always handle old relays that existed in a previous version?

mrice32 (commented):
Looks great! I have one question around how we handle backwards compatibility and versioning.

src/clients/ConfigStoreClient.ts
nicholaspai: do you see a reason why we'd want to track the versions over time? is there somewhere in the bot we can use this now? Otherwise I was thinking we could add this feature in the future, since the version change is forever stored on the block chain, but I don't see why we'd want to check them right now

src/clients/ConfigStoreClient.ts
```diff
@@ -143,15 +149,40 @@ export class AcrossConfigStoreClient {
       (config) => config.blockNumber <= blockNumber
     );
     if (!config) throw new Error(`Could not find MaxRefundCount before block ${blockNumber}`);
-    return config.value;
+    return Number(config.value);
```
nicholaspai, line 152: Miscellaneous fix here and on 160 where the function type expects to return a `number` but sometimes returns a `string` so I explicitly cast it here

src/clients/ConfigStoreClient.ts
```diff
@@ -143,15 +149,40 @@ export class AcrossConfigStoreClient {
...
+
+  getConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number {
+    const config = (sortEventsDescending(this.cumulativeConfigStoreVersionUpdates) as ConfigStoreVersionUpdate[]).find(
+      (config) => config.timestamp <= timestamp
+    );
+    if (!config) return DEFAULT_CONFIG_STORE_VERSION;
+    return Number(config.value);
+  }
+
+  getTimeForConfigStoreVersion(version: number): number {
```
nicholaspai, line 171: This function isn't used yet in this PR but I imagine we will use it to start adding "feature flags" to the relayer. Imagine we want to add the `routeRateModel` feature, we might want to only enable that after the timestamp returned by `getTimeForConfigStoreVersion(1)`

src/clients/ConfigStoreClient.ts
```diff
@@ -247,6 +282,22 @@ export class AcrossConfigStoreClient {
         if (!isNaN(args.value)) this.cumulativeMaxRefundCountUpdates.push(args);
       } else if (args.key === utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.MAX_POOL_REBALANCE_LEAF_SIZE)) {
         if (!isNaN(args.value)) this.cumulativeMaxL1TokenCountUpdates.push(args);
+      } else if (args.key === utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.VERSION)) {
+        // Skip version updates that aren't larger than the previous version. The first version must be > 0.
```
nicholaspai, line 286: This logic is to avoid shooting ourselves in foot where we update to a version that is accidentally lower or equal to the previous one, or is not a number

mrice32: +1

src/clients/ConfigStoreClient.ts
```diff
@@ -247,6 +282,22 @@ export class AcrossConfigStoreClient {
...
+            Number(args.value) <=
+            Number(this.cumulativeConfigStoreVersionUpdates[this.cumulativeConfigStoreVersionUpdates.length - 1].value)
+          ) {
+            continue;
+          }
+          this.cumulativeConfigStoreVersionUpdates.push({
+            ...args,
+            timestamp: globalConfigUpdateTimes[i],
+          });
+        }
```
mrice32, line 300: nit: I would suggest trying to decompose these conditions a bit and unnesting, so it's clearer what we're checking for:

```ts
// If not a number, skip.
if (isNaN(args.value)) continue;
const value = Number(args.value);

// If not an integer, skip.
if (!Number.isInteger(value)) continue;

// Extract last version
const lastValue = this.cumulativeConfigStoreVersionUpdates.length === 0 ? DEFAULT_CONFIG_STORE_VERSION : Number(this.cumulativeConfigStoreVersionUpdates[this.cumulativeConfigStoreVersionUpdates.length - 1].value);

// If version is not >= last version, skip.
if (value <= lastValue) continue;

this.cumulativeConfigStoreVersionUpdates.push({
  ...args,
  timestamp: globalConfigUpdateTimes[i],
});
```

src/dataworker/Dataworker.ts
```diff
@@ -234,6 +234,18 @@ export class Dataworker {
       return;
     }
 
+    // If config store version isn't up to date, return early. This is a simple rule that is perhaps too aggressive
+    // but the proposer role is a specialized one and the user should always be using updated software.
```
mrice32, line 238: +1, I think we can tailor this more if needed, but I really like the simplicity of this. +1

src/dataworker/Dataworker.ts
```diff
@@ -647,6 +659,23 @@ export class Dataworker {
       1,
       this.chainIdListForBundleEvaluationBlockNumbers
     )[1];
+    // If config store version isn't up to date, return early. This is a simple rule that is perhaps too aggressive
+    // but the proposer role is a specialized one and the user should always be using updated software.
+    if (!this.clients.configStoreClient.hasValidConfigStoreVersionForTimestamp()) {
+      this.logger.debug({
+        at: "Dataworke#validate",
```
mrice32, line 666:
```suggestion
        at: "Dataworker#validate",
```

src/dataworker/Dataworker.ts
```diff
@@ -39,15 +39,15 @@ import {
...
+import { CONFIG_STORE_VERSION, IGNORED_SPOKE_BUNDLES } from "../common";
 import { isOvmChain, ovmWethTokens } from "../clients/bridges";
 
 // Internal error reasons for labeling a pending root bundle as "invalid" that we don't want to submit a dispute
 // for. These errors are due to issues with the dataworker configuration, instead of with the pending root
 // bundle. Any reasons in IGNORE_X we emit a "debug" level log for, and any reasons in ERROR_X we emit an "error"
 // level log for.
 const IGNORE_DISPUTE_REASONS = new Set(["bundle-end-block-buffer"]);
-const ERROR_DISPUTE_REASONS = new Set(["insufficient-dataworker-lookback"]);
+const ERROR_DISPUTE_REASONS = new Set(["insufficient-dataworker-lookback", "out-of-date-config-store-version"]);
```
mrice32, line 50: +1 -- really like how clear this is.

src/utils/FillUtils.ts
```diff
@@ -186,7 +197,14 @@ export function getUnfilledDeposits(
     }
   }
 
-  return unfilledDeposits;
+  return unfilledDeposits.map((unfilledDeposit) => {
+    return {
+      ...unfilledDeposit,
+      requiresNewConfigStoreVersion: !configStoreClient.hasValidConfigStoreVersionForTimestamp(
```
mrice32, line 203: Do you have any concerns about us looping over these config store events on every deposit? Do you think this will slow things down?

Thoughts on skipping this deeper check if we already know the config is 100% up-to-date?

mrice32 (approved):
LGTM! only minor comments

src/utils/FillUtils.ts
nicholaspai: This is a good point, i'll add an additional check in the config store client to let the relayer know when its 100% up to date

nicholaspai:
https://github.com/across-protocol/relayer-v2/pull/398#discussion_r1059495480

@mrice32 i thnk its because of the way some tests are written and also how data is written differently on UI's versus hardhat, this function is a bit untested because we rarely run into the refund limit per chain or l1 token count per chain (we usually only refund 1-2 relayers and only refund <5 tokens)

src/clients/ConfigStoreClient.ts
nicholaspai: great call and i love the additional integer check. This is a lot cleaner

mrice32:
> [#398 (comment)](https://github.com/across-protocol/relayer-v2/pull/398#discussion_r1059495480)
> 
> @mrice32 i thnk its because of the way some tests are written and also how data is written differently on UI's versus hardhat, this function is a bit untested because we rarely run into the refund limit per chain or l1 token count per chain (we usually only refund 1-2 relayers and only refund <5 tokens). So I wanted to be safe

Gotcha, makes sense. That's really annoying, as it should be the same regardless.

src/clients/ConfigStoreClient.ts
```diff
@@ -143,15 +149,36 @@ export class AcrossConfigStoreClient {
...
+
+  getConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number {
+    const config = (sortEventsDescending(this.cumulativeConfigStoreVersionUpdates) as ConfigStoreVersionUpdate[]).find(
+      (config) => config.timestamp <= timestamp
+    );
+    if (!config) return DEFAULT_CONFIG_STORE_VERSION;
+    return Number(config.value);
+  }
+
+  hasLatestConfigStoreVersion(): boolean {
```
mrice32, line 171: nit: since we call this many many times in the relayer loop, any reason we can't compute the return value at the end of the update() method once and avoid having to call through to getConfigStoreVersionForTimestamp() on each call since the value shouldn't change?

I only say this because getConfigStoreVersionForTimestamp does a sort + find, which is a little heavy.

mrice32 (approved):
Looks great! One comment

src/clients/ConfigStoreClient.ts
nicholaspai: this is a great idea, we should 100% take advantage of the fact that it won't change after the initial update!

PR #400, opened 2023-01-01
improve(ProviderUtils): log about disagreeing providers

nicholaspai:
We want to know anytime a provider disagrees with another. Currently we only know this if we don't reach quorum. So this PR logs any time we do reach quorum but have disagreeing providers

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -216,10 +216,16 @@ class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     const mismatchedProviders = [...values, ...fallbackValues]
       .filter(([, result]) => !lodash.isEqual(result, quorumResult))
       .map(([provider]) => provider.connection.url);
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -173,6 +188,13 @@ class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
         this._trySend(provider, method, params)
           .then((result): [ethers.providers.StaticJsonRpcProvider, any] => [provider, result])
           .catch((err) => {
+            const errString = err?.stack || err?.toString();
+            logger.debug({
+              at: "ProviderUtils`",
+              message: "Fallback provider failed",
+              provider: provider.connection.url,
+              errString,
+            });
```
nicholaspai, line 197: I'm not sure if we're gonna miss any details about fallback providers that only request after the first batch of providers fail, return, but don't match with the quorum.

mrice32 (approved):
LGTM!

PR #401, opened 2023-01-03
feat: Ignore set keys for different RPC methods when comparing

nicholaspai:


src/utils/ProviderUtils.ts
```diff
@@ -153,7 +153,29 @@ class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
 
     // Start at element 1 and begin comparing.
     // If _all_ values are equal, we have hit quorum, so return.
-    if (values.slice(1).every(([, output]) => lodash.isEqual(values[0][1], output))) return values[0][1];
+    function compareResults(rpcResultA: any, rpcResultB: any): boolean {
```
mrice32, line 156: nit: if we passed in `method`, could we move this function completely out of this class so it's a pure utility function that's only created once in global scope? This might speed things up a little.

src/utils/ProviderUtils.ts
```diff
@@ -153,7 +153,29 @@ class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+      const ignoreFilteredKeys = (ignoredKeys: string[]) => {
+        // This uses native JS spread to create a copy of the object and set ignored keys undefined rather than use
+        // lodash.pick because of performance reasons.
+        const copyA = { ...rpcResultA };
+        const copyB = { ...rpcResultB };
```
mrice32, line 161: I think this is _probably_ fine, but if we have lots of concurrent RPC calls, this could be a big memory and performance hit. I think we can just go with this for now, but if we do see a perf problem, we could try:
1. Instead of a full copy, we could remove the ignored keys from the original and store them in a temp objects (only those keys, not the full thing), compare the originals without the ignored keys, and then write the removed elements back.
2. (A little weird) if we assume we will never use the ignored fields, we could just set `rpcResultB[key] = rpcResultA[key]` to avoid all copying or moving values around. This would modify the original and could cause the result to have an odd value for the ignored fields, but it could speed things up.

Both of these are more complex, so I think the existing solution is fine -- we just may need to resort to these if we see perf issues.

mrice32 (approved):
LGTM!

src/utils/ProviderUtils.ts
nicholaspai: agreed!

src/utils/ProviderUtils.ts
nicholaspai: I actually really like algorithm 1 you listed above because it should be faster than cloning the `rpcResult`. The only side effect that I can see is that the `rpcResult` object will have keys in a different order. For example if we delete `miner` and then re-add it it should get added to the end of the object. This shouldn't be an issue but I've added a comment

nicholaspai: https://github.com/across-protocol/relayer-v2/pull/401/commits/db03453c6f1d51abce50883fca6b9c4610bff8ca

mrice32 (approved):
LGTM!

PR #404, opened 2023-01-09
improve(ProviderUtils): Override ethers 429 backoff

pxrl:
The default ethers implementation is very aggressive and will regularly retry immediately after a 429 response. This might be OK for serialised requests, but is not very compatible with the heavily-parallel request load of this bot.

This change sets a new minimum delay of (1000 * failure) ms, where failure is [1, 2, 3, ...]. An additional random delay between 0 and (1000 * failure) ms is subsequently added to the minimum delay. This is an aggressive backoff that also delivers some randomisation in order to avoid further rate-limiting as a result of retries being submitted in batches.

There are some limitations with this approach:
 - Each RPC provider has different recommendations for handling 429 responses.
 - If the 429 response includes a retry-after header, this new custom delay will be imposed in _addition_ to the retry-after delay. This is an unfortunate necessity that results from the [internal implementation of ethers](https://github.com/ethers-io/ethers.js/blob/44cbc7fa4e199c1d6113ceec3c5162f53def5bb8/packages/web/src.ts/index.ts#L275).
 - The algorithm has limited responsivity to en-masse rate-limiting. Ultimately, the operator must manually tune NODE_MAX_CONCURRENCY if they are seeing regular rate-limiting with > 1 retry.

It should be noted that the reason for not simply tuning the ethers throttleSlotInterval is that the resulting delay can still resolve to 0 ms as [a result of the use of parseInt()](https://github.com/ethers-io/ethers.js/blob/44cbc7fa4e199c1d6113ceec3c5162f53def5bb8/packages/web/src.ts/index.ts#L278).

A nice long-term solution would be to make the RateLimitProvider maxConcurrency automagically self-regulate downwards in response to 429 rate-limiting. This would allow for an more optimistic default NODE_MAX_CONCURRENCY.

This change configuration has been tested against a free-tier Alchemy plan with NODE_MAX_CONCURRENCY=5. I was regularly rate-limited with this approach, but was only rarely rate-limited more than once per request. This is OK per Alchemy's recommendations: they specify that it's OK for up to 30% of requests to be rate-limited.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -309,6 +309,7 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   // Default to a max concurrency of 1000 requests per node.
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
+<<<<<<< HEAD
   // Custom delay + logging for RPC rate-limiting.
   const rpcRateLimited =
     ({ nodeMaxConcurrency, logger }) =>
@@ -332,12 +333,14 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
       return attempt < retries;
     };
 
+  // See ethers ConnectionInfo for field descriptions.
+  // https://docs.ethers.org/v5/api/utils/web/#ConnectionInfo
   const constructorArgumentLists = getNodeUrlList(chainId).map((nodeUrl): [ethers.utils.ConnectionInfo, number] => [
     {
       url: nodeUrl,
       timeout,
       allowGzip: true,
-      throttleSlotInterval: 1,
+      throttleSlotInterval: 1, // Effectively disables ethers' internal backoff algorithm.
       throttleCallback: rpcRateLimited({ nodeMaxConcurrency, logger }),
     },
     chainId,
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -309,7 +309,6 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   // Default to a max concurrency of 1000 requests per node.
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
-<<<<<<< HEAD
   // Custom delay + logging for RPC rate-limiting.
   const rpcRateLimited =
     ({ nodeMaxConcurrency, logger }) =>
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -313,6 +313,8 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   const rpcRateLimited =
     ({ nodeMaxConcurrency, logger }) =>
     async (attempt: number, url: string): Promise<boolean> => {
+      // Implement a slightly aggressive expontential backoff to account for fierce paralellism.
+      // @todo: Start w/ maxConcurrency low and increase until 429 responses start arriving.
       const baseDelay = 1000 * Math.pow(2, attempt); // ms; attempt = [0, 1, 2, ...]
       const delayMs = baseDelay + baseDelay * Math.random();
 
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -276,10 +276,19 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   // Default to a max concurrency of 1000 requests per node.
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
-  // Temporarily added to expose RPC rate-limiting.
-  const rpcRateLimited =
-    ({ nodeMaxConcurrency, logger }) =>
+  const sleep = async (duration: number): Promise<number> => {
+    const startMs = Date.now();
+    await new Promise((resolve) => setTimeout(resolve, duration));
+    return Date.now() - startMs;
+  };
+
+  // Custom delay + logging for RPC rate-limiting.
+  const rpcRateLimited = ({ nodeMaxConcurrency, logger }) =>
     async (attempt: number, url: string): Promise<boolean> => {
+
+      const baseDelay = 1000 * Math.pow(2, attempt); // ms; attempt = [0, 1, 2, ...]
+      const delayMs = await sleep(baseDelay + baseDelay * Math.random());
+
       if (logger) {
         // Make an effort to filter out any api keys.
         const regex = url.match(/https?:\/\/([\w.-]+)\/.*/);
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -276,20 +276,31 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
...
+    await new Promise((resolve) => setTimeout(resolve, duration));
+    return Date.now() - startMs;
+  };
+
+  // Custom delay + logging for RPC rate-limiting.
   const rpcRateLimited =
     ({ nodeMaxConcurrency, logger }) =>
     async (attempt: number, url: string): Promise<boolean> => {
+      const baseDelay = 1000 * Math.pow(2, attempt); // ms; attempt = [0, 1, 2, ...]
+      const delayMs = await sleep(baseDelay + baseDelay * Math.random());
```
nicholaspai, line 290: Why randomize it?

src/utils/ProviderUtils.ts
```diff
@@ -276,20 +276,31 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   // Default to a max concurrency of 1000 requests per node.
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
-  // Temporarily added to expose RPC rate-limiting.
+  const sleep = async (duration: number): Promise<number> => {
+    const startMs = Date.now();
+    await new Promise((resolve) => setTimeout(resolve, duration));
```
nicholaspai, line 281: Why isn't calling `setTimeout` by itself sufficient?

src/utils/ProviderUtils.ts
```diff
@@ -298,6 +309,7 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
       url: nodeUrl,
       timeout,
       allowGzip: true,
+      throttleSlotInterval: 1,
```
nicholaspai, line 312: What does this do?

nicholaspai (commented):
Some questions on how this works but nice initiative!

src/utils/ProviderUtils.ts
pxrl: This overrides the ethers default of 100ms. Ethers uses this equation to compute its delay:

```const timeout = throttleSlotInterval * parseInt(String(Math.random() * Math.pow(2, attempt)));```

The problem is that this expression often resolves to 0 on the first attempt because Math.random() * Math.pow(2,0) should be 0.5 or less about half of the time, and parseInt() rounds down...so we end up immediately submitting another request without any delay. When trying to run with a free-tier provider and our concurrency setup, this just results in even more 429 responses and further compounds the problem.

So the effect of setting throttleSlotInterval to 1ms should be that the timeout computation always resolves to 0ms. This should be OK because we are inserting our own delay the throttleCallback function. If we get a retry-after response from the RPC provider then it does however mean that the bot will do a bit of extra waiting before retrying.

File: src/utils/ProviderUtils.ts (hunk 1 above)
Comment (pxrl) [reply src/utils/ProviderUtils.ts:290]: Randomising the delay is important because otherwise we'll end up with _n_ * workers trying to resubmit their requests all at approximately the same time. It would make the behaviour vastly more bursty and more subject to rate-limiting. By randomising the delay, we can even out the requests such that it should settle on a behavioural pattern that's less-likely to trigger further rate-limiting.

File: src/utils/ProviderUtils.ts (hunk 2 above)
Comment (pxrl) [reply src/utils/ProviderUtils.ts:281]: It seems to be a standard pattern. Apparently there are dragons if you do that, so wrapping it is often suggested: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#the_this_problem

I did however realise that [there's already a similar delay() function in ProviderUtils.ts](https://github.com/across-protocol/relayer-v2/blob/93b1d3d4fd0dbe974b2c83914ccdc7a8c8270ee9/src/utils/ProviderUtils.ts#L66), so I've updated the PR to use that.

File: src/utils/ProviderUtils.ts (hunk 3 above)
Comment (mrice32) [reply src/utils/ProviderUtils.ts:340]: I see. So the goal is just to make their delay deterministic so ours will take precedent?

File: src/utils/ProviderUtils.ts [hunk 4]
```diff
@@ -309,20 +309,26 @@ export function getProvider(chainId: number, logger?: winston.Logger) {
   // Default to a max concurrency of 1000 requests per node.
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
-  // Temporarily added to expose RPC rate-limiting.
+  // Custom delay + logging for RPC rate-limiting.
   const rpcRateLimited =
     ({ nodeMaxConcurrency, logger }) =>
     async (attempt: number, url: string): Promise<boolean> => {
+      const baseDelay = 1000 * Math.pow(2, attempt); // ms; attempt = [0, 1, 2, ...]
```
mrice32, line 316: I think this totally makes sense. If we still continue having issues with rate limiting, it could make sense to either a) tune concurrency algorithmically based on the global proportion of requests that are rate limited or b) tune the backoff based on the global proportion of requests that are rate limited. This is far more complex, so I don't think we should add such complex logic unless absolutely necessary.

mrice32 (approved):
Looks great!

src/utils/ProviderUtils.ts
pxrl: > I see. So the goal is just to make their delay deterministic so ours will take precedent?

Yep, that's correct. Unfortunately ethers doesn't permit it to be outright disabled, so the best I can manage is to set it to the lowest possible value (1ms).

In general I am not a fan of the way ethers implements this - it's quite inflexible and forces the user to resort to workarounds (like this one). I'd much prefer if there was more flexibility - including the ability to know whether retry-after was returned by the RPC provider.

src/utils/ProviderUtils.ts
pxrl: > it could make sense to either a) tune concurrency algorithmically based on the global proportion of requests that are rate limited

I'd previously thought that this was pretty difficult, but I think I was working within some artificial constraints. There seems to be a path forward on this, and in general I think it is the right approach. The user shouldn't have to manually tune how many concurrent workers they have - we should be able to scale up to the something sensible automagically (i.e. a bit like TCP window scaling).

I might chew on that in the background and see if I can produce anything sensible. It would be a win if we could remove the need for yet another config item, or could adjust it so that it was a "starting point" rather than a "max".

src/utils/ProviderUtils.ts
nicholaspai: gotcha makes sense!

nicholaspai: Actually @pxrl can you add a summary of the above into a code comment so we don't ask ourselves this question down the line?

src/utils/ProviderUtils.ts
nicholaspai: Gotcha +1 on `delay()`

src/utils/ProviderUtils.ts
nicholaspai: Can you add a comment about what this does? Also you should comment on what `allowGzip` does as wel

nicholaspai (approved):
This looks great. I have asked for comments in a few places so once those are added LGTM

src/utils/ProviderUtils.ts
pxrl: I've documented the intention behind setting throttleSlotInterval to 1, because that's not really intuitive without looking into the code. Separately, I referenced the ethers ConnectionInfo documentation so we don't need to duplicate it :+1:.

PR #406, opened 2023-01-10
improve: Dataworker should throw explicitly when it can't construct a slow fill excess

nicholaspai:
Right now there unknown behavior if dataworker can't find the first fill associated with a slow fill (i.e. the one that triggered the slow fill). We should loudly throw when this happens.

I think what would happen is that `getFillDataForSlowFillFromPreviousRootBundle` would throw: `Cannot read properties of undefined (reading 'blockNumber')`

I don't think this has caused any self-disputes but its possible?

### DIFF
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -118,6 +118,15 @@ export function getFillDataForSlowFillFromPreviousRootBundle(
     (_fill: FillWithBlock) => filledSameDeposit(_fill, fill) && isFirstFillForDeposit(_fill as Fill)
   );
 
+  // If there is no first fill for the same deposit, then throw an error.
+  if (firstFillForSameDeposit === undefined) {
+    // TODO: Use something similar to SpokePoolClient.queryHistoricalDepositForFill to look up all fills
+    // for the same deposit. I would add in PR#382 but its too complex so I'll leave for another PR.
+    throw new Error(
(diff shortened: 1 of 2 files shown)


src/utils/FillUtils.ts
```diff
@@ -118,6 +118,15 @@ export function getFillDataForSlowFillFromPreviousRootBundle(
     (_fill: FillWithBlock) => filledSameDeposit(_fill, fill) && isFirstFillForDeposit(_fill as Fill)
   );
 
+  // If there is no first fill for the same deposit, then throw an error.
+  if (firstFillForSameDeposit === undefined) {
+    // TODO: Use something similar to SpokePoolClient.queryHistoricalDepositForFill to look up all fills
+    // for the same deposit. I would add in PR#382 but its too complex so I'll leave for another PR.
+    throw new Error(
+      `FillUtils#getFillDataForSlowFillFromPreviousRootBundle: Cannot find first fill for for deposit ${fill.depositId} on chain ${fill.destinationChainId}, set a larger DATAWORKER_FAST_LOOKBACK_COUNT`
+    );
```
mrice32, line 127: +1 -- have we ever seen this issue in the wild? What would happen if we hit it (without this fix)? Would we just produce an incorrect bundle? Ignore the fill? Something else?

mrice32 (approved):
LGTM!

src/utils/FillUtils.ts
nicholaspai: Nope I think it would throw still but its good to explicitly throw

pxrl (approved):
Good to be explicit on this!

PR #407, opened 2023-01-10
improve(runTransaction): Permit caller to specify nonce

pxrl:
To be useful in a pending MultiCallerClient update...

Prettier makes this change look bigger than it really is.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -27,7 +27,8 @@ export async function runTransaction(
     });
     // TX config has gas (from gasPrice function), value (how much eth to send) and an optional gasLimit. The reduce
     // operation below deletes any null/undefined elements from this object. If the gasLimit is not specified, for example,
-    // then leave this up to ethers to compute.
+    // then leave this up to ethers to compute. Similarly, if nonce is not specified, 
+    // then ethers will query the RPC provider for the latest nonce
     const txConfig = Object.entries({ ...gas, value, nonce, gasLimit }).reduce(
       (a, [k, v]) => (v ? ((a[k] = v), a) : a),
       {}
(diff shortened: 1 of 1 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -10,15 +10,28 @@ export async function runTransaction(
...
+      args,
+      value,
+      nonce,
+      gas,
+    });
     // TX config has gas (from gasPrice function), value (how much eth to send) and an optional gasLimit. The reduce
     // operation below deletes any null/undefined elements from this object. If the gasLimit is not specified, for example,
     // then leave this up to ethers to compute.
-    const txConfig = Object.entries({ ...gas, value, gasLimit }).reduce((a, [k, v]) => (v ? ((a[k] = v), a) : a), {});
+    const txConfig = Object.entries({ ...gas, value, nonce, gasLimit }).reduce(
```
pxrl, line 31: Just for clarification, supplying nonce = null is the same as not specifying nonce at all. In this case, ethers will query the RPC provider for the latest nonce.

src/utils/TransactionUtils.ts
```diff
@@ -10,15 +10,28 @@ export async function runTransaction(
...
+      target: getTarget(contract.address),
+      method,
+      args,
+      value,
+      nonce,
+      gas,
+    });
     // TX config has gas (from gasPrice function), value (how much eth to send) and an optional gasLimit. The reduce
     // operation below deletes any null/undefined elements from this object. If the gasLimit is not specified, for example,
     // then leave this up to ethers to compute.
```
nicholaspai, line 30:
```suggestion
    // then leave this up to ethers to compute. Similarly, if nonce is not specified, 
    // then ethers will query the RPC provider for the latest nonce
```

nicholaspai (approved):
One comment suggestion https://github.com/across-protocol/relayer-v2/pull/407/files#r1065879761

LGTM

PR #409, opened 2023-01-10
fix(finalizer): Handle undefined error?.reason

nicholaspai:
IMG_3146

Handle this error so it doesn't go to PD error level

### DIFF
--- a/src/finalizer/utils/polygon.ts
+++ b/src/finalizer/utils/polygon.ts
@@ -105,7 +105,7 @@ export async function getFinalizableTransactions(
         });
         return { status: POLYGON_MESSAGE_STATUS.CAN_EXIT };
       } catch (err) {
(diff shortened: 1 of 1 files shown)


src/finalizer/utils/polygon.ts
```diff
@@ -105,7 +105,7 @@ export async function getFinalizableTransactions(
         });
         return { status: POLYGON_MESSAGE_STATUS.CAN_EXIT };
       } catch (err) {
-        if (err.reason.includes("EXIT_ALREADY_PROCESSED"))
```
pxrl, line 108: It seems that catching an exception with an undefined err would be more likely due to weird transient issues at the transport/infrastructure level. Suppressing them in this way could mean we go for an extended period of time without detecting an actual error.

Should we consider instead catching the case where err is undefined, and logging that with a low-severity message?

pxrl (approved):
:+1:

PR #414, opened 2023-01-11
improve: Import token addresses from sdk-v2

nicholaspai:


src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -14,35 +14,54 @@ import { SpokePoolClient } from "../../clients";
...
-  "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": "0xcEe284F754E854890e311e3280b767F80797180d", // USDC
-  "0xdAC17F958D2ee523a2206206994597C13D831ec7": "0xcEe284F754E854890e311e3280b767F80797180d", // USDT
-  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": "0xd92023E9d9911199a6711321D1277285e6d4e2db", // WETH
-  "0x6B175474E89094C44Da98b954EedeAC495271d0F": "0xD3B5b60020504bc3489D6949d545893982BA3011", // DAI
-  "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599": "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC", // WBTC
-  "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828": "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC", // UMA
-  "0x3472A5A71965499acd81997a54BBA8D852C6E53d": "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC", // BADGER
-  "0xba100000625a3754423978a60c9317c58a424e3D": "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC", // BAL
-  "0x44108f0223A3C3028F5Fe7AEC7f9bb2E66beF82F": "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC", // ACX
+  [sdkConstants.TOKEN_SYMBOLS_MAP.USDC.addresses[sdkConstants.CHAIN_IDs.MAINNET]]:
```
pxrl, line 21: sdkConstants.TOKEN_SYMBOL_MAP.<token>.addresses[sdkConstants.CHAIN_IDs.<network>] is quite a mouthful and it clutters the code quite a bit. Should we consider dereferecing?

```
const tokenMap = sdkConstants.TOKEN_SYMBOL_MAP;
const chainMap = sdkConstants.CHAIN_IDs;
```

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -14,35 +14,37 @@ import { SpokePoolClient } from "../../clients";
...
+  [TOKEN_SYMBOLS_MAP.DAI.addresses[CHAIN_IDs.MAINNET]]: "0x467194771dAe2967Aef3ECbEDD3Bf9a310C76C65", // DAI
+  [TOKEN_SYMBOLS_MAP.WBTC.addresses[CHAIN_IDs.MAINNET]]: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe", // WBTC
+  [TOKEN_SYMBOLS_MAP.UMA.addresses[CHAIN_IDs.MAINNET]]: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe", // UMA
+  [TOKEN_SYMBOLS_MAP.BADGER.addresses[CHAIN_IDs.MAINNET]]: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe", // BADGER
+  [TOKEN_SYMBOLS_MAP.BAL.addresses[CHAIN_IDs.MAINNET]]: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe", // BAL
+  [TOKEN_SYMBOLS_MAP.ACX.addresses[CHAIN_IDs.MAINNET]]: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe", // ACX
 } as const;
 
-type SupportedL1Token = keyof typeof l1Gateways & keyof typeof l2Gateways;
+type SupportedL1Token = string;
```
mrice32, line 47: If we change the SDK to export a more specific type from the SDK for TOKEN_SYMBOLS_MAP (export it `as const`), we don't have to drop the more specific types I don't think.

If the more specific types break things, we could also export a version of it that is casted to the less specific (string) types.

src/clients/AcrossAPIClient.ts
```diff
@@ -50,8 +53,12 @@ export class AcrossApiClient {
...
     const data = await Promise.all(
       tokensQuery.map((l1Token) => {
-        const validDestinationChainForL1Token = l2TokensToL1TokenValidation[l1Token]
-          ? Number(Object.keys(l2TokensToL1TokenValidation[l1Token])[0])
+        const validDestinationChainForL1Token = getL2TokenAddresses(l1Token)
+          ? Number(
+              Object.keys(getL2TokenAddresses(l1Token)).find(
+                (chainId) => Number(chainId) !== CHAIN_IDs.MAINNET && CHAIN_ID_LIST_INDICES.includes(Number(chainId))
+              )
+            )
```
mrice32, line 61: What is this logic change? Is this related to importing token addresses?

mrice32 (approved):
LGTM!

src/clients/AcrossAPIClient.ts
nicholaspai: yes so previously we filled in `validDestinationChainForL1Token` with the first destination chain in  the `l2TokensToL1TokenValidation` struct for the `l1Token`, which included all L2 chains. Now that we're using the TOKEN_MAPPING.addresses constant, which includes mainnet destination chains, I look up the first destination chain in `CHAIN_ID_LIST_INDICES` that is not mainnet

src/clients/bridges/ArbitrumAdapter.ts
mrice32: See docs on [const assertions in typescript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions). If we export TOKEN_SYMBOLS_MAP `as const` as described there, the type will be something like:

```
{
  USDC: {
    addresses: {
      1: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
      ...
}
```

Right now, I suspect the type is something like:
```
{ 
  [tokenName: string]: {
    addresses: {
      [chainId: number]: string;
    }
}
```

The former type is the narrowest form, but you can use it when you know a struct is const and the fields won't change. This means you can do more compile-time checking for things like whether the key is in the map or whether a string is in the list.

To be honest, it's not a huge deal, it just comes down to having typescript be more strict and infer more at compile time. If it's too much trouble to change it, we don't have to.

src/clients/AcrossAPIClient.ts
mrice32: I see. Makes sense!

src/clients/AcrossAPIClient.ts
```diff
@@ -50,8 +53,12 @@ export class AcrossApiClient {
     // when deciding whether a relay will be refunded.
     const data = await Promise.all(
       tokensQuery.map((l1Token) => {
-        const validDestinationChainForL1Token = l2TokensToL1TokenValidation[l1Token]
-          ? Number(Object.keys(l2TokensToL1TokenValidation[l1Token])[0])
+        const validDestinationChainForL1Token = getL2TokenAddresses(l1Token)
```
mrice32, line 56: nit: can we just put the result of getL2TokenAddresses into a local variable so we don't have to call it twice?

src/clients/AcrossAPIClient.ts
```diff
@@ -50,8 +53,12 @@ export class AcrossApiClient {
...
+          ? Number(
+              Object.keys(getL2TokenAddresses(l1Token)).find(
+                (chainId) => Number(chainId) !== CHAIN_IDs.MAINNET && CHAIN_ID_LIST_INDICES.includes(Number(chainId))
+              )
+            )
           : 10;
```
mrice32, line 62: Somewhat unrelated to this change, why do we need this default to Optimism? When would `getL2TokenAddresses` return null/undefined and should it be an error?

Note: we can address this in a separate PR, just wanted to raise the issue.

mrice32 (approved):
LGTM!

src/clients/bridges/ArbitrumAdapter.ts
nicholaspai: ah I see I do think that'd be a nice feature, PTAL: https://github.com/across-protocol/sdk-v2/pull/149

src/clients/AcrossAPIClient.ts
nicholaspai: hmm yeah now we don't need to default to Optimism anymore

PR #416, opened 2023-01-11
improve(MultiCallerClient): De-scope buldMultiCallBundle()

pxrl:
Simplify the code and make it more modular by returning an AugmentedTransaction. This permits it to be reused without imposing that the resulting multicall bundle is submitted.

A test to use this change (and fix a bug) will follow in the next commit.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -186,15 +186,7 @@ export class MultiCallerClient {
       // Construct multiCall transaction for each target chain.
       const multiCallTransactionsResult = await Promise.allSettled(
         Object.values(chunkedTransactions)
-          .map((transactions) =>
-            transactions.map((chunk) => {
-              try {
-                return this.submitTxn(this.buildMultiCallBundle(chunk));
-              } catch (err) {
-                return Promise.reject(err);
-              }
-            })
-          )
+          .map((transactions) => transactions.map((chunk) => this.submitTxn(this.buildMultiCallBundle(chunk))))
           .flat()
       );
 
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -186,7 +186,15 @@ export class MultiCallerClient {
       // Construct multiCall transaction for each target chain.
       const multiCallTransactionsResult = await Promise.allSettled(
         Object.values(chunkedTransactions)
-          .map((transactions) => transactions.map((chunk) => this.buildMultiCallBundle(chunk)))
+          .map((transactions) =>
+            transactions.map((chunk) => {
+              try {
+                return this.submitTxn(this.buildMultiCallBundle(chunk));
```
pxrl, line 192: nb. This seems like an effective no-op, but I will use buildMultiCallBundle() directly in a subsequent change.

nicholaspai (approved):
This makes sense and is more modular

src/clients/MultiCallerClient.ts
```diff
@@ -186,7 +186,15 @@ export class MultiCallerClient {
...
+              } catch (err) {
+                return Promise.reject(err);
```
mrice32, line 194: Why do we need this vs just making the function async and using await?

mrice32 (commented):
Looks great! Minor comment

src/clients/MultiCallerClient.ts
pxrl: Yeah, actually the whole construction was pretty superfluous...It collapses back into nothing: https://github.com/across-protocol/relayer-v2/pull/416/commits/817beee575cc1ab21c87bb66a31a46505639f08c

PR #417, opened 2023-01-12
fix(MultiCallerClient): Correct multicall txn validation

pxrl:
The existing implementation incorrectly checks that _each_ enqueued transaction has the wrong contract address. This is impossible because the contract address used as a reference is sourced from the first transaction in the queue.

While here, add a test for this and expand the logic to validate that all enqueued transactions are destined for the same chain ID as well..

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -261,12 +261,12 @@ export class MultiCallerClient {
   }
 
   buildMultiCallBundle(transactions: AugmentedTransaction[]): AugmentedTransaction {
-    // Validate all transactions in the batch have the same target contract.
-    const target = transactions[0].contract;
-    if (transactions.every((tx) => tx.contract.address !== target.address)) {
+    // Validate all transactions have the same chainId and target contract.
+    const { chainId, contract } = transactions[0];
+    if (transactions.some((tx) => tx.contract.address !== contract.address || tx.chainId !== chainId)) {
       this.logger.error({
-        at: "MultiCallerClient",
-        message: "some transactions in the bundle contain different targets",
+        at: "MultiCallerClient#buildMultiCallBundle",
+        message: "Some transactions in the queue contain different target chain or contract address",
         transactions: transactions.map(({ contract, chainId }) => {
           return { target: getTarget(contract.address), chainId };
         }),
@@ -279,13 +279,13 @@ export class MultiCallerClient {
     callData = [...new Set(callData)];
     this.logger.debug({
       at: "MultiCallerClient",
-      message: "Made bundle",
-      target: getTarget(target.address),
+      message: `Made multicall bundle for ${getNetworkName(chainId)}.`,
+      target: getTarget(contract.address),
       callData,
     });
 
     return {
-      contract: target,
+      contract,
       method: "multicall",
       args: [callData],
     } as AugmentedTransaction;
--- a/test/MultiCallerClient.ts
+++ b/test/MultiCallerClient.ts
@@ -120,7 +120,7 @@ describe("MultiCallerClient", async function () {
         multiCaller.enqueueTransaction({
           chainId: chainId,
           contract: {
-            address: "0x1234",
+            address,
             interface: { encodeFunctionData },
           },
           method: "test",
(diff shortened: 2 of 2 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -261,12 +261,12 @@ export class MultiCallerClient {
   }
 
   buildMultiCallBundle(transactions: AugmentedTransaction[]): AugmentedTransaction {
-    // Validate all transactions in the batch have the same target contract.
-    const target = transactions[0].contract;
-    if (transactions.every((tx) => tx.contract.address !== target.address)) {
+    // Validate all transactions have the same chainId and target contract.
+    const { chainId, contract } = transactions[0];
+    if (transactions.some((tx) => tx.contract.address !== contract.address || tx.chainId !== chainId)) {
```
nicholaspai, line 266: Maybe this works too? not sure if its more efficient though?
```suggestion
    if (transactions.slice(1).some((tx) => tx.contract.address !== contract.address || tx.chainId !== chainId)) {
```

test/MultiCallerClient.ts
```diff
@@ -108,4 +108,47 @@ describe("MultiCallerClient", async function () {
...
+      }
+
+      expect(txns.length).to.not.equal(0);
+      expect(() => multiCaller.buildMultiCallBundle(txns)).to.not.throw();
+
+      const badTxn: AugmentedTransaction = txns.pop();
+      switch (badField) {
+        case "address":
+          badTxn.contract = {
+            address: badTxn.contract.address.concat("badaddr"),
```
nicholaspai, line 140: its kind of weird that you can concat chars to this address to be greater than 20 chars:
- you should probbaly test with a correct length address
- should we enforce that the `contract.address` is actually a contract?

I definitely don't want to add overkill to our logic, since this is called so often and it seems fairly safe. But just something I thought of

src/clients/MultiCallerClient.ts
```diff
@@ -261,12 +261,12 @@ export class MultiCallerClient {
   }
 
   buildMultiCallBundle(transactions: AugmentedTransaction[]): AugmentedTransaction {
-    // Validate all transactions in the batch have the same target contract.
-    const target = transactions[0].contract;
-    if (transactions.every((tx) => tx.contract.address !== target.address)) {
```
nicholaspai, line 266: How did this even work before? It looks completely wrong, checking that each contract is wrong?

nicholaspai (commented):
Left some q's

src/clients/MultiCallerClient.ts
pxrl: It only throws an error if _every_ address is mismatched. So in practice I don't think we could have ever ended up throwing an exception here.

test/MultiCallerClient.ts
pxrl: > its kind of weird that you can concat chars to this address to be greater than 20 chars:

The address I'm testing with is "0x1234", so it's not even ~20~40 chars in length. In practice, the address isn't really used much in the MultiCallerClient, and external dependencies like willSucceed() and runTransaction() (which do examine the address) are stubbed off in the mocked implementation. So in practice, I don't think any code in this test environment is actually requiring anything special of the address string.
 
> * you should probbaly test with a correct length address

Yeah - I can see the argument for this, and it's kinda annoying looking at my fudged 0x1234 address. I'll update to define a constant address of standard length, and will then overwrite it with another address in order to exercise this test.

> * should we enforce that the `contract.address` is actually a contract?

We could do, although for us to have made it this far, we actually must have already completed a successful simulation. So from that we can probably infer that it is a real address.

> I definitely don't want to add overkill to our logic, since this is called so often and it seems fairly safe. But just something I thought of

Good inputs, and appreciate the feedback :+1:

src/clients/MultiCallerClient.ts
pxrl: Ah yeah - it would be nice to skip over the first element, since we know it will always be false in this test.

I checked on the documentation for Array.slice(). It returns an entirely new object with a shallow copy of a subset of the input array. I suspect that it would probably cost more than just wasting a loop on the first element, and it would also cost more cycles as the length of transactions increases (i.e. there are just additional references to create in the new array).

It does annoy me that we make a redundant check on the first element, though :-1:. It'd be really nice to just specify an offset into the array without creating an entirely new object...as far as I can tell, Javascript doesn't allow us to do that :(

test/MultiCallerClient.ts
nicholaspai: > We could do, although for us to have made it this far, we actually must have already completed a successful simulation for each transaction in the queue. So from that we can probably infer (...within reason) that it is a real contract address.

Good point, agreed

nicholaspai (approved):
LGTM

PR #418, opened 2023-01-12
improve(MultiCallerClient): Simplify sim failure logging

pxrl:
Makes the code a bit more compact and more clearly separates between simulation and logging. This also allows for the logging function to be verified in test by mocking it out.

As a minor benefit, the existing implementation performs 3 complete loops over the array of transaction simulation results. This change performs only one complete loop, and then a secondary loop over the list of failures to determine which to log. It should be marginally faster, and probably more memory efficient.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -310,7 +310,7 @@ export class MultiCallerClient {
       if (txn.succeed) validTxns.push(txn.transaction);
       else invalidTxns.push(txn);
     });
-    this.logSimulationFailures(invalidTxns);
+    if (invalidTxns.length > 0) this.logSimulationFailures(invalidTxns);
 
     return validTxns;
   }
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -315,10 +315,9 @@ export class MultiCallerClient {
     return validTxns;
   }
 
-  // Filter out transactions that revert for non-critical expected reasons. For example, the "relay filled" error will
+  // Filter out transactions that revert for non-critical, expected reasons. For example, the "relay filled" error may
   // will occur frequently if there are multiple relayers running at the same time. Similarly, the "already claimed"
-  // error will occur if there are overlapping dataworker executor runs. These are non critical errors we can ignore to
-  // filter out the noise.
+  // error will occur if there are overlapping dataworker executor runs.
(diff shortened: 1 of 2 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -45,8 +45,10 @@ const unknownRevertReasonMethodsToIgnore = new Set([
   "executeRootBundle",
 ]);
 
-// Ignore the general unknown revert reason for specific methods or uniformly ignore specific revert reasons
-// for any contract method.
+// Ignore the general unknown revert reason for specific methods or uniformly ignore specific revert reasons for any
+// contract method. Note: Check for exact revert reason instead of using .includes() to partially match reason string
+// in order to not ignore errors thrown by non-contract reverts. For example, a NodeJS error might result in a reason
+// string that includes more than just the contract revert reason.
 const canIgnoreRevertReasons = (obj: TransactionSimulationResult): boolean => {
```
pxrl, line 52: Thoughts on migrating this function down into logSimulationFailures() ? There might be an argument for that now that the split means both simulatieTransactionQueue() and logSimulationFailures() are quite small.

Alternatively, making canIgnoreRevertReasons() a protected member of MultiCallerClient would allow us to test that we are only suppressing the specific errors that we intend to.

nicholaspai: Sure I'm ok with that unless you want to unit test this logic specifically

nicholaspai (approved):
+1 this is a LOT cleaner

src/clients/MultiCallerClient.ts
pxrl: I migrated it so that we can call it from within the MockedMultiCallerClient and validate its behaviour :+1:.

test/MultiCallerClient.ts
```diff
@@ -106,6 +135,54 @@ describe("MultiCallerClient", async function () {
...
+
+      // Simulation succeeded but submission failed => multiCaller.simulationFailures should be empty.
+      expect(multiCaller.simulationFailureCount()).to.equal(0);
+    }
+  });
+
+  it("Correctly filters loggable vs. ignorable simulation failures", async function () {
+    const txn: AugmentedTransaction = {
+      chainId: chainIds[0],
+      contract: { address: "0x1234" },
```
pxrl, line 147: Address will be modified to a proper full-length address as part of another commit.

PR #419, opened 2023-01-12
improve(PriceClient): Add DefiLlama as a backstop

pxrl:
The recent CoinGecko blip led to a short period of relayer unavailability. DefiLlama have a price feed that is primarily backed by CoinGecko, but which can also fall back to on-chain sources for the major tokens.

Resolves ACX-597.


nicholaspai (approved):
This is awesome lgtm. Really like how the priceClient is build and constructed with fallbacks that are very easy to set

PR #420, opened 2023-01-12
improve: Set 90 day expiry on redis keys

nicholaspai:
RedisClient currently stores two keys:
- block for timestamp
- utilizations at block

These are used by ConfigStoreClient to compute a realized lp fee % and by dataworker to construct spoke pool clients' `fromBlocks`. They're also used by the validateRootBundle script which shouldnt ever be used to validate bundles older than 90 days

Expiring these keys after 90 days seems very safe assuming bundle block ranges stay steady

RedisJS [documentation](https://github.com/redis/node-redis#redis-commands)

Redis.API [documentation](https://redis.io/commands/set/)

### DIFF
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -8,6 +8,16 @@ export type RedisClient = ReturnType<typeof createClient>;
 // Current time must be >= 5 minutes past the event timestamp for it to be stable enough to cache.
 export const REDIS_CACHEABLE_AGE = 300;
 
+export async function setRedisKey(
+  key: string,
+  val: string,
+  expirySeconds: number,
+  redisClient: RedisClient
+): Promise<void> {
+  // EX: Expire key after expirySeconds.
+  await redisClient.set(key, val, { EX: expirySeconds });
+}
+
 // Get the block number for a given timestamp fresh from on-chain data if not found in redis cache.
 export async function getBlockForTimestamp(
   hubPoolChainId: number,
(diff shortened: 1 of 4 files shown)


src/utils/RedisUtils.ts
```diff
@@ -8,6 +8,16 @@ export type RedisClient = ReturnType<typeof createClient>;
 // Current time must be >= 5 minutes past the event timestamp for it to be stable enough to cache.
 export const REDIS_CACHEABLE_AGE = 300;
 
+export async function setRedisKey(
+  key: string,
+  val: string,
+  expirySeconds: number,
```
pxrl, line 14: Might be nice to have this default to 0, in which case the expiry is not set for this key?

src/utils/RedisUtils.ts
```diff
@@ -11,8 +11,8 @@ export const REDIS_CACHEABLE_AGE = 300;
 export async function setRedisKey(
   key: string,
   val: string,
-  expirySeconds: number,
-  redisClient: RedisClient
+  redisClient: RedisClient,
+  expirySeconds = 0,
 ): Promise<void> {
   // EX: Expire key after expirySeconds.
   await redisClient.set(key, val, { EX: expirySeconds });
```
pxrl, line 18: Sorry, I was unclear about the expirySeconds idea. EX = 0 is not a valid input, so I think the EX key would need to be conditional on expirySeconds being > 0.

Full disclosure on this input: If we had a generic redis wrapper/interface module then I think the ability to specify no expiry would be important, but our usage here is pretty specific...so I'm not sure whether this is ever going to be used in practice...

```
127.0.0.1:6379> SET testkey 5 EX 0
(error) ERR invalid expire time in 'set' command
```

nicholaspai: good point

pxrl (approved):
No further comments :+1:

PR #422, opened 2023-01-12
improve: Remove extra properties from cumulative l1TokenUpdate objects

nicholaspai:
Per the comment on line 221 "Drop value and key before passing args." we should be constructing cumulative l1 token config updates with `passedArgs` not `args` which includes the full key-value string which is unused after parsing it

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -94,11 +94,7 @@ export class Relayer {
     const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
 
     // Require that all fillable deposits meet the minimum specified number of confirmations.
-    const confirmedUnfilledDeposits = getUnfilledDeposits(
-      this.clients.spokePoolClients,
-      this.config.maxRelayerLookBack,
-      this.clients.configStoreClient
-    )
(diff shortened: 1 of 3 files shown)


src/relayer/Relayer.ts
```diff
@@ -94,11 +94,7 @@ export class Relayer {
     const latestHubPoolTime = this.clients.hubPoolClient.currentTime;
 
     // Require that all fillable deposits meet the minimum specified number of confirmations.
-    const confirmedUnfilledDeposits = getUnfilledDeposits(
-      this.clients.spokePoolClients,
-      this.config.maxRelayerLookBack,
-      this.clients.configStoreClient
-    )
+    const confirmedUnfilledDeposits = unfilledDeposits
```
nicholaspai, line 97: Removes duplicate `getUnfilledDeposits` call

pxrl: Yeah looks like we call it twice in this function - nice catch!

Is this finding related somehow to the other files? I'm trying to find a common thread but can't quite grasp it.

mrice32 (approved):
LGTM!

PR #423, opened 2023-01-13
improve(MultiCallerClient): Caller-supplied chunksize

pxrl:
Makes it possible to supply the chunkSize from within test cases. Also allows for multiple concurrent instances of the MultiCallerClient to use different chunk sizes, if that should ever be desirable.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -1,4 +1,4 @@
-import { DEFAULT_MULTICALL_CHUNK_SIZE, CHAIN_MULTICALL_CHUNK_SIZE } from "../common";
+import { DEFAULT_MULTICALL_CHUNK_SIZE, DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE } from "../common";
 import {
   winston,
   getNetworkName,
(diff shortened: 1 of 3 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -51,7 +51,10 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
 export class MultiCallerClient {
   private transactions: AugmentedTransaction[] = [];
   // eslint-disable-next-line no-useless-constructor
-  constructor(readonly logger: winston.Logger) {}
+  constructor(
+    readonly logger: winston.Logger,
+    readonly chunkSize: { [chainId: number]: number } = CHAIN_MULTICALL_CHUNK_SIZE
```
mrice32, line 56: Should we change the name of this env to DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE?

mrice32 (approved):
Looks good!

src/clients/MultiCallerClient.ts
pxrl: Yeah, good call. Done in https://github.com/across-protocol/relayer-v2/pull/423/commits/4509356db93f31d6708f8e35497c98bb0412f342.

PR #424, opened 2023-01-13
[2/3] feat(MultiCallerClient): Execute chain sequences in parallel

pxrl:
This PR introduces a new, more parallel-oriented implementation for the
MultiCallerClient. The work takes inspiration from the the pre-existing
implementation in executeTransactionQueue() (now 
executeTransactionQueueLegacy()). Once it has proven to be stable, legacy
mode and executeTransactionQueueLegacy() should be safe to remove.

Key differences to the existing implementation are:
  - Transaction sorting by value/non-value and chainId is performed in
    enqueueTransaction(). This eliminates the need for repetitive
    processing/filtering of the pending transactions at the time of
    execution (in executeTransactionQueue()).

  - Each chain is permitted to execute completely in parallel, and there
    are no intermediate synchronisation points until each chain has
    completed its run. This should reduce latency and improve robustness
    against poor performance or availability issues affecting adjacent
    RPC providers.

  - Transaction submission occurs only once per chain, rather than twice
    as in the existing implementation (value transactions first, then
     multicall transactions). Multicall bundles are now produced in
     advance of any transaction submission occurring, and the resulting
     multicall transactions are simply concatenated to any existing value
     transactions. This complete set is then handed of for submission to
     the RPC provider. This enables more compact, modular code.

  - The implementation is subdivided across multiple functions, enabling
    targeted testing of the key pieces of functionality.

Additional unlocks/benefits from this change:
  - Multi-chunk multicall transactions are properly supported - nonce
    issues begone!

  - Bypass multicall if we only have 1 candidate transaction in the
    queue. This saves ~3k gas (and improves the experience of browsing
    transactions in block explorers).

Fixes ACX-89, ACX-156, ACX-559.


pxrl:
See [here](https://github.com/across-protocol/relayer-v2/pull/425/files) for an example of the post-migration cleanup that would occur in the MultiCallerClient.

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,165 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+
+    // Collate the results for each chain.
+    const txnHashes: { [chainId: number]: string[] } = Object.fromEntries(
+      results.map((result, idx) => {
+        const chainId = chainIds[idx];
+        switch (result.status) {
+          case "fulfilled":
+            return [chainId, result.value.map((txnResponse) => txnResponse.hash)];
+          case "rejected":
+            return [chainId, result.reason];
```
pxrl, line 132: Seems like this should be: `return [chainId, [result.reason]]` in order to live up to the type annotation for txnHashes.

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,165 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+          default: // Should never occur; log if it ever does.
+            this.logger.warn({
+              at: "MultiCallerClient#executeTxnQueues",
+              message: "Unexpected transaction queue resulting status.",
+              result,
+            });
+            return [chainId, "Unknown error"];
```
pxrl, line 139: Same as above; the value should be an array of strings, not a single string.

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,165 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    );
+
+    // Flush out the currently enqueued transactions (for this chain only). Note that this is not async-safe; if the
+    // caller has enqueued additional transactions whilst awaiting executeTransactionQueue() then those will be lost.
+    this.clearTransactionQueue(chainId);
+
+    return txnResponses;
+  }
+
+  // @todo: Consider adding a "wait" argument to wait for n txn confirmations per txn.
```
pxrl, line 170: Add a comment here to describe what this function does (hint: it's a very "mechanical" function => simulate the pending txn queue, then submit the residual txns that passed simulation).

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,165 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+      });
+      this.logger.info({ at: "MulticallerClient#executeTxnQueue", message: "Exiting simulation mode 🎮" });
+      return [];
+    }
+
+    return await this.submitTxns(chainId, txns);
   }
 
-  async executeTransactionQueue(simulationModeOn = false): Promise<string[]> {
+  async executeTxnQueueLegacy(simulationModeOn = false): Promise<string[]> {
```
pxrl, line 210: Note: This is just renaming the pre-existing executeTransactionQueue() function in order to maintain the same functionality, but to provide existing callers with the same entry point into this class for commencing execution of the enqueued transactions. It should probably be marked as a private function, though...

src/clients/MultiCallerClient.ts
```diff
@@ -254,6 +391,54 @@ export class MultiCallerClient {
     return await runTransaction(this.logger, contract, method, args, value, null, nonce);
   }
 
+  private async submitTxns(chainId: number, txnQueue: AugmentedTransaction[]): Promise<TransactionResponse[]> {
```
pxrl, line 394: I think this method, and submitTxn(), could ultimately be moved out of this class and into a separate class dedicated to handling txn submission. They don't really have anything to do with multicall, per-se - I see multicall as a layer above the functionality that's handled here.

src/clients/MultiCallerClient.ts
```diff
@@ -279,16 +470,34 @@ export class MultiCallerClient {
...
+    chunkSize = DEFAULT_MULTICALL_CHUNK_SIZE
+  ): Promise<AugmentedTransaction[]> {
+    const txnChunks: AugmentedTransaction[][] = await lodash.chunk(
+      await this.simulateTransactionQueue(txns),
+      chunkSize
+    );
+
+    return txnChunks.map((txnChunk: AugmentedTransaction[]) => {
+      // Don't wrap single transactions in a multicall.
+      return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0];
```
pxrl, line 497: iirc. this saves ~3k gas when submitting a single transaction.

pxrl:
Based on some test failures in [this](https://github.com/across-protocol/relayer-v2/actions/runs/3916020881/jobs/6694975023) PR, it looks like we're not _quite_ ready for the big dance just yet.

pxrl:
> Based on some test failures in [this](https://github.com/across-protocol/relayer-v2/actions/runs/3916020881/jobs/6694975023) PR, it looks like we're not _quite_ ready for the big dance just yet.

Just to follow up on this, tests pass now when I run them locally. They're failing with OoM errors in CI, though: https://github.com/across-protocol/relayer-v2/actions/runs/3924751056/jobs/6709217940.

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+
+    // Collate the results for each chain.
+    const txnHashes: { [chainId: number]: string[] } = Object.fromEntries(
+      results.map((result, idx) => {
+        const chainId = chainIds[idx];
+        switch (result.status) {
+          case "fulfilled":
+            return [chainId, result.value.map((txnResponse) => txnResponse.hash)];
+          case "rejected":
+            return [chainId, result.reason];
```
mrice32, line 132: nit: might be simpler to use an if/else based on [this function](https://github.com/across-protocol/relayer-v2/blob/e762204a3a275e6cacc0d5a252f6709151a234af/src/utils/TypeGuards.ts#L1-L5). Typescript should infer the correct types in the if/else blocks and you should be able to drop the `default`.

src/clients/MultiCallerClient.ts
mrice32: Why are you returning a string? Why not an array? Does it fail to compile or something?

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+        })
+        .map((_chainId) => {
+          const chainId = Number(_chainId);
+          return this.executeChainTxnQueue(chainId, simulate);
+        })
+    );
+
+    // Collate the results for each chain.
+    const txnHashes: { [chainId: number]: string[] } = Object.fromEntries(
+      results.map((result, idx) => {
```
mrice32, line 126: If you're having issues with the types in this function, it could be helpful to provide a type annotation for the return type.

src/clients/MultiCallerClient.ts
```diff
@@ -49,28 +49,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    return txnHashes;
   }
 
-  clearTransactionQueue() {
-    this.transactions = [];
+  // For a single chain, simulate all potential multicall txns and group the ones that pass into multicall bundles.
+  // Then, submit a concatenated list of value txns + multicall bundles. Flush the existing queues on completion.
+  async executeChainTxnQueue(chainId: number, simulate = false): Promise<TransactionResponse[]> {
+    const multicallTxns: AugmentedTransaction[] =
+      this.txns[chainId].length > 0
```
mrice32, line 151: Should we make sure this isn't undefined?
```suggestion
      this.txns[chainId]?.length > 0
```

mrice32 (commented):
Looks great! A few minor comments. Great simplification!

I also want to dig a bit deeper on the txn execution/nonce logic if it's not identical to what we had before. Parallelism can be really tricky wrt stability and resilience.

src/clients/MultiCallerClient.ts
pxrl: It's wrong :) I earmarked it to be fixed here as well: https://github.com/across-protocol/relayer-v2/pull/424#discussion_r1070180575.

src/clients/MultiCallerClient.ts
pxrl: > Since this is a big refactor, any reason not to go ahead and move this out?

Yeah, I definitely think it's the right call for it to be moved out. I wasn't sure if others would have the same appetite for that as me, but I'll go ahead and get it done as a precursor to this commit.

> Also, is this all new code? Or is this copied? I didn't review it super closely because I assumed it was copied, but will dig deeper if not.

It's all new code. Parts of it are inspired by the existing executeTransactionQueue() function - mostly to try to maintain consistency with the existing logging behaviour. There _are_ some differences in logging, though.

pxrl:
Note: I've proposed some new PRs to factor the transaction-specific parts out into a new class, creatively named "TransactionClient": https://github.com/across-protocol/relayer-v2/pull/426 https://github.com/across-protocol/relayer-v2/pull/427

I think it makes sense to revisit this PR once those are resolved.

src/clients/MultiCallerClient.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/153d3888a837b88edd3927a3d7bd4bdc2762fd0b https://github.com/across-protocol/relayer-v2/pull/424/commits/c968ab38c108d6ee668982c81a29958cddf7b571

src/clients/MultiCallerClient.ts
pxrl: > If you're having issues with the types in this function, it could be helpful to provide a type annotation for the return type.

I haven't had issues per se; if anything I wonder whether TypeScript might be _missing_ something, because it seemed surprisingly non-strict here (and obviously missed the mistakes on some of the return statements.

I've got a type annotation on the method; likewise on the declaration of txnHashes. Where else could I put one to nudge TS in the right direction?

src/clients/MultiCallerClient.ts
pxrl: Nice catch; fixed here: https://github.com/across-protocol/relayer-v2/pull/424/commits/98a66acb394fde49caa7e71ced54237fcbabef87

test/MultiCallerClient.ts
```diff
@@ -29,55 +29,18 @@ class MockedMultiCallerClient extends MultiCallerClient {
...
-      hash: "0x4321",
-    } as TransactionResponse;
-
-    this.logger.debug({
-      at: "MockMultiCallerClient#submitTxns",
-      message: "Transaction submission succeeded!",
-      txn: txnResponse,
-    });
-
-    return txnResponse;
```
pxrl, line 48: This was relocated to the TransactionClient class. The MockedMultiCallerClient class instantiates a MockedTransactionClient that implements this feature.

test/MultiCallerClient.ts
```diff
@@ -29,55 +29,18 @@ class MockedMultiCallerClient extends MultiCallerClient {
...
-    this.logger.debug({
-      at: "MockMultiCallerClient#simulateTxn",
-      message: `Forcing simulation ${this.failSimulate ? "failure" : "success"}.`,
-      txn,
-    });
-    return {
-      transaction: txn,
-      succeed: this.failSimulate === "",
-      reason: this.failSimulate ?? null,
-    };
```
pxrl, line 61: Same as above; relocated to TransactionClient and inherited via the MultiCallerClient txnClient instance.

test/MultiCallerClient.ts
```diff
@@ -29,55 +29,18 @@ class MockedMultiCallerClient extends MultiCallerClient {
...
-      at: "MockMultiCallerClient#submitTxn",
-      message: "Transaction submission succeeded!",
-      txn,
-    });
-
-    return {
-      chainId: txn.chainId,
-      nonce: nonce || 1,
-      hash: "0x4321",
-    } as TransactionResponse;
```
pxrl, line 80: Relocated; same as above.

pxrl:
Note that this PR is currently pretty complex, so I've split part of it out here: https://github.com/across-protocol/relayer-v2/pull/431

Once that is merged, this one should become much more manageable.

src/clients/MultiCallerClient.ts
pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/pull/424/commits/78a6ed3192346176563b575ce58bd12d8d429412

src/clients/MultiCallerClient.ts
pxrl: Addressed here: https://github.com/across-protocol/relayer-v2/pull/424/commits/cf8753d64278f50cd8bda3af32dc0b9d9017d8c9

src/clients/MultiCallerClient.ts
pxrl: Has been resolved as part of the promise handling change you proposed: https://github.com/across-protocol/relayer-v2/pull/424/commits/153d3888a837b88edd3927a3d7bd4bdc2762fd0b.

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+      const txnQueue = txn.value && txn.value.gt(0) ? this.valueTxns : this.txns;
+      if (txnQueue[txn.chainId] === undefined) txnQueue[txn.chainId] = [];
+      txnQueue[txn.chainId].push(txn);
+    } else this.transactions.push(txn);
   }
 
   transactionCount() {
-    return this.transactions.length;
+    let nTxns = this.transactions.length;
+    Object.values(this.txns).forEach((txnQueue) => (nTxns += txnQueue.length));
```
nicholaspai, line 71: this could be a one liner if you use `.reduce`: 

```ts
const nTxns = Object.values(this.txns.concat(this.valueTxns)).reduce((count, txnQueue) => count += txnQueue.length, 0)
```

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    Object.values(this.valueTxns).forEach((txnQueue) => (nTxns += txnQueue.length));
+    return nTxns;
   }
 
-  clearTransactionQueue() {
-    this.transactions = [];
+  clearTransactionQueue(chainId: number = null) {
+    if (this.newMulticaller) {
+      if (chainId) {
```
nicholaspai, line 78:
```suggestion
      if (chainId !== null) {
```

to handle chain ID 0

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+      return Array.from(Object.values(txnHashes)).flat();
+    }
+
+    return this.executeTxnQueueLegacy(simulate);
+  }
+
+  // For each chain, collate the enqueued transactions and process them in parallel.
+  async executeTxnQueues(simulate = false): Promise<{ [chainId: number]: string[] }> {
+    // @todo: This feels like a bodge...Find a better way to produce the set of chainIds that have enqueued txns.
+    const chainIds = [...new Set(Object.keys(this.valueTxns).concat(Object.keys(this.txns)))];
```
nicholaspai, line 102: this is pretty elegant IMO

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    );
+
+    // Collate the results for each chain.
+    const txnHashes: { [chainId: number]: string[] } = Object.fromEntries(
+      results.map((result, idx) => {
+        const chainId = chainIds[idx];
+        if (isPromiseFulfilled(result)) return [chainId, result.value.map((txnResponse) => txnResponse.hash)];
+        else if (isPromiseRejected(result)) return [chainId, [result.reason]];
+
+        this.logger.warn({
```
nicholaspai, line 124: how could you get to this branch?

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    if (simulate) {
+      let mrkdwn = "";
+      results.forEach((result, idx) => {
+        const { succeed, transaction: txn } = result;
+        mrkdwn += `  *${idx + 1}. (${succeed} ? "Passed" : "Failed"})`;
+        mrkdwn += ` ${txn.message || "No message"}: ${txn.mrkdwn || "No markdown"}\n`;
+      });
+      this.logger.info({
+        at: "MultiCallerClient#executeTxnQueue",
+        message: `${txns.length}/${results.length} ${networkName} transaction simulation(s) succeeded!`,
```
nicholaspai, line 186: nit: consider divide by 0 errors here

nicholaspai (commented):
This all makes sense. And the follow up PR would remove legacy code and the `newMulticall` flag right? That flag and associated logic makes this client a bit messy.

This is great code overall!

src/clients/MultiCallerClient.ts
pxrl: It would have to be in "pending" state, which I admit _should_ not be possible because allSettled should enforce that all promises are resolved.

I've been conditioned to be skeptical of programmers who say "_x_ is impossible", because circumstance can still find ways to make it possible anyway. In this case I think it's safe to assume that we can never arrive here and just remove the code.

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +42,167 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    // caller has enqueued additional transactions whilst awaiting executeTransactionQueue() then those will be lost.
+    this.clearTransactionQueue(chainId);
+
+    return txnResponses;
   }
 
-  async executeTransactionQueue(simulationModeOn = false): Promise<string[]> {
+  // Simulate the entire transaction queue and hand the ones that pass off to the TransactionClient for RPC submission.
+  // When simulation is requested, the simulation results are logged and the function returns early.
+  // @todo: Consider adding a "wait" argument to wait for n txn confirmations per txn.
```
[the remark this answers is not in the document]
pxrl: My understanding at the moment is that there is no waiting, so we only collect the transaction hashes and log them. Because we're applying market rates for gas we usually get confirmed quickly (EIP-1559 helps), but I'm pretty confident that there would be a window where the txn is sitting in the mempool, yet we've logged that it was successful because we don't wait for confirmation.

src/clients/MultiCallerClient.ts
pxrl: It's not an actual division - it's only a log message, so it's fine to print x/0. But you've caught me red handed on this before when there was actual division happening :)

pxrl:
> This all makes sense. And the follow up PR would remove legacy code and the `newMulticall` flag right? That flag and associated logic makes this client a bit messy.

Correct. The retention of the existing code and the "new.." flag are only there to support a progressive migration in production. I've got a draft PR to prepare for removal of the redundant code, once we've migrated off it: https://github.com/across-protocol/relayer-v2/pull/425

There is an issue over there because the tests consistently fail with OoM errors. I'm a bit surprised by that; I don't think I'm doing anything especially memory-inefficient here, and try to scope all my objects tightly so they have short lifespans. It might be related to the way we use the MultiCallerClient in test, though. 


> This is great code overall!

:pray:

src/clients/MultiCallerClient.ts
pxrl: Tidied it up here: https://github.com/across-protocol/relayer-v2/pull/424/commits/553fbfd056ed784dc096f29c42e70f7e881fa901.

src/clients/MultiCallerClient.ts
pxrl: Nice - have implemented that here: https://github.com/across-protocol/relayer-v2/pull/424/commits/6627403e545aafd4502cbf1923c10e31529c8d88.

src/clients/MultiCallerClient.ts
pxrl: Seems unlike that we'd ever have a chainId of 0, but it'd also be weird if a caller passed in a chainId of 0 and we flushed everything. Addressed here: https://github.com/across-protocol/relayer-v2/pull/424/commits/c6d85c4b792e8179925dc27548f2118363e676e7

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+  protected txns: { [chainId: number]: AugmentedTransaction[] } = {};
+  protected valueTxns: { [chainId: number]: AugmentedTransaction[] } = {};
+  // Legacy mode is a temporary feature to support transition to the updated multicaller implementation.
   constructor(
     readonly logger: winston.Logger,
-    readonly chunkSize: { [chainId: number]: number } = DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE
+    readonly chunkSize: { [chainId: number]: number } = DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE,
+    newMulticaller = false
   ) {
+    this.newMulticaller = newMulticaller || process.env.NEW_MULTICALLER === "true";
```
mrice32, line 55: nit: why do we need two ways to enable this?

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    // @todo: This feels like a bodge...Find a better way to produce the set of chainIds that have enqueued txns.
+    const chainIds = [...new Set(Object.keys(this.valueTxns).concat(Object.keys(this.txns)))];
+
+    // One promise per chain for parallel execution.
+    const results = await Promise.allSettled(
+      chainIds
+        .filter((_chainId) => {
+          const chainId = Number(_chainId);
+          return this.txns[chainId]?.length > 0 || this.valueTxns[chainId]?.length > 0;
+        })
```
mrice32, line 113: We're checking the existence and length of these in a bunch of places. Can we simplify and drop some of these branches? Checking at each use can feel like it gives more assurance of correctness, but (IMO) it may end up leading to confusion about where the code is expected to branch off -- prefer removing the uncertainty about existence as early as possible and then have the type system guarantee it the rest of the way.

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    );
+
+    return txnHashes;
+  }
+
+  // For a single chain, simulate all potential multicall txns and group the ones that pass into multicall bundles.
+  // Then, submit a concatenated list of value txns + multicall bundles. Flush the existing queues on completion.
+  async executeChainTxnQueue(chainId: number, simulate = false): Promise<TransactionResponse[]> {
+    const multicallTxns: AugmentedTransaction[] =
+      this.txns[chainId]?.length > 0
```
mrice32, line 137: If you pass in these transaction arrays, you can avoid the undefined check because by the time you make it here, the type system can guarantee the array is not undefined.

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+        ? await this.buildMultiCallBundles(this.txns[chainId], this.chunkSize[chainId])
```
mrice32, line 138: nit: you don't need to protect this function from empty arrays. lodash should deal gracefully with empty inputs. Also, this function does internal filtering so you could end up with empty arrays anyway!

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+        : [];
+
+    // Concatenate the new multicall txns onto any existing value txns and pass the queue off for submission.
+    const txnResponses: TransactionResponse[] = await this.executeTxnQueue(
+      chainId,
+      (this.valueTxns[chainId] ?? []).concat(multicallTxns),
```
mrice32, line 144: Same here. If you pass in the value transaction array, you can avoid the check.

src/clients/MultiCallerClient.ts
```diff
@@ -40,29 +43,164 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+      simulate
+    );
+
+    // Flush out the currently enqueued transactions (for this chain only). Note that this is not async-safe; if the
+    // caller has enqueued additional transactions whilst awaiting executeTransactionQueue() then those will be lost.
+    this.clearTransactionQueue(chainId);
```
mrice32, line 150: This is a good point. Why not move these arrays into a local variables at the top of executeTxnQueues and clear them at that point to avoid sync issues?

src/clients/MultiCallerClient.ts
```diff
@@ -271,14 +416,29 @@ export class MultiCallerClient {
...
   }
 
-  protected async simulateTxn(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
-    return await willSucceed(txn);
+  async buildMultiCallBundles(
+    txns: AugmentedTransaction[],
+    chunkSize = DEFAULT_MULTICALL_CHUNK_SIZE
+  ): Promise<AugmentedTransaction[]> {
+    // Enqueued multicall txns that fail simulation are silently dropped.
+    const txnChunks: AugmentedTransaction[][] = await lodash.chunk(
```
mrice32, line 432: Why is there an await in front of lodash.chunk? Is lodash.chunk an async function?

mrice32 (commented):
Looks great! A few comments

src/clients/MultiCallerClient.ts
pxrl: :eyes: That's a good question...shouldn't be there at all.

src/clients/MultiCallerClient.ts
pxrl: It's intended to give us some additional flexibility in choosing how to test this. My thinking was that it's potentially nice to enable it in the env initially, and then make the change in code once we're OK with it.

The scenario that I'd really like to avoid is that it breaks in prod during initial testing, and then we have to revert the change and redeploy, or roll back our deployed version (which may not be feasible if subsequent, unrelated changes have been introduced). I think env-based configuration has some advantages there.

I'm not anchored to this approach and can see it has some downsides - i.e. it might be confusing to debug if only looking at the code, and not considering the environment. Happy to consider alternative approaches!

src/clients/MultiCallerClient.ts
pxrl: Really nice suggestion. Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/44a3a8e16c249197d38eefc3bee090d9ee3a19c0.

src/clients/MultiCallerClient.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/44a3a8e16c249197d38eefc3bee090d9ee3a19c0.

src/clients/MultiCallerClient.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/44a3a8e16c249197d38eefc3bee090d9ee3a19c0.

src/clients/MultiCallerClient.ts
pxrl: ++agree on this. It was clumsy. Have implemented your suggestion here: https://github.com/across-protocol/relayer-v2/pull/424/commits/44a3a8e16c249197d38eefc3bee090d9ee3a19c0.

src/clients/MultiCallerClient.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/424/commits/44a3a8e16c249197d38eefc3bee090d9ee3a19c0.

src/clients/MultiCallerClient.ts
pxrl: Fixed: https://github.com/across-protocol/relayer-v2/pull/424/commits/fc97098d93fd8880e839c615d7c81adaa1d24692.

pxrl:
Update: After making the adjustments per Matt's comments, I dug a bit further and found I could remove a lot of the new code. The implementation should now be simpler and more concise.

src/clients/MultiCallerClient.ts
mrice32: Why not enable it just with env and not have the constructor variable at all? Would that compromise anything we were trying to do?

src/clients/MultiCallerClient.ts
```diff
@@ -41,28 +44,151 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+  async executeTxnQueues(simulate = false): Promise<{ [chainId: number]: string[] }> {
+    const chainIds = [...new Set(Object.keys(this.valueTxns).concat(Object.keys(this.txns)))];
+
+    // One promise per chain for parallel execution.
+    const results = await Promise.allSettled(
+      chainIds
+        .filter((_chainId) => {
+          const chainId = Number(_chainId);
+          return this.txns[chainId]?.length > 0 || this.valueTxns[chainId]?.length > 0;
+        })
```
mrice32, line 112: nit: why do we need this filter at all? The map below still has to deal with one of these potentially being undefined or empty.

src/clients/MultiCallerClient.ts
```diff
@@ -41,28 +44,151 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+        .map((_chainId) => {
+          const chainId = Number(_chainId);
+          const txns: AugmentedTransaction[] = this.txns[chainId];
+          const valueTxns: AugmentedTransaction[] = this.valueTxns[chainId];
```
mrice32, line 116: Typescript will let you get away without the undefined, but I think the undefined is more correct if we're trying to be explicit about the types.

```suggestion
          const txns: AugmentedTransaction[] | undefined = this.txns[chainId];
          const valueTxns: AugmentedTransaction[] | undefined = this.valueTxns[chainId];
```

src/clients/MultiCallerClient.ts
```diff
@@ -41,28 +44,151 @@ export const unknownRevertReasonMethodsToIgnore = new Set([
...
+    const txnSims = await Promise.allSettled([
+      this.simulateTransactionQueue(txns),
+      this.simulateTransactionQueue(valueTxns),
+    ]);
+
+    let _txns: AugmentedTransaction[];
+    let _valueTxns: AugmentedTransaction[];
+    [_txns, _valueTxns] = txnSims.map((result) => {
+      return isPromiseFulfilled(result) ? result.value : ([] as AugmentedTransaction[]);
+    });
```
mrice32, line 162: nit: would this work?
```suggestion
    const [_txns, _valueTxns] = txnSims.map((result): AugmentedTransaction[]  => {
      return isPromiseFulfilled(result) ? result.value : [];
    });
```

mrice32 (approved):
LGTM! A few pretty minor comments

src/clients/MultiCallerClient.ts
pxrl: Fair point - simplified here: https://github.com/across-protocol/relayer-v2/pull/424/commits/5add4e6f5e078d1e569f1314a5e6da4c200fa57d.

src/clients/MultiCallerClient.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/0b8a3d0e4e7d22581d11848560575478c0a5620b.

src/clients/MultiCallerClient.ts
pxrl: Nice suggestion - much more concise! Implemented here: https://github.com/across-protocol/relayer-v2/pull/424/commits/d2c7ed520e057c5f8bc23523af1bbb7bc7c98ca4.

PR #425, opened 2023-01-14
chore(MultiCallerClient):  Remove legacy implementation

pxrl:
The new MultiCallerClient implementation has been trialed in production over the
past ~week, in both the executor and relayer bots. The relayer bots in particular have
been a good stress test, since we have overlapping instances that will tend to race
against each other.

One minor error was corrected relating to dropping the transaction message for
transactions that are bundled into a multicall transaction. This only impacted logging,
and I otherwise have not observed any adverse effects: everything seems stable.

I think it's OK to remove the old implementation and defacto-migrate all users over
to the new implementation.


test/Relayer.BasicFill.ts
```diff
@@ -116,7 +116,6 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
     expect(multiCallerClient.transactionCount()).to.equal(1); // One transaction, filling the one deposit.
 
     const tx = await multiCallerClient.executeTransactionQueue();
-    expect(lastSpyLogIncludes(spy, "Multicall batch sent")).to.be.true;
```
pxrl, line 119: Note: Have removed these log messages because the upper-layer relayer logic shouldn't really know (or care) whether a relay is submitted as a multicall batch or not.

pxrl:
Have observed that the tests are consistently failing with OoM errors :( Need to investigate that.

mrice32 (approved):
LGTM!

PR #426, opened 2023-01-17
feat: Add TransactionClient

pxrl:
This factors out some of the transaction submission functions from the MultiCallerClient implementation. As a result, the MultiCallerClient can become more compact and focused on the task of sorting transactions and bundling the eligible ones.

Note that this PR does not introduce use of the TransactionClient into the MultiCallerClient at this stage, since the TransactionClient includes new behaviour around nonces.

More to come.

### DIFF
--- a/src/clients/TransactionClient.ts
+++ b/src/clients/TransactionClient.ts
@@ -1,0 +1,78 @@
+import {
+  winston,
+  getNetworkName,
+  Contract,
+  runTransaction,
+  BigNumber,
+  etherscanLink,
+  TransactionResponse,
+  TransactionSimulationResult,
+} from "../utils";
+
+export interface AugmentedTransaction {
+  contract: Contract;
+  chainId: number;
+  method: string;
+  args: any[];
+  message: string;
+  mrkdwn: string;
+  value?: BigNumber;
+}
+
+export class TransactionClient {
+  // eslint-disable-next-line no-useless-constructor
+  constructor(readonly logger: winston.Logger) {}
+
+  protected async _submit(txn: AugmentedTransaction, nonce: number | null = null): Promise<TransactionResponse> {
+    const { contract, method, args, value } = txn;
+    return await runTransaction(this.logger, contract, method, args, value, null, nonce);
+  }
+
+  async submit(chainId: number, txns: AugmentedTransaction[]): Promise<TransactionResponse[]> {
+    const networkName = getNetworkName(chainId);
+    const txnResponses: TransactionResponse[] = [];
+
+    this.logger.debug({
+      at: "TransactionClient#submit",
+      message: `Processing ${txns.length} transactions.`,
+      txns,
+    });
+
+    // Transactions are submitted sequentially to avoid nonce collisions. More
+    // advanced nonce management may permit them to be submitted in parallel.
+    let mrkdwn = "";
+    let idx = 1;
+    let nonce: number = null;
+    for (const txn of txns) {
+      let response: TransactionResponse;
+      if (nonce !== null) this.logger.debug({ at: "TransactionClient#submit", message: `Using nonce ${nonce}.` });
+
+      try {
+        response = await this._submit(txn, nonce);
+      } catch (error) {
+        this.logger.info({
+          at: "TransactionClient#submit",
+          message: `Transaction ${idx} submission on ${networkName} failed or timed out.`,
+          mrkdwn,
+          error,
+          notificationPath: "across-error",
+        });
+        return txnResponses;
+      }
+
+      nonce = response.nonce + 1;
+      mrkdwn += `  ${idx}. ${txn.message || "No message"}: ${txn.mrkdwn || "No markdown"}\n`;
+      mrkdwn += `  *Block Explorer:* ${etherscanLink(response.hash, txn.chainId)}\n`;
+      txnResponses.push(response);
+      ++idx;
+    }
+
+    this.logger.info({
+      at: "TransactionClient#submit",
+      message: `Completed ${networkName} transaction submission! 🧙`,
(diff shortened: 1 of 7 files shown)


src/clients/TransactionClient.ts
```diff
@@ -0,0 +1,78 @@
...
+      try {
+        response = await this._submit(txn, nonce);
+      } catch (error) {
+        this.logger.info({
+          at: "TransactionClient#submit",
+          message: `Transaction ${idx} submission on ${networkName} failed or timed out.`,
+          mrkdwn,
+          error,
+          notificationPath: "across-error",
+        });
```
pxrl, line 59: This method handles some informative logging because it's been lifted as-is from the existing MultiCallerClient PR. I am however uncertain about we should be doing informative logging here, since this code is increasingly low-level in our stack. Maybe it should be the responsibility of the caller - perhaps we can consider that in future.

src/clients/TransactionClient.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export class TransactionClient {
+  // eslint-disable-next-line no-useless-constructor
+  constructor(readonly logger: winston.Logger) {}
+
+  protected async _submit(txn: AugmentedTransaction, nonce: number | null = null): Promise<TransactionResponse> {
+    const { contract, method, args, value } = txn;
+    return await runTransaction(this.logger, contract, method, args, value, null, nonce);
+  }
```
pxrl, line 29: Note: This pattern of wrapping the "external" dependency in a small private method is intended to make it easier to mock the method out in test, since this is typically where I want to simulate some success of failure.

Happy to take pointers if there are better ways of achieving this.

src/clients/TransactionClient.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  winston,
+  getNetworkName,
+  Contract,
+  runTransaction,
```
pxrl, line 5: External dependencies like runTransaction() and willSucceed() would probably me candidates for moving directly _into_ this class. They are currently used quite a bit in our code though, so migrating them would involve updating a few files to use the TransactionClient instead.

src/clients/TransactionClient.ts
```diff
@@ -0,0 +1,78 @@
...
+          at: "TransactionClient#submit",
+          message: `Transaction ${idx} submission on ${networkName} failed or timed out.`,
+          mrkdwn,
+          error,
+          notificationPath: "across-error",
+        });
+        return txnResponses;
+      }
+
+      nonce = response.nonce + 1;
```
pxrl, line 63: Relating specifically to nonce handling, this comment might be relevant: https://github.com/across-protocol/relayer-v2/pull/424#issuecomment-1384297917

src/clients/TransactionClient.ts
nicholaspai: I agree, we could make them static functions maybe?

src/clients/TransactionClient.ts
nicholaspai: I think this is a good way to test it!

src/clients/TransactionClient.ts
```diff
@@ -0,0 +1,78 @@
...
+      mrkdwn += `  ${idx}. ${txn.message || "No message"}: ${txn.mrkdwn || "No markdown"}\n`;
+      mrkdwn += `  *Block Explorer:* ${etherscanLink(response.hash, txn.chainId)}\n`;
+      txnResponses.push(response);
+      ++idx;
```
nicholaspai, line 67: nit: Since you're using `idx` like a counter why not use the traditional `for(let idx = 0; idx < txns.length; idx++)` syntax?

test/MultiCallerClient.ts
```diff
@@ -29,6 +29,25 @@ class MockedMultiCallerClient extends MultiCallerClient {
...
   }
 
+  protected async _submit(txn: AugmentedTransaction, nonce: number | null = null): Promise<TransactionResponse> {
+    const result = txn.args[0]?.result;
+    if (result && result !== "pass") return Promise.reject(result);
+
+    const txnResponse = {
+      chainId: txn.chainId,
+      nonce: nonce ?? 1,
+      hash: "0x4321",
```
nicholaspai, line 39: prefer using a real length hash to test more realistically, or an empty hash for edge case testing

nicholaspai (approved):
LGTM!

test/MultiCallerClient.ts
pxrl: Had some fun here: https://github.com/across-protocol/relayer-v2/pull/426/commits/d9a2277288b787d080e9f0565844bd95567ecd82

src/clients/TransactionClient.ts
pxrl: Fair call; have adjusted it here: https://github.com/across-protocol/relayer-v2/pull/426/commits/72d8639e1a146fb9698af35822e46d6a73efff22

src/clients/TransactionClient.ts
mrice32: That was very helpful. I think this all makes sense as a good next step!

One case I do worry a little about, especially when we have multiple relayers running in parallel, is collisions with other runs.

One enhancement we could consider (at the price of some speed) is taking the max of eth_getTransactionCount and our local nonce tracker. An even more extreme measure could be to try and ask the node for information about the mempool, but these methods would probably be complex and differ by chain.

I don't think we should even think about doing any of this now, but it gives us some options if we continue to have issues with nonces.

mrice32 (approved):
Nonce tracking looks great! +1

https://github.com/across-protocol/relayer-v2/pull/426/files#r1080623605

PR #427, opened 2023-01-18
improve(TransactionClient): Add in txn simulation

pxrl:
Note. This PR a delta to the existing PR here: https://github.com/across-protocol/relayer-v2/pull/426.

### DIFF
--- a/src/clients/TransactionClient.ts
+++ b/src/clients/TransactionClient.ts
@@ -30,8 +30,8 @@ export class TransactionClient {
 
   // Each transaction is simulated in isolation; but on-chain execution may produce different
   // results due to execution sequence or intermediate changes in on-chain state.
-  async simulate(txns: AugmentedTransaction[]): Promise<TransactionSimulationResult[]> {
-    return await Promise.all(txns.map((txn: AugmentedTransaction) => this._simulate(txn)));
+ simulate(txns: AugmentedTransaction[]): Promise<TransactionSimulationResult[]> {
+    return Promise.all(txns.map((txn: AugmentedTransaction) => this._simulate(txn)));
   }
 
   protected async _submit(txn: AugmentedTransaction, nonce: number | null = null): Promise<TransactionResponse> {
--- a/src/clients/TransactionClient.ts
+++ b/src/clients/TransactionClient.ts
@@ -1,0 +1,78 @@
+import {
+  winston,
+  getNetworkName,
+  Contract,
+  runTransaction,
+  BigNumber,
+  etherscanLink,
+  TransactionResponse,
+  TransactionSimulationResult,
+} from "../utils";
+
+export interface AugmentedTransaction {
+  contract: Contract;
+  chainId: number;
+  method: string;
+  args: any[];
+  message: string;
+  mrkdwn: string;
+  value?: BigNumber;
+}
+
+export class TransactionClient {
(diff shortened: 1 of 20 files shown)


src/clients/TransactionClient.ts
```diff
@@ -23,6 +24,16 @@ export class TransactionClient {
   // eslint-disable-next-line no-useless-constructor
   constructor(readonly logger: winston.Logger) {}
 
+  protected async _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
+    return await willSucceed(txn);
+  }
```
pxrl, line 29: Note: Implemented separately as a protected method so it can be easily mocked out in test.

src/clients/TransactionClient.ts
```diff
@@ -23,6 +24,16 @@ export class TransactionClient {
...
+
+  // Each transaction is simulated in isolation; but on-chain execution may produce different
+  // results due to execution sequence or intermediate changes in on-chain state.
+  async simulate(txns: AugmentedTransaction[]): Promise<TransactionSimulationResult[]> {
+    return await Promise.all(txns.map((txn: AugmentedTransaction) => this._simulate(txn)));
```
pxrl, line 34: This implementation is really lightweight because I believe the caller should have the choice of how they want to sort/reconcile the simulation failures, and whether they want to automatically drop them, or potentially abort the transaction sequence (i.e. if subsequent transactions in the queue are dependent on earlier transactions).

test/TransactionClient.ts
```diff
@@ -4,12 +4,25 @@ import { TransactionResponse, TransactionSimulationResult } from "../src/utils";
...
+
 class MockedTransactionClient extends TransactionClient {
 
   constructor(logger: winston.Logger) {
     super(logger);
   }
 
+  protected async _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
+    const result = txn.args[0]?.result;
+    const pass = !(result && result !== passResult);
```
pxrl, line 17: If args[0].result is defined and contains a string other than passResult ("pass"), then we'll mark the simulation failed with that string provided as the reason. This permits the caller to be quite flexible in testing various scenarios.

nicholaspai:
Will review this after merging #426

src/clients/TransactionClient.ts
```diff
@@ -23,6 +24,16 @@ export class TransactionClient {
   // eslint-disable-next-line no-useless-constructor
   constructor(readonly logger: winston.Logger) {}
 
+  protected async _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
+    return await willSucceed(txn);
```
mrice32, line 28:
```suggestion
  protected _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
    return willSucceed(txn);
```

src/clients/TransactionClient.ts
mrice32, line 34: Can ts not infer the AugmentedTransaction type in the inner function?

```suggestion
 simulate(txns: AugmentedTransaction[]): Promise<TransactionSimulationResult[]> {
    return Promise.all(txns.map((txn: AugmentedTransaction) => this._simulate(txn)));
```

mrice32 (approved):
LGTM!

PR #429, opened 2023-01-18
feat: cachable block queries

mrice32:
This PR is the first step in updating the block querying logic to cache event queries more effectively. This changes the algorithm to compute block ranges, so that they repeat as often as possible. With this algorithm, a set of queries can have a maximum of _one_ nonstandard query and if that exists, it will always be the last in the series of queries. I suspect this is the best we can do.

Details:
- The goal of the algorithm is to crate "standard" ranges for a particular maxBlockLookBack.
- The algorithm effectively floors the start of the range to the nearest smaller multiple of `maxBlockLookBack`. The operation to do this is `floor(fromBlock / maxBlockLookBack) * maxBlockLookBack`.
- This means that the blocks produced can be outside of the desired range. This just means we need to filter the events we get back. This filtering is pretty naive right now, but we could optimize it by, for instance, just filtering the first query since that's the only one that can contain blocks outside of the range.
- This PR also changes the definition of maxBlockLookBack because I felt the previous definition was a bit non-intuitive. Previously, it was the max difference of toBlock and fromBlock, but because that range is inclusive, that means that that value is the max blocks we're querying minus one, so it didn't align with the provider's definition of their max ranges.

Next steps:
- With this change, I think it should be simple to add caching. We could:
  - Add a redis parameter to the paginatedEventQuery method. When it's provided, it checks a specific redis key for the range before querying it and stores the returned value if it doesn't find it. Note: it's important that the key preserves some info about the providers used to get the result.
  - Build a wrapper method that can be used where caching is desired and swap over different parts of the system one-by-one.
  - Build caching into the provider, itself.
  - Use proxyd.


src/utils/EventUtils.ts
```diff
@@ -65,25 +70,54 @@ export async function paginatedEventQuery(
...
+ * The range is inclusive, so the results will include events in the fromBlock and in the toBlock.
+ * maxBlockLookback defined the maximum number of blocks to search. Because the range is inclusive, the maximum diff
+ * in the returned pairs is maxBlockLookBack - 1. This is a bit non-intuitive here, but this is meant so that this
+ * parameter more closely aligns with the more commonly understood definition of a max query range that node providers
+ * use.
+ * @returns an array of disjoint fromBlock, toBlock ranges that should be queried. These cover at least the entire
+ * input range, but can include blocks outside of the desired range, so results should be filtered. Results
+ * are ordered from smallest to largest.
+ */
+export function getPaginatedBlockRanges({
```
nicholaspai, line 87: How does this handle edge cases `fromBlock == toBlock` or `maxBlockLookback >= (toBlock - fromBlock)`?

nicholaspai (approved):
This is some nice looking code, just one comment about behavior at edge cases! https://github.com/across-protocol/relayer-v2/pull/429#discussion_r1082660447

src/utils/EventUtils.ts
mrice32: There are test cases for each of these. It does the following:

- fromBlock == toBlock: assuming the fromBlock is a multiple of maxBlockLookback, it returns [fromBlock, fromBlock]. If it's not a multiple, it returns [flooredFromBlock, fromBlock].

- maxBlockLookBack >= toBlock - fromBlock: It may actually return multiple ranges depending on where fromBlock is. It will try to break this query into a "standard" query and a nonstandard query. Calling with (2, 3, 4) => [[0, 3]]. Calling with (3, 5, 4) => [[0,3], [4, 5]]. Does that make sense?

- I also just added a line to ensure that it has the same behavior as the previous code when fromBlock > toBlock (return empty array).

nicholaspai (approved):

src/utils/EventUtils.ts
```diff
@@ -65,25 +70,57 @@ export async function paginatedEventQuery(
...
+  fromBlock,
+  toBlock,
+  maxBlockLookBack,
+}: EventSearchConfig): [number, number][] {
+  // If the maxBlockLookBack is undefined, we can look back as far as we like. Just return the entire range.
+  if (maxBlockLookBack === undefined) return [[fromBlock, toBlock]];
+
+  // If the fromBlock is > toBlock, then return no ranges.
+  if (fromBlock > toBlock) return [];
```
pxrl, line 96: Is this something we ever expect to see in practice? I wonder if it might instead be worth assert()ing on, because it might indicate a bug on the calling side.

pxrl (approved):
This is super clean and nicely documented!

src/utils/EventUtils.ts
mrice32: I would suspect not. However, I tested the previous implementation and this is what it returned, so I wanted to preserve the behavior.

Because it's possible for this to cause instability in the bots in the case of node synchronization issues, I think we should leave this as is and put an assert in a separate change (so we can distinguish issues coming from the assert from issues stemming from this change). Does that make sense?

mrice32: Because I think this is just optional, I'm going to go ahead and merge, so @evaldofelipe can begin testing this with proxyd.

PR #431, opened 2023-01-19
[1/3] improve(MultiCallerClient): Use txnClient for simulation

pxrl:
This is _mostly_ done now because this change is relatively simple, yet it significantly reduces the complexity of the subsequent diff that introduces the more impactful MultiCallerClient changes. This is mostly because there is code deletion and subsequent addition in the same space between the two changes, so the combined diff becomes pretty difficult to follow.

I had to disable a single test as part of this change. This was done so as to avoid making short-term changes to the existing MultiCallerClient implementation: modifying that code too much is an unnecessary risk for no real benefit.


test/MultiCallerClient.ts
```diff
@@ -143,16 +94,16 @@ describe("MultiCallerClient", async function () {
...
           value: toBN(0),
           message: `Test transaction on chain ${chainId}`,
           mrkdwn: `Sample markdown string for chain ${chainId} value transaction`,
         } as AugmentedTransaction);
       });
       expect(multiCaller.transactionCount()).to.equal(chainIds.length);
 
-      const result: string[] = await multiCaller.executeTransactionQueue();
-      expect(result.length).to.equal(fail ? 0 : chainIds.length);
+      const results: string[] = await multiCaller.executeTransactionQueue();
```
nicholaspai, line 105: Why does this now fail/ can't be mocked?

pxrl: Ah; this is still using the legacy implementation in executeTransactionQueue(), where I am reluctant to make changes. Specifically I don't want to migrate it to use TransactionClient.submit() for transaction submission, because that also implies nonce tracking when transactions are submitted in succession, and that's a change from the behaviour that we have today. Because we don't migrate to using submit(), we can't take advantage of mocking out _submit() in MockedTransactionClient(), so we have no way of controlling txn pass/fail on submission in this test.

This test is re-enabled in the PR that would immediately follow this one (https://github.com/across-protocol/relayer-v2/pull/424/files), because in that case we are actually testing the new implementation.

mrice32 (approved):
LGTM!

PR #432, opened 2023-01-19
WIP: Look up older deposits when computing slow fill excesses

nicholaspai:


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -171,66 +173,69 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
   // This assumption depends on the rule that slow fills can only be sent after a partial fill for a non zero amount
   // of the deposit. This is why "1 wei" fills are important, otherwise we'd never know which fills originally
   // triggered a slow fill payment to be sent to the destination chain.
-  allValidFillsInRange
-    .filter((fill) => fill.totalFilledAmount.eq(fill.amount) && !fill.fillAmount.eq(fill.amount))
-    .forEach((fill: interfaces.FillWithBlock) => {
-      const { lastMatchingFillInSameBundle, rootBundleEndBlockContainingFirstFill } =
-        getFillDataForSlowFillFromPreviousRootBundle(
```
nicholaspai, line 178: the only diff here is adding an `await Promise.all` around the map and `await`-ing the `getFillDataForSlowFillFromPreviousRootBundle` function which is now async

PR #434, opened 2023-01-20
improve(MultiCallerClient): Correctness & test improvements

pxrl:
- Verify robustness against a 0-length array being supplied to buildMultiCallBundles().
 - Remove async on buildMultiCallBundles().
 - Improve test naming.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -172,7 +172,7 @@ export class MultiCallerClient {
 
     // Generate the complete set of txns to submit to the network. Anything that failed simulation is dropped.
     const txnRequests: AugmentedTransaction[] = _valueTxns.concat(
-      await this.buildMultiCallBundles(_txns, this.chunkSize[chainId])
+      this.buildMultiCallBundles(_txns, this.chunkSize[chainId])
     );
 
     const txnResponses: TransactionResponse[] =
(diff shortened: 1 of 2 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -407,12 +407,13 @@ export class MultiCallerClient {
...
   }
 
-  async buildMultiCallBundles(
+  buildMultiCallBundles(
     txns: AugmentedTransaction[],
     chunkSize = DEFAULT_MULTICALL_CHUNK_SIZE
-  ): Promise<AugmentedTransaction[]> {
-    const txnChunks: AugmentedTransaction[][] = lodash.chunk(txns, chunkSize);
+  ): AugmentedTransaction[] {
+    if (txns.length === 0) return [];
```
mrice32, line 414: nit: why is this needed? Does lodash.chunk not handle length === 0?

mrice32 (approved):
Looks great!

src/clients/MultiCallerClient.ts
pxrl: The problem is actually on line 419:

```
return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0];
```

If we pass in zero-length array to `buildMultiCallBundles()`, lodash will (correctly) produce the same, and then at line 419 we return element 0 because txnChunks.length is <= 1.

Demonstration:
````
> let y = lodash.chunk([], 50)
undefined
> y
[]
> y[0]
undefined
```

I considered chaining ternary statements on line 419 but it's pretty messy, and I think it's conceptually wrong to make it that far with zero-length input:

```
return txnChunk.length > 1
  ? this.buildMultiCallBundle(txnChunk)
  : txnChunk.length === 1 ? txnChunk[0] : [];
```

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/434#discussion_r1085536427

src/clients/MultiCallerClient.ts
pxrl: Schooled once again! Thanks :)

PR #436, opened 2023-01-25
feat: Add function to SpokePoolClient that fetches deposit older than search range

nicholaspai:


src/clients/SpokePoolClient.ts
```diff
@@ -58,8 +64,13 @@ export class SpokePoolClient {
     readonly configStoreClient: AcrossConfigStoreClient | null,
     readonly chainId: number,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0
+    public spokePoolDeploymentBlock: number = 0
```
nicholaspai, line 67: We allow the deployment block to be overridden since its used to set the `firstDepositIdForSpokePool` (by calling `SpokePool.numberOfDeposits({ blockTag: spokePoolDeploymentBlock })`, and in some of the tests we want to override this variable to test edge cases. See [here](https://github.com/across-protocol/relayer-v2/pull/436/files#diff-784ea628cdd13e0f9b2b3ccf74af9aa48f408c8185d3d3cbeca8ae4c65ee0ff3R190) for example.

src/clients/SpokePoolClient.ts
```diff
@@ -58,8 +64,13 @@ export class SpokePoolClient {
...
   ) {
+    try {
+      this.spokePoolDeploymentBlock = getDeploymentBlockNumber("SpokePool", chainId);
```
nicholaspai, line 70: We should be loading spokePoolDeploymentBlock if possible

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -93,7 +144,110 @@ describe("SpokePoolClient: Fill Validation", async function () {
         destinationToken: zeroAddress,
         realizedLpFeePct: toBN(0),
       })
-    ).to.deep.equal(expectedDeposit);
+    )
+      .excludingEvery(["logIndex", "transactionIndex", "transactionHash"])
+      .to.deep.equal(expectedDeposit);
+  });
+
+  it("Can fetch older deposit matching fill", async function () {
```
nicholaspai, line 152: This test and two following are the new unit tests testing out the new spoke pool client function

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +250,87 @@ export class SpokePoolClient {
...
+    }
+    if (cachedDeposit) {
+      deposit = cachedDeposit as DepositWithBlock;
+    } else {
+      // Binary search between spoke pool deployment block and earliest block searched to find the block range containing
+      // the deposited event, where numberOfDeposits incremented from depositId-1 to depositId.
+      const [blockBeforeDeposit, blockAfterDeposit] = await Promise.all([
+        this.binarySearchForBlockContainingDepositId(Math.max(fill.depositId - 1, this.firstDepositIdForSpokePool)),
+        this.binarySearchForBlockContainingDepositId(Math.min(fill.depositId + 1, this.earliestDepositIdQueried)),
+      ]);
```
mrice32, line 291: Two points:
1. I think line 289 could use depositId, not depositId - 1 because the numberOfDeposits will equal the deposit id starting at the deposit _before_ the one we care about.
2. I think this handles the case where there are multiple blocks between deposits, but I don't think it handles the case where there are multiple deposits in the same block since that could result in many increments for numberOfDeposits in the same block. 

This is roughly what that would look like from the binary search algo's perspective:
block 5: 100
block 6: 100
block 7: 100
block 8: 105
block 9: 105

In that case, I think the binary search algo will fall out of the while and hit the error. I'm not 100% sure, but I think in that case, it should work to return high because in the case that we see an increment like that, we always want to return the block that returns the result _after the increment_ (the block 8 in the example above) since that is the block where the increment occured.

Thoughts?

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +250,87 @@ export class SpokePoolClient {
...
+    if (fill.originChainId !== this.chainId) throw new Error("fill.originChainId !== this.chainid");
+    if (fill.depositId < this.firstDepositIdForSpokePool) return undefined;
+    if (fill.depositId >= this.earliestDepositIdQueried) return this.getDepositForFill(fill);
+
+    let deposit: DepositWithBlock, cachedDeposit: Deposit | undefined;
+    if (this.configStoreClient.redisClient) {
+      cachedDeposit = await getDeposit(getRedisDepositKey(fill), this.configStoreClient.redisClient);
+    }
+    if (cachedDeposit) {
+      deposit = cachedDeposit as DepositWithBlock;
```
mrice32, line 284: Just to be safe, thoughts on adding an assert here to ensure that the deposit id and origin chain id for the deposit match the fill?

test/SpokePoolClient.ValidateFill.ts
mrice32: Awesome, very very nice!

mrice32 (commented):
Looks awesome! Only minor comments

src/clients/SpokePoolClient.ts
```diff
@@ -58,8 +64,13 @@ export class SpokePoolClient {
...
     readonly chainId: number,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0
+    public spokePoolDeploymentBlock: number = 0
   ) {
+    try {
+      this.spokePoolDeploymentBlock = getDeploymentBlockNumber("SpokePool", chainId);
+    } catch (err) {
+      // Probably a test environment. Ignore. Test should have set deployment block manually.
+    }
```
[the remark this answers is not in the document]
nicholaspai: This changes the behavior a little but but I do agree we should error out if `_spokePoolDeploymentBlock` is not inputted and getDeploymentBlockNumber fails.

I was thinking we should default to `getDeploymentBlockNumber` if that succeeds otherwise use the inputted value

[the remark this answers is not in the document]
nicholaspai: wdyt about:

```ts
    try {
      this.spokePoolDeploymentBlock = getDeploymentBlockNumber("SpokePool", chainId);
    } catch (err) {
      // If getDeploymentBlockNumber fails then we're probably in a test environment and we can use the
      // the fallback value. If the deployment block is set before the spoke pool's actual deployment, the update()
      // method will likely fail when calling SpokePool.numberOfDeposits({ blockTag: this.spokePoolDeploymentBlock })
      this.spokePoolDeploymentBlock = fallbackSpokePoolDeploymentBlock;
    }
  }
```

src/clients/SpokePoolClient.ts
nicholaspai: sure, curious why you think we need extra checks here. Is it because there' a chance `getRedisDepositKey` could be implemented incorrectly?

src/clients/SpokePoolClient.ts
nicholaspai: Hm this is a good point, how do you propose we handle this case when we fall out of the while?

src/clients/SpokePoolClient.ts
nicholaspai: I've added some unit tests if you want to take a deeper look at the binary search behavior https://github.com/across-protocol/relayer-v2/pull/436/commits/2e00182deaab3a2149c161ffd5c9f3de9091663b

src/clients/SpokePoolClient.ts
nicholaspai: Ok I've added some more unit tests here, im not sure 100% that this is the right behavior... can you take a look at the tests and see if its what youd expect? https://github.com/across-protocol/relayer-v2/pull/436/commits/9ec74d85ad06e55e387a2ff84772532d6f74cf72

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +257,109 @@ export class SpokePoolClient {
...
+  // where `numberOfDeposits == targetDepositId`.
+  async binarySearchForBlockContainingDepositId(
+    targetDepositId: number,
+    fallback: BINARY_SEARCH_FALLBACK = "NONE"
+  ): Promise<number> {
+    // After an update(), firstBlockToSearch increments to the toBlock+1 so clamp that at the latestBlockNumber to
+    // avoid calling spokePool.numberOfDeposits for a block that doesn't exist yet.
+    let high = this.latestBlockNumber
+      ? Math.min(this.firstBlockToSearch, this.latestBlockNumber)
+      : this.firstBlockToSearch;
```
pxrl, line 271: Potential simplification:

```suggestion
    let high = Math.min(this.firstBlockToSearch, this.latestBlockNumber ?? this.firstBlockToSearch);
```

src/utils/EventUtils.ts
```diff
@@ -1,5 +1,5 @@
 import { delay } from "@uma/financial-templates-lib";
-import { DepositWithBlock, FillWithBlock, SortableEvent } from "../interfaces";
+import { SortableEvent } from "../interfaces";
```
pxrl, line 2: Is this just a driveby change?

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +257,109 @@ export class SpokePoolClient {
...
+    } else {
+      // Binary search between spoke pool deployment block and earliest block searched to find the block range containing
+      // the deposited event. Find a block before and after the deposit event, which was emitted when
+      // depositId incremented from fill.depositId to fill.depositId+1
+      const [blockBeforeDeposit, blockAfterDeposit] = await Promise.all([
+        // Ensure that `blockBeforeDeposit` returns and is less than the block where depositId incremented
+        // from depositId-1 to depositId. Similarly ensure that blockAfter deposit returns and is greater than
+        // the block where depositId incremented from depositId to depositId+1.
+        this.binarySearchForBlockContainingDepositId(fill.depositId, "LOW"),
+        this.binarySearchForBlockContainingDepositId(fill.depositId + 1, "HIGH"),
```
pxrl: I'm not sure I grasp the scenario where the lookup fails and we revert to "HIGH" and/or "LOW". Do you have particular scenarios in mind that would cause a failure?

In case of a failure, wouldn't that typically produce a really large range of blocks? I think that would probably cause us to fail further on, in paginatedEventQuery(), since we can't query the RPC provider over such large block ranges.

src/utils/EventUtils.ts
nicholaspai: yep

src/clients/SpokePoolClient.ts
nicholaspai: See this example by matt: https://github.com/across-protocol/relayer-v2/pull/436#discussion_r1087280547

After the binary search completes, the range of blocks [low, high] gets constrained a bit. That plus filtering on `depositId` should produce a quick RPC request IMO

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
pxrl: > This seems like it could be a source of errors/issues if this ever starts failing since we seem to depend on it.
> 
> Why not require it to either be provided or for the lookup to succeed? For instance, you could have the input be something like:
> 
> ```ts
>     _spokePoolDeploymentBlock?: number
>   ) {
>     this.spokePoolDeploymentBlock = _spokePoolDeploymentBlock === undefined ? getDeploymentBlockNumber("SpokePool", chainId) : _spokePoolDeploymentBlock;
>     ...
> ```

This seems like the simplest approach; if the caller has specified a block to use then it seems counter-intuitive to me that the SpokePool would ignore that and prefer something else instead, only use it anyway if the getDeploymentBlockNumber() lookup fails.

If the getDeploymentBlockNumber() error is not suppressed then we'll also spot any remaining places that it needs to be fixed in test, and I also like that it'd fail hard in prod, rather than presenting via some consequent failure later on.

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +257,121 @@ export class SpokePoolClient {
...
+    // Note: now that low > high, return high for LOW and low for HIGH
+    if (fallback === "LOW") return high;
+    else if (fallback === "HIGH") return low;
+    else
+      throw new Error(
+        `Binary search between [${initLow}, ${initHigh}] failed to find block containing deposit ID ${targetDepositId}`
+      );
+  }
+
+  // Load a deposit for a fill if the fill's deposit ID is less than this client's earliest searched
```
pxrl, line 291: "...if the fill's deposit ID is less than this client's earliest searched deposit ID..." is contradicted by line 301, which handles that case anyway.

src/clients/SpokePoolClient.ts
```diff
@@ -301,9 +434,12 @@ export class SpokePoolClient {
...
     timerStart = Date.now();
-    const queryResults = await Promise.all(
-      eventSearchConfigs.map((config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig))
-    );
+    const [_earliestDepositIdForSpokePool, ...queryResults] = await Promise.all([
+      this.spokePool.numberOfDeposits({ blockTag: this.spokePoolDeploymentBlock }),
+      ...eventSearchConfigs.map((config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig)),
+    ]);
+
+    this.firstDepositIdForSpokePool = _earliestDepositIdForSpokePool;
```
pxrl, line 442: Is there a new "set deposit ID" feature to be added to the SpokePool contracts? That might be a problem for this logic if the deposit ID defaults to 0 and can jump significantly post-deployment.

src/clients/SpokePoolClient.ts
pxrl: Ah - I missed that thread. Makes sense. I assumed it was intended to handle the case of the RPC provider failing somehow and hadn't considered the instance of multiple deposits in a single block.

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +257,121 @@ export class SpokePoolClient {
...
+          "HIGH",
+          this.spokePoolDeploymentBlock,
+          this.eventSearchConfig.fromBlock
+        ),
+      ]);
+      assert(blockBeforeDeposit <= blockAfterDeposit, "blockBeforeDeposit > blockAfterDeposit");
+
+      const query = await paginatedEventQuery(
+        this.spokePool,
+        this.spokePool.filters.FundsDeposited(null, null, null, null, fill.depositId),
```
mrice32, line 336: Sorry, @pxrl helped me realize that I made a bad suggestion previously. I thought depositId was indexed, but that was _after_ the recent updates. The deployed contracts do not have depositId indexed, so I think this filter does nothing. Sorry for the bad suggestion!

src/clients/SpokePoolClient.ts
```diff
@@ -239,6 +257,121 @@ export class SpokePoolClient {
...
+      // the deposited event. Find a block before and after the deposit event, which was emitted when
+      // depositId incremented from fill.depositId to fill.depositId+1
+      const [blockBeforeDeposit, blockAfterDeposit] = await Promise.all([
+        // Ensure that `blockBeforeDeposit` returns and is less than the block where depositId incremented
+        // from depositId-1 to depositId. Similarly ensure that blockAfter deposit returns and is greater than
+        // the block where depositId incremented from depositId to depositId+1. We use the "LOW"/"HIGH" fallback
+        // options here to ensure that blockBeforeDeposit <= targetBlock <= blockAfterDeposit.
+        this.binarySearchForBlockContainingDepositId(
+          fill.depositId,
+          "LOW",
```
mrice32, line 321: Can this be HIGH, too? If we're looking for deposit id 4, block 15 has numberOfDeposits == 3 and block 16 has numberOfDeposits == 10, we should be able to start the query at block 16, right? Since the increment happened in 16 and block ranges are inclusive? Do you agree?

I don't think this matters a ton (an extra block of querying), I just want to make sure we all are on the same page about the casework here.

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
mrice32: The reason I worry about the fallback approach as it's currently configured is that we could accidentally end up in the fallback case in production and not realize it. If this is something we think should _never_ happen, then I think we should be absolutely sure the bot errors in that case.

Is there a reason that is unworkable?

src/clients/SpokePoolClient.ts
nicholaspai: This new feature actually sets the deposit ID on construction :)

nicholaspai: so I think the test cases handle this

src/utils/RedisUtils.ts
```diff
@@ -20,6 +21,37 @@ export async function setRedisKey(
...
+
+export async function setDeposit(deposit: Deposit, redisClient: RedisClient, expirySeconds = 0): Promise<void> {
+  await setRedisKey(
+    getRedisDepositKey(deposit),
+    JSON.stringify({
+      ...deposit,
+      amount: deposit.amount.toString(),
+      relayerFeePct: deposit.relayerFeePct.toString(),
+      realizedLpFeePct: deposit.realizedLpFeePct.toString(),
+    }),
```
mrice32, line 36: nit: there are ways to make the json stringifier and parser do these conversions automatically. Have you considered doing that instead of manually modifying each field?

JSON.parse() has a second argument that can detect particular types and rework them (called revive).

JSON.stringify() detects objects with a toJSON method (can be added to prototype) and uses that method to stringify the object. You could create a specific structure in the json to represent the BN and detect that structure in the revive function you pass into JSON.parse().

References:
https://yonatankra.com/custom-json-stringify-to-classes-with-the-tojson-method/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

src/clients/SpokePoolClient.ts
nicholaspai: `depositId` will be indexed in the new contracts so let's leave it in but I will add a filter on `fill.depositor` which is indexed so it should optimize the RPC request a bit

src/clients/SpokePoolClient.ts
nicholaspai: So I guess the fallback == HIGH/LOW only gets hit if the deposit ID we are searching is in a multiblock deposit right? In that case then HIGH always works?

Is that statement accurate? If so we can eliminate the LOW option and just allow caller to set fallback=true in which case we default to `high` if we fall out of the while loop

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
nicholaspai: Ok i hear you guys, i'll revert to the suggestion of using default when provided

src/utils/RedisUtils.ts
nicholaspai: Can you add toJSON to an interface? I assume we'd want to add this to the `Deposit` interface

src/clients/SpokePoolClient.ts
nicholaspai: This method is only intended to be used as a fall back if `getDepositForFill` would return `undefined`. So if "the fill's deposit ID is less than this client's earliest searched deposit ID" then this function should run normally, otherwise we should just return `getDepositForFill` which should contain a deposit with ID >= fill's deposit ID right?

src/utils/RedisUtils.ts
nicholaspai: Ohh this is super nice, replaced the code here: https://github.com/across-protocol/relayer-v2/pull/436/commits/20627aec75a068b436547a6255189536f9fabf5f

src/clients/SpokePoolClient.ts
nicholaspai: thinking about this more I agree with you perhaps the default fallback behavior should be what HIGH returns now. However, does this mean we should always fallback to high if we fall out of the while loop? If so, then how do we distinguish between true errors (i.e depositId's we can't find) versus multiple deposits in the same block?

Do we assume that the former error can never happen, because all depositID's are >= the `firstDepositIdForSpokePool` and depositId's are monotonically increasing?

mrice32: This is a great point. Would that error be caught in the lookup _after_ the binary search, when the intended deposit id is not found? I can't really see how to detect it explicitly here, since we only have block granularity. Thoughts?

src/utils/RedisUtils.ts
mrice32: Nice!!

src/clients/SpokePoolClient.ts
```diff
@@ -58,8 +68,12 @@ export class SpokePoolClient {
     readonly configStoreClient: AcrossConfigStoreClient | null,
     readonly chainId: number,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    readonly spokePoolDeploymentBlock: number = 0
+    readonly fallbackSpokePoolDeploymentBlock?: number
```
mrice32, line 71: nit: with the new logic, this probably isn't a fallback. Thoughts on calling it `spokePoolDeploymentBlock` as it was before just optional now?

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
nicholaspai: it would have to be caught after the binary search, but we have an additional check here that we can rely on. Basically if the event isn't returned between the blocks returned by the two binary searches then we'll error out. So I think the fallback is OK to default to HIGH since we'd always want that based on how multiple deposits in the same block work.

```ts
      const query = await paginatedEventQuery(
        this.spokePool,
        // TODO: When SpokePool V2 is deployed we can begin to filter on fill.depositId as well as fill.depositor,
        // which should optimize the RPC request.
        this.spokePool.filters.FundsDeposited(null, null, null, null, null, null, null, null, fill.depositor),
        {
          fromBlock: blockBeforeDeposit,
          toBlock: blockAfterDeposit,
          maxBlockLookBack: 0,
        }
      );
      const event = (query as FundsDepositedEvent[]).find((deposit) => deposit.args.depositId === fill.depositId);
      if (event === undefined)
        throw new Error(
          `Could not find deposit ID ${fill.depositId} for fill between blocks [${blockBeforeDeposit}, ${blockAfterDeposit}]`
        );
```

src/clients/SpokePoolClient.ts
nicholaspai: good call

src/clients/SpokePoolClient.ts
pxrl: Oh nice - that's neat.

src/clients/SpokePoolClient.ts
pxrl: It seems a _little_ bit redundant to have the fallback capability even though it's intended that the caller has already tried that themselves. But that's mostly just nitpicking :)

pxrl (commented):
No further comments from me :+1:

mrice32 (approved):
LGTM!

PR #437, opened 2023-01-25
improve(validate-script): Script can efficiently reconstruct very old bundle

nicholaspai:
Script  essentially figures out the latest events it should need to query to reconstruct an older bundle, so the range of events to query is fixed based on DATAWORKER_FAST_LOOKBACK_COUNT

### DIFF
--- a/src/scripts/validateRootBundle.ts
+++ b/src/scripts/validateRootBundle.ts
@@ -22,13 +22,14 @@ import {
   getDisputedProposal,
   getBlockForTimestamp,
   getCurrentTime,
+  sortEventsDescending,
 } from "../utils";
 import {
   constructSpokePoolClientsForFastDataworker,
   getSpokePoolClientEventSearchConfigsForFastDataworker,
   updateDataworkerClients,
 } from "../dataworker/DataworkerClientHelper";
-import { PendingRootBundle } from "../interfaces";
+import { PendingRootBundle, ProposedRootBundle } from "../interfaces";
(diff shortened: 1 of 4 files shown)


src/scripts/validateRootBundle.ts
```diff
@@ -79,13 +80,26 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet): Pro
...
+      dvm,
+      clients.configStoreClient,
+      priceRequestTime,
+      priceRequestBlock
+    );
+  } catch (_err) {
+    logger.debug({
+      at: "Dataworker#validate",
+      message: "No dispute found for request time, falling back to most recent root bundle before request time",
+    });
```
mrice32, line 96: nit: prefer to not use errors as branching logic because it means that other errors might be miscategorized as implying that a dispute was not found.

What do you think of these two options:
1. Split getDisputedProposal into two methods: one to find a dispute and the other to find a proposal associated with a passed in dispute?
2. Have getDisputedProposal return the proposal and/or an enum that tells you which part it was unable to find.

Thoughts?

mrice32 (approved):
LGTM!

src/scripts/validateRootBundle.ts
nicholaspai: I agree with this I went with option 1 as I want to avoid returning tuples where possible https://github.com/across-protocol/relayer-v2/pull/437/commits/32da25b33c995a777ffce7bab3ba8d808d2cc59c

mrice32 (approved):
LGTM!

PR #438, opened 2023-01-26
chore(config): Prevent non-zero POLLING_DELAY

pxrl:
This is an administrative change to temporarily prevent operators from running in
looping mode. The justification for this is that we currently do not apply the same
QA to non-zero polling delay (looping mode) as we do to serverless mode.

Looping mode is useful for third-party operators and the intention is ultimately to
support it.

### DIFF
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -40,7 +40,8 @@ export class CommonConfig {
     this.maxRelayerLookBack = Number(MAX_RELAYER_DEPOSIT_LOOK_BACK ?? Constants.MAX_RELAYER_DEPOSIT_LOOK_BACK);
     this.hubPoolChainId = Number(HUB_CHAIN_ID ?? 1);
     this.spokePoolChains = CONFIGURED_NETWORKS ? JSON.parse(CONFIGURED_NETWORKS) : Constants.CHAIN_ID_LIST_INDICES;
-    this.pollingDelay = Number(POLLING_DELAY ?? 60);
+    this.pollingDelay = Number(POLLING_DELAY ?? 0);
+    assert(this.pollingDelay === 0, "Non-zero POLLING_DELAY is currently unsupported");
     this.maxBlockLookBack = MAX_BLOCK_LOOK_BACK ? JSON.parse(MAX_BLOCK_LOOK_BACK) : {};
     if (Object.keys(this.maxBlockLookBack).length > 0)
       for (const chainId of this.spokePoolChains)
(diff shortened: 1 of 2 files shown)


src/common/Config.ts
```diff
@@ -40,7 +40,8 @@ export class CommonConfig {
     this.maxRelayerLookBack = Number(MAX_RELAYER_DEPOSIT_LOOK_BACK ?? Constants.MAX_RELAYER_DEPOSIT_LOOK_BACK);
     this.hubPoolChainId = Number(HUB_CHAIN_ID ?? 1);
     this.spokePoolChains = CONFIGURED_NETWORKS ? JSON.parse(CONFIGURED_NETWORKS) : Constants.CHAIN_ID_LIST_INDICES;
-    this.pollingDelay = Number(POLLING_DELAY ?? 60);
+    this.pollingDelay = Number(POLLING_DELAY ?? 0);
```
nicholaspai, line 43: `pollingDelay` is used in many other files in this repo can you do a search and remove those instances/branches?

nicholaspai (changes requested):
https://github.com/across-protocol/relayer-v2/pull/438/files#r1087757261

src/common/Config.ts
pxrl: I found 19 instances of "polligDelay" in the code, but this is the only place where it's set, so all of the existing code should continue to run OK as it does in the prod environment.

Note I'm assuming that we only want to disable this functionality until we determine how to properly test and support a non-zero POLLING_DELAY. If we don't want to support it at all then it would definitely make sense to remove it.

In general I'd like to see it retained as a future possibility, because there are a few overheads during startup for the bot. Some of these overheads should become smaller as we get caching of getLogs requests in redis. wdyt?

test/Monitor.ts
```diff
@@ -84,6 +84,7 @@ describe("Monitor", async function () {
       MONITOR_REPORT_INTERVAL: "10",
       MONITORED_RELAYERS: `["${depositor.address}"]`,
       CONFIGURED_NETWORKS: JSON.stringify(configuredNetworks),
+      POLLING_DELAY: "60",
```
pxrl, line 87: I consider this specific change to be a bodge, since the test fails without it. Per current implementation the Monitor test seems to only work with a non-zero polling delay. This one needs a bit more investigation.

src/common/Config.ts
nicholaspai: Hm ok I see so this PR amounts to defaulting to 0 instead of "preventing" non zero

test/Monitor.ts
nicholaspai: This is definitely a bodge/hack. I'd feel more comfortable if we knew why this test only works with polling delay = 60, and fixed it so that it works with 0. This is because we run monitor with polling delay = 0 in prod so the test would be closer to prod

src/common/Config.ts
pxrl: It does two things:
 - If POLLING_DELAY is not env-specified, change the default from 60 to 0.
 - If POLLING_DELAY is env-specified and is NOT 0, assert to prevent further execution.

PR #439, opened 2023-01-27
fix(MultiCallerClient): Retain txn message

pxrl:
The new MultiCallerClient is dropping the transaction message when
it is creating bundles.


mrice32 (approved):
LGTM!

PR #440, opened 2023-01-31
feat: add caching logic to intercept provider calls

mrice32:
This PR adds a new provider in the inheritance tree that performs caching on requests.

Today, it only caches eth_getLogs, but the design is intended to be flexible s.t. any request that only depends on older block info can be cached.

**Rough architecture**

The caching provider sits at the top of the inheritance tree. This is intentional because each of our custom providers intercepts send() calls and passes them up the stack if they intend to make the send call to the node. Because this provider is a simple caching layer above the node, it is hit _after_ the retry layer and the rate limiting layer. Notes:
- It's possible to swap this and the rate limit layer if we do not intend to rate limit our redis calls, but it seemed somewhat reasonable to try to rate limit the redis calls too, especially since some of our rate limiting needs come from overloading _our own_ networking stack.
- It cannot go below the retry layer, since the retry layer farms out its calls to multiple nodes, and caching is intended to be per-node.
- Caching can be disabled by not providing a redisClient. The cache then becomes a passthrough.

The process it follows is effectively the following:
- At construction time, it builds a redis key prefix that includes all unchanging information. For instance, the key takes into account the base of the node url (i.e. mainnet.infura.io), the chain id, and the method (eth_getLogs). The only information that can't be preconstructed is the parameters of each request.
- On each send, the provider checks to see if it has caching enabled (i.e. the redisClient is defined).
- Then it checks if this request is cacheable:
  - To do this, it grabs eth_blockNumber (or uses a cached value if previously requested within the last 60 seconds). Note: this internal block number caching is built into the base provider.
  - The provider has a hardcoded mapping of chainId to the reorg distance for that chain.
  - If the latest block being touched by the query is within the reorg distance of the current block, then the query cannot be cached.
  - This is done even when the reorg distance is 0, to avoid caching queries that request information about future blocks (nodes do allow this).
- If it's cachable, the redis key is constructed, and redis is asked for the value.
- If the value does not exist, then the value is queried via the node.
- The value is then stored in redis with a hardcoded TTL of an hour.
- If it's not cachable, it's just passed through.


src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,95 @@ import lodash from "lodash";
...
 import createQueue, { QueueObject } from "async/queue";
-import { Logger } from ".";
+import { Logger, RedisClient, setRedisKey } from ".";
 
 const logger = Logger;
 
+// This is the max distance on each chain that reorgs can happen.
+// Caching will not be allowed for queries whose responses depend on blocks closer than this many blocks.
+// This is intended to be conservative.
+const MAX_REORG_DISTANCE: { [chainId: number]: number } = {
```
nicholaspai, line 13: Can you merge this logic with `RedisUtils.shouldCache`? They are designed to serve the same use case IIUC

src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,95 @@ import lodash from "lodash";
...
+    return this.getLogsCachePrefix + JSON.stringify(params);
+  }
+
+  private async shouldCache(method: string, params: Array<any>): Promise<boolean> {
+    // Today, we only cache eth_getLogs. We could add other methods here, where convenient.
+    if (method !== "eth_getLogs") return false;
+    const [{ fromBlock, toBlock }] = params;
+
+    // Handle odd cases where the ordering is flipped, etc.
+    const latestQueriedBlock = Math.max(parseInt(fromBlock, 16), parseInt(toBlock, 16));
```
nicholaspai, line 76: What happens if you call `eth_getLogs` with toBlock > fromBlock? Does that even return anything?

src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,95 @@ import lodash from "lodash";
...
+
+    // Note: this method is an internal method provided by the BaseProvider. It allows the caller to specify a maxAge of
+    // the block that is allowed. This means if a block has been retrieved withint the last n seconds, no provider
+    // query will be made.
+    const blockNumber = await super._getInternalBlockNumber(BLOCK_NUMBER_TTL * 1000);
+
+    // We ensure that the toBlock is not within the max reorg distance to avoid caching unstable information.
+    const firstUnsafeBlockNumber = blockNumber - this.maxReorgDistance;
+
+    // toBlock is in hex, so it must be parsed before being compared to the first unsafe block.
```
nicholaspai, line 90: This comment seems outdated

src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,95 @@ import lodash from "lodash";
...
+    const latestQueriedBlock = Math.max(parseInt(fromBlock, 16), parseInt(toBlock, 16));
+
+    // Handle cases where the input block numbers are not hex values ("latest", "pending", etc).
+    // This would result in the result of the above being NaN.
+    if (Number.isNaN(latestQueriedBlock)) return false;
+
+    // Note: this method is an internal method provided by the BaseProvider. It allows the caller to specify a maxAge of
+    // the block that is allowed. This means if a block has been retrieved withint the last n seconds, no provider
+    // query will be made.
+    const blockNumber = await super._getInternalBlockNumber(BLOCK_NUMBER_TTL * 1000);
```
nicholaspai, line 85: Is there documentation on this function? I'm not sure I understand it, does it return the block that is `BLOCK_NUMBER_TTL * 1000` seconds older than its latest block?

src/utils/ProviderUtils.ts
```diff
@@ -49,6 +131,9 @@ class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
   }
 
   override async send(method: string, params: Array<any>): Promise<any> {
+    // Ignore rate limiting for eth_blockNumber calls. Otherwise, The internal blocknumber calls in the caching
+    // provider can hang.
+    if (method === "eth_blockNumber") return super.send(method, params);
```
nicholaspai, line 136: This seems like it could be a big change. We send eth_blockNumber requests a lot, right? So are we openining up the bots to more 429 errors?

src/utils/RedisUtils.ts
```diff
@@ -9,6 +10,26 @@ export type RedisClient = ReturnType<typeof createClient>;
 // Current time must be >= 5 minutes past the event timestamp for it to be stable enough to cache.
 export const REDIS_CACHEABLE_AGE = 300;
 
+// Make the redis client for a particular url essentially a singleton.
+const redisClients: { [url: string]: Promise<RedisClient> } = {};
+
+export function getRedis(url: string, logger?: winston.Logger) {
```
nicholaspai, line 16: +1 we can use this elsewhere

src/utils/ProviderUtils.ts
mrice32: This is by block, not timestamp, so they work a bit differently, no? Or is there some common logic I'm missing?

src/utils/ProviderUtils.ts
mrice32: I know alchemy errors, but their error is a bit odd: "One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found."

I'm not sure if it's guaranteed to error or produce consistent results.

src/utils/ProviderUtils.ts
mrice32: https://github.com/ethers-io/ethers.js/blob/ec1b9583039a14a0e0fa15d0a2a6082a2f41cf5b/packages/providers/src.ts/base-provider.ts#L903-L905

It basically takes in a maxAge in ms (which is why we multiply by 1000). This is just the _most stale_ the block number can be to avoid a fresh query. So this is saying "give me the most recent blocknumber query you've done, unless it was over 60 seconds ago. In that case, re-query the block number value".

The goal is just to avoid making a ton of blocknumber queries to the node. So, instead, we allow the old value to be used for up to 60 seconds before making a new query. The BaseProvider class just happens to already have a method to do this.

src/utils/ProviderUtils.ts
mrice32: Yeah, I would prefer to not have to do this. The issue is that if you have a rate limit of 10 concurrent requests, and then you have 10 eth_getLogs queries come in at the same time (common in our code), then the code hangs indefinitely because the eth_getLogs queries are waiting on the block number, but the block number query can't be processed until the eth_getLogs finish. It's essentially a deadlock.

One way to work around this is to flip the order of inheritance. This would mean we don't rate limit how fast requests hit the CacheProvider, we only rate limit how fast the CacheProvider can forward on calls to the node. The only disadvantage to that setup is that we aren't putting redis behind the rate-limiter. That might be okay, though. Not sure if the redis client is smart enough to deal with overloading.

Thoughts?

src/utils/ProviderUtils.ts
mrice32: Yep. Will fix.

src/utils/ProviderUtils.ts
nicholaspai: even though `RedisUtils.shouldCache` uses timestamp, it really is a protection against Reorgs so by block would make sense too

src/utils/ProviderUtils.ts
nicholaspai: hmm but if `fromBlock > toBlock`, even though the `latestQueriedBlock = fromBlock`, does that even query anything..?

src/utils/ProviderUtils.ts
nicholaspai: Would we need to add our own redis rate [limiting algo](https://stackoverflow.com/questions/13175050/how-to-implement-rate-limiting-using-redis) if we added rate limiting after the CacheProvider?

The issue with Redis is its single-threaded so you might end up with a big queue of requests waiting to be sent to redis server

src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,94 @@ import lodash from "lodash";
...
 import createQueue, { QueueObject } from "async/queue";
-import { Logger } from ".";
+import { Logger, RedisClient, setRedisKey } from ".";
 
 const logger = Logger;
 
+// This is the max distance on each chain that reorgs can happen.
+// Caching will not be allowed for queries whose responses depend on blocks closer than this many blocks.
+// This is intended to be conservative.
+const MAX_REORG_DISTANCE: { [chainId: number]: number } = {
```
pxrl, line 13: Did you consider defining this in src/common/Constants.ts? The "minimum deposit confirmations" has a similar need to know max reorg distance, so it seems like something that could be shared.

src/utils/ProviderUtils.ts
```diff
@@ -3,10 +3,94 @@ import lodash from "lodash";
...
+
+      // If cache has the result, parse the json and return it.
+      if (redisResult) return JSON.parse(redisResult);
+
+      // Cache does not have the result. Query it directly and cache.
+      const result = await super.send(method, params);
+
+      // Commit result to redis.
+      await setRedisKey(redisKey, JSON.stringify(result), this.redisClient, PROVIDER_CACHE_TTL);
+    }
```
pxrl, line 60: Is it intended to fall through here and repeat the call to super.send() ?

src/utils/ProviderUtils.ts
mrice32: Yep, exactly! And we can always tweak this number. If we set it to 0, it always does a fresh call.

src/utils/ProviderUtils.ts
mrice32: I think it _should_ either return [] or error. If we want, we could force this to always error by producing an error here, since technically we should never make queries like that.

src/utils/ProviderUtils.ts
mrice32: I will look into it. I think in the other case, we have the opposite situation, we have the timestamp and are going to look up (and cache) the block. Before changing the way the other one works, I need to see if using the block instead of timestamp would cause us to make any additional queries. If not, we can definitely make that change.

src/utils/ProviderUtils.ts
mrice32: If I'm not mistaken, that SO post is talking about implementing rate limiting _using_ redis.

I looked into the redis nodejs client, and I think it internally does something called "auto pipelining" where it batches the requests to the server within a single "tick". This apparently enhances performance a lot.

My intuition is that we aren't even close to hitting the performance limits of redis. However, maybe to make this super configurable, we should make an env variable that allows us to _disable_ provider caching? Thoughts?

src/utils/ProviderUtils.ts
mrice32: No, it isn't! That is definitely a silent bug. We would be querying twice in that case!

Great catch @pxrl!

mrice32: Fixed!

src/utils/ProviderUtils.ts
pxrl: I prefer erroring, since it's a logic error on the calling side.

src/utils/ProviderUtils.ts
pxrl: > My intuition is that we aren't even close to hitting the performance limits of redis.

I'd also be surprised if we were stressing redis - it's designed to be highly performant, and we're reading a lot more regularly than we are writing.

src/utils/ProviderUtils.ts

src/utils/ProviderUtils.ts
mrice32: Okay, I reversed the order of the inheritance to put redis _outside_ of the rate limiter. This could also offer us speedups in prod, since redis can be hit as hard as it can handle.

src/utils/ProviderUtils.ts
mrice32: Added an error! PTAL.

src/utils/ProviderUtils.ts
```diff
@@ -194,6 +110,82 @@ function compareRpcResults(method: string, rpcResultA: any, rpcResultB: any): bo
...
+    const toBlockNumber = parseInt(toBlock, 16);
+
+    // Handle cases where the input block numbers are not hex values ("latest", "pending", etc).
+    // This would result in the result of the above being NaN.
+    if (Number.isNaN(fromBlockNumber) || Number.isNaN(toBlockNumber)) return false;
+    if (toBlockNumber < fromBlockNumber)
+      throw new Error("CacheProvider::shouldCache toBlock cannot be smaller than fromBlock.");
+
+    // Note: this method is an internal method provided by the BaseProvider. It allows the caller to specify a maxAge of
+    // the block that is allowed. This means if a block has been retrieved withint the last n seconds, no provider
```
pxrl, line 179:
```suggestion
    // the block that is allowed. This means if a block has been retrieved within the last n seconds, no provider
```

src/utils/ProviderUtils.ts
```diff
@@ -404,6 +407,17 @@ export function getProvider(chainId: number, logger?: winston.Logger, redisClien
...
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "1000");
 
+  // Provider caching defaults to being enabled if a redis instance exists. It can be manually disabled by setting
+  // DISABLE_PROVIDER_CACHING=true.
+  const disableProviderCache = DISABLE_PROVIDER_CACHING !== "true";
+
+  // This environment variable allows the operator to namespace the cache. This is useful if multiple bots are using
+  // the cache and the operator intends to have them not share.
+  // It's also useful as a way to synthetically "flush" the provider cache by modifying this value.
+  // A reccomended naming strategy is "NAME_X" where NAME is a string name and 0 is a numerical value that can be
```
pxrl, line 417:
```suggestion
  // A recommended naming strategy is "NAME_X" where NAME is a string name and 0 is a numerical value that can be
```

src/utils/ProviderUtils.ts
```diff
@@ -388,7 +383,15 @@ export function getProvider(chainId: number, logger?: winston.Logger, redisClien
...
   }
-  const { NODE_RETRIES, NODE_RETRY_DELAY, NODE_QUORUM, NODE_TIMEOUT, NODE_MAX_CONCURRENCY } = process.env;
+  const {
+    NODE_RETRIES,
+    NODE_RETRY_DELAY,
+    NODE_QUORUM,
+    NODE_TIMEOUT,
+    NODE_MAX_CONCURRENCY,
+    DISABLE_PROVIDER_CACHING,
+    PROVIDER_CACHE_NAMESPACE,
```
pxrl, line 393: _Very_ minor: Since these env vars are all related, are we interested in using the existing common prefix NODE_*?

src/utils/ProviderUtils.ts
```diff
@@ -194,6 +110,82 @@ function compareRpcResults(method: string, rpcResultA: any, rpcResultB: any): bo
...
+    // Handle cases where the input block numbers are not hex values ("latest", "pending", etc).
+    // This would result in the result of the above being NaN.
+    if (Number.isNaN(fromBlockNumber) || Number.isNaN(toBlockNumber)) return false;
+    if (toBlockNumber < fromBlockNumber)
+      throw new Error("CacheProvider::shouldCache toBlock cannot be smaller than fromBlock.");
+
+    // Note: this method is an internal method provided by the BaseProvider. It allows the caller to specify a maxAge of
+    // the block that is allowed. This means if a block has been retrieved withint the last n seconds, no provider
+    // query will be made.
+    const currentBlockNumber = await super._getInternalBlockNumber(BLOCK_NUMBER_TTL * 1000);
```
pxrl, line 181: Something to be aware of: we might have been rugged by ethers on _getInternalBlockNumber.

I spotted that they now have a `main` branch in addition to `master`, and it seems like ethers v6 will be produced from `main`. I can't find base-provider.ts, and _getInternalBlockNumber() seems to be missing entirely: https://github.com/ethers-io/ethers.js/tree/main/src.ts/providers

pxrl (approved):
Nice! I spotted a couple of minor typos and had an observation about the futureproofness of this RE ethers v6. But this should deliver a huge benefit for us...excited to see it!

src/utils/ProviderUtils.ts
mrice32: I think a lot of our provider logic may need to be overhauled in v6. Until it's out of beta, I don't think we should worry about getting in sync with it.

We can (and may have to!) just replicate this functionality internally.

src/utils/ProviderUtils.ts
mrice32: Sure

pxrl (approved):

PR #441, opened 2023-01-31
improve: Remove test that we've .skip()-ed for a while

nicholaspai:


pxrl:
IIRC this was disabled because it used to have a lot of sporadic failures right?

pxrl (approved):
We now have tests that cover the multicall txn submission on various chains, so it seems like the only real advantage of keeping this test would seem to be that it resembles an e2e test. Gut feel is that we don't need it.

mrice32 (approved):
If this is covered elsewhere LGTM!

PR #442, opened 2023-02-01
feat(dataworker): Look up older deposits when computing slow fill excesses for bundles

nicholaspai:
Follows similar logic to https://github.com/across-protocol/relayer-v2/pull/382 by adding methods to the `SpokePoolClient` to query historical fills matching an input Fill when it can't be found in the client's memory


src/utils/FillUtils.ts
```diff
@@ -98,25 +106,40 @@ export function getLastMatchingFillBeforeBlock(
...
   // Find the first fill chronologically for matched deposit for the input fill.
-  const firstFillForSameDeposit = allValidFills.find(
-    (_fill: FillWithBlock) => filledSameDeposit(_fill, fill) && isFirstFillForDeposit(_fill as Fill)
+  const allMatchingFills = sortEventsAscending(
+    allValidFills.filter((_fill: FillWithBlock) => filledSameDeposit(_fill, fill))
   );
+  let firstFillForSameDeposit = allMatchingFills.find((_fill) => isFirstFillForDeposit(_fill));
 
-  // If there is no first fill for the same deposit, then throw an error.
-  if (firstFillForSameDeposit === undefined) {
```
nicholaspai, line 114: The main benefit of this PR is that this error can now be resolved by trying to load historical fills instead of default throwing if a fill can't be found

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -171,66 +173,69 @@ export function subtractExcessFromPreviousSlowFillsFromRunningBalances(
   // This assumption depends on the rule that slow fills can only be sent after a partial fill for a non zero amount
   // of the deposit. This is why "1 wei" fills are important, otherwise we'd never know which fills originally
   // triggered a slow fill payment to be sent to the destination chain.
-  allValidFillsInRange
-    .filter((fill) => fill.totalFilledAmount.eq(fill.amount) && !fill.fillAmount.eq(fill.amount))
-    .forEach((fill: interfaces.FillWithBlock) => {
-      const { lastMatchingFillInSameBundle, rootBundleEndBlockContainingFirstFill } =
-        getFillDataForSlowFillFromPreviousRootBundle(
+  await Promise.all(
```
nicholaspai, line 176: the only diff here is adding an `await Promise.all` around the map and `await`-ing the `getFillDataForSlowFillFromPreviousRootBundle` function which is now async

src/clients/SpokePoolClient.ts
```diff
@@ -350,6 +352,51 @@ export class SpokePoolClient {
...
+        undefined,
+        undefined,
+        undefined,
+        matchingFill.depositor,
+        undefined,
+        undefined
+      ),
+      searchConfig
+    );
+    // Should we cache this query? Will it be automatically cached by https://github.com/across-protocol/relayer-v2/pull/440 ?
```
mrice32, line 395: I think the provider caching will perform pretty well for this. It won't be perfect since the end block will likely not align on each run, but it will cache the majority of the query after the first query. The cache will go stale after an hour. So we'll ultimately do the full query two or three times when this does happen since the bots will need to repeatedly check the info when the bundle is being validated.

My suggestion would be to not worry about caching here to keep complexity low and because the provider caching will work pretty well.

src/utils/FillUtils.ts
```diff
@@ -98,24 +106,43 @@ export function getLastMatchingFillBeforeBlock(
...
+  // If `allValidFills` doesn't contain the first fill for this deposit then we have to perform a historical query to
+  // find it. This is inefficient, but should be rare. Save any other fills we find to the
+  // allMatchingFills array, at the end of this block, allMatchingFills should contain all fills for the same
+  // deposit as the input fill.
+  if (!firstFillForSameDeposit) {
+    const matchingFills = await spokePoolClientsByChain[fill.destinationChainId].queryHistoricalMatchingFills(
+      fill,
+      allMatchingFills[0].blockNumber
+    );
+    firstFillForSameDeposit = sortEventsAscending(matchingFills).find((_fill) => isFirstFillForDeposit(_fill));
```
mrice32, line 136: Technically, this first fill (and the other fills returned in the query) could be _invalid_, but it doesn't seem like we do any validation and only match on the id and origin chain. Is that a problem? It may not be! I just want to double check that we don't need to validate that _all_ the parameters match the input fill.

mrice32 (commented):
Looks great! Only one question

This is super clean and simple!

src/clients/SpokePoolClient.ts
nicholaspai: I agree thanks for looking here! I wanted to confirm my instinct

src/utils/FillUtils.ts
nicholaspai: Ahhh this is a great catch, fixed in this commit. I think this would have been a problem otherwise we could compute slow fill excesses incorrectly: https://github.com/across-protocol/relayer-v2/pull/442/commits/731fb0b8585c4b8887ccd64f56bd2bbba1b532ea

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
```diff
@@ -1249,225 +1252,225 @@ export class Dataworker {
...
-
-        // Filter out roots that are not in the latest N root bundles. This assumes that
-        // relayerRefundRoot+slowFillRoot combinations are unique.
-        rootBundleRelays = this._getRelayedRootsFromBundles(latestRootBundles, rootBundleRelays);
-
-        // Filter out empty relayer refund root:
-        rootBundleRelays = rootBundleRelays.filter((rootBundle) => rootBundle.relayerRefundRoot !== EMPTY_MERKLE_ROOT);
+    // Execute each chain's leaves sequentially to take advantage of pool rebalance root caching. If we executed
+    // each chain in parallel, then we'd have to reconstruct identical pool rebalance root more times than necessary.
+    for (const [_chainId, client] of Object.entries(spokePoolClients)) {
```
nicholaspai, line 1257: Change here to execute chains sequentially rather than in parallel

PR #443, opened 2023-02-03
improve(RedisUtils): Add delay to cache historical deposits

nicholaspai:
Mitigates chance of caching deposit that gets re-orged

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -350,7 +350,14 @@ export class SpokePoolClient {
         deposit,
       });
       if (this.configStoreClient.redisClient)
-        await setDeposit(deposit, this.configStoreClient.redisClient, 24 * 60 * 60);
+        await setDeposit(
+          deposit,
+          (
+            await this.spokePool.provider.getBlock("latest")
+          ).timestamp,
+          this.configStoreClient.redisClient,
+          24 * 60 * 60
+        );
     }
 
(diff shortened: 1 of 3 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -343,7 +343,14 @@ export class SpokePoolClient {
         deposit,
       });
       if (this.configStoreClient.redisClient)
-        await setDeposit(deposit, this.configStoreClient.redisClient, 24 * 60 * 60);
+        await setDeposit(
+          deposit,
+          (
+            await this.spokePool.provider.getBlock("latest")
```
mrice32, line 349: Why not use the current system timestamp?

mrice32 (commented):
Looks good! Just one question

nicholaspai:
@mrice32 do you think we even need to cache this deposit now that we have the provider cache?

src/clients/SpokePoolClient.ts
nicholaspai: done

mrice32:
> @mrice32 do you think we even need to cache this deposit now that we have the provider cache? If not I can open a new PR to remove deposit caching and reduce complexity

I think we do, mainly because when we binary search blocks, we always take a different path, so I suspect that many of the blocks will not be cached on each run. So caching the result avoids the search entirely. We also don't cache eth_getBlockByNumber or eth_call right now, so removing it today would result in us not caching this at all.

However, this isn't a super strong opinion, I just suspect that this caching is adding value.

Thoughts?

mrice32 (approved):

PR #444, opened 2023-02-03
improve: Dump mismatched RPC provider responses

pxrl:
Some RPC provider errors that we see are transient, so it's hard to debug them any further unless we capture the actual response that was issued.

This will generate fairly lengthy logging when we do have provider issues, but will also enable us to diagnose and report errors much more quickly.

Fixes ACX-755.


mrice32 (approved):
LGTM!

PR #446, opened 2023-02-06
improve(SpokePoolClient): Include network names in exception

pxrl:
The existing message logs only the deposit ID and the block range that was searched.


pxrl:
Existing error:

`error: Error: Could not find deposit ID 228137 for fill between blocks [72642091, 72642152]`

mrice32 (approved):

PR #449, opened 2023-02-10
improve: If `spokeRootsLookbackCount` is set, only try to execute root bundles from N most recent proposed root bundles rather than root bundle relays

nicholaspai:
Imagine we only want to execute leaves from the last 10 bundles. Intuitively this should only include leaves from the last 10 root bundles, but currently this means we count the latest 10 relayed root bundles for a chain. This means that some chains like Boba which have infrequent relayed root bundles, the lookback needs to be much larger.

This results in a lot of warnings: "Cannot validate bundle with insufficient event data. Set a larger DATAWORKER_FAST_LOOKBACK_COUNT". We should quiet these warnings

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -811,7 +811,7 @@ export class Dataworker {
         let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter((rootBundle) =>
           isKeyOf(chainId, IGNORED_SPOKE_BUNDLES)
             ? !IGNORED_SPOKE_BUNDLES[chainId].includes(rootBundle.rootBundleId)
-            : true && rootBundle.blockNumber >= client.eventSearchConfig.fromBlock
+            : rootBundle.blockNumber >= client.eventSearchConfig.fromBlock
         );
 
         // Filter out roots that are not in the latest N root bundles. This assumes that
@@ -1253,7 +1253,7 @@ export class Dataworker {
         let rootBundleRelays = sortEventsDescending(client.getRootBundleRelays()).filter((rootBundle) =>
           isKeyOf(chainId, IGNORED_SPOKE_BUNDLES)
             ? !IGNORED_SPOKE_BUNDLES[chainId].includes(rootBundle.rootBundleId)
-            : true && rootBundle.blockNumber >= client.eventSearchConfig.fromBlock
+            : rootBundle.blockNumber >= client.eventSearchConfig.fromBlock
         );
 
         // Filter out roots that are not in the latest N root bundles. This assumes that
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1245,8 +1245,9 @@ export class Dataworker {
     });
 
     let latestRootBundles = sortEventsDescending(this.clients.hubPoolClient.getValidatedRootBundles());
-    if (this.spokeRootsLookbackCount !== 0)
+    if (this.spokeRootsLookbackCount !== 0) {
       latestRootBundles = latestRootBundles.slice(0, this.spokeRootsLookbackCount);
+    }
 
     await Promise.all(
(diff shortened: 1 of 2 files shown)


src/scripts/validateRootBundle.ts
```diff
@@ -135,7 +135,8 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet): Pro
   // We want the bundle end of blocks following the target bundle so add +1 to the index.
   const overriddenConfig = {
     ...config,
-    dataworkerFastStartBundle: closestFollowingValidatedBundleIndex + 1,
+    dataworkerFastStartBundle:
+      closestFollowingValidatedBundleIndex === -1 ? "latest" : closestFollowingValidatedBundleIndex + 1,
```
nicholaspai, line 139: Unrelated fix to this script if trying to validate bundle that is too recent and there no validated bundles after it

src/dataworker/Dataworker.ts
```diff
@@ -800,18 +801,22 @@ export class Dataworker {
       message: "Executing slow relay leaves",
     });
 
+    let latestRootBundles = sortEventsDescending(this.clients.hubPoolClient.getValidatedRootBundles());
+    if (this.spokeRootsLookbackCount !== 0)
+      latestRootBundles = latestRootBundles.slice(0, this.spokeRootsLookbackCount);
```
james-a-morris, line 806: Nit: for readability, should we surround this statement with brackets?
```suggestion
    if (this.spokeRootsLookbackCount !== 0) {
      latestRootBundles = latestRootBundles.slice(0, this.spokeRootsLookbackCount);
    }
```

src/dataworker/Dataworker.ts
```diff
@@ -1238,18 +1243,22 @@ export class Dataworker {
       message: "Executing relayer refund leaves",
     });
 
+    let latestRootBundles = sortEventsDescending(this.clients.hubPoolClient.getValidatedRootBundles());
+    if (this.spokeRootsLookbackCount !== 0)
+      latestRootBundles = latestRootBundles.slice(0, this.spokeRootsLookbackCount);
```
james-a-morris, line 1248: Nit: same as above

src/dataworker/Dataworker.ts
```diff
@@ -1564,4 +1573,20 @@ export class Dataworker {
     if (leaf.groupIndex === 0) requiredAmount = requiredAmount.add(toBNWei("0.02"));
     return requiredAmount;
   }
+
+  // Filters out any root bundles that don't have a matching relayerRefundRoot+slowRelayRoot combination in the
+  // list of proposed root bundles `allRootBundleRelays`.
```
james-a-morris, line 1578: Nit: should we include valid JSDoc comments here so that IDE users have a full representation of this function?

src/dataworker/Dataworker.ts
```diff
@@ -1564,4 +1573,20 @@ export class Dataworker {
...
+    rootBundles: ProposedRootBundle[],
+    allRootBundleRelays: RootBundleRelayWithBlock[]
+  ): RootBundleRelayWithBlock[] {
+    return allRootBundleRelays.filter(
+      (rootBundle) =>
+        rootBundles.find(
+          (_rootBundle) =>
+            _rootBundle.relayerRefundRoot === rootBundle.relayerRefundRoot &&
+            _rootBundle.slowRelayRoot === rootBundle.slowRelayRoot
+        ) !== undefined
```
james-a-morris, line 1589: Nit: semantically, would it make more sense to use `.some` instead of `.find` given that we're only interested in the boolean output?

james-a-morris (commented):
Left some nits. This is a really interesting modification 🙂

james-a-morris (approved):
LGTM! Thanks for thoroughly reviewing my comments. Great work 😎

mrice32 (approved):
LGTM!

PR #452, opened 2023-02-16
feat(ProfitClient): Permit route-based minimum relayer fees

pxrl:
This change permits relayer operators to specify source- & destination-specific minimum
fees per token. The existing global config is retained as a fallback for any combinations
that do not match on a more specific configuration.

Also, make isFillProfitable(..., ..., l1Token, ...) a mandatory argument, since the only caller
of this method always supplies l1token.

Fixed ACX-850.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -189,8 +189,6 @@ export class ProfitClient {
       `Unsupported destination chain ID: ${deposit.destinationChainId}`
     );
 
-    minRelayerFeePct ??= this.minRelayerFeePct(l1Token.symbol, deposit.originChainId, deposit.destinationChainId);
-
     const tokenPriceUsd = this.getPriceOfToken(l1Token.address);
     if (tokenPriceUsd.lte(0)) throw new Error(`Unable to determine ${l1Token.symbol} L1 token price`);
 
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -58,7 +58,7 @@ const { acrossApi, coingecko, defiLlama } = priceClient.adapters;
 
 export class ProfitClient {
   private readonly priceClient;
-  protected minRelayerFees: { [symbol: string]: { [dstChainId: number]: { [srcChainId: number]: BigNumber } } } = {};
+  protected minRelayerFees: { [route: string]: BigNumber } = {};
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
   private unprofitableFills: { [chainId: number]: { deposit: Deposit; fillAmount: BigNumber }[] } = {};
 
@@ -164,7 +164,8 @@ export class ProfitClient {
   // 0.1bps on USDC from Optimism to Arbitrum:
   //   - MIN_RELAYER_FEE_PCT_USDC_42161_10=0.00001
   minRelayerFeePct(token: string, srcChainId: number, dstChainId: number): BigNumber {
-    let minRelayerFeePct: BigNumber | undefined = this.minRelayerFees[token]?.[dstChainId]?.[srcChainId];
+    const routeKey = `${token}_${dstChainId}_${srcChainId}`;
+    let minRelayerFeePct = this.minRelayerFees[routeKey];
 
     if (!minRelayerFeePct) {
       const tokenOverride = `MIN_RELAYER_FEE_PCT_${token}`;
@@ -174,15 +175,10 @@ export class ProfitClient {
         ?? process.env[`${tokenOverride}_${dstChainId}`]
         ?? process.env[tokenOverride];
 
+      minRelayerFeePct = _minRelayerFeePct ? toBNWei(_minRelayerFeePct) : this.defaultRelayerFeePct;
+
       // Save the route for next time.
-      if (!this.minRelayerFees[token]) this.minRelayerFees[token] = {};
-      if (!this.minRelayerFees[token][dstChainId]) this.minRelayerFees[token][dstChainId] = {};
-
-      this.minRelayerFees[token][dstChainId][srcChainId] = _minRelayerFeePct
-        ? toBNWei(_minRelayerFeePct)
-        : this.defaultRelayerFeePct;
-
-      minRelayerFeePct = this.minRelayerFees[token][dstChainId][srcChainId];
+      this.minRelayerFees[routeKey] = minRelayerFeePct;
     }
 
     return minRelayerFeePct as BigNumber;
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -58,6 +58,7 @@ const { acrossApi, coingecko, defiLlama } = priceClient.adapters;
 
 export class ProfitClient {
   private readonly priceClient;
+  protected minRelayerFees: { [symbol: string]: { [dstChainId: number]: { [srcChainId: number]: BigNumber } } } = {};
   protected tokenPrices: { [l1Token: string]: BigNumber } = {};
   private unprofitableFills: { [chainId: number]: { deposit: Deposit; fillAmount: BigNumber }[] } = {};
 
@@ -74,7 +75,7 @@ export class ProfitClient {
     spokePoolClients: SpokePoolClientsByChain,
     readonly ignoreProfitability: boolean,
     readonly enabledChainIds: number[],
-    readonly minRelayerFeePct: BigNumber = toBNWei(constants.RELAYER_MIN_FEE_PCT),
+    readonly defaultRelayerFeePct: BigNumber = toBNWei(constants.RELAYER_MIN_FEE_PCT),
     readonly debugProfitability: boolean = false,
     protected gasMultiplier: BigNumber = toBNWei(1)
   ) {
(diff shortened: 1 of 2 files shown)


src/clients/ProfitClient.ts
```diff
@@ -153,20 +154,59 @@ export class ProfitClient {
...
+  // - Token
+  // - Bridge route (dst/src chain Ids + token)
+  // 0.8bps on USDC global default:
+  //   - MIN_RELAYER_FEE_PCT_USDC=0.00008
+  // 0.2bps on USDC from any to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161=0.00002
+  // 0.1bps on USDC from Optimism to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161_10=0.00001
+  minRelayerFeePct(token: string, srcChainId: number, dstChainId: number): BigNumber {
+    let minRelayerFeePct: BigNumber | undefined = this.minRelayerFees[token]?.[dstChainId]?.[srcChainId];
```
nicholaspai, line 167: nit: the type of `minRelayerFeePct` should be automatically inferred no?

src/clients/ProfitClient.ts
```diff
@@ -58,6 +58,7 @@ const { acrossApi, coingecko, defiLlama } = priceClient.adapters;
 
 export class ProfitClient {
   private readonly priceClient;
+  protected minRelayerFees: { [symbol: string]: { [dstChainId: number]: { [srcChainId: number]: BigNumber } } } = {};
```
nicholaspai, line 61: instead of having a double nested object, why not use a `key` where `key = "${src}-${dst}"`? Could even make this a flat object by making `key="${token}-${src}-${dst}"`

src/clients/ProfitClient.ts
```diff
@@ -153,20 +154,59 @@ export class ProfitClient {
...
+  // 0.8bps on USDC global default:
+  //   - MIN_RELAYER_FEE_PCT_USDC=0.00008
+  // 0.2bps on USDC from any to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161=0.00002
+  // 0.1bps on USDC from Optimism to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161_10=0.00001
+  minRelayerFeePct(token: string, srcChainId: number, dstChainId: number): BigNumber {
+    let minRelayerFeePct: BigNumber | undefined = this.minRelayerFees[token]?.[dstChainId]?.[srcChainId];
+
+    if (!minRelayerFeePct) {
```
nicholaspai, line 169: Does this correctly handle when `minRelayerFeePct == 0`? Is it safer to check for `== undefined`?

src/clients/ProfitClient.ts
```diff
@@ -153,20 +154,59 @@ export class ProfitClient {
...
+    minRelayerFeePct?: BigNumber
+  ): FillProfit {
     assert(fillAmount.gt(0), `Unexpected fillAmount: ${fillAmount}`);
     assert(
       Object.keys(GAS_TOKEN_BY_CHAIN_ID).includes(deposit.destinationChainId.toString()),
       `Unsupported destination chain ID: ${deposit.destinationChainId}`
     );
 
-    l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
-    assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
```
nicholaspai, line 169: why remove this assert?

nicholaspai (commented):
I think this is a very useful PR and we should make sure to bump the config store version to prod other relayers to use it, (although its probably not a breaking change, rather a minor version bump).

I agree with you that a minRelayerFee per token should come first in priority. Do you think we need token-destination chain level granularity, or is it good enough to set priority as follows:
- per token and route
- per token
- per route

that might be simpler?. So for any deposit, if the token+route has an overridden min fee, then use that. Otherwise, if the token has an overidden min fee, then use that. Lastly, use the route.

src/clients/ProfitClient.ts
pxrl: Yeah - minRelayerFeePct is a BN, so it will either resolve to a BN object, or undefined. The content of the BN doesn't have any effect.

src/clients/ProfitClient.ts
pxrl: Yeah, fair enough :+1: Fixed: https://github.com/across-protocol/relayer-v2/pull/452/commits/313602f890a56a246eed7f5fdebb80333ab5350c.

pxrl:
> I think this is a very useful PR and we should make sure to bump the config store version to prod other relayers to use it, (although its probably not a breaking change, rather a minor version bump).
> 
> I agree with you that a minRelayerFee per token should come first in priority. Do you think we need token-destination chain level granularity, or is it good enough to set priority as follows:
> 
> * per token and route
> * per token
> * per route
> 
> that might be simpler?. So for any deposit, if the token+route has an overridden min fee, then use that. Otherwise, if the token has an overidden min fee, then use that. Lastly, use the route.

Hm, so PR as-is does no support specifying a destination-specific minRelayerFeePct, nor a route-based minRelayerFeePct that ignores the token.

I think that a destination-specific minRelayerFeePct could be useful, but having that _and_ a token-specific minRelayerFeePct starts to get confusing to configure (i.e. in case where MIN_RELAYER_FEE_PCT_USDC and MIN_RELAYER_FEE_PCT_42161 are set, the operator must know which takes precedence.

The alternative is just to do the following - and I think we can probably live with this, since we are mostly making adjustments to the "big 4" tokens:
 - MIN_RELAYER_FEE_PCT_USDC_42161=x 
 - MIN_RELAYER_FEE_PCT_WETH_42161=x
 - MIN_RELAYER_FEE_PCT_WBTC_42161=x
 - MIN_RELAYER_FEE_PCT_DAI_42161=x

So, if we support per-destination configuration, I'd probably prefer to drop the global per-token configuration. What do you think is most useful?

Token-agnostic route-based configuration is similarly not supported as-is. I'm not sure that we really need it, since in practice it removes flexibility, and the adjustments made so far have been to individual token/route combinations.

nicholaspai:
> > I think this is a very useful PR and we should make sure to bump the config store version to prod other relayers to use it, (although its probably not a breaking change, rather a minor version bump).
> > I agree with you that a minRelayerFee per token should come first in priority. Do you think we need token-destination chain level granularity, or is it good enough to set priority as follows:
> > 
> > * per token and route
> > * per token
> > * per route
> > 
> > that might be simpler?. So for any deposit, if the token+route has an overridden min fee, then use that. Otherwise, if the token has an overidden min fee, then use that. Lastly, use the route.
> 
> Hm, so PR as-is does not support specifying a destination-specific minRelayerFeePct, nor a token-agnostic route-based minRelayerFeePct.
> 
> I think that a destination-specific minRelayerFeePct could be useful, but having that _and_ a token-specific minRelayerFeePct starts to get confusing to configure. For example, in case where MIN_RELAYER_FEE_PCT_USDC and MIN_RELAYER_FEE_PCT_42161 are set, the operator must know which takes precedence.
> 
> The alternative is just to do the following - and I think we can probably live with this, since we are mostly making adjustments to the "big 4" tokens:
> 
> * MIN_RELAYER_FEE_PCT_USDC_42161=x
> * MIN_RELAYER_FEE_PCT_WETH_42161=x
> * MIN_RELAYER_FEE_PCT_WBTC_42161=x
> * MIN_RELAYER_FEE_PCT_DAI_42161=x
> 
> So, if we support per-destination configuration, I'd probably prefer to drop the global per-token configuration. What do you think is most useful?
> 
> Token-agnostic route-based configuration is similarly not supported as-is. I'm not sure that we really need it, since in practice it removes flexibility, and the adjustments made so far have been to individual token/route combinations.

Why not make it stupid simple and make .env setter specify the full token_srcChain_destChain or have to use the default?

src/clients/ProfitClient.ts
```diff
@@ -153,20 +154,55 @@ export class ProfitClient {
...
+  // - Token
+  // - Bridge route (dst/src chain Ids + token)
+  // 0.8bps on USDC global default:
+  //   - MIN_RELAYER_FEE_PCT_USDC=0.00008
+  // 0.2bps on USDC from any to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161=0.00002
+  // 0.1bps on USDC from Optimism to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161_10=0.00001
+  minRelayerFeePct(token: string, srcChainId: number, dstChainId: number): BigNumber {
+    const routeKey = `${token}_${dstChainId}_${srcChainId}`;
```
james-a-morris, line 167: Nit: should we be mindful of token casing?

james-a-morris (commented):
Left a few nits.

A question I thought of during this which could apply to all Across repos: Would it make sense to create enumerated types for tokens & chainIds? There's a decent amount of assumptions that are made when passing tokens as arbitrary strings that would be reduced with enumerators. In a similar way, adding tokens & chains is an intentional action and rarely a dynamic process. cc @nicholaspai @mrice32

src/clients/ProfitClient.ts
nicholaspai: hmm thats a good point. Usually the token should be assumed to be in checksum, but it doesn't hurt to always enforce that. The only hesitancy is that we should be checking for checksum everywhere a token is mentioned. Maybe its better to make sure that `minRelayerFeePct` is only called internally where we know `token` is being passed in as a checksummed address?

pxrl: `token` is actually a symbol, not a string. If we had internally inconsistent naming for token symbols then I'd be concerned, but the token info is queried directly from the HubPool so I think we can rely on the symbols being upper case.

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -280,6 +280,41 @@ describe("ProfitClient: Consider relay profit", async function () {
...
+            if (minRelayerFeePct) {
+              spyLogger.debug({
+                message: `Expect relayerFeePct === ${minRelayerFeePct}`,
+                envVar,
+                symbol,
+                dstChainId,
+                srcChainId,
+              });
+              const computedMinRelayerFeePct = profitClient.minRelayerFeePct(symbol, srcChainId, dstChainId);
+              expect(computedMinRelayerFeePct.eq(minRelayerFeePct as BigNumber)).to.be.true;
```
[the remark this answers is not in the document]
pxrl: My intention was to describe it in the Across relayer docs at docs.across.to. .env.sample is pretty sparse at the moment; I have an old draft PR that I need to dust off, so I can also capture this as part of that.

src/clients/ProfitClient.ts
pxrl: I renamed `token` => `symbol` here: https://github.com/across-protocol/relayer-v2/pull/452/commits/a0bd14cc587933de3a7c9ae9b66d7792c03d2bd5.

pxrl:
> > > I think this is a very useful PR and we should make sure to bump the config store version to prod other relayers to use it, (although its probably not a breaking change, rather a minor version bump).
> > > I agree with you that a minRelayerFee per token should come first in priority. Do you think we need token-destination chain level granularity, or is it good enough to set priority as follows:
> > > 
> > > * per token and route
> > > * per token
> > > * per route
> > > 
> > > that might be simpler?. So for any deposit, if the token+route has an overridden min fee, then use that. Otherwise, if the token has an overidden min fee, then use that. Lastly, use the route.
> > 
> > 
> > Hm, so PR as-is does not support specifying a destination-specific minRelayerFeePct, nor a token-agnostic route-based minRelayerFeePct.
> > I think that a destination-specific minRelayerFeePct could be useful, but having that _and_ a token-specific minRelayerFeePct starts to get confusing to configure. For example, in case where MIN_RELAYER_FEE_PCT_USDC and MIN_RELAYER_FEE_PCT_42161 are set, the operator must know which takes precedence.
> > The alternative is just to do the following - and I think we can probably live with this, since we are mostly making adjustments to the "big 4" tokens:
> > 
> > * MIN_RELAYER_FEE_PCT_USDC_42161=x
> > * MIN_RELAYER_FEE_PCT_WETH_42161=x
> > * MIN_RELAYER_FEE_PCT_WBTC_42161=x
> > * MIN_RELAYER_FEE_PCT_DAI_42161=x
> > 
> > So, if we support per-destination configuration, I'd probably prefer to drop the global per-token configuration. What do you think is most useful?
> > Token-agnostic route-based configuration is similarly not supported as-is. I'm not sure that we really need it, since in practice it removes flexibility, and the adjustments made so far have been to individual token/route combinations.
> 
> Why not make it stupid simple and make .env setter specify the full token_srcChain_destChain or have to use the default?

Have restricted it to the fully-qualified symbol/route combination here: https://github.com/across-protocol/relayer-v2/pull/452/commits/a0bd14cc587933de3a7c9ae9b66d7792c03d2bd5.

src/clients/ProfitClient.ts
```diff
@@ -153,20 +154,43 @@ export class ProfitClient {
...
+  ): FillProfit {
     assert(fillAmount.gt(0), `Unexpected fillAmount: ${fillAmount}`);
     assert(
       Object.keys(GAS_TOKEN_BY_CHAIN_ID).includes(deposit.destinationChainId.toString()),
       `Unsupported destination chain ID: ${deposit.destinationChainId}`
     );
 
-    l1Token ??= this.hubPoolClient.getTokenInfoForDeposit(deposit);
-    assert(l1Token !== undefined, `No L1 token found for deposit ${JSON.stringify(deposit)}`);
+    minRelayerFeePct ??= this.minRelayerFeePct(l1Token.symbol, deposit.originChainId, deposit.destinationChainId);
```
nicholaspai, line 192: I'm not too familiar with the ??= syntax. Could minRelayerFeePct ever be undefined? It should always be defined right

nicholaspai (approved):
One comment then LGTM!

src/clients/ProfitClient.ts
pxrl: ??= seems pretty useful and we could probably use it a bit more.

In any case, minRelayerFeePct was only undefined when called from tests, but I just changed that here: https://github.com/across-protocol/relayer-v2/pull/452/commits/738a118ad57fac607cb9660cc987c06a0508c243.

PR #453, opened 2023-02-17
fix(dataworker): amount to send to SpokePool for slow fill should be unfilled amount minus LP fee

nicholaspai:
When a slow fill is executed, its `unfilledAmount` is sent to the `recipient` minus any fees where `relayerFeePct == 0` and `lpFeePct >= 0`. Therefore, this fixes an issue where every slow fill executed is leaving the `lpFee` in the SpokePool.

I discovered this bug while running the `validateRunningBundle` script and noticed that the `excess` amount increases after slow fill. This is corroborated by looking at spoke pool balances over time, for example the BAL balance of the EtheruemSpokePool which recently had some slow fills: https://etherscan.io/token/0xba100000625a3754423978a60c9317c58a424e3d?a=0x4d9079bb4165aeb4084c526a32695dcfd2f77381#tokenAnalytics

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -206,8 +206,7 @@ export class BundleDataClient {
       allValidFills.push(fillWithBlock);
 
       // If fill is outside block range, we can skip it now since we're not going to add a refund for it.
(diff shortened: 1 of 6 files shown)


mrice32 (approved):
LGTM!

PR #455, opened 2023-02-17
fix: rework ci to be specific about node versions

mrice32:
This reworks the ci in the repo to be more specific about the node versions that are used. This will also allow us to technically add more versions in the future via the matrix feature (which we were using incorrectly before).

On a related note: this condenses the workflows into fewer files since lint and test were triggered on the same action.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
+name: Check PR
```
mrice32, line 1: This just condenses the old lint and test workflows since they have the same trigger.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
...
+          cache: "yarn"
+
+      - name: Lint
+        run: yarn lint
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [16]
```
mrice32, line 27: This strategy configuration is what we were missing before. We were using matrix.node to specify versions, but it was empty because we were missing this config.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Use Node ${{ matrix.node }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+          cache: "yarn"
```
mrice32, line 18: yarn caching was added.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [16]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Use Node ${{ matrix.node }}
+        uses: actions/setup-node@v3
```
mrice32, line 15: This was upgraded to v3.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [16]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
```
mrice32, line 12: This was also upgraded to v3.

evaldofelipe: nice

evaldofelipe (approved):
LGTM!

nicholaspai (approved):

PR #457, opened 2023-02-17
chore(ProfitClient): Remove ignoreProfitability

pxrl:
ignoreProfitability has not been used since late October, and it was marked
in the code as a temporary feature. Now that we have minimum fee
granularity at the per token/route level, it seems like can probably do
without this config item.
    
For reference, it should still be possible to emulate ignoreProfitability by
setting GAS_MULTIPLIER=0.01 to ignore the cost of a making fills, and then
and tuning any token/route-specific minimums as necessary.

Fixes ACX-919.


nicholaspai (approved):

src/clients/ProfitClient.ts
```diff
@@ -153,20 +153,41 @@ export class ProfitClient {
...
+
+    if (!minRelayerFeePct) {
+      const _minRelayerFeePct = process.env[`MIN_RELAYER_FEE_PCT_${routeKey}`];
+      minRelayerFeePct = _minRelayerFeePct ? toBNWei(_minRelayerFeePct) : this.defaultMinRelayerFeePct;
+
+      // Save the route for next time.
+      this.minRelayerFees[routeKey] = minRelayerFeePct;
+    }
+
+    return minRelayerFeePct as BigNumber;
```
mrice32, line 171: Why does this need to be cast? Shouldn't ts know it's a BigNumber?

mrice32, line 171: Optional suggestion to streamline this flow a little.

```suggestion
    if (!this.minRelayerFees[routeKey]) {
      const _minRelayerFeePct = process.env[`MIN_RELAYER_FEE_PCT_${routeKey}`];
      this.minRelayerFees[routeKey] = _minRelayerFeePct ? toBNWei(_minRelayerFeePct) : this.defaultMinRelayerFeePct;
    }
    
    return this.minRelayerFees[routeKey];
```

src/clients/ProfitClient.ts
```diff
@@ -153,20 +153,41 @@ export class ProfitClient {
     this.unprofitableFills = {};
   }
 
+  // Allow the minimum relayer fee to be overridden per token/route:
+  // 0.1bps on USDC from Optimism to Arbitrum:
+  //   - MIN_RELAYER_FEE_PCT_USDC_42161_10=0.00001
+  minRelayerFeePct(symbol: string, srcChainId: number, dstChainId: number): BigNumber {
```
mrice32, line 159: Is this change ralted to ignoreProfitability?

mrice32 (approved):
LGTM!

src/clients/ProfitClient.ts
pxrl: Nope - this PR only focuses on removing ignoreProfitability, but I think a Github hiccup has meant that it's showing additional changes.

This PR was building on top of https://github.com/across-protocol/relayer-v2/commit/759351e66c4b4cbb8b332fd688002e0dedd947e2, and was initially set to merge into that branch (to only show the delta, which was the ignoreProfitability changes). Github _should_ have auto-updated to merge into master when that PR was merged, but for some reason that didn't work (maybe due to failing CI?).

pxrl:
Just to clarify the situation here - these tests pass on my local machine, but fail consistently in CI. They seem to be failing due to log messages that are supposedly not being emitted. I haven't yet understood why these tests behave differently on my local machine vs. CI, but am working on it.

PR #461, opened 2023-02-23
feat: Allow gas fee scaler's to be settable via env variables

nicholaspai:
So we can experiment with relayer gas pricing

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -20,6 +20,7 @@ export async function runTransaction(
   gasLimit: BigNumber | null = null,
   nonce: number | null = null
 ): Promise<TransactionResponse> {
+  const chainId = (await contract.provider.getNetwork()).chainId;
   try {
     const priorityFeeScaler =
       Number(process.env[`PRIORITY_FEE_SCALER_${chainId}`] || process.env.PRIORITY_FEE_SCALER) || undefined;
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -21,14 +21,12 @@ export async function runTransaction(
   nonce: number | null = null
 ): Promise<TransactionResponse> {
   try {
-    const priorityFeeScaler = Number(process.env[`PRIORITY_FEE_SCALER_${chainId}`] || process.env.PRIORITY_FEE_SCALER) || undefined;
-    const maxFeePerGasScaler = Number(process.env[`MAX_FEE_PER_GAS_SCALER_${chainId}`] || process.env.MAX_FEE_PER_GAS_SCALER) || undefined;
+    const priorityFeeScaler =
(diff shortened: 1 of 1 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -20,7 +21,11 @@ export async function runTransaction(
   nonce: number | null = null
 ): Promise<TransactionResponse> {
   try {
-    const gas = await getGasPrice(contract.provider);
+    const gas = await getGasPrice(
+      contract.provider,
+      process.env.PRIORITY_FEE_SCALER && Number(process.env.PRIORITY_FEE_SCALER),
+      process.env.MAX_FEE_PER_GAS_SCALER && Number(process.env.MAX_FEE_PER_GAS_SCALER)
+    );
```
mrice32, line 28: Thoughts on something like this where we have a chain-specific one and then an omnichain one to fall back to? That way we can tune it for each chain if we like. Something like:

```suggestion
    const priorityFeeScaler = Number(process.env[`PRIORITY_FEE_SCALER_${chainId}`] || process.env.PRIORITY_FEE_SCALER) || undefined;
    const maxFeePerGasScaler = Number(process.env[`MAX_FEE_PER_GAS_SCALER_${chainId}`] || process.env.MAX_FEE_PER_GAS_SCALER) || undefined;

    const gas = await getGasPrice(
      contract.provider,
      priorityFeeScaler,
      maxFeePerGasScaler
    );
```

mrice32 (commented):
Just one minor comment!

src/utils/TransactionUtils.ts
pxrl: Yeah it'd be useful for this to be per-chain configurable.

nicholaspai: Great idea!

src/utils/TransactionUtils.ts
```diff
@@ -20,7 +21,12 @@ export async function runTransaction(
   nonce: number | null = null
 ): Promise<TransactionResponse> {
   try {
-    const gas = await getGasPrice(contract.provider);
+    const priorityFeeScaler =
+      Number(process.env[`PRIORITY_FEE_SCALER_${chainId}`] || process.env.PRIORITY_FEE_SCALER) || undefined;
```
pxrl, line 25: I think you can fish the chainId out from contract.provider.getNetwork().chainId (or something similar).

nicholaspai: yeah this adds an RPC request but `runTransaction` is once per run so i think this is OK, WDYT?

nicholaspai: Alternatively we could try to load in chain ID from calling code where possible (sometimes its already loaded) and then add an RPC request where its not

pxrl: We use StaticJsonRpcProvider so the lookup just returns ethers' local copy of the chainId - so there should be no cost to this :+1:

pxrl (approved):

PR #462, opened 2023-02-27
improve(TxUtils): Should try to resubmit txn when encountering certain errors

nicholaspai:
For  nonce collisions and gas price estimation errors, bots should try to resubmit

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -18,9 +18,11 @@ export async function runTransaction(
   args: any,
   value: BigNumber = toBN(0),
   gasLimit: BigNumber | null = null,
-  nonce: number | null = null
+  nonce: number | null = null,
+  retriesRemaining = 2
 ): Promise<TransactionResponse> {
   const chainId = (await contract.provider.getNetwork()).chainId;
+
   try {
     const priorityFeeScaler =
       Number(process.env[`PRIORITY_FEE_SCALER_${chainId}`] || process.env.PRIORITY_FEE_SCALER) || undefined;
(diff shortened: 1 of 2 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -47,14 +47,13 @@ export async function runTransaction(
...
-      // If error is `insufficient funds for intrinsic transaction cost` the gas price is probably spiking, just
-      // resubmit to get the new gas price.
-      return await runTransaction(logger, contract, method, args, value, gasLimit, nonce);
+    if (
+      error?.code === "NONCE_EXPIRED" ||
+      error?.code === "INSUFFICIENT_FUNDS" ||
+      error?.message.includes("intrinsic gas too low")
+    ) {
+      // If error is due to a nonce collision or gas underpricement then re-submit to fetch latest params.
+      return await runTransaction(logger, contract, method, args, value, gasLimit, null);
```
pxrl, line 56: In general I think we should protect against the potential for endless recursion here. It seems like there's a scenario where a txn could pass the initial simulation, but then end up in a recursive INSUFFICIENT_FUNDS loop due to a gas price spike coinciding with insufficient funds to pay the new gas price.

One possible way of mitigating that would be to modify the input arguments to runTransaction(), such that it accepts a counter that is decremented on subsequent calls into runTransaction(), or a boolean flag that determines whether it can recursively call down to runTransaction().

```
      return await runTransaction(logger, contract, method, args, value, gasLimit, null, --permitRetry);
```

nicholaspai: Good call, i implemented with a default retries number we could configure in the future

src/utils/TransactionUtils.ts
```diff
@@ -47,13 +49,31 @@ export async function runTransaction(
...
-      notificationPath: "across-error",
-    });
-    throw error;
+    if (
+      retriesRemaining > 0 &&
+      (error?.code === "NONCE_EXPIRED" ||
+        error?.code === "INSUFFICIENT_FUNDS" ||
+        error?.message.includes("intrinsic gas too low"))
+    ) {
+      // If error is due to a nonce collision or gas underpricement then re-submit to fetch latest params.
```
mrice32, line 58: Will runTransaction automatically fetch the latest nonce or do we need to manually compute it?

mrice32 (commented):
Looks great! Only one question!

src/utils/TransactionUtils.ts
nicholaspai: it should automatically fetch it as it does now right?

mrice32: That's what I'm unsure about. Does ethers grab the nonce at the initial connection time and never requery it? Or does it query it on an interval? Or does it query it on every transaction just before sending? I'll try to look at the provider stack to get a better understanding of how the autopopulation works.

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/462#discussion_r1120637185

src/utils/TransactionUtils.ts
nicholaspai: hm i'm not sure perhaps its better to just manually reset it higher in this case if there is uncertainty

mrice32: Okay, I think your assumption is right @nicholaspai. I did some digging in ethers and found [this](https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/abstract-signer/src.ts/index.ts#L293), which implies that the nonce is pulled each time the transaction is sent, and it includes "pending" transactions.

src/utils/TransactionUtils.ts
```diff
@@ -47,13 +49,40 @@ export async function runTransaction(
...
+        retriesRemaining,
+      });
+      return await runTransaction(
+        logger,
+        contract,
+        method,
+        args,
+        value,
+        gasLimit,
+        await contract.signer.getTransactionCount(),
```
mrice32, line 73: nit: I take back my original point. You can revert this change, and it should work! see https://github.com/across-protocol/relayer-v2/pull/462#discussion_r1120805208

mrice32 (approved):
LGTM!

PR #464, opened 2023-02-27
improve: Log quorum result data

nicholaspai:
Currently we log only mismatched provider data but we should log the expected quorum result


pxrl (approved):
I considered logging this in the past, but figured it's not necessary because we assume that the quorum result was actually correct and will persist. Therefore, because we log the specific query that was used, we should be able to reproduce the quorum response by making another query.

mrice32 (approved):
OOC, was this just something you noticed? Or did you end up needing this?

nicholaspai:
> OOC, was this just something you noticed? Or did you end up needing this?

I wanted to show the diff in logs to an RPC provider and wish I had logged this

PR #465, opened 2023-02-28
feat: Support disabling chains

nicholaspai:
## Overview
- The majority of the diff in this PR comes from refactoring and removing re-used code. I think it will give many quality of life improvements. I can try to break some of the changes in to separate PR's if need be.
- This PR introduces way for the bots to handle a chain being disabled on-chain in the `ConfigStore` by being set in the list `DISABLED_CHAINS`.
- The behavior for dealing with a disabled chain is different for the different bot functions:
        - _In deciding whether to construct spoke pool clients for a chain:_ Start with all chains. From the oldest block queried on mainnet to the latest, remove all chain ID's that are on the config store's `DISABLED_CHAINS` list. Construct spoke pool clients for all remaining chain ID's.
    - When validating a bundle or proposing a new one, check if a chain is enabled for the bundle's block range. If a chain is not enabled, then set a special block range for it where `endBlock == startBlock == previousBundleEndBlockForChain`.
         - In the rare case where we didn't create a spoke pool client for an enabled chain, then the bot will refuse to propose or validate a bundle. This is because of this [line](https://github.com/across-protocol/relayer-v2/pull/465/files#diff-9535349f4b8beb74323e6d78b13634dc9351a05c53fc08785c124cd5a60c8474R115) in `DataworkerUtils#blockRangesAreInvalidForSpokePoolClients()`.
    - The disabled chain list returned by the on-chain `ConfigStore` can be overridden via the `SPOKE_POOL_CHAINS_OVERRIDE` environment variable.  For example, this can be used to finalize withdrawals that originated on a chain prior to it being disabled.

## Reviewers guide
- First place to look is `ConfigStoreClient` which will read the "DISABLED_CHAINS" key from the ConfigStore. This should allow the user of the client to get the list of disabled chains at a mainnet block, or get the list of enabled chains in a block range. Design decisions were made here in determining how to parse non-integer values.
- The entry point next is in the `ClientHelper` and `DataworkerClientHelper` where I modified the code that constructs `SpokePoolClients` to only constructs clients for chains that were enabled at least once in a block range.
- An important logical change is in `PoolRebalanceUtils#getWidestPossibleExpectedBlockRange`, which has a new rule for returning bundle block ranges for a disabled chain.
- Another important change is a new function `getBundleBlockRanges` that is used all throughout the `Dataworker` to construct bundle block ranges for historical bundles. This includes the logic to handle block ranges for disabled chains. A lot of existing code that replicates this logic was refactored to instead call this function

Fixes ACX-854


src/clients/BundleDataClient.ts
```diff
@@ -178,9 +178,6 @@ export class BundleDataClient {
 
     if (!this.clients.hubPoolClient.isUpdated) throw new Error("HubPoolClient not updated");
     if (!this.clients.configStoreClient.isUpdated) throw new Error("ConfigStoreClient not updated");
-    this.chainIdListForBundleEvaluationBlockNumbers.forEach((chainId) => {
```
nicholaspai, line 181: This client only loads events from the chain IDs for each spoke pool client, so we don't need this check.

src/dataworker/DataworkerConfig.ts
```diff
@@ -82,10 +82,8 @@ export class DataworkerConfig extends CommonConfig {
     if (this.executorEnabled)
       assert(this.spokeRootsLookbackCount > 0, "must set spokeRootsLookbackCount > 0 if executor enabled");
     else if (this.disputerEnabled || this.proposerEnabled)
-      assert(
```
nicholaspai, line 85: Unrelated but useful fix to set lookback count to 0 if not executing leaves

src/finalizer/index.ts
```diff
@@ -213,6 +213,7 @@ export async function finalize(
 
 export async function constructFinalizerClients(_logger: winston.Logger, config, baseSigner: Wallet) {
   const commonClients = await constructClients(_logger, config, baseSigner);
+  await updateFinalizerClients(commonClients);
```
nicholaspai, line 216: need the config store client updated before checking for disabled chains in `constructSpokePoolClientsWithLookback`

mrice32 (commented):
This is cool!

One question on the finalizer being disabled... wouldn't we want the finalizer to still run, at least for a period of time after the chain is disabled? Otherwise, couldn't we miss the last few transfers?

src/common/ClientHelper.ts
```diff
@@ -43,46 +51,120 @@ export async function constructSpokePoolClientsWithLookback(
...
+  config: CommonConfig,
+  baseSigner: Wallet,
+  startBlockOverride: { [chainId: number]: number } = {},
+  toBlockOverride: { [chainId: number]: number } = {}
+): Promise<SpokePoolClientsByChain> {
+  // By default, construct spoke clients for all chains that are not disabled at the time of the first block that we'll
+  // query. This does not handle the case where a chain was disabled at the start block and later un-disabled, but
+  // this case is very rare. This is a good solution for handling the case where a chain is disabled and we want to
+  // eventually stop sending RPC requests for it, but we might need to reconstruct bundles (for the executor) for
+  // a while even after we disable it.
```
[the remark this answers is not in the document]
nicholaspai: I was just thinking about this and I think this is better and would be easy to do. It also would make it easier to support an `ENABLED_CHAIN_IDS` list where chains are added in the middle of a block range

src/common/ClientHelper.ts
```diff
@@ -43,46 +51,120 @@ export async function constructSpokePoolClientsWithLookback(
...
+  // as input into `getDisabledChainsForBlock`
+  const disabledChains =
+    config.disabledChainsOverride.length > 0
+      ? config.disabledChainsOverride
+      : configStoreClient.getDisabledChainsForBlock(startBlockOverride[1]);
+
+  // In no cases do we want to override the disabled chain list for Dataworker functions. If we want to reconstruct
+  // and older bundle, then the `startBlockOverride`-`toBlockOverride` should cover that older bundle's range,
+  // and in that case if the chain was enabled at `startBlockOverride` then we will construct a spoke client
+  // for it.
```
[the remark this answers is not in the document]
nicholaspai: Legacy code, removed

nicholaspai:
> This is cool!
> 
> One question on the finalizer being disabled... wouldn't we want the finalizer to still run, at least for a period of time after the chain is disabled? Otherwise, couldn't we miss the last few transfers?

I agree we'd want the finalizer to still run for a bit after a chain is disabled. This is why we'd still construct spoke pool clients for all chains that were enabled at some point in the lookback. The finalizer uses a 14 day lookback, so as long as the chain was enabled at some point in the last 14 days, the finalizer could still run, assuming the caller sets `FINALIZER_CHAINS` to include the chain ID that was recently disabled.

src/common/ClientHelper.ts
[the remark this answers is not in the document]
nicholaspai: I added some code to this effect [here](https://github.com/across-protocol/relayer-v2/pull/465/commits/2ec564aa62be09c70ff8b81bedee4d5238d4aa08)

src/clients/ConfigStoreClient.ts
```diff
@@ -162,6 +166,50 @@ export class AcrossConfigStoreClient {
...
+        if (disabledChainUpdate.blockNumber > toBlock || disabledChainUpdate.blockNumber < fromBlock)
+          return enabledChains;
+        allPossibleChains.forEach((chainId) => {
+          if (!disabledChainUpdate.chainIds.includes(chainId) && !enabledChains.includes(chainId))
+            enabledChains.push(chainId);
+        });
+        return enabledChains;
+      }, enabledChainsAtFromBlock)
+      .sort((a, b) => a - b);
+  }
```
mrice32, line 204: nit: would be great to get a few minor explanatory comments in here.

src/clients/ConfigStoreClient.ts
```diff
@@ -162,6 +166,50 @@ export class AcrossConfigStoreClient {
...
+
+  getDisabledChainsForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number[] {
+    return (
+      sortEventsDescending(this.cumulativeDisabledChainUpdates).find((config) => config.blockNumber <= blockNumber)
```
mrice32, line 208: Note: in the future, it might make sense to sort the common way (descending in this case) when pulling events in so we don't have to sort on every access.

mrice32 (approved):
LGTM!

src/clients/ConfigStoreClient.ts
nicholaspai: hmm thats a good point, maybe at the end of `update()` we can sort all data in memory once?

src/clients/ConfigStoreClient.ts
nicholaspai: done! sorry about that i thought it was more clear

PR #468, opened 2023-03-07
improve: Get spoke pool addresses and deployment blocks using on-chain events

nicholaspai:
Removes need to update hardcoded mapping of spoke pool addresses and deployment blocks. Will make upgrading spoke pools easier in future.

This PR also introduces some optimizations:
- `getProvider` always tries to load `redisClient` instead of asking caller to pass it in as a param
- `getBlockForTimestamp` similarly always tries to load `redisClient` and fetch the block from the redis cache instead of asking user to pass it in

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -64,7 +64,7 @@ export class HubPoolClient {
     return this.disputedRootBundles;
   }
 
-  getSpokePoolForBlock(block: number, chain: number): string {
+  getSpokePoolForBlock(chain: number, block: number = Number.MAX_SAFE_INTEGER): string {
     if (!this.crossChainContracts[chain]) throw new Error(`No cross chain contracts set for ${chain}`);
     const mostRecentSpokePoolUpdateBeforeBlock = (
       sortEventsDescending(this.crossChainContracts[chain]) as CrossChainContractsSet[]
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -9,7 +9,7 @@ import {
   MakeOptional,
 } from "../utils";
 import { sortEventsDescending, spreadEvent, spreadEventWithBlockNumber, paginatedEventQuery, toBN } from "../utils";
-import { IGNORED_HUB_EXECUTED_BUNDLES, IGNORED_HUB_PROPOSED_BUNDLES } from "../common";
+import { IGNORED_HUB_EXECUTED_BUNDLES, IGNORED_HUB_PROPOSED_BUNDLES, DISABLED_CHAINS } from "../common";
 import { Deposit, L1Token, CancelledRootBundle, DisputedRootBundle, LpToken } from "../interfaces";
 import { ExecutedRootBundle, PendingRootBundle, ProposedRootBundle } from "../interfaces";
 import { CrossChainContractsSet, DestinationTokenWithBlock, SetPoolRebalanceRoot } from "../interfaces";
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -9,7 +9,7 @@ import {
   MakeOptional,
 } from "../utils";
 import { sortEventsDescending, spreadEvent, spreadEventWithBlockNumber, paginatedEventQuery, toBN } from "../utils";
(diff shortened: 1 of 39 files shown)


src/common/ClientHelper.ts
```diff
@@ -172,9 +160,28 @@ export async function constructSpokePoolClientsWithStartBlocks(
...
+      // spoke pool.
+      const latestSpokePool = configStoreClient.hubPoolClient.getSpokePoolForBlock(chainId, toBlockOverride[1]);
+      const spokePoolContract = new Contract(latestSpokePool, SpokePool.abi, spokePoolSigners[chainId]);
+      const spokePoolRegistrationBlock = configStoreClient.hubPoolClient.getSpokePoolActivationBlock(
+        chainId,
+        latestSpokePool
+      );
+      const time = (await configStoreClient.hubPoolClient.hubPool.provider.getBlock(spokePoolRegistrationBlock))
+        .timestamp;
+      const registrationBlock = await getBlockForTimestamp(
```
mrice32, line 175: +1 this is clever.

src/clients/HubPoolClient.ts
```diff
@@ -64,14 +64,23 @@ export class HubPoolClient {
...
+    if (!mostRecentSpokePoolUpdateBeforeBlock)
       throw new Error(`No cross chain contract found before block ${block} for chain ${chain}`);
-    else return mostRecentSpokePoolUpdatebeforeBlock.spokePool;
+    else return mostRecentSpokePoolUpdateBeforeBlock.spokePool;
+  }
+
+  getSpokePoolActivationBlock(chain: number, spokePool: string): number {
+    // Return first time that this spoke pool was registered in the HubPool as a cross chain contract. We can use
+    // this block as the oldest block that we should query for SpokePoolClient purposes.
+    const mostRecentSpokePoolUpdateBeforeBlock = this.crossChainContracts[chain].find(
```
mrice32, line 80: I noticed that these are not explicitly sorted when we query them. This would be unlikely to cause problems with this event, but should we consider sorting when we pull in events and storing them as sorted. I think this would bring some advantages:
1. We rarely have to sort after the query. We would almost always use the data as is (if it's already in the direction we need), work through it backwards (annoying, but we could build a simple generator for this I think), or just call reverse.
2. We won't run into any bugs if a node provider returns values out of order (I suspect this wouldn't happen, but it's probably better to be safe).

mrice32 (approved):
LGTM!

src/clients/HubPoolClient.ts
nicholaspai: yea this is a good call I'll work on this in a follow up PR

src/utils/RedisUtils.ts
```diff
@@ -61,16 +63,25 @@ export async function getDeposit(key: string, redisClient: RedisClient): Promise
...
   chainId: number,
   timestamp: number,
   currentChainTime: number,
-  blockFinder: BlockFinder<Block>,
-  redisClient?: RedisClient
+  redisClient?: RedisClient,
+  blockFinder?: BlockFinder<Block>
 ): Promise<number> {
-  if (!redisClient) return (await blockFinder.getBlockForTimestamp(timestamp)).number;
+  if (blockFinder === undefined) blockFinder = getBlockFinder(chainId, redisClient);
```
nicholaspai, line 83: One improvement introduced in this PR is to not require the `blockFinder` to be passed in, rather we should load it if undefined

src/utils/RedisUtils.ts
```diff
@@ -1,30 +1,39 @@
...
         logger.debug({
           at: "Dataworker#ClientHelper",
           message: `Connected to redis server at ${url} successfully!`,
           dbSize: await redisClient.dbSize(),
         });
-      return redisClient;
-    });
+      redisClients[url] = redisClient;
+    } catch (err) {
+      // No redis client at URL
```
mrice32, line 35: Maybe we should put a debug log here with the error?

mrice32 (approved):
LGTM! I think this is a good addition -- making redis, blockfinder, and other objects that are used everywhere globals.

src/utils/RedisUtils.ts
nicholaspai: good call!

PR #469, opened 2023-03-08
feat: Cache providers by chainId and cache/noCache flag

nicholaspai:
This fixes an issue where the order in which `getProvider()` is called determines which kind of provider is cached: either one with redis or without. However, `getProvider` should always return the expected "cached" provider, with redis or without as the caller expects.

This PR fixes the issue by caching the provider with its redis flag.

Additionally, any calls in the repo to `getProvider` are all modified to be called the same way, with `redis` enabled.


mrice32 (approved):
LGTM!

PR #471, opened 2023-03-08
feat: Add simple oSnap monitor

nicholaspai:
Calls upon the [Snapshot API](https://docs.snapshot.org/guides/graphql-api) and logs of any new proposals created within the lookback.

In production I imagine that this script can be run serverlessly `ts-node ./src/scripts/querySnapshotProposals.ts`


nicholaspai:
Will wait to work more on this PR until we've deployed the oSnap module on-chain and we can filter on proposals that can be used with that module

PR #478, opened 2023-03-09
fix: Filter out tokens that don't have any non-mainnet destination chains before calling /limits

nicholaspai:
For example, BOBA has no destination chains besides 1 so there is no need to call /limits on it


mrice32 (approved):
LGTM! I assume this is the fix for the 400s?

PR #479, opened 2023-03-09
fix: Dataworker should count a chain as disabled if it is disabled at mainnet bundle end block

nicholaspai:
Currently dataworker only treats a chain as disabled if it disabled for entire block range, this is a bug as it should be disabled as soon as the bundle's end block is greater than the time the chain was added to the disabled list


mrice32 (approved):
LGTM!

PR #480, opened 2023-03-10
fix: Subtract buffer from mainnet bundle end block when proposing new  bundle

nicholaspai:
This opens the dataworker to slight chance that "latest" disabled chain list isn't valid for a bundle because the bundle lags `latest` by 100 blocks

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -269,7 +269,7 @@ export class Dataworker {
     // to construct the potential next bundle block range.
     const blockRangesForProposal = await this._getWidestPossibleBlockRangeForNextBundle(
       spokePoolClients,
-      this.clients.hubPoolClient.latestBlockNumber
+      this.clients.hubPoolClient.latestBlockNumber - this.blockRangeEndBlockBuffer[1]
     );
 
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
Sorry, this makes it more likely that there's a discrepency? What bug does it fix?

nicholaspai:
> Sorry, this makes it more likely that there's a discrepency? What bug does it fix?

This fixes the bug where the DISABLED_CHAIN was added after a mainnet bundle end block but before `latest`. Therefore the ConfigStoreClient will return the chain as disabled, so the bundle blocks  will be constructed as if it were disabled, even though the bundle end block < block when chain was disabled

mrice32:
> > Sorry, this makes it more likely that there's a discrepency? What bug does it fix?
> 
> This fixes the bug where the DISABLED_CHAIN was added after a mainnet bundle end block but before `latest`. Therefore the ConfigStoreClient will return the chain as disabled, so the bundle blocks will be constructed as if it were disabled, even though the bundle end block < block when chain was disabled

Ah, I see, that makes sense. I think I misunderstood. LGTM!

mrice32 (approved):

PR #481, opened 2023-03-10
fix: Should pass in latestMainnetBlock into `hubPoolClient.getLatestBundleEndBlockForChain`

nicholaspai:
Also includes a refactor to pass in `enabledChains` into `getWidestPossibleBlockNumbers` which means we don't need to pass in `mainnetBundleEndBlock` to the function because its only used to get enabled chains

This will also make it easier to unit test `getWidestPossibleBlockNumbers`

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -470,7 +470,10 @@ export function generateMarkdownForRootBundle(
   // Create helpful logs to send to slack transport
   let bundleBlockRangePretty = "";
   chainIdListForBundleEvaluationBlockNumbers.forEach((chainId, index) => {
-    bundleBlockRangePretty += `\n\t\t${chainId}: ${JSON.stringify(bundleBlockRange[index])}`;
+    const isChainDisabled = bundleBlockRange[index][0] === bundleBlockRange[index][1];
+    bundleBlockRangePretty += `\n\t\t${chainId}: ${JSON.stringify(bundleBlockRange[index])}${
+      isChainDisabled ? " 🥶" : ""
+    }`;
   });
 
   const convertTokenListFromWei = (chainId: number, tokenAddresses: string[], weiVals: string[]) => {
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -389,20 +389,15 @@ export async function getWidestPossibleExpectedBlockRange(
   endBlockBuffers: number[],
   clients: Clients,
   latestMainnetBlock: number,
-  mainnetBundleEndBlock: number
+  enabledChains: number[]
 ): Promise<number[][]> {
-  // A chain is only enabled for a bundle range if it was not disabled at the time of the mainnet bundle end block.
-  const enabledChains = clients.configStoreClient.getEnabledChains(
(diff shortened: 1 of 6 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -389,20 +389,15 @@ export async function getWidestPossibleExpectedBlockRange(
...
-    mainnetBundleEndBlock,
-    chainIdListForBundleEvaluationBlockNumbers
-  );
   return chainIdListForBundleEvaluationBlockNumbers.map((chainId: number, index) => {
     // If chain is disabled, re-use the latest bundle end block for the chain as both the start
     // and end block.
     if (!enabledChains.includes(chainId)) {
       const lastEndBlockForDisabledChain = clients.hubPoolClient.getLatestBundleEndBlockForChain(
         chainIdListForBundleEvaluationBlockNumbers,
-        mainnetBundleEndBlock,
```
nicholaspai, line 405: This was the most important change in this PR. We should be looking for latest validated bundle as of `latest` rather than mainnet bundle end block. In practice this doesn't really make a difference unless there's a huge gap between the bundle end block and `latest` and/or bundles are proposed very quickly after each other, in which case its possible that the recently validated bundle occurs at a block < mainnet bundle end block

james-a-morris: QQ: does the relayer's source of truth not end at the latest mainnet block it has captured? 

Is it possible/would any errors arise from a case that a chain is not enabled between block [0, mainnet's captured end block] but it was enabled between (mainnet's captured end block, latest]

james-a-morris (commented):
Left a comment for some additional insight into the relayer. Looks great so far 🙂

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: so this function is only used by the `dataworker` and the dataworker's unit of time is bundles which are demarcated by "bundle end blocks". The source of truth is the bundle end block, so if a chain is not disabled prior to the bundle end block, then its not disabled until the next bundle, which begins at current bundle end block + 1. Does that make sense?

> a chain is not enabled between block [0, mainnet's captured end block] but it was enabled between (mainnet's captured end block, latest]

For example, in this case, the chain would only be treated as disabled in the next bundle [mainnet's captured end block, latest]

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -475,7 +470,10 @@ export function generateMarkdownForRootBundle(
   // Create helpful logs to send to slack transport
   let bundleBlockRangePretty = "";
   chainIdListForBundleEvaluationBlockNumbers.forEach((chainId, index) => {
-    bundleBlockRangePretty += `\n\t\t${chainId}: ${JSON.stringify(bundleBlockRange[index])}`;
+    const isChainDisabled = bundleBlockRange[index][0] === bundleBlockRange[index][1];
+    bundleBlockRangePretty += `\n\t\t${chainId}: ${JSON.stringify(bundleBlockRange[index])}${
+      isChainDisabled ? " 🥶" : ""
```
nicholaspai, line 475: Screenshot 2023-03-09 at 22 17 14
 

Preview

PR #485, opened 2023-03-11
improve(TransactionUtils): Retry txn on REPLACEMENT_UNDERPRICED

pxrl:
This likely happens when there is a collision between multiple relayer instances.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -53,6 +53,7 @@ export async function runTransaction(
       retriesRemaining > 0 &&
       (error?.code === "NONCE_EXPIRED" ||
         error?.code === "INSUFFICIENT_FUNDS" ||
+        error?.code === "REPLACEMENT_UNDERPRICED" ||
         error?.message.includes("intrinsic gas too low"))
     ) {
       // If error is due to a nonce collision or gas underpricement then re-submit to fetch latest params.


nicholaspai (commented):
I know this isn't your fault but thoughts on using a `Set` of known error codes/messages versus expanding this `if` conditional? Similar to how we do it in `MulticallerClient` [here](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/MultiCallerClient.ts#LL22C14-L22C32)

mrice32:
> I know this isn't your fault but thoughts on using a `Set` of known error codes/messages versus expanding this `if` conditional? Similar to how we do it in `MulticallerClient` [here](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/MultiCallerClient.ts#LL22C14-L22C32)

+1 good idea

pxrl:
> I know this isn't your fault but thoughts on using a `Set` of known error codes/messages versus expanding this `if` conditional? Similar to how we do it in `MulticallerClient` [here](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/MultiCallerClient.ts#LL22C14-L22C32)



Good suggestion. Have implemented it now :+1:

src/utils/TransactionUtils.ts
```diff
@@ -9,6 +9,8 @@ export type TransactionSimulationResult = {
   reason: string;
 };
 
+const txnRetryErrors = new Set(["INSUFFICIENT_FUNDS", "NONCE_EXPIRED", "REPLACEMENT_UNDERPRICED"]);
```
james-a-morris, line 12: Nit: should this be defined within a constants file instead of in the utilities?

src/utils/TransactionUtils.ts
```diff
@@ -49,12 +51,7 @@ export async function runTransaction(
     );
     return await contract[method](...args, txConfig);
   } catch (error) {
-    if (
-      retriesRemaining > 0 &&
-      (error?.code === "NONCE_EXPIRED" ||
-        error?.code === "INSUFFICIENT_FUNDS" ||
-        error?.message.includes("intrinsic gas too low"))
-    ) {
+    if (retriesRemaining > 0 && (txnRetryErrors.has(error?.code) || error?.message.includes("intrinsic gas too low"))) {
```
james-a-morris, line 54: Nit: would it make sense to create a function specifically for testing whether an `error: unknown | Error` is an error in which a retry is valid?

i.e. `const isErrorRetryable = (error?: unknown) =>  txnRetryErrors.has((error as Error)?.code)`

james-a-morris: ^ I'm not sure if we'll ever re-use this logic but it may be easier to codify it now.

james-a-morris (commented):
Looks great!

src/utils/TransactionUtils.ts
mrice32: +1

mrice32 (approved):
LGTM!

src/utils/TransactionUtils.ts
nicholaspai: I think its OK here for now but if do find we need to export this elsewhere we should migrate to constants

src/utils/TransactionUtils.ts
```diff
@@ -9,6 +9,20 @@ export type TransactionSimulationResult = {
...
+type EthersError = Error & {
+  code: string;
+  reason: string;
+};
+
+const isEthersError = (error?: unknown): error is EthersError =>
+  (error as EthersError)?.code in Object.values(ethers.utils.Logger.errors);
+const txnRetryErrors = new Set(["INSUFFICIENT_FUNDS", "NONCE_EXPIRED", "REPLACEMENT_UNDERPRICED"]);
+const txnRetryable = (error?: unknown): boolean => {
+  if (isEthersError(error)) return txnRetryErrors.has(error.code);
```
pxrl, line 21: nb. I initially tried the following, but was getting "unreachable code" warnings on the `txnRetryErrors.has(...)` statement because the predicate didn't seem to work when used as the first conditional in the && statement.

```
return
  (isEthersError(error) && txnRetryErrors.has(error.code))
  || (error as Error)?.message?.includes(...);
```

james-a-morris: The logic you have now LGTM - I think it will be helpful if we add more logic around catching errors.

james-a-morris (approved):

PR #487, opened 2023-03-13
refactor(Dataworker): Execute all mainnet leaves before other chains

nicholaspai:
Fixes ACX-945

Test execution: https://etherscan.io/tx/0x75105c5227bdcefc8f1e2f2744cddcbd5502dbaac6ab67c3e0c5f07c84806289#eventlog

# Summary

This PR introduces 1 new feature to the `MultiCallerClient` and one optimization to the `Dataworker`. This PR has a large diff because I had to copy the `Multicall2.sol` contract into the Repo in order to deploy for tests and fetch the ABI for the `MultiCallerClient`.

## MultiCallerClient

Extends the `AugmentedTransaction` type with props `unpermissioned: boolean` and `canFailInSimulation: boolean`. These can be used to mark transactions that can be executed by anybody (i.e. they are "unpermissioned") and that we can batch together by sending the transaction through a `Multicall` [contract](https://github.com/mds1/multicall).

Additionally, the `canFailInSimulation` flag is used to mark transactions that can if we individually simulate them. These might be transactions that are path dependent on other transactions executing before them. This will make more sense with the changes to the Dataworker's execution logic.

## Dataworker 

Currently all `PoolRebalanceLeaves` are executed together in a batch. This makes sense as each leaf is independent of each other and they do not draw upon the same pool of L1 tokens.

This can sometimes fail when trying to execute an L2 leaf (i.e. one that is not for mainnet) at the same time as the mainnet leaf. Imagine that an L2 leaf wants to send 100 tokens to Optimism. This leaf might require 100 tokens to be returned from the `Ethereum_SpokePool` in order to send to Optimism. This is a rare case but its possible. 

This PR addresses this issue by changing the default pool rebalance leaf execution logic to perform executions in the following order:

1. Execute chainId=1 `PoolRebalanceLeaves`
2. Execute chainId=1 `RelayerRefundLeaves` and `SlowRelayLeaves`
3. Execute all other `PoolRebalanceLeaves`

Since the `Ethereum_SpokePool` is on the same network as the `HubPool`, we can atomically batch together these transactions. This reduces the probability that any leaf in step 3 fails to execute because of the added liquidity returned to the `HubPool` by step 2.

## Back to the Multicaller changes

So, today's `MulticallerClient` simulates individually all transactions before executing them. However, we can't simply simulate step 2 or step 3 above, because they might depend on capital returned by previous steps. This is where we can take advantage of the new `canFailInSimulation` property in the `AugmentedTransaction` type.

In order to not miss unexpected simulation failures, I've added a check to the `MultiCallerClient` to simulate the full batch (e.g. the ultimate call to `Multicall2#aggregate` or `Multicaller#multicall`) before proceeding with the execution.


src/clients/BalanceAllocator.ts
```diff
@@ -110,6 +110,11 @@ export class BalanceAllocator {
     this.used[chainId][token][holder] = used.add(amount);
   }
 
+  async addBalance(chainId: number, token: string, holder: string, amount: BigNumber) {
+    const balance = await this.getBalance(chainId, token, holder);
+    this.balances[chainId][token][holder] = balance.add(amount);
```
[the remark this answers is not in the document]
nicholaspai: I don't think so as `getBalance` will guarantee that the `balances` entry is set. I'll add a comment on that

src/clients/MultiCallerClient.ts
```diff
@@ -134,6 +143,38 @@ export class MultiCallerClient {
...
+        });
+        throw new Error("Failed to simulate transaction batch!");
+      } else {
+        this.logger.debug({
+          at: "MultiCallerClient#executeTxnQueue",
+          message: `Successfully simulated ${networkName} transaction batch!`,
+          batchTxn: {
+            ...batchTxn.transaction,
+            contract: batchTxn.transaction.contract.address,
+          },
```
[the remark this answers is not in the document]
nicholaspai: This is a good call! https://github.com/across-protocol/relayer-v2/pull/487/commits/829f9528fdad0c43f346346e14ac4cd50e5507ae

src/clients/BalanceAllocator.ts
```diff
@@ -110,6 +110,19 @@ export class BalanceAllocator {
     this.used[chainId][token][holder] = used.add(amount);
   }
 
+  /**
+   * @notice Manually add balance to the balance entry for the holder for the token and chain ID.
+   * Caller should be careful in calling this function otherwise the user of this client
+   * could be led to think that the holder has more on-chain balance than it actually has.
+   * @param amount Amount to increase balance by. Can be negative.
+   */
+  async addBalance(chainId: number, token: string, holder: string, amount: BigNumber): Promise<void> {
```
mrice32, line 119: nit: instead of modifying the balance (which is supposed to be a direct representation of on-chain info), have you considered just adding a negative "used" amount.

For instance, this should _raise_ the effective balance that you have to work with by 100 wei:
```ts
balanceAllocator.addUsed(chainId, token, holder, BigNumber.from("-100"));
```

Only issue I see is that we may need to rework lines 51-68 to not use balances, but instead use the "balance - used" amount.

Thoughts?

test/MultiCallerClient.ts
```diff
@@ -291,4 +294,127 @@ describe("MultiCallerClient", async function () {
     const txnQueue: AugmentedTransaction[] = await multiCaller.buildMultiCallBundles([], 10);
     expect(txnQueue.length).to.equal(0);
   });
+
+  it("Correctly handles unpermissioned transactions", async function () {
+    const _multisender = new ethers.ContractFactory(
+      hre.artifacts.readArtifactSync("Multicall3").abi,
+      hre.artifacts.readArtifactSync("Multicall3").bytecode,
```
mrice32, line 301: nit: why do we have mixed use of multicall2 from contracts-node and multicall3 from this repo? Can we use just one version of multicall to keep everything simple and consistent?

src/clients/MultiCallerClient.ts
```diff
@@ -149,17 +183,63 @@ export class MultiCallerClient {
...
+    const mrkdwn: string[] = [];
+    let callData: Multicall2Call[] = transactions.map((txn, idx) => {
+      mrkdwn.push(`\n  *txn. ${idx + 1}:* ${txn.message ?? "No message"}: ${txn.mrkdwn ?? "No markdown"}`);
+      return {
+        target: txn.contract.address,
+        callData: txn.contract.interface.encodeFunctionData(txn.method, txn.args),
+      };
+    });
+
+    // There should not be any duplicate call data blobs within this array. If there are there is likely an error.
```
mrice32, line 224: Two notes on this line:
1. I don't think it does much (if anything) in this case (vs multicall). The reason is that we're passing in objects here, rather than actual callData (which is what we do in the other function). This means that the set is comparing these based on _object references_. This means if I pass in two different objects with the exact same data in them, they would compare as different. If we want to correctly compare them, I think we would need to JSON.stringify on the way in and JSON.parse on the way out.
See these console examples:
```
> [...new Set([{a: "a", b: "b"}, {a: "a", b: "b"}])]
[ { a: 'a', b: 'b' }, { a: 'a', b: 'b' } ]
> [...new Set([{a: "a", b: "b"}, {a: "a", b: "b"}].map(JSON.stringify))].map(JSON.parse)
[ { a: 'a', b: 'b' } ]
```


2. If we think this is an error, should we throw/log an error if this happens? Note: we could check this by verifying the lengths of the array before is equal to the array after.

src/clients/MultiCallerClient.ts
```diff
@@ -204,11 +284,37 @@ export class MultiCallerClient {
...
+    const { chainId } = txns[0];
+
+    // If we can't construct multisender contract, then multicall everything. If any of the transactions
+    // is for a contract that can't be multicalled, then this function will throw. This client should only be
+    // used on contracts that extend Multicaller.
+    if (this.getMultisender(chainId) === undefined) {
+      const txnChunks = lodash.chunk(txns, chunkSize);
+      return txnChunks.map((txnChunk) => {
+        // Don't wrap single transactions in a multicall.
+        return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0];
```
mrice32, line 297: This may be a dumb question, but where do we make sure that these calls are all to the same contract? Would this break otherwise?

src/clients/BalanceAllocator.ts
nicholaspai: This is a great idea, but doesn't line 51-68 already use "balance-used"? Isn't that what this line does https://github.com/across-protocol/relayer-v2/blob/master/src/clients/BalanceAllocator.ts#L54

test/MultiCallerClient.ts
nicholaspai: Good callout I think we should switch to just using Multicall3 which is slightly more gas efficient. I'll create a helper method

src/clients/MultiCallerClient.ts
nicholaspai: I blindly copied this but upon reading it again I don't think we should care if there are duplicates. This doesn't seem like an error to me, what do you think?

src/clients/MultiCallerClient.ts
nicholaspai: In `buildMultiCallBundle` we check that all calls are to the same contract, but in `buildMultiSendBundle` we don't check it because it doesn't matter if they are to different contracts. That makes sense right?

src/clients/MultiCallerClient.ts
mrice32: Agreed, I'm not sure why we have a special case for this.

src/clients/BalanceAllocator.ts
mrice32: Sorry, I think 51-61 is fine. I do think 61-68 determines which token to use based on the balance only, not the _balance minus the used amount_. Not sure whether this is a bug or if this worked fine assuming used was always positive, but I think we probably want to look at the `balance - used` rather than just balance in that second loop. Thoughts?

Here's how I would change it:
```ts
 if (success) 
	   requestsWithbalances.forEach(({ chainId, tokens, holder, balances, amount }) => 
	     tokens.forEach((token) => {
               const availableBalance = balances[token].sub(this.getUsed(chainId, token, holder));
	       const used = amount.gt(availableBalance) ? availableBalance : amount; 
	       this.addUsed(chainId, token, holder, used); 
	       amount = amount.sub(used); 
	     }) 
	   ); 
```

https://github.com/across-protocol/relayer-v2/blob/6845dc7e37c77463dcfb66556b28962b77aeb33f/src/clients/BalanceAllocator.ts#L61-L68

src/clients/MultiCallerClient.ts
mrice32: I think I agree with that, but this function can call buildMultiCallBundle -- is there anything here or upstream that stops us from hitting that error condition?

Put differently, if we don't have a multisender contract (i.e. this._getMultisender(chainId) === undefined), would the executor bot completely break since it tries to bundle calls to two separate contracts together?

nicholaspai: Yeah `buildMultiCallBundle` does have this check [here](https://github.com/across-protocol/relayer-v2/pull/487/files#diff-24dbb5604d0001ac862fc1a30b22a6d1dd732855e25d548a3cf88f31c283b840L166) unless I'm misunderstanding.

I think if ` this._getMultisender(chainId) === undefined`, then the executor bot will break yes because we have calls to different contracts. I think this is basically the case today but we never run into it in practice because we can never execute PoolRebalanceLeaves and RelayerRefundLeaves in the same run.

One thing we could do to protect ourselves is add some logic to the line `const txnChunks = lodash.chunk(txns, chunkSize)` where we first sort all txns into groups for the same `target`. Thoughts?

src/dataworker/Dataworker.ts
```diff
@@ -1130,10 +1162,72 @@ export class Dataworker {
...
+    // takes advantage of the fact that mainnet transfers between HubPool and SpokePool are atomic.
+    const mainnetLeaf = unexecutedLeaves.find((leaf) => leaf.chainId === hubPoolChainId);
+    if (mainnetLeaf) {
+      await this._executePoolRebalanceLeaves(
+        spokePoolClients,
+        [mainnetLeaf],
+        balanceAllocator,
+        expectedTrees.poolRebalanceTree.tree,
+        submitExecution
+      );
```
mrice32, line 1182: This doesn't happen really in the current system, but just for the sake of correctness, I think there can be multiple pool rebalance leaves for a particular chain.

Thoughts on something more like:
```suggestion
      await this._executePoolRebalanceLeaves(
        spokePoolClients,
        unexecutedLeaves.filter((leaf) => leaf.chainId === hubPoolChainId),
        balanceAllocator,
        expectedTrees.poolRebalanceTree.tree,
        submitExecution
      );
```

src/dataworker/Dataworker.ts
```diff
@@ -1130,10 +1162,72 @@ export class Dataworker {
...
+
+      // We need to know the next root bundle ID for the mainnet spoke pool in order to execute leaves for roots that
```
mrice32, line 1184: +1 the code below is very clean and clear.

src/dataworker/Dataworker.ts
```diff
@@ -1154,33 +1248,94 @@ export class Dataworker {
...
+                    await balanceAllocator.addBalance(
+                      leaf.chainId,
+                      leaf.l1Tokens[i],
+                      spokePoolClients[leaf.chainId].spokePool.address,
+                      amount
+                    );
+                  }
+                })
+              );
+            }
```
mrice32, line 1281: I think this technically works assuming we run this function the way it is called now, but it's pretty coupled/dependent on the order in which these functions are all called in the executor code path -- if we moved the mainnet relayer refund leaf to the very end of the calls, the balance allocator would think we had the balance, even though we wouldn't.

Put differently, why modify the HubPool balance here, since the PoolRebalanceLeaf execution is not the action that actually adds the balance? The RelayerRefundLeaf execution does impact the balance, so why not increment there, and then no matter how we order these functions the balance allocator will have the correct view of the balances and block us from sending underfunded transactions?

nicholaspai: I think I agree with your reasoning it seems safer to future changes but I'm not sure I agree with this statement:

> since the PoolRebalanceLeaf execution is not the action that actually adds the balance? The RelayerRefundLeaf execution does impact the balance, so why not increment there

This line you highlighted adds balance to the `SpokePool`, while `_executeRelayerRefundLeaves` adds balance to the `HubPool`. Doesn't executing the pool leaf add balance to the SpokePool when `netSendAmount > 0`, while executing relayer refund add balance back to the HubPool?

src/dataworker/Dataworker.ts
nicholaspai: Yep you're right added this in this commit a11d0b10c768079b4e9cafbcdded9d465cff5ab4

src/dataworker/Dataworker.ts
mrice32: Oh, you are totally right! I completely misunderstood this line. I will do another pass, but I think you are right. You are adding to the _spoke pool_ in cases where funds are sent, not adding to the hub pool when funds are pulled. I had it backwards. Sorry for the confusion!

mrice32:
@nicholaspai I think #505 will solve some of the issues we discussed in this PR, so that you can safely just addUsed with a negative value to increase the effective balance of a contract. Do you agree?

src/clients/BalanceAllocator.ts
nicholaspai: bc00a85407a01a2f1a3af356cd30ead6c548e9a2

nicholaspai:
> @nicholaspai I think #505 will solve some of the issues we discussed in this PR, so that you can safely just addUsed with a negative value to increase the effective balance of a contract. Do you agree?

Agreed, this captures that change assuming BalanceAllocator has no bugs, right? bc00a85407a01a2f1a3af356cd30ead6c548e9a2

src/clients/MultiCallerClient.ts
```diff
@@ -204,11 +284,46 @@ export class MultiCallerClient {
...
     chunkSize = DEFAULT_MULTICALL_CHUNK_SIZE
   ): AugmentedTransaction[] {
-    const txnChunks: AugmentedTransaction[][] = lodash.chunk(txns, chunkSize);
-    return txnChunks.map((txnChunk: AugmentedTransaction[]) => {
-      // Don't wrap single transactions in a multicall.
-      return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0];
-    });
+    if (txns.length === 0) return [];
+    const { chainId } = txns[0];
+
```
mrice32, line 289: Note: it might be simpler to just sort out the different types of transactions up here.
```ts
const { multicallerTxns = [], multisenderTxns = [], unsendableTxns = [] } = 
  lodash.groupBy(txns, (txn) => {
    if (txn.contract.multicall) return "multicallerTxns";
    if (txn.unpermissioned) return "multisenderTxns";
    return "unsendableTxns";
  });

if (unsendableTxns.length > 0) {
  // Log error about not being able to send txns.
}

if (this._getMultisender(chainId) === undefined) {
  if (multisenderTxns.length > 0) {
    // Log error about unsendable txns.
  }
  const txnChunks = lodash.chunk(
      multicallerTxns.sort((a, b) => a.contract.address.localeCompare(b.contract.address)),
      chunkSize
    );
    ...
} else {
  // No checks needed here.
  // No filtering necessary.
  const multicallerTxnChunks = lodash.chunk(multicallerTxns.sort(...), chunkSize);
  ...
```

Just a thought. Might not actually be simpler. I just had a bit of trouble following where the checks were.

mrice32 (approved):
LGTM!

src/clients/MultiCallerClient.ts
```diff
@@ -134,6 +138,31 @@ export class MultiCallerClient {
...
+          ? `Successfully simulated ${networkName} transaction batch!`
+          : `Failed to simulate ${networkName} transaction batch!`,
+        batchTxn: {
+          ...batchTxn.transaction,
+          contract: batchTxn.transaction.contract.address,
+        },
+      });
+      if (!batchTxn.succeed) {
+        throw new Error("Failed to simulate transaction batch!");
+      }
```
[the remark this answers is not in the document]
nicholaspai: Yeah we typically bail out if any transactions fail. This is why I added the `canFailInSimulation` prop in `TransactionUtils`. I think its useful to fail loudly if any txns fail but I can also see why we'd want to partially execute the transactions. What do you think is best? I think the flexibility of choosing to bail is important. For example, in the relayer its ok to only send a few of the relays, but the dataworker usually should successfully simulate all txns

src/clients/MultiCallerClient.ts
nicholaspai: I didn't know about the `lodash.groupBy` function that's super helpful! I don't think there's an easy way to identify `unsendableTxns` in your example though, by default any transaction that gets to this client is assumed to target a Multicaller contract

2206188436a669d5939f335473d2b59958004e4a

mrice32: Can't you tell by making sure that the contract instance has a method called multicall? (`if(txn.contract.multicall)`)

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
mrice32: Hmmmm, interesting. I didn't realize we did that elsewhere. If that's the pattern, I think it's best to be consistent, then.

mrice32 (approved):
Just this comment remaining! https://github.com/across-protocol/relayer-v2/pull/487#discussion_r1150025082

src/clients/MultiCallerClient.ts
nicholaspai: ah great point: https://github.com/across-protocol/relayer-v2/pull/487/commits/dc2a86b883a72ec9948279b86cf0d24de204ca45

this wasn't getting caught in tests so I've updated the tests here to include the `multicall` prop (or not)

mrice32 (approved):
LGTM!

PR #493, opened 2023-03-16
feat(Monitor): Add Refiller

nicholaspai:
Refiller is a new Monitor mode that will refill balances. The user must set a trigger and target per account, chain, token combination. The caller is responsible for setting these triggers above the monitored targets/thresholds so as not to create a lot of alert noise. I think this is a reasonable ask of the bot runner since the refiller is sending transactions on-chain. Moreover, the refiller will likely only be refilling a couple accounts.

The initial accounts that should be added are the dataworker and hubpool addresses, as the relayer already has its own inventory management logic.


src/monitor/Monitor.ts
```diff
@@ -50,13 +53,16 @@ interface CategorizedTransfers {
   other: TokenTransfer[];
 }
 
+type BalanceRequest = { chainId: number; token: string; account: string };
```
james-a-morris, line 56: Should we store types in a typedef file?

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
   }
 
+  /**
+   * @notice Checks if any accounts on refill balances list are under their target, if so tries to refill them.
+   * This functionality compliments the report-only mode of `checkBalances`. Its expected that some accounts are
+   * listed in `monitorBalances`. These accounts might also be listed in `refillBalances` with a higher target than
+   * the `monitorBalances` target. This function will ensure that `checkBalances` will rarely alert for those
+   * balances.
+   */
+  async refillBalances() {
```
james-a-morris, line 356: Nit: Should we explicitly provide a `: Promise<void>` return type to indicate that this is a non-returning function? Alternatively, should we return an acknowledgment `true`/`false` value or is the intended effect to unwind the stack in the event of an error?

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
+          account,
+          currentBalance: currentBalances[i].toString(),
+          target: ethers.utils.parseUnits(target.toString(), decimalValues[i]),
+        };
+      }),
+    });
+
+    // Compare current balances with triggers and send tokens if signer has enough balance.
+    const signerAddress = await this.clients.hubPoolClient.hubPool.signer.getAddress();
+    await Promise.all(
```
james-a-morris, line 378: Should we use `promise.all` due to its ability to fail if one inner promise fails? I ask because this may cause a situation where one or more refill requests succeeds but other potential requests are rejected. We may want to use `Promise.allSettled` to give all the requests the chance to be refilled. Also, we can use the return type of `Promise.allSettled` to determine exactly which `refillBalance` request failed.

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
+      refillEnabledBalances.map(async ({ chainId, isHubPool, token, account, target, trigger }, i) => {
+        const currentBalance = currentBalances[i];
+        const decimals = decimalValues[i];
+        const balanceTrigger = ethers.utils.parseUnits(trigger.toString(), decimals);
+        const isBelowTrigger = currentBalance.lt(balanceTrigger);
```
james-a-morris, line 383: QQ: does the threshold trigger activate at `[0, balanceTrigger)` or `[0, balanceTrigger]`?

src/monitor/Monitor.ts
```diff
@@ -743,9 +883,7 @@ export class Monitor {
   }
 
   // Returns balances from cache or from provider if there's a cache miss.
-  private async _getBalances(
-    balanceRequests: { chainId: number; token: string; account: string }[]
-  ): Promise<BigNumber[]> {
+  private async _getBalances(balanceRequests: BalanceRequest[]): Promise<BigNumber[]> {
```
james-a-morris, line 886: +1 on consolidating this to a specific type

james-a-morris (commented):
Left a few comments. This is excellent code 🔥 🔥

src/monitor/MonitorConfig.ts
```diff
@@ -69,6 +81,31 @@ export class MonitorConfig extends CommonConfig {
     this.monitoredRelayers = parseAddressesOptional(MONITORED_RELAYERS);
     this.knownV1Addresses = parseAddressesOptional(KNOWN_V1_ADDRESSES);
 
+    // Used to send tokens if available in wallet to balances under target balances.
+    if (REFILL_BALANCES) {
+      this.refillEnabledBalances = JSON.parse(REFILL_BALANCES).map(
+        ({ chainId, account, isHubPool, target, trigger, token }) => {
+          if (Number.isNaN(target) || target <= 0) throw new Error("target must be > 0");
+          if (Number.isNaN(trigger) || trigger <= 0) throw new Error("target must be > 0");
+          if (trigger >= target) throw new Error("trigger must be < target");
```
pxrl, line 90: It'd be easier to debug these if the exception messages included some identifier from the specific entry that's being parsed - i.e. maybe chainId, account and target (or some other combination?).

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
+      }),
+    });
+
+    // Compare current balances with triggers and send tokens if signer has enough balance.
+    const signerAddress = await this.clients.hubPoolClient.hubPool.signer.getAddress();
+    await Promise.all(
+      refillEnabledBalances.map(async ({ chainId, isHubPool, token, account, target, trigger }, i) => {
+        const currentBalance = currentBalances[i];
+        const decimals = decimalValues[i];
+        const balanceTrigger = ethers.utils.parseUnits(trigger.toString(), decimals);
```
pxrl, line 382: If the trigger or target are "numerically mis-specified" (i.e. as numbers that can't be promoted to _n_ decimals) then this would fail with a BN underflow (or something similar), and present as an obscure failure mid-way through execution (instead of during startup, where we parse configs). One example might be if the number was incorrectly specified with the required number of decimal places by the user.

In this instance, i'm in two minds about whether it's worth wrapping in a try/catch statement to at least produce a meaningful exception message. An alternative might be to pre-elevate trigger and target to 18 decimals and then perform any necessary on-the-fly conversions in the case of decimals < 18 (the ProfitClient does this for relayer fee computations). The latter approach would at least ensure that number specified is a valid BN.

nb. I think there are a few other instances in the codebase where we have config items that can't be fully validated until we've completed on-chain queries. In instances of misconfiguration they will similarly tend to throw obscure runtime errors about failed BN conversions (or similar). IMO it'd be nice if we found a general solution to that so we could fail fast (and with helpful messages) in these cases.

src/monitor/Monitor.ts
pxrl, line 356: It looks like we have a `test/Monitor.ts`; worth adding tests for this feature?

src/monitor/index.ts
```diff
@@ -46,6 +46,13 @@ export async function runMonitor(_logger: winston.Logger, baseSigner: Wallet) {
         logger.debug({ at: "AcrossMonitor", message: "Report disabled" });
       }
 
+      if (config.botModes.refillBalancesEnabled) {
+        await acrossMonitor.refillBalances();
+        await clients.multiCallerClient.executeTransactionQueue();
```
pxrl, line 51: This seems like a layering breach; is there any reason that executeTransactionQueue() can't be called as the final step of refillBalances() ?

src/monitor/Monitor.ts
nicholaspai: This seems like possible overengineering. Caller is setting `trigger` as a number and `target` is also a number. Why would the caller ever set `trigger` and `target` using different denominations?

src/monitor/Monitor.ts
nicholaspai: That' a good feature, one refill shouldn't block another

src/monitor/index.ts
mrice32: Why not put this outside of the if/else? Other modes shouldn't enqueue transactions, so this should just do nothing if refillBalancesEnabled is false, right?

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
+                  transactionHash: receipt.transactionHash,
+                });
+              } else {
+                const erc20 = new Contract(
+                  "ERC20",
+                  ERC20.abi,
+                  this.clients.spokePoolClients[chainId].spokePool.provider
+                );
+                const symbol = await erc20.symbol();
+                this.clients.multiCallerClient.enqueueTransaction({
```
mrice32, line 442: Will these ERC20s be multicall-compliant? I don't think it will be super common, but if we ended up with two of these in the same run, I think this would error trying to call erc20.multicall(...).

This probably isn't the right PR to do this, but I think more and more, we're trying to send different types of transactions that aren't always hitting the multicall contract. However, I still think it's a good idea to send all the transactions at the end of the run rather than doing some inside the run and others batched at the end.

Thoughts on our multicall implementation taking any transaction and dynamically figure out how to order and bundle them?

src/monitor/Monitor.ts
nicholaspai: We probably should, i've never usedone. Thoughts on how we should organize it?

src/monitor/Monitor.ts
nicholaspai: good point, added Promise< void>

src/monitor/Monitor.ts
nicholaspai: Good point, I'm thinking we can merge in the Multicaller logic from https://github.com/across-protocol/relayer-v2/pull/487 and set `unpermissioned=true` so that all of these refill txns go through the Multisender instead of the Multicaller

src/monitor/Monitor.ts
nicholaspai: It triggers at [0, balanceTrigger). Do you think we should use `lte`? I guess that's more intuitive to what a "trigger" is

src/monitor/Monitor.ts
```diff
@@ -339,6 +346,139 @@ export class Monitor {
...
+            // 2. The account is not a HubPool and we want to load ETH.
+            if (isHubPool) {
+              // Note: We ignore the `token` if the account is HubPool because we can't call the method with other tokens.
+              this.clients.multiCallerClient.enqueueTransaction({
+                contract: this.clients.hubPoolClient.hubPool,
+                chainId: this.clients.hubPoolClient.chainId,
+                method: "loadEthForL2Calls",
+                args: [],
+                message: "Reloaded ETH in HubPool 🫡!",
+                mrkdwn: `Loaded ${ethers.utils.formatUnits(deficit, decimals)} ETH from ${signerAddress}.`,
```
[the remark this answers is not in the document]
nicholaspai: yea that would be awesome actually. Have one file with all log constants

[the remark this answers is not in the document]
nicholaspai: Do you have any suggestions? Feel like an FE app would use this a lot more so I feel like you might be the expert onit

src/monitor/Monitor.ts
nicholaspai: Sorry I'm not fully understanding what you're suggesting here, is there an easy fix that can make this a lot safer?

src/monitor/Monitor.ts
nicholaspai: Good point, will add in a later commit

nicholaspai:
@mrice32 @james-a-morris @pxrl thanks for all your reviews, have addressed everything in this commit with a placeholder for unit tests that I'll add after we converge on implementation following more reviews 434aab0831c9d0316fb5557681f2954f85ffa62d

src/monitor/Monitor.ts
james-a-morris: My thought would be to account for the trigger inclusively. However, the tech spec takes priority here

src/monitor/Monitor.ts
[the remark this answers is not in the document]
james-a-morris: I think this package is very powerful for efficient internationalization efforts: https://github.com/ivanhofer/typesafe-i18n

src/monitor/Monitor.ts
james-a-morris: We could create a file called `types.d.ts` and store all type references in that specific file. This way, the types are consolidated & centralized in the codebase.

src/monitor/Monitor.ts
mrice32: Aren't ERC20 transfers implicitly permissioned? Since we can't call transfer via another contract?

Also, I assume we'd primarily use this for ETH transfers (to refill the dataworker and hub pool) today? Do you foresee us doing a lot of ERC20 refills?

mrice32 (commented):
https://github.com/across-protocol/relayer-v2/pull/493#discussion_r1148103473

src/monitor/Monitor.ts
pxrl: I think it'd be an improvement if target and trigger were promoted to an 18-decimal BN when parsing config, and then other numbers were normalised to 18 decimals for the purpose of making the comparison. It's definitely not _significantly_ safer, but just makes it a bit more robust to weird edge cases (which we admittedly may never see).

src/monitor/Monitor.ts
nicholaspai: i like this idea

src/monitor/Monitor.ts
nicholaspai: yeah thats a good point. Perhaps we can remove the ERC20 refills for now, though a way to solve this would be to add a prop to the TransactionUtils to handle these sort of transacrtions (not multicaller, permissioned)

nicholaspai: removed ERC20 refills, this simplifies the logic and tests and is more useful in production c07dd0fbf0e38e09cfe0b1befbda64539f148c70

src/monitor/Monitor.ts
nicholaspai: c07dd0fbf0e38e09cfe0b1befbda64539f148c70

james-a-morris (approved):
LGTM - I don't have much to add outside my initial comments and the comments of @pxrl and @mrice32

mrice32 (approved):
LGTM! Thanks for addressing the comments @nicholaspai!

PR #505, opened 2023-03-22
fix: fix multiple bugs in BalanceAllocator

mrice32:
There are a few bugs in the BalanceAllocator.

1. If you send a request containing two requested allocations for the same token, it will not account for the first allocation impacting the second.
2. If you send a request with multiple tokens and it needed to use multiple tokens to fulfil it, it would ignore the previously used amount when deciding how to increment used.

These two scenarios are the last two tests added in the test file. I verified that these failed in the previous version, but succeed after these changes.

### DIFF
--- a/src/clients/BalanceAllocator.ts
+++ b/src/clients/BalanceAllocator.ts
@@ -1,8 +1,8 @@
-import { BigNumber, ERC20, ethers, ZERO_ADDRESS } from "../utils";
+import { BigNumber, ERC20, ethers, ZERO_ADDRESS, min } from "../utils";
 import lodash from "lodash";
 
 // This type is used to map used and current balances of different users.
-interface BalanceMap {
+export interface BalanceMap {
   [chainId: number]: {
     [token: string]: {
       [holder: string]: BigNumber;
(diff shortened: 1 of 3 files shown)


src/clients/BalanceAllocator.ts
```diff
@@ -82,15 +93,15 @@ export class BalanceAllocator {
...
-      // Note: cannot use assign because it breaks the BigNumber object.
-      if (!this.balances[chainId]) this.balances[chainId] = {};
-      if (!this.balances[chainId][token]) this.balances[chainId][token] = {};
-      this.balances[chainId][token][holder] = balance;
+      const balance = await this._queryBalance(chainId, token, holder);
+      // To avoid inconsitencies, we recheck the balances value after the query.
+      // If it exists, skip the assignment so the value doesn't change after being set.
+      if (!this.balances?.[chainId]?.[token]?.[holder]) {
+        // Note: cannot use assign because it breaks the BigNumber object.
+        if (!this.balances[chainId]) this.balances[chainId] = {};
```
pxrl, line 101: Minor: wdyt about using the ??= operator here? It seems like a more concise way of doing this and it avoids the need to repeat the lookup path into the object, so it _might_ be less prone to copy/paste errors creeping in with future changes.

nb. I should note that the `if (!x[a][b]) x[a][b] = {};` pattern is pre-existing in the code and it appears in several other places that are not touched by this change, so if it makes sense to apply this change then maybe it could be done in a follow-up commit to avoid polluting this one?

```suggestion
        this.balances[chainId] ??= {};
```

pxrl (commented):
Only one minor comment :+1:

nicholaspai (approved):
This is very readable code I think it should work. LGTM

src/clients/BalanceAllocator.ts
mrice32: I think it's going to be tough to root out all the places where we have `if(!x) x = y` throughout the codebase (and it's somewhat risky), but I think using this simpler syntax is great and we should do it when modifying/adding code for sure!!

PR #510, opened 2023-03-23
fix(SpokePoolClient): Don't call eth_call on "latest"

nicholaspai:
We noticed that eth_call doesn't seem to be using the quorum provider, and this is the likely suspect

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -260,7 +260,9 @@ export class SpokePoolClient {
   async binarySearchForBlockContainingDepositId(
     targetDepositId: number,
     initLow = this.spokePoolDeploymentBlock,
-    initHigh = this.latestBlockNumber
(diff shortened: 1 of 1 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -260,7 +260,9 @@ export class SpokePoolClient {
   async binarySearchForBlockContainingDepositId(
     targetDepositId: number,
     initLow = this.spokePoolDeploymentBlock,
-    initHigh = this.latestBlockNumber
+    initHigh = this.latestBlockNumber-1 // Calling eth_call at 'latest' doesn't use quorum. Moreover, we 
+    // will never realistically be looking for the "latest" block using this function, so its safe to set
+    // this earlier than latest.
```
james-a-morris, line 265: Good call - we were having issues several months ago with the `"latest"` block instead of a specific number.

james-a-morris (approved):

nicholaspai:
I'm not actually sure this fixes anything as it seems `this.latestBlock` is a number not the string 'latest'

PR #517, opened 2023-03-24
improve: Speedup event helper functions used when constructing bundles with slow fills

nicholaspai:
## HubPoolClient
- `getRootBundleEvalBlockNumberContainingBlock` should return bundle that contained a target block. To do that it iterates through all proposed bundles until it finds one that meets the criteria. This function is used only when computing the slow fill excess produced by a partial fill going in front of a slow fill execution. This function needs to find the bundle that originally triggered a SlowFIll, so its called once per partial fill that beat out a slow fill leaf execution. Some recently bundles had hundreds of slow fills, notably one had ~250. This function currently runs in ~500ms, but searching in reverse order brings this down to ~1ms.
- Removed repeated sorting of events

## SpokePoolClient
- Reduces number of binary searches to make per historical deposit query. Searching for historical deposits shouldn't happen much for the proposer at all. The executor might need to use it when validating older bundles but usually only < 10 deposits. So, this speed up shaves off 1-2 seconds per query (out of 4) but its not a huge speedup, totalling <30seconds total in run time

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -1,14 +1,4 @@
-import {
-  assign,
-  Contract,
-  winston,
-  BigNumber,
-  ERC20,
-  sortEventsAscending,
-  EventSearchConfig,
-  MakeOptional,
-  BigNumberish,
-} from "../utils";
+import { assign, Contract, winston, BigNumber, ERC20, EventSearchConfig, MakeOptional, BigNumberish } from "../utils";
 import { sortEventsDescending, spreadEvent, spreadEventWithBlockNumber, paginatedEventQuery, toBN } from "../utils";
 import { IGNORED_HUB_EXECUTED_BUNDLES, IGNORED_HUB_PROPOSED_BUNDLES } from "../common";
 import { Deposit, L1Token, CancelledRootBundle, DisputedRootBundle, LpToken } from "../interfaces";
@@ -25,7 +15,6 @@ export class HubPoolClient {
   private l1Tokens: L1Token[] = []; // L1Tokens and their associated info.
   private lpTokens: { [token: string]: LpToken } = {};
   private proposedRootBundles: ProposedRootBundle[] = [];
-  private reverseProposedRootBundles: ProposedRootBundle[] = [];
   private canceledRootBundles: CancelledRootBundle[] = [];
   private disputedRootBundles: DisputedRootBundle[] = [];
   private executedRootBundles: ExecutedRootBundle[] = [];
@@ -228,7 +217,8 @@ export class HubPoolClient {
   ): number | undefined {
     let endingBlockNumber: number | undefined;
     // Search proposed root bundles in reverse chronological order.
-    for (const rootBundle of this.reverseProposedRootBundles) {
+    for (let i = this.proposedRootBundles.length - 1; i >= 0; i--) {
+      const rootBundle = this.proposedRootBundles[i];
       const nextRootBundle = this.getFollowingRootBundle(rootBundle);
       if (!this.isRootBundleValid(rootBundle, nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock)) {
         continue;
@@ -257,7 +247,7 @@ export class HubPoolClient {
   // TODO: This might not be necessary since the cumulative root bundle count doesn't grow fast enough, but consider
   // using _.findLast/_.find instead of resorting the arrays if these functions begin to take a lot time.
   getProposedRootBundlesInBlockRange(startingBlock: number, endingBlock: number): ProposedRootBundle[] {
-    return this.reverseProposedRootBundles.filter(
+    return this.proposedRootBundles.filter(
       (bundle: ProposedRootBundle) => bundle.blockNumber >= startingBlock && bundle.blockNumber <= endingBlock
     );
   }
@@ -279,13 +269,15 @@ export class HubPoolClient {
   }
 
   getFollowingRootBundle(currentRootBundle: ProposedRootBundle): ProposedRootBundle {
-    const index = this.reverseProposedRootBundles.findIndex(
+    const index = _.findLastIndex(
+      this.proposedRootBundles,
       (bundle) => bundle.blockNumber === currentRootBundle.blockNumber
     );
-    if (index === 0) {
+    // If index of current root bundle is not found or is the last bundle, return undefined.
+    if (index === -1 || index === this.proposedRootBundles.length - 1) {
       return undefined;
     }
-    return this.reverseProposedRootBundles[index - 1];
+    return this.proposedRootBundles[index + 1];
   }
 
   getExecutedLeavesForRootBundle(
@@ -315,7 +307,7 @@ export class HubPoolClient {
   getLatestFullyExecutedRootBundle(latestMainnetBlock: number): ProposedRootBundle | undefined {
     // Search for latest ProposeRootBundleExecuted event followed by all of its RootBundleExecuted event suggesting
     // that all pool rebalance leaves were executed. This ignores any proposed bundles that were partially executed.
-    return this.reverseProposedRootBundles.find((rootBundle: ProposedRootBundle) => {
+    return _.findLast(this.proposedRootBundles, (rootBundle: ProposedRootBundle) => {
       if (rootBundle.blockNumber > latestMainnetBlock) {
         return false;
       }
@@ -538,17 +530,13 @@ export class HubPoolClient {
         .map((event) => spreadEventWithBlockNumber(event) as ExecutedRootBundle)
     );
 
-    // TODO: Consider storing all of the above methods in descending order for reducing the number of times
-    // you need to resort the events. As a shortcut, store an extra set of events sorted in reverse chronological order.
-    this.reverseProposedRootBundles = sortEventsDescending(this.proposedRootBundles);
-
     // If the contract's current rootBundleProposal() value has an unclaimedPoolRebalanceLeafCount > 0, then
     // it means that either the root bundle proposal is in the challenge period and can be disputed, or it has
     // passed the challenge period and pool rebalance leaves can be executed. Once all leaves are executed, the
     // unclaimed count will drop to 0 and at that point there is nothing more that we can do with this root bundle
     // besides proposing another one.
     if (pendingRootBundleProposal.unclaimedPoolRebalanceLeafCount > 0) {
-      const mostRecentProposedRootBundle = this.reverseProposedRootBundles[0];
+      const mostRecentProposedRootBundle = this.proposedRootBundles[this.proposedRootBundles.length - 1];
       this.pendingRootBundle = {
         poolRebalanceRoot: pendingRootBundleProposal.poolRebalanceRoot,
         relayerRefundRoot: pendingRootBundleProposal.relayerRefundRoot,
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -25,6 +25,7 @@ export class HubPoolClient {
   private l1Tokens: L1Token[] = []; // L1Tokens and their associated info.
   private lpTokens: { [token: string]: LpToken } = {};
   private proposedRootBundles: ProposedRootBundle[] = [];
+  private reverseProposedRootBundles: ProposedRootBundle[] = [];
   private canceledRootBundles: CancelledRootBundle[] = [];
   private disputedRootBundles: DisputedRootBundle[] = [];
   private executedRootBundles: ExecutedRootBundle[] = [];
@@ -226,26 +227,24 @@ export class HubPoolClient {
     chainIdList: number[]
   ): number | undefined {
     let endingBlockNumber: number | undefined;
-    for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+    // Search proposed root bundles in reverse chronological order.
+    for (const rootBundle of this.reverseProposedRootBundles) {
       const nextRootBundle = this.getFollowingRootBundle(rootBundle);
       if (!this.isRootBundleValid(rootBundle, nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock)) {
         continue;
       }
 
+      // 0 is the default value bundleEvalBlockNumber.
       const bundleEvalBlockNumber = this.getBundleEndBlockForChain(
         rootBundle as ProposedRootBundle,
         chain,
         chainIdList
       );
 
-      // If chain list doesn't contain chain, then bundleEvalBlockNumber returns 0 and the following check
-      // always fails.
-      if (bundleEvalBlockNumber >= block) {
-        endingBlockNumber = bundleEvalBlockNumber;
-        // Since events are sorted from oldest to newest, and bundle block ranges should only increase, exit as soon
-        // as we find the first block range that contains the target block.
-        break;
-      }
+      // Since we're iterating from newest to oldest, bundleEvalBlockNumber is only decreasing, and if the
+      // bundleEvalBlockNumber is smaller than the target block, then we should return the last set `endingBlockNumber`.
+      if (bundleEvalBlockNumber <= block) break;
+      endingBlockNumber = bundleEvalBlockNumber;
     }
     return endingBlockNumber;
   }
@@ -253,7 +252,7 @@ export class HubPoolClient {
   // TODO: This might not be necessary since the cumulative root bundle count doesn't grow fast enough, but consider
   // using _.findLast/_.find instead of resorting the arrays if these functions begin to take a lot time.
   getProposedRootBundlesInBlockRange(startingBlock: number, endingBlock: number): ProposedRootBundle[] {
-    return sortEventsDescending(this.proposedRootBundles).filter(
+    return this.reverseProposedRootBundles.filter(
       (bundle: ProposedRootBundle) => bundle.blockNumber >= startingBlock && bundle.blockNumber <= endingBlock
     );
   }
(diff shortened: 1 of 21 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -202,33 +203,31 @@ export class HubPoolClient {
     chainIdList: number[]
   ): number | undefined {
     let endingBlockNumber: number | undefined;
-    for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+    // Search proposed root bundles in reverse chronological order.
+    for (const rootBundle of this.reverseProposedRootBundles) {
```
nicholaspai, line 207: This function usually runs once per slow fill in previous bundle, so this would have added 500ms * 250 slow fills = 125 seconds to the dataworker run to the dataworker run to construct that bundle.

src/clients/SpokePoolClient.ts
```diff
@@ -47,7 +47,13 @@ const FILL_DEPOSIT_COMPARISON_KEYS = [
 ] as const;
 
 // These events are numerous and we should use a fixed lookback when querying them.
-const FIXED_LOOKBACK_EVENTS = ["FundsDeposited", "FilledRelay", "RelayedRootBundle", "ExecutedRelayerRefundRoot"];
+const FIXED_LOOKBACK_EVENTS = [
```
nicholaspai, line 50: Really old instances of these events are not needed in the Dataworker so we can save some time in `update` by fixing lookback for every event that is expected to be numerous. So this means that the only event that we'll query from the `deploymentBlock` is `EnabledDepositRoute` which is expected to be emitted very rarely

src/clients/SpokePoolClient.ts
```diff
@@ -263,29 +269,50 @@ export class SpokePoolClient {
...
     let low = initLow;
     let high = initHigh;
+    let i = 0;
     do {
       const mid = Math.floor((high + low) / 2);
       const searchedDepositId = await this.spokePool.numberOfDeposits({ blockTag: mid });
+
+      // Caller can set maxSearches to minimize number of binary searches and eth_call requests.
+      // Caller then needs to make a subsequent search between the latest [low, high].
+      if (i++ >= maxSearches) return { low, mid, high };
```
nicholaspai, line 298: Using `maxSearches` can speed up `queryHistoricalDepositForFill` by 50% (about 1000ms) per query for deposit that is close to the beginning of the spoke pool client's event search config. For much older deposits it doesn't seem to be as useful but that isn't the normal use case

src/scripts/validateRootBundle.ts
```diff
@@ -173,12 +178,14 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet): Pro
...
   );
+
+  // Get widest possible block range that could be used at time of root bundle proposal.
   const widestPossibleBlockRanges = await getWidestPossibleExpectedBlockRange(
     dataworker.chainIdListForBundleEvaluationBlockNumbers,
     spokePoolClients,
     getEndBlockBuffers(dataworker.chainIdListForBundleEvaluationBlockNumbers, dataworker.blockRangeEndBlockBuffer),
     clients,
-    priceRequestBlock,
+    mainnetBundleEndBlock,
```
nicholaspai, line 188: This was a bug fix. We want to always find the expected start block for a root bundle proposal at the time of the `rootBundle` that we are validating. So we should pass in the latest block to search for the bundle start block as the end block of the `rootBundle`. This way it is impossible that `getWidestPossibleExpectedBlockRange` will set its start block equal to `rootBundle.bundleEndBlock +`, which is possible sometimes with an unlucky `priceRequestBlock` setting.

src/scripts/validateRunningBalances.ts
```diff
@@ -255,7 +255,7 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
           // Not many bundles are expected to have slow fills so we can load them as necessary.
           const { slowFills, bundleSpokePoolClients } = await _constructSlowRootForBundle(
             mostRecentValidatedBundle,
-            validatedBundles[x + 1 + 1],
+            validatedBundles[x + 1 + 2],
```
nicholaspai, line 258: When constructing a spoke pool client to try to find slow fills, set its start blocks equal to end blocks from bundles 2 older than the target bundle, rather than 1. This matches other similar logic in reconstructing bundles in this script

src/clients/HubPoolClient.ts
```diff
@@ -21,6 +21,7 @@ export class HubPoolClient {
   private l1Tokens: L1Token[] = []; // L1Tokens and their associated info.
   private lpTokens: { [token: string]: LpToken } = {};
   private proposedRootBundles: ProposedRootBundle[] = [];
+  private reverseProposedRootBundles: ProposedRootBundle[] = [];
```
mrice32, line 24: Not super strongly opinionated, but instead of storing an extra array, have you considered using lodash methods to loop right to left:
- forEachRight: https://lodash.com/docs/4.17.15#forEachRight
- findLast: https://lodash.com/docs/4.17.15#findLast

Note: you could always write a for loop that loops from the last element to the first, but lodash seems like it introduces less bug risk.

src/clients/HubPoolClient.ts
```diff
@@ -202,33 +203,34 @@ export class HubPoolClient {
...
-      // If chain list doesn't contain chain, then bundleEvalBlockNumber returns 0 and the following check
-      // always fails.
-      if (bundleEvalBlockNumber >= block) {
-        endingBlockNumber = bundleEvalBlockNumber;
-        // Since events are sorted from oldest to newest, and bundle block ranges should only increase, exit as soon
-        // as we find the first block range that contains the target block.
+      // Since we're iterating from newest to oldest, bundleEvalBlockNumber is only decreasing, and if the
+      // bundleEvalBlockNumber is smaller than the target block, then we should return the last set `endingBlockNumber`.
+      if (bundleEvalBlockNumber <= block) {
+        if (bundleEvalBlockNumber === block) endingBlockNumber = bundleEvalBlockNumber;
```
mrice32, line 222: Do we need this if? Could 221 just be <, have it break in that if, and then let the === case hit the next loop where endingBlockNumber = bundleEvalBlockNumber for this loop?

Or is there an edge case I'm missing, maybe close to the current bundle or something?

src/clients/SpokePoolClient.ts
mrice32: Does this just add TokensBridged? Or does it change the behavior for the existing events in some way?

src/clients/SpokePoolClient.ts
```diff
@@ -306,23 +333,26 @@ export class SpokePoolClient {
...
-      ]);
-      assert(blockBeforeDeposit <= blockAfterDeposit, "blockBeforeDeposit > blockAfterDeposit");
-
+      // Binary search for block where SpokePool.numberOfDeposits === fill.depositId. Stop searches after a maximum
+      // # of searches to limit number of eth_call requests. Make an eth_getLogs call on the remaining block range
+      // (i.e. the [low, high] remaining from the binary search) to find the target deposit ID.
+      // @dev Limiting between 5-10 searches empirically performs best when there are ~300,000 deposits
+      // for a spoke pool and we're looking for a deposit <5 days older than HEAD.
+      const searchBounds = await this._binarySearchForBlockContainingDepositId(
+        fill.depositId,
```
mrice32, line 342: Is this guaranteed to give you a low, high bounds that always contain the increment block?

Consider the following possibility:
- The block where the deposit id goes from 4 to 5 is block 100.
- On the 5th iteration, low is passed in as 75, high is passed in as 125.
- The midpoint has a numDeposits of 5, which is > 4, so we make 99 the new high.
- We return low: 75, high 99. But he deposit we care about is in block 100, so we get no result.

Thoughts?

mrice32 (commented):
Looks great! A few questions/comments

src/clients/SpokePoolClient.ts
nicholaspai: yep it just adds it

src/clients/HubPoolClient.ts
nicholaspai: Don't lodash `findLast` methods have to re-construct the array?

src/clients/HubPoolClient.ts
nicholaspai: No you're totally right +1

src/clients/SpokePoolClient.ts
nicholaspai: huh good point this could failin some edge cases. Is it worth using `high = Math.min(this.latestBlock, searchBounds.high)`, `low = Math.max(this.spokePoolDeploymentBlock, searchBounds.low` or is there a cleaner way you think?

nicholaspai: What do you think of this solution? 44b1d34193f747af24aa1b4173e4236a51dc7ad8

mrice32: I think that probably works -- I don't think the low decrement is necessary, though.

I think some of the weirdness of the casework stems from the fact that we're binary searching for a block with deposit count x, but that's not what we really want. We really want to find the block that satisfies these conditions:
- the previous block had deposit count <= X
- this block has a deposit count > X.

Not necessarily advocating a specific change, just thinking out loud.

mrice32: This also feels like a good candidate for fuzz-style testing -- have we considered trying to mock out or abstract the contract call in this method so we can run heavy testing (with no blockchain interactions) on just the binary search itself to ensure it does what we expect in all cases?

mrice32 (approved):
Just remaining comments on the binary search!

src/clients/SpokePoolClient.ts
nicholaspai: ill add some fuzzing

src/clients/HubPoolClient.ts
```diff
@@ -202,33 +193,34 @@ export class HubPoolClient {
     chainIdList: number[]
   ): number | undefined {
     let endingBlockNumber: number | undefined;
-    for (const rootBundle of sortEventsAscending(this.proposedRootBundles)) {
+    // Search proposed root bundles in reverse chronological order.
+    for (let i = this.proposedRootBundles.length - 1; i >= 0; i--) {
```
james-a-morris, line 197: Since we removed `sortEventsAscending(this.proposedRootBundles)` is the `this.proposedRootBundles` guaranteed to be in ascending order?

src/clients/HubPoolClient.ts
```diff
@@ -202,33 +193,34 @@ export class HubPoolClient {
...
+      const rootBundle = this.proposedRootBundles[i];
       const nextRootBundle = this.getFollowingRootBundle(rootBundle);
       if (!this.isRootBundleValid(rootBundle, nextRootBundle ? nextRootBundle.blockNumber : latestMainnetBlock))
```
james-a-morris, line 200: Nit: could we reduce this logic into `nextRootBundle?.blockNumber ?? latestMainnetBlock`

src/clients/HubPoolClient.ts
```diff
@@ -246,17 +238,20 @@ export class HubPoolClient {
...
   getFollowingRootBundle(currentRootBundle: ProposedRootBundle) {
-    return sortEventsAscending(this.proposedRootBundles).find(
-      (_rootBundle: ProposedRootBundle) => _rootBundle.blockNumber > currentRootBundle.blockNumber
-    ) as ProposedRootBundle;
+    const index = _.findLastIndex(
+      this.proposedRootBundles,
+      (rootBundle) => rootBundle.blockNumber === currentRootBundle.blockNumber
+    );
+    if (index === this.proposedRootBundles.length - 1) return undefined;
+    return this.proposedRootBundles[index + 1];
```
james-a-morris, line 250: Could we do something like:

```
return this.proposedRootBundles.find((, ind, arr) => ind > 0 && arr[ind-1].blockNumber === currentRootBundle.blockNumber);
```

james-a-morris (commented):
Left a few comments. This looks great!

src/clients/SpokePoolClient.ts
nicholaspai: Fuzzing here: cb4863498b2422cab9094e5267c7fee290182ade

I think I set this up correctly but its a bit weird with how `depositIds` increment.

Also, I think searching for depositId + 1 instead of depositId might be more correct. We want to find the block where depositId incremented from depositId to depositId + 1. `high` can be any block where `id at block = depositId + 1 (or higher)` and `low` can be any block where `id at block = deposit (or lower)`. So I think we wouldn't need to modify the binary search's returned `low/high` if we search for depositId + 1. Does that make sense..?

src/clients/HubPoolClient.ts
nicholaspai: yep we can assume that since we query the RPC's in chronological order for events and we push into an array that way

src/clients/HubPoolClient.ts
nicholaspai: we definitely want to search from right to left, hence the `.findLast` call, but I do think we can condense this. Any ideas with the constraint we must go from right to left?

src/clients/SpokePoolClient.ts
nicholaspai: Hm thinking out loud, searching for `depositId+1` might result in `low` being too high, where `depositId` at `low` is already `depositId+1` so we miss the block where depositId increments to depositId+1.

Perhaps we need an asymmetrical binary search rule like in this commit 625f87212cca552715009bda9d09b7843cb92ade

nicholaspai: Thinking more outloud, the previous algorithm is kind of strange because it will return some `mid` where the deposit counter is either the target or target + 1. That's kind of weird.

I'm reframing this whole algorithm as a "binary search for deposit ID" to a "fetch block range bounds for for deposit ID". The new invariant is that `depositId at low <= target` and `depositId at high < target`

mrice32: I agree, although, I think you need to flip it for high. target < high. I think this works as long as you search _inclusively_, meaning you include low and high in your search (note: you could actually make it exclusive on the lower end, but it's simpler to just be inclusive on both sides).

test/mocks/MockSpokePoolClient.ts
```diff
@@ -0,0 +1,22 @@
+import { SpokePoolClient } from "../../src/clients";
+
+export class MockedSpokePoolClient extends SpokePoolClient {
+  // Each index in the array is assumed to be a block number, with the mocked deposit ID at the block heigh
+  // as the value.
+  public depositIds: number[] = [];
```
mrice32, line 6: The naming here is a little misleading!

This is the deposit _count_ at that block. So the number that is returned should be 1 higher than the most recent deposit id. I don't think there's anything technically wrong with the code, but consider the following:
- Let's say the value in the array is 5 at index 100 and 6 at index 101.
- This means that depositId 5 was made in block 101!

Labeling them depositIds just makes it seem like that's the deposit id that was sent in that block.

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -256,62 +254,82 @@ describe("SpokePoolClient: Fill Validation", async function () {
...
+    const depositIds = Array(initHigh - initLow + 1).fill(0);
+
+    const testIterations = 100;
+    for (let i = 0; i < testIterations; i++) {
+      // Randomize deposit ID's between initLow and initHigh. ID's should only increase
+      // and will do so 50% of the time. The other 50% of the time they will stay the same.
+      for (let j = 1; j < depositIds.length; j++) {
+        const increment = Math.random() > 0.5 ? 1 : 0;
+        depositIds[j] = depositIds[j - 1] + increment;
+      }
```
mrice32, line 317: This is awesome!! However, I think we should consider increments by more than 1, since that's a real edge case that we probably want to handle. Thoughts on this code change to add that?

```suggestion
      for (let j = 1; j < depositIds.length; j++) {
        const increment = Math.max(0, Math.floor((Math.random() - 0.5) * 10));
        depositIds[j] = depositIds[j - 1] + increment;
      }
```

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -256,62 +254,82 @@ describe("SpokePoolClient: Fill Validation", async function () {
...
+
+    const initLow = 0;
+    const initHigh = 1000000;
+    const depositIds = Array(initHigh - initLow + 1).fill(0);
+
+    const testIterations = 100;
+    for (let i = 0; i < testIterations; i++) {
+      // Randomize deposit ID's between initLow and initHigh. ID's should only increase
+      // and will do so 50% of the time. The other 50% of the time they will stay the same.
+      for (let j = 1; j < depositIds.length; j++) {
```
mrice32, line 314: Thoughts on doing this before the outer loop? This would make this faster since we only need to initialize once.

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -256,62 +254,82 @@ describe("SpokePoolClient: Fill Validation", async function () {
...
+
+      // Randomize max # of searches.
+      const maxSearches = Math.floor(Math.random() * 19) + 1;
+      const results = await fuzzClient._getBlockRangeForDepositId(target, initLow, initHigh, maxSearches);
+
+      expect(results.low >= initLow).to.be.true;
+      expect(results.high <= initHigh).to.be.true;
+      expect(results.mid <= results.high && results.mid >= results.low).to.be.true;
+      expect(depositIds[results.low] <= target).to.be.true;
+      expect(depositIds[results.high] > target).to.be.true;
```
mrice32, line 331: Can we add one more check to actually find the exact block and make sure it's in those bounds?
```suggestion
      expect(depositIds[results.high] > target).to.be.true;
      // The correct block is the first block to increment beyond the target.
      const correctBlock = depositIds.find(depositId => depositId > target);
      expect(correctBlock <= results.high).to.be.true;
      expect(results.low <= correctBlock).to.be.true;
```

mrice32 (approved):
Looks great! Only minor comments

src/clients/HubPoolClient.ts
james-a-morris: Fair point! Could we include something like below in the utils to accomplish this in-place?

```
/**
 * Maps over an array in reverse order.
 * @notice Maps over an array in reverse order.
 * @param {T[]} array Array to map over.
 * @param {(value: T, index: number, array: T[]) => U} callback Function to call on each element of the array.
 * @returns {U[]} Array of mapped values.
 */
export function mapReverse<T, U>(array: T[], callback: (value: T, index: number, array: T[]) => U): U[] {
  const result: U[] = [];
  for (let i = array.length - 1; i >= 0; i--) {
    result.push(callback(array[i], i, array));
  }
  return result;
}
```
```
/**
 * Finds the last element in an array that satisfies the provided testing function.
 * @notice Finds the last element in an array that satisfies the provided testing function.
 * @param {T[]} array Array to find element in.
 * @param {(value: T, index: number, obj: T[]) => boolean} callback Function to call on each element of the array.
 * @returns {T | undefined} The last element in the array that satisfies the provided testing function. Otherwise, undefined.
 */
export function findReverse<T>(array: T[], callback: (value: T, index: number, obj: T[]) => boolean): T | undefined {
  for (let i = array.length - 1; i >= 0; i--) {
    if (callback(array[i], i, array)) return array[i];
  }
  return undefined;
}
```

james-a-morris (commented):
Left a comment: https://github.com/across-protocol/relayer-v2/pull/517#discussion_r1155973582

src/clients/SpokePoolClient.ts
nicholaspai: this is a good point, the inclusivity definitely is needed when you can have multiple deposits in the same block

test/mocks/MockSpokePoolClient.ts
nicholaspai: good point, renamed to `depositIdAtBlock`

test/SpokePoolClient.ValidateFill.ts
nicholaspai: Ahh yes this exposed some bugs, nice catch

test/SpokePoolClient.ValidateFill.ts
nicholaspai: This is a good catch, have refactored the tests a bit

nicholaspai:
@mrice32 have reworked and simplified the "binary search" algo a bit. I think I just had to get away from the fact that this wasn't a true binary search and required a bespoke algorithm PTAL!

mrice32 (approved):
LGTM!

PR #518, opened 2023-03-25
improve: Fix lookback for SpokePool relayed root-related events

nicholaspai:
These events increase with every bundle so I imagine the lookback is starting to take up noticeable time


james-a-morris (approved):
LGTM

PR #522, opened 2023-03-28
improve: Set default NODE_MAX_CONCURRENCY to 25

nicholaspai:


pxrl (approved):
LGTM!

src/utils/ProviderUtils.ts
```diff
@@ -430,7 +430,7 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
   const nodeQuorumThreshold = Number(process.env[`NODE_QUORUM_${chainId}`] || NODE_QUORUM || "1");
 
   // Default to a max concurrency of 1000 requests per node.
```
pxrl, line 432:
```suggestion
```

pxrl: ^^ We can probably live without the comment, right? It's so easy to change the defaults without updating them! >.<

PR #523, opened 2023-03-28
refactor(SpokePoolClient): Split update() and add validation

pxrl:
This started as a driveby change, but it now rearranges some furniture and performs
additional validation of data received from the RPC provider on the SpokePool chain:
 - Split `update()`, such that the parts specifically querying the RPC are performed
 a single function: `_update()`. This significantly reduces the length of `update()`
and hopefully makes it easier to reason about when it's OK to conclude that
SpokePoolClient updating was successful.
 - Be more strict when validating the responses to:
   - spokePool.numberOfDeposits()
   - spokePool.getCurrentTime()
   - provider.getBlockNumber()
 - Generate array of known event names once in the constructor, not on each call to
   `update()`.

As a side benefit, because _update() is not split off into a separate method, it will be
easy to mock it out in test and inject a variety of data to verify the remaining update()
behaviour.


nicholaspai (dismissed):

src/clients/SpokePoolClient.ts
```diff
@@ -409,12 +409,10 @@ export class SpokePoolClient {
     );
 
     // Require that all Deposits meet the minimum specified number of confirmations.
-    const [latestBlockNumber, currentTime] = await Promise.all([
+    [ this.latestBlockNumber, this.currentTime ] = await Promise.all([
```
pxrl, line 412:
```suggestion
    [this.latestBlockNumber, this.currentTime] = await Promise.all([
```

src/clients/SpokePoolClient.ts
```diff
@@ -409,12 +409,10 @@ export class SpokePoolClient {
     );
 
     // Require that all Deposits meet the minimum specified number of confirmations.
-    const [latestBlockNumber, currentTime] = await Promise.all([
+    [this.latestBlockNumber, this.currentTime] = await Promise.all([
       this.spokePool.provider.getBlockNumber(),
       this.spokePool.getCurrentTime(),
     ]);
-    this.latestBlockNumber = latestBlockNumber;
-    this.currentTime = currentTime;
```
james-a-morris, line 416: I don't know if this addition makes the code more efficient/readable. If the first element in `Promise.all` fails, is the `this.latestBlockNumber` updated?

I can pre-approve it since I'm sure the code will work fine 🙂

src/clients/SpokePoolClient.ts
```diff
@@ -409,12 +409,10 @@ export class SpokePoolClient {
     );
 
     // Require that all Deposits meet the minimum specified number of confirmations.
-    const [latestBlockNumber, currentTime] = await Promise.all([
+    [this.latestBlockNumber, this.currentTime] = await Promise.all([
```
mrice32, line 412: I agree this is equivalent to today's code, but do we want to set the `this` values at the end of the update rather than the beginning? Are there concurrency concerns? Is it possible that users may think that the client has finished the update when it actually hasn't?

src/clients/SpokePoolClient.ts
pxrl: > If the first element in Promise.all fails, is the this.latestBlockNumber updated?

These calls both ripple quickly down to the provider layer, where we have retry logic and configurable redundancy. If they fail to resolve then there's actually not much we can do up at the SpokePoolClient, and we should probably just permit any underlying exceptions to be escalated (ultimately terminating the program). It should result in an exception being thrown, so we'd typically just bail out here.

As for whether would even want to continue with only a subset of the data we requested, I'd argue that's undesirable - in the SpokePoolClient we basically want all or nothing.

One thing we might consider up here is to apply some additional sanity checks on the data received - i.e. to ensure that the numbers actually increased (or at least remained the same). They should never go backwards, and there are other invariants that should be satisfied too.

src/clients/SpokePoolClient.ts
pxrl: > Are there concurrency concerns? Is it possible that users may think that the client has finished the update when it actually hasn't?

I do think there is potential for concurrency problems to arise. I hadn't really intended to resolve any of that in this change, but it definitely makes sense to get them in whilst we're lifting the hood on the SpokePoolClient for the UBA upgrades.

> I agree this is equivalent to today's code, but do we want to set the this values at the end of the update rather than the beginning?

SpokePoolClient.update() needs to be atomic, so applying that pattern would be an improvement :+1:. It'll require a bit more work to implement it consistently across the method, but I'm happy to work in that direction as part of this ongoing work.

pxrl:
Have moved this in a different direction - I'm now applying some basic sanity checks to the data received before saving it. Will see how the tests like it; might apply some additional changes based on how that goes.

src/clients/SpokePoolClient.ts
```diff
@@ -413,8 +413,10 @@ export class SpokePoolClient {
       this.spokePool.provider.getBlockNumber(),
       this.spokePool.getCurrentTime(),
     ]);
-    this.latestBlockNumber = latestBlockNumber;
-    this.currentTime = currentTime;
+    if (latestBlockNumber < this.latestBlockNumber)
+      throw new Error(`SpokePoolClient::update: latestBlockNumber ${latestBlockNumber} < ${this.latestBlockNumber}`);
+    else if (currentTime < this.currentTime)
+      throw new Error(`SpokePoolClient::update: currentTime ${currentTime} < ${this.currentTime}`);
```
pxrl, line 419: We might want retry logic on update() in the event of this happening, since it might mean that RPC provider instability/de-sync renders the bots unable to complete a run.

src/clients/SpokePoolClient.ts
```diff
@@ -65,7 +65,7 @@ export class SpokePoolClient {
   public lastDepositIdForSpokePool = Number.MAX_SAFE_INTEGER;
   public isUpdated = false;
   public firstBlockToSearch: number;
-  public latestBlockNumber: number | undefined;
+  public latestBlockNumber = 0;
```
nicholaspai, line 68: changing this from being possibly `undefined` to 0 seems dangerous

pxrl: Hm, could be...I'll take a closer look :+1:.

pxrl: Just circling back to this comment. I've checked the usage of SpokePoolClient.latestBlockNumber and I haven't found any instances where it'd be a problem for it to not to be undefined. Additionally, there are several parts of the codebase that assume it's numeric and do not handle the potential undefined case. Coupled with the fact that we expect all users of the SpokePoolClient to have performed an `update()` call prior to accessing any of its other members of the SpokePoolClient, and I _think_ we should be fine with a number here.

src/clients/SpokePoolClient.ts
```diff
@@ -465,21 +470,33 @@ export class SpokePoolClient {
...
+  }
+
+  async update(eventsToQuery: string[] = this.queryableEventNames): Promise<void> {
+    if (this.configStoreClient !== null && !this.configStoreClient.isUpdated) throw new Error("RateModel not updated");
+
+    const queryResults = await this._update(eventsToQuery);
+    if (queryResults.length === 0) {
+      this.isUpdated = true;
+      return;
+    }
```
mrice32, line 499: why do this out here? Why not just set it along with the other variables (this.currentTime, etc) inside the _update() function where it's clearer whether the call made it to the end?

src/clients/SpokePoolClient.ts
```diff
@@ -413,35 +416,37 @@ export class SpokePoolClient {
...
-    // Require that all Deposits meet the minimum specified number of confirmations.
     const [latestBlockNumber, currentTime] = await Promise.all([
       this.spokePool.provider.getBlockNumber(),
       this.spokePool.getCurrentTime(),
     ]);
-    this.latestBlockNumber = latestBlockNumber;
-    this.currentTime = currentTime;
+    if (isNaN(latestBlockNumber) || latestBlockNumber < this.latestBlockNumber)
+      throw new Error(`SpokePoolClient::update: latestBlockNumber ${latestBlockNumber} < ${this.latestBlockNumber}`);
+    else if (isNaN(currentTime) || currentTime < this.currentTime)
```
mrice32, line 435: Just double checking, what's the type of currentTime? Is it BigNumberIsh or is it number?

mrice32 (commented):
Looks good, just a few comments!

src/clients/SpokePoolClient.ts
```diff
@@ -51,7 +52,7 @@ const FILL_DEPOSIT_COMPARISON_KEYS = [
 const FIXED_LOOKBACK_EVENTS = ["FundsDeposited", "FilledRelay", "RelayedRootBundle", "ExecutedRelayerRefundRoot"];
 
 export class SpokePoolClient {
-  private currentTime: number;
+  private currentTime = 0;
```
james-a-morris, line 55: Does it make sense to define this in the constructor? Also, could we use something like:
```suggestion
  private currentTime = Math.floor(Date.now() / 1000);
```

src/clients/SpokePoolClient.ts
```diff
@@ -60,13 +61,14 @@ export class SpokePoolClient {
...
   private earlyDeposits: FundsDepositedEvent[] = [];
+  private queryableEventNames: string[] = [];
   public earliestDepositIdQueried = Number.MAX_SAFE_INTEGER;
   public latestDepositIdQueried = 0;
   public firstDepositIdForSpokePool = Number.MAX_SAFE_INTEGER;
   public lastDepositIdForSpokePool = Number.MAX_SAFE_INTEGER;
   public isUpdated = false;
   public firstBlockToSearch: number;
-  public latestBlockNumber: number | undefined;
+  public latestBlockNumber = 0;
```
james-a-morris, line 71: Could we use `"latest"` and have the variable take a `string | number` datatype?

james-a-morris (commented):
Left some comments - this is good stuff!

src/clients/SpokePoolClient.ts
pxrl: I don't think so; latestBlockNumber is an important reference that indicates how far the SpokePoolClient has updated to; it must refer to a specific block because it's a synchronisation point for some other lookups.

src/clients/SpokePoolClient.ts
pxrl: Good you asked - it's actually a BigNumber. I've updated the validation and check explicitly for a BigNumber type, and I've made sure to perform the necessary conversions to number once the value is validated. Given that currentTime is a uint32, there should be no overflow concerns.

src/clients/SpokePoolClient.ts
```diff
@@ -445,37 +448,41 @@ export class SpokePoolClient {
     return sortEventsAscending(fills.filter((_fill) => filledSameDeposit(_fill, matchingFill)));
   }
 
-  async update(eventsToQuery?: string[]): Promise<void> {
-    if (this.configStoreClient !== null && !this.configStoreClient.isUpdated) {
-      throw new Error("RateModel not updated");
+  async _update(eventsToQuery: string[]): Promise<Event[][]> {
```
pxrl, line 451:
```suggestion
  protected async _update(eventsToQuery: string[]): Promise<Event[][]> {
```

src/clients/SpokePoolClient.ts
pxrl: I've corrected the identified issue in BaseAdapter.ts and have made another pass at identifying users of latestBlockNumber. I haven't been able to find any remaining checks for latestBlockNumber being undefined.

src/clients/SpokePoolClient.ts
nicholaspai: Yeah I think we might check for `0` in other places, but I do think this should always be referencing the `SpokePool` block timestamp. Another idea is since we check of `SpokePoolClient.isUpdated` in other places, we might want to set `isUpdated = SpokePool.getCurrentTime()` in the update method

src/clients/SpokePoolClient.ts
```diff
@@ -500,21 +507,35 @@ export class SpokePoolClient {
...
-    queryResults.forEach((events) => {
-      sortEventsAscendingInPlace(events);
-    });
+    queryResults.forEach((events) => sortEventsAscendingInPlace(events));
+
+    // Next iteration should start off from where this one ended.
+    this.currentTime = currentTime.toNumber(); // currentTime is uint32
+    this.latestBlockNumber = latestBlockNumber;
+    this.lastDepositIdForSpokePool = latestDepositIdForSpokePool;
+    this.firstBlockToSearch = searchConfig.toBlock + 1;
```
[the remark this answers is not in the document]
pxrl: This is a good observation...I implemented it in the inner method in order to avoid having to pass it back to the caller, but maybe that's just laziness. There are advantages to passing it back to the caller, since the caller might want the data directly (rather than querying class variables). I think it would also make mocking of the _update() method easier too.

src/clients/SpokePoolClient.ts
```diff
@@ -532,12 +576,12 @@ export class SpokePoolClient {
...
       ];
       const { earlyDeposits = [], depositEvents = [] } = groupBy(allDeposits, (depositEvent) => {
-        if (depositEvent.args.quoteTimestamp > this.currentTime) {
+        if (depositEvent.args.quoteTimestamp > currentTime) {
           const { args, transactionHash } = depositEvent;
           this.logger.debug({
             at: "SpokePoolClient#update",
             message: "Deferring early deposit event.",
-            currentTime: this.currentTime,
+            currentTime: currentTime,
```
pxrl, line 584:
```suggestion
            currentTime,
```

src/clients/bridges/BaseAdapter.ts
```diff
@@ -60,13 +60,13 @@ export class BaseAdapter {
   // Note: this must be called after the SpokePoolClients are updated.
   getUpdatedSearchConfigs(): { l1SearchConfig: EventSearchConfig; l2SearchConfig: EventSearchConfig } {
     // Update search range based on the latest data from corresponding SpokePoolClients' search ranges.
-    // This needs to be called before fetching any events because spokePoolClients need to be updated first so
-    // latestBlockNumber is defined.
     const l1LatestBlock = this.spokePoolClients[1].latestBlockNumber;
     const l2LatestBlock = this.spokePoolClients[this.chainId].latestBlockNumber;
-    if (l1LatestBlock === undefined || l2LatestBlock === undefined) {
-      throw new Error("Spoke pool clients don't have an updated block number");
+
```
pxrl, line 65:
```suggestion
```

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
pxrl: Implemented in https://github.com/across-protocol/relayer-v2/pull/523/commits/075ed016e1f8effd454095bcb9087aca920c5f29.

src/clients/SpokePoolClient.ts
pxrl: I also think that a timestamp could be a good idea, because `.isUpdated` (as it's currently used) isn't well-suited to looping mode. I think we can incorporate an update to this as part of the UBA work.

src/clients/SpokePoolClient.ts
pxrl: Confirmed - have checked for all instances of the use of SpokePoolClient.latestBlockNumber. In the majority of cases it's supplied directly to a numerical expression; there are a small number that evaluate it for correctness beforehand - those have been adjusted from `=== undefined` to `=== 0` as part of this PR.

mrice32 (approved):
LGTM!

PR #524, opened 2023-03-28
feat(RefundRequests): SpokePool RefundRequested beachhead

pxrl:
Initial introduction of the RefundRequested type into the SpokePoolClient, plus some
very basic tests. This is backwards compatible with the existing SpokePoolClient
implementation; additional changes will aim to maintain that until we need to branch
off for actual breaking changes (i.e. importing the new contracts-v2 repo).

Note that this change simply replicates the existing pattern established for pulling in
events. We'll probably want to adjust that in follow-up work.

Fixes: ACX-1003


src/clients/SpokePoolClient.ts
```diff
@@ -27,8 +27,9 @@ import {
   Deposit,
   DepositWithBlock,
   Fill,
-  SpeedUp,
   FillWithBlock,
+  RefundRequestWithBlock,
+  SpeedUp,
```
james-a-morris, line 32: We can probably move `SpeedUp` back to line 30 to prevent the diff

test/SpokePoolClient.RefundRequests.ts
```diff
@@ -0,0 +1,113 @@
...
+  originChainId,
+  repaymentChainId,
+  deploySpokePoolWithToken,
+  toBN,
+  toBNWei,
+} from "./utils";
+
+function randomInt(min = 0, max = Number.MAX_SAFE_INTEGER): number {
+  return Math.floor(Math.random() * (max - min) + min);
+}
```
james-a-morris, line 20: Can we use `_.random()` instead of this fn?

test/SpokePoolClient.RefundRequests.ts
```diff
@@ -0,0 +1,113 @@
...
+    const refundFilter = { fromBlock: deploymentBlock + 1, toBlock: latestBlockNumber - 1 };
+    const { filteredRequests = [], excludedRequests = [] } = groupBy(refundRequests, (refundRequest) => {
+      return refundRequest.blockNumber > deploymentBlock && refundRequest.blockNumber < latestBlockNumber
+        ? "filteredRequests"
+        : "excludedRequests";
+    });
+
+    expect(spokePoolClient.getRefundRequests()).to.have.deep.members(refundRequests);
+    expect(spokePoolClient.getRefundRequests(refundFilter)).to.have.deep.members(filteredRequests);
+    expect(spokePoolClient.getRefundRequests(refundFilter)).to.not.have.deep.members(excludedRequests);
```
james-a-morris, line 111: Nit: feel free to disregard, but the github UI line wrapping seems to have differing indentation levels for these various `expect` statements

test/SpokePoolClient.RefundRequests.ts
nicholaspai: What is _.random()? We use Math.random() throughout codebase so I think its safe to stick with that

nicholaspai (approved):

test/SpokePoolClient.RefundRequests.ts
james-a-morris: It would be the `lodash` variant. It has a similar definition of this function

src/clients/SpokePoolClient.ts
```diff
@@ -86,6 +94,7 @@ export class SpokePoolClient {
       FundsDeposited: this.spokePool.filters.FundsDeposited(),
       RequestedSpeedUpDeposit: this.spokePool.filters.RequestedSpeedUpDeposit(),
       FilledRelay: this.spokePool.filters.FilledRelay(),
+      // RefundRequested: this.spokePool.filters.refundRequested(), // @todo update contracts-v2
```
mrice32, line 97: Two comments on one line?

src/clients/SpokePoolClient.ts
```diff
@@ -133,6 +142,14 @@ export class SpokePoolClient {
     return this.getFills().filter((fill) => fill.blockNumber >= startingBlock && fill.blockNumber <= endingBlock);
   }
 
+  getRefundRequests(
+    filter: EventSearchConfig = { fromBlock: this.spokePoolDeploymentBlock, toBlock: this.latestBlockNumber }
```
mrice32, line 146: This is a bit different than the other interfaces -- why did you choose to pass in an event search config rather than individual filter parameters or just allowing the caller to filter?

mrice32 (approved):
Looks good! Only two minor comments

test/SpokePoolClient.RefundRequests.ts
pxrl: Weird...no idea what's going on there. I've checked the diff but I can't see anything unusual in the file :shrug:.

test/SpokePoolClient.RefundRequests.ts
pxrl: Happy to use the lodash variant - it seems to do what I need out of the box and allows me to delete some code :+1:.

src/clients/SpokePoolClient.ts
pxrl: Good question...

The pattern of having the SpokePoolClient perform filtering _before_ returning events is already established, so it seems like we prefer to have this (i.e. as opposed to having the caller filter the results). I'm less of a fan of having separate methods for this (i.e. one method to return the full dataset, and another method to return a filtered variant), so if we want filtering functionality then I'd prefer to merge it into a single method.

In terms of the use of the EventSearchConfig type, I liked the idea of passing in an object with filter criteria, and EventSearchConfig seemed relevant there. I've thought about it some more, and EventSearchConfig seems less-relevant because it contains the concept of block lookbacks, so using it here seems like a misuse. I have reverted to the use of raw number types for filtering instead.

One argument against the initial implementation that I made is that it imposed filtering overheads even if no filter was supplied. I've adjusted the implementation now so that it only filters if filter criteria was actually supplied - otherwise it will just return the full array.

Happy to hear if you have any preference here, or alternative suggestions :+1:

pxrl:
Have temporarily marked this as a draft because this PR is dependent on James' work being merged first: https://github.com/across-protocol/relayer-v2/pull/551.

test/ProfitClient.ConsiderProfitability.ts
pxrl, line 1: This is just incidental and will disappear once the "merge to" branch reverts to `master`.

pxrl:
FWIW this will be submitted first thing on Wednesday.

src/clients/SpokePoolClient.ts
```diff
@@ -133,6 +136,12 @@ export class SpokePoolClient {
     return this.getFills().filter((fill) => fill.blockNumber >= startingBlock && fill.blockNumber <= endingBlock);
   }
 
+  getRefundRequests(fromBlock?: number, toBlock?: number): RefundRequestWithBlock[] {
+    return fromBlock !== undefined && toBlock !== undefined
+      ? this.refundRequests.filter((request) => request.blockNumber >= fromBlock && request.blockNumber <= toBlock)
+      : this.refundRequests;
```
pxrl, line 142:
```suggestion
    return isNan(fromBlock) || isNaN(toBlock)
      ? this.refundRequests
      : this.refundRequests.filter((request) => request.blockNumber >= fromBlock && request.blockNumber <= toBlock);
```

PR #530, opened 2023-03-29
 improve(scripts): Improved logging for validateRunningBalances script

nicholaspai:
I think parts of or the entirety of this script can be used in some serverless function that can alert us when a dataworker has produced a bundle such that running balances are incorrect, or funds are in the wrong places/missing for some reason
- Crash script with error about missing relayed root, so caller can verify if a root message didn't get sent from L1 to L2 for some reason
- Log everything via logger to make it easier to integrate in other clients in future


src/scripts/validateRunningBalances.ts
```diff
@@ -285,37 +277,66 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
...
         // - A slow fill from a bundle more than 2 bundles ago has not been executed and has not been replaced
         // by a partial fill.
         // - A deposit from HubPool to Spoke took too long to arrive and those deposits are not trackable via
         // Transfer events where the sender is 0x0.
         let excess = toBN(tokenBalanceAtBundleEndBlock).add(netSendAmount).add(runningBalance);
 
         if (relayedRoot === undefined || relayedRoot[l2Token] === undefined) {
-          console.log(`- No relayed root for chain ID ${leaf.chainId} and token ${l2Token}`);
+          if (!netSendAmount.eq(0)) {
+            // We shouldn't get here for any bundle since we start with the i-1'th most recent bundle.
```
nicholaspai, line 290: I want to have the script loudly exit if a root hasn't been relayed when we are expecting one. This will allow us to catch dropped messages on Arbitrum for example, faster

james-a-morris (approved):
LGTM - I left this comment (https://github.com/across-protocol/relayer-v2/pull/530#discussion_r1155965435) - I think it may be worth looking into.

src/scripts/validateRunningBalances.ts
```diff
@@ -285,37 +277,66 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
...
+            // If so, then a relayed root message might have gotten stuck in a canonical bridge and we will
+            // want to know about it.
+            throw new Error(`No relayed root for chain ID ${leaf.chainId} and token ${l2Token}`);
```
pxrl, line 293: I could imagine a scenario where we are investigating multiple things, and we want to see the result of something that might happen _after_ this part in the code. In that case, it'd be frustrating for it to terminate prematurely.

Could it instead make sense to collect cases like these in a separate warnings variable, and then log an error (or throw) at the conclusion of the script?

src/scripts/validateRunningBalances.ts
```diff
@@ -285,37 +277,66 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
...
       }
     }
-    console.groupEnd();
+    logger.debug({
+      at: "validateRunningBalances#index",
+      message: `Bundle #${x} proposed at block ${mostRecentValidatedBundle.blockNumber}`,
+      mrkdwn,
+    });
+    // This script is often run locally, log it to console:
+    console.log(mrkdwn);
```
pxrl, line 317: Rather than sending the markdown to logger and console, should we just ensure that we have a console transport for the logger?

src/scripts/validateRunningBalances.ts
mrice32: Sorry, I don't quite understand, why do we only care about the case where netSendAmount != 0?

mrice32 (approved):
Looks good after @pxrl and @james-a-morris's comments are resolved!

mrice32 (approved):

src/scripts/validateRunningBalances.ts
nicholaspai: Because these are the times that a root bundle should have been relayed from Hub to Spoke via a canonical messaging bridge, so we should loudly emit when this message hasn't made its way over

src/scripts/validateRunningBalances.ts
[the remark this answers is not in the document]
nicholaspai: I think using the `logger` is fine but the logger compresses the `mrkdwn` so I added the `console.log` for convenience

src/scripts/validateRunningBalances.ts
nicholaspai: That's a good point but I think we should save that work for when we actually integrate this in other scripts so we don't overengineer for now until we know the intended use case

src/scripts/validateRunningBalances.ts
nicholaspai: The Logger's console transport isn't great it doesn't render the new lines and tabs well

PR #531, opened 2023-03-29
chore: refactor `common` and `clients` directories

james-a-morris:
This PR aims to refactor the `clients` directory and the `common` directory. To accomplish this, changes were made to the `tsconfig`, `prettierrc`, and `eslint` files.


src/monitor/Monitor.ts
```diff
@@ -114,7 +114,7 @@ export class Monitor {
           l1Token: l1Token.address,
           chainId: this.monitorConfig.hubPoolChainId,
           poolCollateralSymbol: this.clients.hubPoolClient.getTokenInfoForL1Token(l1Token.address).symbol,
-          utilization: toBN(utilization),
+          utilization: toBN(utilization.toString()),
```
pxrl, line 117: I applied this change after building this branch locally. It's a bit clunky to convert BigNumberish back to string, just to convert it back to BN...I think it's due to a slight incompatibility between our toBN() typing and ethers.BigNumberish. Maybe we want to tweak that, but toBN() is heavily used so I avoided it for now.

scripts/disableDepositRoutes.ts
```diff
@@ -7,8 +7,10 @@ import {
   constructSpokePoolClientsWithStartBlocks,
   updateClients,
 } from "../src/common";
-const args = require("minimist")(process.argv.slice(2), {
-  number: ["chainId"],
+
+import minimist from "minimist";
+const args = minimist(process.argv.slice(2), {
+  string: ["chainId"],
 });
```
pxrl, line 14:
```suggestion
const args = minimist(process.argv.slice(2), { string: ["chainId"] });
```

src/clients/bridges/BaseAdapter.ts
```diff
@@ -69,11 +83,16 @@ export class BaseAdapter {
     };
   }
 
-  getSearchConfig(chainId: number) {
+  getSearchConfig(chainId: number): {
+    fromBlock: number;
+    maxBlockLookBack?: number;
+    concurrency?: number;
+    toBlock?: number;
+  } {
```
pxrl, line 91:
```suggestion
  getSearchConfig(chainId: number): EventSearchConfig {
```

Requires: `import { EventSearchConfig } from "../../utils";`

nicholaspai (approved):

PR #535, opened 2023-03-30
chore(yarn.lock): Re-sync

pxrl:
I've noticed this change being included by dependabot when it proposes other, unrelated updates.

Am pulling it out here so it can be considered independently.


pxrl:
Was wondering about how this happened; it looks like yarn.lock became slightly out of sync in this commit: https://github.com/across-protocol/relayer-v2/commit/612a610cd5c6f2dd4f3182d4db891bde723e963a

james-a-morris (approved):

PR #537, opened 2023-03-30
improve(SpokePoolClient): Adjust searchConfig validation

pxrl:
- Relocate searchConfig validation so it follows immediately after
   constructing searchConfig.
 - Warn about instances of searchConfig.fromBlock > searchConfig.toBlock.

 Ultimately, I think that SpokePoolClient should throw an error if it is
 requested to update with an invalid searchConfig, but would like to
 detect any instances of it in production before applying that change.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -421,7 +421,7 @@ export class SpokePoolClient {
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
     if (searchConfig.fromBlock > searchConfig.toBlock) {
-      this.log("warn", `Invalid update() searchConfig.`, searchConfig);
+      this.log("warn", `Invalid update() searchConfig.`, { searchConfig });
       return; // If the starting block is greater than the ending block return.
     }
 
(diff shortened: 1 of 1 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -420,15 +420,17 @@ export class SpokePoolClient {
       toBlock: this.eventSearchConfig.toBlock || this.latestBlockNumber,
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
+    if (searchConfig.fromBlock > searchConfig.toBlock) {
+      this.log("warn", `Invalid update() searchConfig.`, searchConfig);
```
pxrl, line 424: Fix use of this.log():

```suggestion
      this.log("warn", `Invalid update() searchConfig.`, { searchConfig });
```

src/clients/SpokePoolClient.ts
```diff
@@ -420,15 +420,17 @@ export class SpokePoolClient {
       toBlock: this.eventSearchConfig.toBlock || this.latestBlockNumber,
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
+    if (searchConfig.fromBlock > searchConfig.toBlock) {
+      this.log("warn", `Invalid update() searchConfig.`, { searchConfig });
+      return; // If the starting block is greater than the ending block return.
+    }
```
james-a-morris, line 426: Great catch!

james-a-morris (approved):

src/clients/SpokePoolClient.ts
nicholaspai: nice yeah this can happen sometimes in the tests because the hardhat test network is weird

nicholaspai (approved):

PR #539, opened 2023-03-30
chore: Cleanup unused variable

pxrl:
Should have been removed as part of the following commit:
 - 8391525a430bb15432b89c267674ef3493177e92.


james-a-morris:
@pxrl would it make sense to append this into the `staging/refactor-ts` branch? I think it fits nicely with the scope of that effort

PR #541, opened 2023-03-30
refactor: :recycle: enable strict on `finalizer` directory

james-a-morris:
This change corrects for eslint and typescript strict in the finalizer directory


pxrl (approved):
LGTM

PR #544, opened 2023-03-31
refactor: :recycle: update `util` directory to comply with new tsconfig

james-a-morris:
This PR aims to bring the `util` functions within compliance of the new tsconfig & eslint config.

### DIFF
--- a/src/utils/ObjectUtils.ts
+++ b/src/utils/ObjectUtils.ts
@@ -1,7 +1,9 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 // Append value along the keyPath to object. For example assign(deposits, ['1337', '31337'], [{depositId:1}]) will create
 // deposits = {1337:{31337:[{depositId:1}]}}. Note that if the path into the object exists then this will append. This
 // function respects the destination type; if it is an object then deep merge and if an array effectively will push.
-export function assign(obj: any, keyPath: any[], value: any) {
+export function assign(obj: any, keyPath: any[], value: any): void {
   const lastKeyIndex = keyPath.length - 1;
   for (let i = 0; i < lastKeyIndex; ++i) {
     const key = keyPath[i];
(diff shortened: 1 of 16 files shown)


src/utils/ProviderUtils.ts
james-a-morris, line 1: This file makes heavy usage of `any` and I think we should open a new PR to specifically tackle this in the future. I would recommend a heavy amount of care.

src/utils/ObjectUtils.ts
james-a-morris, line 1: This file makes heavy usage of `any` and I think we should open a new PR to specifically tackle this in the future. I would recommend a heavy amount of care.

My first pass thoughts to resolve this is to use type `generics`

src/utils/TransactionPropBuilder.ts
```diff
@@ -23,7 +27,7 @@ export function buildFillRelayWithUpdatedFeeProps(
   deposit: Deposit,
   repaymentChainId: number,
   maxFillAmount: BigNumber
-) {
+): (string | number | BigNumber)[] {
```
pxrl, line 30: Candidate for BigNumberish ?

src/utils/TransactionUtils.ts
```diff
@@ -96,7 +97,15 @@ export async function runTransaction(
...
-export async function getGasPrice(provider: ethers.providers.Provider, priorityScaler = 1.2, maxFeePerGasScaler = 3) {
+export async function getGasPrice(
+  provider: ethers.providers.Provider,
+  priorityScaler = 1.2,
+  maxFeePerGasScaler = 3
+): Promise<{
+  maxFeePerGas?: ethers.BigNumber;
+  maxPriorityFeePerGas?: ethers.BigNumber;
+  gasPrice?: ethers.BigNumber;
+}> {
```
pxrl, line 108: ethers 5.7.0 defines a FeeData type for this: https://docs.ethers.org/v5/api/providers/types/#providers-FeeData

src/utils/TransactionPropBuilder.ts
james-a-morris: Yep! Good call

src/utils/TransactionUtils.ts
james-a-morris: SGTM! However I needed to make the return type a `Promsie<Partial<FeeData>>` since the fn doesn't always return the entire type.

PR #550, opened 2023-04-03
improve(test): Add fixed-value ProfitClient tests

pxrl:
The existence of the randomised route-specific fee is only a value-add, rather than the main test.


james-a-morris (approved):
LGTM - nice addition

PR #551, opened 2023-04-03
feat: use sdk types in relayer

james-a-morris:
This PR adds interface support to the `relayer-v2` from the `sdk-v2` package.


src/interfaces/index.ts
pxrl, line 1: This one seems a bit cumbersome to maintain :( Were there any other options for specifying it that you considered and discounted for any reason?

src/interfaces/InventoryManagement.ts
```diff
@@ -1,4 +1,4 @@
-import { BigNumber } from "../utils";
+import { BigNumber } from "ethers";
```
pxrl, line 1: Just to confirm, is this an incidental change?

I _believe_ there's an existing pattern of importing external dependencies in a single file, and then exporting it from there. I'm speculating, but I think the intention might be to ensure the rest of the codebase is always referencing the same dependencies. Unfortunately this isn't consistently adhered to in the code, and I'm also not sure whether it's desirable to maintain it, or just import external dependencies directly in the files where they're used.

Note that this is also evident in the use of some UMA utility functions - toBN() and toBNWei() spring to mind. In some cases we import them directly from an upstream UMA package, and in other cases we reference a local implementation. We should really decide on one or the other and clean it up.

@mrice32 any guidance on what you consider to be best practice here?

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/boba-sdk": "^1.0.4",
     "@across-protocol/contracts-v2": "^2.0.2",
-    "@across-protocol/sdk-v2": "^0.4.1",
+    "@across-protocol/sdk-v2": "^0.4.2",
```
pxrl, line 15:
```suggestion
    "@across-protocol/sdk-v2": "^0.4.3",
```

In order to incorporate the UBA types added in this commit: https://github.com/across-protocol/sdk-v2/commit/f4e6c4d429d3bf2c521b3b5d7a6def7e684c6a95.

src/interfaces/InventoryManagement.ts
pxrl: I'm OK for it to go in as-is so that we can keep moving, but let's pick it up as part of the broader discussion that we're having about the structure of the codebase :+1:

pxrl (approved):

PR #552, opened 2023-04-03
feat: add tsconfig strict and modify eslint

james-a-morris:
This PR performs various refactors across the relayer to improve developer experience and reduce errors from common linting patterns.


src/utils/ContractUtils.ts
```diff
@@ -26,10 +26,10 @@ export function castSpokePoolName(networkId: number): string {
...
 }
 
-export function getParamType(contractName: string, functionName: string, paramName: string) {
-  const artifact: any = typechain[`${[contractName]}__factory`];
-  const fragment = artifact.abi.find((fragment) => fragment.name === functionName);
-  return fragment!.inputs.find((input) => input.name === paramName) || "";
+export function getParamType(contractName: string, functionName: string, paramName: string): string {
+  const artifact = typechain[`${[contractName]}__factory`];
+  const fragment = artifact.abi.find((fragment: { name: string }) => fragment.name === functionName);
+  return fragment.inputs.find((input: { name: string }) => input.name === paramName) || "";
```
pxrl, line 32: Curious on this one - I presume fragment can theoretically be undefined, so it's interesting that we don't get a warning when unconditionally dereferencing it on line 32.

nb. I don't think this is too important, since this is likely something that will always fail or never fail, and it obviously works OK now.

james-a-morris: I left the `strictNullChecks: false` parameter to prevent a significant overhaul of logic. As a result, the `undefined` values are not considered warnings.

tasks/integration-tests.ts
```diff
@@ -8,15 +9,20 @@ import sinon from "sinon";
...
+  }
+
   const spyLogger = winston.createLogger({
     level: "debug",
     format: winston.format.combine(winston.format(bigNumberFormatter)(), winston.format.json()),
-    transports: [
-      new SpyTransport({ level: "debug" }, { spy }),
-      process.env.LOG_IN_TEST ? new winston.transports.Console() : null,
-    ].filter((n) => n),
+    transports: transports,
```
pxrl, line 25:
```suggestion
    transports,
```

pxrl (approved):
LGTM - lots of changes, but it gives some comfort knowing that we've been through them once before. I did spot a few minor points, but no showstoppers.

src/utils/TransactionUtils.ts
```diff
@@ -128,7 +134,15 @@ export async function willSucceed(transaction: AugmentedTransaction): Promise<Tr
   }
 }
 
-export function getTarget(targetAddress: string) {
+export function getTarget(targetAddress: string):
+  | {
```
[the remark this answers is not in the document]
james-a-morris: No worries! This is a way to put the union types on a separate line: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "skipLibCheck": true,
     "resolveJsonModule": true
   },
-  "include": ["./src", "index.ts"],
+  "strict": true,
```
mrice32, line 11: Does this work?

I think this option goes inside compilerOptions, no? Not sure what it does outside. Does it still compile if you move it into compilerOptions?

See https://github.com/tsconfig/bases/blob/7db25a41bc5a9c0f66d91f6f3aa28438afcb2f18/bases/node18.json#L10 as an example.

mrice32 (commented):
Just one comment on the tsconfig!

tsconfig.json
pxrl: I tried migrating it into compilerOptions but I get 246 errors trying to build. Placing it inside compilerOptions also seems to be correct [per the docs](https://www.typescriptlang.org/tsconfig#compilerOptions).

src/clients/SpokePoolClient.ts
```diff
@@ -34,6 +35,7 @@ import {
 } from "../interfaces";
 import { RootBundleRelayWithBlock, RelayerRefundExecutionWithBlock } from "../interfaces";
 
+
```
pxrl, line 38:
```suggestion
```

pxrl (changes requested):
Given the number of things that break when strict: true is migrated into the compilerOptions section, maybe we should just remove strict: true for now in order to get this in as-is and then incrementally tackle the remaining issues. I think it's an acceptable approach, because this PR _does_ implement a lot of good changes. Alternatively, it seems like this PR will need to remain open for a longer period, and we'll risk hitting more merge conflicts.

james-a-morris:
> Given the number of things that break when strict: true is migrated into the compilerOptions section, maybe we should just remove strict: true for now in order to get this in as-is and then incrementally tackle the remaining issues. I think it's an acceptable approach, because this PR _does_ implement a lot of good changes. Alternatively, it seems like this PR will need to remain open for a longer period, and we'll risk hitting more merge conflicts.

This is a fair point! I have a PR open to enable 5 of the 7 strict parameters https://github.com/across-protocol/relayer-v2/pull/559

tsconfig.json
james-a-morris: I have this resolved in this PR: https://github.com/across-protocol/relayer-v2/pull/559

pxrl (approved):
LGTM!

tsconfig.json
pxrl, line 11:
```suggestion
```

Can maybe leave this one out, given that it doesn't do what we thought, and this is being merged into master.

james-a-morris: Fair - there is an open PR which removes this and it's planned to be merged into this branch prior to this branch's merge into master.

pxrl: Ah nice - perfect.

src/clients/BundleDataClient.ts
```diff
@@ -182,7 +182,7 @@ export class BundleDataClient {
 
     const unfilledDepositsForOriginChain: UnfilledDepositsForOriginChain = {};
     const fillsToRefund: FillsToRefund = {};
-    const allRelayerRefunds: any[] = [];
+    const allRelayerRefunds: { repaymentChain: string; repaymentToken: string }[] = [];
```
mrice32, line 185: These sorts of changes are great! Going to make the codebase much safer. +1

nit: why use string for chainId?

src/utils/ContractUtils.ts
```diff
@@ -26,10 +26,10 @@ export function castSpokePoolName(networkId: number): string {
   return `${networkName}_SpokePool`;
 }
 
-export function getParamType(contractName: string, functionName: string, paramName: string) {
-  const artifact: any = typechain[`${[contractName]}__factory`];
-  const fragment = artifact.abi.find((fragment) => fragment.name === functionName);
-  return fragment!.inputs.find((input) => input.name === paramName) || "";
+export function getParamType(contractName: string, functionName: string, paramName: string): string {
+  const artifact = typechain[`${[contractName]}__factory`];
```
mrice32, line 30: ❤️

tasks/integration-tests.ts
```diff
@@ -8,15 +9,20 @@ import sinon from "sinon";
...
-export function createSpyLogger() {
+export function createSpyLogger(): {
+  spy: sinon.SinonSpy<any[], any>;
+  spyLogger: winston.Logger;
+} {
   const spy = sinon.spy();
+  const transports: winston.transport[] = [new SpyTransport({ level: "debug" }, { spy })];
+  if (process.env.LOG_IN_TEST) {
+    transports.push(new winston.transports.Console());
+  }
```
mrice32, line 20: +1 way more clear

mrice32 (commented):
Looks awesome! Only minor nits and questions! Big step in improving the state of the codebase!

scripts/unwrapWeth.ts
```diff
@@ -28,7 +28,7 @@ export async function run(): Promise<void> {
   if (!Object.keys(args).includes("amount")) throw new Error("Define `amount` as how much you want to unwrap");
   const baseSigner = await getSigner();
   const connectedSigner = baseSigner.connect(await getProvider(Number(args.chainId)));
-  const token = WETH_ADDRESSES[Number(args.chainId)];
+  const token = WETH_ADDRESSES[Number(args.chainId) as keyof typeof WETH_ADDRESSES];
```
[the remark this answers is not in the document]
james-a-morris: Resolved https://github.com/across-protocol/relayer-v2/pull/565

src/clients/BundleDataClient.ts
james-a-morris: Fair point! Resolved in a companion PR (that will be merged in here): https://github.com/across-protocol/relayer-v2/pull/566/commits/c20f84be403fb0f1570b2a861fd2e672a0e17f0c

.eslintrc.js
```diff
@@ -2,49 +2,27 @@ module.exports = {
...
-    "node/no-unsupported-features/es-syntax": ["error", { ignores: ["modules"] }],
-    // Disable warnings for { a, b, ...rest } variables, since this is typically used to remove variables.
-    "@typescript-eslint/no-unused-vars": ["warn", { ignoreRestSiblings: true }],
-    "chai-expect/missing-assertion": 2,
-  },
-  settings: {
-    node: {
-      tryExtensions: [".js", ".ts"],
-    },
+    "no-console": 0,
```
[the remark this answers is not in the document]
james-a-morris: Fair point - I've reverted the eslint back to the master branch. https://github.com/across-protocol/relayer-v2/pull/566/commits/62f9c2f4e243a965b238e941aab286f7389c0296

.eslintrc.js
```diff
@@ -2,49 +2,27 @@ module.exports = {
...
     "prettier/prettier": ["warn"],
     indent: 0, // avoid conflict with prettier's indent system
     "linebreak-style": ["error", "unix"],
     quotes: ["error", "double", { avoidEscape: true }],
     semi: ["error", "always"],
     "spaced-comment": ["error", "always", { exceptions: ["-", "+"] }],
-    "no-console": 2,
-    camelcase: "off",
-    "@typescript-eslint/camelcase": "off",
-    "mocha/no-exclusive-tests": "error",
```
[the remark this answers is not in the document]
james-a-morris: Similar to above: https://github.com/across-protocol/relayer-v2/pull/566/commits/62f9c2f4e243a965b238e941aab286f7389c0296

.eslintrc.js
```diff
@@ -2,49 +2,27 @@ module.exports = {
...
-    "@typescript-eslint/no-var-requires": "off",
```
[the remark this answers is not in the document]
james-a-morris: Similar to above: https://github.com/across-protocol/relayer-v2/pull/566/commits/62f9c2f4e243a965b238e941aab286f7389c0296

scripts/unwrapWeth.ts
```diff
@@ -1,9 +1,9 @@
-/* eslint-disable no-process-exit */
 import { ethers, getSigner, getProvider, WETH9, toBN } from "../src/utils";
 import { askYesNoQuestion } from "./utils";
-const args = require("minimist")(process.argv.slice(2), {
-  string: ["amount"],
-  number: ["chainId"],
```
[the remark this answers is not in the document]
james-a-morris: The official TS documents don't specify `number` as a valid option type. Here are the defined types:
```
declare namespace minimist {
    interface Opts {
        /**
         * A string or array of strings argument names to always treat as strings
         */
        string?: string | string[] | undefined;

        /**
         * A boolean, string or array of strings to always treat as booleans. If true will treat
         * all double hyphenated arguments without equals signs as boolean (e.g. affects `--foo`, not `-f` or `--foo=bar`)
         */
        boolean?: boolean | string | string[] | undefined;

        /**
         * An object mapping string names to strings or arrays of string argument names to use as aliases
         */
        alias?: { [key: string]: string | string[] } | undefined;

        /**
         * An object mapping string argument names to default values
         */
        default?: { [key: string]: any } | undefined;

        /**
         * When true, populate argv._ with everything after the first non-option
         */
        stopEarly?: boolean | undefined;

        /**
         * A function which is invoked with a command line parameter not defined in the opts
         * configuration object. If the function returns false, the unknown option is not added to argv
         */
        unknown?: ((arg: string) => boolean) | undefined;

        /**
         * When true, populate argv._ with everything before the -- and argv['--'] with everything after the --.
         * Note that with -- set, parsing for arguments still stops after the `--`.
         */
        '--'?: boolean | undefined;
    }
}
```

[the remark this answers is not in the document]
mrice32: Great find! I'm not sure how "number" ever worked!

TLDR: I think making this a string is the right call. +1!

I found this in their docs: https://github.com/minimistjs/minimist/blob/093bc85eb9b2a06f0d7fa3f7f1940cd135495abf/README.md?plain=1#L64.

Basically if the argument isn't labeled a string or bool and it looks like a number, it will treat it as a number. However, I think treating numbers as strings gives us much more type safety since we know it will always come as a string vs being conditional on the input.

src/clients/BundleDataClient.ts
mrice32: Sounds good!

.eslintrc.js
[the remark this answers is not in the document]
mrice32: I think changing this is totally fine!! I was only calling out potential issues that I saw! Not necessary to completely revert IMO. Ultimately up to you, but I do think there is value in simplicity of the config, which seems to be what you're pushing for 👍

mrice32 (approved):
LGTM!

PR #556, opened 2023-04-04
improve(SpokePoolClient): Optimise EnabledDepositRoute

pxrl:
Per current logic, we must search for EnabledDepositRoute events back to SpokePool deployment. Whether or not this is desirable for all bot instances, we can optimise it by dramatically increasing the block range over which we are searching. This should result in fewer RPC eth_getLogs() queries for this topic.


src/clients/SpokePoolClient.ts
```diff
@@ -440,8 +440,10 @@ export class SpokePoolClient {
       // By default, an event's query range is controlled by the `eventSearchConfig` passed in during instantiation.
       // However, certain events have special overriding requirements to their search ranges:
       // - EnabledDepositRoute: The full history is always required, so override the requested fromBlock.
-      if (eventName === "EnabledDepositRoute" && !this.isUpdated) {
-        _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
+      if (eventName === "EnabledDepositRoute") {
+        // EnabledDepositRoute events are rare, so we can dramatically increase the block range per query.
+        _searchConfig.maxBlockLookBack *= 10;
```
pxrl, line 445: Note: I plucked the 10x multiplier out of thin air.

Infura don't restrict on block ranges anymore; their constraints are:
 - Request must return within 10s.
 - No more than 10k events in the response.
 
Alchemy similarly impose:
 - Up to 2k block range with 150MB response limit, OR
 - Unlimited block range with max 10k events in the response.

So with both of the major providers, it seems fairly safe to apply a much larger multiplier.

pxrl: Per [this line](https://github.com/across-protocol/relayer-v2/blob/6a0273e078e005d5922d626cd9183ead45e12965/src/utils/EventUtils.ts#L40) in paginatedEventQuery(), we could also just set maxBlockLookBack to 0. Or alternatively, get the size of the range [fromBlock, toBlock] and divide it by some fixed integer in order to cap the complete lookup at _x_ requests.

mrice32: Are we sure that these restrictions are the same _on all chains_? Have you tested these larger queries on each provider?

I'm a little worried about this silently erroring and causing certain providers to just fail consistently on this query... especially if infura's limit is just based on a node timeout.

We don't actually use this information in that many places. I wonder if the right way to optimize this might be to try and look at whether we need to do the query in the relayer.

mrice32 (commented):
Just one comment!

src/clients/SpokePoolClient.ts
```diff
@@ -449,8 +449,10 @@ export class SpokePoolClient {
...
+        if (!this.isUpdated) _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
```
james-a-morris, line 455: This line is relatively important. Would it make sense from a clarity perspective to make the logic a bit more verbose?
```suggestion
        if (!this.isUpdated) {
            _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
        }
```

james-a-morris (commented):
Left one comment.

src/clients/SpokePoolClient.ts
james-a-morris: I'm unsure which exact chain this was (if it's Boba feel free to disregard this comment), but I distinctly remember at least one chain stipulated a 5K block number range during my work in the ETL project.

src/clients/SpokePoolClient.ts
pxrl: Yeah - I actually prefer to use braces but noticed that the team style is normally to keep things on a single line, if it fits. But definitely happy to apply it here :+1:

src/clients/SpokePoolClient.ts
nicholaspai: Yeah the original constraint was that some chains (maybe just Boba though but I think Polygon also) set a limit for their RPCs

nicholaspai (commented):
One comment

src/clients/SpokePoolClient.ts
```diff
@@ -487,7 +487,9 @@ export class SpokePoolClient {
       if (eventName === "EnabledDepositRoute") {
         // EnabledDepositRoute events are rare, so we can dramatically increase the block range per query.
         _searchConfig.maxBlockLookBack *= 10;
-        if (!this.isUpdated) _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
+        if (!this.isUpdated) {
+            _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
```
pxrl, line 491:
```suggestion
          _searchConfig.fromBlock = this.spokePoolDeploymentBlock;
```

pxrl:
Pulling this back to a draft for now.

pxrl:
Closing this for now...the problem has been temporarily mitigated since the migration to the new SpokePools. I'd like to circle back to it in future to find a better solution.

PR #557, opened 2023-04-05
improve(ProviderUtils): Redis cache TTL enhancements

pxrl:
Two changes:
 - Modify the configured cache TTL by up to 15% when setting new keys. This is intended to spread cache expiry more evenly over time in order to reduce expiry spikes.
 - Allow the minimum Redis cache TTL to be user-configured via the PROVIDER_CACHE_TTL env var. This is effectively now setting the average TTL.

Note that I'd prefer if ProviderUtils didn't read env vars directly, but it's the existing pattern established in that file, so I don't want to go on a yak shaving odyssey by implementing it now. I think we'll probably make that change separately.

Fixes ACX-1002


src/utils/ProviderUtils.ts
```diff
@@ -4,11 +4,18 @@ import winston from "winston";
...
+  MAX_REORG_DISTANCE,
+  PROVIDER_CACHE_TTL,
+  PROVIDER_CACHE_TTL_MODIFIER as ttl_modifier,
+  BLOCK_NUMBER_TTL,
+} from "../common";
 import { Logger } from ".";
 
 const logger = Logger;
 
+const _ttl = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
```
james-a-morris, line 17: Nit: does this potentially have the side-effect of causing a NaN cascade? If `process.env.PROVIDER_CACHE_TTL` is a non-numeric string, `_ttl` will be set to `NaN`. I'm not sure how `lodash.random` handles NaN values, but I figured I'd call it out for caution

james-a-morris (commented):
Left one comment 🙂

mrice32 (approved):
LGTM once @james-a-morris's comment is resolved!

src/utils/ProviderUtils.ts
pxrl: Added in https://github.com/across-protocol/relayer-v2/pull/557/commits/c8fc80f26fdf93684cf1085f67a23391a701785f. There's some repetition in the resolution of PROVIDER_CACHE_TTL, but since this error should never happen I'd rather not pollute the global namespace any further.

PR #558, opened 2023-04-05
fix(SpokePoolClient): lint

pxrl:
I introduced this in a previous commit and now it's showing up as a change in all PRs because lint-fix is auto-correcting it. Fix it here so it doesn't pollute others' PRs.


james-a-morris (approved):
Good call

PR #559, opened 2023-04-05
refactor: ♻️ enforce typescript strict

james-a-morris:
This PR adds typescript `strict` to the `tsconfig.json`. The following check constraints have been added:
1. [alwaysStrict](https://www.typescriptlang.org/tsconfig#alwaysStrict)
1. [strictBindCallApply](https://www.typescriptlang.org/tsconfig#strictBindCallApply)
1. [strictFunctionTypes](https://www.typescriptlang.org/tsconfig#strictFunctionTypes)
1. [strictPropertyInitialization](https://www.typescriptlang.org/tsconfig#strictPropertyInitialization)
1. [noImplicitThis](https://www.typescriptlang.org/tsconfig#noImplicitThis)
1. [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables)

To reduce the size of this PR, the following check constraints have been left explicitly disabled:
1. [noImplicitAny](https://www.typescriptlang.org/tsconfig#noImplicitAny)
1. [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks)


src/clients/ProfitClient.ts
```diff
@@ -214,7 +214,8 @@ export class ProfitClient {
     const netRelayerFeePct = netRelayerFeeUsd.mul(toBNWei(1)).div(relayerCapitalUsd);
 
     // If token price or gas cost is unknown, assume the relay is unprofitable.
-    const fillProfitable = tokenPriceUsd.gt(0) && gasCostUsd.gt(0) && netRelayerFeePct.gte(minRelayerFeePct);
+    const fillProfitable =
+      tokenPriceUsd.gt(0) && gasCostUsd.gt(0) && (!minRelayerFeePct || netRelayerFeePct.gte(minRelayerFeePct));
```
pxrl, line 218: I'm pretty sure we can change the definition of minRelayerFeePct on line 183, so that it's not optional anymore. I've checked the users of it and it seems to be unconditionally passed in, and I also don't think it should ever be undefined. Then we could revert the change made here.

src/clients/SpokePoolClient.ts
```diff
@@ -80,6 +80,7 @@ export class SpokePoolClient {
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 }
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
+    this.currentTime = getCurrentTime();
```
pxrl, line 83: currentTime refers to the on-chain timestamp supplied by SpokePool.getCurrentTime(), whereas getCurrentTime() just provides the local system time. In [this PR](https://github.com/across-protocol/relayer-v2/pull/523) I am proposing that we initialise currentTime to 0.

src/clients/SpokePoolClient.ts
```diff
@@ -280,8 +281,8 @@ export class SpokePoolClient {
   async binarySearchForBlockContainingDepositId(
     targetDepositId: number,
     initLow = this.spokePoolDeploymentBlock,
-    initHigh = this.latestBlockNumber
-  ): Promise<number> {
+    initHigh = this.latestBlockNumber as number
```
pxrl, line 284: In [this PR](https://github.com/across-protocol/relayer-v2/pull/523) I'm proposing that we initialise this.latestBlockNumber to 0 and remove the potential for it to be undefined. It _seems_ like we can do that, subject to some cleanup in tests.

src/clients/SpokePoolClient.ts
james-a-morris: Fair point! I've removed the change and will defer to your PR. With the removal of the `noImplicitAny` typecheck it's not considered an error anymore and the compiler will accept this current code block

src/interfaces/index.ts
pxrl, line 1: Nice!

src/interfaces/index.ts
```diff
@@ -3,6 +3,7 @@ import { interfaces } from "@across-protocol/sdk-v2";
 export * from "./InventoryManagement";
 export * from "./SpokePool";
 export * from "./Token";
+export * from "./Error";
```
pxrl, line 6: For follow-up, maybe we want to put this in the SDK.

pxrl (commented):
LGTM! Will give Matt the opportunity to comment on anything additional.

pxrl (approved):
OK from my end :+1:

scripts/unwrapWeth.ts
```diff
@@ -28,7 +28,7 @@ export async function run(): Promise<void> {
   if (!Object.keys(args).includes("amount")) throw new Error("Define `amount` as how much you want to unwrap");
   const baseSigner = await getSigner();
   const connectedSigner = baseSigner.connect(await getProvider(Number(args.chainId)));
-  const token = WETH_ADDRESSES[Number(args.chainId)];
+  const token = WETH_ADDRESSES[Number(args.chainId) as keyof typeof WETH_ADDRESSES];
```
mrice32, line 31: In general, I think we should avoid casting in these situations.

typescript assumes that token will be a string after this, even though it could definitely be undefined if the key is absent.

Would suggest using [this generic function](https://github.com/across-protocol/relayer-v2/blob/3c655e3c6318a187590fa8fe6774099ec42174d2/src/utils/TypeGuards.ts#L17-L27) to ensure that that chainId is in the struct beforehand and throwing if not.

```ts
const chainId = Number(args.chainId);
if (!isKeyOf(chainId, WETH_ADDRESSES)) throw new Error("chainId does not have a defined WETH address");
// Now we have runtime assurance that this is a real key in the struct
// and typescript "knows" that that validation has been done and will
// apply that assumption to the type.
const token = WETH_ADDRESSES[chainId];

mrice32 (approved):
LGTM! This is awesome!

src/interfaces/index.ts
james-a-morris: This has been merged in a separate PR.

scripts/unwrapWeth.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/565

src/clients/ProfitClient.ts
james-a-morris: This has been resolved in a separate PR

PR #564, opened 2023-04-06
fix: fix bug in slow relay construction

mrice32:
This was causing the executor to fail silently.


pxrl (approved):

src/dataworker/Dataworker.ts
```diff
@@ -1051,7 +1051,7 @@ export class Dataworker {
             leaf.realizedLpFeePct,
             leaf.relayerFeePct,
             leaf.depositId,
-            slowRelayTree.getHexRoot(),
+            rootBundleId,
```
james-a-morris, line 1054: Good find

james-a-morris (approved):

PR #565, opened 2023-04-06
fix: resolve type casting comment

james-a-morris:
This PR ensures that there are no type casting, among other fixes per @mrice32 comments in https://github.com/across-protocol/relayer-v2/pull/559


mrice32 (approved):
Thanks for coming back to this!!

PR #566, opened 2023-04-06
fix: change chainId string to number

james-a-morris:


.eslintrc.js
```diff
@@ -2,27 +2,49 @@ module.exports = {
...
     browser: false,
     es2021: true,
+    mocha: true,
     node: true,
   },
+  plugins: ["node", "prettier", "@typescript-eslint", "mocha", "chai-expect"],
   extends: [
+    "plugin:prettier/recommended",
     "eslint:recommended",
-    "plugin:@typescript-eslint/eslint-recommended",
```
mrice32, line 9: Just wanted to reference [this comment](https://github.com/across-protocol/relayer-v2/pull/552#discussion_r1160028411). Not a blocker, 100% up to you.

mrice32 (approved):
LGTM!

.eslintrc.js
james-a-morris: I think we can revert in this case. Ultimately it's not broken - just slightly different from the SDK/FE.

PR #567, opened 2023-04-06
improve(SpokePoolClient): Better early deposit logging

pxrl:
The existing logging tends to include a lot of information that we don't need, so this will
be more concise and easy to use.


james-a-morris (approved):

mrice32 (approved):
LGTM!

PR #568, opened 2023-04-06
fix: :bug: improve logging support for failed multicall txns

james-a-morris:
This change adds a logging and exit condition to the multicaller txn execution. In the event that the transactions fail, the new logic will collate failed txn states and produce an error log. Additionall this fn will propagate the error and lead to an eventual (caught) stack unwind to prevent any further logic

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -112,25 +112,19 @@ export class MultiCallerClient {
     // We need to iterate over the results to determine if any of the transactions failed.
     // If any of the transactions failed, we need to log the results and throw an error. However, we want to
     // only log the results once, so we need to collate the results into a single object.
-    const { failed } = Object.entries(txnHashes).reduce(
-      (accumulator, [chainId, { result, isError }]) => {
-        const reference = isError ? "failed" : "succeeded";
-        accumulator[reference].chains.push(chainId);
-        accumulator[reference].count += result.length;
-        return accumulator;
-      },
-      { succeeded: { chains: [], count: 0 }, failed: { chains: [], count: 0 } }
-    );
+    const failedChains = Object.entries(txnHashes)
+      .filter(([, { isError }]) => isError)
+      .map(([chainId]) => chainId)
+      .filter((chainId, idx, arr) => arr.indexOf(chainId) === idx);
 
-    // If any of the transactions failed, log the results and throw an error.
-    if (failed.count > 0) {
+    if (failedChains.length > 0) {
       // Log the results.
       this.logger.error({
         at: "MultiCallerClient#executeTxnQueues",
-        message: `Failed to execute ${failed.count} transaction(s) on chain(s) ${failed.chains.join(", ")}`,
+        message: `Failed to execute ${failedChains.length} transaction(s) on chain(s) ${failedChains.join(", ")}`,
         error: txnHashes,
       });
-      throw new Error(`Failed to execute ${failed.count} transaction(s) on chain(s) ${failed.chains.join(", ")}`);
+      throw new Error(`Failed to execute ${failedChains.length} transaction(s) on chain(s) ${failedChains.join(", ")}`);
     }
     // Recombine the results into a single object that match the legacy implementation.
     return Object.fromEntries(Object.entries(txnHashes).map(([chainId, { result }]) => [chainId, result]));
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -100,16 +98,42 @@ export class MultiCallerClient {
...
+          return [chainId, { result: result.reason, isError: true }];
+        }
       })
     );
 
-    return txnHashes;
+    // We need to iterate over the results to determine if any of the transactions failed.
+    // If any of the transactions failed, we need to log the results and throw an error. However, we want to
+    // only log the results once, so we need to collate the results into a single object.
+    const { failed } = Object.entries(txnHashes).reduce(
```
mrice32, line 115: I think this works as is! So feel free to take or leave this suggestion to simplify.
```ts
// Could include the error messages in the return object so we filter the txnHashes that we log in the error, but not 100% necessary.
const failedChains = Object.entries(txnHashes)
  .filter(([, { isError }]) => isError)
  .map(([chainId]) => chainId);

  if (failed.length > 0) {
    // Log the results.
    this.logger.error({
      at: "MultiCallerClient#executeTxnQueues",
      message: `Failed to execute ${failedChains.length} transaction(s) on chain(s) ${failedChains.join(", ")}`,
      error: txnHashes,
    });
    throw new Error(`Failed to execute ${failedChains.length} transaction(s) on chain(s) ${failedChains.join(", ")}`);
  }

File: src/clients/MultiCallerClient.ts [hunk 2]
```diff
@@ -100,16 +98,42 @@ export class MultiCallerClient {
...
+
+    // If any of the transactions failed, log the results and throw an error.
+    if (failed.count > 0) {
+      // Log the results.
+      this.logger.error({
+        at: "MultiCallerClient#executeTxnQueues",
+        message: `Failed to execute ${failed.count} transaction(s) on chain(s) ${failed.chains.join(", ")}`,
+        error: txnHashes,
+      });
+      throw new Error(`Failed to execute ${failed.count} transaction(s) on chain(s) ${failed.chains.join(", ")}`);
```
mrice32, line 133: Are you worried about this blocking later operations in the bot by throwing an error here? For instance, this queue execution is the last operation in the executor, but when used elsewhere it may not be (I'm not sure).

mrice32 (commented):
Looks awesome! Thank you for putting this together!! Only a few minor comments.

src/clients/MultiCallerClient.ts
james-a-morris: I'm a big fan of this suggestion. I've had to add one additional item to prevent any duplicates, but added here ( https://github.com/across-protocol/relayer-v2/pull/568/commits/9c674441ddad317aace1fd90aa0961886ad1b556 )

```
    const failedChains = Object.entries(txnHashes)
      .filter(([, { isError }]) => isError)
      .map(([chainId]) => chainId)
      .filter((chainId, idx, arr) => arr.indexOf(chainId) === idx);
```

src/clients/MultiCallerClient.ts
james-a-morris: I suppose, in this case, we specifically want to have a stack unwind in the executor - but I'm personally comfortable with the idea that the calling code should have to explicitly try/catch to handle error conditions.

src/clients/MultiCallerClient.ts
```diff
@@ -100,16 +98,36 @@ export class MultiCallerClient {
...
     );
 
-    return txnHashes;
+    // We need to iterate over the results to determine if any of the transactions failed.
+    // If any of the transactions failed, we need to log the results and throw an error. However, we want to
+    // only log the results once, so we need to collate the results into a single object.
+    const failedChains = Object.entries(txnHashes)
+      .filter(([, { isError }]) => isError)
+      .map(([chainId]) => chainId)
+      .filter((chainId, idx, arr) => arr.indexOf(chainId) === idx);
```
mrice32, line 118: How are duplicates possible? Isn't chainId the key of the txnHashes struct, which, by definition, must be unique?

src/clients/MultiCallerClient.ts
mrice32: That seems reasonable to me!

mrice32 (approved):
Looks great! Just one minor comment about duplicates!

src/clients/MultiCallerClient.ts
james-a-morris: Fair point - I've removed it here: https://github.com/across-protocol/relayer-v2/pull/568/commits/1586d79a5509e9829f34ff0593435536a39b3eac

PR #569, opened 2023-04-06
feat(refactor): ✨ include eslint config for curly 

james-a-morris:
This change enforces the eslint formatting directive to enforce the "curly" rule. The "curly" rule requires that control statements have set of closing brackets. This helps to improve readability.


mrice32 (approved):
LGTM! I think the reasoning of it not being obvious when you have the statement on the next line is a fair point as to why single-line ifs should be avoided.

james-a-morris:
@pxrl @mrice32 had to include a rather intense refactor after moving code. I've re-requested so that nothing gets lost.

mrice32 (approved):
LGTM!

PR #573, opened 2023-04-07
improve: updates linter settings

mrice32:
This makes a few changes:
1. Adds eslint to `yarn lint`.
2. Fixes a few existing eslint errors around process.exit.
3. Adds unused variables as an error rather than a warning.
4. Removes all existing unused variables (mostly in tests).

### DIFF
--- a/scripts/unwrapWeth.ts
+++ b/scripts/unwrapWeth.ts
@@ -49,7 +49,9 @@ export async function run(): Promise<void> {
     }
     console.log(`Wrap ${amountFromWei} ETH`);
     // Check the user is ok with the info provided. else abort.
-    if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) process.exit(0);
+    if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) {
+      return;
+    }
     console.log("sending...");
     const tx = await weth.deposit({ value: args.amount });
     const receipt = await tx.wait();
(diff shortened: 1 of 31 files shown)


scripts/unwrapWeth.ts
```diff
@@ -85,10 +85,12 @@ export async function run(): Promise<void> {
 if (require.main === module) {
   run()
     .then(async () => {
+      // eslint-disable-next-line no-process-exit
```
james-a-morris, line 88: Nit: would it make sense to create a util function for exiting a program and place the `// eslint-disable...` there? In that case, we only need the disable once

james-a-morris (approved):
Just a minor nit - this all tracks well

scripts/unwrapWeth.ts
mrice32: For small scripts like this, I'd vote no mainly to avoid indirection and extra dependencies. Not strongly opinionated, but IMO it's nice for scripts to be as self-contained as possible.

src/utils/RedisUtils.ts
```diff
@@ -5,7 +5,7 @@ import { createClient } from "redis4";
 import winston from "winston";
 import { Deposit, Fill } from "../interfaces";
 import dotenv from "dotenv";
-import { BigNumberish } from "@across-protocol/sdk-v2/dist/utils";
+import { BigNumberish } from "../utils";
```
pxrl, line 8:
```suggestion
import { BigNumberish } from ".";
```

james-a-morris (approved):
This is a good change

PR #576, opened 2023-04-10
fix: fix TTL resolution bug

mrice32:
This moves TTL resolution into the constructor of the cache provider so the default is defined by the time it runs.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -140,10 +140,10 @@ class CacheProvider extends RateLimitedProvider {
       this.network.chainId
     }:eth_getLogs,`;
 
-    const _ttl = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
+    const _ttlVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
+    const _ttl = Number(_ttlVar);
     if (isNaN(_ttl) || _ttl <= 0) {
-      const envVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
-      throw new Error(`PROVIDER_CACHE_TTL (${envVar}) must be numeric and > 0`);
+      throw new Error(`PROVIDER_CACHE_TTL (${_ttlVar}) must be numeric and > 0`);
     }
     this.baseTTL = _ttl;
   }
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -140,11 +140,12 @@ class CacheProvider extends RateLimitedProvider {
       this.network.chainId
     }:eth_getLogs,`;
 
-    this.baseTTL = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
-    if (isNaN(this.baseTTL) || this.baseTTL <= 0) {
+    const _ttl = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
+    if (isNaN(_ttl) || _ttl <= 0) {
       const envVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
       throw new Error(`PROVIDER_CACHE_TTL (${envVar}) must be numeric and > 0`);
     }
+    this.baseTTL = _ttl;
   }
   override async send(method: string, params: Array<any>): Promise<any> {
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -144,6 +139,12 @@ class CacheProvider extends RateLimitedProvider {
     this.getLogsCachePrefix = `${providerCacheNamespace},${new URL(this.connection.url).hostname},${
       this.network.chainId
     }:eth_getLogs,`;
+
+    this.baseTTL = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
+    if (isNaN(this.baseTTL) || this.baseTTL <= 0) {
```
james-a-morris, line 144: Nit: would it make sense to use a regex check to verify if the baseTTL is non-numeric?

src/utils/ProviderUtils.ts
```diff
@@ -144,6 +139,12 @@ class CacheProvider extends RateLimitedProvider {
     this.getLogsCachePrefix = `${providerCacheNamespace},${new URL(this.connection.url).hostname},${
       this.network.chainId
     }:eth_getLogs,`;
+
+    this.baseTTL = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
```
james-a-morris, line 143: Should we assign this before we verify the validity of the ttl?

mrice32: Sure!

src/utils/ProviderUtils.ts
mrice32: I would argue that we should rely on existing numerical libraries to do this validation. Home-rolled regexes are notorious sources of bugs in my experience, so I prefer not to rely on them for anything mission critical if there are alternatives.

src/utils/ProviderUtils.ts

james-a-morris (approved):
Thanks for addressing my comments! LGTM 😎

src/utils/ProviderUtils.ts
```diff
@@ -144,6 +139,13 @@ class CacheProvider extends RateLimitedProvider {
     this.getLogsCachePrefix = `${providerCacheNamespace},${new URL(this.connection.url).hostname},${
       this.network.chainId
     }:eth_getLogs,`;
+
+    const _ttl = Number(process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL);
+    if (isNaN(_ttl) || _ttl <= 0) {
+      const envVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
+      throw new Error(`PROVIDER_CACHE_TTL (${envVar}) must be numeric and > 0`);
+    }
```
pxrl, line 147:
```suggestion
    const ttlVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
    const _ttl = Number(ttlVar);
    if (isNaN(_ttl) || _ttl <= 0) {
      throw new Error(`PROVIDER_CACHE_TTL (${_ttl}) must be numeric and > 0`);
    }
```

pxrl: Since it's in the constructor now, perhaps we can just resolve the env var once?

mrice32: Shouldn't it be ttlVar in the error?

pxrl: Yes, my bad!

pxrl:
```suggestion
    const _ttlVar = process.env.PROVIDER_CACHE_TTL ?? PROVIDER_CACHE_TTL;
    const _ttl = Number(_ttlVar);
    if (isNaN(_ttl) || _ttl <= 0) {
      throw new Error(`PROVIDER_CACHE_TTL (${ttlvar}) must be numeric and > 0`);
    }
```

nicholaspai: nit: why not `Number(...)` around the first line instead of defining `_ttl` and `_ttlVar`

nicholaspai (approved):

src/utils/ProviderUtils.ts
pxrl: > nit: why not `Number(...)` around the first line instead of defining `_ttl` and `_ttlVar`

That's how it was originally, and how this PR was originally proposed, but it means that in the error case then the input variable needs to be resolved again.

It's OK either way, I just thought it's nicer to only resolve it once.

PR #577, opened 2023-04-10
refactor(test): ♻️ remove most eslint errors

james-a-morris:
This PR aims to remove most ESLint errors/warnings from the master branch. Multiple errors were left due to invalid constructor instantiation. More dialog will be needed to determine how to best remove the leftover errors.

### DIFF
--- a/test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
+++ b/test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
@@ -3,6 +3,7 @@ import { toBN } from "./utils";
 import { BaseAdapter } from "../src/clients/bridges";
 import { SpokePoolClient } from "../src/clients";
 import { OutstandingTransfers } from "../src/interfaces";
+import { ZERO_ADDRESS } from "@uma/common";
 
 class TestAdapter extends BaseAdapter {
   constructor() {
@@ -18,14 +19,30 @@ class TestAdapter extends BaseAdapter {
 
   public setDepositEvents(amounts: number[]) {
     const deposits = amounts.map((amount) => {
-      return { amount: toBN(amount) };
+      return {
+        // Added additional fields to satisfy the DepositEventType
+        amount: toBN(amount),
+        to: ZERO_ADDRESS,
+        transactionIndex: 0,
+        blockNumber: 0,
+        logIndex: 0,
+        transactionHash: "",
+      };
     });
     this.l1DepositInitiatedEvents = { "0xmonitored": { token: deposits } };
   }
 
   public setFinalizationEvents(amounts: number[]) {
     const deposits = amounts.map((amount) => {
-      return { amount: toBN(amount) };
+      return {
+        // Added additional fields to satisfy the DepositEventType
+        amount: toBN(amount),
+        to: ZERO_ADDRESS,
+        transactionIndex: 0,
+        blockNumber: 0,
+        logIndex: 0,
+        transactionHash: "",
+      };
     });
     this.l2DepositFinalizedEvents = { "0xmonitored": { token: deposits } };
   }
--- a/test/Dataworker.buildRoots.ts
+++ b/test/Dataworker.buildRoots.ts
@@ -19,6 +19,7 @@ import {
   mockTreeRoot,
   buildPoolRebalanceLeaves,
   modifyRelayHelper,
+  RelayData,
 } from "./constants";
 import { MAX_REFUNDS_PER_RELAYER_REFUND_LEAF, MAX_L1_TOKENS_PER_POOL_REBALANCE_LEAF } from "./constants";
 import { refundProposalLiveness, CHAIN_ID_TEST_LIST } from "./constants";
@@ -450,7 +451,7 @@ describe("Dataworker: Build merkle roots", async function () {
         "ethers.getSigners doesn't have enough signers"
       ).to.be.true;
       const sortedAllSigners = [...allSigners].sort((x, y) => compareAddresses(x.address, y.address));
-      const fills = [];
+      const fills: Fill[] = [];
       for (let i = 0; i < MAX_REFUNDS_PER_RELAYER_REFUND_LEAF + 1; i++) {
         await setupTokensForWallet(spokePool_1, sortedAllSigners[i], [erc20_1]);
         fills.push(await buildFillForRepaymentChain(spokePool_1, sortedAllSigners[i], deposit1, 0.1, originChainId));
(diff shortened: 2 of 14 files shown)


test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
```diff
@@ -18,14 +19,28 @@ class TestAdapter extends BaseAdapter {
 
   public setDepositEvents(amounts: number[]) {
     const deposits = amounts.map((amount) => {
-      return { amount: toBN(amount) };
+      return {
+        amount: toBN(amount),
+        to: ZERO_ADDRESS, // Added to satisfy the DepositEventType
```
nicholaspai, line 24: nit: only need to use this comment "DepositEventType" once. Maybe just add it above the `return` statement

test/Dataworker.buildRoots.ts
```diff
@@ -779,6 +802,9 @@ describe("Dataworker: Build merkle roots", async function () {
 
       const fill7 = await buildFillForRepaymentChain(spokePool_1, depositor, deposit4, 0.5, originChainId);
       const fill8 = await buildFillForRepaymentChain(spokePool_1, depositor, deposit4, 1, originChainId);
+      if (!fill7 || !fill8) {
+        throw new Error("Fill 7 or 8 is undefined");
```
nicholaspai, line 806: Is there a way we can better annotate the type `fill8` to remove the need to add all these Error throws to fulfill the TS?

nicholaspai (approved):
Left a couple nit comments, hoping we can resolve some of the types without having to add new `if ... throw new Error` lines

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -354,10 +361,11 @@ describe("ProfitClient: Consider relay profit", async function () {
...
     expect(fill.grossRelayerFeePct.eq(deposit.relayerFeePct)).to.be.true;
 
-    deposit.relayerFeePct = toBNWei(".001");
-    expect(deposit.relayerFeePct.lt(deposit.newRelayerFeePct)).to.be.true; // Sanity check
+    const updatedRelayerFeePct = toBNWei("0.0001");
+    deposit.relayerFeePct = updatedRelayerFeePct;
+    expect(deposit.relayerFeePct.lt(updatedRelayerFeePct)).to.be.true; // Sanity check
     fill = profitClient.calculateFillProfitability(deposit, fillAmount, l1Token, minRelayerFeePct);
-    expect(fill.grossRelayerFeePct.eq(deposit.newRelayerFeePct)).to.be.true;
+    expect(fill.grossRelayerFeePct.eq(updatedRelayerFeePct)).to.be.true;
```
pxrl, line 368: Thoughts on just asserting that deposit.newRelayerFeePct is a BigNumber? i.e.

```
expect(deposit.relayerFeePct.lt(deposit.newRelayerFeePct as BN)).to.be.true; // Sanity check
```

It's correct that the Deposit interface specifies that these types are `BigNumber | undefined`, so we should normally handle the undefined case. In this instance, we know that it's a BigNumber because it's hardcoded immediately above. I also did a quick REPL test and it seems impossible to assert that undefined should be a BigNumber (it just remains undefined), so the sanity check should still catch a scenario where deposit.newRelayerFeePct somehow is undefined (see below).

Note that I do also see valid arguments for being strict - predominantly the possibility for future code changes (in relayer-v2 or upstream) to invalidate some of the assumptions that are baked in now. If we choose that path, there are a few other places in this code that would need to be fixed - i.e. the instantiation of deposit object itself, because it only implements the subset of the actual Deposit interface that is necessary for computing profitability, and it relies on a similar type assertion to appease the transpiler. My reasoning for the assertions is that defining the complete set of fields adds boilerplate and (IMO) distracts from the actual test that is being performed.

```
> ethers.BigNumber.isBigNumber(undefined as BigNumber);
false

> ethers.BigNumber.from("1").gt(undefined as BigNumber);
across-protocol/relayer-v2/node_modules/@ethersproject/logger/lib/index.js:247
        throw this.makeError(message, code, params);
        ^

Uncaught:
Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.7.0)
    at Logger.makeError (across-protocol/relayer-v2/node_modules/@ethersproject/logger/src.ts/index.ts:269:28)
    at Logger.throwError (across-protocol/relayer-v2/node_modules/@ethersproject/logger/src.ts/index.ts:281:20)
    at Logger.throwArgumentError (across-protocol/relayer-v2/node_modules/@ethersproject/logger/src.ts/index.ts:285:21)
    at Function.BigNumber.from (across-protocol/relayer-v2/node_modules/@ethersproject/bignumber/src.ts/bignumber.ts:289:23)
    at toBN (across-protocol/relayer-v2/node_modules/@ethersproject/bignumber/src.ts/bignumber.ts:345:27)
    at BigNumber.gt (across-protocol/relayer-v2/node_modules/@ethersproject/bignumber/src.ts/bignumber.ts:167:30)
    at across-protocol/relayer-v2/<repl>.ts:1:37
    at Script.runInThisContext (node:vm:129:12)
    at runInContext (across-protocol/relayer-v2/node_modules/ts-node/src/repl.ts:673:19)
    at Object.execCommand (across-protocol/relayer-v2/node_modules/ts-node/src/repl.ts:639:28) {
  reason: 'invalid BigNumber value',
  code: 'INVALID_ARGUMENT',
  argument: 'value',
  value: undefined
}
```

test/mocks/MockTokenClient.ts
```diff
@@ -8,40 +8,28 @@ export class MockTokenClient extends TokenClient {
...
-  getBalance(chainId: number, token: string) {
-    if (!this.tokenData[chainId]) {
-      return toBN(0);
-    }
-    if (!this.tokenData[chainId][token]) {
-      return toBN(0);
-    }
-    return this.tokenData[chainId][token].balance;
+  getBalance(chainId: number, token: string): BigNumber {
+    return this.tokenData[chainId]?.[token]?.balance || toBN(0);
```
pxrl, line 25: Does the || have any impact here? For numeric values it'd activate if balance was 0, but this seems to be a BigNumber object.

```suggestion
    return this.tokenData[chainId]?.[token]?.balance ?? toBN(0);
```

test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
pxrl: I left a longer comment below on ProfitClient.ConsiderProfitability.ts, and it might have some relevance here. In general, it'd be good to confirm whether we want to require that an object is populated with all required member fields, or whether it's OK to assert the type _if_ it's easy to check that we have populated the fields necessary for completing the test.

As an example, the alternative approach here would be:

```
return { amount: toBN(amount) } as DepositEvent;
```

Possible advantages:
 - Less boilerplate required to test => more focus on what's actually being tested.

Possible disadvantages:
 - We take greater assumptions, meaning greater chance for things to break in future (i.e. due to changes here or upstream).

Note that we are (in some respect) also fudging via an implicit type assertion with the current proposed change, since most of the fields are being set to non-sensical values (and it seems borderline impossible to set them to sensible defaults anyway, given the current surrounding implementation).

pxrl (commented):
Left a few comments - nice work :+1:

test/mocks/MockTokenClient.ts
james-a-morris: The `||` here will convert even if the `balance` is falsy (potentially 0). I think there should be no functional issue with this.

test/Dataworker.buildRoots.ts
james-a-morris: Yes - I've made changes to move the exception to one place to remove the need for these try-catch blocks.

test/mocks/MockTokenClient.ts
pxrl: Agree - no functional issue since || will catch a superset of ??. Was just curious if there was a specific motivation for it.

mrice32 (approved):
LGTM one @pxrl and @nicholaspai's comments are resolved!

nicholaspai:
@james-a-morris want to revisit this one? I think its a great initiative but a lot of changes have gone into `master`

james-a-morris:
Close for now

PR #578, opened 2023-04-11
feat: Let bot runner set `toBlockOverride` to construct clients for deprecated SpokePools

nicholaspai:
This can be used for example to make sure that the `finalizer` continues to finalize events for deprecated spoke pools by setting `TO_BLOCK_OVERRIDE_1=<block-before-spoke-pool-was-deprecated>`.

This is because the function `constructSpokePoolClientsWithLookback` used to create `SpokePoolClient`'s  looks up the enabled SpokePools as of the `toBlock` for chainId=1. Currently the `toBlocks` default to "latest", so this PR allow these to be overridden.

### DIFF
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -18,6 +18,7 @@ export class CommonConfig {
   readonly multiCallChunkSize: { [chainId: number]: number };
   readonly version: string;
   readonly blockRangeEndBlockBuffer: { [chainId: number]: number };
+  readonly toBlockOverride: Record<number, number> = {};
 
   constructor(env: ProcessEnv) {
     const {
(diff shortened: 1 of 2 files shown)


src/common/Config.ts
```diff
@@ -18,6 +18,7 @@ export class CommonConfig {
   readonly multiCallChunkSize: { [chainId: number]: number };
   readonly version: string;
   readonly blockRangeEndBlockBuffer: { [chainId: number]: number };
+  readonly toBlockOverride: Record<number, number> = {};
```
james-a-morris, line 21: Should we set this to `this.toBlockOverride = {}` in the constructor?

nicholaspai: Do you think that is cleaner code? I thought setting it here was pretty much the same?

mrice32 (approved):
LGTM!

src/common/Config.ts
```diff
@@ -46,6 +47,15 @@ export class CommonConfig {
         );
       }
     }
+
+    for (const chainId of Constants.CHAIN_ID_LIST_INDICES) {
+      if (env[`TO_BLOCK_OVERRIDE_${chainId}`] !== undefined) {
```
pxrl, line 52: Do you plan to keep this env var around for the long term, or only whilst we have the Boba hurdle to clear? If it's intended for long-term use then it might be nice to scope this as a SpokePool-specific override, since it has no impact on the other clients.

nicholaspai: @pxrl this isn't just a boba hurdle, its the hurdle of all finalizations that originate from the deprecated spoke pool, so I think we're talking about the same thing.

I imagine this will never be used after this hurdle is cleared. Do you think its better that we rename this to `spokePoolClientToBlockOverride` since its only used in that one area? We can also leave a comment to mark for eventual deprecation?

pxrl: > Do you think its better that we rename this to `spokePoolClientToBlockOverride` since its only used in that one area? We can also leave a comment to mark for eventual deprecation?

Yeah I think it's nice to scope it to the Spoke Pool. Also OK with leaving a comment (open question) about whether it could be deprecated again in future :+1:

PR #579, opened 2023-04-11
improve(SpokePoolClient): Support test event injection

pxrl:
This change implements a mocked SpokePoolClient, with the SpokePoolClient._update() method overridden in order to suppress on-chain queries and instead inject ethers Event objects directly.

It _should_ have a noticeable speed difference over running end-to-end test via Hardhat, and it permits test of Events that are not yet emitted by the contracts-v2 package. With this feature we can develop further without a hard dependency on bumping the contracts ABI. Further, it should be trivially reusable for other SpokePoolClient events in test.


test/SpokePoolClient.RefundRequests.ts
```diff
@@ -11,97 +12,85 @@ import {
...
 
 describe("SpokePoolClient: Refund Requests", async function () {
   beforeEach(async function () {
     [relayer] = await ethers.getSigners();
-    ({
-      spokePool,
-      weth: refundToken,
-      deploymentBlock,
-    } = await deploySpokePoolWithToken(originChainId, destinationChainId));
+    ({ spokePool, deploymentBlock } = await deploySpokePoolWithToken(originChainId, destinationChainId));
```
pxrl, line 25: We're still dependent on deploying a contract instance to bootstrap the SpokePoolClient. Some future hacking might be able to do away with that; though I didn't want to tackle it now.

test/mocks/MockSpokePoolClient.ts
```diff
@@ -0,0 +1,138 @@
...
+      currentTime,
+      events,
+      searchEndBlock: this.eventSearchConfig.toBlock || latestBlockNumber,
+    };
+  }
+
+  generateRefundRequest(inputs: EthersEventTemplate): Event {
+    // @todo: Source these from contracts-v2, when available.
+    const event = "RefundRequested";
+    const eventSignature = "RefundRequested(uint256,uint256,uint256,uint256,uint256,int65,uint32,uint256,uint256)";
```
pxrl, line 79:
```suggestion
    const eventSignature = "RefundRequested(address,address,uint256,uint256,uint256,int64,uint32,uint256,uint256)";
```

mrice32 (approved):
LGTM!

PR #580, opened 2023-04-12
fix(Monitor): Suppress "unknown relayer" alerts on slow fills

pxrl:
Slow fills can trigger "unknown relayer" alerts now that multicall3 is used to fill slow
relays. The address of the relayer is perceived to be the multicall3 contract address,
and the address of tx.origin is not available directly in the FillWithBlock event (would
need to look it up via transaction hash instead).

We can't include multicall3 in the list of known relayers, so given that slow relays are
trusted to the extent that they have cleared liveness and have been sent out of the
HubPool, simply suppress alerts on slow fills from unknown relayers.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -167,7 +167,7 @@ export class Monitor {
       );
       for (const fill of fills) {
         // Skip notifications for known relay caller addresses.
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):
LGTM

mrice32 (approved):

PR #581, opened 2023-04-12
feat(UBA): :sparkles: add stub for new UBAFeeCalculator

james-a-morris:
This change adds the UBAFeeCalculator client class. This class will eventually work with @pxrl's logic to extract and sequence an array of UBA flows as well as logic that is used to calculate the latest running validated balance.

### DIFF
--- a/src/utils/UBAFeeCalculator.ts
+++ b/src/utils/UBAFeeCalculator.ts
@@ -101,7 +101,7 @@ export default class UBAFeeCalculator {
   }
 
   /**
-   * @description Get the most recent request flow
+   * @description Get the most recent request flow array from the most recent validated propsoal block
    * @returns Promise<UbaFlow[]>
    * @todo Implement this function
    * @private
--- a/src/utils/UBAFeeCalculator.ts
+++ b/src/utils/UBAFeeCalculator.ts
@@ -1,0 +1,107 @@
+import { BigNumber, ethers } from "ethers";
+import { UbaFlow, UbaRunningRequest, isUbaInflow, isUbaOutflow } from "../interfaces";
+import { toBN } from "./FormattingUtils";
+
+/**
+ * @file UBAFeeCalculator.ts
+ * @description UBA Fee Calculator
+ * @author Across Bots Team
+ */
+export default class UBAFeeCalculator {
+  private readonly provider: ethers.providers.Provider;
+  private blockNumber: number;
+  private lastValidatedRunningBalance?: BigNumber;
+  private recentRequestFlow: UbaFlow[];
+  private runningBalance?: BigNumber;
+
+  constructor(provider: ethers.providers.Provider, initialBlockNumber: number) {
+    this.provider = provider;
+    this.blockNumber = initialBlockNumber;
+  }
+
+  /**
+   * @description Recalculates running balance and from a given block number
+   * @param blockNumber An optional block number to start from - otherwise will start from the lastest blockNumber
+   */
+  public async recalculateRunningBalance(blockNumber?: number): Promise<void> {
+    // Initially set the blockNumber to a new block with either the given input or the last blockNumber available
+    this.blockNumber = blockNumber ?? (await this.provider.getBlockNumber());
+    // Clear the recent request flow
+    this.recentRequestFlow = [];
+    // Resolve the most recent flows
+    this.recentRequestFlow = await this.getRecentRequestFlow();
+    // Recalculate the last validated running balance
+    this.lastValidatedRunningBalance = await this.calculateRunningBalance();
+    // Resolve the running balance
+    this.runningBalance = this.getRunningBalance();
+  }
+
+  /**
+   * @description Get the recent request flow
+   * @param action The action to get the fee for
+   * @returns The relevant fee
+   */
+  public async getUBAFee(action: UbaRunningRequest): Promise<BigNumber> {
+    // Destructure the action
+    const { amount, type } = action;
+    // First verify that both the last validated running balance and the runningBalance is
+    // set to a non-undefined value
+    if (this.lastValidatedRunningBalance === undefined || this.runningBalance === undefined) {
+      await this.recalculateRunningBalance();
+    }
+    // Set the amount to add to the running balance
+    const amountToModify = amount.mul(type === "deposit" ? 1 : -1);
+    // Resolve the new modified balance
+    const modifiedBalance = this.runningBalance.add(amountToModify);
+    // Return the running balance
+    return modifiedBalance;
+  }
+
+  /**
+   * @description Get the running balance
+   * @returns Promise<BigNumber>
+   * @private
+   * @method getRunningBalance
+   * @memberof UBAFeeCalculator
+   */
+  private getRunningBalance(): BigNumber {
+    // Reduce over the recent request flow and add the amount to
+    // the last validated running balance. If there is no last validated running balance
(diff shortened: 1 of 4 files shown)


src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,107 @@
+import { BigNumber, ethers } from "ethers";
```
mrice32, line 1: I really liked the way you built this stub. I think this is incredibly useful for visualizing  how this will work/be used.

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,107 @@
...
+  private runningBalance?: BigNumber;
+
+  constructor(provider: ethers.providers.Provider, initialBlockNumber: number) {
+    this.provider = provider;
+    this.blockNumber = initialBlockNumber;
+  }
+
+  /**
+   * @description Recalculates running balance and from a given block number
+   * @param blockNumber An optional block number to start from - otherwise will start from the lastest blockNumber
```
mrice32, line 24: When you say "start from" does that mean working _forward_ in time? Or backwards from that block?

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,107 @@
...
+    // Resolve the running balance
+    this.runningBalance = this.getRunningBalance();
+  }
+
+  /**
+   * @description Get the recent request flow
+   * @param action The action to get the fee for
+   * @returns The relevant fee
+   */
+  public async getUBAFee(action: UbaRunningRequest): Promise<BigNumber> {
```
mrice32, line 44: I'm having a bit of trouble seeing how all these methods fit together. Do you think you could add a comment at the top to kind of explain how the typical flow or happy path might work?

mrice32 (commented):
Overall, I think putting together stubs like this to start is awesome! Just one question to get a better understanding of how these functions fit together.

src/utils/UBAFeeCalculator.ts
james-a-morris: It moves backwards in time - i've updated the docs here: https://github.com/across-protocol/relayer-v2/pull/581/commits/a415ba4cfca8f9b52286c44a4dfd650df6564f3c

src/utils/UBAFeeCalculator.ts
james-a-morris: I've added comments here: https://github.com/across-protocol/relayer-v2/pull/581/commits/46d80c816f8830c7d030909558060b4a2f549a92

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @todo Implement this function
+   * @private
+   * @async
+   * @method calculateRunningBalance
+   * @memberof UBAFeeCalculator
+   * @returns {BigNumber}
+   */
+  private async calculateRunningBalance(): Promise<BigNumber> {
+    this.blockNumber;
+    throw new Error("Not implemented");
```
mrice32, line 122: Just a random thought -- do you think this method needs to return the block number that that last running balance was for as well? So our recent flows don't take into account requests from before that bundle?

mrice32 (approved):
Looks great! Just a few minor comments.

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @description Calculate the last validated running balance
+   * @returns Promise<BigNumber>
+   * @todo Implement this function
+   * @private
+   * @async
+   * @method calculateRunningBalance
+   * @memberof UBAFeeCalculator
+   * @returns {BigNumber}
+   */
+  private async calculateRunningBalance(): Promise<BigNumber> {
```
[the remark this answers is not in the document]
james-a-morris: Ah definitely - in this case, my goal is to completely replace this for @pxrl's `UBASequencer` class that he's working on. The ultimate goal is to remove this fn.

I've updated this in the meantime to hopefully be more readable: https://github.com/across-protocol/relayer-v2/pull/581/commits/13094a0a821822259b806101808bb06ab6c8e395

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,124 @@
...
+ * @author Across Bots Team
+ */
+export default class UBAFeeCalculator {
+  private readonly provider: ethers.providers.Provider;
+  private blockNumber: number;
+  private lastValidatedRunningBalance?: BigNumber;
+  private recentRequestFlow: UbaFlow[];
+  private runningBalance?: BigNumber;
+
+  constructor(provider: ethers.providers.Provider, initialBlockNumber: number) {
```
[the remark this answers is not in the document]
james-a-morris: Yep! We'll get the chain id from the spoke pool client provided in the constructor. I've preemtively added a token symbol [here](https://github.com/across-protocol/relayer-v2/pull/581/commits/a8d320b5e2d90cafa96adf0a3cc337419806c7d9)

src/utils/UBAFeeCalculator.ts
james-a-morris: This is a good callout - the ultimate goal of this function is to be replaced by the extraction/sequencer logic that @pxrl is writing. I think he's adding some layer of caching to optimize RPC calls

pxrl: By using the SpokePoolClient we should inherit the pre-existing "sliding window" logic that is implemented there :+1:

src/utils/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * @description Get the most recent request flow
+   * @returns Promise<UbaFlow[]>
+   * @todo Implement this function
+   * @private
+   * @async
+   * @method getRecentRequestFlow
+   * @memberof UBAFeeCalculator
+   */
+  private async getRecentRequestFlow(): Promise<UbaFlow[]> {
```
mrice32, line 112: One last question: the description says that we're going to get the most recent singular flow, but this returns an array of flows. So I guess I'm just left with the question of:
1. What is a single UbaFlow object in this context?
2. Should this be returning a single flow to match the description? Or are we returning a group of flows based on some filter (like after a certain block, or similar?

mrice32 (approved):
LGTM! One last comment, which is more of just a question. Feel free to merge when ready!

src/utils/UBAFeeCalculator.ts
james-a-morris: This function is meant to be replaced by the code that @pxrl has written - I left it as more of a placeholder than a stub. I've updated the docs to be a bit more descriptive. This was a good callout

pxrl: > One last question: the description says that we're going to get the most recent singular flow, but this returns an array of flows. So I guess I'm just left with the question of:
> 
> 1. What is a single UbaFlow object in this context?
> 2. Should this be returning a single flow to match the description? Or are we returning a group of flows based on some filter (like after a certain block, or similar?

A single flow is the union of a FundsDeposited, RelayFilled and RefundRequested event. Type predicates are available to distinguish each subtype and act accordingly, though ideally we won't need to do that very much.

So the request flow is the sequence array of events that have been made against the SpokePool since the request/update.

pxrl (approved):
LGTM - this is a good structure to work up against.

PR #582, opened 2023-04-12
feat!: Support new SpokePool interfaces

nicholaspai:
Relayer and Dataworker should be able to use this code to run on new spoke pools


src/clients/SpokePoolClient.ts
```diff
@@ -187,18 +188,24 @@ export class SpokePoolClient {
...
+    const maxSpeedUp = speedupsForDeposit?.reduce((prev, current) =>
       prev.newRelayerFeePct.gt(current.newRelayerFeePct) ? prev : current
     );
 
     // Only if there is a speedup and the new relayer fee is greater than the current relayer fee, save the new fee.
     if (!maxSpeedUp || maxSpeedUp.newRelayerFeePct.lte(deposit.relayerFeePct)) {
       return deposit;
     }
+
+    // Return deposit with updated params from the speedup with the highest updated relayer fee pct.
```
nicholaspai, line 202: @mrice32 @pxrl @james-a-morris this is a design choice that makes the code easier to write but it wouldn't be too hard to change. We take the `updatableRelayExecutionData` from the speed up call with the _highest_ `relayerFeePct` and set the relay's `recipient` and `message` accordingly. I could see us implementing:
- Take the highest `relayerFeePct`
- Take the latest `recipient` and `message`. I just didn't implement this for now to get the tests to pass initially

src/clients/SpokePoolClient.ts
```diff
@@ -101,7 +102,7 @@ export class SpokePoolClient {
       FundsDeposited: this.spokePool.filters.FundsDeposited(),
       RequestedSpeedUpDeposit: this.spokePool.filters.RequestedSpeedUpDeposit(),
       FilledRelay: this.spokePool.filters.FilledRelay(),
-      // RefundRequested: this.spokePool.filters.refundRequested(), @todo update contracts-v2
+      // RefundRequested: this.spokePool.filters.refundRequested(), @todo Only used once we implement UBA fee model
```
pxrl, line 105: Oh - I thought we'd be updating to the new contracts-v2 as part of the migration to the new SpokePools, but before the UBA migration. Is there some other blocker that I missed?

nicholaspai: Good point we can uncomment this but I was thinking that we don't use this event in the new spoke pools so no point in querying it

src/clients/SpokePoolClient.ts
nicholaspai: > there's no scope for mixing fields from multiple speed-ups, right?

This is the crux of the issue. I can see why a depositor might not realize they need to bump their `relayerFeePct` when setting a new `message` and `recipient` (perhaps because the `message` reverted on the `recipient` contract). But then again, relayers have no incentive to use the new message/recipient unless the fee is higher

src/relayer/Relayer.ts
```diff
@@ -277,7 +302,8 @@ export class Relayer {
   }
 
   zeroFillDeposit(deposit: Deposit): void {
-    const repaymentChainId = 1; // Always refund zero fills on L1 to not send dust over the chain unnecessarily.
+    // We can only overwrite repayment chain ID if we can fully fill the deposit.
+    const repaymentChainId = deposit.destinationChainId;
```
nicholaspai, line 306: This is an unfortunate artifact of the new SpokePools.  Slow fills will always be refunded on the destination chain. This might exacerbate high volume situations in which a lot of slow fills are sent. Whereas previously these slow fills could lead the HubPool refunding relayers on mainnet, refilling their balances, versus now refunding them on the destination chain.

src/clients/SpokePoolClient.ts
pxrl: Yeah that's true, but uncommenting this line only means that we have the option of querying for RefundRequested events. We'd have to separately update SpokePoolClient users to actually request the event, so I think it's fine to uncomment this line as part of this PR.

src/clients/SpokePoolClient.ts
pxrl: > This is the crux of the issue. I can see why a depositor might not realize they need to bump their `relayerFeePct` when setting a new `message` and `recipient` (perhaps because the `message` reverted on the `recipient` contract). But then again, relayers have no incentive to use the new message/recipient unless the fee is higher

Fair point; there is a potential UX issue around that. It seems like a can of worms to permit SpeedUp events to be combined according to arbitrary relayer logic, though.

I think it's very clean to follow a ruleset of "the depositor authorises SpeedUps in isolation of each other", and this keeps the relayer lookup simple: find the SpeedUp with the highest relayerFeePct, and use all of its fields. SpeedUp requests are already fairly complex for the user to create so it seems like they'll always need a helper UI - could the "minimum fee" issue be handled in the UI instead?

src/clients/SpokePoolClient.ts
```diff
@@ -397,9 +404,18 @@ export class SpokePoolClient {
...
         this.spokePool,
-        // TODO: When SpokePool V2 is deployed we can begin to filter on fill.depositId as well as fill.depositor,
-        // which should optimize the RPC request.
-        this.spokePool.filters.FundsDeposited(null, null, null, null, null, null, null, null, fill.depositor),
+        this.spokePool.filters.FundsDeposited(
+          null,
+          null,
+          fill.destinationChainId,
+          null,
+          fill.depositId,
```
pxrl, line 412: It seems counter-intuitive, but I think this will result in a net increase in RPC usage.

Currently we issue a very broad query, so we pick up all Deposit events from a single depositor. In the case of aggregators, we are probably caching quite a few additional Deposits, which we _may_ end up searching for later anyway.

Now that we apply such a fine-grained filter, the vast majority of eth_getLogs() lookups will return empty results, but we'll still pay the cost of the RPC request.

With such a fine-grained filter, I think it makes makes sense to increase the range of blocks over which we are searching. The trick is to figure out a value/pattern that applies safely to all supported chains and providers.

nicholaspai: hmm, but we are caching general event queries, rather than the individual deposit events. So for us to have cached this specific search for a depositor seems not as likely. And this way we can make that initial event search very fast. I'm not sure though, this requires some testing in prod

src/clients/SpokePoolClient.ts
nicholaspai: ok done

src/clients/SpokePoolClient.ts
nicholaspai: I agree with keeping it simple at the expenses of slightly (very very slightly) degraded depositor-X

scripts/runRelayerOnGoerli.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "Running relayer on Goerli. Will fill any deposits for origin chain [5/421613] to destination chain [421613/5]"
+
+echo "Killing redis server if it's running to avoid production data corruption"
+redis-cli shutdown
+
+export RELAYER_IGNORE_LIMITS="true"
+export RELAYER_TOKENS='["0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6"]'
+export MAX_RELAYER_DEPOSIT_LOOK_BACK=7200
+export NODE_DISABLE_PROVIDER_CACHING=true
+export IGNORE_PROFITABILITY=true
```
pxrl, line 12: You might be able to try with the following combination in order to avoid reintroducing IGNORE_PROFITABILITY:
```
RELAYER_GAS_MULTIPLIER=.01
RELAYER_MIN_FEE_PCT=.00001
```

src/relayer/Relayer.ts
```diff
@@ -225,10 +244,35 @@ export class Relayer {
     }
 
     try {
+      // TODO: Consider adding some way for Relayer to delete transactions in Queue for fills for same deposit.
```
pxrl, line 247: I'm not sure I understand the scenario that could lead to this - can you clarify?

nicholaspai: Imagine the queue has two partial fills for one deposit, the first one is a 1-wei fill, the second is a full fill. The second full fill sets its repayment chain != destination chain.

The transactions will revert because the first partial fill will execute before the full fill, rendering the full fill into a partial fill and partial fills must take repayment on destination chain.

This is an incredibly rare case in reality as I mention in the comment but I somehow created it on test nets

scripts/runRelayerOnGoerli.sh
nicholaspai: The main issue with following the normal ProfitClient flow is that the client can't figure out how to fetch the ETH token price on goerli. We can opt for this logic although it also has its tradeoffs and isn't totally clean: ab8522fee070e3f8f58d60eb17a60e3c315d40f8

src/clients/ProfitClient.ts
```diff
@@ -352,7 +377,7 @@ export class ProfitClient {
     }
   }
 
-  private async updateGasCosts(): Promise<void> {
+  async updateGasCosts(): Promise<void> {
```
pxrl, line 380: updateGasCost() isn't called externally anymore, so we can probably revert this change.

```suggestion
  private async updateGasCosts(): Promise<void> {
```

src/clients/SpokePoolClient.ts
```diff
@@ -201,18 +202,26 @@ export class SpokePoolClient {
   }
 
   appendMaxSpeedUpSignatureToDeposit(deposit: DepositWithBlock): DepositWithBlock {
-    const maxSpeedUp = this.speedUps[deposit.depositor]?.[deposit.depositId]?.reduce((prev, current) =>
+    const speedupsForDeposit = this.speedUps[deposit.depositor]?.[deposit.depositId];
+
+    // Find the SpeedUp with the highest relayerFeePct
+    const maxSpeedUp = speedupsForDeposit?.reduce((prev, current) =>
```
pxrl, line 208: I think this started out as a functional change, but it seems to have reverted to the pre-existing behaviour (albeit with different formatting). Should we just revert that so it disappears from the diff?

```suggestion
const maxSpeedUp = this.speedUps[deposit.depositor]?.[deposit.depositId]?.reduce((prev, current) =>
```

src/relayer/Relayer.ts
pxrl: > This is an incredibly rare case in reality as I mention in the comment but I somehow created it on test nets

Ah - now it makes sense :) I don't see how it could happen in prod, but I could imagine it happening in hardhat if the block numbers are jumping around (as we have seen them do).

pxrl:
@nicholaspai See https://github.com/across-protocol/relayer-v2/pull/603 for a proposed fix for the current breakage.

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
nicholaspai: Its reproducible by the following test:
- run checkForUnfilledDeposits with insufficient balance, relayer enqueues a `zeroFill`
- send more balance to relayer
- run checkForUnfilledDeposits again, relayer enqueues a full `fill`

yarn.lock
```diff
@@ -22,14 +22,30 @@
...
+    "@openzeppelin/contracts" "^4.8.3"
+    "@openzeppelin/contracts-upgradeable" "^4.8.3"
+    "@uma/common" "^2.29.0"
+    "@uma/contracts-node" "^0.3.18"
+    "@uma/core" "^2.41.0"
+    arb-bridge-eth "^0.7.4"
+    arb-bridge-peripherals "^1.0.5"
+
+"@across-protocol/contracts-v2@^2.1.0-beta.17":
+  version "2.1.0"
```
pxrl, line 41: This seems to resolve to a previously released (but deprecated) 2.1.0: https://www.npmjs.com/package/@across-protocol/contracts-v2/v/2.1.0

Should we bump contracts-v2 to 2.1.1 to avoid this?

pxrl (approved):

PR #585, opened 2023-04-14
fix(SpokePoolClient): Don't reorder on getDeposits()

pxrl:
getDeposits() internally sorts deposit events according to blockNumber. The deposit blockNumber is however the block number corresponding to the quoteTimestamp, not the block number of the deposit transaction. This is a quirk that seems unique to FundsDeposited events.

This method is only used in test, but the re-ordering is incompatible with the new UBA design, where the ordering of SpokePool transactions takes precedence.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -114,7 +114,7 @@ export class SpokePoolClient {
   }
 
   getDeposits(): DepositWithBlock[] {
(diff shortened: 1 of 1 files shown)


pxrl:
nb. I am now proposing to shuffle the DepositWithBlock type, so that blockNumber is the event block number and "quoteBlockNumber" is the block number corresponding to the quoteTimestamp field.

PR #586, opened 2023-04-14
refactor(FundsDeposited): Realign blockNumber purpose

pxrl:
The FundsDeposited event is unique in that its blockNumber currently refers to the block number according to its quoteTimestamp. All other events use blockNumber to specify the timestamp of their event.

This change re-aligns FundsDeposited to be more compatible with the other type definitions, which will make it easier to sort them for the UBA model.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -428,7 +428,6 @@ export class SpokePoolClient {
         destinationToken: this.getDestinationTokenForDeposit(processedEvent),
         quoteBlockNumber: dataForQuoteTime.quoteBlock,
         blockNumber: event.blockNumber,
-      } as DepositWithBlock;
       this.logger.debug({
         at: "SpokePoolClient#queryHistoricalDepositForFill",
         message: "Queried RPC for deposit outside SpokePoolClient's search range",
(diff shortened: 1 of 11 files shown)


nicholaspai (approved):
This is a good change, have you tested in prod?

pxrl:
> This is a good change, have you tested in prod?

I'm about to start that now. Note that it's dependent on an SDK bump here: https://github.com/across-protocol/sdk-v2/pull/171, and there would be some follow-up changes necessary here before we can submit:
 - Bump SDK version to 0.4.7.
 - Adjust the instances of deposit["quoteBlockNumber"] => deposit.quoteBlockNumber.

james-a-morris (approved):
This looks like a straightforward refactor! LGTM 🙂

src/clients/SpokePoolClient.ts
```diff
@@ -426,9 +426,9 @@ export class SpokePoolClient {
         ...processedEvent,
         realizedLpFeePct: dataForQuoteTime.realizedLpFeePct,
         destinationToken: this.getDestinationTokenForDeposit(processedEvent),
-        blockNumber: dataForQuoteTime.quoteBlock,
-        originBlockNumber: event.blockNumber,
-      };
+        quoteBlockNumber: dataForQuoteTime.quoteBlock,
+        blockNumber: event.blockNumber,
+      } as DepositWithBlock;
```
pxrl, line 431:
```suggestion
```

pxrl:
Have repeated testing with the same script, against the latest commit (a9a2926). 61 bundles check out OK again.
[validateRootBundle-results.txt](https://github.com/across-protocol/relayer-v2/files/11251978/validateRootBundle-results.txt)

pxrl:
Have additionally tested with the relayer a few times. Attempted fills correctly identify the blockNumber and quoteBlockNumber, and I haven't seen any signs of misbehaviour.

PR #587, opened 2023-04-14
improve(MockSpokePoolClient): Fixes & simplifications

pxrl:
- The ethers Event type is a weird hybrid of an array _and_ an Object with k/v pairs. Internally in relayer-v2 EventUtils, we ignore the array members and focus on the named k/v pairs. Therefore, switch the args type from an array to an Object. This enables the named args to be properly extracted and transformed into a proper "...WithBlock" object.

 - Consolidate MockSpokePoolClient event generation, such that the event name is passed in as part of the EthersEventTemplate. This makes it easier to use and reduces repetitive code fragments.

 - Minor optimisation to the way events are fed into the base SpokePoolClient during mocked _update().

 - Increase randomness in transactionHash generation.

### DIFF
--- a/test/SpokePoolClient.RefundRequests.ts
+++ b/test/SpokePoolClient.RefundRequests.ts
@@ -19,6 +19,8 @@ let relayer: SignerWithAddress;
 let deploymentBlock: number;
 let spokePoolClient: MockSpokePoolClient;
 
+const event = "RefundRequested";
+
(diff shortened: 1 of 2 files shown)


james-a-morris (commented):
Left a comment - looks great!

nicholaspai (approved):

pxrl:
> Left a comment - looks great!

Thanks, nice suggestion.

PR #589, opened 2023-04-17
feat: add piecewise function logic

james-a-morris:
This PR adds logic to calculating the UBA fees given a provided on-chain configuration.

Partially closes ACX-1021


src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -60,24 +64,44 @@ export default class UBAFeeCalculator {
...
+    // Resolve the balancing fee tuples that are relevant to this operation
+    const originBalancingFeeTuples = this.config.getBalancingFeeTuples(originChain);
+    const destinationBalancingFeeTuples = this.config.getBalancingFeeTuples(destinationChain);
+
+    // If the action is a deposit, then we need to add the origin and destination balancing fee
+    // to the total UBA fee. We can use the getDepositBalancingFee function to do this
+    // Find both of these fees from the origin and destination chains
+    if (type === "deposit") {
+      totalUBAFee = totalUBAFee.add(getDepositBalancingFee(originBalancingFeeTuples, this.runningBalance, amount));
+      totalUBAFee = totalUBAFee.add(getDepositBalancingFee(destinationBalancingFeeTuples, this.runningBalance, amount));
```
mrice32, line 94: As discussed offline, I think this should be refund balancing fee on the destination.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -60,24 +64,44 @@ export default class UBAFeeCalculator {
...
+    }
+    // If the action is a refund, then we need to add the origin and destination balancing fee
+    // to the total UBA fee. We can use the getRefundBalancingFee function to do this
+    // Find both of these fees from the origin and destination chains
+    else {
+      totalUBAFee = totalUBAFee.add(getRefundBalancingFee(originBalancingFeeTuples, this.runningBalance, amount));
```
mrice32, line 100: As discussed offline, I think this should be deposit balancing fee on the origin side when doing a refund.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -60,24 +64,44 @@ export default class UBAFeeCalculator {
...
+      totalUBAFee = totalUBAFee.add(getRefundBalancingFee(destinationBalancingFeeTuples, this.runningBalance, amount));
+    }
+
+    return totalUBAFee;
```
mrice32, line 104: As we discussed earlier, I think this should be broken down into LP fee and relayer fee.

Relayer fee = gas fee + destination refund balancing fee.

LP fee = base fee + utilization fee + origin deposit balancing fee.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -60,24 +64,44 @@ export default class UBAFeeCalculator {
...
+
+    let totalUBAFee = alphaFee.add(utilizationFee);
+    // Resolve the balancing fee tuples that are relevant to this operation
+    const originBalancingFeeTuples = this.config.getBalancingFeeTuples(originChain);
+    const destinationBalancingFeeTuples = this.config.getBalancingFeeTuples(destinationChain);
+
+    // If the action is a deposit, then we need to add the origin and destination balancing fee
+    // to the total UBA fee. We can use the getDepositBalancingFee function to do this
+    // Find both of these fees from the origin and destination chains
+    if (type === "deposit") {
```
mrice32, line 92: Thinking more about it, I'm not sure what the actions really mean here.

If you're trying to construct a deposit, you need to know how much the deposit "action" will cost, but also how much the refund "action" will cost on the other side to approximate the relayer fee (with some cushion). So you're really simulating both actions.

If you're trying to decide where to take your refund, you'll probably only want the gas costs + the refund balancing fee on different chains. However, to perform the relay, you'll still need to know the _actual_ LP fee the depositor should pay, since your relay's validity depends on it.

Thoughts?

mrice32 (commented):
Looks great! A few high-level comments, I dug into the integral code a bit and it looked reasonable, but will go further tomorrow.

src/clients/ConfigStoreClient.ts
```diff
@@ -452,4 +453,32 @@ export class AcrossConfigStoreClient {
       return { current, post };
     }
   }
+
+  /**
+   * Call to onchain to get the UBA config for a given L1 token.
+   * @param l1TokenAddress L1 token address to get UBA config for.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async getUBAConfig(l1TokenAddress: string): Promise<UBAConfig> {
```
pxrl, line 462: Have we considered whether there's a time-based component to this? The UBA config can change from one block to the next, so I think we might need to be able to lookup the UBA config at a particular block/timestamp.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: Updated in the changes below

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: Updated in the changes below

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: Updated in the changes below

src/clients/ConfigStoreClient.ts
james-a-morris: Good point! I've updated it here https://github.com/across-protocol/relayer-v2/pull/589/commits/7ccb2eae2dfb080ecffaac7b090ce3b4d5e63963

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: I'm not entirely sure about this - I'll follow up with you offline.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,121 @@
...
+    relayerFee = relayerFee.add(utilizationFee);
+
+    // Resolve the balancing fee tuples that are relevant to this operation
+    const originBalancingFeeTuples = this.config.getBalancingFeeTuples(originChain);
+    const destinationBalancingFeeTuples = this.config.getBalancingFeeTuples(destinationChain);
+
+    relayerFee = relayerFee.add(
+      getRefundBalancingFee(destinationBalancingFeeTuples, destinationRunningBalance, amount)
+    );
+    lpFee = lpFee.add(getDepositBalancingFee(originBalancingFeeTuples, originRunningBalance, amount));
```
nicholaspai, line 72: Why are we naming this the "LPFee", is it better described as a depositorFee?

src/utils/UBAFeeCalculator/UBAFeeCalculatorWithRefresh.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @param blockNumber An optional block number to act as the current time. If not provided, the current block number will be used. The block number is used to calculate the running balance from the most recent proposed bundle.
+   */
+  public async updateRunningBalance(blockNumber?: number): Promise<void> {
+    const fn = async (spokeClient: SpokePoolClient, spoke: UBASpokeBalanceType) => {
+      // Initially set the blockNumber to a new block with either the given input or the last blockNumber available
+      spoke.blockNumber = blockNumber ?? (await spokeClient.spokePool.getBlockNumber());
+      // Clear the recent request flow
+      spoke.recentRequestFlow = [];
+      // Resolve the most recent flows
+      spoke.recentRequestFlow = await this.getRecentRequestFlow();
```
nicholaspai, line 45: nit: should we be running these promises in parallel?

nicholaspai (approved):
This is a really well written PR.

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: Fair point! Updated here: https://github.com/across-protocol/relayer-v2/pull/589/commits/fe2f503344e177550847a70d8dd21d55b6d3efdc

src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -0,0 +1,121 @@
...
+    relayerFee = relayerFee.add(utilizationFee);
+
+    // Resolve the balancing fee tuples that are relevant to this operation
+    const originBalancingFeeTuples = this.config.getBalancingFeeTuples(originChain);
+    const destinationBalancingFeeTuples = this.config.getBalancingFeeTuples(destinationChain);
+
+    relayerFee = relayerFee.add(
+      getRefundBalancingFee(destinationBalancingFeeTuples, destinationRunningBalance, amount)
+    );
+    depositorFee = depositorFee.add(getDepositBalancingFee(originBalancingFeeTuples, originRunningBalance, amount));
```
mrice32, line 72: +1 -- we may need more granularity here in the future, but LGTM!

mrice32 (approved):
LGTM!

PR #591, opened 2023-04-18
improve: Fetch gas price when sending gas token

nicholaspai:
- This specifically fixes a bug often seen on chains like Matic where the ethers automatic gas price setting usually doesn't work
- Motivation for this was to add MATIC config to monitor's refillBalances logic


james-a-morris (approved):
LGTM! No significant comments to add here

PR #592, opened 2023-04-18
improve(UBA): Support retrieving SpokePool flows

pxrl:
This change adds a helper function to retrieve and sort UBA flows. The final resting place for this functionality is tbd, but for now it can safely live as a utility function in utils. The primary user of this function is expected to be the UBAFeeCalculator class, also in utils.

The implementation implicitly (unfortunately) involves multiple rounds of sorting, by virtue of the fact that the SpokePoolClient already performs in-place sorting on all retrieved events before they are parsed and stored in the SpokePoolClient. This is maintained in order to avoid changing existing SpokePoolClient behaviour, but may be re-evaluated as the UBA implementation matures.

Some additional modifications were needed to the mocked SpokePoolClient implementation. These changes should incidentally make it easier to add further tests in the future.

Fixes: ACX-1004


nicholaspai (commented):
Can you rebase this off across-2.5 branch?

src/interfaces/index.ts
```diff
@@ -69,3 +69,4 @@ export type UbaFlow = interfaces.UbaFlow;
 export type UbaRunningRequest = interfaces.UbaRunningRequest;
 export const isUbaInflow = interfaces.isUbaInflow;
 export const isUbaOutflow = interfaces.isUbaOutflow;
+export const outflowIsFill = interfaces.outflowIsFill;
```
pxrl, line 72: Remove outflowIsFill, since it's not used yet, and it creates some challenges for the ABI jump to the new SpokePools.

```suggestion
```

pxrl:
> Can you rebase this off across-2.5 branch?

It applies cleanly for me after I removed a single line from src/interfaces/index.ts (https://github.com/across-protocol/relayer-v2/pull/592/commits/0cee69445cf184a15f4309282eceab3f5d487188). We should be fine to get this into master and then merge master into across-2.5 :+1:

src/utils/UBA.ts
```diff
@@ -0,0 +1,44 @@
...
+      (fill: FillWithBlock) =>
+        fill.repaymentChainId === spokePoolClient.chainId && fill.blockNumber > fromBlock && fill.blockNumber < toBlock
+    );
+
+  const refundRequests: RefundRequestWithBlock[] = spokePoolClient.getRefundRequests(fromBlock, toBlock);
+
+  // This is probably more expensive than we'd like... @todo: optimise.
+  const flows = sortEventsAscending(
+    (deposits as UbaFlow[]).concat(fills as UbaFlow[]).concat(refundRequests as UbaFlow[])
+  );
```
mrice32, line 41: Prefer to not cast. Is it possible to have these values have the right type from the start?

src/utils/UBA.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // @todo: Must discard subsequent partial fills that are _not_ the first partial fill.
+  const fills: FillWithBlock[] = spokePoolClient
+    .getFills()
+    .filter(
+      (fill: FillWithBlock) =>
+        fill.repaymentChainId === spokePoolClient.chainId && fill.blockNumber > fromBlock && fill.blockNumber < toBlock
+    );
+
+  const refundRequests: RefundRequestWithBlock[] = spokePoolClient.getRefundRequests(fromBlock, toBlock);
```
mrice32, line 36: Since these can be spoofed, are these validated by the spokepoolclient? Or are you expecting the caller to validate them?

src/utils/UBA.ts
```diff
@@ -0,0 +1,44 @@
...
+  fromBlock ??= spokePoolClient.deploymentBlock;
+  toBlock ??= spokePoolClient.latestBlockNumber;
+
+  const deposits: DepositWithBlock[] = spokePoolClient
+    .getDeposits()
+    .filter((deposit: DepositWithBlock) => deposit.blockNumber >= fromBlock && deposit.blockNumber <= toBlock);
+
+  // @todo: Must discard subsequent partial fills that are _not_ the first partial fill.
+  const fills: FillWithBlock[] = spokePoolClient
+    .getFills()
```
mrice32, line 30: Same with these fills -- are fills returned by this function guaranteed to be valid fills?

mrice32 (commented):
Looks great! Just a few minor comments

src/utils/UBA.ts
pxrl: I touched on it briefly in the comment at the top of the function, but this function will aim to outsource as much validation as possible down to the SpokePoolClient - i.e. it should be filtered out as early as possible. Examples of this would be (for the block in which the transaction was confirmed):
 - Are the chainIds valid?
 - Is the refund token valid? 
 - Is the corresponding fill known and valid? (nb. this implies that the fill has been validated in advance).

At this layer, I expect some additional filtering out:
 - Remove slow fills.
 - Remove subsequent partial fills.

src/utils/UBA.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number, toBlock?: number): UbaFlow[] {
+  fromBlock ??= spokePoolClient.deploymentBlock;
+  toBlock ??= spokePoolClient.latestBlockNumber;
+
+  const deposits: DepositWithBlock[] = spokePoolClient
+    .getDeposits()
+    .filter((deposit: DepositWithBlock) => deposit.blockNumber >= fromBlock && deposit.blockNumber <= toBlock);
+
+  // @todo: Must discard subsequent partial fills that are _not_ the first partial fill.
```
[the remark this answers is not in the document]
pxrl: Has been implemented here: https://github.com/across-protocol/relayer-v2/pull/592/commits/4e85a3a5f544772b427dc113697b807043731163.

src/utils/UBA.ts
pxrl: Good call - should be fixed here: https://github.com/across-protocol/relayer-v2/pull/592/commits/52f0dc98364ec0ac2ad31963c3734214cfd1e2e6.

src/utils/UBA.ts
```diff
@@ -0,0 +1,44 @@
...
+  const fills: FillWithBlock[] = spokePoolClient
```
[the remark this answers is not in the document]
pxrl: It works OK if everything is promoted to UbaFlow[] - that was a good suggestion. Fixed inhttps://github.com/across-protocol/relayer-v2/pull/592/commits/52f0dc98364ec0ac2ad31963c3734214cfd1e2e6.

src/utils/UBA.ts
pxrl: Have responded in more detail here: https://github.com/across-protocol/relayer-v2/pull/592#discussion_r1172567155, but in short: yes.

The SpokePoolClient already has to perform a lot of validation logic, and since UBA is a layer above the SpokePoolClient, it should inherit all of that. We _may_ find cases where the existing validation and verification is not as strict as we'd like, but the intention would be to resolve that at the lowest feasible layer - so it should be rare for any of this to ripple up to the UBA layer.

james-a-morris (commented):
LGTM pending the CI tests running

james-a-morris (approved):
LGTM once the most recent test passes.

src/utils/UBA.ts
pxrl: One additional clarification: There is a quirk in that verification of anything that spans multiple SpokePools is difficult to do from within the context of an individual SpokePoolClient (and it feels wrong to try to put it there).

Examples: End-to-end a RefundRequest to a Fill, and a Fill to a Deposit. So there is a set of intermediate functionality that handles verification of events that are implicitly cross-chain, and this code implicitly relies on that verification having already occurred. Will need to be mindful of that...

mrice32: Yep, I agree that that's out of scope!

src/utils/UBA.ts
mrice32: SGTM!

mrice32 (approved):
LGTM!

PR #593, opened 2023-04-18
chore(across-v2.5): Import hardhat config changes

pxrl:
Pulled from Nick's changes on the Across-2.5 branch. Can be merged to master without
affecting prod, so the strategy is to migrate it in advance so that the remaining delta is
smaller and confined (as much as possible) to breaking changes.

Co-authored-by: nicholaspai <9457025+nicholaspai@users.noreply.github.com>

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -15,12 +15,12 @@ import "hardhat-deploy";
 
 dotenv.config();
 
-const solcVersion = "0.8.13";
+const solcVersion = "0.8.18";
 const mnemonic = getMnemonic();
 
 const config: HardhatUserConfig = {
   solidity: {
-    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 } } }],
+    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 }, viaIR: true } }],
   },
   networks: {
     hardhat: { accounts: { accountsBalance: "1000000000000000000000000" } },


hardhat.config.ts
```diff
@@ -15,12 +15,12 @@ import "hardhat-deploy";
...
 dotenv.config();
 
-const solcVersion = "0.8.13";
+const solcVersion = "0.8.18";
 const mnemonic = getMnemonic();
 
 const config: HardhatUserConfig = {
   solidity: {
-    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 } } }],
+    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 }, viaIR: true } }],
```
james-a-morris, line 23: Just to confirm, what does `viaIR` mean?

james-a-morris (commented):
Left a comment

hardhat.config.ts
nicholaspai: It allows us to compile contracts without the "stack too deep" error, which is necessary when compiling the new SpokePool contracts

nicholaspai (approved):

hardhat.config.ts
pxrl: The solidity docs give a nice description:

_The IR-based code generator was introduced with an aim to not only allow code generation to be more transparent and auditable but also to enable more powerful optimization passes that span across functions._

https://docs.soliditylang.org/en/v0.8.19/ir-breaking-changes.html#solidity-ir-based-codegen-changes

PR #595, opened 2023-04-19
refactor(SpokePoolClient): Adjust DepositWithBlock production

pxrl:
This makes the code a bit more compact and requires fewer type assertions, and should also be compatible with both master and across-2.5.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -658,7 +658,6 @@ export class SpokePoolClient {
           realizedLpFeePct: dataForQuoteTime[index].realizedLpFeePct,
           destinationToken: this.getDestinationTokenForDeposit(partialDeposit),
           quoteBlockNumber: dataForQuoteTime[index].quoteBlock,
-          blockNumber: event.blockNumber,
         };
 
         assign(this.depositHashes, [this.getDepositHash(deposit)], deposit);
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -433,23 +433,16 @@ export class SpokePoolClient {
             ` between ${srcChain} blocks [${searchBounds.low}, ${searchBounds.high}]`
         );
       }
-      const processedEvent: {
-        originChainId: number;
-        originToken: string;
-        destinationChainId: number;
-      } = spreadEventWithBlockNumber(event) as DepositWithBlock;
-      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number } = await this.computeRealizedLpFeePct(
-        event
-      );
+      const partialDeposit = spreadEventWithBlockNumber(event) as DepositWithBlock;
+      const { realizedLpFeePct, quoteBlock: quoteBlockNumber } = await this.computeRealizedLpFeePct(event);
       // Append the realizedLpFeePct.
       // Append destination token and realized lp fee to deposit.
       deposit = {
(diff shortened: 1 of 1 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -657,23 +650,16 @@ export class SpokePoolClient {
...
-          destinationToken: this.getDestinationTokenForDeposit(
-            processedEvent as {
-              originChainId: number;
-              originToken: string;
-              destinationChainId: number;
-            }
-          ),
+          destinationToken: this.getDestinationTokenForDeposit(partialDeposit),
           quoteBlockNumber: dataForQuoteTime[index].quoteBlock,
           blockNumber: event.blockNumber,
```
pxrl, line 661:
```suggestion
```

nicholaspai (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -433,23 +433,16 @@ export class SpokePoolClient {
...
-        realizedLpFeePct: dataForQuoteTime.realizedLpFeePct,
-        destinationToken: this.getDestinationTokenForDeposit(processedEvent),
-        quoteBlockNumber: dataForQuoteTime.quoteBlock,
-        blockNumber: event.blockNumber,
-      } as DepositWithBlock;
+        ...partialDeposit,
+        realizedLpFeePct,
+        destinationToken: this.getDestinationTokenForDeposit(partialDeposit),
+        quoteBlockNumber,
+      };
```
james-a-morris, line 445: Good refactor here

james-a-morris (approved):
LGTM

src/clients/SpokePoolClient.ts
```diff
@@ -433,23 +433,16 @@ export class SpokePoolClient {
...
       }
-      const processedEvent: {
-        originChainId: number;
-        originToken: string;
-        destinationChainId: number;
-      } = spreadEventWithBlockNumber(event) as DepositWithBlock;
-      const dataForQuoteTime: { realizedLpFeePct: BigNumber; quoteBlock: number } = await this.computeRealizedLpFeePct(
-        event
-      );
+      const partialDeposit = spreadEventWithBlockNumber(event) as DepositWithBlock;
```
mrice32, line 436: My only minor issue with this change is that we're sort of lying to getDestinationTokenForDeposit about what type it's being passed. Implicitly, we assume that it will not use realizedLpFeePct or destinationToken, but if down the line, somebody decided to access those fields, they wouldn't be present, right?

I don't think the code before was better, to be clear, but in general, I've been tripped up quite a bit by types that claim to have fields x, y, and z, but when you use them, you find that y and z are both not present at runtime.

If there's no way to get this without a lot of complexity, I have no problem with keeping this code as is! Just wanted to point out where this code can mislead the reader, as any reader would assume that partialDeposit has all the deposit fields.

mrice32 (commented):
Looks great! Only one high-level comment

src/clients/SpokePoolClient.ts
pxrl: It's a good observation and I agree on the nit. There are a couple of places where we're a bit flexible with the truth in order to produce a "...WithBlock" type, and it means that extra care is needed in order to avoid triggering the foot gun.

I think we can address this by trading off a little bit of flexibility for increased type safety. I'll follow up with a change to implement that, but it probably makes sense to defer it until after the SpokePool migration.

PR #601, opened 2023-04-20
feat(UBA): :sparkles: Add function to compute a range of fees

james-a-morris:
This PR adds the ability for the UBA Fee Calculator to replay the fees of each of the running events within its running balance.

### DIFF
--- a/src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
+++ b/src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
@@ -37,7 +37,6 @@ export default class UBAFeeCalculator {
   /**
    * @description Get the recent request flow
    * @param action The action to get the fee for
-   * @param tokenSymbol The token symbol to get the fee for
    * @returns The relevant fee
    */
   public async getUBAFee(action: UbaRunningRequest): Promise<UBAFeeResult> {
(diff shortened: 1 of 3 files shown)


src/utils/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -100,6 +99,17 @@ export default class UBAFeeCalculator {
     };
   }
 
+  public getHistoricalUBAFees(flows: UbaFlow[]): Promise<UBAFeeResult[]> {
```
pxrl, line 102: What's the UBA state when these fees are computed?

I'm a bit confused by the fact that this is a "historical" computation, but I don't know what the time/state reference is, since it only uses the amount and type from the flow (i.e. not the block number or UBA sequence).

james-a-morris: Fair point! I've made a modification that should resolve this.

mrice32 (approved):
LGTM! I think this algorithm makes sense and is what we need. One thing we may need to think about once we have all this in is whether this is going to be super costly to compute on the fly and if we need to optimize. I worry there might be some n^2 or n^3 loops floating around that we may need to tame.

james-a-morris:
> LGTM! I think this algorithm makes sense and is what we need. One thing we may need to think about once we have all this in is whether this is going to be super costly to compute on the fly and if we need to optimize. I worry there might be some n^2 or n^3 loops floating around that we may need to tame.

This is a very fair point - I plan to look for some optimization efficiencies once all of these outstanding PRs are merged.

PR #602, opened 2023-04-21
refactor(SpokePoolClient): Relocate validateFillForDeposit

pxrl:
This function compares a Deposit with a Fill and doesn't relate to the SpokePool or SpokePoolClient. Move it to FillUtils instead.

Motivation: This helps to slim down the SpokePoolClient and is one enabler for further functionalities to be relocated.


src/clients/SpokePoolClient.ts
```diff
@@ -49,18 +50,6 @@ export type SpokePoolUpdate = {
   searchEndBlock: number;
 };
 
-const FILL_DEPOSIT_COMPARISON_KEYS = [
```
pxrl, line 52: Note: This array is touched by the across-2.5 branch (`message` field is added). If this goes into master as-is then I will follow up with the corresponding change to align across-2.5.

src/clients/SpokePoolClient.ts
```diff
@@ -223,9 +212,7 @@ export class SpokePoolClient {
     if (depositWithMatchingDepositId === undefined) {
       return undefined;
     }
-    return this.validateFillForDeposit(fillCopy, depositWithMatchingDepositId)
-      ? depositWithMatchingDepositId
-      : undefined;
+    return validateFillForDeposit(fillCopy, depositWithMatchingDepositId) ? depositWithMatchingDepositId : undefined;
```
james-a-morris, line 215: Does the above conditional need to be present now that we have a validate fill function?

james-a-morris (commented):
Left a comment

src/clients/SpokePoolClient.ts
pxrl: Hm..I think so? `validateFillForDeposit()` doesn't expect to receive `deposit = undefined`, so we can't blindly `depositWithMatchingDepositId` in.

Do you see an opportunity to simplify even further?

james-a-morris (commented):
Replied to a comment

src/utils/FillUtils.ts
```diff
@@ -265,3 +277,18 @@ export function getUnfilledDeposits(
     });
   }
 }
+
+// Ensure that each deposit element is included with the same value in the fill. This includes all elements defined
+// by the depositor as well as the realizedLpFeePct and the destinationToken, which are pulled from other clients.
+export function validateFillForDeposit(fill: Fill, deposit?: Deposit): boolean {
+  if (deposit === undefined) {
+    return false;
+  }
```
mrice32, line 286: nit: why do we allow deposit to be optional?

mrice32 (approved):
LGTM!

src/utils/FillUtils.ts
pxrl: Was discussed here: https://github.com/across-protocol/relayer-v2/pull/602#discussion_r1175397986. It saves multiple callers from doing the validation.

src/utils/FillUtils.ts
```diff
@@ -11,6 +11,18 @@ import {
...
+const FILL_DEPOSIT_COMPARISON_KEYS = [
+  "amount",
+  "originChainId",
+  "relayerFeePct",
+  "realizedLpFeePct",
+  "depositId",
+  "depositor",
+  "recipient",
+  "destinationChainId",
+  "destinationToken",
```
pxrl, line 23:
```suggestion
  "destinationToken",
  "message",
```

nicholaspai (approved):
LGTM, long time change incoming

james-a-morris (approved):
LGTM

PR #603, opened 2023-04-21
Fix test after SpokePool contract update

pxrl:
This change fixes the breakage introduced by the recent SpokePool change (to reject partial fills on deposits with a message).

https://github.com/across-protocol/relayer-v2/pull/582 currently depends on this (or alternative fix).

cc @nicholaspai.


nicholaspai (changes requested):
I have a fix for this in this commit 60dc34e3a783aa011c6a4d191a312d8ec3cf70b3 wdyt?

pxrl:
> I have a fix for this in this commit [60dc34e](https://github.com/across-protocol/relayer-v2/commit/60dc34e3a783aa011c6a4d191a312d8ec3cf70b3) wdyt?

Yeah that's even simpler. I wasn't sure whether we actually wanted to verify the reversion or not, but if not then your fix is better.

PR #608, opened 2023-04-25
improve(package): Automate user upgrades

pxrl:
This should make it easier and less error-prone for end users to update their installations.

The intended flow is that the user simply types `yarn update`, and following completion, verifies that the commit hash shown matches their intended target.


james-a-morris (approved):
LGTM

PR #613, opened 2023-04-26
fix: Handle executor for first root bundle ID

nicholaspai:
Doesn't handle case where there are no root bundles yet


src/dataworker/Dataworker.ts
```diff
@@ -1221,10 +1221,8 @@ export class Dataworker {
       );
 
       // We need to know the next root bundle ID for the mainnet spoke pool in order to execute leaves for roots that
-      // will be relayed after executing the above pool rebalance root. However, since we batch together all
-      // transactions, the root bundle won't have been relayed yet. So, we need to infer it by adding 1 to the last
-      // relayed root bundle. This assumes that all previous root bundles have been relayed already.
-      const nextRootBundleIdForMainnet = spokePoolClients[1].getLatestRootBundleId() + 1;
```
nicholaspai, line 1227: The bug is that this always increments 1, which shouldn't be the case for the first root bundle, which really `getLatestRootBundleId` should return -1 for

PR #614, opened 2023-04-26
fix(UBA): Restore tests

pxrl:
The Ethers Event 'args' field is an unusual Array/Object hybrid, so we can dereference its members directly, rather than using their array indices. However, EventUtils doesn't handle recursion down into args members that are themselves Objects, so add that capability in.

This broader event unpacking functionality seems ripe for some re-work in order to address some previously discussed type that's necessary to appease the various type checkers.

Fixes ACX-1057

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -694,21 +694,14 @@ export class SpokePoolClient {
       }
       for (const event of fillEvents) {
         const fill = spreadEventWithBlockNumber(event) as FillWithBlock;
-        /**
-         * struct RelayExecutionInfo {
-         *   address recipient;
-         *   bytes message;
-         *   int64 relayerFeePct;
-         *   bool isSlowRelay;
-         *   int256 payoutAdjustmentPct;
-         * }
-         */
+        // speadEventWithBlockNumber() doesn't recurse down into Objects within event.args.
+        // Unpack updatableRelayData here and perform the necessary type conversions.
         fill.updatableRelayData = {
-          recipient: fill.updatableRelayData[0],
-          message: fill.updatableRelayData[1],
-          relayerFeePct: toBN(fill.updatableRelayData[2]),
-          isSlowRelay: fill.updatableRelayData[3],
-          payoutAdjustmentPct: toBN(fill.updatableRelayData[4]),
+          recipient: fill.updatableRelayData.recipient,
+          message: fill.updatableRelayData.message,
+          relayerFeePct: toBN(fill.updatableRelayData.relayerFeePct),
+          isSlowRelay: fill.updatableRelayData.isSlowRelay,
+          payoutAdjustmentPct: toBN(fill.updatableRelayData.payoutAdjustmentPct),
         };
         assign(this.fills, [fill.originChainId], [fill]);
         assign(this.depositHashesToFills, [this.getDepositHash(fill)], [fill]);
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -694,14 +694,15 @@ export class SpokePoolClient {
       }
       for (const event of fillEvents) {
         const fill = spreadEventWithBlockNumber(event) as FillWithBlock;
+        const { recipient, message, relayerFeePct, isSlowRelay, payoutAdjustmentPct } = fill.updatableRelayData;
         // speadEventWithBlockNumber() doesn't recurse down into Objects within event.args.
         // Unpack updatableRelayData here and perform the necessary type conversions.
         fill.updatableRelayData = {
-          recipient: fill.updatableRelayData.recipient,
-          message: fill.updatableRelayData.message,
-          relayerFeePct: toBN(fill.updatableRelayData.relayerFeePct),
-          isSlowRelay: fill.updatableRelayData.isSlowRelay,
-          payoutAdjustmentPct: toBN(fill.updatableRelayData.payoutAdjustmentPct),
+          recipient,
+          message,
+          isSlowRelay,
+          relayerFeePct: toBN(relayerFeePct),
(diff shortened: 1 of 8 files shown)


nicholaspai (commented):
One question https://github.com/across-protocol/relayer-v2/pull/614/files#r1177877633

src/clients/SpokePoolClient.ts
```diff
@@ -694,21 +694,15 @@ export class SpokePoolClient {
...
-         *   address recipient;
-         *   bytes message;
-         *   int64 relayerFeePct;
-         *   bool isSlowRelay;
-         *   int256 payoutAdjustmentPct;
-         * }
-         */
+        // speadEventWithBlockNumber() doesn't recurse down into Objects within event.args.
+        // Unpack updatableRelayData here and perform the necessary type conversions.
+        const { recipient, message, relayerFeePct, isSlowRelay, payoutAdjustmentPct } = fill.updatableRelayData;
```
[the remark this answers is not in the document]
pxrl: > I parse the `updatableRelayData` as an array currently, how can you destructure it like this ?

It's both an array and an object. It's pretty counter-intuitive. I had to dig into Ethers to figure it out.

You can see that `args` is defined as a type `Result` [here](https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/contracts/src.ts/index.ts#L84), and `Result` is defined as follows [here](https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/abi/src.ts/coders/abstract-coder.ts#L11):
```
export interface Result extends ReadonlyArray<any> {
    readonly [key: string]: any;
}
```

Here is where the args object is updated with the key/value pair: https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/abi/src.ts/interface.ts#L623

test/mocks/MockSpokePoolClient.ts
```diff
@@ -152,7 +152,9 @@ export class MockSpokePoolClient extends SpokePoolClient {
     depositor ??= randomAddress();
 
     const topics = [originChainId, depositId, depositor];
+    const recipient = fill.recipient ?? randomAddress();
```
james-a-morris, line 155: QQ: is there any case where `fill.recipient === ""`? If so, we could replace `??` with `||`

test/UBA.SpokePoolClient.ts
```diff
@@ -220,7 +220,7 @@ describe("UBA: SpokePool Events", async function () {
     });
   });
 
-  it.skip("Correctly filters slow fills", async function () {
```
james-a-morris, line 223: This is a cool feature of `it(...)`

james-a-morris (approved):
LGTM - I'm also interested in the Array/Object nature of the `updatedableRelayData`

src/clients/SpokePoolClient.ts
```diff
@@ -694,22 +694,8 @@ export class SpokePoolClient {
...
-         */
-        fill.updatableRelayData = {
-          recipient: fill.updatableRelayData[0],
-          message: fill.updatableRelayData[1],
-          relayerFeePct: toBN(fill.updatableRelayData[2]),
-          isSlowRelay: fill.updatableRelayData[3],
-          payoutAdjustmentPct: toBN(fill.updatableRelayData[4]),
-        };
+        // speadEventWithBlockNumber() doesn't recurse down into Objects within event.args.
+        // Unpack updatableRelayData here and perform the necessary type conversions.
```
pxrl, line 698:
```suggestion
```

pxrl:
@nicholaspai @james-a-morris I pivoted on this - preferring to make spreadEvent() more capable instead of just gaffer-taping updatableRelayData in SpokePoolClient. It works OK on my end; updatableRelayData ends up as a nice object and we don't need to fiddle with its members anymore. wdyt?

james-a-morris:
> @nicholaspai @james-a-morris I pivoted on this - preferring to make `spreadEvent()` more capable instead of just gaffer-taping `updatableRelayData` in SpokePoolClient. It works OK on my end; `updatableRelayData` ends up as a nice object and we don't need to fiddle with its members anymore. wdyt?

If possible, can you ping me when you merge this PR? I want to make sure these changes make it into the SDK PR :https://github.com/across-protocol/sdk-v2/pull/183

james-a-morris (approved):
LGTM

pxrl:
Have verified the new behaviour by running the validateRootBundle.ts script against the latest root bundle proposal. This one was a nice test because it included a slow fill, so we know that we were evaluating `updatableRelayData.isSlowRelay`.

```
2023-04-26 20:47:07 [debug]: {
  "at": "Dataworker#validate",
  "message": "Pending root bundle matches with expected",
}
2023-04-26 20:47:07 [info]: {
  "at": "RootBundleValidator",
  "message": "Validation results",
  "rootBundle": {
    "poolRebalanceRoot": "0x4410ba300c57266a9c30e22962c8e13cdbeb3f71c3cec16868c36d2ef493dbfb",
    "relayerRefundRoot": "0xbcede773d988d694dafa75ebca10d4c1d81428803a1e68f8136615013158839d",
    "slowRelayRoot": "0x69ac71e6b91398011a030cf4e01c70291c59b0a2f114adc321bc7f0bc01ebc6e",
    "proposer": "0xf7bAc63fc7CEaCf0589F25454Ecf5C2ce904997c",
    "unclaimedPoolRebalanceLeafCount": 4,
    "challengePeriodEndTimestamp": 1682548247,
    "bundleEvaluationBlockNumbers": [
      17132754,
      94534954,
      41992767,
      977154,
      84600904
    ],
    "proposalBlockNumber": 17132782
  },
  "valid": true,
}
```

nicholaspai (approved):

PR #616, opened 2023-04-26
improve(Monitor): Update refill balances log

nicholaspai:


src/monitor/Monitor.ts
```diff
@@ -436,8 +437,8 @@ export class Monitor {
                 message: `Reloaded ${ethers.utils.formatUnits(
                   deficit,
                   decimals
-                )} ETH for ${account} from ${signerAddress} 🫡!`,
-                transactionHash: receipt.transactionHash,
+                )} ${nativeSymbolForChain} for ${account} from ${signerAddress} 🫡!`,
+                transactionHash: etherscanLink(receipt.transactionHash, chainId),
```
james-a-morris, line 441: Nit: It looks like we use `nativeSymbolForChain` only in this else statement. Would it make sense to move the instantiation of that variable into this else block?

james-a-morris (approved):
LGTM - one nit

src/monitor/Monitor.ts
nicholaspai: Agreed, i was just testing it but i'll move it there good reminder

PR #626, opened 2023-04-28
refactor(HubPoolClient): Subdivide update() method

pxrl:
Split it up to permit the external RPC lookups to be mocked. This follows a similar pattern to the recent SpokePoolClient changes.


src/clients/HubPoolClient.ts
```diff
@@ -427,43 +450,70 @@ export class HubPoolClient {
...
+    this.logger.debug({
+      at: "HubPoolClient",
+      message: "Updating HubPool client",
+      searchConfig,
+      eventNames,
+    });
+    const timerStart = Date.now();
+    const [currentTime, pendingRootBundleProposal, ...events] = await Promise.all([
+      this.hubPool.getCurrentTime({ blockTag: latestBlockNumber }),
+      this.hubPool.rootBundleProposal({ blockTag: searchConfig.toBlock }),
```
pxrl, line 483: Note one change here: previously, `getCurrentTime` and `rootBundleProposal` were performed against no specific block, so there was potential for them to resolve to different blocks (i.e. if the RPC requests are routed to different hosts).

The strategy now is to resolve the "latestBlockNumber" at the very beginning, and use that as the reference point for subsequent lookups via searchConfig.toBlock (or whatever value it's overridden with).

src/clients/HubPoolClient.ts
```diff
@@ -427,43 +450,70 @@ export class HubPoolClient {
...
+
+    this.logger.debug({
+      at: "HubPoolClient",
+      message: "Updating HubPool client",
+      searchConfig,
+      eventNames,
+    });
+    const timerStart = Date.now();
+    const [currentTime, pendingRootBundleProposal, ...events] = await Promise.all([
+      this.hubPool.getCurrentTime({ blockTag: latestBlockNumber }),
```
pxrl, line 482:
```suggestion
      this.hubPool.getCurrentTime({ blockTag: searchConfig.toBlock }),
```

src/clients/HubPoolClient.ts
```diff
@@ -13,6 +15,15 @@ import { ExecutedRootBundle, PendingRootBundle, ProposedRootBundle } from "../in
 import { CrossChainContractsSet, DestinationTokenWithBlock, SetPoolRebalanceRoot } from "../interfaces";
 import _ from "lodash";
 
+type HubPoolUpdate = {
+  success: boolean;
+  currentTime: number;
+  latestBlockNumber: number;
+  pendingRootBundleProposal: PendingRootBundle;
+  events: Record<string, Event[]>;
```
pxrl, line 23: Instead of a 2d-array as per the SpokePoolClient, events is an Object that's indexed by the exact event name that was queried.

src/clients/HubPoolClient.ts
```diff
@@ -13,6 +15,15 @@ import { ExecutedRootBundle, PendingRootBundle, ProposedRootBundle } from "../in
 import { CrossChainContractsSet, DestinationTokenWithBlock, SetPoolRebalanceRoot } from "../interfaces";
 import _ from "lodash";
 
+type HubPoolUpdate = {
+  success: boolean;
+  currentTime: number;
+  latestBlockNumber: number;
+  pendingRootBundleProposal: PendingRootBundle;
```
nicholaspai, line 22: Why define `pendingRootBundleProposal` here?  Doesn't the client have this available as a state variable (i.e. this. pendingRootBundle)

src/clients/HubPoolClient.ts
```diff
@@ -427,43 +450,70 @@ export class HubPoolClient {
...
-      paginatedEventQuery(this.hubPool, this.hubPool.filters.RootBundleExecuted(), searchConfig),
-      paginatedEventQuery(this.hubPool, this.hubPool.filters.CrossChainContractsSet(), searchConfig),
-      this.hubPool.rootBundleProposal(),
+      this.logger.warn("Invalid update() searchConfig.", { searchConfig });
+      return {
+        success: false,
+        currentTime: this.currentTime,
+        latestBlockNumber: this.latestBlockNumber,
+        pendingRootBundleProposal: this.pendingRootBundle,
+        searchEndBlock: searchConfig.fromBlock - 1,
```
nicholaspai, line 469: Shouldn't `searchEndBlock` be equal to `searchConfig.toBlock`?

nicholaspai (commented):
I find the `HubPoolUpdate` type a bit strange. I would expect `success` to be returned as well as a `searchConfig`, but the other types like `currentTime`, `latestBlockNumber` and `pendingRootBundleProposal` don't make intuitive sense to me

src/clients/HubPoolClient.ts
pxrl: This scenario would indicate a failed update, so we don't actually want to migrate past where we are currently in our searchConfig. This way, when a subsequent `_update()` is called, we'll repeat the current search (and hopefully succeed this time).

In practice these variables are not actually used, since the calling update() method first evaluates the `success` variable, and if it's false, returns immediately without reading any additional variables. However, because we're returning a typed object, this is necessary to keep tsc happy.

src/clients/HubPoolClient.ts
pxrl: Yeah, pendingRootBundle is available as a private class member of the HubPoolClient. The reason for passing it back via a HubPoolUpdate type is so that _update() never modifies HubPoolClient state, so the caller can receive it all as a single Object and update state in a single place, rather than having various sections of code update them incrementally. This also eases test by making it easier to mock out `_update()`.

src/clients/HubPoolClient.ts
```diff
@@ -45,6 +56,18 @@ export class HubPoolClient {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
   }
 
+  protected hubPoolEventFilters(): { [eventName: string]: EventFilter } {
+    return {
+      SetPoolRebalanceRoot: this.hubPool.filters.SetPoolRebalanceRoute(),
```
pxrl, line 61: Typo:

```suggestion
      SetPoolRebalanceRoute: this.hubPool.filters.SetPoolRebalanceRoute(),
```

src/clients/HubPoolClient.ts
```diff
@@ -479,7 +529,7 @@ export class HubPoolClient {
       );
     }
 
-    for (const event of poolRebalanceRouteEvents) {
+    for (const event of events["SetPoolRebalanceRoot"]) {
```
pxrl, line 532:
```suggestion
    for (const event of events["SetPoolRebalanceRoute"]) {
```

src/clients/HubPoolClient.ts
```diff
@@ -479,7 +529,7 @@ export class HubPoolClient {
       );
     }
 
-    for (const event of poolRebalanceRouteEvents) {
+    for (const event of events["SetPoolRebalanceRoute"]) {
       const args = spreadEventWithBlockNumber(event) as SetPoolRebalanceRoot;
```
pxrl, line 533: @nicholaspai Do we have a mistake in the naming of event `SetPoolRebalanceRoot` ? Seems like it should instead be named `SetPoolRebalanceRoute`, and is easily confused with the `poolRebalanceRoot` member of a root bundle proposal, which is something completely different.

nicholaspai: I'm not sure where you're getting `SetPoolRebalanceRoot` from, it looks like the HubPool's event is named `SetPoolRebalanceRoute` did you accidentally confuse them?

src/clients/HubPoolClient.ts
mrice32: I would argue that a clearer way to illustrate this in typescript is this return type:
```
async _update(eventNames: string[]): Promise<{ success: true, currentTime: number ... } | { success: false }>
```

Note: I'm just using shorthand here to make the point clear, but the basic idea is that success _must_ be true for the other fields to be defined. If false, the caller must assume the other fields are not present. This means that caller _must_ check success and is unable to use fields that aren't intended to be used.

src/clients/HubPoolClient.ts
```diff
@@ -45,6 +56,18 @@ export class HubPoolClient {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
   }
 
+  protected hubPoolEventFilters(): { [eventName: string]: EventFilter } {
+    return {
+      SetPoolRebalanceRoute: this.hubPool.filters.SetPoolRebalanceRoute(),
```
mrice32, line 61: nit: it might be nice to have a type that defines all the event names we use:

```ts
type HubEvents = "SetPoolRebalanceRoute" | "L1TokenEnabledForLiquidityProvision" | ...;

...

protected hubPoolEventFilters(): { [eventName: HubEvents]: EventFilter } {
```

Using this type throughout this class would mean that we get compile-time checking for things like typos or other issues when accessing objects whose keys are those event names. Not a huge deal either way, but just a way to get more help from the compiler.

mrice32 (approved):
LGTM!

src/clients/HubPoolClient.ts
pxrl: Have implemented this here: https://github.com/across-protocol/relayer-v2/pull/626/commits/2934bca9ecf72ddbcfea85f5d86d19b24756b2ff

There are still some gaps, and I did need to add a type assertion on the default "no events requested" case, because tsc (fairly) doesn't recognise that `Object.keys(hubPoolEvents)` overlaps 100% with `type HubPoolEvent` in practice.

src/clients/HubPoolClient.ts
pxrl: Good suggestion - implemented: https://github.com/across-protocol/relayer-v2/pull/626/commits/04e05070ff8603ecfa4a0328275a8062531c8b1c

PR #632, opened 2023-05-01
fix: Make entire package a module

nicholaspai:
Addresses the [err](https://stackoverflow.com/questions/58211880/uncaught-syntaxerror-cannot-use-import-statement-outside-a-module-when-import)


james-a-morris (commented):
Is this related to the GCP issue? If so, I think this will produce more errors than it will solve.

PR #633, opened 2023-05-01
chore(package): Bump minimum nodejs version

pxrl:
In practice, relayer-v2 won't run on node 12.x, and supporting older versions makes it difficult to support the relayer because some users see weird issues on older versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "AGPL-3.0-only",
   "private": true,
   "engines": {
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):
This is a good change that I don't think will have too many draws (outside of compatibility). 

I haven't noticed this as an issue because I always run Node 16+, so this is a good call and great catch!

nicholaspai (approved):

PR #635, opened 2023-05-01
improve(multicallerclient): Log multicaller execution failures

nicholaspai:
This  PR tries to filter out any errors that occur when a transaction is successfully simulated but then fails when it is mined. This can happen for the relayer commonly if they get front-run by another 
relay, which is expected behavior

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -117,7 +117,8 @@ export class MultiCallerClient {
         if (isPromiseFulfilled(result)) {
           return [chainId, { result: result.value.map((txnResponse) => txnResponse.hash), isError: false }];
         } else {
-          return [chainId, { result: result.reason, isError: true }];
+          // Don't mark transaction as failed if error reason is a known benign one.
+          return [chainId, { result: result.reason, isError: !knownRevertReasons.has(result.reason) }];
         }
       })
     );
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -117,26 +117,29 @@ export class MultiCallerClient {
...
+          return [chainId, { result: result.reason, isError: !knownRevertReasons.has(result.reason) }];
         }
       })
     );
 
     // We need to iterate over the results to determine if any of the transactions failed.
     // If any of the transactions failed, we need to log the results and throw an error. However, we want to
     // only log the results once, so we need to collate the results into a single object.
-    const failedChains = Object.entries(txnHashes)
-      .filter(([, { isError }]) => isError)
```
james-a-morris, line 129: Could we improve this filter and remove the ignorable reversions here?

james-a-morris (commented):
Left a comment

src/clients/MultiCallerClient.ts
nicholaspai: Actually I think this filter won't help that much, first things first i'd like to just log the failures better

src/clients/MultiCallerClient.ts
```diff
@@ -99,7 +99,7 @@ export class MultiCallerClient {
     const chainIds = [...new Set(Object.keys(this.valueTxns).concat(Object.keys(this.txns)))];
 
     // One promise per chain for parallel execution.
-    const results = await Promise.allSettled(
+    const resultsByChain = await Promise.allSettled(
```
nicholaspai, line 102: This is a more intuitive name IMO as results is actually a list of PromiseSettledResults for each chain. So i'm curious what a failure looks like if one of the txns fails

PR #639, opened 2023-05-02
improve(test): Run PR tests in parallel

pxrl:
For Across I noticed that we run our tests serialised, and it usually takes about 15 minutes to complete. When we have lots of PRs queued up it can mean a long delay between having a PR ready to merge, and actually getting it merged (in my case I often miss an opportunity to submit my PRs, and then have to re-run after merging in the latest changes).

[Hardhat supports running tests in parallel](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#running-tests-in-parallel), and I think we can live with the drawbacks that come with it. On my quad-core host, I can complete the same test in about 4 - 5 minutes. [I tried this in CI here](https://github.com/across-protocol/relayer-v2/actions/runs/4865938927/jobs/8676886658?pr=639) and it took 12 minutes, so I assume that the CI workers only have two CPU cores available, rather than the 4 I have in my system.


pxrl:
Just testing to see how fast it completes (...and whether it blows up).

nicholaspai (approved):

pxrl:
Note: There seems to be some variance in the time to complete tests in parallel. I've seen ~12 minutes and ~14 minutes so far. When tests are run in parallel, the test execution order is randomised; this might impact the end-to-end execution time.

pxrl:
Will hold off on merging this until we've had time to discuss it in quroum.

mrice32 (approved):
This seems good to me!

pxrl:
Circling back to this one...it unfortunately seems unfeasible to enable parallel tests.

At best, the execution environment only has 2 CPU cores available, but there's also some hint that there might occasionally be only 1 available core: https://github.com/orgs/community/discussions/37306

It is theoretically possible to get better-resourced workers ("large runners"), but it's an experimental feature that requires a special request to GitHub, and it requires the org to migrate to GitHub Teams or Enterprise.

https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners

PR #643, opened 2023-05-03
refactor(test): Factor out event generation functionalities

pxrl:
This slims down the MockSpokePoolClient implementation and makes the generic implementation available for use in the MockHubPoolClient, which will get some new capabilities in a following commit.

Also:
 - Support generating "EnabledDepositRoute" SpokePool events.
 - Fill out placeholder SpokePool event signatures and topics.
 - Automagically generate topic[0] for each event based on the event signature.
 - Don't permit fractional units of time in SpokePoolClient update.
 - Cleanup incorrect isSlowRelay member from old fill events during test.
 - Export MockSpokePoolClient from test/utils.


test/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
...
+  transactionIndex?: number;
+};
+
+export class EventManager {
+  private logIndexes: Record<string, number> = {};
+
+  constructor(public readonly eventSignatures: Record<string, string>) {}
+
+  generateEvent(inputs: EthersEventTemplate): Event {
+    const { address, event, _topics, data, args } = inputs;
```
pxrl, line 25:
```suggestion
    const { address, event, topics: _topics, data, args } = inputs;
```

test/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
...
+    const decodeError = new Error(`${event} decoding error`);
+    const removeListener = (): void => {
+      return;
+    };
+
+    blockNumber ??= random(1, 100_000, false);
+    transactionIndex ??= random(1, 32, false);
+    const transactionHash = ethers.utils.id(
+      `Across-v2-${event}-${blockNumber}-${transactionIndex}-${random(1, 100_000)}`
+    );
```
pxrl, line 54:
```suggestion

    blockNumber ??= random(1, 100_000, false);
    transactionIndex ??= random(1, 32, false);
    const transactionHash = ethers.utils.id(
      `Across-v2-${event}-${blockNumber}-${transactionIndex}-${random(1, 100_000)}`
    );

    const _logIndex = `${blockNumber}-${transactionIndex}`;
    this.logIndexes[_logIndex] ??= 0;
    const logIndex = this.logIndexes[_logIndex]++;

    // Populate these Event functions, even though they appear unused.
    const getBlock = async (): Promise<Block> => {
      return {} as Block;
    };
    const getTransaction = async (): Promise<TransactionResponse> => {
      return {} as TransactionResponse;
    };
    const getTransactionReceipt = async (): Promise<TransactionReceipt> => {
      return {} as TransactionReceipt;
    };
    const decodeError = new Error(`${event} decoding error`);
    const removeListener = (): void => {
      return;
    };
```

test/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
+import { random } from "lodash";
+import { Event } from "../../src/utils";
+import { ethers } from "../utils";
+
+type Block = ethers.providers.Block;
+type TransactionResponse = ethers.providers.TransactionResponse;
+type TransactionReceipt = ethers.providers.TransactionReceipt;
+
+export type EthersEventTemplate = {
```
pxrl, line 9:
```suggestion
type EthersEventTemplate = {
```

test/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
...
+    // Populate these Event functions, even though they appear unused.
+    const getBlock = async (): Promise<Block> => {
+      return {} as Block;
+    };
+    const getTransaction = async (): Promise<TransactionResponse> => {
+      return {} as TransactionResponse;
+    };
+    const getTransactionReceipt = async (): Promise<TransactionReceipt> => {
+      return {} as TransactionReceipt;
+    };
```
mrice32, line 50: Should we error instead of just returning an empty object? Or do you expect them to be used somewhere?

mrice32 (approved):
LGTM! Nice idea!

nicholaspai (approved):

test/mocks/MockEvents.ts
pxrl: From what I can see, we probably never use these (and it would be interesting to know otherwise!). Have relocated them, and now they throw errors: https://github.com/across-protocol/relayer-v2/pull/643/commits/f0e2ae643535f2cbc1021c93f83bf3c4bf49aa11.

PR #646, opened 2023-05-04
improve(UBA): Utility to find UBA period opening balance and block

pxrl:
Notes:
   - Introduction of the UBA will allow alt-L1 Spoke Pools. Where possible, effort is made to avoid terms like `l1Token` and `l2Token`, preferring isntead `hubPoolToken` and `spokePoolToken`.
 - The function `getUBAFlows()` from UBA.ts will probably be migrated into the UBAClient class.
 - Since there is now a `UBAClient` class, this probably belongs under src/clients. However, I'll hold off on that migration because this will ultimately be moved to the SDK anyway.

Fixes ACX-1048


test/mocks/MockHubPoolClient.ts
```diff
@@ -85,4 +104,104 @@ export class MockHubPoolClient extends HubPoolClient {
...
+    const event = "SetPoolRebalanceRoute";
+
+    const topics = [destinationChainId, l1Token, destinationToken];
+    const args = {
+      destinationChainId,
+      l1Token,
+      destinationToken,
+    };
+
+    // console.log(`HubPoolClient latestBlockNumber is ${this.latestBlockNumber}.`);
```
pxrl, line 126:
```suggestion
```

test/mocks/MockHubPoolClient.ts
```diff
@@ -85,4 +104,104 @@ export class MockHubPoolClient extends HubPoolClient {
...
+
+    return this.eventManager.generateEvent({
+      event,
+      address: this.hubPool.address,
+      topics: topics.map((topic) => topic.toString()),
+      args,
+      blockNumber: this.latestBlockNumber + 1,
```
pxrl, line 133:
```suggestion
      blockNumber: this.firstBlockToSearch,
```

test/mocks/MockHubPoolClient.ts
```diff
@@ -85,4 +104,104 @@ export class MockHubPoolClient extends HubPoolClient {
...
+      runningBalances,
+      caller,
+    };
+
+    return this.eventManager.generateEvent({
+      event,
+      address: this.hubPool.address,
+      topics: topics.map((topic) => topic.toString()),
+      args,
+      blockNumber: this.latestBlockNumber + 1,
```
pxrl, line 204:
```suggestion
      blockNumber: this.firstBlockToSearch,
```

src/clients/HubPoolClient.ts
```diff
@@ -61,10 +61,12 @@ export class HubPoolClient {
   constructor(
     readonly logger: winston.Logger,
     readonly hubPool: Contract,
+    public deploymentBlock = 0,
     readonly chainId: number = 1,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 }
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
+    this.latestBlockNumber = deploymentBlock === 0 ? deploymentBlock : deploymentBlock - 1;
```
pxrl, line 69:
```suggestion
```

src/utils/UBA.ts
```diff
@@ -46,3 +47,44 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+  }
+
+  getOpeningBalance(
+    chainId: number,
+    spokePoolToken: string,
+    hubPoolBlockNumber?: number
+  ): { balance: BigNumber; blockNumber: number } {
+    assert(Array.isArray(this.closingBlockNumbers[chainId]), `Invalid chainId: ${chainId}`);
+
+    hubPoolBlockNumber = this.hubPoolClient.latestBlockNumber;
```
pxrl, line 76:
```suggestion
    hubPoolBlockNumber ??= this.hubPoolClient.latestBlockNumber;
```

test/UBA.HubPoolClient.ts
```diff
@@ -0,0 +1,125 @@
...
+  toBNWei,
+} from "./utils";
+import { MockHubPoolClient, MockSpokePoolClient } from "./mocks";
+
+type Event = ethers.Event;
+
+let spokePoolClients: { [chainId: number]: MockSpokePoolClient };
+let hubPool: Contract, dai: Contract, weth: Contract;
+let uba: UBAClient;
+// let relayer: SignerWithAddress;
```
pxrl, line 22:
```suggestion
```

test/UBA.HubPoolClient.ts
```diff
@@ -0,0 +1,125 @@
...
+let hubPoolClient: MockHubPoolClient;
+let hubPoolDeploymentBlock: number;
+
+const logger = createSpyLogger().spyLogger;
+
+const chainIds = [10, 137];
+
+describe("UBA: HubPool Events", async function () {
+  beforeEach(async function () {
+    // [relayer] = await ethers.getSigners();
```
pxrl, line 32:
```suggestion
```

test/UBA.HubPoolClient.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    ({ hubPool, dai, weth } = await hubPoolFixture());
+    hubPoolDeploymentBlock = random(1, 100, false);
+    hubPoolClient = new MockHubPoolClient(logger, hubPool, hubPoolDeploymentBlock);
+    await hubPoolClient.update(); // Set initial latestBlockNumber
```
pxrl, line 37:
```suggestion
    await hubPoolClient.update();
```

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
 }
+
+export class UBAClient {
+  private closingBlockNumbers: { [chainId: number]: number[] };
+
+  // Note: chainIds is used to index members of root bundle proposals submitted to the HubPool.
+  // It must include the complete set of chain IDs ever supported by the HubPool.
+  // SpokePoolClients may be a subset of the SpokePools that have been deployed.
+  constructor(
+    private readonly chainIds: number[],
```
nicholaspai, line 58: I prefer naming this list `chainIdIndices` or something including "indices" to make it clear its not just any listof chain ID's, the order matters, right?

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+    private readonly spokePoolClients: { [chainId: number]: SpokePoolClient }
+  ) {
+    assert(chainIds.length > 0, "No chainIds provided");
+    assert(Object.values(spokePoolClients).length > 0, "No SpokePools provided");
+    this.closingBlockNumbers = Object.fromEntries(this.chainIds.map((chainId) => [chainId, []]));
+  }
+
+  private resolveClosingBlockNumber(chainId: number, blockNumber: number): number {
+    const endBlock = this.hubPoolClient.getLatestBundleEndBlockForChain(this.chainIds, blockNumber, chainId);
+    return endBlock ?? this.spokePoolClients[chainId].deploymentBlock;
```
nicholaspai, line 69: Should it be deployment block + 1?

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+  }
+
+  getOpeningBalance(
+    chainId: number,
+    spokePoolToken: string,
+    hubPoolBlockNumber?: number
+  ): { balance: BigNumber; blockNumber: number } {
+    assert(Array.isArray(this.closingBlockNumbers[chainId]), `Invalid chainId: ${chainId}`);
```
nicholaspai, line 77: Should we be make an even stronger assertion that the closingBlockNumber values are all of length 2?

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+    assert(Object.values(spokePoolClients).length > 0, "No SpokePools provided");
+    this.closingBlockNumbers = Object.fromEntries(this.chainIds.map((chainId) => [chainId, []]));
+  }
+
+  private resolveClosingBlockNumber(chainId: number, blockNumber: number): number {
+    const endBlock = this.hubPoolClient.getLatestBundleEndBlockForChain(this.chainIds, blockNumber, chainId);
+    return endBlock ?? this.spokePoolClients[chainId].deploymentBlock;
+  }
+
+  getOpeningBalance(
```
nicholaspai, line 72: How do you intend to use this function? Some examples would help me understand how its different from just calling `getRunningBalanceBeforeBlockForChain`. Why would you want to look up the opening balance before a `hubPoolBlockNumber`: converting the hubPool block into a spoke pool block and getting the running balance before that time

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+    hubPoolBlockNumber ??= this.hubPoolClient.latestBlockNumber;
+
+    const hubPoolToken = this.hubPoolClient.getL1TokenCounterpartAtBlock(chainId, spokePoolToken, hubPoolBlockNumber);
+    if (!isDefined(hubPoolToken)) {
+      throw new Error(`Could not resolve ${chainId} token ${spokePoolToken} at block ${hubPoolBlockNumber}`);
+    }
+
+    const spokePoolClient = this.spokePoolClients[chainId];
+    let blockNumber = spokePoolClient.deploymentBlock;
+    const prevEndBlock = this.resolveClosingBlockNumber(chainId, hubPoolBlockNumber);
```
nicholaspai, line 88: Just wondering, have you seen the logic [here](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/DataworkerUtils.ts#L111) that returns the the start block for an end block? I think you should try to reuse some of the logic because there are important edge cases to handle, such as when the `chainId` is not found in the `this.chainIds` list

nicholaspai (commented):
I'm struggling to see how this is going to be used for UBA functions and why existing functions can't achieve that?

src/utils/UBA.ts
pxrl: Hm, good point - but it should be -1 then right?

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+    if (prevEndBlock > blockNumber) {
```
mrice32, line 89: Can you explain the cases this is meant to handle? The logic seems a little opaque to me. Are we trying to skip this when there have been no bundles for that chain before and so the prevEnvBlock === spokePool.deploymentBlock === blockNumber (if this is the idea, then it seems like it might be clearer just to return null from resolveClosingBlockNumber in that case)? Or some weird case where the spoke pool was upgraded after the last bundle that included this chain or something?

Would be great to see a quick comment on this!

src/utils/UBA.ts
```diff
@@ -46,3 +47,55 @@ export function getUBAFlows(spokePoolClient: SpokePoolClient, fromBlock?: number
...
+
+  private resolveClosingBlockNumber(chainId: number, blockNumber: number): number {
+    const endBlock = this.hubPoolClient.getLatestBundleEndBlockForChain(this.chainIds, blockNumber, chainId);
+    return endBlock ?? this.spokePoolClients[chainId].deploymentBlock;
+  }
+
+  getOpeningBalance(
+    chainId: number,
+    spokePoolToken: string,
+    hubPoolBlockNumber?: number
```
mrice32, line 75: What does it mean to leave this undefined?

mrice32 (commented):
Looks great! Only two minor comments/questions.

src/utils/UBA.ts
pxrl: This method uses `getRunningBalanceBeforeBlockForChain()` under the hood, but it also resolves the starting block of the UBA period and returns them as a combined object. It's mostly a convenience function so a caller can easily resolve the starting block and opening balance, and can proceed to compute inflows and outflows from there.

james-a-morris (commented):
Left a few comments

nicholaspai (approved):

src/utils/UBA.ts
pxrl: There were various places in the HubPoolClient that are capable of returning 0 on this path. For instance, if no bundle has been executed, or if the chainId was not found in the bundle. I'd have preferred to resolve it via a ternary statement:

`
const blockNumber = this.resolveClosingBlockNumber(chainId, hubPoolBlockNumber) || spokePoolClient.deploymentBlockNumber;
`

The challenge is that the former path requires incrementing the block to the _next_ block, whereas the fallback does not. This makes it clunky to work with, and I didn't see any obvious way to do it better than this (which I think is unacceptable).

`
const blockNumber = 1 + (this.resolveClosingBlockNumber(chainId, hubPoolBlockNumber) || (spokePoolClient.deploymentBlock - 1));
`

PR #647, opened 2023-05-04
improve: Refuse to execute SlowFill leaves for deposits with messages

nicholaspai:
Protects dataworker from executing fill for deposit with very expensive message

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -984,16 +984,32 @@ export class Dataworker {
   }
 
   async _executeSlowFillLeaf(
-    leaves: SlowFillLeaf[],
+    _leaves: SlowFillLeaf[],
     balanceAllocator: BalanceAllocator,
     client: SpokePoolClient,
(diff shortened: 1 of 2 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -984,16 +984,32 @@ export class Dataworker {
...
-    if (leaves.length === 0) {
+    if (_leaves.length === 0) {
       return;
     }
+
+    // Ignore slow fill leaves for deposits with messages as these messages might be very expensive to execute.
+    // The original depositor can always execute these and pay for the gas themselves.
+    const leaves = _leaves.filter((leaf) => {
+      if (leaf.relayData.message !== "0x") {
+        this.logger.warn({
```
mrice32, line 1002: Should we make this an error (just for now)? Since we don't expect any messages?

mrice32 (approved):
LGTM!

src/dataworker/Dataworker.ts
nicholaspai: yea thats safer +1

PR #652, opened 2023-05-05
improve(dataworker): Print out slow leaf execution args if ignoring

nicholaspai:
This way we can more easily manually execute the leaf to quiet alerts, or pass the `recipient` the execution args to execute themselves


src/dataworker/Dataworker.ts
```diff
@@ -993,12 +993,25 @@ export class Dataworker {
...
+            relayData.amount,
+            relayData.originChainId,
+            relayData.realizedLpFeePct,
+            relayData.relayerFeePct,
+            relayData.depositId,
+            rootBundleId,
+            relayData.message,
+            payoutAdjustmentPct,
+            slowRelayTree.getHexProof({ relayData, payoutAdjustmentPct }),
+          ],
```
james-a-morris, line 1014: +1 for printing only the necessary items. 

Nit: would this make sense to include in the FormattingUtils?

james-a-morris (commented):
Left a nit

src/dataworker/Dataworker.ts
nicholaspai: which part would be included in `FormattingUtils`?

james-a-morris: My thought would be the code from 1002 to 1013 (the code that decomposes the object into an array)

PR #656, opened 2023-05-06
improve(ProviderUtils): New RPC provider config syntax

pxrl:
This commit introduces new syntax for RPC provider configuration. With this change, providers can be configured independently with the following env var:

  `RPC_PROVIDER_<PROVIDER_NAME>_<CHAIN_ID>=<ENDPOINT_URL>`

The use and preference of a provider is determined by the ordering of its appearance a comma-separated string in the `RPC_PROVIDERS` env var:

  `RPC_PROVIDERS=<PROVIDER_1>,<PROVIDER_...>,<PROVIDER_n>`

Chain-specific provider preferences can be expressed by appending `_<CHAIN_ID>` to the env var:

  `RPC_PROVIDERS_1=<PROVIDER_1>,<PROVIDER_...>,<PROVIDER_n>`

The new config parser will optimistically try for the new syntax, and if it fails, will silently fall back to the existing NODE_URL_X and NODE_URLS_X syntax. This allows users the opportunity for a seamless migration. Ultimately, I'd propose to deprecate the existing NODE_URL* syntax after one or more weeks, to allow sufficient time for third-party operators to adjust their config.

Provider configuration examples:

  `RPC_PROVIDER_INFURA_1=https://mainnet.infura.io/v3/...`
  `RPC_PROVIDER_ALCHEMY_1=https://eth-mainnet.g.alchemy.com/v2/...`
  `RPC_PROVIDER_QUICKNODE_1=...`

  `RPC_PROVIDERS=INFURA,ALCHEMY,QUICKNODE`

To re-order, such that Alchemy is the preferred provider on Optimism:

  `RPC_PROVIDERS_10=ALCHEMY,INFURA,QUICKNODE`

To specify only a single RPC provider for a specific chain:

  `RPC_PROVIDERS_1=<PROVIDER>`

Some advantages of this new approach are:
 - The same syntax is used for when there are one or multiple RPC providers. The existing configuration can be unintuitive because both `NODE_URL_X` and `NODE_URLS_X` is supported, and this can confuse users.

 - No JSON is needed, so JSON parsing errors can not trigger a failure. This is especially convenient for users who manage their environemnt via .env.

 - When multiple providers are used, re-ordering the providers does not require touching the endpoint URLs, which are often  long in length and cumbersome to adjust.

 - It's OK to leave a provider defined (uncommented) in the environment, since only the RPC providers referenced by the `RPC_PROVIDERS` (or `RPC_PROVIDERS_<CHAIN_ID>`) env var will be used. This should make testing easier for third parties.

 - Per-provider configuration is an obvious follow-on (i.e. permitting a unique block range look back to be specified for each provider, rather than a global config per chain.


nicholaspai (commented):
Would love to see an example for why this is a quality of life improvement

pxrl:
> Would love to see an example for why this is a quality of life improvement

I've made a draft PR here to serve as an example of how it'd be used: https://github.com/UMAprotocol/bot-configs/pull/1193

In addition to what I've written in the PR message, I'll throw up reasons I think this is a useability improvement:
 - For RL to change provider priorities in bot-configs, it's a one-line change (vs. a 3 line change at present).
 - We've discussed previously that it's preferred to avoid JSON in env vars, since it's additional complexity and throws up unhelpful errors when we're parsing config. It's especially annoying to troubleshoot JSON-encoding errors when working directly in a .env file.
 -  NODE_URLS_X env vars end up especially long (easily >120 characters for >1 defined endpoint). Most people will end up with line wrapping. It's currently not possible to spread these over multiple lines because [multi-line env vars are broken in dotenv 16 for almost 1 year](https://github.com/motdotla/dotenv/issues/664).
 - I have received redacted copies of .env files from third-party relayers where they have both NODE_URL_X and NODE_URLS_X defined. This indicates that they're unclear about how these should be used, so they define both.
 - See below for a screenshot from my own .env. In order to switch providers, I have to comment or uncomment out multiple lines, and I need to ensure that _only one_  instance of NODE_URL[S] for each chain. It's easy to miss this and end up with multiple variants of the NODE_URL[S]_X uncommented (the last one probably wins). The new syntax means that it doesn't matter whether I have all these providers defined in my env - all that matters is the order they appear in the RPC_PROVIDERS[_X] env var.
 
2023-05-09-113124_831x484_scrot

Additionally, this new syntax opens up for follow-up changes like:
 - RPC_PROVIDER_INFURA_1_BLOCK_RANGE=99999
 - RPC_PROVIDER_QUICKNODE_1_BLOCK_RANGE=10000

At the moment, we specify the "max block lookback" (maximum block range for getLogs queries) per _chain_, rather than per provider. This means that users of QuickNode and Infura have to limit their max block range for queries to the lowest supported value, which is limited by QuickNode's 10k block range. If we move to provider-specific config, the limitations of each individual provider can be applied onto to provider instances where they are used.

src/utils/ProviderUtils.ts
```diff
@@ -574,22 +574,48 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
...
-    if (nodeUrls?.length === 0) {
-      throw new Error(`Provided ${retryConfigKey}, but parsing it as json did not result in an array of urls.`);
+  const resolveUrls = (): string[] => {
+    const providers = process.env[`RPC_PROVIDERS_${chainId}`] ?? process.env["RPC_PROVIDERS"];
+    if (providers === undefined) {
+      throw new Error(`No RPC providers defined for chainId ${chainId}`);
     }
+    const nodeUrls = providers.split(",").map((provider) => {
+      const envVar = `RPC_PROVIDER_${provider}_${chainId}`;
+      const url = process.env[envVar];
```
james-a-morris, line 584: QQ: Is the `providers` just a list of provider keys like `alchemy, infura,...`? If so, should we use `superstruct` or something similar to validate the providers array to ensure that the user did not supply an invalid provider name?

src/utils/ProviderUtils.ts
```diff
@@ -574,22 +574,48 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
...
-  const retryConfigKey = `NODE_URLS_${chainId}`;
-  if (process.env[retryConfigKey]) {
-    const nodeUrls = JSON.parse(process.env[retryConfigKey]) || [];
-    if (nodeUrls?.length === 0) {
-      throw new Error(`Provided ${retryConfigKey}, but parsing it as json did not result in an array of urls.`);
+  const resolveUrls = (): string[] => {
+    const providers = process.env[`RPC_PROVIDERS_${chainId}`] ?? process.env["RPC_PROVIDERS"];
+    if (providers === undefined) {
+      throw new Error(`No RPC providers defined for chainId ${chainId}`);
     }
```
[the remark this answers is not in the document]
pxrl: That's the intention, yeah.

In the majority of cases, I expect users to run with the same provider priorities across all chains, so it's unnecessary to enumerate the same config for each individual chain.

Exceptions would be in cases like Boba (before removal), where the set of available RPC providers was significantly different. This situation may arise again with new chains as they are added - especially if they are somewhat exotic.

src/utils/ProviderUtils.ts
pxrl: The env var that seeds `providers` should take the form `RPC_PROVIDERS=<PROVIDER-1>,<PROVIDER-2>,<PROVIDER-3>`, and this should result in `providers` being an array of `[<PROVIDER-1>, <PROVIDER-2>, <PROVIDER-3>]`. The names of each provider is actually arbitrary, the important thing is that it's possible to resolve a defined RPC endpoint via the subsequent `RPC_PROVIDER_<PROVIDER>_<CHAIN_ID>` env var.

If the user supplies an invalid name, it will fail at line 585, and the user will get an error explaining which env var was not found. I _think_ this covers the scenario that you're asking about, so I'm not sure that superstruct helps in this instance. In general I'd hope that the validation here is simple enough that we don't need to use external tooling, unless it really offers some nice benefit.

mrice32 (commented):
Will this be backwards compatible?

pxrl:
> Will this be backwards compatible?

The new config syntax isn't backwards compatible, but the implementation will still try to use the existing syntax if it's unable to use the new format. This is transparent, so it doesn't result in an error for the user.

If we adopt the change, I'd propose to have an overlap period of at least a couple of weeks to allow migration of all configs.

nicholaspai:
> UMAprotocol/bot-configs#1193

That sounds

src/utils/ProviderUtils.ts
nicholaspai: I think super struct might be overkill because we might actually want to use shortened, non intuitive names. Instead of `infura` name it `1` etc.

src/utils/ProviderUtils.ts
```diff
@@ -574,22 +574,48 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
...
+    const nodeUrls = providers.split(",").map((provider) => {
+      const envVar = `RPC_PROVIDER_${provider}_${chainId}`;
+      const url = process.env[envVar];
+      if (url === undefined) {
+        throw new Error(`Missing RPC provider URL (${envVar})`);
```
nicholaspai, line 586:
```suggestion
        throw new Error(`Missing RPC provider URL for chain ${chainId} (${envVar})`);
```

Something like this to make it more clear?

src/utils/ProviderUtils.ts
```diff
@@ -574,22 +574,48 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
...
+      }
+      return url;
+    });
+
+    if (nodeUrls.length === 0) {
```
nicholaspai, line 591: Should we be comparing the `nodeUrls` to the quorum length to make sure nodeUrls.length >= quorum?

nicholaspai (approved):
@pxrl accidentally closed this by accident. 

After seeing the example config I can see how this can be really powerful. It does make switching the order of RPC's more intuitive (because of the names you can assign providers like `INFURA` and `ALCHEMY`), and also requires less code diff to actually switch the order of them. It also means that you can unilaterally switch the order of all providers.

mrice32 (approved):
@pxrl I like this as long as we retain backwards compatibility.

One concern that the UMA team has raised is that we have switched the env setup for node urls many times. We used to have CUSTOM_NODE_URL, then RETRY_CONFIG, then RETRY_CONFIG_[chainID], then NODE_URL_[chainId], then NODE_URLS_[chainId], and now this.

I think this is an improvement nonetheless, but a) would be great to document this in the README and the sample env and b) I think it would be nice to keep backwards compatibility for a while -- if this becomes a lot to manage, we could try to isolate this processing logic more than it is now by building a distinct class/module just for just grabbing the node url from the env :).

src/utils/ProviderUtils.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/656/commits/9565eae94cc7de91a117bb0aeb4b7b7834245c78

Tested OK on my end with both the new schema and the old. Seems to work OK with no identifiable regressions. I'll probably submit this on Monday.

pxrl:
> @pxrl I like this as long as we retain backwards compatibility.
> 
> One concern that the UMA team has raised is that we have switched the env setup for node urls many times. We used to have CUSTOM_NODE_URL, then RETRY_CONFIG, then RETRY_CONFIG_[chainID], then NODE_URL_[chainId], then NODE_URLS_[chainId], and now this.

> I think this is an improvement nonetheless, but a) would be great to document this in the README and the sample env and b) I think it would be nice to keep backwards compatibility for a while -- if this becomes a lot to manage, we could try to isolate this processing logic more than it is now by building a distinct class/module just for just grabbing the node url from the env :).

Totally on board with all of the above :+1:

PR #662, opened 2023-05-08
chore: defer client definitions from SDK clients

james-a-morris:


src/clients/AcrossAPIClient.ts
```diff
@@ -1,6 +1,6 @@
 import { winston, BigNumber, getL2TokenAddresses } from "../utils";
 import axios, { AxiosError } from "axios";
-import { HubPoolClient } from "./HubPoolClient";
+import { HubPoolClient } from ".";
```
pxrl, line 3: I recall tsc complaining at me for using `"."` instead of `"./"`.

```suggestion
import { HubPoolClient } from "./";
```

src/clients/HubPoolClient.ts
pxrl, line 1: I thought the intention was that we'd have a local variant of the HubPoolClient that inherited from the SDK-v2 HubPoolClient. Do we no longer have any local need for that?

src/common/ClientHelper.ts
```diff
@@ -241,7 +241,7 @@ export async function constructClients(
     // ProposeRootBundle in order to match a bundle block evaluation block range with a pending root bundle.
     maxBlockLookBack: config.maxBlockLookBack[config.hubPoolChainId],
   };
-  const hubPoolClient = new HubPoolClient(logger, hubPool, config.hubPoolChainId, hubPoolClientSearchSettings);
+  const hubPoolClient = new HubPoolClient(logger, hubPool, 0, config.hubPoolChainId, hubPoolClientSearchSettings);
```
pxrl, line 244: Here's the eventual change that we want :+1:

```suggestion
  const hubPoolClient = new HubPoolClient(
    logger,
    hubPool,
    Number(getDeploymentBlockNumber("HubPool", config.hubPoolChainId)),
    config.hubPoolChainId,
    hubPoolClientSearchSettings
  );
```

james-a-morris: Updated in: https://github.com/across-protocol/relayer-v2/pull/662/commits/b3e1538a564521abbff11801d9f3c9e4e3e74c17

src/clients/AcrossAPIClient.ts
james-a-morris: This is now removed

src/clients/SpokePoolClient.ts
```diff
@@ -1,44 +1,17 @@
...
-  RelayerRefundExecutionWithBlock,
-  RootBundleRelayWithBlock,
-  SpeedUp,
-  TokensBridged,
-  FundsDepositedEvent,
-} from "../interfaces";
-import { HubPoolClient } from ".";
+import { Deposit, DepositWithBlock, Fill, FundsDepositedEvent } from "../interfaces";
 
 export type SpokePoolUpdate = {
```
pxrl, line 16: This seems out of place now. It's used internally within the SpokePoolClient, so I kinda assume it should be in the SDK now. It's currently exported, but I think we could probably get away without that.

pxrl (commented):
Looks OK to me for the most part. Will hold off on final approval to allow others time to check it out.

nicholaspai (approved):
Very impressive you had to make no changes to the unit tests which should mean that the clients were migrated perfectly!

src/clients/SpokePoolClient.ts
james-a-morris: Good call! I've removed it. It's exposed in the SDK already.

pxrl (approved):
LGTM!

PR #675, opened 2023-05-12
refactor(UBA): Relocate getUBAFlows

pxrl:
This function becomes a method of the UBAClient class. This is relocated as an enabler for ACX-1076.


james-a-morris (approved):
This seems like a refactor - LGTM!

mrice32 (approved):
LGTM!

test/mocks/MockEvents.ts
pxrl, line 1: Not strictly related, but this is a hangover from previous refactoring over to the SDK.

PR #676, opened 2023-05-12
improve(UBA): Implement individual RefundRequest validation

pxrl:
This change implements basic validation for isolated RefundRequest events. There are certain attributes that can't be evaluated at this level; these are identified in the @dev comments for the new method.

Fixes ACX-1076.


src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+  /**
+   * @description Evaluate an RefundRequest object for validity.
+   * @dev  Callers should evaluate 'valid' before 'reason' in the return object.
+   * @dev  The following RefundRequest attributes are not evaluated for validity and should be checked separately:
+   * @dev  - previousIdenticalRequests
+   * @dev  - Age of blockNumber (i.e. according to SpokePool finality)
+   * @param chainId       ChainId of SpokePool where refundRequest originated.
+   * @param refundRequest RefundRequest object to be evaluated for validity.
+   */
+  refundRequestIsValid(chainId: number, refundRequest: RefundRequestWithBlock): { valid: boolean; reason?: string } {
```
pxrl, line 125: Per discussion, we might not want this to be directly in the UBAClient. The reason I've parked it here for now is because it spans the HubPoolClient and the set of SpokePoolClients, and I find it very annoying to always have to pass those objects into a validation function each time. So perhaps we can find another class for this to live in, but I'd like to avoid it living directly in the SpokePoolClient.

src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+    // Validate relayer and depositId.
+    const fill = destSpoke.getFillsForRelayer(relayer).find((fill) => {
+      // prettier-ignore
+      return (
+        fill.amount === amount
+        && fill.originChainId === originChainId
+        && fill.destinationChainId === destinationChainId
+        && fill.depositId === depositId
+        && fill.realizedLpFeePct === realizedLpFeePct
+        && fill.blockNumber === fillBlock.toNumber()
```
pxrl, line 152: It's a bit irritating that fillBlock is a BN. We usually normalise block numbers to JS numbers. Thoughts on applying the same for the RefundRequested type?

test/UBA.SpokePoolClient.ts
```diff
@@ -45,6 +102,7 @@ describe("UBA: SpokePool Events", async function () {
     ({ hubPool, weth, dai } = await hubPoolFixture());
     const deploymentBlock = await hubPool.provider.getBlockNumber();
     hubPoolClient = new MockHubPoolClient(logger, hubPool, deploymentBlock);
+    refundToken = originToken = weth.address;
```
james-a-morris, line 105: Nit: should we spread this over two statements? This is a nice succinct line of code, but it may not be completely intuitive as it relies on the return type of the assignment operator.

james-a-morris (commented):
Left some comments/nits

src/utils/UBA.ts
```diff
@@ -12,7 +12,8 @@ export class UBAClient {
   constructor(
     private readonly chainIds: number[],
     private readonly hubPoolClient: HubPoolClient,
-    private readonly spokePoolClients: { [chainId: number]: SpokePoolClient }
+    private readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    private readonly logger?: winston.Logger
```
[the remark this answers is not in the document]
pxrl: My gut feel is no, since this is a mid-layer component that probably shouldn't make any assumptions about its environment. It does mean that some log events are suppressed, but IMO it's really up to the caller to provide the logging object.

src/utils/UBA.ts
pxrl, line 125: Since we're looking up both the deposit and fill event here, does it make sense to include this in the returned object?

Otherwise we potentially end up with redundant looping by later having to make the necessary associations...

src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+    const fill = destSpoke.getFillsForRelayer(relayer).find((fill) => {
+      // prettier-ignore
+      return (
+        fill.amount === amount
+        && fill.originChainId === originChainId
+        && fill.destinationChainId === destinationChainId
+        && fill.depositId === depositId
+        && fill.realizedLpFeePct === realizedLpFeePct
+        && fill.blockNumber === fillBlock.toNumber()
+      );
```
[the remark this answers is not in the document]
pxrl: Oops...well spotted. BN does seem to allow the caller to coax an === comparison, but we might overflow the resulting number. So it needs to be a proper BN .eq comparison.

src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+    }
+    const destSpoke = this.spokePoolClients[destinationChainId];
+
+    if (fillBlock.lt(destSpoke.deploymentBlock) || fillBlock.gt(destSpoke.latestBlockNumber)) {
+      return { valid: false, reason: `Invalid FillBlock (${fillBlock})` };
+    }
+
+    // Validate relayer and depositId.
+    const fill = destSpoke.getFillsForRelayer(relayer).find((fill) => {
+      // prettier-ignore
```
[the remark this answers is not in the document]
pxrl: You mean the prettier-ignore? I prefer prepending the boolean operator, but I'm not married to it. I'd propose for prettier to ignore it centrally, but prettier apparently doesn't support this :(

src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+      return (
+        fill.amount === amount
```
mrice32, line 147: Is the fill validated _before_ being passed in here? For instance, what would happen if I created an invalid fill, but then created a _valid_ refund request _for that invalid fill_? Would the dataworker think that the refund request is valid and send a refund?

mrice32 (commented):
Looks great! Only a few minor questions/comments!

src/utils/UBA.ts
```diff
@@ -99,4 +112,65 @@ export class UBAClient {
...
+      return { valid: false, reason: "Invalid destinationChainId" };
+    }
+    const destSpoke = this.spokePoolClients[destinationChainId];
+
+    if (fillBlock.lt(destSpoke.deploymentBlock) || fillBlock.gt(destSpoke.latestBlockNumber)) {
+      return { valid: false, reason: `Invalid FillBlock (${fillBlock})` };
+    }
+
+    // Validate relayer and depositId.
+    const fill = destSpoke.getFillsForRelayer(relayer).find((fill) => {
```
pxrl, line 144: Instead of `find`, this should probably be a `filter`. There might be multiple valid fills from a single relayer for a single deposit.

src/utils/UBA.ts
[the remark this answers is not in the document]
pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/pull/676/commits/aa0229f525bf42c6da504c1822c17ea81f847c10

src/utils/UBA.ts
pxrl: The fill that is returned is subsequently validated against a Deposit, so that _shouldn't_ be possible.

I would prefer it if we accessed SpokePoolClient data via a wrapper that implemented all the necessary validation, so we could centralise that logic and ensure that nobody is accidentally bypassing it (as can be done today).

nicholaspai (approved):

PR #680, opened 2023-05-15
chore(package): Remove dependency on bluebird

pxrl:
I've isolated our use of bluebird.Promise to EventUtils, so have extracted it from all other files.

Can get rid of it from EventUtils with a little more refactoring.


nicholaspai (approved):

mrice32 (commented):
I'm a bit surprised, we don't need bluebird anywhere anymore? I thought we used it for throttling?

mrice32 (approved):
LGTM!

mrice32:
I think we had this idea that redis could be getting overloaded, and so rate limiting outside the provider might be helpful, but I suspect that's not the case!

PR #681, opened 2023-05-15
feat(Dataworker): Update exchange rates opportunistically in disputer path

nicholaspai:
The disputer already does all of the work of reconstructing pool rebalance leaves, which contain L1 tokens that could be executed in the next leaf. We should make sure these are always up to date and update them opportunistically before the bundle is executed.

This PR moves the `_updateExchangeRateCurrent` call out of the `proposer` path into the `disputer` path. This is because sometimes simulating the pool rebalance leaves fails when a token's reserves are out of `sync`. If the `disputer` handles this, then the tokens will always be in sync and the `bundleLpFees` will also be kept updated from the last bundle. This optimizes compounding effects for LPs

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -486,7 +486,7 @@ export class Dataworker {
     // Exit early if challenge period timestamp has passed:
     if (this.clients.hubPoolClient.currentTime > pendingRootBundle.challengePeriodEndTimestamp) {
       this.logger.debug({
-        at: "Dataworke#validate",
+        at: "Dataworker#validate",
         message: "Challenge period passed, cannot dispute",
         expirationTime: pendingRootBundle.challengePeriodEndTimestamp,
       });


scripts/runRelayerOnGoerli.sh
```diff
@@ -10,7 +10,7 @@ export RELAYER_TOKENS='["0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6"]'
 export MAX_RELAYER_DEPOSIT_LOOK_BACK=14400
 export NODE_DISABLE_PROVIDER_CACHING=true
 export RELAYER_GAS_MULTIPLIER=.01
-export RELAYER_MIN_FEE_PCT=.00001
+export MIN_RELAYER_FEE_PCT=.00001
```
nicholaspai, line 13: Unrelated fix in this script

mrice32 (commented):
Sorry, I'm a little confused about this change. Why would the disputer update the exchange rates? Is this a simulation thing? Or will they dispute all bundles that push us beyond what we can do?

nicholaspai:
> Sorry, I'm a little confused about this change. Why would the disputer update the exchange rates? Is this a simulation thing? Or will they dispute all bundles that push us beyond what we can do?

This PR arose from an issue where the `executor` failed to execute PoolRebalanceLeaves because an L1 token was not synced and therefore the reserves were not updated to pay out the `netSendAmount`.

In order to account for this, we'd need to add to the simulation logic some ability to simulate updating the reserves, or just ignore failed simulations of the PoolRebalanceLeaf execution. I think this is much more complex than just making sure we keep all L1 tokens synced that are included in a pending bundle.

The disputer will always make multiple runs and validate a pending bundle so it makes sense for the disputer to update the exchange rate, no?

pxrl:
> > Sorry, I'm a little confused about this change. Why would the disputer update the exchange rates? Is this a simulation thing? Or will they dispute all bundles that push us beyond what we can do?
> 
> This PR arose from an issue where the `executor` failed to execute PoolRebalanceLeaves because an L1 token was not synced and therefore the reserves were not updated to pay out the `netSendAmount`.
> 
> In order to account for this, we'd need to add to the simulation logic some ability to simulate updating the reserves, or just ignore failed simulations of the PoolRebalanceLeaf execution. I think this is much more complex than just making sure we keep all L1 tokens synced that are included in a pending bundle.
> 
> The disputer will always make multiple runs and validate a pending bundle so it makes sense for the disputer to update the exchange rate, no?

I can see the convenience of appending this functionality. It looks like a pretty simple change, and it seems guarded such that it can only occur if the disputer doesn't want to submit a dispute. Something does feel a little off about "overloading" the disputer in this way, though.

nicholaspai:
> > > > Sorry, I'm a little confused about this change. Why would the disputer update the exchange rates? Is this a simulation thing? Or will they dispute all bundles that push us beyond what we can do?
> > > 
> > > 
> > > This PR arose from an issue where the `executor` failed to execute PoolRebalanceLeaves because an L1 token was not synced and therefore the reserves were not updated to pay out the `netSendAmount`.
> > > In order to account for this, we'd need to add to the simulation logic some ability to simulate updating the reserves, or just ignore failed simulations of the PoolRebalanceLeaf execution. I think this is much more complex than just making sure we keep all L1 tokens synced that are included in a pending bundle.
> > > The disputer will always make multiple runs and validate a pending bundle so it makes sense for the disputer to update the exchange rate, no?
> > 
> > 
> > I can see the convenience of appending this functionality. It looks like a pretty simple change, and it seems guarded such that it can only occur if the disputer doesn't want to submit a dispute. Something does feel a little off about "overloading" the disputer in this way, though.
> 
> Fair points. Maybe I'll just add a separate function for a separate "exchange-rate-updater" bot

But my thinking is: we should only update exchange rates for L1 tokens that are included in pending bundles. We should only do so for valid bundles, ideally. 

So this validation of pending bundles is exactly what the disputer does. So isn't it exactly in the right place?

pxrl:
> > > I can see the convenience of appending this functionality. It looks like a pretty simple change, and it seems guarded such that it can only occur if the disputer doesn't want to submit a dispute. Something does feel a little off about "overloading" the disputer in this way, though.
> > 
> > 
> > Fair points. Maybe I'll just add a separate function for a separate "exchange-rate-updater" bot
> 
> But my thinking is: we should only update exchange rates for L1 tokens that are included in pending bundles. We should only do so for valid bundles, ideally.
> 
> So this validation of pending bundles is exactly what the disputer does. So isn't it exactly in the right place?

I do agree there are some practical justifications for including it. It feels to me as though the executor is probably the "spiritual" home for this functionality, though I concede that it's probably non-trivial to put it there.

One practical issue with the current PR: it forces every disputer instance to take on this role. That doesn't work very well in the case of third-parties who might be running a disputer (i.e. if they have their own relayer, or whatever).

src/dataworker/Dataworker.ts
```diff
@@ -534,6 +534,11 @@ export class Dataworker {
           this._submitDisputeWithMrkdwn(hubPoolChainId, reason);
         }
       }
+    } else {
+      // Once we've validated it, we should see if there are any L1 tokens included in the proposed bundle
+      // who's exchange rates need to be updated. This ensures that any future PoolRebalanceLeaf execution
+      // does not revert due to liquidReserves being out of sync.
+      await this._updateExchangeRates(expectedTrees.poolRebalanceTree.leaves, submitDisputes);
```
james-a-morris, line 541: Good refactor 🔥

src/dataworker/Dataworker.ts
```diff
@@ -486,7 +486,7 @@ export class Dataworker {
     // Exit early if challenge period timestamp has passed:
     if (this.clients.hubPoolClient.currentTime > pendingRootBundle.challengePeriodEndTimestamp) {
       this.logger.debug({
-        at: "Dataworke#validater",
+        at: "Dataworke#validate",
```
james-a-morris, line 489:
```suggestion
        at: "Dataworker#validate",
```

james-a-morris (commented):
Left a few nits

mrice32:
> > > > > I can see the convenience of appending this functionality. It looks like a pretty simple change, and it seems guarded such that it can only occur if the disputer doesn't want to submit a dispute. Something does feel a little off about "overloading" the disputer in this way, though.
> > > > 
> > > > 
> > > > Fair points. Maybe I'll just add a separate function for a separate "exchange-rate-updater" bot
> > > 
> > > 
> > > But my thinking is: we should only update exchange rates for L1 tokens that are included in pending bundles. We should only do so for valid bundles, ideally.
> > > So this validation of pending bundles is exactly what the disputer does. So isn't it exactly in the right place?
> > 
> > 
> > I do agree there are some practical justifications for including it. It feels to me as though the executor is probably the "spiritual" home for this functionality, though I concede that it's probably non-trivial to put it there.
> > One practical issue with the current PR: it forces every disputer instance to take on this role. That doesn't work very well in the case of third-parties who might be running a disputer (i.e. if they have their own relayer, or whatever).
> 
> So the updating logic already exists in the executor and this PR removes it. I think the executor doesn't work because it is in the one in charge of executing PoolRebalanceLeaves. The issue is that if one of those leaves wants to send tokens out of the HubPool but the HubPool doesn't currently have enough `liquidReserves` to service that outflow. The executor sees this and wants to call `exchangeRateCurrent` and update the `liquidReserves`, however the executor also enqueues transactions to execute those `PoolRebalanceLeaves`.
> 
> When we get to the `MulticallerClient`, it fails to simulate all of the leaves because the `BallanceAllocator` doesn't currently have logic to deal with the out-of-sync-reserves situation. So my thinking behind this PR is to avoid this situation by making sure that all L1 tokens are synced by the time we get to executing PoolRebalanceLeaves by adding the updating logic to the disputer, which only runs before the PoolRebalanceLeaves are executable.

I'm assuming it's technically challenging to have it just take the out-of-sync reserves? It does seem most natural for this all to happen atomically and have the multicaller client simulate the txns together to prove that when sequenced correctly that they work.

nicholaspai:
> have it just take the out-of-sync reserves

That seems like we'd have to add some logic to the Balance Allocator that I wanted to avoid. Currently we kind of hack it, so I thought it'd be very safe and strictly better to move it to the disputer. Do you see downsides to this temporary hack?

nicholaspai:
Going to replace this with https://github.com/across-protocol/relayer-v2/pull/686 which resolves the underlying the issue

PR #685, opened 2023-05-17
chore: Bump SDK for new UBA toys

pxrl:
Significant work has gone into the UBAClient post-migration to the SDK, and the SDK HubPoolClient now supports UBA-style runningBalances.

Fixes ACX-1220.


src/utils/UBA.ts
pxrl, line 1: Has previously been migrated to the SDK.

nicholaspai (approved):
LGTM

PR #686, opened 2023-05-17
improve(MulticallerClient): First try to simulate txns as a batch before individually simulating them

nicholaspai:
Improvement following up on https://github.com/across-protocol/relayer-v2/pull/637 that doesn't increase run-time by sending extra RPC requests in best case scenario where batch simulation succeeds.

Additionally, handles the case addressed by https://github.com/across-protocol/relayer-v2/pull/681 where simulations can fail individually but succeed in a batch

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -168,7 +168,7 @@ export class MultiCallerClient {
     const txnRequestsToSubmit: AugmentedTransaction[] = [];
 
     // First try to simulate the transaction as a batch. If the full batch succeeded, then we don't
-    // need to simulate transactions individually. If the batch failed, then we need to 
+    // need to simulate transactions individually. If the batch failed, then we need to
     // simulate the transactions individually and pick out the successful ones.
     const batchTxns: AugmentedTransaction[] = valueTxns.concat(
       await this.buildMultiCallBundles(txns, this.chunkSize[chainId])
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -165,36 +165,61 @@ export class MultiCallerClient {
...
+        batchesAllSucceeded = false;
+      }
+    });
+    if (batchesAllSucceeded) {
+      txnRequestsToSubmit.push(...batchTxns);
+    } else {
+      const individualTxnSimResults = await Promise.allSettled([
+        this.simulateTransactionQueue(txns),
+        this.simulateTransactionQueue(valueTxns),
+      ]);
```
james-a-morris, line 199: Will this transact all the simulations again?

james-a-morris (commented):
Left a few comments

src/clients/MultiCallerClient.ts
nicholaspai: yes it will, so in the worst case that the batch fails, we re sim the individual txns

src/clients/MultiCallerClient.ts
```diff
@@ -165,36 +165,61 @@ export class MultiCallerClient {
...
+          : `Failed to simulate ${networkName} transaction batch!`,
+        batchTxn: {
+          ...result.transaction,
+          contract: result.transaction.contract.address,
+        },
+      });
+      if (!result.succeed) {
+        batchesAllSucceeded = false;
+      }
+    });
```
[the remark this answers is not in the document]
nicholaspai: i.e. I don't want `.every()` to exit early.

PR #687, opened 2023-05-17
improve: Log deposit txn hashes for unprofitable deposits

nicholaspai:
In  practice we find it would be helpful to link directly to deposit transaction hashes in order to debug why a deposit was sent with unprofitable parameters

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -401,11 +401,11 @@ export class Relayer {
         const { symbol, decimals } = this.clients.hubPoolClient.getTokenInfoForDeposit(deposit);
         const formatFunction = createFormatFunction(2, 4, false, decimals);
         const gasFormatFunction = createFormatFunction(2, 10, false, 18);
+        const depositEtherscanLink = etherscanLink(deposit.transactionHash, deposit.originChainId);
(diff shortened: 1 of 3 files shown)


src/relayer/Relayer.ts
```diff
@@ -400,10 +402,13 @@ export class Relayer {
         const formatFunction = createFormatFunction(2, 4, false, decimals);
         const gasFormatFunction = createFormatFunction(2, 10, false, 18);
         depositMrkdwn +=
-          `- DepositId ${deposit.depositId} of amount ${formatFunction(deposit.amount)} ${symbol}` +
+          `- DepositId ${deposit.depositId} (tx: ${etherscanLink(
+            deposit.transactionHash,
+            deposit.originChainId
+          )}) of amount ${formatFunction(deposit.amount.toString())} ${symbol}` +
```
james-a-morris, line 408: Should we place the `etherscanLink` in its own variable to hoist it out of the formatting fn?

james-a-morris (approved):
LGTM - left a nit.

src/relayer/Relayer.ts
nicholaspai: yea its good hygiene you're right

PR #688, opened 2023-05-17
refactor(ConfigStoreClient): Remove getUtilization override

pxrl:
This method seems surprisingly unused by any other code in this repository!

It's proposed for removal because it introduces a dependency on the HubPoolClient, which seems like an inversion. Per a pending change to the HubPoolClient in the sdk-v2 repository, the HubPoolClient will instead become dependent on the ConfigStoreClient when validating the runningBalances member of the ExecutedRootBundle.

See https://github.com/across-protocol/sdk-v2/pull/213#discussion_r1197043597 for reference.


james-a-morris (approved):
I did a grep on the relayer and couldn't find any real usage of this function. I think this is gtg, but I'll wait pending signoff from Nick/Matt

pxrl:
Also to clarify, if we want to retain the caching logic, I think it's more logical for it to live in the HubPoolClient itself.

mrice32 (approved):
LGTM!

PR #692, opened 2023-05-19
refactor(SpokePoolClient): Simplify deposit lookup

pxrl:
Split queryHistoricalDepositForFill() into two, factoring out the on-chain paginated lookup from the caching logic.

The on-chain paginated lookup (SpokePoolClient.findDeposit()) can probably be relocated to the parent SpokePoolClient class in the sdk-v2 repository as a result.

This is an enabler for future work to shuffle the relationship between the SpokePoolClient, HubPoolClient and ConfigStoreClient, which is needed for the pending UBA rollout.


nicholaspai (approved):
> The on-chain paginated lookup (SpokePoolClient.findDeposit()) can probably be relocated to the parent SpokePoolClient class in the sdk-v2 repository

+1

PR #693, opened 2023-05-19
refactor(HubPoolClient): Update after SDK refactor

pxrl:
Per the recent change to refactor functionality between the ConfigStoreClient and HubPoolClient, corresponding changes are needed in relayer-v2.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -44,11 +44,11 @@ async function getSpokePoolSigners(
  */
 export async function constructSpokePoolClientsWithLookback(
   logger: winston.Logger,
+  hubPoolClient: HubPoolClient,
   configStoreClient: AcrossConfigStoreClient,
   config: CommonConfig,
   baseSigner: Wallet,
   initialLookBackOverride: number,
-  hubPoolChainId: number
 ): Promise<SpokePoolClientsByChain> {
   // Construct spoke pool clients for all chains that were enabled at least once in the block range.
   // Caller can optionally override the disabled chains list, which is useful for executing leaves or validating
@@ -56,10 +56,11 @@ export async function constructSpokePoolClientsWithLookback(
   // running the disputer or proposer functionality as it can lead to proposing disputable bundles or
   // disputing valid bundles.
 
-  if (!configStoreClient.isUpdated) {
+  if (!hubPoolClient.isUpdated) {
     throw new Error("Config store client must be updated before constructing spoke pool clients");
   }
 
+  const hubPoolChainId = hubPoolClient.chainId;
   const currentTime = getCurrentTime();
 
   // Use the first block that we'll query on mainnet to figure out which chains were enabled between then
@@ -93,7 +94,7 @@ export async function constructSpokePoolClientsWithLookback(
   // @dev: If toBlocks = {} then  construct spoke pool clients that query until the latest blocks.
   return await constructSpokePoolClientsWithStartBlocks(
     logger,
-    configStoreClient,
+    hubPoolClient,
     config,
     baseSigner,
     fromBlocks,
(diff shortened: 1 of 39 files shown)


pxrl:
Note: This fails because it requires an SDK bump first.

src/common/ClientHelper.ts
```diff
@@ -140,12 +141,13 @@ export async function constructSpokePoolClientsWithStartBlocks(
 ): Promise<SpokePoolClientsByChain> {
   if (!enabledChains) {
     enabledChains = getEnabledChainsInBlockRange(
-      configStoreClient,
+      hubPoolClient.configStoreClient,
       config.spokePoolChainsOverride,
-      startBlocks[1],
-      toBlockOverride[1]
+      startBlocks[hubPoolClient.chainId],
+      toBlockOverride[hubPoolClient.chainId]
```
pxrl, line 147: Improvement: Remove a constraint that would prevent the HubPoolClient from working on non-mainnet chains.

nicholaspai: I can't forsee a reason that we'd want the HubPool client to work on any chain but mainnet. Maybe in the future following a relaunch but I actually view pegging the chainId to mainnet as a safety feature

test/Dataworker.buildRoots.ts
```diff
@@ -69,7 +69,6 @@ describe("Dataworker: Build merkle roots", async function () {
 
     // Submit deposits for multiple destination chain IDs.
     const deposit1 = await buildDeposit(
-      configStoreClient,
```
nicholaspai, line 72: This is nice to see +1

nicholaspai (commented):
I like seeing that the tests generally are removing code and we're removing the `configStoreClient` from a bunch of function calls!

I think you're missing a package.json update to the latest sdk though

src/common/ClientHelper.ts
pxrl: > I can't forsee a reason that we'd want the HubPool client to work on any chain but mainnet. Maybe in the future following a relaunch but I actually view pegging the chainId to mainnet as a safety feature

If we have a hard-coupling to chainId 1 then we'll have a really difficult time testing the UBA on a testnet. I think that as long as the HubPoolClient uses its chainId correctly then we should be able to implement the necessary safeguards at the top-level relayer.ts and dataworker.ts files. wdyt?

I've proposed the rest of the necessary here: https://github.com/across-protocol/relayer-v2/pull/694

test/ConfigStoreClient.ts
```diff
@@ -49,20 +50,25 @@ const tokenConfigToUpdate = JSON.stringify({
...
+  });
+
   beforeEach(async function () {
     [owner] = await ethers.getSigners();
     ({ spokePool, erc20: l2Token } = await deploySpokePoolWithToken(originChainId, repaymentChainId));
     ({ hubPool, timer, dai: l1Token, weth } = await hubPoolFixture());
     await hubPool.enableL1TokenForLiquidityProvision(l1Token.address);
 
     configStore = await (await getContractFactory("AcrossConfigStore", owner)).deploy();
-    hubPoolClient = new HubPoolClient(createSpyLogger().spyLogger, hubPool);
```
pxrl, line 59: The ConfigStoreClient is no longer dependent on the HubPoolClient, so I could relegate it to the specific test where it's actually needed.

test/ConfigStoreClient.ts
```diff
@@ -245,30 +254,48 @@ describe("AcrossConfigStoreClient", async function () {
       ).to.throw(/Could not find TransferThreshold/);
     });
 
+    // @note: expect(...)to.deep.equals() coerces BigNumbers incorrectly and fails. Why?
```
pxrl, line 257: For reference, I identified that one of the BigNumbers, which was originally of the standard Ethers BigNumber format (`{ type: "BigNumber", hex: "0x00" }`)..., was being coerced into `{ value: "0" }`.

mrice32 (approved):
LGTM!

PR #696, opened 2023-05-22
refactor(fixture): Remove redundant array mapping

pxrl:
The mapping does not modify the input chainId list, despite the comment above indicating that it does.

Note: See lines 15 and 16 where destinationChainId and originChainId are imported.


pxrl:
Will be proposed in another way.

PR #697, opened 2023-05-22
fix(fixture): Fix chainIds in Dataworker fixture

pxrl:
Driveby fix:
 - Use the correct chainIds after SpokePool deployment.
 - Correct minor typo (originChainid => originChainId).


james-a-morris (approved):
LGTM

PR #698, opened 2023-05-22
improve: Bail immediately on CI test fail

pxrl:
In cases where a CI test is failing, this forces it to be aborted immediately, rather than spending time completing all subsequent tests. It should save time, and the developer can test locally where they almost certainly have a faster machine.

This is actually a mocha option that hardhat passes straight through. See https://mochajs.org/api/mocha for a basic description.


james-a-morris (approved):
LGTM - I personally like this because I can see the error output faster via push notifications on GitHub. The downside is that this may lead to a sequence of fixes that resolve one error after the other whereas currently having all the failures be presented in the CI may provide additional context/patterns.

mrice32 (approved):

mrice32:
> LGTM - I personally like this because I can see the error output faster via push notifications on GitHub. The downside is that this may lead to a sequence of fixes that resolve one error after the other whereas currently having all the failures be presented in the CI may provide additional context/patterns.

I think the question is around workflow -- do y'all prefer just seeing the error and then running locally to get more context and a full report? Or do you prefer ci taking a bit longer, but getting a full report to avoid running all tests locally. Personally, I often use the latter to guide _which_ tests I need to run, but the former is good too if y'all prefer ci being snappier.

james-a-morris:
> > LGTM - I personally like this because I can see the error output faster via push notifications on GitHub. The downside is that this may lead to a sequence of fixes that resolve one error after the other whereas currently having all the failures be presented in the CI may provide additional context/patterns.
> 
> 
> 
> Agreed @james-a-morris, I think the question is around workflow -- do y'all prefer just seeing the error and then running locally to get more context and a full report? Or do you prefer ci taking a bit longer, but getting a full report to avoid running all tests locally? Personally, I often use the latter to guide _which_ tests I need to run, but the former is good too if y'all prefer ci being snappier.

Can we configure the CI where on-push we run with the `--bail` flag but the PR rule runs fully?

PR #700, opened 2023-05-23
fix: Ensure ordering of client updates

pxrl:
The HubPoolClient must now be updated _after_ the ConfigStoreClient.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -261,7 +261,8 @@ export async function constructClients(
 }
 
 export async function updateClients(clients: Clients): Promise<void> {
-  await Promise.all([clients.hubPoolClient.update(), clients.configStoreClient.update()]);
+  await clients.configStoreClient.update();
+  await clients.hubPoolClient.update();
 }
 
 export function spokePoolClientsToProviders(spokePoolClients: { [chainId: number]: SpokePoolClient }): {


mrice32 (approved):

src/common/ClientHelper.ts
```diff
@@ -261,7 +261,8 @@ export async function constructClients(
 }
 
 export async function updateClients(clients: Clients): Promise<void> {
-  await Promise.all([clients.hubPoolClient.update(), clients.configStoreClient.update()]);
```
james-a-morris, line 264: Will this impact performance now that we're no longer updating in parallel?

src/common/ClientHelper.ts
```diff
@@ -261,7 +261,8 @@ export async function constructClients(
...
+  await clients.configStoreClient.update();
```
james-a-morris, line 264: Should we add meaningful documentation to this function about how the order of updates matters?

james-a-morris (approved):
LGTM - left a few nits

src/common/ClientHelper.ts
pxrl: Updating in parallel seems like it was always incorrect, and I'm puzzled about why it never seemed to throw an error. What changed in the last commit was that the precedence was switched from HubPoolClient to ConfigStoreClient (i.e. previously we required that the HubPoolClient was updated before the ConfigStoreClient, and now we require the opposite).

In terms of performance, I think it's minimal. It'll be slow when running for the first time, and thereafter it should only require a couple of seconds.

PR #702, opened 2023-05-23
fix: Restore service to finalizer

pxrl:


pxrl:
This was failing for the same reason as the previous issue. The finalizer has its own local copy of an "update" function, and it wasn't fixed previously.

Have tested OK locally with this PR, so we should be good to go.

mrice32 (approved):
LGTM!

PR #704, opened 2023-05-23
fix: Sync yarn.lock

pxrl:
Somehow the recent dependabot PR to bump contracts-v2 to 2.2.3 did not include removing contracts-v2 2.2.2.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11,19 +11,6 @@
     "@uma/common" "^2.17.0"
     hardhat "^2.9.3"
 
-"@across-protocol/contracts-v2@2.2.2":
-  version "2.2.2"
-  resolved "https://registry.yarnpkg.com/@across-protocol/contracts-v2/-/contracts-v2-2.2.2.tgz#ff7684317e93678c88f440eca9e7d398bfbd485a"
-  integrity sha512-y4ouHyRafRv/+L3CgvYE4V4CFWgK9Qm7cOSVusc+l1XTlgdRNgh0JIA41fXE6pG7eu+V6GmayMSUO9tAbH9QDA==
-  dependencies:
-    "@defi-wonderland/smock" "^2.3.4"
-    "@eth-optimism/contracts" "^0.5.11"
-    "@openzeppelin/contracts" "4.8.3"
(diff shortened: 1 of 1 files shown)


james-a-morris (commented):
Was this a manual removal?

mrice32:
> Was this a manual removal?

Even if it's manual, the thing I would want to double check is that if you re-run yarn after making this change, that it doesn't make any changes to the lockfile

pxrl:
> > Was this a manual removal?
>
> Even if it's manual, the thing I would want to double check is that if you re-run yarn after making this change, that it doesn't make any changes to the lockfile. @pxrl did you verify this?

Nope - nothing manual (editing yarn.lock by hand is a no-no!) This is what yarn spits out when I run it against current HEAD of master.

james-a-morris (approved):

mrice32 (approved):

mrice32:
> > > Was this a manual removal?
> > 
> > 
> > Even if it's manual, the thing I would want to double check is that if you re-run yarn after making this change, that it doesn't make any changes to the lockfile. @pxrl did you verify this?
> 
> Nope - nothing manual (editing yarn.lock by hand is a no-no!) This is what yarn spits out when I run it against current HEAD of master.
> 
> So, I'm not sure what to conclude other than something went amiss in the preparation of this commit: [7716fb9](https://github.com/across-protocol/relayer-v2/commit/7716fb9707394debf91e911580f1071d78f5b1aa).
> 
> [update]: I figured it out. This PR from dependabot (#689) was open for a few days before it got merged. Inbetween, I made changes to yarn.lock in this pr: #693. My changes were later merged into the PR from dependabot, but `yarn` was not re-run in that PR before it was submitted.
> 
> So the lesson is that we should be careful merging changes into a PR from dependabot if there is any risk that `package.json` was touched. Dependabot's `recreate` feature should help to avoid this, though it's slower to update the PR. This also applies to normal PRs that are submitted by devs, but I think dependabot PRs have a relatively higher chance of these collisions since they're "ownerless" and sometimes hang open for a relatively long time.

I see, no big deal, this sort of thing happens! LGTM!

PR #706, opened 2023-05-26
feat: Support basic message relay rules

pxrl:
This will permit guarded testing of the messaging feature. Example config:

	# Identify a list of rules to parse.
	RELAYER_MESSAGE_RULES=<NAME_1>,...

	# Config for individual rule. <NAME_1> is a placeholder for a descriptive name.
	RELAYER_MESSAGE_RULE_<NAME_1>_ORIGIN_CHAIN_IDS=10,137
	RELAYER_MESSAGE_RULE_<NAME_1>_DESTINATION_CHAIN_IDS=10,137
	RELAYER_MESSAGE_RULE_<NAME_1>_DEPOSITOR=0x012...def
	RELAYER_MESSAGE_RULE_<NAME_1>_RECIPIENTS=0x012...def,0xabc...012
	RELAYER_MESSAGE_RULE_<NAME_1>_TOKEN_SYMBOLS=USDC,WETH
	RELAYER_MESSAGE_RULE_<NAME_1>_TOKEN_AMOUNTS=50,0.005

Note: This PR is not properly tested yet, but is indiciative of the general concept that I propose.

Fixes ACX-1165


src/relayer/Relayer.ts
```diff
@@ -26,7 +28,53 @@ export class Relayer {
...
+    }
+
+    // The recipient might have changed via a speed-up event.
+    const recipient = isDepositSpedUp(deposit) ? deposit.updatedRecipient : deposit.recipient;
+    const rule = rules.find((rule) => {
+      return (
+        deposit.depositor === rule.depositor &&
+        rule.originChainIds.includes(deposit.originChainId) &&
+        rule.destinationChainIds.includes(deposit.destinationChainId) &&
+        rule.recipients.includes(recipient) &&
```
pxrl, line 53: Might benefit from some case-normalisation in these comparisons.

pxrl:
Adding the do not merge label until tests are added to verify the filtering.

src/relayer/RelayerConfig.ts
```diff
@@ -1,8 +1,18 @@
...
 
+export type MessageRelayRule = {
+  name: string;
+  originChainIds: number[];
+  depositor: string;
+  destinationChainIds: number[];
+  recipients: string[];
+  tokenSymbols: string[];
+  tokenAmounts: BigNumber[];
+};
```
mrice32, line 14: Just OOC, why do we need anything other than a recipient address + destination chain id pair (per rule)?

Just thinking about what we're worried about: we're mainly concerned about people abusing this to use up a lot of gas, right, but as long as we know the recipient contract cannot use a ton of gas, I would assume this is safe to enable without worrying about the depositor. Do you have concerns beyond that for this limited beta?

mrice32 (commented):
Looks great! Just one question about the number of parameters.

src/relayer/Relayer.ts
```diff
@@ -26,7 +28,53 @@ export class Relayer {
     readonly logger: winston.Logger,
     readonly clients: RelayerClients,
     readonly config: RelayerConfig
-  ) {}
+  ) {
+    this.logger.debug({
+      at: "Relayer",
+      message: "Instantiated with message relay rules",
+      messageRules: config.messageRules,
```
james-a-morris, line 35: Nit: would this output log produce additional info we may not need? In general, I'm not super comfortable with printing out a large object.

src/relayer/Relayer.ts
```diff
@@ -26,7 +28,53 @@ export class Relayer {
...
+        rule.tokenSymbols.includes(l1Token.symbol)
+      );
+    });
+
+    if (!isDefined(rule)) {
+      return false;
+    }
+
+    const idx = rule.tokenSymbols.indexOf(l1Token.symbol);
+    assert(idx !== -1, `Unexpected MessageRelayRule format (${rule.name})`); // invariant
```
james-a-morris, line 63: Should we assert here? If I understand correctly, this would unwind the stack. Should we have retry logic to mitigate against transient issues?

src/relayer/Relayer.ts
```diff
@@ -26,7 +28,53 @@ export class Relayer {
...
-  ) {}
+  ) {
+    this.logger.debug({
+      at: "Relayer",
+      message: "Instantiated with message relay rules",
+      messageRules: config.messageRules,
+    });
+  }
+
+  // Temporary feature to allow controlled use of message transfers.
```
james-a-morris, line 39: The comments below are related to this temporary logic - feel free to ignore them if the code is not going to be around for long.

src/relayer/Relayer.ts
```diff
@@ -189,14 +237,18 @@ export class Relayer {
       // We need to build in better simulation logic for deposits with non-empty messages. Currently we only measure
       // the fill's gas cost against a simple USDC fill with message=0x. This doesn't handle the case where the
       // message is != 0x and it ends up costing a lot of gas to execute, resulting in a big loss to the relayer.
-      if (isDepositSpedUp(deposit) && deposit.updatedMessage !== "0x") {
+      if (
+        isDepositSpedUp(deposit) &&
+        deposit.updatedMessage !== "0x" &&
```
james-a-morris, line 242: What if the `updatedMessage` is empty or undefined? Is this still considered a different case?

james-a-morris (commented):
Left a few comments

src/relayer/Relayer.ts
pxrl: Yeah; it can be a bit spammy because it's not guarded by whether `config.messageRules` is populated. It should be guarded, but if message rules are defined then I think it makes sense to dump them, since it's not expected to be arbitrarily long.

src/relayer/Relayer.ts
pxrl: It's an invariant, so if idx evaluates to -1 then something is wrong with the config parsing. I think it's preferable to just fail hard and fast in that case.

src/relayer/Relayer.ts
pxrl: We're only interested in Deposits where we'd actually be executing a message via a callback in the fill. If the message has been removed as part of the speed-up then we just treat it like a normal Deposit and fill it accordingly.

src/relayer/Relayer.ts
pxrl: Have guarded it here: https://github.com/across-protocol/relayer-v2/pull/706/commits/b92892c1f59438e1bcda90800bbb9d8b3a88adfd.

pxrl:
This has tested OK on my end and it has been used to relay messages between Optimism and Polygon. I think this is OK to go in, pending a decision about whether to ease off on the rule restrictions.

src/relayer/RelayerConfig.ts
mrice32: Makes sense!

mrice32 (commented):
Should we drop this PR in favor of a full productionization of messaging?

pxrl:
> Should we drop this PR in favor of a full productionization of messaging?

Sure, I'll park it for now. I was thinking that it would be useful for third-parties to use to self-relay messages - i.e. a project might want to run a relayer to handle their own messages, but not any other fills (with or without messages).

PR #708, opened 2023-06-02
chore: Cleanup fetchTokenInfo after SDK-v2 migration

pxrl:
fetchTokenInfo() was copied to the SDK a while ago, but it persisted in relayer-v2 until now.

Note: There are currently no users of fetchTokenInfo() in relayer-v2, so we could alternatively remove it.


mrice32 (approved):
LGTM!

james-a-morris (approved):
Good refactor

PR #709, opened 2023-06-03
fix: fix field comparison error

mrice32:
This handles an issue where one of the rpc response passed into the compareRpcResults method is undefined or null and can't be accessed as an object. In this case, the method that checks for ignored fields first checks if the value is present before attempting to modify it.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -89,6 +89,9 @@ function compareRpcResults(method: string, rpcResultA: any, rpcResultB: any): bo
   // back after we do the comparison with unignored keys. This is a faster algorithm than cloning an object but it has
   // some side effects such as the order of keys in the rpcResults object changing.
   const deleteIgnoredKeys = (ignoredKeys: string[], rpcResults: any) => {
+    if (!rpcResults) {
+      return;
+    }
     const ignoredMappings = {};
     for (const key of ignoredKeys) {
       ignoredMappings[key] = rpcResults[key];


src/utils/ProviderUtils.ts
```diff
@@ -89,6 +89,9 @@ function compareRpcResults(method: string, rpcResultA: any, rpcResultB: any): bo
   // back after we do the comparison with unignored keys. This is a faster algorithm than cloning an object but it has
   // some side effects such as the order of keys in the rpcResults object changing.
   const deleteIgnoredKeys = (ignoredKeys: string[], rpcResults: any) => {
+    if (!rpcResults) {
+      return;
+    }
```
james-a-morris, line 94: It seems that we're using `rpcResults` to represent an object. I think it would be a better typing to use something like `rpcResults: Record<string, any>` since that's more closely set to what we're looking for. 

Additionally, if we do this, we could be more specific with like 92 - for example, we could use either `rpcResults === undefined` or `Object.keys(obj).length > 0`

james-a-morris (approved):
LGTM preemptively

src/utils/ProviderUtils.ts
mrice32: This is a little tricky. The return types for these RPC calls (from ethers) are any. I think they can be primitives, like numbers or strings. The only types I'm aware of that will generate an error on being accessed like value["someKey"] is null or undefined.

I don't love the way it's represented now -- I wish the types from ethers were more constrained or specific, but I think Record might be too narrow, as that could imply that we could do things to the value that you can't do on a primitive.

Thoughts?

pxrl (approved):
LGTM. I sympathise with the frustration of having to deal with wide-open types :(

src/utils/ProviderUtils.ts
james-a-morris: If we receive a primitive, wouldn't the logic below fail? 

```
    const ignoredMappings = {};
    for (const key of ignoredKeys) {
      ignoredMappings[key] = rpcResults[key];
      delete rpcResults[key];
    }
```

PR #710, opened 2023-06-03
chore: Bump contracts-v2

pxrl:
Since contracts-v2 no longer exports hre, import it directly in the files that need it.


james-a-morris (approved):
Good workaround!

mrice32 (approved):
LGTM!

PR #711, opened 2023-06-04
fix: Permit proposals when there is no outstanding proposal

pxrl:


pxrl:
Have worked around this by running the proposer locally with this change. It should break us out of the indefinite wait, but we need to follow up on the underlying issue.

PR #712, opened 2023-06-04
fix(dataworker): Wait to propose for some time after latest valid proposal, not just any proposal

nicholaspai:
Fixes the case where previous bundle was disputed and we need to re-propose


james-a-morris (approved):
Good change.

PR #715, opened 2023-06-05
improve: Handle case where chain has stopped advancing

nicholaspai:
Adds one additional check to function that returns next bundle block range to propose to handle case where the `latest` block has stopped advancing.

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -425,34 +425,43 @@ export function getWidestPossibleExpectedBlockRange(
   latestMainnetBlock: number,
   enabledChains: number[]
 ): number[][] {
+  // We subtract a buffer from the end blocks to reduce the chance that network providers
+  // for different bot runs produce different contract state because of variability near the HEAD of the network.
+  // Reducing the latest block that we query also gives partially filled deposits slightly more buffer for relayers
+  // to fully fill the deposit and reduces the chance that the data worker includes a slow fill payment that gets
+  // filled during the challenge period.
+  const latestPossibleBundleEndBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
+    (chainId: number, index) =>
+      spokeClients[chainId] && Math.max(spokeClients[chainId].latestBlockNumber - endBlockBuffers[index], 0)
+  );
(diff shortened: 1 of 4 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -425,34 +425,45 @@ export function getWidestPossibleExpectedBlockRange(
   latestMainnetBlock: number,
   enabledChains: number[]
 ): number[][] {
+  // We subtract a buffer from the end blocks to reduce the chance that network providers
+  // for different bot runs produce different contract state because of variability near the HEAD of the network.
+  // Reducing the latest block that we query also gives partially filled deposits slightly more buffer for relayers
+  // to fully fill the deposit and reduces the chance that the data worker includes a slow fill payment that gets
+  // filled during the challenge period.
+  const latestPossibleBundleEndBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
```
mrice32, line 433: Just want to double check: this function isn't used to determine if the blocks ranges are disputable, right? Presumably we wouldn't configurable buffers to determine the validity of a bundle, right?

mrice32 (approved):
Just so I understand: the only thing this changes is that there's an extra case where the chain is _not disabled_ but has not progressed since the last end block. In that case, it's treated as frozen, so 1 is not added to the start block.

Is that logic allowed according to the UMIP?

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: That's right. This function `blockRangesAreInvalidForSpokeClients` validates the block ranges and it includes a condition [here](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/DataworkerUtils.ts#L144) that handles the case where the range is 0 blocks.

nicholaspai:
> This is great!
> 
> Just so I understand: the only thing this changes is that there's an extra case where the chain is _not disabled_ but has not progressed since the last end block. In that case, it's treated as frozen, so 1 is not added to the start block.
> 
> Is that logic allowed according to the UMIP?

That's right, do you think this makes sense? And great point about the UMIP, added a corresponding PR here https://github.com/UMAprotocol/UMIPs/pull/586

PR #716, opened 2023-06-06
fix: Respect relayer dest chains in API instantiation

pxrl:


pxrl:
Dropped in favour of Nick's change: https://github.com/across-protocol/relayer-v2/pull/717

PR #717, opened 2023-06-06
fix(relayer): Don't construct API client with disabled destination chains

nicholaspai:
There is no need to call /limits on disabled destination chains


pxrl (approved):
Looks fine. I tested with something similar and it works OK for me.

PR #718, opened 2023-06-06
improve(TransactionUtils): Improve error logging for failed transactions

nicholaspai:


src/utils/TransactionUtils.ts
```diff
@@ -76,22 +76,32 @@ export async function runTransaction(
...
         message: "Retrying txn due to expected error",
         error: JSON.stringify(error),
         retriesRemaining,
       });
       return await runTransaction(logger, contract, method, args, value, gasLimit, null, retriesRemaining);
     } else {
+      // If error is "UNPREDICTABLE_GAS_LIMIT" then the transaction likely reverted on-chain when estimated by the
+      // RPC server. Ethers might populate an "error" object which could contain additional
+      // information about the error encountered when the RPC simulated the transaction.
+      const innerError = (error as any)?.error;
```
james-a-morris, line 89: Can we use `(error as Error)?.error` to use the generic error typing?

src/utils/TransactionUtils.ts
```diff
@@ -76,22 +76,32 @@ export async function runTransaction(
...
+
       // If transaction error reason is known to be benign, then reduce the log level to warn.
-      logger[txnRetryable(error) ? "warn" : "error"]({
-        at: "TxUtil",
+      logger[txnRetryable(error) || txnRetryable(innerError) ? "warn" : "error"]({
+        at: "TxUtil#runTransaction",
```
james-a-morris, line 93: Should we begin to list these `at` locations in an enum somewhere or a constants file? It may make sense to spend time to codify these specific location names so that re-use is consistent.

src/utils/TransactionUtils.ts
```diff
@@ -76,22 +76,32 @@ export async function runTransaction(
...
     } else {
+      // If error is "UNPREDICTABLE_GAS_LIMIT" then the transaction likely reverted on-chain when estimated by the
+      // RPC server. Ethers might populate an "error" object which could contain additional
+      // information about the error encountered when the RPC simulated the transaction.
+      const innerError = (error as any)?.error;
+
       // If transaction error reason is known to be benign, then reduce the log level to warn.
-      logger[txnRetryable(error) ? "warn" : "error"]({
-        at: "TxUtil",
+      logger[txnRetryable(error) || txnRetryable(innerError) ? "warn" : "error"]({
```
james-a-morris, line 92: Could we make `txnRetryable` take a list of `Error` objects so that we could do something like `txnsRetryable(error, innerError)`

james-a-morris (commented):
Left some feedback

src/utils/TransactionUtils.ts
nicholaspai: it doesn't work though because `Error` doesn't have an `error` prop

src/utils/TransactionUtils.ts
nicholaspai: We should probably but in a separate PR!

src/utils/TransactionUtils.ts
nicholaspai: Do you think we would use this often?

james-a-morris: It's possible it wouldn't be used as often. I think this is a good point to leave

james-a-morris (approved):
LGTM

src/utils/TransactionUtils.ts
nicholaspai: @pxrl sorry are you suggesting a change to the code as is? It would be awesome if we can re-use code from the sdk-v2 to avoid this `any` typecast!

src/utils/TransactionUtils.ts
```diff
@@ -76,22 +76,33 @@ export async function runTransaction(
...
+        at: "TxUtil#runTransaction",
         message: "Error executing tx",
         retriesRemaining,
-        error: JSON.stringify(error),
+        errorReason: innerError?.reason ?? (error as any)?.reason,
```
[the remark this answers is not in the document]
nicholaspai: good call.. will do

nicholaspai:
@james-a-morris @pxrl I've taken in your feedback and changed the PR a bit. Its a bit hard to test but I think this could work? 720f685c8ec643bd0522f2c2555f9e41db9bdab9

src/interfaces/Error.ts
```diff
@@ -1,4 +1,5 @@
 export type EthersError = Error & {
   code: string;
   reason: string;
+  error?: EthersError;
```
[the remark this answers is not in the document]
nicholaspai: So should we delete this file and import it from the SDK? We first need to export a new SDK version then

src/utils/TransactionUtils.ts
```diff
@@ -79,23 +79,54 @@ export async function runTransaction(
...
+        let topError: EthersError = error as EthersError;
+        while (topError !== undefined) {
+          ethersErrors.push({ reason: topError.reason, err: topError.error });
+          topError = topError.error;
+        }
+        logger[ethersErrors.some((e) => txnRetryable(e.err)) ? "warn" : "error"]({
+          at: "TxUtil#runTransaction",
+          message: "Error executing tx",
+          retriesRemaining,
+          errorReasons: ethersErrors.map((e, i) => `\n\t ${i}: ${e.reason}`),
```
james-a-morris, line 106: Should we join these errors?

```suggestion
          errorReasons: ethersErrors.map((e, i) => `\t ${i}: ${e.reason}`).join('\n'),
```

james-a-morris (approved):
LGTM

src/interfaces/Error.ts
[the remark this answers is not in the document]
pxrl: FWIW we should be able to update interfaces/Error.ts now because we have bumped the SDK since we last discussed this.

src/utils/TransactionUtils.ts
```diff
@@ -79,23 +79,54 @@ export async function runTransaction(
...
-        retriesRemaining,
-        error: JSON.stringify(error),
-        notificationPath: "across-error",
-      });
-      throw error;
+      // Empirically we have observed that Ethers can produce nested errors, so we try to recurse down them
+      // and log them as clearly as possible. For example:
+      // - Top-level (Contract method call): "reason":"cannot estimate gas; transaction may fail or may require manual gas limit" (UNPREDICTABLE_GAS_LIMIT)
+      // - Mid-level (eth_estimateGas): "reason":"execution reverted: delegatecall failed" (UNPREDICTABLE_GAS_LIMIT)
+      // - Bottom-level (JSON-RPC/HTTP): "reason":"processing response error" (SERVER_ERROR)
```
pxrl, line 94: How about factoring out the common fields from the log object so that it can be unpacked in each call to logger , with only the differences being supplied? It should save on a lot of vertical space here.
 
```suggestion
      // - Bottom-level (JSON-RPC/HTTP): "reason":"processing response error" (SERVER_ERROR)
      const commonFields = {
        at: "TxUtil#runTransaction",
        message: "Error executing tx",
        retriesRemaining,
        target: getTarget(contract.address),
        method,
        args,
        value,
        nonce,
      };  
```

src/utils/TransactionUtils.ts
```diff
@@ -79,23 +79,54 @@ export async function runTransaction(
...
+        let topError: EthersError = error as EthersError;
+        while (topError !== undefined) {
+          ethersErrors.push({ reason: topError.reason, err: topError.error });
+          topError = topError.error;
+        }
+        logger[ethersErrors.some((e) => txnRetryable(e.err)) ? "warn" : "error"]({
+          at: "TxUtil#runTransaction",
+          message: "Error executing tx",
+          retriesRemaining,
+          errorReasons: ethersErrors.map((e, i) => `\t ${i}: ${e.reason}`).join('\n'),
```
pxrl, line 106: Github seems to object to the use of single quotes...

```suggestion
          errorReasons: ethersErrors.map((e, i) => `\t ${i}: ${e.reason}`).join("\n"),
```

src/utils/TransactionUtils.ts
```diff
@@ -79,23 +79,54 @@ export async function runTransaction(
...
+          retriesRemaining,
+          error: JSON.stringify(error),
+          target: getTarget(contract.address),
+          method,
+          args,
+          value,
+          nonce,
+          notificationPath: "across-error",
+        });
+        throw error;
```
pxrl, line 128: Error seems to be thrown in either case - can it be moved out of the `if ... else` statement?

src/utils/TransactionUtils.ts
nicholaspai: great call

PR #723, opened 2023-06-07
improve: Add default Optimism reorg values

nicholaspai:


james-a-morris (approved):
Good call on increasing the reorg / look back distance

pxrl (approved):
LGTM

PR #726, opened 2023-06-08
chore(monitor): Remove CONFIGURED_NETWORKS

pxrl:
The MonitorConfig hasn't read this from the environment for quite a while.


james-a-morris (approved):
This seems like a reasonable change if we no longer need it

PR #743, opened 2023-06-16
fix: Bump SDK for better Arbitrum gas pricing

pxrl:
The SDK recently migrated to using EIP1559 pricing for Arbitrum, but this is misleading because the priority fee is always refunded (and Ethers hardcodes it to 1.5 Gwei). This caused deposits destined for Arbitrum to be considered unprofitable.


james-a-morris (approved):
LGTM

PR #744, opened 2023-06-17
improve(runTransaction): Pad gas for multicall3 aggregate

pxrl:
This change implements three key things:
        
1) Capture the gas estimates from successful txn simulations and pass it
through to txn submission.
        
2) Modify TransactionClient() to check for a gasLimitMultiplier argument
If a multiplier was supplied, apply it to the gasLimit at that point,
before passing a transaction request over to runTransaction() for 
submission to the RPC provider.
        
3) Modify the MultiCallerClient class to apply 20% gas padding on
multicall3 aggregate() transactions. This addresses a specific issue
that has been seen with the executor personality of the dataworker
recently, resulting in a significant number of OoG failures.
         
Side benefit: This might actually save a separate estimateGas() RPC 
call, so it's possibly faster and a bit cheaper in terms of RPC usage.

Fixes ACX-1236.


nicholaspai (approved):

src/clients/MultiCallerClient.ts
```diff
@@ -380,9 +386,7 @@ export class MultiCallerClient {
         chunkSize
       );
       const multicallerTxnBundle = multicallerTxnChunks
-        .map((txnChunk) => {
-          return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0];
-        })
+        .map((txnChunk) => (txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0]))
```
pxrl, line 389: This is just a minor driveby refactor, since it can fit on a single line.

src/utils/TransactionUtils.ts
```diff
@@ -12,7 +12,8 @@ dotenv.config();
 export type TransactionSimulationResult = {
   transaction: AugmentedTransaction;
   succeed: boolean;
-  reason: string;
+  gasLimit?: BigNumber;
+  reason?: string;
```
pxrl, line 16: Driveby change: made `reason` optional so we don't have to return it on simulation successes.

src/clients/MultiCallerClient.ts
```diff
@@ -269,6 +272,7 @@ export class MultiCallerClient {
       contract: multisender,
       method: "aggregate",
       args: [callData],
+      gasLimitMultiplier: 1.2,
```
nicholaspai, line 275: Nit can we make this 1.5 or even 2 to be safe?

nicholaspai (approved):
This is a good patch lgtm

src/clients/MultiCallerClient.ts
```diff
@@ -188,8 +188,11 @@ export class MultiCallerClient {
       });
       if (!result.succeed) {
         batchesAllSucceeded = false;
+      } else {
+        batchTxns[idx].gasLimit = result.transaction.gasLimit;
```
james-a-morris, line 192: Should the `batchesAllSucceeded` break the loop?

src/clients/MultiCallerClient.ts
```diff
@@ -269,6 +272,7 @@ export class MultiCallerClient {
       contract: multisender,
       method: "aggregate",
       args: [callData],
+      gasLimitMultiplier: 2, // @note: Lots of headroom for the executor.
```
james-a-morris, line 275: Should we proactively place this in a constant

src/clients/TransactionClient.ts
```diff
@@ -24,6 +26,8 @@ export interface AugmentedTransaction {
   canFailInSimulation?: boolean;
 }
 
+const GAS_MULTIPLIER_1X = 1.0;
```
james-a-morris, line 29: Should we place this in a constants file?

src/clients/TransactionClient.ts
```diff
@@ -63,6 +67,16 @@ export class TransactionClient {
         this.logger.debug({ at: "TransactionClient#submit", message: `Using nonce ${nonce}.` });
       }
 
+      const gasLimitMultiplier = txn.gasLimitMultiplier ?? GAS_MULTIPLIER_1X;
+      if (gasLimitMultiplier != GAS_MULTIPLIER_1X) {
```
james-a-morris, line 71: Since `gastLimitMultiplier` is potentially a non-integer (neither is `GAS_MULTIPLIER_1X`), should we perform fuzzy comparisons?

I.e. `Math.abs(gasLimitMultiplier - GAS_MULTIPLIER_1X) < .001`

james-a-morris (commented):
Left a few nits

src/clients/MultiCallerClient.ts
pxrl: Addressed here: https://github.com/across-protocol/relayer-v2/pull/744/commits/b30cc5644a4ba131068af05cde6178caaa89ebdb

src/clients/TransactionClient.ts
pxrl: Fair call - have implemented something similar here: https://github.com/across-protocol/relayer-v2/pull/744/commits/6c42b6040d9c9d635a58375250ffaaf434719641

I don't think there's really a case for ever setting a multiplier < DEFAULT_GAS_MULTIPLIER, so I've just ignored that.

james-a-morris: Fair enough!

james-a-morris (approved):
My comments have been addressed. Thanks!

PR #748, opened 2023-06-20
refactor(Relayer): Factor out repaymentChainId determination

pxrl:
No functional change - this is mostly just re-arranging the furniture prior to introducing UBA fee determination.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -242,40 +242,10 @@ export class Relayer {
       return;
     }
 
+    const repaymentChainId = await this.resolveRepaymentChain(deposit, fillAmount);
+    this.logger.debug({ at: "Relayer", message: "Filling deposit", deposit, repaymentChainId });
+
     try {
-      // TODO: Consider adding some way for Relayer to delete transactions in Queue for fills for same deposit.
-      // This way the relayer could set a repayment chain ID for any fill that follows a 1 wei fill in the queue.
-      // This isn't implemented due to complexity because its a very rare case in production, because its very
-      // unlikely that a relayer could enqueue a 1 wei fill (lacking balance to fully fill it) for a deposit and
-      // then later on in the run have enough balance to fully fill it.
-      const fillsInQueueForSameDeposit =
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -347,6 +317,40 @@ export class Relayer {
...
+    // then later on in the run have enough balance to fully fill it.
+    const fillsInQueueForSameDeposit =
+      this.clients.multiCallerClient.getQueuedTransactions(deposit.destinationChainId).some((tx) => {
+        const { method, args } = tx;
+        const { depositId, originChainId } = deposit;
+        return (
+          (method === "fillRelay" && args[9] === depositId && args[6] === originChainId) ||
+          (method === "fillRelayWithUpdatedDeposit" && args[11] === depositId && args[7] === originChainId)
+        );
+      });
```
pxrl, line 334: Made a minor adjustment to the resolution of `fillsInQueueForSameDeposit`:
 - Use `Array.some()` instead of `Array.filter().length > 0`.
 - Dereference members of the tx and deposit objects in order to condense the code => fits on fewer lines.

Functionally this should behave identically as the pre-existing implementation.

pxrl: As additional background, it's hard to see how we can ever hit this scenario in production, so it'd be good to see whether we can remove it in future.

src/relayer/Relayer.ts
```diff
@@ -347,6 +317,40 @@ export class Relayer {
...
+
+    // Fetch the repayment chain from the inventory client. Sanity check that it is one of the known chainIds.
+    // We can only overwrite repayment chain ID if we can fully fill the deposit.
+    let repaymentChainId = deposit.destinationChainId;
+
+    if (fillAmount.eq(deposit.amount) && !fillsInQueueForSameDeposit) {
+      const destinationChainId = deposit.destinationChainId.toString();
```
pxrl, line 342: Minor tweak here - resolve the destinationChainId to a string once, rather than once per loop in the lower `Array.includes()` statement.

nicholaspai (approved):

PR #749, opened 2023-06-20
improve: Suppress additional "relay filled" errors

pxrl:
Sometimes the error is returned as `execution reverted: <reason>`, so add detection for that. Additionally, refactor the detection logic to be more concise (no intended change in behaviour).

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -424,7 +424,9 @@ export class MultiCallerClient {
     const knownReason = [...knownRevertReasons].some((knownReason) =>
       [knownReason, `execution reverted: ${knownReason}`].includes(reason)
     );
-    return knownReason || (unknownRevertReasonMethodsToIgnore.has(_txn.method) && reason.includes(unknownRevertReason));
+    return !txn.succeed && (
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -420,13 +420,11 @@ export class MultiCallerClient {
...
-    return (
-      !txn.succeed && (
-        knownRevertReasons.has(txn.reason) ||
-        (unknownRevertReasonMethodsToIgnore.has(txn.transaction.method) && txn.reason === unknownRevertReason)
-      )
+    const { transaction: _txn, reason } = txn;
+    const knownReason = [...knownRevertReasons].some((knownReason) =>
+      [knownReason, `execution reverted: ${knownReason}`].includes(reason)
     );
+    return knownReason || (unknownRevertReasonMethodsToIgnore.has(_txn.method) && reason.includes(unknownRevertReason));
```
nicholaspai, line 427:
```suggestion
    return !txn.succeed && (
        knownReason || (unknownRevertReasonMethodsToIgnore.has(_txn.method) && reason.includes(unknownRevertReason))
    );
```

PR #751, opened 2023-06-20
refactor(Relayer): Remove exception wrapper in fillRelay()

pxrl:
I think this large exception wrapper is a hangover from a much older implementation where something could fail at runtime (perhaps the repaymentChainId determination, which was recently relocated). In any case, there are no instances of the "Error creating fillRelayTx" message in the logs over the past 180 days, and it's difficult to see what could fail here without being an actual bug in the code.

Additionally, since there are now no async calls in fillRelay(), remove its async declaration.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -206,7 +206,7 @@ export class Relayer {
         const refundFee = toBN(0);
 
         if (profitClient.isFillProfitable(deposit, unfilledAmount, refundFee, l1Token)) {
-          await this.fillRelay(deposit, unfilledAmount, repaymentChainId);
+          this.fillRelay(deposit, unfilledAmount, repaymentChainId);
         } else {
           profitClient.captureUnprofitableFill(deposit, unfilledAmount);
         }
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -244,48 +244,39 @@ export class Relayer {
     }
 
     this.logger.debug({ at: "Relayer", message: "Filling deposit", deposit, repaymentChainId });
-    try {
```
james-a-morris, line 247: QQ: why are we removing the `try{ } catch(e) { }`

james-a-morris (commented):
Left a comment

james-a-morris (approved):

src/relayer/Relayer.ts
pxrl: The wrapper was added back in March 2022 at a time when the relayer had a lot less maturity than it does now. I'm pretty confident that whatever errors could be thrown back then have either been fixed or mitigated in other places. With the code that we have remaining, I think an error would actually be indicative of a bug, and we'd rather crash hard so it is detected and fixed, rather than suppress it and take the chance that it goes undetected.

PR #752, opened 2023-06-20
improve(ProfitClient): Permit gas costs to be ignored

pxrl:
Recent testing has been difficult because third-party relayers are eager to make fills, and are sometimes doing so at a loss. This means it can be surprisingly difficult to set a relayer fee low enough to be ignored by others. This change permits a 0x gas multiplier to be used, thereby instructing the relayer to ignore the cost of gas entirely. Deposits can thus be set with a 0% relayerFeePct, which is good enough to be ignored by most third parties. The pre-existing profitability protections are still in place, and control remains in the hands of the user as to how they want to configure their relayer.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -93,7 +93,7 @@ export class ProfitClient {
   ) {
     // Require 1% <= gasMultiplier <= 400%
     assert(
-      this.gasMultiplier.gte(toBNWei("0.01")) && this.gasMultiplier.lte(toBNWei(4)),
+      this.gasMultiplier.gte(toBNWei("0.00")) && this.gasMultiplier.lte(toBNWei(4)),
       `Gas multiplier out of range (${this.gasMultiplier})`
     );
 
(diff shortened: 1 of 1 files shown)


src/clients/ProfitClient.ts
```diff
@@ -239,7 +239,7 @@ export class ProfitClient {
     const netRelayerFeePct = netRelayerFeeUsd.mul(fixedPoint).div(relayerCapitalUsd);
 
     // If token price or gas cost is unknown, assume the relay is unprofitable.
```
pxrl, line 241:
```suggestion
    // If token price or gas price is unknown, assume the relay is unprofitable.
```

james-a-morris (approved):
LGTM pending review from @nicholaspai . Out of curiosity, this is meant to be a very temporary change to allow for testing? If so, should we encode this multiplier in an ENV var

nicholaspai (approved):

pxrl:
> LGTM pending review from @nicholaspai . Out of curiosity, this is meant to be a very temporary change to allow for testing? If so, should we encode this multiplier in an ENV var

Just to follow up on this, this is already controlled by the `RELAYER_GAS_MULTIPLIER` env var. The PR is just expanding the range of possible values to allow a 0 multiplier, which is useful for testing. The default value still remains at a multiplier of 1.2 (set in `src/common/Constants.ts`).

PR #753, opened 2023-06-20
improve(multicaller-client): Ignore expected revert reasons for batch txns

nicholaspai:
Today  the Multicaller client ignores known revert reasons when simulating individual transactions but not batch transactions. We should do the same for batch txns because this handles the common case where 
the batch contains only 1 txn and that has a clear revert reason that we expect.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -183,7 +183,7 @@ export class MultiCallerClient {
     );
     const batchSimResults = await this.txnClient.simulate(batchTxns);
     const batchesAllSucceeded = batchSimResults.every(({ succeed, transaction, reason }, idx) => {
-      if (this.canIgnoreRevertReason({ succeed, transaction, reason })) {
+      if (!succeed && this.canIgnoreRevertReason({ succeed, transaction, reason })) {
         this.logger.debug({
           at: "MultiCallerClient#executeChainTxnQueue",
           message: `Ignoring revert reason for ${networkName} transaction batch!`,
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -190,7 +190,6 @@ export class MultiCallerClient {
           batchTxn: { ...transaction, contract: transaction.contract.address },
           reason,
         });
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -183,12 +183,21 @@ export class MultiCallerClient {
...
+          reason,
+        });
+      } else {
+        this.logger[succeed ? "debug" : "error"]({
+          at: "MultiCallerClient#executeChainTxnQueue",
+          message: `${succeed ? "Successfully simulated" : "Failed to simulate"} ${networkName} transaction batch!`,
+          batchTxn: { ...transaction, contract: transaction.contract.address },
+          reason,
+        });
+      }
```
pxrl, line 200: How about this? The only difference is the log level of the emitted message, rather than the message being different. But I think that's OK (??)

```suggestion
        this.logger[succeed || this.canIgnoreRevertReason({ succeed, transaction, reason }) ? "debug" : "error"]({
          at: "MultiCallerClient#executeChainTxnQueue",
          message: `${succeed ? "Successfully simulated" : "Failed to simulate"} ${networkName} transaction batch!`,
          batchTxn: { ...transaction, contract: transaction.contract.address },
          reason,
        });
      }
```

nicholaspai: Oh nice great refactor

pxrl (approved):
So as I understand it, the input to this `txnClient.simulate()` is an array of transactions that _may_ have been bundled into a multicall3 `aggregate()` call, but may also be a regular Across `multicall()` call. We optimistically try try that because it's faster and cheaper to simulate the complete bundle.

If the simulation is fine, then we hand the batch off for submission and everything is OK (subject to any on-chain state changes that might subsequently cause them to fail anyway).

_If_ any of the transactions fail, then we emit the log message that is the subject of this PR, and then go through and individually simulate each transaction in isolation to identify which one failed. Any failing transactions are filtered out from the batch, and then the remaining transactions are submitted for re-packaging and submission.

Would be good to confirm that my understanding is correct; if so, LGTM :+1:

PR #758, opened 2023-06-21
feat(UBA): Derive UBA relayer fees via UBAClient

pxrl:
Introduces UBA fees to the relayer. UBA fees are only applied if the deposit version is >= the minimum applicable UBA version.

Closes ACX-1189

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1,4 +1,5 @@
 import { groupBy } from "lodash";
+import { clients as sdkClients, utils as sdkUtils } from "@across-protocol/sdk-v2";
 import {
   BigNumber,
   winston,
@@ -13,6 +14,8 @@ import { createFormatFunction, etherscanLink, formatFeePct, toBN, toBNWei } from
 import { RelayerClients } from "./RelayerClientHelper";
 import { Deposit, DepositWithBlock } from "../interfaces";
 import { RelayerConfig } from "./RelayerConfig";
+
+const { UBAActionType } = sdkClients;
 
 const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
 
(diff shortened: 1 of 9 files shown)


src/relayer/Relayer.ts
```diff
@@ -14,6 +15,8 @@ import { RelayerClients } from "./RelayerClientHelper";
 import { Deposit, DepositWithBlock } from "../interfaces";
 import { RelayerConfig } from "./RelayerConfig";
 
+const { UBAActionType } = sdkClients;
```
pxrl, line 18: Have imported this one directly because `UBAActionType` is used as both a type and an Object, and typescript seems really picky about how to re-export those from clients/interfaces.

src/relayer/Relayer.ts
```diff
@@ -200,10 +203,13 @@ export class Relayer {
       }
 
       if (tokenClient.hasBalanceForFill(deposit, unfilledAmount)) {
+        // The pre-computed realizedLpFeePct is for the pre-UBA fee model. Update it to the UBA fee model if necessary.
+        // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
+        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit, l1Token.symbol);
```
pxrl, line 208: I've realised there's an issue with this. This line updates the deposit object that's held by the SpokePoolClient, and that object has already been stored via its deposit hash in a separate SpokePoolClient object. So this will _probably_ mean that we need to pull realizedLpFeePct computation out of the SpokePoolClient.

src/relayer/Relayer.ts
```diff
@@ -357,6 +363,54 @@ export class Relayer {
...
+  ): Promise<BigNumber> {
+    if (!sdkUtils.isUBA(version)) {
+      return deposit.realizedLpFeePct;
+    }
+
+    const { hubPoolClient, ubaClient } = this.clients;
+    hubPoolBlockNumber ??= hubPoolClient.latestBlockNumber;
+
+    const { originChainId, destinationChainId, amount } = deposit;
+    const { systemFee: realizedLpFeePct } = await ubaClient.computeSystemFee(
```
nicholaspai, line 380: Can we log a breakdown of this fee for debugging?

nicholaspai (approved):
This looks great. I think we will want some debug level log that breaks down the fee components at least at the outset of the UBA launch

src/relayer/Relayer.ts
pxrl: Good call - added here: https://github.com/across-protocol/relayer-v2/pull/758/commits/bb936e24fa5247fbe0cc3925dfb9cba455384959

PR #759, opened 2023-06-21
feat: Add slow fill "grey" list mode to Relayer

nicholaspai:
Enables relayer to mark certain depositors as slow-fill only. Could also use this pattern for a depositor "blacklist"

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -161,19 +161,6 @@ export class Relayer {
         continue;
       }
 
-      // If depositor is on the slow deposit list, then send a zero fill to initiate a slow relay and return early.
-      if (config.slowDepositors.includes(deposit.depositor) && fillCount === 0) {
-        if (sendSlowRelays && tokenClient.hasBalanceForZeroFill(deposit) && fillCount === 0) {
-          this.logger.debug({
-            at: "Relayer",
-            message: "Initiating slow fill for grey listed depositor",
-            depositor: deposit.depositor,
-          });
-          this.zeroFillDeposit(deposit);
-        }
-        continue;
-      }
-
       // We query the relayer API to get the deposit limits for different token and destination combinations.
       // The relayer should *not* be filling deposits that the HubPool doesn't have liquidity for otherwise the relayer's
       // refund will be stuck for potentially 7 days.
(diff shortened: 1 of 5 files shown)


src/relayer/Relayer.ts
```diff
@@ -161,6 +161,19 @@ export class Relayer {
         continue;
       }
 
+      // If depositor is on the slow deposit list, then send a zero fill to initiate a slow relay and return early.
+      if (config.slowDepositors.includes(deposit.depositor) && fillCount === 0) {
+        if (sendSlowRelays && tokenClient.hasBalanceForZeroFill(deposit) && fillCount === 0) {
```
mrice32, line 166: nit: do we need to re-check fillCount === 0 in this second if?

src/relayer/Relayer.ts
```diff
@@ -161,6 +161,19 @@ export class Relayer {
         continue;
       }
 
+      // If depositor is on the slow deposit list, then send a zero fill to initiate a slow relay and return early.
```
mrice32, line 164: nit: should we put this _below_ the limits check (the block below) so this doesn't accidentally push us over the hubpool limits? (I'm not entirely sure whether it's better to just ignore these in that case)

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
nicholaspai: good point. I also want to put it under the empty message check

src/relayer/RelayerConfig.ts
```diff
@@ -53,6 +56,9 @@ export class RelayerConfig extends CommonConfig {
     this.relayerTokens = RELAYER_TOKENS
       ? JSON.parse(RELAYER_TOKENS).map((token) => ethers.utils.getAddress(token))
       : [];
+    this.slowDepositors = SLOW_DEPOSITORS
+      ? JSON.parse(RELAYER_TOKENS).map((token) => ethers.utils.getAddress(token))
```
mrice32, line 60:
```suggestion
      ? JSON.parse(SLOW_DEPOSITORS).map((depositor) => ethers.utils.getAddress(depositor))
```

mrice32 (approved):
LGTM! One copy-paste error

mrice32 (approved):

PR #760, opened 2023-06-21
improve(Relayer): More intelligent UBA refund chain selection

pxrl:
This implements the "Repayment Chain Choice Module" described in
ACX-977. It prefers to take refunds on the chainId suggested by the
Inventory Manager, then the destinationChainId, and then the HubPool
chainId, before reverting to an ordered list of the most profitable
refund chains. In each case, it must be profitable for the relayer to
take a refund there for it to use that chainId.

Some minor tweaks were needed in the ProfitClient to make this more
cohesive.

Closes ACX-977.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -326,7 +326,7 @@ export class ProfitClient {
     l1Token: L1Token
   ): boolean {
     const { profitable } =  this.computeFillProfitability(deposit, fillAmount, refundFee, l1Token);
-    return profitable;
+    return profitable ?? false;
   }
 
   captureUnprofitableFill(deposit: DepositWithBlock, fillAmount: BigNumber): void {
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -270,7 +270,7 @@ export class ProfitClient {
     return fillAmount.mul(tokenPriceInUsd).div(toBN(10).pow(l1TokenInfo.decimals));
   }
 
-  isFillProfitable(
+  computeFillProfitability(
     deposit: Deposit,
     fillAmount: BigNumber,
     refundFee: BigNumber,
@@ -317,6 +317,16 @@ export class ProfitClient {
     }
 
     return fill;
+  }
+
+ isFillProfitable(
+    deposit: Deposit,
+    fillAmount: BigNumber,
+    refundFee: BigNumber,
+    l1Token: L1Token
+  ): boolean {
+    const { profitable } =  this.computeFillProfitability(deposit, fillAmount, refundFee, l1Token);
+    return profitable;
   }
 
   captureUnprofitableFill(deposit: DepositWithBlock, fillAmount: BigNumber): void {
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -319,13 +319,8 @@ export class ProfitClient {
     return fill;
   }
 
- isFillProfitable(
-    deposit: Deposit,
-    fillAmount: BigNumber,
-    refundFee: BigNumber,
-    l1Token: L1Token
-  ): boolean {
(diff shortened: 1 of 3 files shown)


src/clients/ProfitClient.ts
```diff
@@ -270,7 +270,12 @@ export class ProfitClient {
     return fillAmount.mul(tokenPriceInUsd).div(toBN(10).pow(l1TokenInfo.decimals));
   }
 
-  isFillProfitable(deposit: Deposit, fillAmount: BigNumber, refundFee: BigNumber, l1Token: L1Token): boolean {
+  isFillProfitable(
+    deposit: Deposit,
+    fillAmount: BigNumber,
+    refundFee: BigNumber,
+    l1Token: L1Token
+  ): FillProfit | undefined {
```
nicholaspai, line 278: Either we should change this function name or have it return `boolean` as having it return an object or undefined here doesn't seem intuitive. Can we rename it `getFillProfitability`?

src/relayer/Relayer.ts
```diff
@@ -313,37 +311,70 @@ export class Relayer {
...
-      this.logger.debug({ at: "Relayer", message: "Skipping repayment chain determination for partial fill" });
+      this.logger.debug({
+        at: "Relayer",
+        message: `Skipping repayment chain determination for partial fill on ${destinationChain}`,
+        deposit: { originChain, depositId, destinationChain, depositHash },
+      });
+      return destinationChainId;
     }
 
+    const preferredChainId = await inventoryClient.determineRefundChainId(deposit);
```
nicholaspai, line 348: Should we expect an array of preferred chain Id's from the inventory client?

src/relayer/Relayer.ts
```diff
@@ -313,37 +311,70 @@ export class Relayer {
...
+    );
+
+    const refundChainsByProfit = refundChainIds.sort((chainA, chainB) =>
+      refundChainProfits[chainA].netRelayerFeePct.gte(refundChainProfits[chainB].netRelayerFeePct) ? 1 : -1
+    );
+
+    // If none of the preferred refund chains are profitable, take the refund wherever it's profitable.
+    // This may also produce no chainId, in which case the fill is truly unprofitable.
+    const repaymentChainId = [
+      preferredChainId,
```
nicholaspai, line 372: nit: should handle case where `preferredChainId` is duplicative with destination chain or the hub chain

src/relayer/Relayer.ts
pxrl: Ultimately we should, yeah. This PR is an step in that direction but doesn't dive into the inventory client yet. My thinking is that that would probably be bundled up into the inventory client overhaul, because it's not really needed for the UBA launch.

src/relayer/Relayer.ts
pxrl: I considered that but didn't land on a concise way of doing it, and ultimately settled for this because it's harmless to evaluate the same chainId twice as long as the ordering is OK. Do you have a suggestion for filtering out the duplicates whilst maintaining the ordering?

src/clients/ProfitClient.ts
pxrl: Fair point - it used to return a boolean but I modified that in the PR. Have added a wrapper named `isFillProfitable()` to restore the old behaviour here: https://github.com/across-protocol/relayer-v2/pull/760/commits/ae3de2c223d3c8208b3ae16540a3848ae37a0c2d.

src/relayer/Relayer.ts
pxrl: Improved it a bit here: https://github.com/across-protocol/relayer-v2/pull/760/commits/f768f6684ec136c475c274a1e145c48b10f15065. It's still possible for duplicates to appear in the set of [preferredChainId, destinationChainId, hubPoolClient.chainId], but it doesn't affect the result because they always appear at the head of the array. Happy to iterate on it if there's something cleaner.

PR #764, opened 2023-06-22
improve(InventoryClient): Use HubPool chainId

pxrl:
Makes the InventoryClient more flexible and should save it from doing some crazy things in a testnet environment. No functional change intended as part of this modification.


nicholaspai (approved):
nice much needed lgtm

src/clients/InventoryClient.ts
```diff
@@ -51,11 +51,14 @@ export class InventoryClient {
...
   getBalanceOnChainForL1Token(chainId: number | string, l1Token: string): BigNumber {
     // We want to skip any l2 token that is not present in the inventory config.
-    if (String(chainId) !== "1" && this.inventoryConfig.tokenConfig[l1Token][String(chainId)] === undefined) {
+    chainId = Number(chainId);
+    if (
+      chainId !== this.hubPoolClient.chainId &&
+      this.inventoryConfig.tokenConfig[l1Token][String(chainId)] === undefined
+    ) {
       return toBN(0);
     }
```
james-a-morris, line 60: I am a tad confused at this configuration. If we guarantee that the chainId should be the hubpool chainId, then why would we need to pass it? 
```
this.inventoryConfig.tokenConfig[l1Token][String(this.hubPoolClient.chainId)] === undefined
```

james-a-morris (commented):
I'm a bit confused with the structure of this function. I left a comment.

cc @nicholaspai

src/clients/InventoryClient.ts
pxrl: As I read the logic, it matches the expected behaviour in the comment on line 53. I have submitted this as-is because this commit just retains the existing logic, rather than modifying it...but we should circle back to it if the described behaviour is actually not corrected cc @james-a-morris @nicholaspai

PR #765, opened 2023-06-22
improve(Relayer): Step towards intelligent UBA refund chain selection

pxrl:
This change initially proposed to implement preferential refund chain
selection based on: 
 - Destination chain ID
 - HubPool chain ID
 - Profitable chain IDs (if any).

As-is, the change now retains the pre-existing repayment chain selection
based solely on input from the Inventory Client (or just the destination
chain, if inventory management is not configured). The ProfitClient can 
now compute the profitability of an array of chain IDs, and will return
their profitability (yes/no) as well as the net profitability to permit
sorting/preferential selection (note: >0% does not guarantee that the
fill would be considered profitable).

The subsequent work to preferentially select the best repayment chain
will be bundled into the InventoryClient update.

Closes ACX-977.


src/relayer/Relayer.ts
```diff
@@ -329,37 +327,85 @@ export class Relayer {
...
+      refundChains[chainA].netRelayerFeePct.gte(refundChains[chainB].netRelayerFeePct) ? 1 : -1
+    );
+
+    // When the refund chainId proposed by the inventory client is one of [destinationChainId, HubPoolChainId],
+    // prioritise taking refunds on the desinationChainId first. This helps to avoid the destination SpokePool running
+    // over-balance. If the preferredChainId is for a third chain, then accept that proposal as-is and preference the
+    // destinationChainId and HubPool chainId as the next best options, subject to profitability. If none of these
+    // refund chains are profitable, go mercenary and take the refund wherever it's most profitable. This may also
+    // produce no chainId, in which case the fill is truly unprofitable and may be ignored.
+    const preferredChainIds = [
```
nicholaspai, line 396: @mrice32 this is the strategy that I suggested to paul: prioritize `destinationChain` and then Ethereum, and then all chains, unless there's a preferred chain.

Is that in line with our offline discussions about the repayment chain choice strategy?

src/relayer/Relayer.ts
```diff
@@ -329,37 +327,85 @@ export class Relayer {
     }
   }
 
-  protected async resolveRepaymentChain(deposit: Deposit, fillAmount: BigNumber): Promise<number> {
+  protected async resolveRepaymentChain(
```
nicholaspai, line 330: As per our discussion offline, I could imagine adding a function in the future `resolveRepaymentChains(...deposit: DepositWithBlock[])` that takes in a list of all deposits it wants to fill and returns the refund chain for each one

src/relayer/Relayer.ts
```diff
@@ -329,37 +327,85 @@ export class Relayer {
...
+      refundChainIds
+        .map((chainId, idx) =>
+          profitClient.computeFillProfitability(deposit, fillAmount, refundFees[idx], hubPoolToken)
+        )
+        .map(({ profitable, netRelayerFeePct }, idx) => [refundChainIds[idx], { profitable, netRelayerFeePct }])
+    );
+
+    // Sort the residual chainIds according to their respective profitabilities.
+    const refundChainsByProfit = refundChainIds.sort((chainA, chainB) =>
+      refundChains[chainA].netRelayerFeePct.gte(refundChains[chainB].netRelayerFeePct) ? 1 : -1
```
mrice32, line 387: I don't think this is a valid comparitor function according to the requirements listed here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort.

If you swap the arguments, it should return 0 in both cases or the signs should flip.

Excerpt:
> More formally, the comparator is expected to have the following properties, in order to ensure proper sort behavior:
>
> Pure: The comparator does not mutate the objects being compared or any external state. (This is important because there's no guarantee when and how the comparator will be called, so any particular call should not produce visible effects to the outside.)
>
> Stable: The comparator returns the same result with the same pair of input.
Reflexive: compareFn(a, a) === 0.
>
> Anti-symmetric: compareFn(a, b) and compareFn(b, a) must both be 0 or have opposite signs.
>
> Transitive: If compareFn(a, b) and compareFn(b, c) are both positive, zero, or negative, then compareFn(a, c) has the same positivity as the previous two.

src/relayer/Relayer.ts
```diff
@@ -329,37 +327,85 @@ export class Relayer {
...
+
+    // When the refund chainId proposed by the inventory client is one of [destinationChainId, HubPoolChainId],
+    // prioritise taking refunds on the desinationChainId first. This helps to avoid the destination SpokePool running
+    // over-balance. If the preferredChainId is for a third chain, then accept that proposal as-is and preference the
+    // destinationChainId and HubPool chainId as the next best options, subject to profitability. If none of these
+    // refund chains are profitable, go mercenary and take the refund wherever it's most profitable. This may also
+    // produce no chainId, in which case the fill is truly unprofitable and may be ignored.
+    const preferredChainIds = [
+      ...new Set(
+        [destinationChainId, hubPoolClient.chainId].includes(preferredChainId)
```
mrice32, line 398: This sorting strategy is a bit confusing to me, tbh.

Why sort differently if the preferred chain is the destination or hub chain? Why not always put the destination chain first if it's profitable (assuming the goal is the over-balance issue mentioned in the comment)?

I don't understand why a 3rd chain being "preferred" changes the sorting logic so substantially...?

mrice32 (commented):
Two minor comments!

src/relayer/Relayer.ts
pxrl: This is a misunderstanding on my part because I hadn't spent enough time studying the InventoryClient.

I'd understood that it could request a refund on a third chain, and in this case wanted to give it the opportunity to propose that, with the expectation it'd only do so if we really had a need for funds on another chain. I've checked again and can see that it will only ever request on the destination chain or mainnet. Per the logic you've highlighted, we are actually unconditionally prioritising in the order of `[destinationChainId, hubPoolClient.chainId]` - which is what we want, but the call to InventoryClient is misleading its proposed refund chain is unconditionally overridden.

I've updated the code to make it more clear by hardcoding the priority to be `[destinationChainId, hubPoolClient.chainId]`. As I see it, it only makes sense to reintroduce the InventoryClient here after we update it with new logic.

src/relayer/Relayer.ts
pxrl: Oof - that's a good catch. Thanks for pointing that out! Has been corrected here: https://github.com/across-protocol/relayer-v2/pull/765/commits/3173bfd3e78d77a7849ac3d23560183d9bd85003

src/relayer/Relayer.ts
```diff
@@ -329,37 +327,81 @@ export class Relayer {
...
+    // Ensure that the refund token exists on the candidate chainId.
+    const refundChainIds = this.config.relayerDestinationChains.filter((chainId) =>
+      hubPoolClient.l2TokenEnabledForL1Token(hubPoolToken.address, chainId)
+    );
+
+    // Estimate the profitability of taking a refund on each candidate refundChainId.
+    const refundFees = await this.computeRefundFees(version, fillAmount, refundChainIds, hubPoolToken.symbol);
+    const refundChains = Object.fromEntries(
+      refundChainIds
+        .map((chainId, idx) => profitClient.getFillProfitability(deposit, fillAmount, refundFees[idx], hubPoolToken))
```
nicholaspai, line 379: Is this `.map` to check profitablity across all `chainId`'s expected to take a non negligible amount of time or is it instant?

If not instant, we might want to lazily get the profitability of each refund chain and pick the first one that is profitable. This way, our average run time stays low as in most cases, the first or second chain will be profitable

pxrl: It should be very fast to compute this for each chain. There are no blocking operations on the profitability determination; it uses pre-fetched token prices and assumes a fixed gas cost for the fill, so it's basically just BigNumber operations.

Computing the balancing fees (prerequisite) [is a blocking operation on the first time it's called](https://github.com/across-protocol/sdk-v2/blob/ca5b92540c621c2e63d589f886b3dfd46790de92/src/clients/UBAClient/UBAClientAbstract.ts#L102), but subsequent calls don't hit the blocking path. We should never incur that cost this late in the process, so I expect it will also be quite fast.

In the case of lazy evaluation, that's probably marginally useful if we are always going on the destination chain, but it would mean that we can't do any profit-based chain prioritisation.

In any case, it seems like we'll go for something a bit simpler now, so we can probably park some of this newer code, for eventual relegation to a specific type of refund strategy (i.e. arb/profit-maxi strategy).

src/relayer/Relayer.ts
```diff
@@ -329,38 +327,69 @@ export class Relayer {
...
+    const preferredChainId = await inventoryClient.determineRefundChainId(deposit);
+    if (!sdkUtils.isUBA(version)) {
+      const profitable = profitClient.isFillProfitable(deposit, fillAmount, toBN(0), hubPoolToken);
+      return profitable ? preferredChainId : undefined;
+    }
+
+    // Prefer the chainId suggested by the InventoryClient, but fall back to one of destinationChainId
+    // or hubPoolClient.chainId if the refunds on the preferred chainId is unprofitable.
+    const refundChainIds = [preferredChainId];
+    [destinationChainId, hubPoolClient.chainId].forEach((chainId) => {
```
pxrl, line 373: To be removed.

src/relayer/Relayer.ts
```diff
@@ -329,38 +327,69 @@ export class Relayer {
...
-      repaymentChainId = await this.clients.inventoryClient.determineRefundChainId(deposit);
-      if (!Object.keys(this.clients.spokePoolClients).includes(destinationChainId)) {
-        throw new Error("Fatal error! Repayment chain set to a chain that is not part of the defined sets of chains!");
+    const preferredChainId = await inventoryClient.determineRefundChainId(deposit);
+    if (!sdkUtils.isUBA(version)) {
+      const profitable = profitClient.isFillProfitable(deposit, fillAmount, toBN(0), hubPoolToken);
+      return profitable ? preferredChainId : undefined;
+    }
+
+    // Prefer the chainId suggested by the InventoryClient, but fall back to one of destinationChainId
```
nicholaspai, line 370: as discussed offline, I recommend removing this fallback logic and just simplifying heuristic to:
- get `preferredChainId` from `inventoryClient`
- If profitable continue
- else skip

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/765/files#r1243815924

src/relayer/Relayer.ts
pxrl: Implemented here: https://github.com/across-protocol/relayer-v2/pull/765/commits/df2185fa57cb0d33fa6c8a6cace774fd3dac2f38

Had to follow up with a could of merge updates and lint fixes.

nicholaspai (approved):

PR #766, opened 2023-06-22
fix: Include logMessage to finalize messages that are contained in same L2 withdrawal txn

nicholaspai:
We fixed this for `proveMessage` but not `finalizeMessage`


james-a-morris (approved):
LGTM - good catch

pxrl (approved):

PR #771, opened 2023-06-23
feat: Add demo scripts for interacting with ZkSync

nicholaspai:
# Motivation

Prepare repo to interact with ZkSync Goerli and Era Mainnet networks. Include scripts that we can use in the future to inform how to build the ZKSync inventory manager adapter

# Details
Adds some configuration to support zksync. Tested this with the `sendTokensAdapter` script and a new script `zkSyncDemo` where I tested out deposits to ZkSync and other actions we'll want from day 1 on the relayer.

Transactions successfully executed on chain `324` (ZkSync mainnet):
- `transfer` ETH: https://explorer.zksync.io/tx/0x254948f1d03ebbbbc533fde06f0735a59cc1e313c79f8f25b65e72c0459e2661
- `deposit` wrapped ETH: 0x9f667f0cd79dc28a08a1372343c9bf7e9f72f6dfeec8a71ddd5f40d0f5035788
- `withdraw` wrapped ETH: 0x4fac492fb8e7771590a92bf9aee64f490467cf80e4ff222cd6fbd596195c7269
- `depositETH` to ZkSync from Ethereum: https://etherscan.io/tx/0xdc9409002251adb1c8dd6559b6f6ddc586e6b4b662ce1d7a4a0b241bb946e895
- `depositERC20` to ZkSync from Ethereum: https://etherscan.io/tx/0x9fc458abe311d5ba7fc9a887e595fa251b24604a286a794ce53f7cb634f29df2


scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+    throw new Error("Define `zkSyncChainId` as the zkSync chain you want to connect on");
+  }
+
+  const baseSigner = await getSigner();
+  const connectedSigner = baseSigner.connect(await getProvider(Number(args.chainId)));
+  console.log("Connected to account", connectedSigner.address);
+  const recipient = args.to;
+  const token = args.token;
+  if (!ethers.utils.isAddress(recipient)) {
+    throw new Error("invalid addresses");
```
pxrl, line 81:
```suggestion
    throw new Error(`invalid \"to\" address: ${recipient}`);
```

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+      l2GasLimit,
+      l2PubdataByteLimit
+    );
+    console.log(`Contract estimated L2 transaction base cost: ${l2TransactionBaseCost.toString()}`);
+    const _txnRequest: AugmentedTransaction = {
+      contract: mailboxContract,
+      chainId: args.chainId,
+      method,
+      args: params,
+      gasLimitMultiplier: 1,
```
pxrl, line 126: This defaults to 1x so you can easily leave it out, though if we suspect that the call might need padding then it's also fine to explicitly define it here.

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+    const amountFromWei = ethers.utils.formatUnits(args.amount, decimals);
+    // Check the user is ok with the info provided. else abort.
+    console.log(`Send ${symbol} with amount ${amountFromWei} tokens to ${recipient} on chain ${args.zkSyncChainId}`);
+    if (!(await askYesNoQuestion("\nConfirm that you want to execute this transaction?"))) {
+      return;
+    }
+    console.log("sending...");
+
+    const l1ERC20Bridge = new Contract(
+      "0x57891966931Eb4Bb6FB81430E6cE0A03AAbDe063",
```
pxrl, line 152: Do we have somewhere central to define hardcoded contract addresses?

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+      l1Erc20BridgeInterface,
+      connectedSigner
+    );
+    const method = "deposit";
+    const l2GasLimit = await zksync.utils.estimateDefaultBridgeDepositL2Gas(
+      connectedSigner.provider,
+      zkSyncProvider,
+      token,
+      toBN(args.amount),
```
pxrl, line 161: Just to be sure, is it intended to use `args.amount` or `amountFromWei` here?

src/common/Constants.ts
```diff
@@ -60,9 +60,11 @@ export const DEFAULT_MIN_DEPOSIT_CONFIRMATIONS = {
   10: 0,
   137: 128, // Commonly used finality level for CEX's that accept Polygon deposits
   288: 0,
+  324: 0,
```
pxrl, line 63: I found [this](https://era.zksync.io/docs/reference/concepts/finality.html#instant-confirmations) section describing zkSync finality. I'm a little unclear about best practice due to the following sentence: `More cautious users may prefer to wait for full finality, or for any of the intermediate steps described above, before considering the assets as received.`

Have you seen any better description?

src/utils/NetworkUtils.ts
```diff
@@ -14,6 +14,12 @@ export function getNetworkName(networkId: number | string): string {
     if (Number(networkId) == 421613) {
       return "ArbitrumGoerli";
     }
+    if (Number(networkId) == 324) {
+      return "ZkSync";
+    }
+    if (Number(networkId) == 280) {
+      return "ZkSync-Goerli";
+    }
```
pxrl, line 22: wdyt about importing our own version of `PublicNetworks` that we can maintain more easily? Could place it in the sdk.

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+      value: l2TransactionBaseCost.add(args.amount),
+      message: "Deposit ETH to ZkSync",
+      mrkdwn: "Deposit ETH to ZkSync",
+    };
+    const simulationResult = (await txnClient.simulate([_txnRequest]))[0];
+    if (simulationResult.succeed) {
+      await txnClient.submit(args.chainId, [simulationResult.transaction]);
+    } else {
+      console.log("Simulation failed", simulationResult);
+    }
```
pxrl, line 136: This part seems common to both use cases - worth dropping it out the bottom of the if statement?

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+  if (!Object.keys(args).includes("to")) {
+    throw new Error("Define `to` as where you want to send funds to");
+  }
+  if (!Object.keys(args).includes("chainId")) {
+    throw new Error("Define `chainId` as the chain you want to connect on");
+  }
+  if (!Object.keys(args).includes("zkSyncChainId")) {
+    throw new Error("Define `zkSyncChainId` as the zkSync chain you want to connect on");
+  }
```
pxrl, line 73:
```suggestion
  const fields: { [key: string]: errMsg } = {
    "token": "Define `token` as the address of the token to send",
    "amount": "Define `amount` as how much you want to send",
    "to": "Define `to` as where you want to send funds to",
    "chainId": "Define `chainId` as the chain you want to connect on",
    "zkSyncChainId": "Define `zkSyncChainId` as the zkSync chain you want to connect on",
  };
  
  Object.entries(fields).forEach(([key, errMsg]) => {
    if (!isDefined(args[key])) {
      throw new Error(errMsg);
    }
  }
```

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+      contract: mailboxContract,
+      chainId: args.chainId,
+      method,
+      args: params,
+      gasLimitMultiplier: 1,
+      value: l2TransactionBaseCost.add(args.amount),
+      message: "Deposit ETH to ZkSync",
+      mrkdwn: "Deposit ETH to ZkSync",
+    };
+    const simulationResult = (await txnClient.simulate([_txnRequest]))[0];
```
pxrl, line 131: I threw up a quick [PR](https://github.com/across-protocol/relayer-v2/pull/777) to make `TransactionClient.submit()` call `self.simulate()` if the gasLimit is not defined. This would probably mean you can skip the call to `txnClient.simulate()` here.

scripts/zkSyncDemo.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+  if (!Object.keys(args).includes("zkSyncChainId")) {
+    throw new Error("Define `zkSyncChainId` as the zkSync chain you want to connect on");
+  }
+
+  const baseSigner = await getSigner();
+  const connectedSigner = baseSigner.connect(await getProvider(Number(args.chainId)));
+  console.log("Connected to account", connectedSigner.address);
+  const recipient = args.to;
+  const token = args.token;
```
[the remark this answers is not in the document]
nicholaspai: good point added

scripts/zkSyncDemo.ts
nicholaspai: We hardcode the bridge addresses in `/clients/bridges` like in `OptimismAdapter` and `ArbitrumAdapter` so maybe we can start a new pattern for `ZkSyncAdapter`.

WDYT of this: f4e18426fdff17065dc2a38c23124781d9b87f3b

src/common/Constants.ts
nicholaspai: The way I understand it, full finality is 24 hours but with a single centralized sequencer in practice (to bootstrap the network), finality is instant: 

_Taken from [source](https://era.zksync.io/docs/reference/concepts/finality.html#instant-confirmations)_

> ## Finality on zkSync Era
>
> As with other L2 rollup systems, finality and security on zkSync Era is tied to the finality and security of the underlying L1 Ethereum chain. Finality on zkSync Era is the time taken from sending a transaction to when the L1 smart contract updates the L2 state. The process has a few steps:
> 
> - Fill a batch with transactions (usually takes a few minutes).
> - Commit the batch to Ethereum.
> - Generate a proof for the whole batch (usually takes around an hour).
> - Submit the proof for verification by the Ethereum smart contract.
> - Finalize the batch on Ethereum (delayed by ~21 hours as a security measure during the alpha phase).
>
> When this process is complete, around ~24 hours in total, the batch is as final as any Ethereum transaction included in the same Ethereum block.
> 
> ## Instant confirmations
> Even though zkSync Era finality is usually around 24 hours, users may consider transactions submitted to zkSync Era as having instant confirmation. Transaction details are instantly displayed in the UI and API (although they are marked as unconfirmed), and the transferred assets can be used immediately to make further transfers (which might even end up in the same zkSync Era batch).
> 
> More cautious users may prefer to wait for full finality, or for any of the intermediate steps described above, before considering the assets as received.

src/utils/NetworkUtils.ts
nicholaspai: Yea that would be great, its much slower to have to add these to the `protocol/packages/common` package

scripts/zkSyncDemo.ts
nicholaspai: Nice refactor +1 70dc556a6f5787dddad02fa67ece0a0b4534cf05

nicholaspai:
https://github.com/across-protocol/relayer-v2/pull/771#discussion_r1241802447

@pxrl wdyt about this refactor https://github.com/across-protocol/relayer-v2/pull/771/commits/5f87d523c4b9cc215862e7447b715aed4a0c3365

PR #777, opened 2023-06-26
improve(TransactionClient): Simulation on undefined gasLimit

pxrl:
If the transaction gasLimit is undefined, simulate it before submission. This saves the caller from having to call simulate() before submit().


nicholaspai (approved):

pxrl:
Parking this PR for now, since it has a minor issue and is otherwise unrelated to ongoing work.

PR #778, opened 2023-06-26
feat(UBA): Support generating refundRequest() txns

pxrl:
This allows the relayer to generate a refundRequest transaction from a FillWithBlock event. The maxCount argument is currently pegged to UINT256_MAX but will need to be updated so that it's (relatively) configurable via the environment.

Closes ACX-1195


src/relayer/Relayer.ts
```diff
@@ -301,6 +302,48 @@ export class Relayer {
...
+      destinationToken,
+      fillAmount: amount,
+      realizedLpFeePct,
+      repaymentChainId: chainId,
+      blockNumber: fillBlock,
+    } = fill;
+
+    const contract = spokePoolClients[chainId].spokePool;
+    const method = "requestRefund";
+    // @todo: Should support specifying max impact against the refund amount (i.e. to mitigate price impact by fills).
```
nicholaspai, line 320: ideally is configurable by setting `MAX_SLIPPAGE_PCT` and then uses UBA Client to solve for what to set here

src/relayer/Relayer.ts
```diff
@@ -301,6 +302,48 @@ export class Relayer {
...
+      depositId,
+      destinationChainId,
+      destinationToken,
+      fillAmount: amount,
+      realizedLpFeePct,
+      repaymentChainId: chainId,
+      blockNumber: fillBlock,
+    } = fill;
+
+    const contract = spokePoolClients[chainId].spokePool;
```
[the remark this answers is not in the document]
pxrl: The renaming of variables is mostly just personal preference in order to 1:1 map to the input arguments on the `requestRefund()` and `enqueueTransaction()` functions respectively (i.e. I am a big fan of being able to instantiate the object literal with only the relevant member names).

Happy to revert to `repaymentChainId` if it helps readability though :+1:

refundRequest:
```
    function requestRefund(
        address refundToken,
        uint256 amount,
        uint256 originChainId,
        uint256 destinationChainId,
        int64 realizedLpFeePct,
        uint32 depositId,
        uint256 fillBlock,
        uint256 maxCount
    )
```

enqueueTransaction() takes an AugmentedTransaction argument:

```
export interface AugmentedTransaction {
  contract: Contract;
  chainId: number;
  method: string;
  args: any[];
  message?: string;
  mrkdwn?: string;
  // ...
}
```

src/relayer/Relayer.ts
```diff
@@ -306,6 +308,105 @@ export class Relayer {
     this.clients.tokenClient.decrementLocalBalance(deposit.destinationChainId, deposit.destinationToken, fillAmount);
   }
 
+  // Strategy for requesting refunds: Query all refunds requests, and match them to fills.
+  // The remaining fills are eligible for new requests.
+  async requestRefunds(): Promise<void> {
+    const { multiCallerClient, ubaClient } = this.clients;
+    const spokePoolClients = Object.values(this.clients.spokePoolClients);
+
+    await ubaClient.update();
```
nicholaspai, line 317: Can't we assume already that the `ubaClient` is up to date? Updating within this function breaks the existing pattern of not updating after the initial update at the `index.ts` level

src/relayer/Relayer.ts
```diff
@@ -306,6 +308,105 @@ export class Relayer {
...
+
+    const eligibleFillsByRefundChain: { [chainId: number]: FillWithBlock[] } = Object.fromEntries(
+      spokePoolClients.map(({ chainId }) => [chainId, []])
+    );
+
+    const refundRequestsByDstChain: { [chainId: number]: RefundRequestWithBlock[] } = {};
+
+    spokePoolClients.forEach((spokePoolClient) => {
```
nicholaspai, line 325: Would it be helpful for there to be a function on the `UBAClient` like so:

`getRefundRequests(relayerAddress: string): { [chainId: number]: RefundRequestWithBlock[] }`

It seems more natural that the UBAClient should have a function that the relayer can call to get its list of **_validated_** refund requests across all chains

src/relayer/Relayer.ts
```diff
@@ -306,6 +308,105 @@ export class Relayer {
...
+      fills.forEach((fill) => {
+        this.logger.debug({ at: "Relayer::requestRefunds", message: "Requesting refund for fill.", fill });
+        this.requestRefund(fill);
+      });
+    });
+
+    await multiCallerClient.executeTransactionQueue();
+  }
+
+  findEligibleFills(
```
nicholaspai, line 354:
```suggestion
  findFillsRequiringSeparateRefundRequest(
```
I think this name is more clear, but not perfect. I think "eligible" is just kind of vague

src/relayer/Relayer.ts
```diff
@@ -306,6 +308,105 @@ export class Relayer {
...
+    spokePoolClient: SpokePoolClient,
+    refundRequests: RefundRequestWithBlock[] = []
+  ): FillWithBlock[] {
+    const refundRequestDepositIds = refundRequests.map(({ depositId }) => depositId);
+
+    // Find fills where repayment was requested on another chain.
+    const eligibleFills = spokePoolClient.getFillsForRelayer(this.relayerAddress).filter((fill) => {
```
nicholaspai, line 361: Can we assume that `getFillsForRelayer` only returns validated fills?

src/relayer/Relayer.ts
```diff
@@ -306,6 +308,105 @@ export class Relayer {
...
+    });
+
+    await multiCallerClient.executeTransactionQueue();
+  }
+
+  findEligibleFills(
+    spokePoolClient: SpokePoolClient,
+    refundRequests: RefundRequestWithBlock[] = []
+  ): FillWithBlock[] {
+    const refundRequestDepositIds = refundRequests.map(({ depositId }) => depositId);
```
nicholaspai, line 358: Do you need this `.map`? Can't you just check `refundRequests.find(x => x.depositId === depositId) !== undefined`?

src/relayer/Relayer.ts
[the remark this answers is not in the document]
nicholaspai: Ok I'm good with your suggestion then

src/relayer/Relayer.ts
```diff
@@ -528,4 +629,22 @@ export class Relayer {
...
+    const { symbol, decimals } = hubPoolClient.getL1TokenInfoForL2Token(destinationToken, destinationChainId);
+
+    const refundChain = getNetworkName(fill.repaymentChainId);
+    const originChain = getNetworkName(fill.originChainId);
+    const destinationChain = getNetworkName(destinationChainId);
+
+    const _fillAmount = createFormatFunction(2, 4, false, decimals)(fill.fillAmount.toString());
+
+    return (
+      `Requested refund on ${refundChain} for ${_fillAmount} ${symbol} for ${originChain} depositId ${depositId},` +
```
nicholaspai, line 646:
```suggestion
      `🌈 Requested refund on ${refundChain} for ${_fillAmount} ${symbol} for ${originChain} depositId ${depositId},` +
```

Adding a very bright and catchy emoji so we can see these more clearly

src/relayer/Relayer.ts
pxrl: Assuming that the ubaClient is up to date will mean that we only submit refund requests associated with previous bot run/iteration, and will not include fills from the current iteration. Updating the ubaClient here is a bandaid for that, though it currently lacks synchronisation with the recently submitted fills so the update might complete before some or all of those fills are confirmed.

I think the logical next step is to modify the MultiCallerClient to return TransactionResponse objects for each transaction that is executed. With this, we can synchronously wait for confirmation of any fills that were just made, and then combine those fills with any previous ones that don't have associated RefundRequested events. This would enable us to get away without the subsequent update.

src/relayer/Relayer.ts
pxrl: Unfortunately not when requesting the events directly via the SpokePoolClient. RefundRequested events need verification across at least one separate SpokePoolClieint instances (originChainId == repaymentChainId), but more often probably two separate SpokePoolClient instances (originChainId != repaymentChainId).

The SpokePoolClient therefore only does very limited verification on the events, and they need to be post-processed by a layer above (nominally the UBAClient). This fits well with your comment here: https://github.com/across-protocol/relayer-v2/pull/778/files#r1244063966.

src/relayer/Relayer.ts
pxrl: I chose to spin up a separate array of `depositId` because [it's significantly faster to evaluate `Array.includes()` relative to `Array.find()`](https://github.com/across-protocol/relayer-v2/pull/778/files#r1244063966).

There is an issue here though - the fills correspond to deposits on multiple SpokePools, so `depositId` can collide. It'll be necessary to apply additional criteria to do the matching, and that won't work as well with `Array.includes()`.

src/clients/UBAClient.ts
```diff
@@ -14,4 +17,24 @@ export class UBAClient extends clients.UBAClient {
   ) {
     super(chainIdIndices, hubPoolClient, spokePoolClients, {} as RelayFeeCalculatorConfig, logger);
   }
+
+  getRefundRequests(chainId: number, relayer?: string): RefundRequestWithBlock[] {
```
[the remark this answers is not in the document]
nicholaspai: that's a good instinct, i think RL would run this by filtering on its own `relayer` address

src/relayer/Relayer.ts
```diff
@@ -13,12 +14,13 @@ import {
...
 import { createFormatFunction, etherscanLink, formatFeePct, toBN, toBNWei } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
-import { Deposit, DepositWithBlock, L1Token } from "../interfaces";
+import { Deposit, DepositWithBlock, FillWithBlock, L1Token, RefundRequestWithBlock } from "../interfaces";
 import { RelayerConfig } from "./RelayerConfig";
 
 const { UBAActionType } = sdkClients;
 
 const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
+export const UINT256_MAX = toBN(2).pow(256).sub(1); // @note: Temporary
```
nicholaspai, line 23: Have you looked into this? https://docs.ethers.org/v5/api/utils/constants/#constants-MaxUint256

src/relayer/Relayer.ts
```diff
@@ -304,6 +306,107 @@ export class Relayer {
...
+  // The remaining fills are eligible for new requests.
+  async requestRefunds(sendRefundRequests = true): Promise<void> {
+    const { multiCallerClient, ubaClient } = this.clients;
+    const spokePoolClients = Object.values(this.clients.spokePoolClients);
+
+    const eligibleFillsByRefundChain: { [chainId: number]: FillWithBlock[] } = Object.fromEntries(
+      spokePoolClients.map(({ chainId }) => [chainId, []])
+    );
+
+    // For each refund/repayment Spoke Pool, group its set of by corresponding destination chain.
```
nicholaspai, line 319:
```suggestion
    // For each refund/repayment Spoke Pool, group by destination chain.
```

src/relayer/Relayer.ts
```diff
@@ -304,6 +306,107 @@ export class Relayer {
...
+      spokePoolClients.map(({ chainId, latestBlockNumber }) => {
+        const lookback = latestBlockNumber - this.config.maxRelayerLookBack;
+        const refundRequests = ubaClient
+          .getRefundRequests(chainId, this.relayerAddress)
+          .filter(({ blockNumber }) => blockNumber > lookback);
+        return [chainId, refundRequests];
+      })
+    );
+
+    for (const spokePoolClient of spokePoolClients) {
```
nicholaspai, line 331:
```suggestion
    // For each destination chain, validate refund requests were sent on refund chain and group validated ones by refund chain
    for (const spokePoolClient of spokePoolClients) {
```

src/relayer/index.ts
```diff
@@ -22,10 +24,19 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Wallet): P
     for (;;) {
       await updateRelayerClients(relayerClients, config);
 
+      // @note: For fills with a different repaymentChainId, refunds are requested on the _subsequent_ relayer run.
```
nicholaspai, line 27: +1

src/relayer/Relayer.ts
```diff
@@ -304,6 +306,107 @@ export class Relayer {
...
+
+  findFillsWithoutRefundRequests(
+    spokePoolClient: SpokePoolClient,
+    refundRequests: RefundRequestWithBlock[] = []
+  ): FillWithBlock[] {
+    const refundRequestDepositIds: { [chainId: number]: number[] } = [];
+
+    refundRequests.forEach(({ originChainId, depositId }) => {
+      refundRequestDepositIds[originChainId] ??= [];
+      refundRequestDepositIds[originChainId].push(depositId);
```
nicholaspai, line 356: Shouldn't we do some validation on the refund request before inserting it into here? So we don't unintentionally skip sending a refund request because of the presence of an invalid refund request

nicholaspai (commented):
This looks almost done, some nits, but biggest issue is https://github.com/across-protocol/relayer-v2/pull/778/files#r1246973317

src/relayer/Relayer.ts
```diff
@@ -304,6 +306,107 @@ export class Relayer {
...
+    });
+
+    // Find fills where repayment was requested on another chain.
+    const fills = spokePoolClient.getFillsForRelayer(this.relayerAddress).filter((fill) => {
```
nicholaspai, line 360: I think we also need to validate these fills. I'd recommend maybe adding a function to the `SpokePoolClient` that allows the caller to `getValidatedFillsForRelayer`

test/Relayer.RefundRequests.ts
```diff
@@ -0,0 +1,224 @@
...
+    await updateAllClients();
+
+    // Set the spokePool's time to the provider time. This is done to enable the block utility time finder identify a
+    // "reasonable" block number based off the block time when looking at quote timestamps.
+    for (const spokePool of [spokePool_1, spokePool_2]) {
+      await spokePool.setCurrentTime(await getLastBlockTime(spokePool.provider));
+    }
+  });
+
+  it("Correctly identifies fills eligible for cross-chain refunds", async function () {
```
nicholaspai, line 141: Suggest adding the following tests:
- Does not submit refund requests for invalid fills => send a fill with repaymentChainId != destinationChainId but set realizedLpFeePct incorrectly, for example
- Does not send refund request if refund request already exists for fill
- Sends refund request if invalid refund requests exists for fill

src/relayer/Relayer.ts
```diff
@@ -304,6 +306,107 @@ export class Relayer {
...
+    // Bound the event range by the relayer deposit lookback.
+    const refundRequestsByDstChain = Object.fromEntries(
+      spokePoolClients.map(({ chainId, latestBlockNumber }) => {
+        const lookback = latestBlockNumber - this.config.maxRelayerLookBack;
+        const refundRequests = ubaClient
+          .getRefundRequests(chainId, this.relayerAddress)
+          .filter(({ blockNumber }) => blockNumber > lookback);
+        return [chainId, refundRequests];
+      })
+    );
```
james-a-morris, line 329: Would this be more readable as a `reduce`?
```suggestion
    const refundRequestsByDstChain = spokePoolClients.reduce((acc, {chainId, latestBlockNumber}) => (
        {
            ...acc,
            [chainId]: ubaClient
                             .getRefundRequests(chainId, this.relayerAddress)
                             .filter(({blockNumber}) => 
                                 blockNumber > (latestBlockNumber - this.config.maxRelayerLookback)
                             )
        }
    ), {});
```

src/relayer/index.ts
```diff
@@ -22,10 +24,19 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Wallet): P
     for (;;) {
```
james-a-morris, line 24: OOC - I wonder if we should be more explicit with this. I know it's not a part of the scope of this PR, but imho `while(true) {` feels more explicit

james-a-morris (commented):
Left two reviews.

src/relayer/Relayer.ts
pxrl: Ah nice - I had looked for this under ethers.BigNumber but hadn't seen the `Constants` section. Fixed here: https://github.com/across-protocol/relayer-v2/pull/778/commits/dd2dbbc9d760350d7cf9efc68e0a7e6b4d8e1e90

src/clients/UBAClient.ts
```diff
@@ -14,4 +21,58 @@ export class UBAClient extends clients.UBAClient {
   ) {
     super(chainIdIndices, hubPoolClient, spokePoolClients, {} as RelayFeeCalculatorConfig, logger);
   }
+
```
pxrl, line 24: Note: `getFills()` and `getRefundRequests()` have been parked here for demonstration, but they obviously don't belong here. My inclination is to put them in a `SpokePoolClientAggregator` class that sits between the `UBAClient` and the set of `SpokePoolClient` instances. This would ensure that the `UBAClient` (and other users) alway get validated and filtered event objects. This functionality should also live in SDK-v2, and is only proposed here for initial agreement on the functionality.

src/relayer/Relayer.ts
pxrl: The set of refund requests supplied as input is now pre-validated, so we have verified the existence of a corresponding deposit. Since this is the top-level application logic, I think it's preferable to ensure that that validation occurs at a lower layer.

src/finalizer/index.ts
pxrl, line 1: These changes relate to the upstream `pxrl/getBlockForTimestamp` branch and will disappear once it is merged.

src/common/ClientHelper.ts
pxrl, line 1: These changes relate to the upstream `pxrl/getBlockForTimestamp` branch and will disappear once it is merged.

src/scripts/validateRootBundle.ts
pxrl, line 1: These changes relate to the upstream `pxrl/getBlockForTimestamp` branch and will disappear once it is merged.

src/utils/BlockUtils.ts
pxrl, line 1: These changes relate to the upstream `pxrl/getBlockForTimestamp` branch and will disappear once it is merged.

src/utils/UmaUtils.ts
pxrl, line 1: These changes relate to the upstream `pxrl/getBlockForTimestamp` branch and will disappear once it is merged.

src/relayer/Relayer.ts
pxrl: This was bundled into a subsequent commit :+1:

src/clients/UBAClient.ts
pxrl: @james-a-morris Tagging you in this one - @nicholaspai has mentioned that you might be working on something similar. WDYT about putting this in the SDK in a `SpokePoolClientManager` or `SpokePoolClientAggregator` class?

pxrl:
FWIW, the latest updates to this PR are expected to fail because they require an additional SDK bump.

src/relayer/Relayer.ts
pxrl: Has been implemented.

src/clients/UBAClient.ts
pxrl: Update: These wrappers are still parked here because `getRefundRequests()` requires a lot of context to be passed in. They'll be migrated down the stack in sdk-v2 in follow-up work.

src/relayer/index.ts
pxrl: If I was writing it from scratch I'd probably use `while(true)`, but (to me) it's within the realm of preference and will inevitably vary from one developer to the next. We could adjust it here, but the next developer that comes along might just adjust it back, so I'm inclined to not touch it unless we're actually refactoring the functionality here.

test/Relayer.RefundRequests.ts
pxrl: The latter two tests have been identified in `Relayer.RefundRequests.ts`, but they're not implemented at this point. I'd like to just get this in as-is and follow up with those two tests, because this PR is at a size where it risks being accidentally impacted again by additional changes in sdk-v2.

nicholaspai (approved):

PR #779, opened 2023-06-27
feat: add support for POOL token

dohaki:
Bumps contracts-v2 and sdk-v2 to latest version with fixed POOL token addresses. Also, add support for POOL in `PolygonAdapter`.

There was a breaking change in `@across-protocol/sdk-v2@0.10.4`. The fix is here https://github.com/across-protocol/relayer-v2/pull/779/commits/661ad21aaab448d1839515b461a110ebbde7ca56. Let me know if this is not correct

### DIFF
--- a/src/clients/bridges/PolygonAdapter.ts
+++ b/src/clients/bridges/PolygonAdapter.ts
@@ -85,6 +85,13 @@ const tokenToBridge = {
     l1AmountProp: "amount",
     l2AmountProp: "value",
   }, // ACX
+  [TOKEN_SYMBOLS_MAP.POOL.addresses[CHAIN_IDs.MAINNET]]: {
+    l1BridgeAddress: "0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf",
+    l2TokenAddress: TOKEN_SYMBOLS_MAP.POOL.addresses[CHAIN_IDs.POLYGON],
+    l1Method: "LockedERC20",
+    l1AmountProp: "amount",
+    l2AmountProp: "value",
+  }, // POOL
   [TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET]]: {
     l1BridgeAddress: "0x8484Ef722627bf18ca5Ae6BcF031c23E6e922B30",
     l2TokenAddress: TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.POLYGON],
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -221,7 +221,12 @@ export class Relayer {
       if (tokenClient.hasBalanceForFill(deposit, unfilledAmount)) {
         // The pre-computed realizedLpFeePct is for the pre-UBA fee model. Update it to the UBA fee model if necessary.
         // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
-        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit, l1Token.symbol);
+        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(
+          version,
+          deposit,
+          l1Token.symbol,
+          l1Token.address
+        );
 
         const repaymentChainId = await this.resolveRepaymentChain(deposit, unfilledAmount);
         // @todo: For UBA, compute the anticipated refund fee(s) for *all* candidate refund chain(s).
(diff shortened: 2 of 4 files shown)


src/relayer/Relayer.ts
```diff
@@ -221,7 +221,12 @@ export class Relayer {
       if (tokenClient.hasBalanceForFill(deposit, unfilledAmount)) {
         // The pre-computed realizedLpFeePct is for the pre-UBA fee model. Update it to the UBA fee model if necessary.
         // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
-        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit, l1Token.symbol);
+        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(
+          version,
+          deposit,
+          l1Token.symbol,
+          l1Token.address
+        );
```
pxrl, line 229:
```suggestion
        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit, l1Token);
```

src/relayer/Relayer.ts
```diff
@@ -367,6 +372,7 @@ export class Relayer {
     version: number,
     deposit: Deposit,
     symbol: string,
+    hubPoolTokenAddress: string,
```
pxrl, line 375:
```suggestion
    hubPoolToken: L1Token,
```
 (Plus would need to remove the `symbol` argument above)

src/relayer/Relayer.ts
```diff
@@ -381,7 +387,14 @@ export class Relayer {
       lpFee,
       depositBalancingFee: depositFee,
       systemFee: realizedLpFeePct,
-    } = await ubaClient.computeSystemFee(originChainId, destinationChainId, symbol, amount, hubPoolBlockNumber);
+    } = await ubaClient.computeSystemFee(
+      originChainId,
+      destinationChainId,
+      symbol,
+      hubPoolTokenAddress,
```
pxrl, line 394:
```suggestion
      hubPoolToken.symbol,
      hubPoolToken.address,
```

pxrl (approved):
LGTM. I left a suggestion to adjust the method prototype for `this.computeRealizedLpFeePct()`, but it's also fine to just get this in as-is because I'm working in this area anyway, so I can always follow up with that change.

src/relayer/Relayer.ts
pxrl: Forget this - seems easier for me to just adjust it afterwards :+1:

src/relayer/Relayer.ts
pxrl: Forget this - seems easier for me to just adjust it afterwards :+1:

src/relayer/Relayer.ts
pxrl: Forget this - seems easier for me to just adjust it afterwards :+1:

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -85,6 +85,13 @@ const tokenToBridge = {
     l1AmountProp: "amount",
     l2AmountProp: "value",
   }, // ACX
+  [TOKEN_SYMBOLS_MAP.POOL.addresses[CHAIN_IDs.MAINNET]]: {
+    l1BridgeAddress: "0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf",
```
nicholaspai, line 89: did you determine this by testing a deposit? Just curious

nicholaspai (approved):
Looks good. Few questions
- do we not need changes in the optimism adapter?
- just confirming how you set the polygon bridge address
- why are we changing how we call sdk functions?

src/clients/bridges/PolygonAdapter.ts
dohaki: I assumed that this is the address after looking at the POOL token holdings here https://etherscan.io/tokenholdings?a=0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf

But now I made sure by testing a deposit https://etherscan.io/tx/0xe4f779b745b56e90dff2fe7bae9d0f769b0331c17f566467eedb82019ddc41ed :)

dohaki:
> do we not need changes in the optimism adapter?

AFAICT POOL doesn't use a custom bridge, so I don't think we need changes in the Optimism adapter. Correct me if I am wrong.

> just confirming how you set the polygon bridge address

Tested a deposit here https://etherscan.io/tx/0xe4f779b745b56e90dff2fe7bae9d0f769b0331c17f566467eedb82019ddc41ed

> why are we changing how we call sdk functions?

There was a breaking change in https://github.com/across-protocol/sdk-v2/releases/tag/v0.10.4 which I released while bumping the contracts-v2 version. The PR https://github.com/across-protocol/sdk-v2/pull/255 introduced new function signatures, so I had to some adjustments.

PR #783, opened 2023-06-28
fix(Relayer): Fix Relayer realizedLpFeePct computation

pxrl:
Rather than the SpokePool block mapped to a HubPool block, the relevant block for the LP fee is resolved via the quoteTimestamp on the deposit.

Spotted by Nick.


src/relayer/Relayer.ts
pxrl, line 1: FWIW we might also decide to relocate this to the relayer-v2 UBAClient as part of this PR, since it's more at home there (and will be needed for the Dataworker integration anyway).

pxrl: Scrap that - there's nothing useful to migrate here. The point of this method is just to translate the UBA "system fee" to the Relayer's concept of a `realizedLpFeePct`, which it needs when making a fill.

nicholaspai (approved):
This won't work yet until the `computeSystemFee` works correctly but LGTM

PR #785, opened 2023-06-30
improve(BalanceAllocator): Allow caller to specify fungible tokens

nicholaspai:
This feature simplifies the BalanceAllocator logic by treating fungible `tokens` passed into `requestBalanceAllocations` as the same 
token (arbitrarily keyed by the first token in the list). This makes accounting of the `balances` and `used` dictionaries simpler.

The tradeoff is that the caller has more responsibility to only pass in truly fungible tokens whose balances are collapsed in reality. 
If the caller doesn't actually combine token0 and token1, for example, then the next call to the BalanceAllocator would be using 
inaccurate numbers.

I think this is a fine tradeoff as the BalanceAllocator is shared only amongst dataworker-executor functions which correctly collapse 
WETH and ETH, the two fungible tokens on OVM chains.

### DIFF
--- a/src/clients/BalanceAllocator.ts
+++ b/src/clients/BalanceAllocator.ts
@@ -16,28 +16,42 @@ export class BalanceAllocator {
 
   constructor(readonly providers: { [chainId: number]: ethers.providers.Provider }) {}
 
-  // Note: The caller is suggesting that `tokens` for a request are interchangeable.
-  // The tokens whose balances are depleted first should be placed at the front of the array.
+  /**
+   * @notice Caller is assumiing that all `tokens` are interchangeable and that if their balances are not collapsed
+   * into one, then the accounting could be messed up and future calls to getBalance or getUsed could be incorrect.
+   * @dev: The caller is suggesting that `tokens` for a request are interchangeable.
+   * This assumes therefore that the balance of all `tokens` are ultimately treated as one, so for accounting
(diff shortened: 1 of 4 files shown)


mrice32 (commented):
General comment: I don't see the bug in the original code, do you have an explanation for what's wrong?

I was looking into the callsite, though, and it doesn't look like it's passing ETH and WETH as fungible tokens... or am I missing something?

Thoughts?

nicholaspai:
> General comment: I don't see the bug in the original code, do you have an explanation for what's wrong?
> 
> 
> 
> I was looking into the callsite, though, and it doesn't look like it's passing ETH and WETH as fungible tokens... or am I missing something?
> 
> 
> 
> Thoughts?

Maybe there wasn't a bug but the original code was a lot more complex than was required. As we really want to treat the two balances as equal and in practice the spoke pools collapse the balances. That's the only way we can pass in multiple tokens. 

I wasn't able to verify the bug before the issue in prod fixed itself

nicholaspai:
> General comment: I don't see the bug in the original code, do you have an explanation for what's wrong?
> 
> I was looking into the callsite, though, and it doesn't look like it's passing ETH and WETH as fungible tokens... or am I missing something?
> 
> Thoughts?

The alternative PR would be this: https://github.com/across-protocol/relayer-v2/pull/788

Maybe we go with 788 instead for now and see if that works, and then we can close this one. Its a bit hard to test in practice however

PR #787, opened 2023-07-04
fix(Relayer): Correct deposit lookback

pxrl:
The MAX_RELAYER_DEPOSIT_LOOKBACK config item is currently intended to be specified in units of seconds, but was previously configured with a number of blocks. The Relayer class however passed MAX_RELAYER_DEPOSIT_LOOKBACK directly in to the getUnfilledDeposits() helper function, where it is compared directly against a block number. This should to be translated from seconds to a block number first.

The effect of this misalignment in units is probably that the relayer looks back more (or less) depending on whether the average block time for each chain is more or less than 1 second. This would be very pronounced on mainnet.

This fix may resolve an issue that we've received vague reports about in the past from the relayer community. Feedback has been that the relayer is less likely to capture fills if it runs in looping mode, so many from the community choose to run in serverless mode and loop their bot externally.

Fixes ACX-1258

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -3,6 +3,10 @@ import { DepositWithBlock, FundsDepositedEvent } from "../interfaces";
 import { getNetworkName, paginatedEventQuery, spreadEventWithBlockNumber } from "../utils";
 
 export class SpokePoolClient extends clients.SpokePoolClient {
+  getCurrentTime(): number {
+    return this.currentTime;
+  }
+
   async findDeposit(depositId: number, destinationChainId: number, depositor: string): Promise<DepositWithBlock> {
     // Binary search for block where SpokePool.numberOfDeposits incremented to fill.depositId + 1.
     // This way we can get the blocks before and after the deposit with deposit ID = fill.depositId
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -3,10 +3,6 @@ import { DepositWithBlock, FundsDepositedEvent } from "../interfaces";
 import { getNetworkName, paginatedEventQuery, spreadEventWithBlockNumber } from "../utils";
 
 export class SpokePoolClient extends clients.SpokePoolClient {
-  getCurrentTime(): number {
-    return this.currentTime;
-  }
-
   async findDeposit(depositId: number, destinationChainId: number, depositor: string): Promise<DepositWithBlock> {
     // Binary search for block where SpokePool.numberOfDeposits incremented to fill.depositId + 1.
     // This way we can get the blocks before and after the deposit with deposit ID = fill.depositId
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -181,10 +181,10 @@ export class Monitor {
   }
 
   async reportUnfilledDeposits(): Promise<void> {
-    const unfilledDeposits = getUnfilledDeposits(
+    const unfilledDeposits = await getUnfilledDeposits(
       this.clients.spokePoolClients,
-      this.monitorConfig.maxRelayerLookBack,
-      this.clients.configStoreClient
+      this.clients.configStoreClient,
+      this.monitorConfig.maxRelayerLookBack
     );
 
     // Group unfilled amounts by chain id and token id.
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -1,7 +1,7 @@
 import assert from "assert";
-import { ConfigStoreClient, HubPoolClient, SpokePoolClient } from "../clients";
+import { ConfigStoreClient, HubPoolClient } from "../clients";
 import { Deposit, DepositWithBlock, Fill, FillsToRefund, FillWithBlock, SpokePoolClientsByChain } from "../interfaces";
-import { getBlockFinder, queryHistoricalDepositForFill } from "../utils";
+import { getBlockForTimestamp, queryHistoricalDepositForFill } from "../utils";
 import {
   BigNumber,
   assign,
@@ -234,11 +234,6 @@ export type RelayerUnfilledDeposit = {
   invalidFills: Fill[];
 };
 
-async function resolveBlockNumber(spokePoolClient: SpokePoolClient, timestamp: number): Promise<number> {
-  const blockFinder = await getBlockFinder(spokePoolClient.chainId);
-  return (await blockFinder.getBlockForTimestamp(timestamp)).number;
-}
-
 // @description Returns an array of unfilled deposits over all spokePoolClients.
 // @param spokePoolClients  Mapping of chainIds to SpokePoolClient objects.
 // @param configStoreClient ConfigStoreClient instance.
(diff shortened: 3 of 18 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -3,6 +3,10 @@ import { DepositWithBlock, FundsDepositedEvent } from "../interfaces";
 import { getNetworkName, paginatedEventQuery, spreadEventWithBlockNumber } from "../utils";
 
 export class SpokePoolClient extends clients.SpokePoolClient {
+  getCurrentTime(): number {
```
pxrl, line 6: I parked this here for now so as to avoid having to bump the SDK, but it can be proposed for migration there in a follow-up commit.

src/utils/FillUtils.ts
```diff
@@ -232,32 +234,52 @@ export type RelayerUnfilledDeposit = {
   invalidFills: Fill[];
 };
 
-// Returns all unfilled deposits over all spokePoolClients. Return values include the amount of the unfilled deposit.
-export function getUnfilledDeposits(
+async function resolveBlockNumber(spokePoolClient: SpokePoolClient, timestamp: number): Promise<number> {
+  const blockFinder = await getBlockFinder(spokePoolClient.chainId);
+  return (await blockFinder.getBlockForTimestamp(timestamp)).number;
```
pxrl, line 239: I'm using the blockFinder directly here because I don't anticipate any benefit from caching the block timestamps via the local getBlockForTimestamp() (which uses redis). It seems unlikely to me that we'd be reusing the exact block timestamp in subequent calls, because it's an arbitrary offset to an arbitrary block (i.e. whatever the user has configured for the lookback, and whatever the head of the chain was at the latest lookup).

pxrl: Also: could maybe consider some additional validation on the return from `getBlockForTimestamp()`. We're not really consistent about whether (and how much) to validate RPC provider responses...

src/utils/FillUtils.ts
```diff
@@ -232,32 +234,52 @@ export type RelayerUnfilledDeposit = {
...
+export async function getUnfilledDeposits(
   spokePoolClients: SpokePoolClientsByChain,
-  maxUnfilledDepositLookBack: number,
-  configStoreClient: ConfigStoreClient
-): RelayerUnfilledDeposit[] {
+  configStoreClient: ConfigStoreClient,
+  depositLookBack?: number
+): Promise<RelayerUnfilledDeposit[]> {
   const unfilledDeposits: RelayerUnfilledDeposit[] = [];
+  const chainIds = Object.values(spokePoolClients).map(({ chainId }) => chainId);
```
pxrl, line 253: `Object.values(spokePoolClients)` is used _repeatedly_ in this function. It'd be nicer to just receive the SpokePoolClients as an array, rather than an Object. It's more convenient for the callers to just pass in an array, though.

src/utils/FillUtils.ts
```diff
@@ -232,32 +234,52 @@ export type RelayerUnfilledDeposit = {
...
-  for (const originChain of chainIds) {
-    const originClient = spokePoolClients[originChain];
-    for (const destinationChain of chainIds) {
-      if (originChain === destinationChain) {
-        continue;
-      }
+  for (const originClient of Object.values(spokePoolClients)) {
+    const { chainId: originChainId } = originClient;
+    const chainIdx = chainIds.indexOf(originChainId);
+    assert(chainIdx !== -1, `Invalid chain index for chainId ${originChainId} (${chainIdx}) (${chainIds}`);
```
pxrl, line 270:
```suggestion
    assert(chainIdx !== -1, `Invalid chain index for chainId ${originChainId} (${chainIdx})`);
```

test/Monitor.ts
```diff
@@ -86,6 +86,11 @@ describe("Monitor", async function () {
     };
     const monitorConfig = new MonitorConfig(defaultMonitorEnvVars);
 
+    // @dev: Force maxRelayerLookBack to undefined to skip lookback block resolution.
```
pxrl, line 89: The reason for this is that trying to use the `BlockFinder` class in test causes problems because RPC providers are ultimately invoked, and the test environment has no `NODE_URL[S]_<CHAIN-ID>` variables defined.

nicholaspai (approved):
Have you tested this out locally?

pxrl:
> Have you tested this out locally?

Yep; tested locally that the existing computation was incorrect, and that it's correct after applying this change. Now, dividing the lookback configuration by the earliest found deposit gives approximately the average block time for the chain.

Unfortunately it's quite difficult to add automated tests for it at the moment because of the use of BlockFinder, which can't be injected. I avoided changing this because it requires a bit of additional refactoring, but I think we should consider some adjustments here post-UBA.

src/clients/SpokePoolClient.ts
james-a-morris: I include this in the SDK's PR here: https://github.com/across-protocol/sdk-v2/pull/267

src/monitor/Monitor.ts
```diff
@@ -181,10 +181,10 @@ export class Monitor {
   }
 
   async reportUnfilledDeposits(): Promise<void> {
-    const unfilledDeposits = getUnfilledDeposits(
+    const unfilledDeposits = await getUnfilledDeposits(
```
james-a-morris, line 184: Was this ever working correctly? AFAICT without the `await` this should fail

src/utils/FillUtils.ts
james-a-morris, line 253: Maybe we can create a transformer fn in our utils for this as a convenience wrapper.

```
export function convertSpokeRecords(Record<string, SpokePoolClient>): SpokePoolClient[];
export function findAllRelevantChainIds(Record<string, SpokePoolClient>): number[];
```

james-a-morris: I left a comment above

james-a-morris (commented):
LGTM!

src/utils/FillUtils.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/787#discussion_r1252291847

src/monitor/Monitor.ts
pxrl: The function was previously synchronous, but I made it async as part of this change because it now needs to make RPC queries via `blockFinder`.

pxrl:
I'm holding off on this PR until https://github.com/across-protocol/relayer-v2/pull/791 can be merged, since it will make it very easy to just use `getBlockForTimestamp()` directly.

src/utils/FillUtils.ts
pxrl: I suspect helper functions like this are probably unlikely to be widely used, so I'm probably OK to just leave it as-is. I just wanted to call it out before anyone else made the same observation :sweat_smile:

pxrl: Responded here: https://github.com/across-protocol/relayer-v2/pull/787#discussion_r1252291847. I'm happy to wait and see whether I see this issue anywhere else, because I'm not sure it'd be used very much.

PR #788, opened 2023-07-04
fix(dataworker): Add ETH to list of tokens we should request balance allocations for

nicholaspai:
When executing OVM SpokePool leaves we need to request both ETH and WETH balances together

Alternative implementation to #785


james-a-morris (approved):

mrice32 (approved):
LGTM! Great call!

PR #789, opened 2023-07-04
improve: include new spoke client fn into relayer

james-a-morris:


nicholaspai (approved):

pxrl (approved):
Nice, thanks to jumping on this!

PR #791, opened 2023-07-05
improve(BlockUtils): Simplify getBlockForTimestamp()

pxrl:
Two improvements here:
 - Remove the need to supply the HubPool chain ID to this function. I've been unable to find any other places in the code where the HubPool chain ID block numbers are accessed via the key `block_number_<timestamp>`, so just normalise all keys to be `<chainId>_block_number_<timestamp>`. This frees the caller from having to know what the HubPool chainId is.

 - Remove the need to supply the externally-supplied `currentChainTime`. This argument was _always_ supplied with the output of `getCurrentTime()`, which is not necessarily the current chain time anyway, and goes against the naming of the variable `currentTimeTime`. Instead, use the timestamp of the latest block to determine whether it's old enough to be cached.

As background, the motivation for this change is that the Relayer now needs to apply its "max lookback" on RefundRequest events as well, so we have a common need for resolving timestamps to a previous block number. Simplifying getBlockForTimestamp() allows us to benefit from the caching that it implements.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -66,13 +66,7 @@ export async function constructSpokePoolClientsWithLookback(
 
   // Use the first block that we'll query on mainnet to figure out which chains were enabled between then
   // and the the latest mainnet block. These chains were enabled via the ConfigStore.
-  const fromBlock_1 = await getBlockForTimestamp(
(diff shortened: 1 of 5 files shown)


nicholaspai (approved):
Good optimizations

PR #795, opened 2023-07-08
improve(utils): Catch expected RPC logs

nicholaspai:
We shouldn't send `error` level logs when the error is a nonce collision, which sometimes gets returned with an RPC specific error msg


src/utils/TransactionUtils.ts
```diff
@@ -16,12 +16,13 @@ export type TransactionSimulationResult = {
 };
 
 const txnRetryErrors = new Set(["INSUFFICIENT_FUNDS", "NONCE_EXPIRED", "REPLACEMENT_UNDERPRICED"]);
+const expectedRpcErrorMessages = new Set(["nonce has already been used", "intrinsic gas too low"]);
```
james-a-morris, line 19: Should this be in a constant folder because it's never changed?

james-a-morris (approved):
Left a nit

PR #796, opened 2023-07-08
fix(relayer): Don't complete slow fill-only fills

nicholaspai:
Follow up to https://github.com/across-protocol/relayer-v2/pull/759 that fixes bug where we currently fast fill any slow fill for certain depositors


mrice32 (approved):
LGTM!

PR #800, opened 2023-07-10
feat(relayer): Support updating Relayer ubaClient

pxrl:
This is a PoC, there are outstanding points to resolve on the arguments to ubaClient.update() (i.e. for relayer looping mode).

Additionally, per current implementation, the ubaClient will update "TokensBridged" events. This will slightly increase RPC usage but should otherwise have no noticeable effect.


src/relayer/RelayerClientHelper.ts
```diff
@@ -105,19 +106,26 @@ export async function constructRelayerClients(
...
-    "EnabledDepositRoute",
-    "RelayedRootBundle",
-    "ExecutedRelayerRefundRoot",
-  ]);
+  const { configStoreClient, spokePoolClients, ubaClient } = clients;
+
+  await configStoreClient.update();
+  const version = configStoreClient.getConfigStoreVersionForTimestamp();
+  if (sdkUtils.isUBA(version)) {
+    await ubaClient.update({}, true);
```
james-a-morris, line 114: This is on me, but the update function should be called with `undefined` if you're not looking to override state

james-a-morris, line 114:
```suggestion
    await ubaClient.update(undefined, true);
```

james-a-morris (commented):
Left a comment

src/relayer/RelayerClientHelper.ts
pxrl: I've updated this PR to handle that, but requiring the caller to pass in `undefined` as part of normal operation is counter-intuitive for me.

nicholaspai (approved):

james-a-morris (approved):
LGTM

src/relayer/RelayerClientHelper.ts
```diff
@@ -105,20 +107,27 @@ export async function constructRelayerClients(
...
-    "EnabledDepositRoute",
-    "RelayedRootBundle",
-    "ExecutedRelayerRefundRoot",
-  ]);
+  const { configStoreClient, spokePoolClients, ubaClient } = clients;
+
+  await configStoreClient.update();
+  const version = configStoreClient.getConfigStoreVersionForTimestamp();
+  if (sdkUtils.isUBA(version)) {
+    assert(configStoreClient.isValidConfigStoreVersion(version), `Relayer does not support UBA transfers`);
```
pxrl, line 115:
```suggestion
    assert(configStoreClient.isValidConfigStoreVersion(version), "Relayer does not support UBA transfers");
```

PR #806, opened 2023-07-12
fix(tests): Fix broken deepEqualsWithBigNumber implementation

nicholaspai:
`deepEqualsWithBigNumber` doesn't actually deeply compare objects as expected because any Object.toString() === Object.toString().

The easiest way I see to fix this is to replace the implementation using a combination of assert.deepStrictEquals and lodash.omit

I had to fix broken tests as a result

### DIFF
--- a/test/ConfigStoreClient.ts
+++ b/test/ConfigStoreClient.ts
@@ -7,7 +7,7 @@ import { MAX_REFUNDS_PER_RELAYER_REFUND_LEAF, MAX_L1_TOKENS_PER_POOL_REBALANCE_L
 import { DEFAULT_POOL_BALANCE_TOKEN_TRANSFER_THRESHOLD } from "./constants";
 import { GLOBAL_CONFIG_STORE_KEYS } from "../src/clients";
 import { SpokePoolTargetBalance } from "../src/interfaces";
-import { DEFAULT_CONFIG_STORE_VERSION } from "./mocks";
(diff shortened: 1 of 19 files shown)


test/ConfigStoreClient.ts
```diff
@@ -64,8 +71,13 @@ describe("AcrossConfigStoreClient", async function () {
     configStore = await (await getContractFactory("AcrossConfigStore", owner)).deploy();
     const receipt = await configStore.deployTransaction.wait();
     const eventSearchConfig = { fromBlock: receipt.blockNumber };
-    configStoreClient = new MockConfigStoreClient(createSpyLogger().spyLogger, configStore, eventSearchConfig);
-    configStoreClient.setConfigStoreVersion(0);
+    configStoreClient = new ConfigStoreClient(
```
nicholaspai, line 74: Unrelated to the `deepEquals` bug, I can't get any of the `MockConfigStoreClient` tests to work. It only works when I use the non mocked version. @pxrl i recall you worked on this recently, any ideas if I'm using it incorrectly?

pxrl: I _think_ this might be related to running tests in parallel - is that how you're testing @nicholaspai ?

pxrl: Confirmed unrelated to running the tests in parallel - there seems to have been a local issue with `yarn link`.

test/Dataworker.loadData.ts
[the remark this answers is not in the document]
nicholaspai: theyve always been broken, yep

PR #816, opened 2023-07-17
improve(Dataworker): Call `exchangeRateCurrent` more frequently

nicholaspai:
- Call `exchangeRateCurrent` when token hasn't been updated within past 2 hours rather than 24 hours
- Call `exchangeRateCurrent` before exiting early because challenge period hasn't passed yet for bundle. This increases chances that token exchange rate is synced before leaves are executed.


james-a-morris (approved):
QQ: does this increase the frequency by 24x to every hour?

nicholaspai:
> QQ: does this increase the frequency by 24x to every hour?

This will only run on `executePoolRebalanceLeaf` calls and only if there are leaves to execute. So it will likely increase it to 1x every bundle execution which is not nearly that often (more like every ~4 hours). This seems like a healthy frequency

james-a-morris (approved):
LGTM

PR #829, opened 2023-07-19
test: getUBAFeeConfig

james-a-morris:
Requires this PR: https://github.com/across-protocol/sdk-v2/pull/333


nicholaspai (commented):
This test doesn't test the situation where there are multipleUBA configs published in the config store at say blocks 10 and 100 and i want to `getUBAConfig` at block 9 (should return undefined or error), 10, 12 (should return first config), 100, 101, (should return second config) unless I'm missing somethingobvious

james-a-morris:
> This test doesn't test the situation where there are multipleUBA configs published in the config store at say blocks 10 and 100 and i want to `getUBAConfig` at block 9 (should return undefined or error), 10, 12 (should return first config), 100, 101, (should return second config) unless I'm missing somethingobvious

Resolved - but I'll need the SDK-V2 PR I opened to be merged before this PR will run

test/UBA.getUBAFeeConfig.ts
```diff
@@ -0,0 +1,183 @@
...
+          omega: { "10": [["0", "0"]], "137": [["0", "0"]], "42161": [["0", "0"]], default: [["0", "0"]] },
+          rebalance: {
+            "10": { threshold_upper: "200000000", target_upper: "100000000" },
+            "137": { threshold_upper: "100000000", target_upper: "0" },
+            "42161": { threshold_upper: "200000000", target_upper: "100000000" },
+          },
+        },
+      ];
+
+      const mockedConfigStore = hubPoolClient.configStoreClient as MockConfigStoreClient;
```
pxrl, line 134: Could alternatively just instantiate `let configStoreClient: MockConfigStoreClient;` up in the top-level `describe` codeblock. Then you should have access to it here without needing to assert the type.

test/UBA.getUBAFeeConfig.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      const blockNumbers: number[] = [];
+      for (const config of realisticConfigs) {
+        const { blockNumber } = mockedConfigStore.updateTokenConfig(
+          validToken.address,
+          JSON.stringify({ uba: config })
+        );
+        blockNumbers.push(blockNumber);
+        // Sleep for 1 second to ensure that the block number is different
+        await new Promise((resolve) => setTimeout(resolve, 500));
```
pxrl, line 144: You _should_ be able to get away without this sleep, since `EventManager.generateEvent()` enforces that the block number of successive events always increments. The only exception is if the caller supplies the block number, that it must be >= the  current block number.

test/UBA.getUBAFeeConfig.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+
+      // Find the starting and ending block range of all the block numbers
+      // Add an additional 10 blocks to the end to ensure that the config is correctly resolved post update
+      const startingBlock = Math.min(...blockNumbers);
+      const endingBlock = Math.max(...blockNumbers) + 10;
```
pxrl, line 150: This should also be the first and last elements of the array, no?

test/mocks/MockConfigStoreClient.ts
```diff
@@ -16,6 +17,25 @@ export class MockConfigStoreClient extends clients.mocks.MockConfigStoreClient {
...
+      chainId,
+      mockUpdate
+    );
+  }
+
+  public getUBAConfig(
+    l1TokenAddress: string,
+    blockNumber?: number | undefined
+  ): interfaces.UBAParsedConfigType | undefined {
+    return this.mockedUBAConfig ?? super.getUBAConfig(l1TokenAddress, blockNumber);
```
pxrl, line 35: +1 nice to fallback to the parent class if the config hasn't been overridden.

test/UBA.getUBAFeeConfig.ts
james-a-morris: Good call: https://github.com/across-protocol/relayer-v2/pull/829/commits/33e5914339f2e889af25eec457acd7a05b1a5448

test/UBA.getUBAFeeConfig.ts
james-a-morris: I had some issues with this. I figure it's more effective to grab the high/low & append 10 to test that the last bundle is found. Although I do agree that there is a more effective solution, the performance impact of the min/max should be minimal

test/UBA.getUBAFeeConfig.ts
```diff
@@ -0,0 +1,178 @@
...
+        .sort((a, b) => a.block - b.block);
+
+      await hubPoolClient.configStoreClient.update();
+      await hubPoolClient.update();
+
+      // Ensure that their length is 3
+      expect(hubPoolClient.configStoreClient.ubaConfigUpdates).to.be.lengthOf(3);
+
+      // Iterate through 30 blocks and ensure that the config is correctly resolved
+      for (const blockNumber of range(startingBlock, endingBlock)) {
```
nicholaspai, line 160: Want to check a few things
- edge cases: 
-     what happens when there are zero configs added (i.e. before you call updateTokenConfig)? It should return undefined right? 
-     what happens when the block number is really high, i.e. no configs after the block, should also return undefined?
- Have you manually logged out the configs to make sure they're printing out different ones? This is just to make sure the test is testing what you expect

james-a-morris: Edge Cases:

- Handled: https://github.com/across-protocol/relayer-v2/pull/829/commits/422ac846ee56f07e42f0ff977a99695bd929a2bb
- This should always default back to the latest config (or undefined if nothing has been updated)
- Yep! I've also added sanity checks

james-a-morris: @nicholaspai as per our offline conversation: config resolution where a config update isn't found will throw an error: https://github.com/across-protocol/relayer-v2/pull/829/commits/e748b3803c9cc0675d0102d8f8252abf3184c92c

test/UBA.getUBAFeeConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+            "137": { threshold_upper: "100000000", target_upper: "0" },
+            "42161": { threshold_upper: "200000000", target_upper: "100000000" },
+          },
+        },
+      ];
+
+      await hubPoolClient.configStoreClient.update();
+      await hubPoolClient.update();
+
+      // Test that the config is correctly resolved to undefined for a scenario where the block number is before the
```
nicholaspai, line 139:
```suggestion
      // Test that the function throws for a scenario where the block number is before the
```

nicholaspai (approved):
LGTM after remaining comments are resolved!

pxrl (approved):
LGTM. I like the structure of the test - it's very clean.

test/UBA.getUBAFeeConfig.ts
pxrl: Interesting - that seems odd and I'm not sure why there should be issues, but what you have is also fine.

PR #832, opened 2023-07-21
WIP: Make relayer work with updated UBAClient

nicholaspai:
Pairs with https://github.com/across-protocol/sdk-v2/pull/332

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -26,7 +26,6 @@ import {
 } from "../interfaces";
 import {
   PendingRootBundle,
-  TreeData,
   RunningBalances,
   PoolRebalanceLeaf,
   RelayerRefundLeaf,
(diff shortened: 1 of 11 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -184,12 +184,8 @@ export class BundleDataClient {
       this.clients.hubPoolClient.chainId,
       this.chainIdListForBundleEvaluationBlockNumbers
     )[0];
-    const version = this.clients.configStoreClient.getConfigStoreVersionForBlock(mainnetStartBlock);
     let isUBA = false;
-    if (utils.isUBA(version)) {
-      if (!this.clients.configStoreClient.isValidConfigStoreVersion(version)) {
```
nicholaspai, line 190: I removed all the `isValidConfigStoreVersion` in this repo as they are redundant. We check in various parts of the relayer and the dataworker as to whether we have the correct version, and then we refuse to act if we don't. There are a lot of config store functions related to checking the version and whether we have the right version and I found it confusing so I condensed the interface to just:
- `getConfigStoreVersionForTimestamp`/`getConfigStoreVersionForBlock`: straight forward, arguably should be collapsed into one func
- `hasValidConfigStoreVersionForTimestamp`: used by relayer/dataworker to decide whether to relay and propose/validate.

src/dataworker/Dataworker.ts
```diff
@@ -947,26 +928,45 @@ export class Dataworker {
...
     // If valid is false, we get a reason and we might get expected trees.
     | {
         valid: false;
         reason: string;
         expectedTrees?: {
-          poolRebalanceTree: TreeData<PoolRebalanceLeaf>;
-          relayerRefundTree: TreeData<RelayerRefundLeaf>;
-          slowRelayTree: TreeData<SlowFillLeaf>;
+          poolRebalanceTree: {
+            tree: MerkleTree<PoolRebalanceLeaf>;
```
nicholaspai, line 940: This fixed a compiler complaint that `MerkleTree` and `TreeData` have non overlapping types. Ultimately I prefer to use straightforward `MerkleTree<T>` as `TreeData<T>` just seems like a wrapper with no additional purpose

PR #835, opened 2023-07-25
feat: Cache eth_getBlockByNumber calls

pxrl:
This will be needed if the we become dependent on knowing the timestamp of the various {FundsDeposited, FilledRelay, RefundRequested} events.

Closes ACX-1344

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -117,6 +117,7 @@ function compareRpcResults(method: string, rpcResultA: any, rpcResultB: any): bo
 }
 
 class CacheProvider extends RateLimitedProvider {
+  public readonly getBlockByNumberPrefix: string;
   public readonly getLogsCachePrefix: string;
   public readonly callCachePrefix: string;
   public readonly maxReorgDistance: number;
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -208,9 +212,10 @@ class CacheProvider extends RateLimitedProvider {
       }
 
       return this.canCacheInformationFromBlock(toBlock);
-    } else if (method === "eth_call") {
-      // Block number is the second argument. Parse as hex.
-      const blockNumber = parseInt(params[1], 16);
+    } else if ("eth_call" === method || "eth_getBlockByNumber" === method) {
+      const idx = method === "eth_getBlockByNumber" ? 0 : 1;
```
nicholaspai, line 216: Can you add a comment explaining what the params for `eth_getBlockByNumber` are versus `eth_call`? Otherwise this logic is a bit hard to understand

pxrl: Clarified here: https://github.com/across-protocol/relayer-v2/pull/835/commits/2136faf19d62e4818774de52e19158ea6a94aeb6

nicholaspai (approved):

PR #836, opened 2023-07-25
feat: add cache agnostic redis cache

james-a-morris:
This PR uses the SDK-V2's agnostic caching interface to wrap a Redis client. We can now use an instance of this caching class in the SDK without the SDK needing to know anything about redis except that it can `set` and `get` keys of a specific type.


nicholaspai (approved):

pxrl (approved):
Looks good!

PR #841, opened 2023-07-26
improve(relayer): Simplify deposit version filtering

pxrl:
The version-filtered array of unfilled deposits was not subsequently used when deriving the confirmedUnfilledDeposits array that is ultimately used to make fills. This was fortunately not a showstopper for the relayer because we still had an old check in `getUnfilledDeposits()`. This PR removes that and implements a test for specific version-based filtering.

In order to make it harder to trip over this in future, factor out the version determination so that there are fewer overall variables to track in the checkForUnfilledDepositsAndFill() method. This also makes it easy to add a specific unit test.

Closes ACX-1346

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -12,6 +12,7 @@ import {
   getCurrentTime,
   buildFillRelayWithUpdatedFeeProps,
   isDepositSpedUp,
+  RelayerUnfilledDeposit,
 } from "../utils";
 import { createFormatFunction, etherscanLink, formatFeePct, toBN, toBNWei } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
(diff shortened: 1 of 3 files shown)


pxrl:
nb. This requires a new SDK version, so it will fail until that's in place.

src/relayer/Relayer.ts
```diff
@@ -61,16 +60,25 @@ export class Relayer {
       });
     }
 
-    const unfilledDepositAmountsPerChain: { [chainId: number]: BigNumber } = supportedDeposits
```
nicholaspai, line 64: I'm confused where the bug was before where we discarded the version field. Was it here?

nicholaspai (commented):
I'm still confused where the prior bug was can you point it out? Otherwise LGTM but i want to understand more

src/relayer/Relayer.ts
pxrl: The bug (...fortunately without consequence) was further down in the code: https://github.com/across-protocol/relayer-v2/blob/e2010dbe5233f70c899c932219ce885d839d66b5/src/relayer/Relayer.ts#L102

In the above code, we totally ignore `supportedDeposits` and source `confirmedUnfilledDeposits` from `unfilledDeposits`. This bypasses the intended version filtering that was done to produce `supportedDeposits`.

Fortunately `unfilledDeposits` was pre-filtered for the version, so we didn't actually risk allowing unsupported deposits to be filled. Subsequent changes to introduce additional filtering in the `Relayer` class might not have realised that the filtering was ignored, and bugs with actual consequences may have crept in as a result.

nicholaspai: ah

nicholaspai (approved):
LGTM

PR #842, opened 2023-07-26
feat: Consider block timestamps when evaluating deposits

pxrl:
Ensure that the relationship between the deposit timestamp and the destination chain timestamp also meets the MDCs criteria.

Closes ACX-1328


pxrl:
nb. This PR depends on bumping the SDK, so tests will fail until that happens.

src/relayer/Relayer.ts
```diff
@@ -85,6 +86,13 @@ export class Relayer {
       .filter((x) => x !== "default")
       .sort((x, y) => Number(x) - Number(y));
 
+    const avgBlockTimes = Object.fromEntries(
+      await sdkUtils.mapAsync(Object.values(spokePoolClients), async ({ chainId }) => [
+        chainId,
+        await umaUtils.averageBlockTimeSeconds(chainId),
```
pxrl, line 92: nb. This won't work well for zksync because the average block times are hardcoded in the UMA repo.

I have a draft to handle this dynamically in sdk-v2: https://github.com/across-protocol/sdk-v2/compare/master...pxrl/blockUtils

src/relayer/Relayer.ts
```diff
@@ -105,24 +115,27 @@ export class Relayer {
...
       .filter((x) => {
+        const { originChainId, destinationChainId, quoteTimestamp, blockNumber, blockTimestamp } = x.deposit;
+        const { blocks: mdcBlocks, seconds: mdcSeconds } = mdcPerChain[originChainId];
         return (
-          x.deposit.quoteTimestamp + config.quoteTimeBuffer <= hubPoolClient.currentTime &&
-          x.deposit.blockNumber <=
-            spokePoolClients[x.deposit.originChainId].latestBlockNumber - mdcPerChain[x.deposit.originChainId]
+          hubPoolClient.currentTime >= quoteTimestamp + config.quoteTimeBuffer &&
+          blockNumber <= spokePoolClients[originChainId].latestBlockNumber - mdcBlocks &&
+          spokePoolClients[destinationChainId].getCurrentTime() >= blockTimestamp + mdcSeconds
```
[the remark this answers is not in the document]
nicholaspai: Can we instead remove this check for the UBA if we force the UBAclient to set `blockTimestamp` for all fills +300 seconds ? This way the buffer gets built in directly into the data structure and doesn't require as much the relayer to add this buffer

nicholaspai (commented):
One high level comment as to the importance of this PR given https://linear.app/uma/issue/ACX-1352/add-buffer-to-blocktimestamp-for-fills

nicholaspai (commented):
This makes sense to me as a safety measure. Two things:
- Can we add a specific unit test for this logic somewhere?
- How would this interplay with this potential future feature [here](https://github.com/across-protocol/relayer-v2/pull/842/files#r1276492009)?

src/relayer/Relayer.ts
```diff
@@ -1,6 +1,7 @@
 import { constants as ethersConstants } from "ethers";
 import { groupBy } from "lodash";
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { utils as umaUtils } from "@uma/sdk";
```
james-a-morris, line 4: Were we planning to remove this dependency?

src/relayer/Relayer.ts
```diff
@@ -93,7 +101,9 @@ export class Relayer {
           return toBNWei(_usdThreshold).gte(unfilledAmount);
         });
         // If no thresholds are greater than unfilled amount, then use fallback which should have largest MDCs.
-        return [chainId, config.minDepositConfirmations[usdThreshold ?? "default"][chainId]];
+        const blocks = config.minDepositConfirmations[usdThreshold ?? "default"][chainId];
```
james-a-morris, line 104: In this case, would we want to use `||` for the event that `usdThreshold` was incorrectly configured to be `""`?

test/Relayer.BasicFill.ts
```diff
@@ -147,12 +147,16 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
     await configStore.updateTokenConfig(l1Token.address, defaultTokenConfig);
 
     await updateAllClients();
+
+    // Set the SpokePool time to the provider time. This is done to enable the block utility time finder identify a
+    // "reasonable" block number based off the block time when looking at quote timestamps. The timestame at the
+    // destination SpokePool must be _at least_ the timestamp of the origin SpokePool.
+    const currentTime = await getLastBlockTime(spokePool_1.provider);
+    await spokePool_1.setCurrentTime(currentTime);
+    await spokePool_2.setCurrentTime(currentTime + 5);
```
james-a-morris, line 156: Should we hardcode this somewhere in the constants? I only worry that if this ever becomes an issue it may be hard to find exactly why 5 was chosen & inlined into the code.

james-a-morris (commented):
Left a few nits

pxrl:
Closing this one for now but may revive it in future.

PR #847, opened 2023-07-31
feat(finalizer): Upgrade eth-optimism/sdk version to take advantage of updated cross-chain messenger

nicholaspai:
This PR removes the dependency on a cloned version of the `@eth-optimism/sdk-2.1.3` repo that added a bug fix to handle multiple withdrawal messages in the same L2 transaction post-bedrock.

The `eth-optimism/sdk` as of version 3.1 now includes this logic

Resolves ACX-1203

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,10 @@
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "2.4.0",
-    "@across-protocol/optimism-sdk": "2.1.3",
     "@across-protocol/sdk-v2": "0.15.0",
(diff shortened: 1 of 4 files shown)


mrice32 (approved):
LGTM! Assume the additional `undefined` in the calls is just because the interfaces changed a bit. Otherwise, seems uncontroversial!

nicholaspai:
> LGTM! Assume the additional `undefined` in the calls is just because the interfaces changed a bit. Otherwise, seems uncontroversial!

Exactly

PR #850, opened 2023-08-02
improve(ProviderUtils): Add filtered key to eth_getLogs

nicholaspai:
We've seen QuickNode add in a `transactionLogIndex` field which isn't in the spec and is causing quorum agreement issues


james-a-morris (approved):
LGTM - one nit

PR #851, opened 2023-08-03
refactor(finalizer): Split out per-chain finalization

pxrl:
In preparation for supporting a zkSync finalizer (...and subsequently other variants for each new chain), slice up the existing finalize() function to make it more modular and maintainable. There's no intended change in functionality as a result of this.

This is only a first step; there will be more opportunity for refactoring in follow-up work.


src/finalizer/index.ts
```diff
@@ -40,9 +42,127 @@ export interface Withdrawal {
...
+  assert(chainId === 10, `Chain ID mismatch: ${chainId} != 10`);
+
+  const crossChainMessenger = getOptimismClient(10, signer);
+
+  // Sort tokensBridged events by their age. Submit proofs for recent events, and withdrawals for older events.
+  const earliestBlockToProve = latestBlockToFinalize + 1;
+  const { recentTokensBridgedEvents = [], olderTokensBridgedEvents = [] } = groupBy(
+    spokePoolClient.getTokensBridged(),
+    (e) => (e.blockNumber >= earliestBlockToProve ? "recentTokensBridgedEvents" : "olderTokensBridgedEvents")
+  );
```
pxrl, line 75: One change in implementation was introduced here - use `lodash.groupBy()` to subdivide `TokensBridged` events based on their block number, rather than running two passes of `Array.filter()`.

src/finalizer/index.ts
```diff
@@ -79,94 +205,18 @@ export async function finalize(
...
-      const recentTokensBridgedEvents = tokensBridged.filter((e) => e.blockNumber >= firstBlockToFinalize);
-      const proofs = await multicallOptimismL1Proofs(
-        chainId,
-        recentTokensBridgedEvents,
-        crossChainMessenger,
-        hubPoolClient,
-        logger
-      );
-      finalizationsToBatch.callData.push(...proofs.callData);
-      finalizationsToBatch.optimismL1Proofs.push(...proofs.withdrawals);
```
pxrl, line 148: Just flagging an error here - this has been lost in the refactor. Will follow up with a fix.

src/finalizer/types.ts
```diff
@@ -0,0 +1,21 @@
+import { Wallet } from "ethers";
+import { HubPoolClient, SpokePoolClient } from "../clients";
+import { Multicall2Call } from "../common";
+
+export type Withdrawal = {
+  l2ChainId: number;
+  l1TokenSymbol: string;
+  amount: string;
+  type: "proof" | "withdrawal";
```
pxrl, line 9: nb. This is a bit janky; would probably revisit in future to clean it up, but it allows work to continue in the mean time.

src/finalizer/index.ts
pxrl: Should be fixed here: https://github.com/across-protocol/relayer-v2/pull/851/commits/d3c2676af1cdd16282469782781acbc29e34518b

src/finalizer/types.ts
nicholaspai: THis is fine for the time being. Lots of networks do this two step finalization

nicholaspai (approved):
This LGTM and I ran finalizer locally with the code and it produced same results as prod.

src/finalizer/index.ts
james-a-morris: This is a great use of lodash - +1

src/finalizer/index.ts
```diff
@@ -31,18 +33,121 @@ import {
...
+  const earliestBlockToProve = latestBlockToFinalize + 1;
+  const { recentTokensBridgedEvents = [], olderTokensBridgedEvents = [] } = groupBy(
+    spokePoolClient.getTokensBridged(),
+    (e) => (e.blockNumber >= earliestBlockToProve ? "recentTokensBridgedEvents" : "olderTokensBridgedEvents")
+  );
+
+  // First submit proofs for any newly withdrawn tokens. You can submit proofs for any withdrawals that have been
+  // snapshotted on L1, so it takes roughly 1 hour from the withdrawal time
+  logger.debug({
+    at: "Finalizer",
```
james-a-morris, line 65: Do we want to specify the function we're in?
```suggestion
    at: "Finalizer#optimismFinalizer",
```

src/finalizer/index.ts
```diff
@@ -31,18 +33,121 @@ import {
...
+    .filter((e) => e.blockNumber < latestBlockToFinalize);
+
+  return await multicallArbitrumFinalizations(olderTokensBridgedEvents, signer, hubPoolClient, logger);
+}
+
+const chainFinalizers: { [chainId: number]: ChainFinalizer } = {
+  10: optimismFinalizer,
+  137: polygonFinalizer,
+  42161: arbitrumOneFinalizer,
+};
```
james-a-morris, line 149: Do we still need a Boba finalizer? I'm aware that we've disabled the chain, but is there any rare use-case where this would fail because we don't have Boba but we need it?

james-a-morris (commented):
Left a few comments

src/finalizer/index.ts
nicholaspai: yes please

pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/851/commits/a1887cc9c2bbfb7b0a650af8641f5e80bd177fc3

src/finalizer/index.ts
pxrl: We can probably remove it as part of the Base work, but I didn't want to touch it as part of this work.

PR #852, opened 2023-08-03
improve: bump sdk version & refactor

james-a-morris:
Closes: ACX-1336


src/utils/FillUtils.ts
```diff
@@ -192,7 +192,7 @@ export async function getFillDataForSlowFillFromPreviousRootBundle(
     latestMainnetBlock,
     firstFillForSameDeposit.blockNumber,
     firstFillForSameDeposit.destinationChainId,
-    hubPoolClient.configStoreClient.enabledChainIds
+    hubPoolClient.configStoreClient.getEnabledChains()
```
nicholaspai, line 195:
```suggestion
    hubPoolClient.configStoreClient.getEnabledChains(latestMainnetBlock)
```

test/ConfigStoreClient.ts
```diff
@@ -339,6 +348,10 @@ describe("AcrossConfigStoreClient", async function () {
       ).to.throw(/Could not find MaxL1TokenCount/);
     });
     it("Get disabled chain IDs for block range", async function () {
+      // set all possible chains for the next several tests
+      const allPossibleChains = [1, 19, 21, 23];
```
nicholaspai, line 352: Why do you need to inject this int this test, shouldn't it just read directly from the `configStore`?

test/ConfigStoreClient.ts
```diff
@@ -347,66 +360,57 @@ describe("AcrossConfigStoreClient", async function () {
...
+      expect(configStoreClient.getEnabledChainsInBlockRange(0, events[2].blockNumber)).to.deep.equal(allPossibleChains);
 
       // When calling with no to block, returns all enabled chains at from block.
-      expect(configStoreClient.getEnabledChainsInBlockRange(0, undefined, allPossibleChains)).to.deep.equal(
-        allPossibleChains
-      );
+      expect(configStoreClient.getEnabledChainsInBlockRange(0, undefined)).to.deep.equal(allPossibleChains);
+
+      // Expect that calling with no available chains returns an empty array.
+      configStoreClient.setAvailableChains([]);
```
nicholaspai, line 375: This is a weird test to me. It's testing the MockConfigStoreClient, right?  As opposed to the actual ConfigStoreClient we care about?

nicholaspai (approved):
This LGTM, as discussed offline i think all its missing are some unit tests for the new SDK imported function you added most recently `getChainIdIndicesForBlock`

james-a-morris:
@nicholaspai this should resolve the outstanding need for the addtl tests. However, I'm waiting on this PR ( https://github.com/across-protocol/sdk-v2/pull/355 ) to get the tests to pass as I've encoded the new protocol contracts as a constant

test/ConfigStoreClient.ts
```diff
@@ -57,13 +58,78 @@ describe("AcrossConfigStoreClient", async function () {
...
+    // We should now expect that after event two, we have a set of chain IDs
+    // that equals the protocol defaults + 100. This is because the second
+    // event is invalid and should be ignored.
+    expect(configStoreClient.getChainIdIndicesForBlock(eventTwo.blockNumber)).to.deep.equal([
+      1, 10, 137, 288, 42161, 100,
+    ]);
+
+    // We can also check that the chain IDs haven't changed after event 3 because
+    // event three is invalid as it contains duplicates.
+    expect(configStoreClient.getChainIdIndicesForBlock(eventThree.blockNumber)).to.deep.equal([
```
nicholaspai, line 113: This is a great set of tests!

nicholaspai (changes requested):
I'm just realizing this behavior will be overrided unless you remove `chainIdListIndices` from the common/Config.ts file

Additionally, this PR from the SDK needs to be incorporated: https://github.com/across-protocol/sdk-v2/pull/356

src/common/Config.ts
```diff
@@ -33,17 +32,10 @@ export class CommonConfig {
...
-      CHAIN_ID_LIST_OVERRIDE,
       ACROSS_BOT_VERSION,
       ACROSS_MAX_CONFIG_VERSION,
     } = env;
 
-    // This should really not be changed unless we're on Testnet as a lot depends on the order of these
-    // indices. If anything, add chains to this list.
-    this.chainIdListIndices = CHAIN_ID_LIST_OVERRIDE
-      ? JSON.parse(CHAIN_ID_LIST_OVERRIDE)
-      : Constants.CHAIN_ID_LIST_INDICES;
```
pxrl, line 45: It'd be nice to retain some way of overriding the chain ID list via the environment. Otherwise it's a bit more difficult to test new chains before they are added to the global config.

src/common/Config.ts
```diff
@@ -55,16 +47,11 @@ export class CommonConfig {
...
-      for (const chainId of this.chainIdListIndices) {
-        assert(
-          Object.keys(this.blockRangeEndBlockBuffer).includes(chainId.toString()),
-          "BLOCK_RANGE_END_BLOCK_BUFFER missing networks"
-        );
-      }
-    }
 
-    for (const chainId of Constants.CHAIN_ID_LIST_INDICES) {
+    // Min deposit confirmations seems like the most likely constant to have all possible chain IDs listed.
```
pxrl, line 51: This is probably OK for now, but as the bots become more dynamic I think we'll ultimately need to split our verification into pre- and post update stages.

src/finalizer/index.ts
```diff
@@ -268,7 +268,9 @@ export async function runFinalizer(_logger: winston.Logger, baseSigner: Wallet):
           commonClients.hubSigner,
           commonClients.hubPoolClient,
           spokePoolClients,
-          config.finalizerChains
+          process.env.FINALIZER_CHAINS
+            ? JSON.parse(process.env.FINALIZER_CHAINS)
+            : commonClients.configStoreClient.getChainIdIndicesForBlock()
```
pxrl, line 273: In general I'm not keen on reading from `process.env` directly in application code, although sometimes it's difficult or impractical to avoid. In this case I think it'd be preferable to allocate this to a new `FinalizerConfig.ts` file, mirroring the approach for the relayer and dataworker.

It is however a bit unrelated to ACX-1336, so we could consider doing it that separately.

src/monitor/MonitorConfig.ts
```diff
@@ -163,7 +163,9 @@ export class MonitorConfig extends CommonConfig {
       );
     }
 
-    this.chainIdListIndices.forEach((chainId) => {
+    // Min deposit confirmations seems like the most likely constant to have all possible chain IDs listed.
+    const allPossibleChainIds = Object.keys(MIN_DEPOSIT_CONFIRMATIONS).map((chainId) => Number(chainId));
```
pxrl, line 167: Rather than recreating from MIN_DEPOSIT_CONFIRMATIONS, I think it'd be better to set `this.allPossibleChainIds` in the `CommonConfig` class, and then everyone else can read from it. It doesn't really make sense to me that the monitor should know anything about MDCs.

src/relayer/RelayerConfig.ts
```diff
@@ -102,8 +102,8 @@ export class RelayerConfig extends CommonConfig {
     (this.minDepositConfirmations = MIN_DEPOSIT_CONFIRMATIONS
       ? JSON.parse(MIN_DEPOSIT_CONFIRMATIONS)
       : Constants.MIN_DEPOSIT_CONFIRMATIONS),
-      this.chainIdListIndices.forEach((chainId) => {
-        Object.keys(this.minDepositConfirmations).forEach((threshold) => {
+      Object.keys(this.minDepositConfirmations).forEach((threshold) => {
+        Object.keys(this.minDepositConfirmations[threshold]).forEach((chainId) => {
```
pxrl, line 106: Same as previous - I think it's safer and more maintainable to infer and store `allPossibleChainIds` centrally.

test/Dataworker.loadData.ts
```diff
@@ -72,18 +73,9 @@ describe("Dataworker: Load data used in all functions", async function () {
...
-      bundleDataClient.loadData(getDefaultBlockRange(0), spokePoolClients),
-      "ConfigStoreClient not updated"
-    );
     await configStoreClient.update();
 
     // Throws error if hub pool client is not updated.
-    await assertPromiseError(
-      bundleDataClient.loadData(getDefaultBlockRange(0), spokePoolClients),
-      "HubPoolClient not updated"
-    );
```
james-a-morris, line 86: These two tests can now be removed as the hubpool/config store is essentially required to be updated

test/ConfigStoreClient.ts
```diff
@@ -57,13 +58,78 @@ describe("AcrossConfigStoreClient", async function () {
...
+    await configStoreClient.update();
+
+    // Sanity check to verify that the config store client is updated
+    expect(configStoreClient.isUpdated).to.be.true;
+
+    // Next, we can test the case where we submit an additional chain ID to our
+    // list of chain IDs.
+    const eventOne = await configStore.updateGlobalConfig(
+      utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.CHAIN_ID_INDICES),
+      JSON.stringify([1, 10, 137, 288, 42161, 100])
```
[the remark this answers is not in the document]
james-a-morris: My only worry about this is that I wanted to keep things explicit because we introduce erroneous configs in later tests

src/finalizer/index.ts
james-a-morris: @pxrl Should we make this a separate linear issue?

src/monitor/MonitorConfig.ts
nicholaspai: We'll fix this in #ACX1390

src/finalizer/index.ts
nicholaspai: We'll fix this in #ACX1390

test/ConfigStoreClient.ts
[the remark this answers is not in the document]
james-a-morris: Good call - I've modified for better readability

test/UBAClientUtilities.ts
```diff
@@ -82,15 +82,16 @@ describe("UBAClientUtilities", function () {
...
     return await publishValidatedBundles(
       chainIds,
       l1Tokens,
       hubPoolClient,
       spokePoolClients,
       bundleCount,
       runningBalances,
-      incentiveBalances
+      incentiveBalances,
+      1000
```
james-a-morris, line 94: By increasing the random jump vector, we effectively make the chance of block collisions much smaller

test/utils/HubPoolUtils.ts
```diff
@@ -79,6 +79,18 @@ export async function publishValidatedBundles(
...
+  // block ranges are invalid for the case of testing purposes. If we find a `start` block
+  // that is equal to or greater than the `end` block, we will set the `end` block to be
+  // equal to the `start` block + 1.
+  Object.values(expectedBlockRanges).forEach((blockRanges) => {
+    blockRanges.forEach((blockRange) => {
+      if (blockRange.start >= blockRange.end) {
+        blockRange.end = blockRange.start + 1;
+      }
+    });
+  });
```
james-a-morris, line 92: The motivation here is that we can further enforce that any random errors can be accounted for

src/relayer/RelayerConfig.ts
nicholaspai: https://github.com/across-protocol/relayer-v2/pull/858

test/utils/BlockchainUtils.ts
```diff
@@ -1,7 +1,11 @@
...
-export async function mineRandomBlocks(): Promise<void> {
-  const randomBlocksToMine = Math.ceil(Math.random() * 10);
+/**
+ * Mines a random number of blocks.
+ * @param amount Amount of blocks to mine. If not provided, a random number of blocks will be mined.
+ */
+export async function mineRandomBlocks(amount?: number): Promise<void> {
+  const randomBlocksToMine = amount ?? Math.ceil(Math.random() * 10);
   for (let i = 0; i < randomBlocksToMine; i++) {
     await hre.network.provider.send("evm_mine");
```
pxrl, line 10: I'm not sure whether these requests are always synchronous - in my experience it's more like sending a request to hre, and have definitely seen that can return without producing a new block. If you want to be sure that the block is actually mined, I'd recommend periodically issuing `evm_mine` requests until the expected block number is actually produced.

FWIW I had to do something similar here: https://github.com/across-protocol/relayer-v2/blob/2727ef9f298abc9515206f22b990b122c4323e8e/test/Relayer.RefundRequests.ts#L54

test/UBAClientUtilities.ts
nicholaspai: Can you add this comment above the hardcoded 1000 just so we know what's going on if this breaks in a future test?

nicholaspai (approved):
Just one comment LGTM

test/UBAClientUtilities.ts
james-a-morris: Done:

```
      // We can override the jump block to be 10_000. By increasing the random jump vector, we
      // can minimize the chance of block collisions that would cause this function to
      // nondeterministically fail in the CI. NOTE: a failure as a result of incorrect
      // block ranges isn't necessarily a failure of the test, but of the randomness that
      // is generated in the test environment to simulate the block ranges.
      10_000
```

test/utils/BlockchainUtils.ts
james-a-morris: Good call - I think is more-or-less just used to set larger numbers than waiting the random 0 to 10 rate

PR #853, opened 2023-08-05
feat: encode sim mode to dataworker

james-a-morris:
This PR adds an explicit ENV flag to the dataworker so that it can be run safely in simulation mode.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -69,6 +69,13 @@ NODE_MAX_CONCURRENCY=25
 # recommended to test with this before setting SEND_RELAYS to "true".
 SEND_RELAYS=false
 
+# Permit the Dataworker from submitting transactions to the RPC provider. This
+# is disabled by default and must be explicitly enabled for the dataworker to
+# execute the transactions that it generates. When SEND_DATAWORKER_TXS is not
+# set to "true", the bot will simulate making fills and will log the results,
+# but will not submit transactions to the RPC provider. It is recommended to
+# test with this before setting SEND_DATAWORKER_TXS to "true".
+SEND_DATAWORKER_TXS=false
 
 # List of destination chains to be supported by the relayer. If set to a
 # non-empty list, only transfers going to these chains will be filled. For
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -162,7 +162,11 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
+      // Execute a series of transactions to reflect the state changes from the above operations.
+      // This is done only if the sendingDataworkerTxsEnabled flag is set to true.
+      if (config.transactionExecutionsEnabled) {
       await clients.multiCallerClient.executeTransactionQueue();
+      }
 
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -69,6 +69,13 @@ NODE_MAX_CONCURRENCY=25
 # recommended to test with this before setting SEND_RELAYS to "true".
 SEND_RELAYS=false
 
+# Permit the Dataworker from submitting transactions to the RPC provider. This
+# is disabled by default and must be explicitly enabled for the dataworker to
+# execute the transactions that it generates. When SEND_DATAWORKER_TXS is not
+# set to "true", the bot will simulate making fills and will log the results,
+# but will not submit transactions to the RPC provider. It is recommended to
+# test with this before setting SEND_DATAWORKER_TXS to "true".
+SEND_DATAWORKER_TXS=false
```
james-a-morris, line 78: Flagging this for discussion: Should we keep the previous implementation where sending txns is essentially opt-out

nicholaspai (commented):
Note we can't merge this without a config change to the prod bots otherwise they won't send anything

james-a-morris:
> Note we can't merge this without a config change to the prod bots otherwise they won't send anything

I agree - I left that in a comment in the event that we'd want to keep the initial functionality & make sim mode opt-in

https://github.com/across-protocol/relayer-v2/pull/853#discussion_r1285058878

james-a-morris:
@nicholaspai Added the do-not-merge label to this PR until resolved. 

In my opinion the opt-in approach is more consistent with the other relayer-v2 operations. If we'd prefer to have sim mode be the `opt-in` variant of the DW, this is quickly done by inverting the conditional & changing the ENV name. It's just a matter of preference/consistency.

.env.example
pxrl: Not sure if I am understanding your point correctly, but the current config is already opt-in: the config must explicitly be set to "true" for it to be enabled.

```
    this.sendingDisputesEnabled = SEND_DISPUTES === "true";
    this.sendingProposalsEnabled = SEND_PROPOSALS === "true";
    this.sendingExecutionsEnabled = SEND_EXECUTIONS === "true";
```

Also, I think we should simplify this even further, and just use the existing `SEND_TRANSACTIONS`. That's a global config item that'll also work fine for the relayer.

src/dataworker/index.ts
```diff
@@ -162,7 +162,11 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
-      await clients.multiCallerClient.executeTransactionQueue();
+      // Execute a series of transactions to reflect the state changes from the above operations.
+      // This is done only if the sendingDataworkerTxsEnabled flag is set to true.
+      if (config.transactionExecutionsEnabled) {
+        await clients.multiCallerClient.executeTransactionQueue();
+      }
```
pxrl, line 169: If you pass in the argument then it'll toggle simulation mode in the `MultiCallerClient`. This is what is done on the relayer side.

```suggestion
        await clients.multiCallerClient.executeTransactionQueue(!config.transactionExecutionsEnabled);
```

.env.example
nicholaspai: agreed with paul

PR #854, opened 2023-08-07
feat: Script for interacting with SpokePools

pxrl:
This is useful for various SpokePool-related activities:
- Make a deposit
- Retrieve information about deposit(s) and/or fill(s) from a transaction.
- Dump basic SpokePool configuration information.

This is especially handy with the new zkSync SpokePool, because zkSync's block explorer doesn't yet support making transactions via proxy contracts, and doesn't decode Across SpokePool events properly either.

Example:

```
# Deposit .001 WETH on Ethereum Görli, destined for zkSync era Görli. 
$ yarn ts-node ./scripts/spokepool.ts deposit --wallet mnemonic --from 5 --to 280 --token weth --amount 0.001
```


pxrl:
Example output:

```
Submitting deposit on chain ID 5
        originChainId       : 5
        destinationChainId  : 280
        depositor           : 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D
        recipient           : 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D
        relayerFeePct       : 0
        quoteTimestamp      : 1691411700

Submitted deposit:
        depositId           : 100019
        transactionHash     : 0x1fa909fea87e41c865555559719ff25a5e6c3cd4e1af61da7d4ed02c6f3f3322

Done in 25.11s.
```

scripts/spokepool.ts
```diff
@@ -0,0 +1,183 @@
...
+  if (address === undefined) {
+    throw new Error(`Token ${symbol} not supported on chain ID ${chainId}`);
+  }
+
+  return { address: addresses[chainId], decimals, symbol };
+}
+
+async function getHubPoolContract(chainId: number): Promise<Contract> {
+  const contractName = "HubPool";
+  const hubPoolChainId = chains.includes(chainId) ? 1 : 5;
```
james-a-morris, line 35: We make this distinction in multiple locations throughout the repo - would it make sense to put this into a util function `resolveHubPoolChainId()` or something similar?

scripts/spokepool.ts
```diff
@@ -0,0 +1,183 @@
...
+  const provider = new ethers.providers.StaticJsonRpcProvider(getNodeUrlList(hubPoolChainId, 1)[0]);
+  return hubPool.connect(provider);
+}
+
+async function getSpokePoolContract(chainId: number): Promise<Contract> {
+  const hubPool = await getHubPoolContract(chainId);
+  const spokePoolAddr = (await hubPool.crossChainContracts(chainId))[1];
+
+  const contract = new Contract(spokePoolAddr, contracts.SpokePool__factory.abi);
+  return contract;
```
james-a-morris, line 50: IIUC we can use 

```suggestion
  const hubPool = await getHubPoolContract(chainId);
  const spokePoolAddr = (await hubPool.crossChainContracts(chainId))[1];

  return SpokePool__factory.connect(spokePoolAddr);
```

scripts/spokepool.ts
```diff
@@ -0,0 +1,183 @@
...
+  const tokenSymbol = (args.token as string).toUpperCase();
+
+  const provider = new ethers.providers.StaticJsonRpcProvider(getNodeUrlList(fromChainId, 1)[0]);
+  signer = signer.connect(provider);
+  const spokePool = (await getSpokePoolContract(fromChainId)).connect(signer);
+
+  const token = getTokenAddress(tokenSymbol, fromChainId);
+  const amount = ethers.utils.parseUnits(baseAmount.toString(), token.decimals);
+
+  const erc20 = new Contract(token.address, contracts.ExpandedERC20__factory.abi, signer);
```
james-a-morris, line 68:
```suggestion
  const erc20 = ExpandedERC20__factory.connect(token.address, signer)
```

james-a-morris (commented):
Left a few nits

scripts/spokepool.ts
pxrl: Yeah that's a nice simplification, although I think we want `.attach()` instead of `.connect()`, since `.connect()` returns another `ContractFactory` instance, whereas `.attach()` returns a `Contract`:

https://docs.ethers.org/v5/single-page/#/v5/api/contract/contract-factory/-%23-ContractFactory-connect
https://docs.ethers.org/v5/single-page/#/v5/api/contract/contract-factory/-%23-ContractFactory-attach

pxrl:
> High level comment: can we make this script's name more clear (or do you intend to add other helper SpokePool related functions in here?) or add a helpful comment at the top explaining why this would be used?

Generally yes - I'd like to support something like the following:

`spokepool.ts deposit <deposit-args>`
`spokepool.ts fill --origin <chainId> --hash <depositHash>`
`spokepool.ts request-refund <refund-args>`

I left a placeholder in the code for adding fill support, but I didn't want to spend any time on actually implementing it now given other priorities: https://github.com/across-protocol/relayer-v2/blob/4b1bb8a633f87a6f34ee2602cf217d3a80a29c45/scripts/spokepool.ts#L165

As general background, it's occurred to me that I miss a good set of cmdline utilities for interacting with our contracts, since it should be possible to execute (or query) standard operations against them without a block explorer (a la zkSync at the moment), and without having to modify code or .env configs.

nicholaspai:
> > High level comment: can we make this script's name more clear (or do you intend to add other helper SpokePool related functions in here?) or add a helpful comment at the top explaining why this would be used?
> 
> Generally yes - I'd like to support something like the following:
> 
> `spokepool.ts deposit <deposit-args>` `spokepool.ts fill --origin <chainId> --hash <depositHash>` `spokepool.ts request-refund <refund-args>`
> 
> I left a placeholder in the code for adding fill support, but I didn't want to spend any time on actually implementing it now given other priorities:
> 
> https://github.com/across-protocol/relayer-v2/blob/4b1bb8a633f87a6f34ee2602cf217d3a80a29c45/scripts/spokepool.ts#L165
> 
> As general background, it's occurred to me that I miss a good set of cmdline utilities for interacting with our contracts, since it should be possible to execute (or query) standard operations against them without a block explorer (a la zkSync at the moment), and without having to modify code or .env configs.

This would be a good interface yes

scripts/spokepool.ts
pxrl: I'm not sure whether this is the same use case. This utility is trying to infer the correct HubPool based on the chainId of a SpokePool. In the main codebase we just read `HUB_CHAIN` from the environment, or fetch the chain ID out of the HubPoolClient, so there shouldn't be any doubt about which chain the HubPool is on.

The reason for ignoring the env var is that I'd like this utility to be almost config-less - i.e. it shouldn't be necessary to fiddle with `.env` in order to make a deposit on a test chain while I am working on my relayer in prod.

scripts/spokepool.ts
```diff
@@ -0,0 +1,236 @@
...
+  ["MNEMONIC", "PRIVATE_KEY"].forEach((envVar) => (process.env[envVar] = ""));
+
+  switch (argv[0]) {
+    case "deposit":
+      return await deposit(args, signer);
+    case "dump":
+      return await dumpConfig(args, signer);
+    case "fill":
+      // @todo Not supported yet...
+      usage(); // no return
```
james-a-morris, line 219: We should add a break here for consistency otherwise it will call usage twice
```suggestion
      // @todo Not supported yet...
      usage(); // no return
      break;
```

scripts/spokepool.ts
```diff
@@ -0,0 +1,236 @@
...
+  switch (args.wallet) {
+    case "mnemonic":
+      signer = Wallet.fromMnemonic(process.env.MNEMONIC);
+      break;
+    case "privateKey":
+      signer = new Wallet(process.env.PRIVATE_KEY);
+      break;
+    default:
+      usage(args.wallet); // no return
+  }
```
james-a-morris, line 209: Could we use `getSigner()` in `src/utils/SignerUtils.ts` to re-use logic?

james-a-morris (commented):
Two more comments

scripts/spokepool.ts
pxrl: It won't call `usage()` because `usage()` never returns ;)

src/utils/AddressUtils.ts
```diff
@@ -29,11 +29,23 @@ export function compareAddressesSimple(addressA: string, addressB: string): bool
...
+ * @returns The number of ERC20 decimals configured for the requested token.
+ */
+export function resolveTokenDecimals(tokenSymbol: string): number {
+  const decimals = TOKEN_SYMBOLS_MAP[tokenSymbol]?.decimals;
+  if (decimals === undefined) {
+    throw new Error(`Unrecognized token symbol: ${tokenSymbol}`);
+  }
+  return decimals;
+}
+
```
james-a-morris, line 48:
```suggestion

/**
```

scripts/spokepool.ts
pxrl: Ended up having to do this anyway because the linter doesn't recognise that usage() never returns.

scripts/spokepool.ts
```diff
@@ -0,0 +1,352 @@
...
+const { MaxUint256, Zero } = ethers.constants;
+const { isAddress } = ethers.utils;
+
+const testChains = [5, 280];
+const chains = [1, 10, 137, 324, 8453, 42161];
+
+const padLeft = 20;
+const padRight = 25;
+
+function formatAddress(address: string, maxWidth = 18): string {
```
mrice32, line 25: Why not just display full addresses?

scripts/spokepool.ts
```diff
@@ -0,0 +1,352 @@
...
+  const contract = new Contract(spokePoolAddr, contracts.SpokePool__factory.abi);
+  return contract;
+}
+
+async function deposit(args: Record<string, number | string>, signer: Wallet): Promise<boolean> {
+  const depositor = await signer.getAddress();
+  const fromChainId = Number(args.from);
+  const toChainId = Number(args.to);
+  const recipient = args.recipient ?? depositor;
+  const baseAmount = Number(args.amount);
```
mrice32, line 137: Should we validate args here and throw if they aren't provided as expected?

scripts/spokepool.ts
```diff
@@ -0,0 +1,352 @@
...
+      `\t${"Cross-domain alias".padEnd(padLeft)}: ${adminAlias}\n` +
+      `\t${"WETH".padEnd(padLeft)}: ${wethAddress}\n` +
+      `\t${"Current time".padEnd(padLeft)}: ${currentTimeStr} (${currentTime})\n`
+  );
+
+  return true;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+async function fetchTxn(args: Record<string, number | string>, _signer: Wallet): Promise<boolean> {
```
mrice32, line 237: +1 great function

mrice32 (approved):
Great place to start! Always todos and things to add/change, but this is great!

scripts/spokepool.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/854/commits/8c1c29d59e301a349959898ce8193a5111535b4a

scripts/spokepool.ts
pxrl: Unfortunately this doesn't work as intended - see https://github.com/across-protocol/relayer-v2/pull/854#discussion_r1286127759

scripts/spokepool.ts
pxrl: Maybe this is just my OCD, but it looks really awkward to have a long address alongside a bunch of significantly shorter values. Some of the values look best right-justified (especially fields with a numerical relationship to each other - i.e. relayerFeePct and amount. I was thinking about adding a `-v` flag to allow the caller to force the address to be displayed in full.

scripts/spokepool.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/854/commits/ea24ac168f4334328c41cf1592eb40c0ad6ae4a5

scripts/spokepool.ts
pxrl: I ended up scrapping the custom formatting and reverted to "raw" dumping of the event. This removes some of the (imo) nice features like grouping related fields, which the SpokePool event data doesn't always do. However, it's very simple and generic, and ensures that all the fields are dumped. Maybe in future I'll introduce a way of ordering the fields, but that's not worth prioritising now.

https://github.com/across-protocol/relayer-v2/pull/854/commits/caccd7b4391479f5cedc1de052074fdf526776fd

PR #856, opened 2023-08-07
chore: Additional defaults for zkSync

pxrl:


src/common/Constants.ts
```diff
@@ -27,6 +27,7 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
   10: 1382400,
   137: 138240,
   288: 11520,
+  324: 4 * 24 * 60 * 60,
```
pxrl, line 30: fwiw I wouldn't mind applying this style to the other entries here.

nicholaspai (approved):

PR #857, opened 2023-08-07
improve(finalizer): Dynamic HubPool chain ID

pxrl:
Removes a barrier for running the finalizer on a testnet.


james-a-morris (approved):
Good call making this dynamic

PR #858, opened 2023-08-07
feat: add validation function to Config class

nicholaspai:
This way we can validate configs that are set per-chain once we know all chains to validate using the config store client's chain ID's fetched from on-chain

Fixes ACX-1390

### DIFF
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -90,7 +90,8 @@ export class CommonConfig {
       if (Object.keys(this.maxBlockLookBack).length > 0) {
         assert(
           Object.keys(this.maxBlockLookBack).includes(chainId.toString()),
-          "MAX_BLOCK_LOOK_BACK is missing chainId ${chainId}");
+          "MAX_BLOCK_LOOK_BACK is missing chainId ${chainId}"
+        );
       }
 
       // Multicall chunk size precedence: Environment, chain-specific config, global default.
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -70,7 +70,7 @@ export class CommonConfig {
    * @dev This should be called by passing in the latest chain ID indices from an updated ConfigStoreClient.
    * @param chainIdIndices All expected chain ID's that could be supported by this config.
    */
-  loadAndValidateConfigForChains(chainIdIndices: number[]) {
+  loadAndValidateConfigForChains(chainIdIndices: number[]): void {
     const multiCallChunkSize: { [chainId: number]: number } = {};
 
     for (const chainId of chainIdIndices) {
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -88,7 +88,9 @@ export class CommonConfig {
 
       // Validate that there is a max block look back for each chain.
       if (Object.keys(this.maxBlockLookBack).length > 0) {
-        assert(Object.keys(this.maxBlockLookBack).includes(chainId.toString()), "MAX_BLOCK_LOOK_BACK missing networks");
+        assert(
+          Object.keys(this.maxBlockLookBack).includes(chainId.toString()),
+          "MAX_BLOCK_LOOK_BACK is missing chainId ${chainId}");
       }
 
       // Multicall chunk size precedence: Environment, chain-specific config, global default.
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -15,11 +15,13 @@ export class CommonConfig {
   readonly sendingTransactionsEnabled: boolean;
   readonly bundleRefundLookback: number;
   readonly maxRelayerLookBack: number;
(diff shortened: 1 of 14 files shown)


src/scripts/validateRootBundle.ts
```diff
@@ -72,6 +72,7 @@ export async function validate(_logger: winston.Logger, baseSigner: Wallet): Pro
   // same block.
   const dvm = await getDvmContract(clients.configStoreClient.configStore.provider);
   await updateDataworkerClients(clients, false);
```
james-a-morris, line 74: Can we nest this call inside of the `updateClients` & `updateDataworkerClients`?

james-a-morris (commented):
I think we could probably get away with a lot less entry points to `config.loadAndValidateConfigForChains` if we called it within `updateClients` & `updateDataworkerClients`. It would also have the benefit of being called everytime we call these functions and enforce that it wouldn't be forgotten.

Alternatively, we could embed this logic into the clients themselves so that after every update everything can be verified on a per-client level.

src/scripts/validateRootBundle.ts
nicholaspai: yeah good call. I didn't want to have to add `config` to the list of function params but its a lot cleaner this way and less diff

src/common/Config.ts
```diff
@@ -71,20 +62,49 @@ export class CommonConfig {
...
-        // prettier-ignore
-        const chunkSize = Number(
-          process.env[`MULTICALL_CHUNK_SIZE_CHAIN_${chainId}`]
-            ?? DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE[chainId]
-            ?? DEFAULT_MULTICALL_CHUNK_SIZE
+  /**
+   * @notice Loads additional configuration state that can only be known after we know all chains that we're going to
+   * support. Throws an error if any of the configurations are not valid.
+   * @dev This should be called by passing in the latest chain ID indices from an updated ConfigStoreClient.
+   * @param chainIdIndices All expected chain ID's that could be supported by this config.
```
james-a-morris, line 71: Nit: we should document when/how this function will throw

nicholaspai: done

src/common/Config.ts
```diff
@@ -71,20 +62,49 @@ export class CommonConfig {
...
+   */
+  loadAndValidateConfigForChains(chainIdIndices: number[]): void {
```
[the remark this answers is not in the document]
nicholaspai: it loads and asserts

james-a-morris (approved):
LGTM pending passing tests

src/common/Config.ts
```diff
@@ -71,20 +62,53 @@ export class CommonConfig {
...
+    const multiCallChunkSize: { [chainId: number]: number } = {};
+
+    for (const chainId of chainIdIndices) {
+      // Validate that there is a block range end block buffer for each chain.
+      if (Object.keys(this.blockRangeEndBlockBuffer).length > 0) {
+        assert(
+          Object.keys(this.blockRangeEndBlockBuffer).includes(chainId.toString()),
+          "BLOCK_RANGE_END_BLOCK_BUFFER missing networks"
```
pxrl, line 85:
```suggestion
          `BLOCK_RANGE_END_BLOCK_BUFFER is missing chainId ${chainId}`
```

src/common/Config.ts
```diff
@@ -71,20 +62,53 @@ export class CommonConfig {
...
         );
-        assert(chunkSize > 0, `Chain ${chainId} multicall chunk size (${chunkSize}) must be greater than 0`);
-        return [chainId, chunkSize];
-      })
+      }
+
+      // Validate that there is a max block look back for each chain.
+      if (Object.keys(this.maxBlockLookBack).length > 0) {
+        assert(Object.keys(this.maxBlockLookBack).includes(chainId.toString()), "MAX_BLOCK_LOOK_BACK missing networks");
```
pxrl, line 91:
```suggestion
        assert(
          Object.keys(this.maxBlockLookBack).includes(chainId.toString()),
          "MAX_BLOCK_LOOK_BACK is missing chainId ${chainId}");
```

src/common/Config.ts
```diff
@@ -71,20 +62,53 @@ export class CommonConfig {
...
+      }
+
+      // Multicall chunk size precedence: Environment, chain-specific config, global default.
+      // prettier-ignore
+      const chunkSize = Number(
+      process.env[`MULTICALL_CHUNK_SIZE_CHAIN_${chainId}`]
+        ?? DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE[chainId]
+        ?? DEFAULT_MULTICALL_CHUNK_SIZE
     );
```
pxrl, line 100: Indentation seems off here.

PR #859, opened 2023-08-08
fix(dataworker): Add changes found after integration testing Dataworker

nicholaspai:
- We need to update the config store client before constructing BundleDataClient
- We need to handle the fact that BundleDataClient and Dataworker might have a longer chainIdIndices list than the actual number of block ranges they will handle

## Fixes
- Fix logic in `blockRangesAreInvalidForSpokeClients` that assumed that `chainIdIndices` length is always equal to `blockRanges` length

## Bots integration tested:
- [x] Dataworker (propose, validate, execute)
- [x] `validateRunningBalances` script used to make sure there are no excesses or deficits created by subtle dataworker running balance compuation bugs. 
- [x] `validateRootBundle` script used to evaluate an UMA dispute over an Across bundle proposal 
- [x] Monitor
- [x] Relayer

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -52,14 +52,7 @@ export class BundleDataClient {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly chainIdListForBundleEvaluationBlockNumbers: number[],
     readonly blockRangeEndBlockBuffer: { [chainId: number]: number } = {}
-  ) {
-    logger.debug({
-      at: "BundleDataClient",
-      message: "Initialized BundleDataClient",
-      chainIdListForBundleEvaluationBlockNumbers,
-      spokePoolClientChains: Object.keys(spokePoolClients),
-    });
-  }
+  ) {}
 
   // This should be called whenever it's possible that the loadData information for a block range could have changed.
   // For instance, if the spoke or hub clients have been updated, it probably makes sense to clear this to be safe.
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -52,7 +52,14 @@ export class BundleDataClient {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     readonly chainIdListForBundleEvaluationBlockNumbers: number[],
     readonly blockRangeEndBlockBuffer: { [chainId: number]: number } = {}
-  ) {}
+  ) {
+    logger.debug({
+      at: "BundleDataClient",
+      message: "Initialized BundleDataClient",
+      chainIdListForBundleEvaluationBlockNumbers,
+      spokePoolClientChains: Object.keys(spokePoolClients),
+    });
+  }
 
   // This should be called whenever it's possible that the loadData information for a block range could have changed.
   // For instance, if the spoke or hub clients have been updated, it probably makes sense to clear this to be safe.
@@ -210,9 +217,9 @@ export class BundleDataClient {
       throw new Error("HubPoolClient not updated");
     }
 
-    if (blockRangesForChains.length !== this.chainIdListForBundleEvaluationBlockNumbers.length) {
+    if (blockRangesForChains.length > this.chainIdListForBundleEvaluationBlockNumbers.length) {
       throw new Error(
-        `Unexpected block range list length of ${blockRangesForChains.length}, should be ${this.chainIdListForBundleEvaluationBlockNumbers.length}`
+        `Unexpected block range list length of ${blockRangesForChains.length}, should be <= ${this.chainIdListForBundleEvaluationBlockNumbers.length}`
       );
     }
 
@@ -298,7 +305,10 @@ export class BundleDataClient {
       return isChainDisabled(blockRangeForChain);
     };
 
-    const allChainIds = Object.keys(spokePoolClients).map(Number);
+    // Infer chain ID's to load from number of block ranges passed in.
+    const allChainIds = blockRangesForChains.map(
+      (_blockRange, index) => this.chainIdListForBundleEvaluationBlockNumbers[index]
+    );
 
     const validateRefundRequestAndSaveData = async (refundRequest: RefundRequestWithBlock): Promise<void> => {
       const result = await refundRequestIsValid(spokePoolClients, this.clients.hubPoolClient, refundRequest);
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -35,7 +35,6 @@ import { SpokePoolClient, UBAClient } from "../clients";
 import * as PoolRebalanceUtils from "./PoolRebalanceUtils";
 import {
   blockRangesAreInvalidForSpokeClients,
-  getBlockForChain,
   getBlockRangeForChain,
   getImpliedBundleBlockRanges,
 } from "../dataworker/DataworkerUtils";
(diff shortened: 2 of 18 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -298,7 +298,10 @@ export class BundleDataClient {
       return isChainDisabled(blockRangeForChain);
     };
 
-    const allChainIds = Object.keys(spokePoolClients).map(Number);
+    // Infer chain ID's to load from number of block ranges passed in.
```
pxrl, line 301: Can this break if someone makes a malformed proposal?

nicholaspai: We'll sanitize the malformed bundle block numbers before we get here. 
- Check start blocks: https://github.com/across-protocol/relayer-v2/blob/6cb7fc96a9e6dbef41d0c500b0f29e3d42f0f59e/src/dataworker/Dataworker.ts#L1021
- Check end blocks: https://github.com/across-protocol/relayer-v2/blob/6cb7fc96a9e6dbef41d0c500b0f29e3d42f0f59e/src/dataworker/Dataworker.ts#L1049
- Check block range length https://github.com/across-protocol/relayer-v2/blob/6cb7fc96a9e6dbef41d0c500b0f29e3d42f0f59e/src/dataworker/Dataworker.ts#L997

src/monitor/MonitorClientHelper.ts
```diff
@@ -60,6 +54,15 @@ export async function constructMonitorClients(
     [baseSigner.address, ...spokePoolAddresses],
     commonClients.hubPoolClient.hubPool.address
   );
+  const adapterSupportedChains = adapterManager.supportedChains();
+  const spokePoolChains = Object.keys(spokePoolClients)
+    .map((chainId) => Number(chainId))
+    .filter((chainId) => adapterSupportedChains.includes(chainId));
+  // We can only pass in providers for chain and spoke pool chains for chains that are supported in the adapter manager.
```
pxrl, line 61: I might be misreading the comment, but it seems like it's saying more than it intends to?

src/dataworker/Dataworker.ts
```diff
@@ -2289,19 +2291,22 @@ export class Dataworker {
...
+   * set of chains we can return block ranges for.
    * @returns [number, number]: [startBlock, endBlock]
    */
   _getWidestPossibleBlockRangeForNextBundle(
     spokePoolClients: SpokePoolClientsByChain,
     mainnetBundleEndBlock: number
   ): number[][] {
     return PoolRebalanceUtils.getWidestPossibleExpectedBlockRange(
-      this.chainIdListForBundleEvaluationBlockNumbers,
+      // We only want as many block ranges as there are chains enabled at the time of the bundle end block.
```
pxrl, line 2303: This is basically the opposite of the policy we have on the VERSION field. I can see reasons to do it like this, but it also seems like it could get complicated when we start adding future changes. Have you considered only applying the set of chains at the start of the bundle? In this way the proposal cycle would basically act like a clock for config changes to propagate through the system.

src/dataworker/index.ts
```diff
@@ -47,14 +47,18 @@ export async function createDataworker(
 }
 export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet): Promise<void> {
   logger = _logger;
+  let loopStart = Date.now();
   const { clients, config, dataworker } = await createDataworker(logger, baseSigner);
```
pxrl, line 51: Seems like we're unconditionally preventing allowances here. Should this instead be toggled by `SEND_TRANSACTIONS` in the environment?

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -145,9 +142,16 @@ export function getSpokePoolClientEventSearchConfigsForFastDataworker(
     fromBundle === undefined
       ? {}
       : Object.fromEntries(
-          dataworker.chainIdListForBundleEvaluationBlockNumbers.map((chainId) => {
+          dataworker.chainIdListForBundleEvaluationBlockNumbers.map((chainId, i) => {
+            // If block for chainId doesn't exist in bundleEvaluationBlockNumbers, then leave undefined which
+            // will reuslt in querying from the spoke activation block.
```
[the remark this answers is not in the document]
nicholaspai: I believe so. We read the `bundleEvaluationBlockNumbers` length from the published bundle, so if a bundle was proposed without the chain, then the chain is considered "disabled" if its index is >= bundleEvaluationBlockNumbers.length

src/dataworker/index.ts
nicholaspai: good point

src/dataworker/Dataworker.ts
nicholaspai: yeah good call it required a few code changes but this is a good call out. I considered doing this, ultimately don't think it matters too much, but its best to keep it consistent. Do note though that we do use the mainnet bundle end block as the config marker in other places...

nicholaspai: We probably need to move all configs to be marked by the mainnet bundle start block since for the UBA, only the next bundle start blocks are deterministic while the bundle end blocks cannot be predicted. This is important therefore for any config that could effect the UBA computation which includes `CHAIN_ID_INDICES`.

Some of the other configs we don't need to care about like `maxRelayerRefundsPerLeaf`.

src/monitor/MonitorClientHelper.ts
nicholaspai: Fixed in this commit a89f6fada85fbe3ba40d2c8727e9b91ae09f18ea

src/dataworker/Dataworker.ts
nicholaspai: Fixed in this commit a89f6fada85fbe3ba40d2c8727e9b91ae09f18ea

src/dataworker/index.ts
nicholaspai: Fixed in this commit a89f6fada85fbe3ba40d2c8727e9b91ae09f18ea

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -130,9 +119,16 @@ export function getSpokePoolClientEventSearchConfigsForFastDataworker(
...
+          dataworker.chainIdListForBundleEvaluationBlockNumbers.map((chainId, i) => {
+            // If block for chainId doesn't exist in bundleEvaluationBlockNumbers, then leave undefined which will
+            // result in querying until the latest block.
+            if (i >= toBundle.bundleEvaluationBlockNumbers.length) {
+              return [chainId, undefined];
+            }
             return [
               chainId,
+              // TODO: I think this has a bug for disabled chains where we don't want to +1 if the chain is disabled
+              // at the time of this bundle.
```
pxrl, line 131: qq. are the two instances of this showstoppers for merging this PR?

src/monitor/Monitor.ts
```diff
@@ -502,6 +510,11 @@ export class Monitor {
 
     const allL1Tokens = this.clients.hubPoolClient.getL1Tokens();
     for (const chainId of this.monitorChains) {
+      // If chain wasn't active in latest bundle, then skip it.
+      const chainIndex = this.clients.hubPoolClient.configStoreClient.getChainIdIndicesForBlock().indexOf(chainId);
+      if (chainIndex >= lastFullyExecutedBundle.bundleEvaluationBlockNumbers.length) {
+        continue;
+      }
```
pxrl, line 517: Will this impose a 1 bundle delay on reporting on a newly activated chain? It's "only" the monitor so I suppose that's a non-issue.

src/dataworker/DataworkerClientHelper.ts
nicholaspai: oh not at all, these exist already and are used a lot. They're not really critical bugs since the from and to blocks are set equal to each other so the chain is still disabled.

src/monitor/Monitor.ts
nicholaspai: It shouldn't. Monitor chains will include all chains we have an `adapter` for so once the zksync adapter is merged in + once the 324 chain is added to the indices list it should be g2g

nicholaspai: I don't think so. This function only checks for rebalances emitted by validated bundles. So once the first bundle is executed with block ranges for chain 324, this function should be able to track it

PR #864, opened 2023-08-08
feat: add ZKSync Inventory Manager Adapter

james-a-morris:
This PR introduces the ZKSync adapter to the across inventory manager.


src/clients/bridges/BaseAdapter.ts
```diff
@@ -29,7 +28,7 @@ interface Events {
     [l1Token: string]: DepositEvent[];
   };
 }
-export class BaseAdapter {
+export abstract class BaseAdapter {
```
james-a-morris, line 31: As an abstract class, we can define the common methods that the other adapters share

nicholaspai (commented):
Three high level points that were also discussed offline
- I think this `zksync` package refers to a deprecated network. It is found in [this repo](https://github.com/matter-labs/zksync) and I believe it refers to `zkSync Lite`. We want the SDK for the new `zksync Era` network which migrated around the time of this [article](https://blog.matter-labs.io/all-aboard-zksync-era-mainnet-8b8964ba7c59). This is the repo we want [here](https://github.com/matter-labs/zksync-era) and the NPM import should be `zksync-web3` as described [https://era.zksync.io/docs/api/js](https://era.zksync.io/docs/api/js)
- We should try to call the `Mailbox` contract (to send ETH) and the `ERC20Bridge` contract directly. For sending ETH we will need to unwrap WETH and then bridge ETH over in two transactions. This will involve using the `MulticallerClient`'s [`enqueueTransaction` function.](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/MultiCallerClient.ts#L73). A lot of this work I've already done for you in this [script](https://github.com/across-protocol/relayer-v2/blob/master/scripts/zkSyncDemo.ts), but do note how sending ETH/WETH is different from sending an ERC20.
    - This matches the pattern where we use the `transactionClient` in [Polygon](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/PolygonAdapter.ts#L229), [Optimism](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/OptimismAdapter.ts#L150) and [Arbitrum](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/ArbitrumAdapter.ts#L176). The `TransactionClient` ultimately [calls `runTransaction`](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/TransactionClient.ts#L49). This reduces our dependency on third party SDK's. I understand we'll still need to use the SDK to estimate L2 gas like we do in the script [here](https://github.com/across-protocol/relayer-v2/blob/master/scripts/zkSyncDemo.ts).
- We will need to implement `wrapEthIfAboveThreshold` the same way that we do in [Optimism](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/OptimismAdapter.ts#L153). This is because we'll receive ETH on the L2 side when we bridge over the canonical bridge and we want our relayer to hold primarily WETH, not ETH.

scripts/zkSyncDemo.ts
```diff
@@ -59,6 +61,14 @@ export async function run(): Promise<void> {
...
   );
   const l2PubdataByteLimit = zksync.utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT;
+  const l1GasPriceData = await gasPriceOracle.getGasPriceEstimate(l1Provider, 1);
+  const estimatedL1GasPrice = l1GasPriceData.maxPriorityFeePerGas.add(l1GasPriceData.maxFeePerGas);
+  // The ZkSync Mailbox contract checks that the msg.value of the transaction is enough to cover the transaction base
+  // cost. The transaction base cost can be queried from the Mailbox by passing in an L1 "executed" gas price,
+  // which is the priority fee plus base fee. This is the same as calling tx.gasprice on-chain as the Mailbox
+  // contract does here:
+  // https://github.com/matter-labs/era-contracts/blob/3a4506522aaef81485d8abb96f5a6394bd2ba69e/ethereum/contracts/zksync/facets/Mailbox.sol#L287
+  console.log("Actual L1 gas price (e.g. tx.gasprice)", estimatedL1GasPrice.toString());
```
nicholaspai, line 71: Leftover console.log, should fail linting

nicholaspai: Actually this is a script, nvm

src/clients/bridges/AdapterManager.ts
```diff
@@ -122,25 +129,12 @@ export class AdapterManager {
...
-        l1Tokens.filter((token) => this.l2TokenExistForL1Token(token, 137))
-      );
-    }
-
-    if (this.adapters[42161] !== undefined) {
-      await this.adapters[42161].checkTokenApprovals(
-        address,
-        l1Tokens.filter((token) => this.l2TokenExistForL1Token(token, 42161))
-      );
+    for (const chainId of [10, 137, 324, 42161]) {
```
nicholaspai, line 132: Should we instead read from `this.supportedChains()` so we don't hardcode these chains in multiple places?

src/clients/bridges/OptimismAdapter.ts
```diff
@@ -208,4 +208,8 @@ export class OptimismAdapter extends BaseAdapter {
   private hasCustomL2Bridge(l1Token: string): boolean {
     return l1Token in this.customOvmBridgeAddresses;
   }
+
+  isSupportedToken(l1Token: string): l1Token is string {
+    return true;
```
nicholaspai, line 213: Rather than return true can we us a similar pattern to the `ZkSyncAdapter` so we don't accidentally send an ERC20 over the default OVM bridge when in fact we should set a custom erc20 bridge? Having an explicit supported token list means that we need to manually do work to add new tokens but its a safety check since this adapter can burn capital

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+    logger: winston.Logger,
+    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    readonly multicallerClient: MultiCallerClient,
+    monitoredAddresses: string[]
+  ) {
+    super(spokePoolClients, 324, monitoredAddresses, logger);
+    this.txnClient = new TransactionClient(logger);
+  }
+
+  // TODO: This will require observing production transfers to determine the correct events and contracts to track on
```
nicholaspai, line 38: Is this TODO still needed?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    // Resolve all the event queries for L1 and L2 finalizations.
+    const [l1Finalizations, l2Finalizations] = await Promise.all([
+      Promise.all(eventQueriesFromL1Finalizations),
+      Promise.all(eventQueriesFromL2Finalizations),
+    ]);
+
+    // WE ARE CURRENTLY NOT USING THESE VARIABLES
+    // WE CAN LIST THEM BELOW AS NO-OPs TO AVOID LINTING ERRORS
+    // FIXME: RESOLVE THIS FOR PROD
```
nicholaspai, line 95: +1

src/clients/bridges/AdapterManager.ts
james-a-morris: Good call - done: https://github.com/across-protocol/relayer-v2/pull/864/commits/679f769346e55c816e66ce10b1530c374bfb1499

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+  private getL1ERC20BridgeContract(): Contract {
+    const { hubChainId } = this;
+    const l1Erc20BridgeContractData = CONTRACT_ADDRESSES[hubChainId]?.zkSyncDefaultErc20Bridge;
+    if (!l1Erc20BridgeContractData) {
+      throw new Error(`l1Erc20BridgeContractData not found for chain ${hubChainId}`);
+    }
+    return new Contract(l1Erc20BridgeContractData.address, l1Erc20BridgeContractData.abi, this.getSigner(hubChainId));
+  }
+
+  private async getL2ERC20BridgeContract(): Promise<Contract> {
```
nicholaspai, line 291: Why is this async?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+    this.log("Getting cross-chain txs", { l1Tokens, l1Config: l1SearchConfig, l2Config: l2SearchConfig });
+
+    // Store all the event queries as promises for finalization events on L1 and L2.
+    const eventQueriesFromL2Finalizations: Promise<Event[]>[] = [];
+    const eventQueriesFromL1Finalizations: Promise<Event[]>[] = [];
+
+    // Resolve the mailbox and bridge contracts for L1 and L2.
+    const mailbox = this.getMailboxContract();
+    const l1ERC20Bridge = this.getL1ERC20BridgeContract();
+    const l2ERC20Bridge = await this.getL2ERC20BridgeContract();
```
nicholaspai, line 51: https://github.com/across-protocol/relayer-v2/pull/864/files#r1290488693

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+    const { provider } = this.spokePoolClients[this.chainId].spokePool;
+    const { L1_MESSENGER, L1_MESSENGER_ADDRESS } = zksync.utils; // @todo Do we want to hardcode these?
+    return new Contract(L1_MESSENGER_ADDRESS, L1_MESSENGER, provider);
+  }
+
+  isSupportedToken(l1Token: string): l1Token is string {
+    const relevantSymbol = Object.values(TOKEN_SYMBOLS_MAP).find(({ addresses }) => {
+      return Object.values(addresses).includes(l1Token);
+    })?.symbol;
+    return (relevantSymbol && this.isWeth(relevantSymbol)) || this.supportedERC20s.includes(relevantSymbol);
```
nicholaspai, line 313: this won't work... `isWeth` expects an address not a symbol. Why can't this be much simpler and just return if `this.supportedERC20s.includes(relevantSymbol)`?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+   */
+  private getl1Messenger(): Contract {
+    const { provider } = this.spokePoolClients[this.chainId].spokePool;
+    const { L1_MESSENGER, L1_MESSENGER_ADDRESS } = zksync.utils; // @todo Do we want to hardcode these?
+    return new Contract(L1_MESSENGER_ADDRESS, L1_MESSENGER, provider);
+  }
+
+  isSupportedToken(l1Token: string): l1Token is string {
+    const relevantSymbol = Object.values(TOKEN_SYMBOLS_MAP).find(({ addresses }) => {
+      return Object.values(addresses).includes(l1Token);
```
nicholaspai, line 316: This is not as safe as it could be. We know `l1Token` is an L1 token address so we shouldn't check for all addresses on all chains. There might be (very unlikely) an address collision where one token's L2 token address == another token's L1 address.
```suggestion
      return addresses[0] === l1Token;
```

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: I've updated to be more descriptive of the outstanding work.

james-a-morris: https://github.com/across-protocol/relayer-v2/pull/864/commits/7c17b8b3487fd14f617b04aeece5044fc91e87d5

src/utils/index.ts
```diff
@@ -8,6 +8,7 @@ export { Logger } from "@uma/financial-templates-lib";
 export { BigNumber, Signer, Contract, ContractFactory, Transaction, BigNumberish } from "ethers";
 export { utils, EventFilter, BaseContract, Event, Wallet } from "ethers";
 export { ethers, providers } from "ethers";
+export { Wallet as ZKSyncWallet } from "zksync-web3";
```
nicholaspai, line 11: Do we still need this?

nicholaspai (commented):
Left several blocking comments

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: It's a holdover: https://github.com/across-protocol/relayer-v2/pull/864/commits/2fd656df4f57283b7341c9b8ebccaecc45729e08

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: Resolved - it was originally using on-chain to resolve

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,315 @@
...
+    }
+    return new Contract(l2Erc20BridgeContractData.address, l2Erc20BridgeContractData.abi, provider);
+  }
+
+  /**
+   * @dev The L1Messenger contract is on zkSync and emits L1MessageSent when a mesage is sent back to mainnet.
+   */
+  private getl1Messenger(): Contract {
+    const { provider } = this.spokePoolClients[this.chainId].spokePool;
+    const { L1_MESSENGER, L1_MESSENGER_ADDRESS } = zksync.utils; // @todo Do we want to hardcode these?
```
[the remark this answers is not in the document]
james-a-morris: It's no longer needed - the function that references it is unused. https://github.com/across-protocol/relayer-v2/pull/864/commits/116e2426b4a57c8d63e322a1630485e51555c9d5

src/utils/index.ts
james-a-morris: No - removed: https://github.com/across-protocol/relayer-v2/pull/864/commits/e3e02ec3b8373d51bb10e9e716b6f27920a7760e

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: This was a holdover from testing - it's resolved here: https://github.com/across-protocol/relayer-v2/pull/864/commits/d9475bc9fd9e8fe8717a50d254a3177150e33aef

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: Tightened it to be more strict then the suggested edit - it will only test against the hub pool chain:
```
const relevantSymbol = Object.values(TOKEN_SYMBOLS_MAP).find(
      ({ addresses }) => addresses[this.hubChainId] === l1Token
    )?.symbol;
```

james-a-morris: https://github.com/across-protocol/relayer-v2/commit/d9475bc9fd9e8fe8717a50d254a3177150e33aef

nicholaspai (commented):
LGTM pending the Optimism return true change!

src/utils/AddressUtils.ts
```diff
@@ -13,3 +14,28 @@ export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0
...
 }
+
+/**
+ * Resolve the token symbol for the given token address and chain ID.
+ * @param tokenAddress The token address to resolve the symbol for.
+ * @param chainId The chain ID to resolve the symbol for.
+ * @returns The token symbol for the given token address and chain ID, or undefined if the token address is not
+ * recognized.
+ */
+export function resolveTokenSymbolFromTokenAddress(tokenAddress: string, chainId: number): string | undefined {
```
pxrl, line 25: Driveby observation on naming: A token address (and implied chainId) is the only way I can think of to resolve a token symbol, so to me there's a lot of redundancy in the name `resolveTokenSymbolFromTokenAddress`.

If I was naming it I'd probably go for `resolveTokenSymbol(tokenAddress: string, chainId: number)`, because the fact that the first argument is called `tokenAddress` is in itself quite self-explanatory. Just my 2c.

(This might just a personal nit but I felt compelled to throw it out there :wink:)

nicholaspai: Agreed with paul thats a better name

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -176,6 +184,10 @@ export class ArbitrumAdapter extends BaseAdapter {
     return await runTransaction(this.logger, this.getL1GatewayRouter(), "outboundTransfer", args, this.l1SubmitValue);
   }
 
+  async wrapEthIfAboveThreshold(): Promise<TransactionResponse | null> {
+    throw new Error("Unneccessary to wrap ETH on Arbitrum");
```
nicholaspai, line 188:
```suggestion
    throw new Error("Unnecessary to wrap ETH on Arbitrum");
```

nicholaspai (changes requested):
Not allowing this to merge until #874 is in

src/utils/AddressUtils.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/864/commits/d644337d2009c8e57d37f9ead39bb7e5644553bc

src/clients/bridges/AdapterManager.ts
```diff
@@ -122,25 +129,12 @@ export class AdapterManager {
...
-        l1Tokens.filter((token) => this.l2TokenExistForL1Token(token, 137))
-      );
-    }
-
-    if (this.adapters[42161] !== undefined) {
-      await this.adapters[42161].checkTokenApprovals(
-        address,
-        l1Tokens.filter((token) => this.l2TokenExistForL1Token(token, 42161))
-      );
+    for (const chainId of this.supportedChains()) {
```
mrice32, line 132: nice

src/clients/bridges/BaseAdapter.ts
```diff
@@ -207,7 +227,64 @@ export class BaseAdapter {
...
+
+  /**
+   * Get L1 Atomic WETH depositor contract
+   * @returns L1 Atomic WETH depositor contract
+   */
+  getAtomicDepositor(): Contract {
+    const { hubChainId } = this;
+    return new Contract(
+      this.atomicDepositorAddress,
+      CONTRACT_ADDRESSES[hubChainId].atomicWeth.abi,
```
mrice32, line 256: nit: what is atomicWeth? Should this be [atomicDepositor](https://github.com/across-protocol/relayer-v2/blob/52552a724300708b1de2326aae4b439ff107c860/src/common/ContractAddresses.ts#L118C5-L118C20)?

src/utils/AddressUtils.ts
```diff
@@ -13,3 +14,24 @@ export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0
...
+export function matchTokenSymbol(tokenAddress: string, chainId: number): string[] {
+  return Object.values(TOKEN_SYMBOLS_MAP)
+    .filter(({ addresses }) => addresses[chainId] === tokenAddress)
+    .map(({ symbol }) => symbol);
+}
+
+export function resolveTokenSymbols(tokenAddresses: string[], chainId: number): string[] {
+  return tokenAddresses
+    .map((tokenAddress) => {
+      return Object.values(TOKEN_SYMBOLS_MAP).find(({ addresses }) => addresses[chainId] === tokenAddress)?.symbol;
```
mrice32, line 34: nit: in some places we do toLowerCase or similar to ensure that there aren't weird casing issues between these addresses. Should we do something similar here?

src/utils/AddressUtils.ts
```diff
@@ -13,3 +14,24 @@ export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0
...
+/**
+ * Match the token symbol for the given token address and chain ID.
+ * @param tokenAddress The token address to resolve the symbol for.
+ * @param chainId The chain ID to resolve the symbol for.
+ * @returns The token symbols for the given token address and chain ID, or undefined if the token address is not
+ * recognized.
+ */
+export function matchTokenSymbol(tokenAddress: string, chainId: number): string[] {
+  return Object.values(TOKEN_SYMBOLS_MAP)
+    .filter(({ addresses }) => addresses[chainId] === tokenAddress)
```
mrice32, line 27: Same question

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,324 @@
...
+  async wrapEthIfAboveThreshold(threshold: BigNumber): Promise<TransactionResponse | null> {
+    const { chainId, txnClient } = this;
+    assert(chainId === 324, `chainId ${chainId} is not supported`);
+
+    const l2WethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[chainId];
+    const ethBalance = await this.getSigner(chainId).getBalance();
+    if (ethBalance.gt(threshold)) {
+      const l2Signer = this.getSigner(chainId);
+      // @dev Can re-use ABI from L1 weth as its the same for the purposes of this function.
+      const contract = new Contract(l2WethAddress, CONTRACT_ADDRESSES[1].weth.abi, l2Signer);
```
mrice32, line 265: nit: hardcoded 1 chainId. Should we use hubchain?

mrice32 (approved):
LGTM! Just a few minor questions

src/clients/bridges/BaseAdapter.ts
nicholaspai: yea i fixed this just now

src/utils/AddressUtils.ts
nicholaspai: yep

src/utils/AddressUtils.ts
nicholaspai: on it

src/clients/bridges/ZKSyncAdapter.ts
nicholaspai: yep

nicholaspai (approved):

src/utils/AddressUtils.ts
pxrl: I like to be optimistic about it the cases matching, but to implement a fallback if not. For example, as implemented here: https://github.com/across-protocol/relayer-v2/pull/864/commits/a7d21e6970d0648d926e172911ee58e7c2344d7e

`return l1Token === this.wethAddress || l1Token.toLowerCase() === this.wethAddress.toLowerCase();`

We _could_ alternatively use `src/utils/compareAddresses()`, which does a mathematical comparison after converting to BigNumber, although this is obviously intended for use in `Array.sort()`.

One advantage of implementing address comparison centrally in a util function we could add some monitoring to record how many times we revert to a lower case comparison _and succeed in matching_. That might help us to give some insight into how consistent we are about  achieve consistent normalisation of addresses on ingress into the bot (i.e. with `getAddress()`).

src/utils/AddressUtils.ts
```diff
@@ -13,3 +14,26 @@ export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0
...
+}
+
+export function resolveTokenSymbols(tokenAddresses: string[], chainId: number): string[] {
+  return tokenAddresses
+    .map((tokenAddress) => {
+      return Object.values(TOKEN_SYMBOLS_MAP).find(
+        ({ addresses }) => addresses[chainId].toLowerCase() === tokenAddress.toLowerCase()
+      )?.symbol;
+    })
+    .filter(Boolean);
```
pxrl, line 38:
```suggestion
  const tokenSymbols = Object.values(TOKEN_SYMBOLS_MAP);
  return tokenAddresses
    .map((tokenAddress) => {
      return tokenSymbols.find(
        ({ addresses }) => addresses[chainId].toLowerCase() === tokenAddress.toLowerCase()
      )?.symbol;
    })
    .filter(Boolean);
```

src/utils/AddressUtils.ts
```diff
@@ -13,3 +14,26 @@ export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0
...
 }
+
+/**
+ * Match the token symbol for the given token address and chain ID.
+ * @param tokenAddress The token address to resolve the symbol for.
+ * @param chainId The chain ID to resolve the symbol for.
+ * @returns The token symbols for the given token address and chain ID, or undefined if the token address is not
+ * recognized.
+ */
+export function matchTokenSymbol(tokenAddress: string, chainId: number): string[] {
```
pxrl, line 25: Given that there's only one input address, should we ever receive multiple symbols out for a given chain ID?

nicholaspai: We shouldn't I'll throw an error

nicholaspai: actually, we can because of ETH/WETH

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,301 @@
...
+        amount: eventSpread["_amount"],
+        to: eventSpread["_to"],
+        ...eventSpread,
+      };
+    };
+
+    await Promise.all(
+      this.monitoredAddresses.map((address) =>
+        Promise.all(
+          supportedL1Tokens.map(async (l1TokenAddress) => {
```
pxrl, line 73: `utils/mapAsync()` from sdk-v2 might help to simplify this one a bit.

contracts/AtomicWethDepositor.sol
james-a-morris, line 1: I left comments here: https://github.com/across-protocol/relayer-v2/pull/874/files#r1293499524

scripts/zkSyncDemo.ts
```diff
@@ -59,6 +61,14 @@ export async function run(): Promise<void> {
     connectedSigner
   );
   const l2PubdataByteLimit = zksync.utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT;
+  const l1GasPriceData = await gasPriceOracle.getGasPriceEstimate(l1Provider, 1);
```
james-a-morris, line 64:
```suggestion
  const l1GasPriceData = await gasPriceOracle.getGasPriceEstimate(l1Provider, l1ChainId);
```

scripts/zkSyncDemo.ts
```diff
@@ -98,6 +107,7 @@ export async function run(): Promise<void> {
       chainId: args.chainId,
       method,
       args: params,
+      gasLimitMultiplier: 3,
```
james-a-morris, line 110: Should we leave a comment on why this is hardcoded to 3?

src/clients/bridges/AdapterManager.ts
```diff
@@ -63,29 +67,32 @@ export class AdapterManager {
...
-  // needs to e done on Boba and Optimism as only these two chains require ETH to be sent over the canonical bridge.
+  // needs to be done on chains where rebalancing WETH from L1 to L2 results in the relayer receiving ETH
+  // (not the ERC20).
   async wrapEthIfAboveThreshold(wrapThreshold: BigNumber): Promise<void> {
-    const optimismCall =
-      this.spokePoolClients[10] !== undefined
-        ? (this.adapters[10] as OptimismAdapter).wrapEthIfAboveThreshold(wrapThreshold)
-        : Promise.resolve(undefined);
-    const [optimismWrapTx] = await Promise.all([optimismCall]);
+    const chainsToWrapEtherOn = [10, 324];
```
james-a-morris, line 83: Should we hardcode this elsewhere? Or leave a `//TODO: Add for Base`?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -207,7 +233,106 @@ export class BaseAdapter {
...
+   * @returns L1 Atomic WETH depositor contract
+   */
+  getAtomicDepositor(): Contract {
+    const { hubChainId } = this;
+    return new Contract(
+      this.atomicDepositorAddress,
+      CONTRACT_ADDRESSES[hubChainId].atomicDepositor.abi,
+      this.getSigner(hubChainId)
+    );
   }
```
james-a-morris, line 256: Is this needed for all base adapter children? We could use an intermediary child class that has this function

src/clients/bridges/OptimismAdapter.ts
```diff
@@ -140,21 +151,22 @@ export class OptimismAdapter extends BaseAdapter {
...
-    return (await txnClient.submit(originChainId, [txnRequest]))[0];
+    return await this._sendTokenToTargetChain(
+      l1Token,
+      l2Token,
+      amount,
+      contract,
+      method,
+      args,
+      gasLimitMultiplier,
+      BigNumber.from(0),
```
james-a-morris, line 167: It may be more consistent to use `toBN(0)`

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,301 @@
...
+    logger: winston.Logger,
+    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    monitoredAddresses: string[]
+  ) {
+    super(
+      spokePoolClients,
+      324,
+      monitoredAddresses,
+      logger,
+      Object.values(TOKEN_SYMBOLS_MAP)
```
james-a-morris, line 38: Did we want to be explicit with these values and not just read the token symbols map?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,301 @@
...
+      ? await zksync.utils.estimateDefaultBridgeDepositL2Gas(
+          l1Provider,
+          zkProvider,
+          l1Token,
+          amount,
+          address,
+          address,
+          gasPerPubdataLimit
+        )
+      : 2_000_000;
```
james-a-morris, line 172: Should we add a link to the docs where this hardcoded value is added?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,301 @@
...
+   * @notice sendTokenToTargetChain will send WETH as ETH to the L2 recipient so we need to implement
+   * this function so that the AdapterManager can know when to wrap ETH into WETH.
+   * @param threshold
+   * @returns
+   */
+  async wrapEthIfAboveThreshold(threshold: BigNumber): Promise<TransactionResponse | null> {
+    const { chainId, txnClient } = this;
+    assert(chainId === 324, `chainId ${chainId} is not supported`);
+
+    const l2WethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[chainId];
```
james-a-morris, line 237: This would fail if we wanted to use the ZKSync testnet - should we allow the testnet to be included in the assertion?

src/utils/AddressUtils.ts
james-a-morris: @pxrl yes, this happens due to Ether/Wrapped Ether having the same entry in our TOKEN_SYMBOLS_MAP

src/utils/RPCUtils.ts
```diff
@@ -0,0 +1,16 @@
+import { assert } from "chai";
```
pxrl, line 1: `chai` is a devDependency and it's only for test.

```suggestion
import assert from "assert";
```

yarn.lock
```diff
@@ -3121,7 +3104,7 @@
     "@typescript-eslint/types" "4.33.0"
     eslint-visitor-keys "^2.0.0"
 
-"@uma/common@2.33.0", "@uma/common@^2.17.0", "@uma/common@^2.28.4", "@uma/common@^2.29.0", "@uma/common@^2.33.0":
+"@uma/common@2.33.0", "@uma/common@^2.17.0", "@uma/common@^2.28.4", "@uma/common@^2.33.0":
```
james-a-morris, line 3107: Should we begin phasing this package out of the repositories?

james-a-morris (commented):
Left a few comments

src/clients/bridges/ZKSyncAdapter.ts
nicholaspai: good call

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -0,0 +1,301 @@
...
+    monitoredAddresses: string[]
+  ) {
+    super(
+      spokePoolClients,
+      324,
+      monitoredAddresses,
+      logger,
+      Object.values(TOKEN_SYMBOLS_MAP)
+        .filter(({ addresses }) => isDefined(addresses[CHAIN_IDs.ZK_SYNC]))
+        .map(({ symbol }) => symbol)
```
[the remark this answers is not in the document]
nicholaspai: yeah we can do that too, I went back and forth on this but I think its safer to explicitly opt-in to supporting new tokens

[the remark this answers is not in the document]
nicholaspai: I'll do the same for Optimism

src/clients/bridges/BaseAdapter.ts
nicholaspai: 3 of the 5 of them so yeah i'd assume its pretty normal

src/clients/bridges/ZKSyncAdapter.ts
nicholaspai: done

src/clients/bridges/ZKSyncAdapter.ts
nicholaspai: nah i don't want this to be used on testnets, if we want to support testnets we need to make another sweeping refactor of the inveitory client

yarn.lock
nicholaspai: definitely

PR #865, opened 2023-08-09
feat(monitor): Add better logs about configured chainIds

nicholaspai:


mrice32 (approved):

james-a-morris (approved):
Should we get all the other logs using # instead of ::?

PR #870, opened 2023-08-10
feat: consolidate supported logic

james-a-morris:


src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -63,7 +63,9 @@ export class ArbitrumAdapter extends BaseAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     monitoredAddresses: string[]
   ) {
-    super(spokePoolClients, 42161, monitoredAddresses, logger);
+    super(spokePoolClients, 42161, monitoredAddresses, logger, {
+      addresses: [...Object.keys(l1Gateways), ...Object.keys(l2Gateways)],
```
nicholaspai, line 67: Do we need to use an intermediate set object to pull out duplicates?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -50,7 +51,14 @@ export abstract class BaseAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     _chainId: number,
     readonly monitoredAddresses: string[],
-    readonly logger: winston.Logger
+    readonly logger: winston.Logger,
+    readonly supportedTokens:
+      | {
+          symbols: SupportedTokenSymbol[];
```
nicholaspai, line 57: Let's stick with either symbols or addresses. Would prefer symbols

src/clients/bridges/BaseAdapter.ts
```diff
@@ -238,6 +246,34 @@ export abstract class BaseAdapter {
     return new Contract(this.wethAddress, CONTRACT_ADDRESSES[hubChainId].weth.abi, this.getProvider(hubChainId));
   }
 
+  /**
+   * Determine whether this adapter supports an l1 token address
+   * @param l1Token an address
+   * @returns True if l1Token is supported
+   */
+  isSupportedToken(l1Token: string): l1Token is SupportedL1Token {
```
nicholaspai, line 254: See comment above about forcing supported tokens to be symbols

nicholaspai (commented):
Sorry if you're still making changes but I've left comments

src/clients/bridges/ArbitrumAdapter.ts
james-a-morris: It shouldn't be necessary, but we definitely could just to be sure

nicholaspai: i assumed the keys for both objects would be L1 token addresses so there should definitely be duplicates

src/clients/bridges/OptimismAdapter.ts
```diff
@@ -33,7 +33,7 @@ export class OptimismAdapter extends BaseAdapter {
     // monitoring transfers from HubPool to SpokePools where the sender is HubPool.
     readonly senderAddress?: string
   ) {
-    super(spokePoolClients, 10, monitoredAddresses, logger, { symbols: [] });
+    super(spokePoolClients, 10, monitoredAddresses, logger, { symbols: Object.keys(TOKEN_SYMBOLS_MAP) });
```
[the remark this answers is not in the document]
james-a-morris: Done

src/clients/bridges/BaseAdapter.ts
james-a-morris: Done

src/clients/bridges/BaseAdapter.ts
james-a-morris: Symbols are allowed only

src/clients/bridges/ArbitrumAdapter.ts
james-a-morris: Right, but we're only using the list to test if an l1Token address/symbol exists, so duplicates won't affect the result. However, for testing/logging purposes, we can remove the duplicates

src/clients/bridges/BaseAdapter.ts
```diff
@@ -34,13 +35,17 @@ interface Events {
 const { TOKEN_SYMBOLS_MAP } = sdkConstants;
 
 type SupportedL1Token = string;
+type SupportedTokenSymbol = string;
 
 export abstract class BaseAdapter {
+  static readonly hubChainId = 1; // @todo: Make dynamic
+
+  readonly hubChainId = BaseAdapter.hubChainId;
```
nicholaspai, line 43: I'm confused, why are there two of these? Should we make the `static` one in ALL_CAPS?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -34,13 +35,17 @@ interface Events {
...
+
   chainId: number;
   baseL1SearchConfig: MakeOptional<EventSearchConfig, "toBlock">;
   baseL2SearchConfig: MakeOptional<EventSearchConfig, "toBlock">;
-  readonly hubChainId = 1; // @todo: Make dynamic
-  readonly wethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[this.hubChainId];
+  readonly wethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[BaseAdapter.hubChainId];
```
nicholaspai, line 48: Shouldn't we be using `this.hubChainId`..?

src/clients/bridges/BaseAdapter.ts
```diff
@@ -238,6 +244,32 @@ export abstract class BaseAdapter {
...
+    // Resolve the symbol for this token
+    let relevantSymbol = tokenInfo?.symbol;
+
+    // If we don't have a symbol for this token, return that the token is not supported
+    if (!isDefined(relevantSymbol)) {
+      return false;
+    }
+
+    // If the symbol result is "ETH", then we need to modify it to be "WETH"
+    if (relevantSymbol === "ETH") {
```
nicholaspai, line 265: Is this true? Shouldn't we independently support ETH or WETH? I actually don't think we want to support ETH

nicholaspai (commented):
Just a few questions!

src/clients/bridges/BaseAdapter.ts
james-a-morris: It's completely interchangeable - but I've reverted back to save on the diff: https://github.com/across-protocol/relayer-v2/pull/870/commits/81854db60ceed7e0aecb66718b2e7daa223642a8

src/clients/bridges/BaseAdapter.ts
james-a-morris: I've made it now all caps. The prevailing issue is that in the child classes, we can't call `this.{...}` before the call to `super( )`. This is normally fine, except we need to access the `this.hubChainId` within the mappings to convert `l1TokenAddresses -> L1Symbols`. Below is one example of where we need the static variable instead of using the `this.hubChainId`

```
  constructor(
    logger: winston.Logger,
    monitoredAddresses: string[]
  ) {
    super(
      spokePoolClients,
      42161,
      monitoredAddresses,
      logger,
      resolveTokenSymbolsFromTokenAddresses(
        Array.from(new Set([...Object.keys(l1Gateways), ...Object.keys(l2Gateways)])),
        BaseAdapter.HUB_CHAIN_ID
      )
    );
  }
```

If we tried to use `this.hubChainId`, we'd get an error because we're trying to access the class scope prior to calling the parent constructor.

james-a-morris: Done here: https://github.com/across-protocol/relayer-v2/pull/870/commits/b933de5eaca0cff8f50c07a4d3e0be5a769642ed

src/clients/bridges/BaseAdapter.ts
james-a-morris: This is one of those few places where `ETH` and `WETH` have conflicting targets in `TOKEN_SYMBOLS_MAP`. The addresses for Ether and WETH are identical:

**WETH**
```
{
  name: "Wrapped Ether",
  symbol: "WETH",
  decimals: 18,
  addresses: {
    [CHAIN_IDs.MAINNET]: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
    [CHAIN_IDs.OPTIMISM]: "0x4200000000000000000000000000000000000006",
    [CHAIN_IDs.POLYGON]: "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619",
    [CHAIN_IDs.BOBA]: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
    [CHAIN_IDs.ARBITRUM]: "0x82aF49447D8a07e3bd95BD0d56f35241523fBab1",
    [CHAIN_IDs.ARBITRUM_GOERLI]: "0xe39Ab88f8A4777030A534146A9Ca3B52bd5D43A3",
    [CHAIN_IDs.GOERLI]: "0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6",
    [CHAIN_IDs.SEPOLIA]: "0x7b79995e5f793A07Bc00c21412e50Ecae098E7f9",
    [CHAIN_IDs.ZK_SYNC]: "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",
    [CHAIN_IDs.ZK_SYNC_GOERLI]: "0x20b28B1e4665FFf290650586ad76E977EAb90c5D",
  },
},
```

**Ether**
```
{
  name: "Ether",
  symbol: "ETH",
  decimals: 18,
  addresses: {
    [CHAIN_IDs.MAINNET]: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
    [CHAIN_IDs.OPTIMISM]: "0x4200000000000000000000000000000000000006",
    [CHAIN_IDs.POLYGON]: "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619",
    [CHAIN_IDs.BOBA]: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
    [CHAIN_IDs.ARBITRUM]: "0x82aF49447D8a07e3bd95BD0d56f35241523fBab1",
    [CHAIN_IDs.ARBITRUM_GOERLI]: "0xe39Ab88f8A4777030A534146A9Ca3B52bd5D43A3",
    [CHAIN_IDs.GOERLI]: "0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6",
    [CHAIN_IDs.SEPOLIA]: "0x7b79995e5f793A07Bc00c21412e50Ecae098E7f9",
    [CHAIN_IDs.ZK_SYNC]: "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",
    [CHAIN_IDs.ZK_SYNC_GOERLI]: "0x20b28B1e4665FFf290650586ad76E977EAb90c5D",
  },
},
```

----
My main thoughts for this is that calling this function with the intention of `ETH` would never really make any sense because `ETH` as a native token doesn't have a contract address as it's the native token.

src/clients/bridges/BaseAdapter.ts
```diff
@@ -238,6 +244,32 @@ export abstract class BaseAdapter {
...
+
+    // Resolve the symbol for this token
+    let relevantSymbol = tokenInfo?.symbol;
+
+    // If we don't have a symbol for this token, return that the token is not supported
+    if (!isDefined(relevantSymbol)) {
+      return false;
+    }
+
+    // If the symbol result is "ETH", then we need to modify it to be "WETH"
```
nicholaspai, line 264:
```suggestion
    // If the symbol result is "ETH", then we need to modify it to be "WETH" because WETH and ETH are both used as symbol keys in TOKEN_SYMBOLS_MAP. This does NOT assume that they have equivalent L1 token addresses
```

nicholaspai (approved):
LGTM pending this one comment change https://github.com/across-protocol/relayer-v2/pull/870/files/b2118db280f9fc3c9ba5a7f7b9b6afdb5efe795a#r1290642361

PR #873, opened 2023-08-11
feat: add Base constants/defaults

mrice32:
This adds a bunch of constants and defaults to support Base.

Note: for some of the code here to work, this sdk PR will need to be merged and released: https://github.com/across-protocol/sdk-v2/pull/369.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -17,10 +17,11 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK = 4 * 60 * 60;
 // Target ~4 days per chain. Should cover all events needed to construct pending bundle.
 export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
   1: 28800,
-  10: 1382400,
+  10: 172800, // 1 block every 2 seconds after bedrock
   137: 138240,
   288: 11520,
   324: 4 * 24 * 60 * 60,
+  8453: 172800, // Same as Optimism.
   42161: 1382400,
 };
 
@@ -55,10 +56,12 @@ export const DEFAULT_MIN_DEPOSIT_CONFIRMATIONS = {
   137: 128, // Commonly used finality level for CEX's that accept Polygon deposits
   288: 0,
   324: 0,
+  8453: 0,
   42161: 0,
   // Testnets:
   5: 0,
   280: 0,
+  84531: 0,
   421613: 0,
 };
 export const MIN_DEPOSIT_CONFIRMATIONS: { [threshold: number | string]: { [chainId: number]: number } } = {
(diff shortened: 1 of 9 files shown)


src/common/Constants.ts
```diff
@@ -102,23 +109,27 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
...
   288: 5, // At 60s/block, 50 blocks = 25 mins
   324: 1500, // At 1s/block, 25 mins = 1500 blocks.
+  8453: 750, // At 2s/block, 25 mins = 750 blocks.
   42161: 300, // At a conservative 1 TPS, 5 mins = 300 seconds = 300 transactions. And 1 block per txn.
   // Testnets:
   5: 0,
   280: 0,
+  84531: 0,
   421613: 0,
 };
```
pxrl, line 135: We _might_ want to set `8453: 75` in `DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE` below. I should note though, that the rough limit of 75 was determined months back (before Bedrock), when we had > 100 fills enqueued and were hitting the gas limit. We might theoretically be able to nudge the limit higher now, though in practice we almost _never_ have so many txns in a multicall so we only benefit from this after some kind out outage.

Side note: Limiting based on the number of transactions in the bundle is fraught because the real issue is the gas consumption of the bundle, so we should probably just be configuring a max gas limit here instead. :wood:  to :axe:...

nicholaspai: These are good points. I wonder why we only override the multicall chunk size for `10` and not other chains, any idea? Was it just because that was the chain with a lot more fills than the others?

src/dataworker/DataworkerUtils.ts
```diff
@@ -416,7 +416,7 @@ export function l2TokensToCountTowardsSpokePoolLeafExecutionCapital(
   l2TokenAddress: string,
   l2ChainId: number
 ): string[] {
-  const spokesThatHoldEthAndWeth = [10, 324];
+  const spokesThatHoldEthAndWeth = [10, 324, 8453];
```
nicholaspai, line 419: nit: do you think we should move this to common/Constants? I'm glad you added this here but I was wondering how you figured out how to add it, did you just do a search for "10" in your IDE?

nicholaspai (approved):
This is comprehensive LGTM

src/common/Constants.ts
pxrl: We had a specific issue a few months back where the relayer was down for a period of time, so the number of deposits destined for Optimism exceeded 100. The `MulticallClient` was slicing the fills into batches of 100, but we couldn't execute until we dropped the batch size to 75. So I think we were implicitly hitting a gas limit, but it wasn't ever confirmed.

src/dataworker/DataworkerUtils.ts
mrice32: I searched for 324 :)

mrice32: Can move

src/common/Constants.ts
mrice32: Will add 75 for Base. Shouldn't impact much!

mrice32: Fixed.

src/dataworker/DataworkerUtils.ts
mrice32: Fixed.

PR #876, opened 2023-08-11
fix(OptimismAdapter): Support custom SNX Adapter ABI

nicholaspai:


src/clients/bridges/OptimismAdapter.ts
```diff
@@ -44,18 +46,25 @@ export class OptimismAdapter extends BaseAdapter {
...
+          ? "DepositInitiated"
+          : "ERC20DepositInitiated";
         let l1SearchFilter = [l1Token, undefined, monitoredAddress];
         let l2SearchFilter = [l1Token, undefined, monitoredAddress];
         if (this.isWeth(l1Token)) {
           l1SearchFilter = [undefined, monitoredAddress];
           l2SearchFilter = [ZERO_ADDRESS, undefined, monitoredAddress];
+        } else if (this.isSNX(l1Token)) {
+          l1SearchFilter = [monitoredAddress];
+          l2SearchFilter = [undefined, monitoredAddress];
```
mrice32, line 61: On the l2 event (DepositFinalized), it looks like the monitored address is the first indexed param, not the second... am I missing something?

mrice32 (approved):
LGTM! Only one question after going through the ABIs.

src/clients/bridges/OptimismAdapter.ts
nicholaspai: thanks for catching this you're right, and matches the ABI

PR #880, opened 2023-08-14
fix: Remove ProfitClient testnet incompatibilities

pxrl:
Rather than trying to correctly translate testnet ERC20 addresses to mainnet addresses, just accept token prices of zero on testnets. This simplifies the code and removes a few potential sources of error.

While here, sub out a couple of toBN(0) for the ethers constant Zero.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -67,7 +67,7 @@ const QUERY_HANDLERS: {
   8453: relayFeeCalculator.BaseQueries,
   // Testnets:
   5: relayFeeCalculator.EthereumQueries,
-  280: relayFeeCalculator.ZkSyncQueries,
+  280: relayFeeCalculator.zkSyncGoerliQueries,
   421613: relayFeeCalculator.ArbitrumQueries,
   84531: relayFeeCalculator.BaseGoerliQueries,
 };
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,5 +1,5 @@
 import { Provider } from "@ethersproject/abstract-provider";
-import { utils as ethersUtils } from "ethers";
+import { constants as ethersConstants, utils as ethersUtils } from "ethers";
 import * as constants from "../common/Constants";
 import { assert, BigNumber, formatFeePct, max, winston, toBNWei, toBN, assign } from "../utils";
 import { HubPoolClient } from ".";
@@ -83,7 +83,7 @@ export class ProfitClient {
   // Queries needed to fetch relay gas costs.
   private relayerFeeQueries: { [chainId: number]: relayFeeCalculator.QueryInterface } = {};
 
-  private isGoerli: boolean;
+  private readonly isTestnet: boolean;
 
   // @todo: Consolidate this set of args before it grows legs and runs away from us.
   constructor(
@@ -114,7 +114,7 @@ export class ProfitClient {
       );
     }
 
-    this.isGoerli = this.hubPoolClient.chainId === CHAIN_IDs.GOERLI;
+    this.isTestnet = this.hubPoolClient.chainId !== CHAIN_IDs.MAINNET;
   }
 
   getAllPrices(): { [address: string]: BigNumber } {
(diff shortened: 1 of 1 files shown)


src/clients/ProfitClient.ts
```diff
@@ -118,7 +118,7 @@ export class ProfitClient {
       );
     }
 
-    this.isGoerli = this.hubPoolClient.chainId === CHAIN_IDs.GOERLI;
+    this.isTestnet = this.hubPoolClient.chainId !== CHAIN_IDs.MAINNET;
```
nicholaspai, line 121: +1

src/clients/ProfitClient.ts
```diff
@@ -329,7 +326,7 @@ export class ProfitClient {
 
   isFillProfitable(deposit: Deposit, fillAmount: BigNumber, refundFee: BigNumber, l1Token: L1Token): boolean {
     const { profitable } = this.getFillProfitability(deposit, fillAmount, refundFee, l1Token);
-    return profitable ?? false;
+    return profitable ?? this.isTestnet;
```
nicholaspai, line 329: Should we force the caller even on testnets to set IGNORE_PROFITABILITY=false rather than silently always return `true` here?

nicholaspai (approved):
One suggestion!

src/clients/ProfitClient.ts
james-a-morris, line 121: This seems to be defined in multiple places and referenced as a ground truth throughout the application. Is it possible that we should definitively define the `HUBPOL_CHAIN_ID` as a constant (maybe to be resolved via an env) and reference that constant?

Right now, we have prod = 1, testnet = 5. AFAIK I don't believe we conditionally set the hub pool chain id differently anywhere in the repo outside of the initial runtime so it would probably remove a few lines of redundant references to resolve the hubpool chain id

src/clients/ProfitClient.ts
```diff
@@ -128,20 +128,17 @@ export class ProfitClient {
   getPriceOfToken(token: string): BigNumber {
     // Warn on this initially, and move to an assert() once any latent issues are resolved.
     // assert(this.tokenPrices[token] !== undefined, `Token ${token} not in price list.`);
-    if (this.tokenPrices[token] === undefined) {
+    if (this.tokenPrices[token] === undefined && !this.isTestnet) {
```
james-a-morris, line 131: Similar to above - our hubpool really only exists on testnet (chainId = 5) or prod (chainId = 1)

src/clients/ProfitClient.ts
```diff
@@ -128,20 +128,17 @@ export class ProfitClient {
...
-    if (this.isGoerli && token === TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.GOERLI]) {
-      return this.tokenPrices[WETH];
-    }
     return this.tokenPrices[token];
   }
 
   // @todo: Factor in the gas cost of submitting the RefundRequest on alt refund chains.
   getTotalGasCost(chainId: number): BigNumber {
-    return this.totalGasCosts[chainId] ? toBN(this.totalGasCosts[chainId]) : toBN(0);
+    return this.totalGasCosts[chainId] ? toBN(this.totalGasCosts[chainId]) : ethersConstants.Zero;
```
james-a-morris, line 141: We use `ethersConstants` in multiple locations, should we export it in `utils` so that if we ever want to just define `BIGNUM_ZERO = toBN(0)` we could save ourselves the import?

james-a-morris (commented):
Left a few reviews related to how we could remove redundancies in the relayer repo

src/clients/ProfitClient.ts
pxrl: `IGNORE_PROFITABILITY` was deprecated a while ago; we now have `MIN_RELAYER_FEE_PCT` and `RELAYER_GAS_MULTIPLIER`. `RELAYER_GAS_MULTIPLIER=0` gives approximately the same behaviour as `IGNORE_PROFITABILITY` did before its deprecation.

My preference is to just implicitly ignore profitability on testnets, rather than to force the operator to fiddle with more config. At the moment, provided that I have the correct RPC endpoints defined, I can switch between prod and test environments by toggling `HUB_CHAIN=5` in my config. This is _really_ nice, and I don't run the risk of making unprofitably fills when I revert back to running on prod.

wdyt?

src/clients/ProfitClient.ts
pxrl: Yeah, I am on board with this :+1: I'll follow up with a PR to implement that, since it'll touch quite a few files.

src/clients/ProfitClient.ts
pxrl: We also have Sepolia (11155111), and that will end up being the eventual testnet hub since Görli will be deprecated at some point. But the code should also be generic and flexible enough that it works in hre (where we have 666 and 31337 defined). And it should theoretically also be possible for others to spin it up in a bespoke environment without everything breaking.

PR #888, opened 2023-08-16
feat: optionally sanitize environment of keys

james-a-morris:
This PR exists as a result of a conversation I've had with @pxrl about improving security around resolving keys from the environment. In a secure and trusted system, we should be able to retrieve a signer using keys from the ENV multiple times without issue. However, there is never a guarantee of security and we should minimize the amount of time that keys can be stored in ENV state as much as possible.

This PR aims to correct this by allowing our primary `getSigner` function which uses the ENV to clean the environment after keys have been read. This follows our current paradigm of only using this function at the top level of our index.ts file


src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,87 @@
 import { Wallet, retrieveGckmsKeys, getGckmsConfig } from "./";
-// const args = require("minimist")(process.argv.slice(2));
-
```
james-a-morris, line 3: This is redundant - I've removed for convenience

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,87 @@
...
   return new Wallet(privateKeys[0]); // GCKMS retrieveGckmsKeys returns multiple keys. For now we only support 1.
 }
 
-function getMnemonicSigner() {
+/**
+ * Retrieves a signer based on the mnemonic set in the env.
+ * @returns A signer based on the mnemonic set in the env.
+ * @throws If the mnemonic is not set.
+ */
+function getMnemonicSigner(): Wallet {
```
james-a-morris, line 70: Added comments & return types for readability

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,87 @@
...
-export async function getSigner(): Promise<Wallet> {
+/**
+ * Retrieves a signer based on the wallet type defined in the args.
+ * @param cleanEnv If true, clears the mnemonic and private key from the env after retrieving the signer.
+ * @returns A signer.
+ * @throws If the wallet type is not defined or the mnemonic/private key is not set.
+ * @note If cleanEnv is true, the mnemonic and private key will be cleared from the env after retrieving the signer.
+ * @note This function will throw if called a second time after the first call with cleanEnv = true.
+ */
+export async function getSigner(cleanEnv = true): Promise<Wallet> {
```
james-a-morris, line 13: I figure there may be the most contention here on whether or not we should have this as `opt-in` or `opt-out`

james-a-morris:
@pxrl @nicholaspai making this **opt-in** so that all normal code will not change. Plus, this code also cleans up/refactors some logic which makes the code more readable.

nicholaspai (approved):

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,87 @@
 import { Wallet, retrieveGckmsKeys, getGckmsConfig } from "./";
-// const args = require("minimist")(process.argv.slice(2));
-
 import minimist from "minimist";
 const args = minimist(process.argv.slice(2));
```
pxrl, line 3: While here, wdyt about removing this and making the key source an input argument to `getSigner()` ?

Background: Parsing argv from a utility function like `getSigner()` seems like an antipattern to me. This (for instance) prevents us from setting `--wallet memonic` as the default, in which case it _doesn't_ appear on argv.

james-a-morris: I think we'll need to create some utility to parse out argv. However, I agree that it shouldn't be in `getSigner()`

I've tried to include this in the new `utils/CLIUtils.ts`

.env.example
```diff
@@ -8,6 +8,10 @@
 #MNEMONIC="your twelve or twenty four word seed phrase..."
 #PRIVATE_KEY=0xabc123...
 
+# Specify which wallet to use. This can be set as one of: {MNEMONIC, PRIVATE_KEY, GCPKMS}
+# NOTE: The command line argument --wallet takes precedence over this configuration.
+# NOTE: If neither the command line argument nor this configuration is set, the default is MNEMONIC.
+WALLET_TYPE=MNEMONIC
```
james-a-morris, line 15: It would be convenient to be able to specify this in the ENV as an optional fallback

src/utils/CLIUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+import { SignerOptions, getSigner } from "./SignerUtils";
+import { Wallet } from "ethers";
+
+/**
+ * Retrieves a signer based on both the CLI args and the env.
+ * @returns A signer based on the CLI args.
+ */
+export function retrieveSignerFromCLIArgs(): Promise<Wallet> {
+  // Call into the process' argv to retrieve the CLI args.
+  const args = minimist(process.argv.slice(2));
```
james-a-morris, line 11: This is now defined dynamically so that we aren't polling argv unless this function is specifically called.

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,103 @@
...
-const args = minimist(process.argv.slice(2));
+/**
+ * Signer options for the getSigner function.
+ */
+export type SignerOptions = {
+  /*
+   * The type of wallet to use.
+   * @note If using a GCKMS wallet, the gckmsKeys parameter must be set.
+   */
+  keyType: string;
```
pxrl, line 11: Since we only support 3 types, is it worth locking this down?

```suggestion
  keyType: "gckms" | "mnemonic" | "privateKey";
```

src/utils/CLIUtils.ts
pxrl: Long term I think we probably want to implement proper argument parsing in our application entry points (i.e. `scripts`, `src/scripts`, and top-level `index.ts`. This seems like an OK approach until we find time to do that, but utility functions reading `argv` directly still jump out at me :)

.env.example
pxrl: Since the wallet type is a command-line argument, I'd rather avoid implementing too many ways of configuring it, because it becomes confusing for users who aren't necessarily clear on the precedence.

minimist defaults and script targets in `package.json` are IMO a cleaner way of doing this - i.e. we already have a script target for `yarn relay` which expands to `... yarn ts-node ./index.ts --wallet mnemonic`. If a user doesn't want to run with the mnemonic then they're welcome to override the defaults via a simple runtime argument.

james-a-morris:
@pxrl great point - I've removed the optional .env parameter to set the `--wallet`

.env.example
james-a-morris: Fair point - removed

src/utils/SignerUtils.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/888/commits/41616e22cdc16c85e2669be3eafa04dc3889e8a4

nicholaspai (approved):

.env.example
```diff
@@ -8,7 +8,6 @@
 #MNEMONIC="your twelve or twenty four word seed phrase..."
 #PRIVATE_KEY=0xabc123...
```
pxrl, line 10: Just restoring the 2 line gap between disparate sections.

```suggestion


```

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,103 @@
...
+    cleanKeysFromEnvironment();
   }
+  return wallet;
 }
 
-function getPrivateKeySigner() {
+/**
+ * Retrieves a signer based on the mnemonic set in the env.
+ * @returns A signer based on the mnemonic set in the env.
+ * @throws If the mnemonic is not set.
```
pxrl, line 59:
```suggestion
 * @throws If a valid private key is not defined in the environment.
```

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,103 @@
...
-  const privateKeys = await retrieveGckmsKeys(getGckmsConfig([args.keys]));
+  const privateKeys = await retrieveGckmsKeys(getGckmsConfig(keys));
   return new Wallet(privateKeys[0]); // GCKMS retrieveGckmsKeys returns multiple keys. For now we only support 1.
 }
 
-function getMnemonicSigner() {
+/**
+ * Retrieves a signer based on the mnemonic set in the env.
+ * @returns A signer based on the mnemonic set in the env.
+ * @throws If the mnemonic is not set.
```
pxrl, line 84:
```suggestion
 * @throws If a valid mnemonic is not defined in the environment.
```

src/utils/SignerUtils.ts
```diff
@@ -1,42 +1,103 @@
...
 
-async function getGckmsSigner() {
-  if (!args.keys) {
+/**
+ * Retrieves a signer based on the GCKMS key set in the args.
+ * @returns A signer based on the GCKMS key set in the args.
+ * @throws If the GCKMS key is not set.
+ */
+async function getGckmsSigner(keys?: string[]): Promise<Wallet> {
+  if (!isDefined(keys)) {
```
pxrl, line 74:
```suggestion
  if (!isDefined(keys) || keys.length === 0) {
```

pxrl (approved):
A couple of minor points, but otherwise LGTM.

pxrl:
@james-a-morris Just following up on this one.

pxrl (approved):
LGTM!

PR #890, opened 2023-08-17
feat: Add zkSync finalizer

pxrl:
Note: zkSync withdrawals are currently auto-finalized, but this feature will at least attempt to finalize any if it sees them.

Closes ACX-1332


src/finalizer/index.ts
```diff
@@ -148,6 +149,8 @@ async function arbitrumOneFinalizer(
 const chainFinalizers: { [chainId: number]: ChainFinalizer } = {
   10: optimismFinalizer,
   137: polygonFinalizer,
+  280: zkSyncFinalizer,
```
pxrl, line 152: We're probably unlikely to use this, but IMO it's nice if we have structured the finalizer such that it can work on testnets.

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * @returns Withdrawal finalizaton calldata and metadata.
+ */
+export async function zkSyncFinalizer(
+  logger: winston.Logger,
+  signer: Wallet,
+  hubPoolClient: HubPoolClient,
+  spokePoolClient: SpokePoolClient,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _latestBlockToFinalize: number
```
pxrl, line 33: This is actually unused here, since all events are dynamically queried for finalization status. We probably want to set it to ~22 hours in order to avoid checking status on very recent events.

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  const tokensBridged = spokePoolClient.getTokensBridged();
+  const { ready, pending } = await sortWithdrawals(l2Provider, tokensBridged);
+  if (pending.length > 0) {
+    logger.debug({
+      at: "zkSyncFinalizer",
+      message: `Found ${pending.length} pending withdrawal transactions.`,
+      pending,
+    });
+  }
```
pxrl, line 50: Contrived example from test:

```
2023-08-17 15:40:16 [debug]: {                 
  "at": "zkSyncFinalizer",                     
  "message": "Found 1 pending withdrawal transactions.",                                       
  "ready": {                                                                                   
    "0": {                                                                                     
      "amountToReturn": "41048027182598217407",                                                
      "chainId": 324,                                                                          
      "leafId": 15,                            
      "l2TokenAddress": "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",                          
      "caller": "0xf7bAc63fc7CEaCf0589F25454Ecf5C2ce904997c",                                  
      "blockNumber": 11322069,                 
      "transactionIndex": 4,                                                                   
      "logIndex": 37,                                                                          
      "transactionHash": "0x21574213dd1b18d71af3e873c2d9dacb49851d937434a80ec2ad59d404074f6a"  
    }                                                                                                                                                                                          
  },                                                                                           
  "bot-identifier": ""                                                                                                                                                                 
}
```

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  const candidates = await filterMessageLogs(wallet, l2Provider, ready);
+  if (candidates.length > 0) {
+    logger.debug({
+      at: "zkSyncFinalizer",
+      message: `Found ${candidates.length} withdrawals ready to be finalized.`,
+      candidates,
+    });
+  }
```
pxrl, line 59: Contrived example from test:

```
2023-08-17 15:40:18 [debug]: {                                                                                                                                                                 
  "at": "zkSyncFinalizer",                                                                     
  "message": "Found 1 withdrawals ready to be finalized.",                                     
  "candidates": {                                                                              
    "0": {                                                                                     
      "amountToReturn": "41048027182598217407",                                                
      "chainId": 324,                                                                          
      "leafId": 15,                                                                                                                                                                            
      "l2TokenAddress": "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",                          
      "caller": "0xf7bAc63fc7CEaCf0589F25454Ecf5C2ce904997c",                                  
      "blockNumber": 11322069,                                                                 
      "transactionIndex": 4,                                                                   
      "logIndex": 37,                                                                          
      "transactionHash": "0x21574213dd1b18d71af3e873c2d9dacb49851d937434a80ec2ad59d404074f6a", 
      "withdrawalIdx": 0                                                                       
    }                                          
  },                                                                                           
  "bot-identifier": ""                                                                 
}
```

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+    const amountFromWei = convertFromWei(amountToReturn.toString(), decimals);
+    const _withdrawal: Withdrawal = {
+      l2ChainId,
+      l1TokenSymbol,
+      amount: amountFromWei,
+      type: "withdrawal",
+    };
+
+    return _withdrawal;
+  });
```
pxrl, line 80: Contrived example from test (log is emitted from `src/finalizer/index.ts` based on the `withdrawal` object generated here.

```
2023-08-17 15:40:19 [info]: {                  
  "at": "Finalizer",                           
  "message": "Finalized ZkSync withdrawal for 41.04 WETH 🪃",                                  
  "bot-identifier": ""                 
}
```

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+    const txnReceipt = await l2Provider.getTransactionReceipt(transactionHash);
+
+    // Search backwards from the TokensBridged for the first L1MessageSent event.
+    const txnLogs = txnReceipt.logs.slice(0, logIndex).reverse();
+    const withdrawal = txnLogs.find((log) => log.topics[0] === l1MessageSent);
+
+    // @dev idx is the "withdrawal number" within the transaction, _not_ the index of the log.
+    const l1MessagesSent = txnReceipt.logs.filter((log) => log.topics[0] === l1MessageSent);
+    const withdrawalIdx = l1MessagesSent.indexOf(withdrawal);
+    return { ...tokenBridged, withdrawalIdx };
```
pxrl, line 131: This was one of the most confusing elements of the finalizer, since zkSync are not very consistent in the terminology within their documentation.

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+async function sortWithdrawals(
+  provider: zksProvider,
+  tokensBridged: TokensBridged[]
+): Promise<{ pending: TokensBridged[]; ready: TokensBridged[] }> {
+  const txnStatus = await Promise.all(
+    tokensBridged.map(({ transactionHash }) => provider.getTransactionStatus(transactionHash))
+  );
+
+  let idx = 0; // @dev Possible to infer the loop index in groupBy ??
+  const { pending = [], ready = [] } = groupBy(tokensBridged, () =>
```
nicholaspai, line 99: Can't this be or am i missing something obvious
```suggestion
  const { pending = [], ready = [] } = groupBy(tokensBridged, (txnStatus) =>
   txnStatus === TransactionStatus.Finalized ? "ready" : "pending"
```

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+  tokensBridged: TokensBridged[]
+): Promise<(TokensBridged & { withdrawalIdx: number })[]> {
+  const l1MessageSent = zkUtils.L1_MESSENGER.getEventTopic("L1MessageSent");
+
+  // Extract the relevant L1MessageSent events from the transaction.
+  const logs = await sdkUtils.mapAsync(tokensBridged, async (tokenBridged) => {
+    const { transactionHash, logIndex } = tokenBridged;
+    const txnReceipt = await l2Provider.getTransactionReceipt(transactionHash);
+
+    // Search backwards from the TokensBridged for the first L1MessageSent event.
```
nicholaspai, line 124: Does this handle multiple messages sent in same l2 txn? It seems that if there are two L1MessageSent logs in the same txn, we'll always only `.find` the last one

nicholaspai (approved):
This seems good to me barring a few questions.

Have you tested the finalizer locally, does it log any existing already finalized withdrawals?

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+  provider: zksProvider,
+  tokensBridged: TokensBridged[]
+): Promise<{ pending: TokensBridged[]; ready: TokensBridged[] }> {
+  const txnStatus = await Promise.all(
+    tokensBridged.map(({ transactionHash }) => provider.getTransactionStatus(transactionHash))
+  );
+
+  let idx = 0; // @dev Possible to infer the loop index in groupBy ??
+  const { pending = [], ready = [] } = groupBy(tokensBridged, () =>
+    txnStatus[idx++] === TransactionStatus.Finalized ? "ready" : "pending"
```
nicholaspai, line 100: Can we also add a filter to already finalized withdrawals? i.e. "finalized"

src/finalizer/utils/zkSync.ts
nicholaspai: Was this one auto finalized or no?

pxrl: Yep: https://etherscan.io/tx/0xb1185edcf5804996b10a2f4700ae5d07bffb04e1d1720efd2763dc4c536a3574

src/finalizer/utils/zkSync.ts
pxrl: In the event that there are multiple withdrawals, we should receive one `TokensBridged` event for each of them. They are passed 1:1 into `filterMessageLogs()`, so we should repeat this once per event. The reason we take the last is that the set of logs is truncated _up to_ the specific `TokensBridged` we are evaluating, so the relevant `L1MessageSent` event should be immediately before.

It could be done slightly more efficiently by filtering the input `TokensBridged` events for duplicates, but then we need to do some additional reconciliation steps to link them back up with the `l2TokenAddress` and `amountToWithdraw` values (we need these for logging later on).

src/finalizer/utils/zkSync.ts
pxrl: We need the proof in order to determine whether the withdrawal has already been completed, so this is done down at line 136.

Also to be clear on the meaning of `TransactionStatus.Finalized`: I understand this to mean that the L2 transaction has been committed (and subsequently finalized) on the L1. At the point of finalization, any L1 messages initiated by the L2 transaction are then able to be executed on the L1 (for example, finalizing a withdrawal). (It seems a bit unfortunate that the term `finalized` is overloaded in this context.)

The implementation of `getTransactionStatus()` [here](https://github.com/matter-labs/zksync-era/blob/e575ec101fe88627ab541a52464ab5025c16e6b4/sdk/zksync-web3.js/src/provider.ts#L759
) is useful for understanding the context - the distinction between `Committed`and `Finalized` for an L2 transaction depends solely on the block number, and not whether any/all L1 messages contained within the L2 transaction have been executed on the L1.

I guess another way to look at it is that this function (`sortWithdrawals()`) filters L2 transaction hashes according to whether they are committed and finalized on L1 or not. `filterMessageLogs()` filters the finalized L2 transaction hashes according to whether their L1 withdrawal events have been _executed_ on L1. The resulting set of transactions should be withdrawals that we are then able to finalize ourselves.

pxrl:
> This seems good to me barring a few questions.
> 
> Have you tested the finalizer locally, does it log any existing already finalized withdrawals?

It detects and excludes the existing 1 zkSync withdrawal for 41 ETH that was auto-finalized, and is currently logging about this recent withdrawal that was initiated on the L2: https://explorer.zksync.io/tx/0x1e02f9c0777ce46c001feefe014c262f34a54c16a8ee46c1e286b25bfbf154bb

```
2023-08-18 09:31:09 [debug]: {
  "at": "zkSyncFinalizer",
  "message": "Found 1 pending withdrawal transactions.",
  "pending": {
    "0": {
      "amountToReturn": "41267645166932497055",
      "chainId": 324,
      "leafId": 13,
      "l2TokenAddress": "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",
      "caller": "0xf7bAc63fc7CEaCf0589F25454Ecf5C2ce904997c",
      "blockNumber": 11531744,
      "transactionIndex": 7,
      "logIndex": 72,
      "transactionHash": "0x1e02f9c0777ce46c001feefe014c262f34a54c16a8ee46c1e286b25bfbf154bb"
    }
  },
  "bot-identifier": ""
}
```

src/finalizer/utils/zkSync.ts
nicholaspai: sure thats reasonable

src/finalizer/index.ts
james-a-morris: +1

src/finalizer/index.ts
```diff
@@ -163,6 +166,8 @@ export async function finalize(
   const finalizationWindows: { [chainId: number]: number } = {
     10: optimisticRollupFinalizationWindow,
     137: polygonFinalizationWindow,
+    280: oneDaySeconds * 4,
```
james-a-morris, line 169: Does this mean the finalization window is 4 days in seconds?

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+  hubPoolClient: HubPoolClient,
+  spokePoolClient: SpokePoolClient,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _latestBlockToFinalize: number
+): Promise<FinalizerPromise> {
+  const { chainId: l1ChainId } = hubPoolClient;
+  const { chainId: l2ChainId } = spokePoolClient;
+
+  const l1Provider = hubPoolClient.hubPool.provider;
+  const l2Provider = new zksProvider(getNodeUrlList(l2ChainId, 1)[0]);
```
james-a-morris, line 39: Could we use `convertEthersRPCToZKSyncRPC` here?

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+    const { decimals, symbol: l1TokenSymbol } = hubPoolClient.getTokenInfo(l1ChainId, l1TokenCounterpart);
+    const amountFromWei = convertFromWei(amountToReturn.toString(), decimals);
+    const _withdrawal: Withdrawal = {
+      l2ChainId,
+      l1TokenSymbol,
+      amount: amountFromWei,
+      type: "withdrawal",
+    };
+
+    return _withdrawal;
```
james-a-morris, line 79:
```suggestion
    return {
      l2ChainId,
      l1TokenSymbol,
      amount: amountFromWei,
      type: "withdrawal",
    };
```

src/finalizer/utils/zkSync.ts
```diff
@@ -0,0 +1,229 @@
...
+): Promise<(TokensBridged & { withdrawalIdx: number })[]> {
+  const l1MessageSent = zkUtils.L1_MESSENGER.getEventTopic("L1MessageSent");
+
+  // Extract the relevant L1MessageSent events from the transaction.
+  const logs = await sdkUtils.mapAsync(tokensBridged, async (tokenBridged) => {
+    const { transactionHash, logIndex } = tokenBridged;
+    const txnReceipt = await l2Provider.getTransactionReceipt(transactionHash);
+
+    // Search backwards from the TokensBridged for the first L1MessageSent event.
+    const txnLogs = txnReceipt.logs.slice(0, logIndex).reverse();
```
james-a-morris, line 125: Are these guaranteed to be in sorted order before the `.reverse()`?

james-a-morris (commented):
Left comments

src/finalizer/utils/zkSync.ts
nicholaspai: I think it's helpful to always know how many of the L2 withdrawals we can see from the the lookback range that are already finalized. 

For example on the other chains: The EXIT_ALREADY_PROCESSED , EXECUTED , RELAYED statuses are what I want to make sure we’re capturing the parallel of in this finalizer:


jsonPayload: {
message: "Polygon message statuses"
metadata: {
at: "PolygonFinalizer"
bot-identifier: "across-v2-finalizer"
statusesGrouped: {
EXIT_ALREADY_PROCESSED: 13
NOT_CHECKPOINTED: 0
}}}
jsonPayload: {
message: "Arbitrum outbox message statuses"
metadata: {
at: "ArbitrumFinalizer"
bot-identifier: "across-v2-finalizer"
statusesGrouped: {
EXECUTED: 36
UNCONFIRMED: 20
}}}
jsonPayload: {
message: "Optimism message statuses"
metadata: {
at: "OptimismFinalizer"
bot-identifier: "across-v2-finalizer"
statusesGrouped: {
IN_CHALLENGE_PERIOD: 24
RELAYED: 38
#}}

src/finalizer/utils/zkSync.ts
pxrl: This doesn't quite work as you'd hope :( The `Array.map()` closure needs to specifically return a `Withdrawal` type, because otherwise tsc is unhappy about the potential for the `type` field to be an arbitrary string.

2023-08-21-112044_1004x364_scrot

The other option is just to do a type assertion as follows:

```
      return {
        l2ChainId,
        l1TokenSymbol,
        amount: amountFromWei,
        type: "withdrawal",
      } as Withdrawal;
```

I'm not keen on this because it will actively coerce the returned object, whereas the former will complain if the returned data doesn't meet the requirements.

src/finalizer/utils/zkSync.ts
pxrl: Ah, I wasn't aware that the others did this. I've implemented something comparable here: https://github.com/across-protocol/relayer-v2/pull/890/commits/f5168dca5f6b6e35e72fa0c3c5c2e8c889709143

Note that I didn't log by the zkSync status, since it's only reporting on the L2 transaction status, not the status of the withdrawal itself. I've instead gone with:
 - withdrawalPending (l2 status < "Finalized")
 - withdrawalReady (l2 status == "Finalized" && l1 withdrawal not complete).
 - withdrawalFinalized (l2 status == "Finalized" && l1 withdrawal complete).

src/finalizer/utils/zkSync.ts
pxrl: In this commit (https://github.com/across-protocol/relayer-v2/pull/890/commits/2a0c30ffd64874dfffa4854ddcd418fa2587f8c9) I simplified the logic, such that we filter out duplicate txn hashes before requesting receipts for each of them. As a side-effect, this hopefully helps to clarify the distinction between the txn hash and the withdrawal event(s) they contain.

src/finalizer/utils/zkSync.ts
pxrl: Yeah :+1:  implemented: https://github.com/across-protocol/relayer-v2/pull/890/commits/12aa8941e9c29b7de02ee97cd6ffbac8e0cac5bc

src/finalizer/index.ts
pxrl: Yeah - it will filter out TokensBridged events older than 4 days :+1:

src/finalizer/utils/zkSync.ts
pxrl: They're supplied by the RPC provider and formatted by ethers. Personally, I think it's probably safe to assume that they'll always be supplied in ascending order, but we don't have any specific guarantees about that. If we want to enforce this, I'd probably advocate that we do it down in the `RetryProvider`, rather than up in application code.

PR #895, opened 2023-08-21
improve: Consolidate RetryProvider cast in RPCUtils

pxrl:
Makes it easier to use the `convertEthersRPCToZKSyncRPC()` helper.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -11,6 +11,7 @@ export type { Block, TransactionResponse, TransactionReceipt, Provider } from "@
 export { config } from "dotenv";
 
 // Utils specifically for this bot.
+export * from "./chains";
 export * from "./ProviderUtils";
 export * from "./SignerUtils";
 export * from "./DepositUtils";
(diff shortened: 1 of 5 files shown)


src/utils/index.ts
```diff
@@ -23,6 +23,7 @@ export * from "./TransactionPropBuilder";
 export * from "./ContractUtils";
 export * from "./ExecutionUtils";
 export * from "./NetworkUtils";
+export * from "./RPCUtils";
```
pxrl, line 26: Meta: Since this file solely contains zkSync helpers, what do people think about relocating it to `src/utils/zkSync/helpers.ts`, or something similar?

If we leave it as is, it's liable to become a basket of functionality for each different chain RPC provider (with their accompanying kitchen sink list of the dependencies). I think it's cleaner to organise per chain/dependency, rather than per functionality.

nicholaspai: I agree it's cleaner to separate by chain! Let's do that

nicholaspai (approved):

src/utils/index.ts
james-a-morris: 👍

james-a-morris (approved):
High level comment: it's currently tech debt, but should we eventually try to have `RetryProvider` extend `ethersProviders.Provider`? This way, it's completely a passthrough class and we don't need to do anything special

pxrl:
> High level comment: it's currently tech debt, but should we eventually try to have `RetryProvider` extend `ethersProviders.Provider`? This way, it's completely a passthrough class and we don't need to do anything special

Could be interesting to consider this... :thinking:

PR #897, opened 2023-08-21
feat: test chain routes

james-a-morris:
This PR adds a bash script and changes to the `spokePool.ts` file. The changes it aims to make are the following:

1. Allow the user to pass a token address as an alternative to a token symbol where applicable.
2. Allow the user to pass amount parameters in terms of base units of that token.

The script itself is now further documented. But, in a summary, it aims to:

1. Test that the chain that we are attempting to test exists in the available routes endpoint
2. Call the deposit function on each route so that we can easily send a test deposit to all available routes that include the target chain.


scripts/testTargetChainRoutes.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
mrice32, line 1: Just so I understand, the idea behind this script is to print out a list of enabled routes according to the API?

If so, could you write that above the explanation of the arguments? Just a one-sentence explainer for the "why" of the script?

scripts/spokepool.ts
```diff
@@ -92,6 +93,16 @@ function getTokenAddress(symbol: string, chainId: number): ERC20 {
   return { address: addresses[chainId], decimals, symbol };
 }
 
+function getTokenSymbol(address: string, chainId: number): ERC20 {
```
mrice32, line 96: Could you provide a quick summary in the PR description as to what these spokepool.ts changes are meant to do? Just to help guide the review?

mrice32 (commented):
Looks great! Just a bit more context would be helpful!

scripts/spokepool.ts
```diff
@@ -124,14 +147,25 @@ async function deposit(args: Record<string, number | string>, signer: Wallet): P
...
   const baseAmount = Number(args.amount);
-  const tokenSymbol = (args.token as string).toUpperCase();
+  const rawToken = args.token;
+
+  // Depending on whether or not the passed in token argument is a symbol or an address,
+  // we need to call either `getTokenSymbol` or `getTokenAddress`.
+  const token = (ethers.utils.isAddress(String(args.token)) ? getTokenSymbol : getTokenAddress)(
+    rawToken as string,
+    fromChainId
+  );
```
james-a-morris, line 157: Added the ability for the user to provide an address instead of just the token. We check if the address is properly formed before attempting to use that function.

scripts/testTargetChainRoutes.sh
james-a-morris: That makes complete sense. I've added additional documentation

scripts/spokepool.ts
james-a-morris: Yep! I've added comments in the PR & I'll change the description

scripts/spokepool.ts
```diff
@@ -92,6 +99,22 @@ function getTokenAddress(symbol: string, chainId: number): ERC20 {
...
 
+/**
+ * Resolves an ERC20 type from an address and chain ID.
+ * @param address The address of the token to resolve.
+ * @param chainId The chain ID to resolve the token on.
+ * @returns The ERC20 type of the token.
+ */
+function getTokenSymbol(address: string, chainId: number): ERC20 {
+  const token = Object.values(contracts.TOKEN_SYMBOLS_MAP).find(({ addresses }) => addresses[chainId] === address);
+  if (token === undefined) {
```
pxrl, line 110: We probably can't rely on the user supplying an address in checksummed-case:

```suggestion
  const _address = address.toLowerCase();
  const token = Object.values(contracts.TOKEN_SYMBOLS_MAP).find(({ addresses }) => addresses[chainId] === _address);
  if (token === undefined) {
```

mrice32 (approved):
LGTM!

scripts/testTargetChainRoutes.sh
```diff
@@ -0,0 +1,87 @@
...
+        to_chain_id=$(get_json_field "$json_obj" "destinationChainId")
+
+        # Extract the origin token address
+        origin_token_address=$(get_json_field "$json_obj" "originToken") 
+
+        # Print out the deposit message
+        echo "Sending a deposit from $from_chain_id to $to_chain_id on token $origin_token_address"
+
+        # Call yarn deposit with the correct arguments
+        echo $(yarn deposit --from $from_chain_id --to $to_chain_id --token $origin_token_address --amount 1 --wallet)
```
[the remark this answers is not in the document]
james-a-morris: The amount is now passed as input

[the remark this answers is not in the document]
pxrl: @james-a-morris I think there's potentially still an issue with this, because 1 USDC is totally different to 1 WETH or 1 WBTC, so the scale of the amount is dependent on the token that's being bridged, right?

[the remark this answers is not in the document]
james-a-morris: I have updated the script to include this logic. I do believe we're running up against overengineering if we push this script any farther

pxrl (approved):
OK from me :+1:

PR #904, opened 2023-08-22
improve: Don't require trailing / on block explorer hosts

pxrl:


src/utils/FormattingUtils.ts
```diff
@@ -38,7 +38,7 @@ export function createEtherscanLinkFromTx(networkId: number): string {
   if (constants.PUBLIC_NETWORKS[networkId]) {
     url = `${constants.PUBLIC_NETWORKS[networkId].etherscan}`;
   } else {
-    url = "https://etherscan.io/";
+    url = "https://etherscan.io";
   }
```
james-a-morris, line 43: Can we reduce this fn to 

```
export function createEtherscanLinkFromTx(networkId: number): string {
    return String(
        constants.PUBLIC_NETWORKS[networkId]?.etherscan ?? 'https://etherscan.io'
    ).replace(/\/*$/, "") + "/"
}
```

pxrl:
Dropping in favour of https://github.com/across-protocol/sdk-v2/pull/377.

PR #912, opened 2023-08-29
feat: add dataworker force bundle range

james-a-morris:
This change adds the ability for an engineer debugging the proposer/disputer to peg the proposed bundle block ranges to a specified value.


mrice32 (commented):
Looks good to me, but want to give @nicholaspai a chance to bring up any concerns.

src/dataworker/DataworkerConfig.ts
```diff
@@ -91,11 +98,39 @@ export class DataworkerConfig extends CommonConfig {
...
     // We NEVER want to force propose if the proposer is enabled.
     if (this.sendingProposalsEnabled) {
       assert(!this.forcePropose, "Cannot force propose if sending proposals is enabled");
     }
 
+    // We NEVER want to force a bundle range if the proposer OR disputer is enabled.
+    if (this.sendingDisputesEnabled || this.sendingProposalsEnabled) {
+      assert(!this.forceProposalBundleRange, "Cannot force bundle range if sending proposals or disputes is enabled");
+    }
+
```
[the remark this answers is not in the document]
james-a-morris: My intuition is that we should enforce a simulation mode whenever we enable these debugging ENV variables.

That said, if there's a valid use case for it we can definitely remove this filter

.env.example
```diff
@@ -249,3 +249,9 @@ RELAYER_TOKENS='["0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", "0xA0b86991c6218b
 # Note: This logic ONLY works if `SEND_PROPOSALS` is set to false.
 #FORCE_PROPOSAL=false
 
+# This variable can be used to simulate a bundle range that a proposal will be created for.
+# This is useful for testing the disputer and proposer logic. The format is:
+# [number, number][] where the two numbers are the start and end bundle ranges and the array
+# represents the bundle ranges that will be proposed per the chain id indices.
+# Note: This logic ONLY works if `SEND_PROPOSALS` and `SEND_DISPUTES` are BOTH set to false.
+# FORCE_PROPOSAL_BUNDLE_RANGE = [[1, 2], [1, 3], ...]
```
pxrl, line 257: I'm in two minds about whether we should document this here, since my hope for .env.example was that it'd be a copy/paste template for relayer operators to start from. Dropping dataworker concepts in here is probably going to confuse these people who might be overwhelmed by the volume of configurables.

wdyt about having separate examples for relayer, proposer, executor, ... ?

src/dataworker/Dataworker.ts
```diff
@@ -97,7 +97,8 @@ export class Dataworker {
     readonly blockRangeEndBlockBuffer: { [chainId: number]: number } = {},
     readonly spokeRootsLookbackCount = 0,
     readonly bufferToPropose = 0,
-    readonly forceProposal = false
+    readonly forceProposal = false,
+    readonly forceBundleRange?: [number, number][]
```
pxrl, line 101: Is there a concrete need for the "from" block to be configurable? Per the UMIP, the fromBlock _must_ be the block immediately following the previous bundle, so making it configurable introduces a possible source of disputes.

src/dataworker/DataworkerConfig.ts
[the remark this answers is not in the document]
pxrl: Hot take: I'd like to deprecate the `PROPOSER_ENABLED`  (...) and use runtime arguments to determine the function. That would prevent them from ever being enabled concurrently.

Additional background: I think it's an improvement to specify this on the command line, vs. having to edit my .env if I want to run the bot in a different personality.

Examples:

_nb. I tweaked the names somewhat in the package.json script targets_

`yarn across-propose` => `yarn ts-node ./index.ts --proposer`
`yarn across-verify` => `yarn ts-node ./index.ts --disputer`
`yarn across-finalize` => `yarn ts-node ./index.ts --finalizer`
`yarn across-execute` => `yarn ts-node ./index.ts --executor`

src/dataworker/Dataworker.ts
james-a-morris: Agreed - but the purpose of this ENV variable is to provide full flexibility for the engineer to decide block ranges. (both the min and the max). 

My intuition behind these changes is that they exist only for simulations. Currently, we do not allow the user to enable either `SEND_PROPOSALS` or `SEND_DISPUTES` enabled when the `FORCED_BUNDLE_RANGE` is specified.

Given that, do you think having control over the start/end ranges are still helpful?

mrice32: Yeah, this is a good point @pxrl. When I overrode this manually, I overrode both because that was easier, but it's unclear if there's a use case for overriding both.

I do think it could be nice as a way to print out a bundle _for any range_ (even an invalid one), as that could be a pretty powerful debugging tool to narrow the source of an error.

On the margin, I think flexibility is preferable here, and if you are not using the correct fromBlock and use this to actually propose, you should be disputed.

src/dataworker/DataworkerConfig.ts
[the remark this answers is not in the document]
mrice32: > My intuition is that we should enforce a simulation mode whenever we enable these debugging ENV variables.

I'm not sure I agree. I think this could be useful for an operator to more precisely control their proposal. Rather than thinking of this exclusively as a debug tool, I think of it as a powerful admin tool that you should only use if you know what you're doing (in the same vein as assembly or unchecked in solidity).

> Additional background: I think it's an improvement to specify this on the command line, vs. having to edit my .env if I want to run the bot in a different personality.

I think this is probably a good call -- definitely something for a separate PR. We'll need to think about backcompat, especially for the relayer since there are many other known operators :)

src/dataworker/Dataworker.ts
pxrl: > On the margin, I think flexibility is preferable here, and if you are not using the correct fromBlock and use this to actually propose, you should be disputed.

This sounds reasonable :+1: Would we consider some special case where a range of `[-1, x] ` implies "start from the previous" ?

mrice32: Seems reasonable, but may also be overkill. Not super opinionated!

mrice32 (commented):
LGTM once the outstanding comment threads are resolved!

src/dataworker/DataworkerConfig.ts
```diff
@@ -91,11 +98,39 @@ export class DataworkerConfig extends CommonConfig {
...
+      // positive whole numbers and start < end.
+      this.forceProposalBundleRange.forEach((bundleRange, index) => {
+        assert(Array.isArray(bundleRange), `forceProposalBundleRange[${index}] is not an array`);
+        assert(bundleRange.length === 2, `forceProposalBundleRange[${index}] does not have length 2`);
+        const [start, end] = bundleRange;
+        assert(typeof start === "number", `forceProposalBundleRange[${index}][start] is not a number`);
+        assert(typeof end === "number", `forceProposalBundleRange[${index}][end] is not a number`);
+        assert(start > 0, `forceProposalBundleRange[${index}][start] is not positive`);
+        assert(end > 0, `forceProposalBundleRange[${index}][end] is not positive`);
+        assert(start < end, `forceProposalBundleRange[${index}][start] >= forceProposalBundleRange[${index}][end]`);
```
pxrl, line 117: I tried to configure a manual proposal range, but it threw here on Boba. What's the expected behaviour for when a chain has been disabled?

james-a-morris: Handled: I've changed the range parameter.

src/dataworker/DataworkerConfig.ts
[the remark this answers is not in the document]
james-a-morris: > I think of it as a powerful admin tool that you should only use if you know what you're doing (in the same vein as assembly or unchecked in solidity).

I like this way of looking at the logic. Resolved to remove this constraint in the dataworker config

src/dataworker/Dataworker.ts
james-a-morris: This might be overengineering. This feature is meant to provide concise controls over a range, with the result being that the operator needs to be sure of what they are doing. I think adding any additional logic would be accounting for this as a feature rather than an admin/debug tool.

@pxrl do you agree?

.env.example
james-a-morris: What would differing examples look like? I.e. should we put a real-world example in?

My gut feel is that this segment of the `.env.sample` should require the operator to know exactly what to place and where. IMO - this section serves as an indicator of what ENV vars exist, not necessarily how to construct them as they are fairly one-off in that providing an example proposal range wouldn't necessarily be helpful for a different range.

mrice32 (approved):
LGTM once @pxrl gives the go-ahead on the two outstanding threads that stemmed from his comments!

.env.example
pxrl: > What would differing examples look like? I.e. should we put a real-world example in?
> 
> My gut feel is that this segment of the `.env.sample` should require the operator to know exactly what to place and where. IMO - this section serves as an indicator of what ENV vars exist, not necessarily how to construct them as they are fairly one-off in that providing an example proposal range wouldn't necessarily be helpful for a different range.

Yeah, I hear you on the issue of `.env.example` being relayer-specific. This is how it's historically been used, but it doesn't necessarily mean it's correct.

I do still think it's important to offer relayers an _almost works out of the box_ example/template, but it doesn't have to be `.env.example`. I'd be equally happy if `.env.example` serves as the complete variant, and then we have a `.env.relayer` that relayer can copy/paste to use. I just really think it's important that they can do that without having to pick their way through dataworker-specific config.

src/dataworker/Dataworker.ts
pxrl: The background on my request is that I was trying to race the proposer during a recent issue.

It's actually a moot point now, since only 0x7f has the ability to propose after the recent changes. So I've lost the use case that was behind my request. I'll mark this as resolved!

pxrl (approved):
OK from me :+1: I'd just like to conclude on this thread, relating to relayer convenience https://github.com/across-protocol/relayer-v2/pull/912#discussion_r1329222856

.env.example
james-a-morris: Ah I see. That makes total sense. In this case, should we do so in a different PR - maybe one that is directed toward updating documentation?

james-a-morris: Created ACX-1545 so this doesn't get lost.

PR #917, opened 2023-09-01
improve(relayer): Permit ignoring origin chains

pxrl:
This is a simple relayer-specific config to permit ignoring deposits based on their origin chain ID.

Note: This is implemented at the relayer level because the anticipated UBA rollout will require relayers to validate events on _all_ supported chains, not only the ones they wish to support.

Closes ACX-1516

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -67,6 +67,20 @@ export class Relayer {
     return supportedDeposits;
   }
 
+  routeEnabled(originChainId: number, destinationChainId: number): boolean {
+    const { relayerOriginChains: originChains, relayerDestinationChains: destinationChains } = this.config;
+
+    if (originChains?.length > 0 && !originChains.includes(originChainId)) {
+      return false;
+    }
+
+    if (destinationChains?.length > 0 && !destinationChains.includes(destinationChainId)) {
+      return false;
+    }
+
+    return true;
+  }
+
   async checkForUnfilledDepositsAndFill(sendSlowRelays = true): Promise<void> {
     // Fetch all unfilled deposits, order by total earnable fee.
     const { config } = this;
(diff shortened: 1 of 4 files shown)


src/relayer/Relayer.ts
```diff
@@ -152,18 +178,6 @@ export class Relayer {
...
-      if (relayerDestinationChains.length > 0 && !relayerDestinationChains.includes(destinationChainId)) {
-        this.logger.debug({
-          at: "Relayer",
-          message: "Skipping deposit for unsupported destination chain",
-          deposit,
-          destinationChain,
-        });
-        continue;
-      }
-
```
pxrl, line 166: This was elevated to be _above_ the "supported token" validation because it's cheaper to validate the chain Ids than resolve a HubPool token from a SpokePool token, so it might save some time for operators who restrict chains but not tokens.

src/relayer/Relayer.ts
```diff
@@ -138,11 +152,23 @@ export class Relayer {
...
     // unprofitable tx to the unprofitable tx tracker to produce an appropriate log.
     for (const { deposit, version, unfilledAmount, fillCount, invalidFills } of confirmedUnfilledDeposits) {
-      const { relayerDestinationChains, relayerTokens, slowDepositors } = config;
+      const { relayerTokens, slowDepositors } = config;
+
+      const { originChainId, destinationChainId, originToken } = deposit;
+      const destinationChain = getNetworkName(destinationChainId);
+
+      if (!this.routeEnabled(originChainId, destinationChainId)) {
+        this.logger.debug({
```
pxrl, line 161: This is likely to be really noisy...we might want to aggregate the skipped deposits and log only a few key details from each one in a single message.

mrice32 (approved):
LGTM! Great how you precisely matched the behavior of the other env var -- very nice!

PR #918, opened 2023-09-01
fix: Avoid logging entire Error object

pxrl:
The logging subsystem doesn't handle Error objects properly.


pxrl:
2023-09-01-173048_616x103_scrot

mrice32 (approved):
LGTM!

PR #919, opened 2023-09-01
improve(SpokePoolClient): Be more robust to _update() failures

pxrl:
Some upstream RPC providers are giving the SpokePoolClient a hard time...

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,3 +1,39 @@
-import { clients } from "@across-protocol/sdk-v2";
+import { clients, typeguards } from "@across-protocol/sdk-v2";
+import { Event, providers } from "ethers";
+import { getNetworkName } from "../utils";
 
-export class SpokePoolClient extends clients.SpokePoolClient {}
+// @dev Duplicated from @across-protocol/sdk-v2/src/clients/SpokePoolClient.ts
+type _SpokePoolUpdate = {
+  success: boolean;
+  currentTime: number;
+  firstDepositId: number;
+  latestBlockNumber: number;
+  latestDepositId: number;
+  events: Event[][];
+  blocks: { [blockNumber: number]: providers.Block };
+  searchEndBlock: number;
+};
+type SpokePoolUpdate = { success: false } | _SpokePoolUpdate;
+
+class SpokePoolClient extends clients.SpokePoolClient {
+
+   // protected async _update(eventsToQuery: string[]): Promise<SpokePoolUpdate> {
+   protected override async _update(eventsToQuery: string[]): Promise<SpokePoolUpdate> {
+    const network = getNetworkName(this.chainId);
+    let i: number;
+    for (i = 0; i < 3; ++i) {
+      try {
+        return await super._update(eventsToQuery);
+      } catch (err) {
+        this.logger.info({
+          at: "SpokePoolClient::_update",
+          message: "Caught unhandled exception in SpokePoolClient::_update()",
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -1,3 +1,39 @@
...
+};
+type SpokePoolUpdate = { success: false } | _SpokePoolUpdate;
+
+class SpokePoolClient extends clients.SpokePoolClient {
+
+   // protected async _update(eventsToQuery: string[]): Promise<SpokePoolUpdate> {
+   protected override async _update(eventsToQuery: string[]): Promise<SpokePoolUpdate> {
+    const network = getNetworkName(this.chainId);
+    let i: number;
+    for (i = 0; i < 3; ++i) {
```
james-a-morris, line 24: I wonder if we should put this directly into the SDK?

Also, it may be beneficial to store the number of retries in a constant

src/clients/SpokePoolClient.ts
```diff
@@ -1,3 +1,39 @@
...
+      try {
+        return await super._update(eventsToQuery);
+      } catch (err) {
+        this.logger.info({
+          at: "SpokePoolClient::_update",
+          message: "Caught unhandled exception in SpokePoolClient::_update()",
```
james-a-morris, line 30: Can we add the number of retries thus far?

I.e. "Caught unhandled exception on run 1/4..."

james-a-morris (commented):
Left some comments

src/clients/SpokePoolClient.ts
pxrl: I'd like to try it in the relayer first, because the RTT for getting it into the SDK and making and subsequent adjustments is unappetising at the moment. We may be able to simply remove this band-aid once the current issues are gone, but we may also decide to retain it.

src/clients/SpokePoolClient.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/919/commits/8d06523d5d461b7080ffca4acf73d3a162f3fc35

src/clients/SpokePoolClient.ts
```diff
@@ -1,3 +1,38 @@
...
+  latestDepositId: number;
+  events: Event[][];
+  blocks: { [blockNumber: number]: providers.Block };
+  searchEndBlock: number;
+};
+type SpokePoolUpdate = { success: false } | _SpokePoolUpdate;
+
+const maxTries = 5;
+
+export class SpokePoolClient extends clients.SpokePoolClient {
```
james-a-morris, line 20: Should we override the constructor to include maxTries?

pxrl: If it was intended to be permanent then I'd do that, but since it's only temporary code that attempts to work around a current issue, I don't really think it's worth doing that. Ultimately this code would be removed, and it's just easier to do that by nuking these lines without having to modify any places where this is instantiated.

mrice32 (approved):
LGTM assuming this is only meant to be temporary!

pxrl:
Have marked this as `do not merge` for now, since the issues _might_ be perstistent and related to a specific provider, in which case retrying won't actually help. Will leave it open though, until this is confirmed.

pxrl:
Closing this one since we've had a very quiet and stable weekend after rotating a single RPC provider out of the hot path, indicating that the issues we experienced were persistent with a single provider, rather than transient.

PR #923, opened 2023-09-05
improve: reduce errors in relayer tests

james-a-morris:


test/Dataworker.loadData.ts
```diff
@@ -729,7 +728,10 @@ describe("Dataworker: Load data used in all functions", async function () {
     );
     const blockNumber = await spokePool_2.provider.getBlockNumber();
     const blockTimestamp = (await spokePool_2.provider.getBlock(blockNumber)).timestamp;
-    const realizedLpFeePctData = await hubPoolClient.computeRealizedLpFeePct(deposit1, l1Token_1.address);
+    const realizedLpFeePctData = await hubPoolClient.computeRealizedLpFeePct(
+      { ...deposit1, blockNumber },
```
mrice32, line 732: Is this a functional change?

test/utils/utils.ts
```diff
@@ -272,138 +286,43 @@ export async function simpleDeposit(
   };
 }
 
-export async function deploySpokePoolForIterativeTest(
```
mrice32, line 275: Were all of these unused?

mrice32 (approved):
Looks good! Just want to double check that there are no functional changes here, and this is only a refactor. Is that the intention?

test/utils/utils.ts
james-a-morris: Yep!

test/Dataworker.loadData.ts
james-a-morris: This PR is related to refactorings & slight rewrites to get the software to be able to compile with tsc. 

Prior to this, we were having actual compiler errors due to signature incongruence on function/constructor calls.

james-a-morris:
@mrice32 resolving as we discussed offline.

PR #927, opened 2023-09-07
improve: Don't trigger quorum fights on zkSync l1 fields

pxrl:
Testing shows that these fields can be updated shortly after the block has been produced.


pxrl:
Am testing this locally but wanted to submit it now to get a head start on CI.

james-a-morris (approved):
Good call - LGTM once the linting errors are resolved.

PR #928, opened 2023-09-07
feat: enable linting for floating promises.

james-a-morris:
This PR adds ESLinting for floating promises. Floating promises are defined as any function that resolves an unhandled promise. A handled promise can either be returned directly, awaited, then/catch'd, or voided.

This change would have immediately solved https://github.com/across-protocol/relayer-v2/commit/723bdcfae365051fcb05a089ebf4e6fc998bb2b0. We can now enforce that un-awaited promises will never be uncaught again.

### DIFF
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -84,7 +84,7 @@ export async function disconnectRedisClient(logger?: winston.Logger): Promise<vo
   if (redisClient !== undefined) {
     // todo understand why redisClient isn't GCed automagically.
     logger.debug("Disconnecting from redis server.");
-    redisClient.disconnect();
+    await redisClient.disconnect();
   }
 }
 
--- a/test/UBAClient.validateFlow.ts
+++ b/test/UBAClient.validateFlow.ts
@@ -422,11 +422,11 @@ describe("UBAClient: Flow validation", function () {
           // Don't add inflow to uba bundle state.
 
           // Throws errors if inflow is before, after, or in same bundle as outflow.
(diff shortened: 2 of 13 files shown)


test/UBAClient.validateFlow.ts
```diff
@@ -422,11 +422,11 @@ describe("UBAClient: Flow validation", function () {
           // Don't add inflow to uba bundle state.
 
           // Throws errors if inflow is before, after, or in same bundle as outflow.
-          assertPromiseError(ubaClient.validateFlow(outflow), "Could not find matched deposit in same bundle");
+          void assertPromiseError(ubaClient.validateFlow(outflow), "Could not find matched deposit in same bundle");
```
mrice32, line 425: why don't we await these promises?

src/utils/RedisUtils.ts
```diff
@@ -84,7 +84,7 @@ export async function disconnectRedisClient(logger?: winston.Logger): Promise<vo
   if (redisClient !== undefined) {
     // todo understand why redisClient isn't GCed automagically.
     logger.debug("Disconnecting from redis server.");
-    redisClient.disconnect();
+    await redisClient.disconnect();
```
mrice32, line 87: Nice find!

mrice32 (approved):
Great addition!

test/UBAClient.validateFlow.ts
james-a-morris: These are in the tests and purely serve to throw errors

pxrl (approved):

PR #930, opened 2023-09-07
feat: enforce async functions to have await

james-a-morris:
This PR enforces that all functions marked with `async` must have at least one `await` within them. This is for two reasons: firstly, it provides a layer of code consistency. Secondly (and more importantly), it can help prevent accidental cases where we should use await but don't.

### DIFF
--- a/scripts/utils.ts
+++ b/scripts/utils.ts
@@ -19,7 +19,7 @@ const fallbackProviders: { [chainId: number]: string } = {
   5: "https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161",
 };
 
-async function askQuestion(query: string) {
+function askQuestion(query: string) {
   const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
 
   return new Promise((resolve) =>
--- a/scripts/utils.ts
+++ b/scripts/utils.ts
@@ -115,7 +115,7 @@ export function resolveHubChainId(spokeChainId: number): number {
  * @param contractName Name of the deployed contract.
  * @returns ethers Contract instance.
  */
-export async function getContract(chainId: number, contractName: string): Promise<Contract> {
+export function getContract(chainId: number, contractName: string): Contract {
   const contract = getDeployedContract(contractName, chainId);
   const provider = new ethers.providers.StaticJsonRpcProvider(getProviderUrl(chainId));
   return contract.connect(provider);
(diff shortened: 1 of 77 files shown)


james-a-morris:
Waiting for https://github.com/across-protocol/sdk-v2/pull/383 to resolve this failed test.

src/dataworker/Dataworker.ts
```diff
@@ -1873,7 +1873,7 @@ export class Dataworker {
     });
   }
 
-  async _updateExchangeRates(l1Tokens: string[], submitExecution: boolean): Promise<void> {
+  _updateExchangeRates(l1Tokens: string[], submitExecution: boolean): void {
```
nicholaspai, line 1876: Are there any calls to `_updateExchangeRates` that needlessly `await` that we can change?

src/relayer/Relayer.ts
```diff
@@ -565,7 +565,7 @@ export class Relayer {
     return profitable ? preferredChainId : undefined;
   }
 
-  protected async computeRealizedLpFeePct(version: number, deposit: DepositWithBlock): Promise<BigNumber> {
+  protected computeRealizedLpFeePct(version: number, deposit: DepositWithBlock): BigNumber {
```
nicholaspai, line 568: Same question as https://github.com/across-protocol/relayer-v2/pull/930/files#r1327228115 but just want to double check that you've removed any `await` calls to this function

nicholaspai (approved):
A couple questions to spur second checks

scripts/utils.ts
```diff
@@ -115,7 +115,7 @@ export function resolveHubChainId(spokeChainId: number): number {
  * @param contractName Name of the deployed contract.
  * @returns ethers Contract instance.
  */
-export async function getContract(chainId: number, contractName: string): Promise<Contract> {
+export function getContract(chainId: number, contractName: string): Contract {
```
mrice32, line 118: Looks like there's a call below in this file that awaits this function. Is there a linter rule that can prevent us from calling await on a non-promise?

src/dataworker/Dataworker.ts
mrice32: Would be great if this could be linted automatically.

mrice32 (approved):
LGTM! Would be great if the linter could catch cases where we needlessly await.

james-a-morris:
This is no longer needed in favor of co-PRs that have been added to the SDK & Relayer

PR #932, opened 2023-09-10
fix: handle early deposits

mrice32:


src/utils/DepositUtils.ts
```diff
@@ -93,6 +93,32 @@ export function getUniqueDepositsInRange(
...
+  originClient: SpokePoolClient,
+  existingUniqueDeposits: typechain.FundsDepositedEvent[]
+): typechain.FundsDepositedEvent[] {
+  const originChainBlockRange = getBlockRangeForChain(
+    blockRangesForChains,
+    originChain,
+    chainIdListForBundleEvaluationBlockNumbers
+  );
+  return (originClient["earlyDeposits"] as unknown as typechain.FundsDepositedEvent[]).filter(
+    (deposit: typechain.FundsDepositedEvent) =>
```
james-a-morris, line 110: Nit: with the typecast above, do we need to specify this function param?

james-a-morris (approved):
Left one nit. I helped with typecasting & unit tests so a second approval may be in order. Otherwise, this code seems good & it passes on my integration tests locally

src/utils/DepositUtils.ts
mrice32: @james-a-morris maybe not, but just in the interest of time, let's do this in a follow-up.

james-a-morris:
LGTM

PR #935, opened 2023-09-11
fix: Correct scripts contract resolution utility

pxrl:
Was recently broken during a refactor.


james-a-morris (approved):
LGTM - it may be useful in another PR, but we should probably add an enum type to `getContract` so we can enforce only specific contracts

mrice32 (approved):

PR #936, opened 2023-09-11
fix(Dataworker): Dereference originChainId by index, not key

pxrl:
Temporary fix to resolve an issue with future quoteTimestamps. We need a more robust solution than this in the long run.

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -85,7 +85,7 @@ export function updateRunningBalanceForEarlyDeposit(
   updateAmount: BigNumber
 ): void {
   // deposit.args is a pure array; there are no mapping to be dereferenced.
-  const originChainId = deposit.args[1].toNumber();
+  const originChainId = Number(deposit.args[1].toString());
   const originToken = deposit.args[6];
 
   const l1TokenCounterpart = hubPoolClient.getL1TokenCounterpartAtBlock(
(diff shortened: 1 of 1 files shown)


pxrl:
Hm; this is actually failing for me in local test, but I believe this fix is just unearthing a separate issue.

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -84,17 +84,21 @@ export function updateRunningBalanceForEarlyDeposit(
   deposit: typechain.FundsDepositedEvent,
   updateAmount: BigNumber
 ): void {
+  // deposit.args is a pure array; there are no mapping to be dereferenced.
+  const originChainId = deposit.args[1].toNumber();
```
james-a-morris, line 88: We may want to wrap this as `Number(String(deposit.args[1]))` to ensure that we won't get a null reference. Alternatively - we can ensure that deposit.args[1] !== undefined

james-a-morris (commented):
LGTM - pending linter

mrice32 (approved):

src/dataworker/PoolRebalanceUtils.ts
mrice32: Will this work on a BN?

mrice32: I think .toString() will work on a BigNumber, not sure about String()

mrice32 (approved):
https://github.com/across-protocol/relayer-v2/pull/936#discussion_r1322129057

PR #938, opened 2023-09-15
improve: leverage default TTL & leverage caching

james-a-morris:
This PR enabled the ability for the hubpool client to utilize caching. Additionally, this change sets the RedisClient to use the default 2 week TTL if not provided.

### DIFF
--- a/src/caching/RedisCache.ts
+++ b/src/caching/RedisCache.ts
@@ -1,4 +1,4 @@
-import { interfaces } from "@across-protocol/sdk-v2";
+import { interfaces, constants } from "@across-protocol/sdk-v2";
 import { RedisClient, getRedis, objectWithBigNumberReviver, setRedisKey, winston } from "../utils";
 
 /**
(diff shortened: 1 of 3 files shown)


src/caching/RedisCache.ts
```diff
@@ -67,7 +67,7 @@ export class RedisCache implements interfaces.CachingMechanismInterface {
       await this.instantiate();
     }
     // Call the setRedisKey function to set the value in redis.
-    await setRedisKey(key, JSON.stringify(value), this.redisClient, ttl);
+    await setRedisKey(key, JSON.stringify(value), this.redisClient, ttl ?? constants.DEFAULT_CACHING_TTL);
```
nicholaspai, line 70: why not set `ttl = constants.DEFAULT_CACHING_TTL` in the function declaration?

nicholaspai (approved):
One comment.

src/caching/RedisCache.ts
james-a-morris: I'm not super opinionated on this. I wanted to keep the function signatures as close to the interface as possible. However, I like this approach because if we ever decide to expand on the logic we only need to define the default once.

james-a-morris: Resolved

PR #940, opened 2023-09-20
refactor: Deprecate NODE_URL[S]_<CHAIN-ID> config

pxrl:
This change deprecates the ability to configure RPC providers with the old NODE_URL[S]_<CHAIN-ID> syntax. Instead, the newer RPC_PROVIDERS format is enforced.

Both formats have been supported for over 4 months now, but we have long since migrated to the new format in the bots.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -569,12 +569,13 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
   return provider;
 }
 
-export function getNodeUrlList(chainId: number, quorum: number): string[] {
+export function getNodeUrlList(chainId: number, quorum = 1): string[] {
   const resolveUrls = (): string[] => {
     const providers = process.env[`RPC_PROVIDERS_${chainId}`] ?? process.env["RPC_PROVIDERS"];
     if (providers === undefined) {
       throw new Error(`No RPC providers defined for chainId ${chainId}`);
     }
+
     const nodeUrls = providers.split(",").map((provider) => {
       const envVar = `RPC_PROVIDER_${provider}_${chainId}`;
       const url = process.env[envVar];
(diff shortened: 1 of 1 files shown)


pxrl:
Background to this change: I'd like to migrate this function over to the SDK so that we can use it for some of the e2e tests there. It seemed like an opportunity to simplify the implementation post-migration.

src/utils/ProviderUtils.ts
```diff
@@ -569,12 +569,13 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
...
 }
 
-export function getNodeUrlList(chainId: number, quorum: number): string[] {
+export function getNodeUrlList(chainId: number, quorum = 1): string[] {
   const resolveUrls = (): string[] => {
     const providers = process.env[`RPC_PROVIDERS_${chainId}`] ?? process.env["RPC_PROVIDERS"];
     if (providers === undefined) {
       throw new Error(`No RPC providers defined for chainId ${chainId}`);
     }
+
```
james-a-morris, line 578:
```suggestion
```

src/utils/ProviderUtils.ts
```diff
@@ -591,29 +592,8 @@ export function getNodeUrlList(chainId: number, quorum: number): string[] {
...
-      if (process.env[nodeUrlKey]) {
-        nodeUrls = [process.env[nodeUrlKey]];
-      }
-    }
-  }
-
-  if (nodeUrls.length === 0) {
-    throw new Error(`Can't get ${chainId} node url(s) because neither ${retryConfigKey} or ${nodeUrlKey} are defined.`);
-  } else if (nodeUrls.length < quorum) {
+  const nodeUrls = resolveUrls();
```
james-a-morris, line 595: QQ: why are we removing the try/catch logic here?

james-a-morris (commented):
Left a few comments

src/utils/ProviderUtils.ts
pxrl: This was used to prefer the new config format, but to fall back to the old one if it was still in use. If we remove the old format then it's instead desirable to throw with the underlying error.

src/utils/ProviderUtils.ts
pxrl: That was intentional because the closure is kinda dense IMO; the linter seems OK with it so I'd rather leave it in unless there are hard objections.

nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #944, opened 2023-09-20
feat: Add protection against computing LP fees for future quote timestamps

nicholaspai:
This resolves the [reverted SDK PR](https://github.com/across-protocol/sdk-v2/pull/399) and re-introduces protection from deposit quote timestamps that are impossible to compute.

This PR essentially moves that protection to the production version of Hub and Spoke clients so that the unit tests can continue to work and not have to add additional logic to set quote times correctly. Ideally in the future we set the quote times how they would in production but for now this solution results in a much smaller code diff

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -1,5 +1,5 @@
 import { clients } from "@across-protocol/sdk-v2";
-import { Contract } from "ethers";
+import { BigNumber, Contract } from "ethers";
 import winston from "winston";
 import { MakeOptional, EventSearchConfig } from "../utils";
 import { IGNORED_HUB_EXECUTED_BUNDLES, IGNORED_HUB_PROPOSED_BUNDLES } from "../common";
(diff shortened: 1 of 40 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -18,4 +19,20 @@ export class HubPoolClient extends clients.HubPoolClient {
...
+    l1Token: string
+  ): Promise<{ realizedLpFeePct: BigNumber | undefined; quoteBlock: number }> {
+    if (deposit.quoteTimestamp > this.currentTime) {
+      throw new Error(
+        `Cannot compute lp fee percent for quote timestamp ${deposit.quoteTimestamp} in the future. Current time: ${this.currentTime}.`
+      );
+    }
+
+    return await super.computeRealizedLpFeePct(deposit, l1Token);
+  }
```
james-a-morris, line 37: Could we add this check directly into the SDK?

james-a-morris (commented):
Left a comment

src/clients/HubPoolClient.ts
nicholaspai: Well, we could, but it seriously messes up the tests :) The main point of putting it here is we don't actually test this client. A lot of the tests make don't opine on the quote timestamps so we'd have to change a lot

james-a-morris (approved):

PR #946, opened 2023-09-22
improve: Simulate finalization txns before failure

pxrl:
Simulate all finalizations before bundling them into the multicall aggregate() transaction, which will fail in aggregate if even a single transaction is unsuccessful. This caused problems today, when a small number of zkSync withdrawals were unsuccessful, resulting in a finalizer halt for a few hours.

Transactions are simulated first for logging purposes; it would be possible to user multicall3.tryAggregate(false, ...), except it's non-trivial to extract the results of each transaction.

Fixes ACX-1551

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -1,4 +1,5 @@
 import assert from "assert";
+import { typeguards, utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { groupBy } from "lodash";
 import {
   Wallet,
@@ -27,6 +28,8 @@ import {
   Multicall2Call,
 } from "../common";
 import { ChainFinalizer, Withdrawal } from "./types";
+
+const { isError, isEthersError } = typeguards;
 
 config();
 let logger: winston.Logger;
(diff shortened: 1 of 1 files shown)


pxrl:
Small example of the logging that will be emitted on a simulation failure (produced via forcing the zkSync adapter to include transactions that were already finalised). Note that the "xx" revert reason is particular to the zkSync contracts, and indicates a specific failure.

```
2023-09-22 22:52:49 [info]: {
  "at": "finalizer",
  "message": "Failed to estimate gas for ZkSync 44.63 WETH withdrawal.",
  "reason": "execution reverted: xx",
  "bot-identifier": ""
}
2023-09-22 22:52:50 [info]: {
  "at": "finalizer",
  "message": "Failed to estimate gas for ZkSync 51,336.17 USDC withdrawal.",
  "reason": "execution reverted: xx",
  "bot-identifier": ""
}
```

nicholaspai (approved):

james-a-morris (approved):

PR #948, opened 2023-09-25
improve: Split CI test

evaldofelipe:
Migrate the current CI tests from GitHub Actions, to CircleCI due the increase of speed


mrice32:
Is this ready to be reviewed?

test/split.ts
```diff
@@ -0,0 +1,40 @@
+import glob from 'glob';
+
+if (!process.env.CI_TOTAL) {
+    console.error('No environment variable CI_TOTAL defined');
+    process.exit(1);
```
pxrl, line 5: Minor FWIW...I had success in [another script](https://github.com/across-protocol/relayer-v2/blob/c210556cf387a4ae807af32fe92b9829fec43622/scripts/hubpool.ts#L265) by just setting `process.exitCode` - this is the "node recommended" way: https://nodejs.org/api/process.html#processexitcode

2023-09-29-115731_1618x378_scrot

james-a-morris (commented):
Should we also do this in the SDK? Also, will this produce a report to github so that external users can verify that tests pass?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+          command: yarn build
+      - run:
+          name: Run lint
+          command: yarn lint
+workflows:
+  version: 2.1
+  test_and_lint:
+    jobs:
+      - test
+      - lint
```
mrice32, line 69: Where's build? 

Shouldn't these also "require" build? So that build always runs first (so we can be sure the cache exists)?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+          name: Install Prereqs
+          command: sudo apt-get update && sudo apt-get install -y rsync libudev-dev libusb-1.0-0-dev yarn
+      - run:
+          name: Install Dependencies
+          command: yarn install
+      - run:
+          name: Run build
+          command: yarn build
+      - save_cache:
+          key: relayer-cache-{{ .Environment.CIRCLE_SHA1 }}
```
mrice32, line 19: Is this cach used anywhere? It looks like we're saving to it and never reading from it?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+      - checkout
+      - run:
+          name: Install Prereqs
+          command: sudo apt-get update && sudo apt-get install -y rsync libudev-dev libusb-1.0-0-dev yarn
+      - run:
+          name: Install Dependencies
+          command: yarn install
+      - run:
+          name: Run build
+          command: yarn build
```
mrice32, line 38: Why reinstall and rebuild? Aren't we supposed to do that in build?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
```
mrice32, line 60: Same -- why reinstall and rebuild?

mrice32 (commented):
A few comments -- not sure these jobs are configured correctly. I think the build job is unused, the cache is disabled, and install + build is repeated on each step.

.circleci/config.yml
evaldofelipe: Just fixed it. We are not using a step to `build`. It's quicker to build on each time, better than build, upload and download.

.circleci/config.yml
evaldofelipe: Same comment as https://github.com/across-protocol/relayer-v2/pull/948#discussion_r1348029417

.circleci/config.yml
evaldofelipe: Same comment as https://github.com/across-protocol/relayer-v2/pull/948#discussion_r1348029417

mrice32 (approved):
@evaldofelipe I made some minor changes [here](https://github.com/across-protocol/relayer-v2/pull/948/commits/6b74f71052697ea8a88047c292b3f31181315df1). Let me know if you disagree. Otherwise, looks great!

.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
...
+    working_directory: ~/relayer-v2
+    resource_class: medium+
+    parallelism: 20
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - cache-node-v1-{{ checksum "yarn.lock" }}
+      - run:
```
mrice32, line 34: Removed yarn install from the test and lint jobs, since that's already done by the cache.

.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
...
+    docker:
+      - image: cimg/node:16.18.0
+    working_directory: ~/relayer-v2
+    resource_class: medium+
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - cache-node-v1-{{ checksum "yarn.lock" }}
```
mrice32, line 13: Added v1 to all of these because that allows us to increment the number to flush the cache.

.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
+version: 2.1
+jobs:
+  install:
```
mrice32, line 3: Changed the name of this to install instead of build, since we aren't building, we are just installing deps.

PR #950, opened 2023-09-27
feat: Update constants for Polyon Mumbai

pxrl:
Note - we still need to bump contracts-v2 to get Mumbai contract addresses.


james-a-morris (approved):
Good addition

PR #951, opened 2023-09-27
improve: Simplify wallet key type

pxrl:
The type is correctly filtered within getSigner(), but it causes unnecessary type annoyances when reading the key type from the environment or on the command-line.

### DIFF
--- a/src/utils/SignerUtils.ts
+++ b/src/utils/SignerUtils.ts
@@ -8,7 +8,7 @@ export type SignerOptions = {
    * The type of wallet to use.
    * @note If using a GCKMS wallet, the gckmsKeys parameter must be set.
    */
-  keyType: "mnemonic" | "privateKey" | "gckms";
+  keyType: string;
   /**
    * Whether or not to clear the mnemonic/private key from the env after retrieving the signer.
    * @note Not including this parameter or setting it to false will not clear the mnemonic/private key from the env.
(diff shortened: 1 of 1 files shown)


src/utils/SignerUtils.ts
```diff
@@ -8,7 +8,7 @@ export type SignerOptions = {
    * The type of wallet to use.
    * @note If using a GCKMS wallet, the gckmsKeys parameter must be set.
    */
-  keyType: "mnemonic" | "privateKey" | "gckms";
+  keyType: string;
```
pxrl, line 11: FWIW I previously thought it was nice to lock this down, but I now think it's unnecessary.

james-a-morris: Sounds good - this is fine. Is there a reason why this change is being made or is it a driveby comment?

james-a-morris (approved):

mrice32 (approved):

src/utils/SignerUtils.ts
pxrl: During the hackathon I was playing with argument parsing for the spokepool script, and tsc was unhappy that a string parsed on the cmdline didn't fit 1:1 with the keyType defintion.

FWIW, you previously raised a question about why the keyType isn't specified in `.env`. I'm probably inclined to a agree - it makes more sense to me for it to be specified in the same place as `MNEMONIC` or `PRIVATE_KEY`.

PR #953, opened 2023-09-28
fix(finalizer): Handle 1:many withdrawal:calldata

pxrl:
The previous change to the finalizer incorrectly assumed that there would be a 1:1 relationship between a txn and a withdrawal or proof. This is incorrect on Polygon, where 2 txns are needed to finalize a single withdrawal.

This change seems like a manageable way of establishing a 1:many relationship between withdrawal/proof logging and the necessary transaction(s), without imposing any changes on the underlying chain adapters. Ultimately it would be nice to backport this into the adapters themselves, but that's invasive and it's not worth prioritising now.

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -114,6 +114,10 @@ export async function finalize(
       message: `Found ${_withdrawals.length} ${network} withdrawals for finalization.`,
     });
 
+    if (_withdrawals.length === 0) {
+      continue;
+    }
+
     if (![1, 2].includes(txns.length / _withdrawals.length)) {
       logger.warn({
         at: "finalize",
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -32,7 +32,7 @@ import { ChainFinalizer, Withdrawal as _Withdrawal } from "./types";
 
 type TransactionReceipt = providers.TransactionReceipt;
 
-type Withdrawal = _Withdrawal & { calldata: Multicall2Call[] };
+type Withdrawal = _Withdrawal & { txns: Multicall2Call[] };
 
 const { isError, isEthersError } = typeguards;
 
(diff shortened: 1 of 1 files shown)


src/finalizer/index.ts
```diff
@@ -100,75 +102,99 @@ export async function finalize(
...
     );
-    logger.debug({ at: "finalize", message: `Found ${callData.length} ${network} withdrawals for finalization.` });
+    logger.debug({
+      at: "finalize",
+      message: `Found ${_withdrawals.length} ${network} withdrawals for finalization.`,
+    });
 
-    const txns = callData.map((callData, i) => {
-      return { callData, withdrawal: withdrawals[i] };
+    if (![1, 2].includes(txns.length / _withdrawals.length)) {
```
pxrl, line 117: Missing length check prior to this causes divide by 0 when there are no withdrawals to finalise.

src/finalizer/index.ts
```diff
@@ -100,75 +102,103 @@ export async function finalize(
...
+    finalizationsToBatch.push(...withdrawals);
   }
 
   // Ensure each transaction would succeed in isolation.
   const finalizations = await sdkUtils.filterAsync(finalizationsToBatch, async (finalization) => {
+    const { calldata } = finalization;
     try {
-      const { target: to, callData: data } = finalization.callData;
-      await multicall2.provider.estimateGas({ to, data });
+      const txn = await multicall2.populateTransaction.aggregate(calldata);
```
pxrl, line 148: This is broken - @todo fix.

src/finalizer/index.ts
```diff
@@ -100,75 +102,103 @@ export async function finalize(
...
+    // Normalise withdawals, such that 1 withdrawal has an array of calldata (usually only 1 call), but can be more.
+    // @todo: Refactor the underlying adapters so they return in this data structure.
+    const withdrawals: Withdrawal[] = _withdrawals.map((withdrawal) => {
+      return { ...withdrawal, calldata: [] };
     });
 
-    finalizationsToBatch.push(...txns);
+    // Append calldata. If multiple calls are needed per withdrawal (i.e. Polygon),
+    // require that the 2nd batch is appended to the first.
+    txns.forEach((txn, i) => withdrawals[i % withdrawals.length].calldata.push(txn));
```
nicholaspai, line 139: what if a future network uses more than 2 calls?

pxrl: As long as they are returned as batches that are multiples of `withdrawals`, i.e. batch 1, batch 2, batch 3, in the same order, then it should be OK.

ATM it seems unlikely that we'd require _more_ than 2 transactions for a single withdrawal, and the fact that we currently need 2 seems to be exclusively a quirk of Polygon. Ultimately I'd like to go back into each of the adapters so that they enforce the 1:many relationship between withdrawal and multicall transactions, but I don't think that's a priority at the moment.

FWIW, I've been front-running the production finalizer successfully for the past ~24 hours, and it's working as expected.

nicholaspai (approved):

pxrl:
I've run this continuously over the weekend. Aside from a small redis issue (unrelated to this change), the change has worked 100% as intended). Will submit as soon as CI passes.

nicholaspai (approved):

PR #956, opened 2023-09-29
refactor: Local definition for TokenRolesEnum

pxrl:
This seems to be entirely internal, so no need to export it or define it centrally.


pxrl:
Closing this one, since this will be migrated to sdk-v2.

PR #959, opened 2023-09-29
refactor: Remove UMA interfaces lookup

pxrl:
These names change very rarely, and we only have need for a small number of the complete set.

### DIFF
--- a/test/TokenClient.Approval.ts
+++ b/test/TokenClient.Approval.ts
@@ -1,4 +1,3 @@
-import { interfaceName } from "@uma/common";
 import { HubPoolClient, SpokePoolClient, TokenClient } from "../src/clients";
 import {
   Contract,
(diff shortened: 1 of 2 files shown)


nicholaspai (commented):
@james-a-morris can you look at this and make sure we don't overwrite this in the SDK changes?

pxrl:
nb. We could alternatively define this centrally somewhere

> @james-a-morris can you look at this and make sure we don't overwrite this in the SDK changes?

Ah good call, maybe this is migrated to sdk anyway.

mrice32 (approved):
LGTM!

pxrl:
Part of this PR will be moved to sdk-v2, so I have re-proposed that part here: https://github.com/across-protocol/sdk-v2/pull/417.

james-a-morris (approved):
LGTM - this is now reflected in the SDK

PR #963, opened 2023-10-02
refactor: Simplify inventory client parsing

pxrl:
This helps to drop an UMA dependency, and is arguably nicer because it doesn't corce the js Object to and from a string type.


src/relayer/RelayerConfig.ts
```diff
@@ -65,11 +65,12 @@ export class RelayerConfig extends CommonConfig {
     this.slowDepositors = SLOW_DEPOSITORS
       ? JSON.parse(SLOW_DEPOSITORS).map((depositor) => ethers.utils.getAddress(depositor))
       : [];
-    this.inventoryConfig = RELAYER_INVENTORY_CONFIG ? JSON.parse(RELAYER_INVENTORY_CONFIG) : {};
+    this.inventoryConfig = JSON.parse(
+      RELAYER_INVENTORY_CONFIG?.replace(/0x[a-fA-F0-9]{40}/g, ethers.utils.getAddress) ?? "{}"
```
nicholaspai, line 69: Do you think we should make this a helper function? I think there are several other places where we expect an address string set in the .env that might not be checksummed

pxrl: We can make it a utility function; I only did it here because this was the only place I could see that we actually normalised addressed on input.

I'll update accordingly :+1:

james-a-morris, line 69: It may be easier to use a reviver in this case. `JSON.parse`'s second parameter is an optional function that looks at keys/values. We could probably create a reviver function like:
```
/**
 * Converts a potential string/array of strings that match an address into a checksummed address
 **/
export function checksumAddress(value: unknown): unknown {
    const isArray = Array.isArray(value);
    const result: unknown[] = (isArray ? value : [value]).map((v) => utils.isAddress(v) ? utils.getAddress(v) : v;
    return isArray ? result : result[0];
}
```
```
JSON.parse(
    RELAYER_INVENTORY_CONFIG ?? {},
    (, v) => checksumAddress(v)
)
```

james-a-morris (commented):
Left a comment to possibly add discussion about the readability of the parse function

PR #964, opened 2023-10-02
feat: Import constants-v2

nicholaspai:
IGets updated BAL address

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,12 +5,12 @@
   "repository": "git@github.com:across-protocol/relayer-v2.git",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
-  "private": true,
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -5,12 +5,12 @@
   "repository": "git@github.com:across-protocol/relayer-v2.git",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
-  "private": true,
```
james-a-morris, line 8: Did we want to keep this as to not accidentally publish to npmjs?

james-a-morris (approved):

package.json
nicholaspai: oh is that why we did that?

PR #965, opened 2023-10-03
chore: remove refactored tests

james-a-morris:
Will require https://github.com/across-protocol/sdk-v2/pull/425 to be merged into the SDK to pass.


src/caching/RedisCache.ts
```diff
@@ -71,4 +71,10 @@ export class RedisCache implements interfaces.CachingMechanismInterface {
     // Return key to indicate that the value was set successfully.
     return key;
   }
+
+  static resolveFromRedisClient(redisClient: RedisClient, logger?: winston.Logger): RedisCache {
```
mrice32, line 75: What are these behavior changes? Are these related to removing tests?

mrice32 (commented):
Looks great! Only one question

src/caching/RedisCache.ts
james-a-morris: This behavior change is related to moving the tests. The goal of `RedisCache` is that it is a wrapped version of our `RedisClient` that complies with the `CachingMechanismInterface`. The `CachingMechanismInterface` is an interface that allows the SDK to leverage caching without knowing the specifics of what cache is being used.

The `RedisClient` isn't specifically a `CachingMechanismInterface` so `RedisCache` is used to wrap it with an additional reviver.

james-a-morris: Effectively, I wanted to create a way for the SDK to be unopinionated on which cache mechanism we use for two reasons:

- Caching mechanisms come and go and as the SDK gets more mature, it would make changing more difficult.
- With this system, we don't actually need to store any cache-specific clients in our dependencies. The onus of storage is placed on the software that calls the SDK. As such, we can keep our SDK lighter in terms of required dependencies.

mrice32: Gotcha, makes sense. I'll take a pass over the caching logic and match it up with the SDK.

I suspect that the tests could be deleted without updating the caching interface, but it looks like we're also upgrading the version, too. In the future, I might suggest trying to break these two things apart into 1) a PR that deletes the tests and 2) a PR that updates the version and changes the cache interface to make it easier to review. Just a minor nit, though!

mrice32 (approved):
LGTM!

nicholaspai (approved):
LGTM have you tested the relayer, dataworker, finalizer in sim mode and made sure there are no regressions?
- relayer looks good to fill deopsits
- relayer can manage inventory correctly
- proposer proposes a bundle that looks normal
- disputer doesnt want to dispute

Also, try running `validateRunningBalances` and `validateRootBundle`

james-a-morris:
@pxrl @nicholaspai @mrice32 this PR passes our validate root bundle at multiple steps but fails the running balance test. This seems to also be failing on master so it may not be related to this PR. In either way, it is open for any additional reviews.

src/utils/BlockUtils.ts
pxrl, line 1: FWIW I'm probably partial to submitting this separately. In this PR we're already bumping SDK and refactoring a bunch of tests. If we see any regressions it might be difficult to identify where they are, and since this particular change isn't integral to that and is easy to submit afterwards, wdyt about handling it afterwards?

james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/965/commits/cc236b7cca24701c3be189c0dfc71cd955f458ee

nicholaspai (approved):
LGTM

PR #970, opened 2023-10-06
feat: Support reading mnemonic or private key from file

pxrl:
Rather than read the mnemonic or private key directly from .env, specify the location of a separate file via the SECRET env var. Then, open that file and read the mnemonic or key from there. Global fs scope is applied to the location, so it can be located in a totally different path, and can even have more restrictive permissions. This significantly reduces the chance of leaking critical secrets - i.e. when sharing screen.

A simple regex is applied to determine whether the format matches a private key, and if not, it's assumed to be a mnemonic. This should make it easier to manage for operators.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ typechain*
 .DS_Store
 dump.rdb*
 .idea
+.mnemonic
+.keyfile
+.secret
 
 # Hardhat files
 cache
--- a/src/utils/CLIUtils.ts
+++ b/src/utils/CLIUtils.ts
@@ -12,7 +12,7 @@ export function retrieveSignerFromCLIArgs(): Promise<Wallet> {
   // Resolve the wallet type & verify that it is valid.
   const keyType = (args.wallet as string) ?? "mnemonic";
   if (!isValidKeyType(keyType)) {
-    throw new Error(`Unsupported key type (${keyType}); expected "mnemonic", "privateKey" or "gckms"`);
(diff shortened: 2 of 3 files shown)


nicholaspai (approved):

nicholaspai:
Should we add an example file like .env.example or nah

pxrl:
> Should we add an example file like .env.example or nah

Ah good call - definitely!

james-a-morris (commented):
How often does this happen? IIUC you'd need to give the program read access to the file system in addition to just execute access on the OS. 

Is there any vulnerabilities introduced here if the user specified a file that exists but isn't a valid secret? 

I.e. ~/.bash_history

mrice32 (approved):

pxrl:
> How often does this happen? IIUC you'd need to give the program read access to the file system in addition to just execute access on the OS.

The bot is already expected to have filesystem access, because it reads its config from `.env`. It's certainly possible for an operator to lock their bot down so much that it doesn't have fs access, but in that case I'd expect them to use `MNEMONIC` or `PRIVATE_KEY` directly.

> Is there any vulnerabilities introduced here if the user specified a file that exists but isn't a valid secret?
> 
> I.e. ~/.bash_history

None that I can see, though I might be missing something In any case, the operator is expected to configure this correctly, and even if they don't, there is no wallet loaded at the time this is called. So the input string would be passed into to `Wallet.fromMnemonic`, and it would immediately blow up because it didn't receive a valid seed phrase. That error is caught and displayed, and the program immediately terminates. So I don't really see any marginal insecurity added a result of this, though there are certainly still things that we can improve on from here :+1:

PR #972, opened 2023-10-08
improve(inventory-manager): Check ERC20 balance before submitting cross chain bridge

nicholaspai:
The  `TokenClient` fetches ERC20 balances for the relayer prior to carrying out Relayer duties. This means that there could be a multiple minute delay between balances between fetched and the inventory 
manager executing, exposing the relayer to the risk of sending a duplicate rebalance.

This PR adds a simple check for the ERC20 balance as close to sending the inventory rebalance as possible. The downside is a slower inventory manager run, which is mitigated by a side effect of allowing the bot-runner to refactor the inventory manager out of the relayer. This can be done by setting the new environment variable `SKIP_RELAYS` which will skip the relayer portion of the `relayer/index.ts` run.

Fixes ACX1595

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -318,7 +318,7 @@ export class InventoryClient {
         // balance then this logic ensures that we only fill the first n number of chains where we can.
         if (amount.lt(balance)) {
           // As a precautionary step before proceeding, check that the token balance for the token we're about to send
-          // hasn't changed on L1. It's possible its changed since we updated the inventory due to one or more of the 
+          // hasn't changed on L1. It's possible its changed since we updated the inventory due to one or more of the
           // RPC's returning slowly.
           const expectedBalance = this.tokenClient.getBalance(1, l1Token);
           const tokenContract = new Contract(l1Token, ERC20.abi, this.hubPoolClient.hubPool.signer);
@@ -326,24 +326,25 @@ export class InventoryClient {
           if (!expectedBalance.eq(currentBalance)) {
             this.logger.warn({
               at: "InventoryClient",
-              message: `🚧 Token balance in relayer on Ethereum changed before sending cross chain transfer, skipping rebalance`,
+              message:
+                "🚧 Token balance in relayer on Ethereum changed before sending cross chain transfer, skipping rebalance",
               l1Token,
               l2ChainId: chainId,
               expectedBalance,
-              currentBalance
+              currentBalance,
             });
             continue;
           } else {
             this.logger.debug({
               at: "InventoryClient",
-              message: `Token balance in relayer on Ethereum is as expected, sending cross chain transfer`,
+              message: "Token balance in relayer on Ethereum is as expected, sending cross chain transfer",
               l1Token,
               l2ChainId: chainId,
-              expectedBalance
+              expectedBalance,
             });
             possibleRebalances.push(rebalance);
             // Decrement token balance in client for this chain and increment cross chain counter.
-            this.trackCrossChainTransfer(l1Token, amount, chainId);  
+            this.trackCrossChainTransfer(l1Token, amount, chainId);
           }
         } else {
           // Extract unexecutable rebalances for logging.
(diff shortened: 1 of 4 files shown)


src/clients/InventoryClient.ts
```diff
@@ -316,9 +317,35 @@ export class InventoryClient {
         // the rebalance to this particular chain. Note that if the sum of all rebalances required exceeds the l1
         // balance then this logic ensures that we only fill the first n number of chains where we can.
         if (amount.lt(balance)) {
-          possibleRebalances.push(rebalance);
-          // Decrement token balance in client for this chain and increment cross chain counter.
-          this.trackCrossChainTransfer(l1Token, amount, chainId);
+          // As a precautionary step before proceeding, check that the token balance for the token we're about to send
+          // hasn't changed on L1. It's possible its changed since we updated the inventory due to one or more of the
+          // RPC's returning slowly.
```
pxrl, line 322:
```suggestion
          // RPC's returning slowly, leading to concurrent/overlapping instances of the bot running.
```

src/relayer/RelayerConfig.ts
pxrl, line 1: Is the `skipRelays` stuff relevant to this change?

pxrl, line 1: Just to double check, the `skipRelays` flag is really a secondary change to permit the InventoryClient to run in a standalone instance, right? I otherwise don't think this change should add any noticeable delay to the existing relayer run (as I read the commit message, it seems to indicate that this change slows the bot down).

pxrl: (So basically, I think it'd be sufficient to just have the final `erc20.balanceOf()` check before submitting any rebalance transactions, as a mitigation for the fact that we don't know whether there are concurrent relayer instances running)

nicholaspai: > Just to double check, the skipRelays flag is really a secondary change to permit the InventoryClient to run in a standalone instance, right?

Yes exactly, it should not add any delay to the existing relayer run. The slowdown is moreso related to the `ERC20.balanceOf` check added in the inventory manager

src/clients/InventoryClient.ts
```diff
@@ -316,9 +317,35 @@ export class InventoryClient {
...
+          // hasn't changed on L1. It's possible its changed since we updated the inventory due to one or more of the
+          // RPC's returning slowly, leading to concurrent/overlapping instances of the bot running.
+          const expectedBalance = this.tokenClient.getBalance(1, l1Token);
+          const tokenContract = new Contract(l1Token, ERC20.abi, this.hubPoolClient.hubPool.signer);
+          const currentBalance = await tokenContract.balanceOf(this.relayer);
+          if (!expectedBalance.eq(currentBalance)) {
+            this.logger.warn({
+              at: "InventoryClient",
+              message:
+                "🚧 Token balance in relayer on Ethereum changed before sending cross chain transfer, skipping rebalance",
```
pxrl, line 330: Just trying to keep this from a weird long line:

```suggestion
                "🚧 Token balance on Ethereum changed before sending transaction, skipping rebalance",
```

src/relayer/RelayerConfig.ts
```diff
@@ -131,7 +133,8 @@ export class RelayerConfig extends CommonConfig {
     }
     this.debugProfitability = DEBUG_PROFITABILITY === "true";
     this.relayerGasMultiplier = toBNWei(RELAYER_GAS_MULTIPLIER || Constants.DEFAULT_RELAYER_GAS_MULTIPLIER);
-    this.sendingRelaysEnabled = SEND_RELAYS === "true";
```
pxrl, line 134: FWIW I think it's worth calling this out in the commit message and I probably think it's better to just do it in a separate commit, since it is a breaking change for existing relayer configs.

nicholaspai: fair point reverted here https://github.com/across-protocol/relayer-v2/pull/972/commits/9818120e3237c05bf70bf7da6afa3f8cc2cb1801

PR #975, opened 2023-10-09
refactor: Remove InventoryClient hardcoded WETH address

pxrl:


nicholaspai (approved):

src/clients/InventoryClient.ts
```diff
@@ -425,7 +425,7 @@ export class InventoryClient {
       if (!this.isInventoryManagementEnabled()) {
         return;
       }
-      const l1Weth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+      const l1Weth = CONTRACT_ADDRESSES[this.hubPoolClient.chainId].weth.address;
```
nicholaspai, line 428: Why not grab this from `TOKEN_SYMBOLS_MAP`?

pxrl: CONTRACT_ADDRESSES was already in use here, but I've updated it to use TOKEN_SYMBOLS_MAP now.

PR #980, opened 2023-10-09
fix: Handle sped-up deposits w/ nullified message

pxrl:
The existing logic didn't work properly when a deposit with message was subsequently sped-up with no message.


pxrl:
Hm, this one fails a relayer test. Will check that out...

src/relayer/Relayer.ts
```diff
@@ -231,18 +231,15 @@ export class Relayer {
...
-        this.logger.warn({
-          at: "Relayer",
-          message: "Skipping fill for sped-up deposit with message",
-          deposit,
-        });
-        continue;
-      } else if (deposit.message !== "0x") {
+      if (
+        (isDepositSpedUp(deposit) && deposit.updatedMessage !== "0x") ||
+        (!isDepositSpedUp(deposit) && deposit.message !== "0x")
```
james-a-morris, line 236:
```suggestion
        deposit[isDepositSpedUp(deposit) ? "updatedMessage" : "message") !== "0x"
```

test/Relayer.BasicFill.ts
```diff
@@ -301,7 +301,7 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
       message: "0x1212",
       recipient: randomAddress(),
     };
-    const unusedSpeedUpSignature = await modifyRelayHelper(
+    const { signature: unusedSpeedUpSignature } = await modifyRelayHelper(
```
james-a-morris, line 304: Is this strictly to make the following code not have to reference `.signature`?

james-a-morris (commented):
Left one nit on the code & one question

src/relayer/Relayer.ts
nicholaspai, line 236: I'm a bit confused how this fixes the problem as described in the PR message:

> when a deposit with message was subsequently sped-up with no message.

It looks like this conditional you're adding is catching when the deposit is **not** sped up and that `message = 0x`. But your commit message suggests that the condition should be `isDepositSpedUp(deposit)...`

test/Relayer.BasicFill.ts
pxrl: Yeah - it's simply to avoid the subsequent dereferencing - there's no functional change.

src/relayer/Relayer.ts
pxrl: The existing flow for a deposit with an updated/nullified message is:
- `if (isDepositSpedUp(deposit) && deposit.updatedMessage !== "0x")` -> false (`deposit.updatedMessage` === "0x")
- `else if (deposit.message !== "0x")` -> true; the original deposit had a non-null message, but since we have a speed-up without a message, we shouldn't care about the original message anymore.

I think the subsequent suggestion from @james-a-morris makes the new flow much more clear - by doing it in a single statement, it's clear that we're only intending to evaluate a single message field.

src/relayer/Relayer.ts
```diff
@@ -231,18 +231,12 @@ export class Relayer {
...
       // message is != 0x and it ends up costing a lot of gas to execute, resulting in a big loss to the relayer.
-      if (isDepositSpedUp(deposit) && deposit.updatedMessage !== "0x") {
-        this.logger.warn({
-          at: "Relayer",
-          message: "Skipping fill for sped-up deposit with message",
-          deposit,
-        });
-        continue;
-      } else if (deposit.message !== "0x") {
+      if (deposit[isDepositSpedUp(deposit) ? "updatedMessage" : "message"] !== "0x") {
```
nicholaspai, line 234: This is much clearer +1

test/Relayer.BasicFill.ts
```diff
@@ -316,27 +316,27 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
...
       deposit1.depositId,
       unusedSpeedUp.recipient,
       unusedSpeedUp.message,
-      unusedSpeedUpSignature.signature
+      unusedSpeedUpSignature
     );
     await updateAllClients();
     await relayerInstance.checkForUnfilledDepositsAndFill();
-    expect(lastSpyLogIncludes(spy, "Skipping fill for sped-up deposit with message")).to.be.true;
+    expect(lastSpyLogIncludes(spy, "Skipping fill for deposit with message")).to.be.true;
```
nicholaspai, line 339: Do we need to add another unit test to test for a deposit who's `message != 0x` and is ignored, but then it gets sped up and its `updatedMessage == 0x` so it doesn't get ignored ?

james-a-morris (approved):
LGTM

mrice32 (approved):

pxrl:
Applied `do not merge` - this is ready, but since we're rolled back from master, it should wait until those issues are resolved.

test/Relayer.BasicFill.ts
pxrl: Just circling back to this - I had to bump contracts-v2 in the mean time. There was an issue with the existing tests where we unconditionally defaulted the message on deposits to "0x", so we weren't ever able to test a scenario where a non-zero message was nullified by a speed-up. I've now added this specific test in this commit: https://github.com/across-protocol/relayer-v2/pull/980/commits/4102236ea889db8ff190a606a17f56248d1a7e89

PR #981, opened 2023-10-09
fix: Handle zero-fills with speed-ups

pxrl:
If a deposit with an updated message happens to be slow-filled, the slow-fill logic will unconditionally use the fillRelay() method. Instead, it should use fillRelayWIthUpdatedDeposit().

To make this simple, just augment the existing fillRelay() method so that it understands how to handle slow fills.

Fixes ACX-1601


pxrl:
nb this might break some tests, so I've submitted it as a draft for now. Will circle back to it on Tuesday.

src/relayer/Relayer.ts
```diff
@@ -311,12 +312,12 @@ export class Relayer {
...
     // original hash with the old fee hasn't been filled.
     const [method, argBuilder, messageModifier] = isDepositSpedUp(deposit)
-      ? ["fillRelayWithUpdatedDeposit", buildFillRelayWithUpdatedFeeProps, "with modified fee "]
+      ? ["fillRelayWithUpdatedDeposit", buildFillRelayWithUpdatedFeeProps, "with modified parameters "]
       : ["fillRelay", buildFillRelayProps, ""];
 
     const message = fillAmount.eq(deposit.amount)
       ? `Filled deposit ${messageModifier}🚀`
-      : `Partially filled deposit ${messageModifier}📫"`;
+      : `${fillAmount.eq(zeroFillAmount) ? "Zero" : "Partially"} filled deposit ${messageModifier}📫"`;
```
pxrl, line 320: One impact here: the same emojji is used for zero fill and partial fill. Can change that if we want to retain the zero fill emoji.

src/relayer/Relayer.ts
```diff
@@ -311,12 +312,12 @@ export class Relayer {
     // accidentally double fill due to the deposit hash being different - SpokePool contract will check that the
     // original hash with the old fee hasn't been filled.
     const [method, argBuilder, messageModifier] = isDepositSpedUp(deposit)
-      ? ["fillRelayWithUpdatedDeposit", buildFillRelayWithUpdatedFeeProps, "with modified fee "]
+      ? ["fillRelayWithUpdatedDeposit", buildFillRelayWithUpdatedFeeProps, "with modified parameters "]
```
nicholaspai, line 315: changing this message is probably breaking tests FYI

src/relayer/Relayer.ts
nicholaspai: Yes please keep the emoji distinction, i genuinely find it helpful when scrolling through the messages

src/relayer/Relayer.ts
```diff
@@ -328,7 +329,7 @@ export class Relayer {
     });
 
     // TODO: Revisit in the future when we implement partial fills.
-    this.fullyFilledDeposits[fillKey] = true;
+    this.fullyFilledDeposits[fillKey] = fillAmount.eq(deposit.amount);
 
     // Decrement tokens in token client used in the fill. This ensures that we dont try and fill more than we have.
     this.clients.tokenClient.decrementLocalBalance(deposit.destinationChainId, deposit.destinationToken, fillAmount);
```
nicholaspai, line 335: Any weird regressions if we decrement by such a small amount?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: `this.fullyFilledDeposits[fillKey]` is a boolean, so we're only recording whether we've completed the fill.

There is an issue here though, since we do occasionally make partial fills that complete the fill (i.e. we have made a zero fill, then get a refund and fill the balance).

pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/pull/981/commits/136c76894e612cb704e0e191fc46a90bd6435d13. This basically maintains the existing logic.

src/relayer/Relayer.ts
pxrl: That's a safe bet, but I checked and we surprisingly don't check for this log message in tests.

src/relayer/Relayer.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/981/commits/1cf29fc31d858062687cdabbda3f939fc9c49611

pxrl:
Applied `do not merge` - this is ready, but since we're rolled back from master, it should wait until those issues are resolved.

src/relayer/Relayer.ts
```diff
@@ -255,7 +256,7 @@ export class Relayer {
             message: "Initiating slow fill for grey listed depositor",
             depositor: deposit.depositor,
           });
-          this.zeroFillDeposit(deposit);
+          this.fillRelay(deposit, zeroFillAmount, deposit.destinationChainId);
```
james-a-morris, line 259: We should add a comment here on why we are changing this

james-a-morris (commented):
Left a few comments

src/relayer/Relayer.ts
pxrl: I'm not sure what I can add here that's not already evident by the surrounding context - i.e. the log message ("initiating slow fill for grey listed depositor") and the existing comments that we're initiating a slow fill.

pxrl: On second thought, it probably makes sense to retain the zeroFillDeposit() wrapper, since we rely on the destinationChainId being set correctly. Restored here: https://github.com/across-protocol/relayer-v2/pull/981/commits/4632731fda2d87e485b1336fa986dd2af63765f5.

PR #982, opened 2023-10-10
feat(relayer): Propagate estimated gas cost for fills

pxrl:
When estimating relay profitability, pass the computed gasLimit back to the caller. This is a prerequisite for dynamically computing the cost of filling deposits with messages, and should incidentally save doing the fill estimate later on.


src/relayer/Relayer.ts
```diff
@@ -267,9 +267,14 @@ export class Relayer {
         // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
         deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit);
 
-        const repaymentChainId = await this.resolveRepaymentChain(version, deposit, unfilledAmount, l1Token);
+        const [repaymentChainId, gasLimit] = await this.resolveRepaymentChain(
```
pxrl, line 270: todo: Need to identify where to apply a gas padding strategy - i.e. such that the estimated gas cost is padded to avoid sporadic OoG.

src/relayer/Relayer.ts
```diff
@@ -547,14 +553,16 @@ export class Relayer {
...
-    const refundFee = this.computeRefundFee(version, deposit);
+    const preferredChainId =
+      fillAmount.eq(deposit.amount) && !fillsInQueueForSameDeposit
+        ? await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)
+        : destinationChainId;
 
-    const profitable = profitClient.isFillProfitable(deposit, fillAmount, refundFee, hubPoolToken);
-    return profitable ? preferredChainId : undefined;
+    const refundFee = this.computeRefundFee(version, deposit);
+    const [profitable, gasLimit] = profitClient.isFillProfitable(deposit, fillAmount, refundFee, hubPoolToken);
```
pxrl, line 564: Problem: requiring profitability here will effectively prevent zero fills, since a 1 wei fill can never be profitable. Will need to figure out how to handle that gracefully.

pxrl: ^^ The above is actually a non-issue, since we only call `resolveRepaymentChain()` for non-zero fills.

src/relayer/Relayer.ts
nicholaspai: Do you think it should be in the relayer or the sdk? My instinct is the relayer should apply the gas padding

src/relayer/Relayer.ts
```diff
@@ -527,14 +533,15 @@ export class Relayer {
...
         deposit: { originChain, depositId, destinationChain, depositHash },
       });
-      return destinationChainId;
     }
 
-    const preferredChainId = await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address);
-    const refundFee = this.computeRefundFee(version, deposit);
+    const preferredChainId = fillAmount.eq(deposit.amount)
+      ? await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)
+      : destinationChainId;
```
nicholaspai, line 540: why default to `destinationChainId` for partial fills?

src/relayer/Relayer.ts
```diff
@@ -527,14 +533,15 @@ export class Relayer {
...
 
-    const profitable = profitClient.isFillProfitable(deposit, fillAmount, refundFee, hubPoolToken);
-    return profitable ? preferredChainId : undefined;
+    const refundFee = this.computeRefundFee(version, deposit);
+    const [profitable, gasLimit] = profitClient.isFillProfitable(deposit, fillAmount, refundFee, hubPoolToken);
```
nicholaspai, line 543: Here is where I'd guess you want to apply the gas padding, perhaps as an argument to `isFillProfitable`

nicholaspai (commented):
Some thoughts on first pass

src/relayer/Relayer.ts
pxrl: This is required by the SpokePool: https://github.com/across-protocol/contracts-v2/blob/406a68566b54d688feac8a503dd85546faec9365/contracts/SpokePool.sol#L1223

test/Relayer.TokenShortfall.ts
```diff
@@ -109,6 +109,7 @@ describe("Relayer: Token balance shortfall", async function () {
     );
     const spokePoolClients = { [originChainId]: spokePoolClient_1, [destinationChainId]: spokePoolClient_2 };
     tokenClient = new TokenClient(spyLogger, relayer.address, spokePoolClients, hubPoolClient);
+
```
james-a-morris, line 112:
```suggestion
```

src/clients/ProfitClient.ts
```diff
@@ -17,16 +17,19 @@ export const USDC = TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET];
...
 export const WETH = TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET];
 
-// note: All FillProfit BigNumbers are scaled to 18 decimals unless specified otherwise.
-export type FillProfit = {
+// @note All FillCost and FillProfit BigNumbers are scaled to 18 decimals unless specified otherwise.
+type FillCost = {
+  nativeGasCost: BigNumber; // Cost of completing the fill in the native gas token.
+  gasPriceUsd: BigNumber; // Price paid per unit of gas in USD.
+  gasCostUsd: BigNumber; // Estimated cost of completing the fill in USD.
+};
```
james-a-morris, line 25: To clarify, does this specifically remove the need to define these types in several locations?

src/clients/ProfitClient.ts
```diff
@@ -326,9 +313,26 @@ export class ProfitClient {
...
 
-  isFillProfitable(deposit: Deposit, fillAmount: BigNumber, refundFee: BigNumber, l1Token: L1Token): boolean {
-    const { profitable } = this.getFillProfitability(deposit, fillAmount, refundFee, l1Token);
-    return profitable ?? this.isTestnet;
+  isFillProfitable(
+    deposit: Deposit,
+    fillAmount: BigNumber,
+    refundFee: BigNumber,
+    l1Token: L1Token
+  ): [boolean, BigNumber] {
```
james-a-morris, line 321: Instead of a tuple, can we return a labeled object?

src/relayer/Relayer.ts
```diff
@@ -295,7 +300,7 @@ export class Relayer {
     }
   }
 
-  fillRelay(deposit: Deposit, fillAmount: BigNumber, repaymentChainId: number): void {
+  fillRelay(deposit: Deposit, fillAmount: BigNumber, repaymentChainId: number, gasLimit?: BigNumber): void {
```
james-a-morris, line 303: Should we set `gasLimit = bnMaxUint256` so that it's always defined?

src/relayer/Relayer.ts
```diff
@@ -515,7 +521,7 @@ export class Relayer {
     deposit: DepositWithBlock,
     fillAmount: BigNumber,
     hubPoolToken: L1Token
-  ): Promise<number | undefined> {
+  ): Promise<[number | undefined, BigNumber]> {
```
james-a-morris, line 524: Thoughts on avoiding typed tuples & using structs as an alternative?

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -306,7 +306,7 @@ describe("ProfitClient: Consider relay profit", () => {
               netRelayerFeeUsd: expected.netRelayerFeeUsd,
             });
 
-            const profitable = profitClient.isFillProfitable(deposit, nativeFillAmount, zeroRefundFee, l1Token);
+            const [profitable] = profitClient.isFillProfitable(deposit, nativeFillAmount, zeroRefundFee, l1Token);
```
james-a-morris, line 309: Similar pushback. 

The idea of restructuring an array may get messy whereas destructuring a struct is more explicit.

src/relayer/Relayer.ts
pxrl: It's actually preferable for it to be specifically set, or undefined. If it's undefined then ethers will simulate the transaction and set an approximately sensible gasLimit on the transaction. If we default to uin256max then we're telling ethers not to bother, and it opens the door for a huge gas spend if the transaction consumes more gas than expected.

pxrl: I have that in a follow-up change where we actually dynamically evaluate the cost of a fill. This PR is mostly just plumbing the gas cost / gasLimit all the way back to the caller, without making any behavioural changes to the code.

src/clients/ProfitClient.ts
pxrl: Yep. I've applied the Pick<...> suggestion so that negates the need for this now :+1:

src/relayer/Relayer.ts
pxrl: It's valid for it to be undefined because in this case ethers will set a sensible gas limit based on its own simulation. When making the fill here we don't actually know whether it will be bundled or not, because the multicaller client will opportunistically bundle it, and only revert to the individual  think it's OK to leave it as-is, with the option of tightening it in the future.

src/relayer/Relayer.ts
pxrl: Padding is in the pipeline for the follow-up change that integrates with the SDK :+1:

src/relayer/Relayer.ts
nicholaspai: oh right

nicholaspai (approved):
I think this refactor makes sense.

src/relayer/Relayer.ts
james-a-morris: Ah fair point

james-a-morris (approved):
LGTM

PR #983, opened 2023-10-10
improve(relayer): Clear out token data explicitly and log token balances

nicholaspai:
Based on an issue we're seeing where L1 to L2 rebalances are still getting duplicated, even after wrapping ETH, perhaps suggesting that the `tokenClient` is using stale data

the logic written [here](https://github.com/across-protocol/relayer-v2/blob/master/src/relayer/RelayerClientHelper.ts#L159) should be reading updated balances post-ETH-unwrap but it may not.

This PR adds more logs we can use to better debug

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -634,6 +634,11 @@ export class InventoryClient {
   }
 
   async wrapL2EthIfAboveThreshold(): Promise<void> {
+    // If inventoryConfig is defined, there will be a default wrapEtherTarget and wrapEtherThreshold
+    // set by RelayerConfig.ts
+    if (!this?.inventoryConfig) {
+      return;
+    }
     this.log("Checking ETH->WETH Wrap status");
     await this.adapterManager.wrapEthIfAboveThreshold(this.inventoryConfig, this.simMode);
   }
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -634,9 +634,6 @@ export class InventoryClient {
   }
 
   async wrapL2EthIfAboveThreshold(): Promise<void> {
-    if (!this.isInventoryManagementEnabled()) {
-      return;
-    }
     this.log("Checking ETH->WETH Wrap status");
     await this.adapterManager.wrapEthIfAboveThreshold(this.inventoryConfig, this.simMode);
   }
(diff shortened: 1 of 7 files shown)


src/clients/InventoryClient.ts
```diff
@@ -634,9 +634,6 @@ export class InventoryClient {
   }
 
   async wrapL2EthIfAboveThreshold(): Promise<void> {
-    if (!this.isInventoryManagementEnabled()) {
-      return;
-    }
```
pxrl, line 639: What happens in this instance if there's no inventory config defined?

src/relayer/RelayerConfig.ts
```diff
@@ -108,28 +108,30 @@ export class RelayerConfig extends CommonConfig {
...
-          this.inventoryConfig.tokenConfig[l1Token][chainId].targetPct = toBNWei(targetPct).div(100);
-          this.inventoryConfig.tokenConfig[l1Token][chainId].thresholdPct = toBNWei(thresholdPct).div(100);
-          if (unwrapWethThreshold !== undefined) {
-            this.inventoryConfig.tokenConfig[l1Token][chainId].unwrapWethThreshold = toBNWei(unwrapWethThreshold);
-          }
-          this.inventoryConfig.tokenConfig[l1Token][chainId].unwrapWethTarget = unwrapWethTarget
-            ? toBNWei(unwrapWethTarget)
-            : toBNWei(2);
+      if (this.inventoryConfig.tokenConfig !== undefined) {
+        Object.keys(this.inventoryConfig.tokenConfig).forEach((l1Token) => {
```
pxrl, line 112: This high level of nesting is pretty hard to read; you could maybe get around it with the following:
```suggestion
        Object.keys(this.inventoryConfig?.tokenConfig ?? {}).forEach((l1Token) => {
```

pxrl (approved):

src/clients/InventoryClient.ts
nicholaspai: hm yeah thanks for pressing on this, think this is what I wanted to do: 3d0678fc2b519a16363414c72c5e5367f7ffcd2e

src/relayer/RelayerConfig.ts
nicholaspai: done

PR #987, opened 2023-10-10
fix(validateRunningBalances): Change events we track to perform accounting

nicholaspai:
Various  fixes to script:
- Tracks correct event on OPStack L2's to account for missing PoolRebalance deposits to the SpokePool. Without this, the excess balance looked too high
- Make excess check less sensitive and ignore dust

### DIFF
--- a/src/scripts/validateRunningBalances.ts
+++ b/src/scripts/validateRunningBalances.ts
@@ -25,6 +25,7 @@ import {
   config,
   Logger,
   toBN,
+  Event,
   fromWei,
   Contract,
   ERC20,
@@ -40,7 +41,7 @@ import { createDataworker } from "../dataworker";
 import { getWidestPossibleExpectedBlockRange } from "../dataworker/PoolRebalanceUtils";
 import { getBlockForChain, getEndBlockBuffers } from "../dataworker/DataworkerUtils";
 import { ProposedRootBundle, SlowFillLeaf, SpokePoolClientsByChain } from "../interfaces";
-import { constructSpokePoolClientsWithStartBlocks, updateSpokePoolClients } from "../common";
+import { CONTRACT_ADDRESSES, constructSpokePoolClientsWithStartBlocks, updateSpokePoolClients } from "../common";
 import { createConsoleTransport } from "@uma/financial-templates-lib";
 import { retrieveSignerFromCLIArgs } from "../utils/CLIUtils";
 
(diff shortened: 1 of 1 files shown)


src/scripts/validateRunningBalances.ts
```diff
@@ -158,11 +159,11 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
...
           // Make sure that previous root bundle's netSendAmount has been deposited into the spoke pool. We only
-          // perform this check for chains 10, 137, 288, and 42161 because transfers from the hub pool to spoke
+          // perform this check for these L2 chains because transfers from the hub pool to spoke
           // pools on those chains can take a variable amount of time, unlike transfers to the spoke pool on
-          // mainnet. Additionally, deposits to those chains emit transfer events where the from address
-          // is the zero address, making it easy to track.
-          if ([10, 137, 288, 42161].includes(leaf.chainId)) {
+          // mainnet. Additionally, deposits to those chains emit Transfer events where the to address
+          // is the SpokePool address, making it easy to track.
+          if (leaf.chainId !== 1) {
```
james-a-morris, line 166: We should probably use `constants.HUBPOOL_CHAIN_ID` from `import { constants } from "across-protocol/sdk-v2"`

src/scripts/validateRunningBalances.ts
```diff
@@ -182,19 +183,54 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
...
+                      {
+                        fromBlock: previousBundleEndBlockForChain.toNumber(),
+                        toBlock: bundleEndBlockForChain.toNumber(),
+                        maxBlockLookBack: config.maxBlockLookBack[leaf.chainId],
+                      }
+                    )
+                  ).filter(
+                    (e) =>
+                      e.args._amount.eq(previousNetSendAmount) &&
+                      e.args._to === spokePoolClients[leaf.chainId].spokePool.address
```
james-a-morris, line 217: nit: we could probably destructure this into a variable above to prevent this lookup happening > 1 times

src/scripts/validateRunningBalances.ts
```diff
@@ -335,9 +371,10 @@ export async function runScript(_logger: winston.Logger, baseSigner: Wallet): Pr
   });
   const unexpectedExcess = Object.entries(excesses).some(([, tokenExcesses]) => {
     return Object.entries(tokenExcesses).some(([, excesses]) => {
-      // We only care that the latest excesses are 0, because sometimes excesses can appear in historical bundles
+      // We only care about the latest excess, because sometimes excesses can appear in historical bundles
       // due to ordering of executing leaves. As long as the excess resets back to 0 eventually it is fine.
-      return Number(excesses[0]).toFixed(6) !== "0.000000";
+      const excess = Number(excesses[0]);
+      return excess > 0.05 || excess < -0.05;
```
james-a-morris, line 377: This may be problematic for the edge case that `excesses[0]` is too large. It would probably make sense to compare something like:

```
return excess.abs().gt(fromWei(0.05))
```

james-a-morris (approved):
Nice work. I only have nits for this one which may help prevent some error paths if our numbers are anomylously large.

james-a-morris:
Has this been broken for some time or was this changed in more recent changes?

src/scripts/validateRunningBalances.ts
nicholaspai: `excess` is not a BN, its already been converted from one above

james-a-morris: Right, but why are we wrapping the `excesses[0]` string in `Number`? We could deal with it as a `BN` and prevent any overflow issues

PR #991, opened 2023-10-11
fix: :ambulance: resolve double stringify

james-a-morris:
The `RedisCache` introduced a double `JSON.stringify` and a `JSON.parse` issue. These ran fine in an isolated environment because `JSON.parse(JSON.parse(JSON.stringify(JSON.stringify(X))))` will compute fine, but key/values clashed with our existing cache. 

This new hot fix removes the double `JSON.stringify` issue that was present with the merge of the latest PR.

The cache should be cleared for good measure.

This was tested using a relayer iteration and a dataworker interation with a lookback of 16 bundles.

### DIFF
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -128,5 +128,17 @@ export async function queryHistoricalDepositForFill(
   spokePoolClient: SpokePoolClient,
   fill: Fill
 ): Promise<DepositWithBlock | undefined> {
-  return utils.queryHistoricalDepositForFill(spokePoolClient, fill, await getRedisCache(spokePoolClient.logger));
+  try {
+    return await utils.queryHistoricalDepositForFill(
+      spokePoolClient,
+      fill,
+      await getRedisCache(spokePoolClient.logger)
+    );
+  } catch (e) {
+    spokePoolClient.logger.error({
+      at: "DepositUtils#queryHistoricalDepositForFill",
+      message: e,
+    });
+    throw e;
+  }
 }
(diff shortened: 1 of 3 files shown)


src/utils/DepositUtils.ts
```diff
@@ -128,5 +128,17 @@ export async function queryHistoricalDepositForFill(
   spokePoolClient: SpokePoolClient,
   fill: Fill
 ): Promise<DepositWithBlock | undefined> {
-  return utils.queryHistoricalDepositForFill(spokePoolClient, fill, await getRedisCache(spokePoolClient.logger));
```
nicholaspai, line 131: Why are you making this change? Are you sure this isn't going to cause an error to be suppressed higher up in the stack?

nicholaspai (commented):
One question

src/utils/DepositUtils.ts
james-a-morris: This change is just meant to log and throw the error further in the stack. We throw e at the end. 

I wanted to keep this here because we lose the stack trace when we call into the SDK with TSNode so it's a way we can log which call resulted in an error

nicholaspai (approved):

src/utils/DepositUtils.ts
mrice32: Sorry, maybe I'm misunderstanding, but why will emitting the log with e show error text that isn't seen if we just allow the error to bubble up? I don't quite see the difference

mrice32 (commented):
Looks great! Just one question about the error part

src/utils/DepositUtils.ts
james-a-morris: I've removed it as it's something we could glean from the error path. It was a holdover from the debugging process that I found an effective way to get a more granular view of the error path.

PR #992, opened 2023-10-12
chore(across-v2): Bump @uma/financial-templates-lib

pxrl:
After discussing w/ Reinis, UMA had also noticed that the Slack
transport is prone to bombing out on long, complex JSON objects, and 
had independently made some improvements over the past couple of months.

Bumping to financial-templates-lib 2.34.0 makes the transport more
robust in that it doesn't die on the first badly-formatted log message.
It also supports notifying of transport errors via a dedicated logging
transport, which might be useful. 

I think it makes sense to jump to this new version immediately. There
are some residual issues - i.e. per our current config, this would
simply drop the log message that we'd like to log. We can try to work
out a common solution w/ the UMA team for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@openzeppelin/hardhat-upgrades": "^1.28.0",
     "@uma/common": "2.33.0",
     "@uma/contracts-node": "^0.4.16",
-    "@uma/financial-templates-lib": "^2.32.11",
+    "@uma/financial-templates-lib": "^2.34.0",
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

pxrl:
Applied do not merge - this is ready, but since we're rolled back from master, it should wait until those issues are resolved.

james-a-morris (approved):

pxrl:
This has tested OK; the issues of dropped log messages seems to be resolved completely.

PR #995, opened 2023-10-12
improve: Catch Finalizer lookback misconfiguration

pxrl:
I discovered this the hard way.


src/finalizer/index.ts
```diff
@@ -246,6 +246,10 @@ export class FinalizerConfig extends DataworkerConfig {
 
     // `maxFinalizerLookback` is how far we fetch events from, modifying the search config's 'fromBlock'
     this.maxFinalizerLookback = Number(FINALIZER_MAX_TOKENBRIDGE_LOOKBACK ?? FINALIZER_TOKENBRIDGE_LOOKBACK);
+    assert(
+      !isNaN(this.maxFinalizerLookback),
```
nicholaspai, line 250: Should we also make sure its an integer?

nicholaspai (approved):
Should we also make sure its an integer?

src/finalizer/index.ts
pxrl: Good call - implemented: https://github.com/across-protocol/relayer-v2/pull/995/commits/4ca99277369d48f9991437a4de1d0a684bf3840d

PR #996, opened 2023-10-13
refactor: Use BlockFinder from SDK

pxrl:
This change migrates from the UMA BlockFinder to the Across-local variant and enables us to drop @uma/sdk as a direct dependency. We still inherit it from other packages, though.

### DIFF
--- a/src/utils/BlockUtils.ts
+++ b/src/utils/BlockUtils.ts
@@ -1,8 +1,4 @@
-import { utils as sdkUtils } from "@across-protocol/sdk-v2";
-import { getProvider, getRedis, isDefined, setRedisKey, shouldCache } from "./";
-
-type BlockFinder = sdkUtils.BlockFinder;
-const { BlockFinder } = sdkUtils;
+import { BlockFinder, getProvider, getRedis, isDefined, setRedisKey, shouldCache } from "./";
 
 const blockFinders: { [chainId: number]: BlockFinder } = {};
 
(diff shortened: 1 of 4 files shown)


src/utils/BlockUtils.ts
```diff
@@ -1,18 +1,21 @@
-import { BlockFinder } from "@uma/financial-templates-lib";
-import { Block, getProvider, getRedis, isDefined, setRedisKey, shouldCache } from "./";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { getProvider, getRedis, isDefined, setRedisKey, shouldCache } from "./";
 
-const blockFinders: { [chainId: number]: BlockFinder<Block> } = {};
+type BlockFinder = sdkUtils.BlockFinder;
+const { BlockFinder } = sdkUtils;
```
pxrl, line 5: It's annoying to have to do this by ts requires it when BlockFinder is used for a type annotation :\

src/utils/BlockUtils.ts
```diff
@@ -1,18 +1,21 @@
...
  * for this chain ID. Otherwise creates a new block finder and adds it to the cache.
  * @param chainId
  * @returns
  */
-export async function getBlockFinder(chainId: number): Promise<BlockFinder<Block>> {
+export async function getBlockFinder(chainId: number): Promise<BlockFinder> {
   if (!isDefined(blockFinders[chainId])) {
     const providerForChain = await getProvider(chainId);
-    blockFinders[chainId] = new BlockFinder<Block>(providerForChain.getBlock.bind(providerForChain), [], chainId);
+    blockFinders[chainId] = new BlockFinder(providerForChain);
```
james-a-morris, line 18: +1 for never having to do `.bind( )` again lol

james-a-morris (approved):
LGTM - this is a straightforward refactor and I expect that any compilation/test errors would be caught by the linter/test clients

src/utils/BlockUtils.ts
pxrl: Ah, nice! That's much better. Thanks!

PR #1000, opened 2023-10-14
improve: disconnect all clients

james-a-morris:
Our disconnect logic currently only handles the default `redis` URL. (logic below). We ignore all the clients held in the  `redisClients` singleton.

```
export async function disconnectRedisClient(logger?: winston.Logger): Promise<void> {
  const redisClient = await getRedis(logger);
  if (redisClient !== undefined) {
    // todo understand why redisClient isn't GCed automagically.
    logger.debug({ at: "disconnectRedisClient", message: "Disconnecting from redis server." });
    await redisClient.disconnect();
  }
}
```

In addition, we're calling `getRedis` to find the client we want to disconnect. The issue is that if the following logic fails (within `getRedis()`), then we have created a redis connection that isn't stored to the `redisClients` singleton. We can never close this because it isn't saved.

```
      const redisClient = createClient({ url });
      await redisClient.connect();
```

Then, when we eventually call `disconnectRedisClient()` in the catch handlers. If the error was related to redis, then the issue circles back to the problem of calling `getRedis()` within the disconnect logic. This will throw in the catch statement and cause an uncaught exception.

---

In all of our try/catch blocks, we attempt to disconnectRedisClient. A more apt usage would be to use the `finally` block which is guaranteed to run after a try/catch regardless of failure. 

---

Lastly, we don't eject the disconnected redis client from our `redisClients` object.

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -180,8 +180,7 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
         break;
       }
     }
-  } catch (error) {
+  } finally {
     await disconnectRedisClients(logger);
-    throw error;
   }
 }
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -1,4 +1,4 @@
-import { processEndPollingLoop, winston, config, startupLogLevel, Wallet, disconnectRedisClient } from "../utils";
+import { processEndPollingLoop, winston, config, startupLogLevel, Wallet, disconnectRedisClients } from "../utils";
 import { spokePoolClientsToProviders } from "../common";
 import { Dataworker } from "./Dataworker";
 import { DataworkerConfig } from "./DataworkerConfig";
@@ -181,7 +181,7 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
       }
     }
   } catch (error) {
-    await disconnectRedisClient(logger);
+    await disconnectRedisClients(logger);
     throw error;
   }
 }
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -82,8 +82,8 @@ export async function getRedis(logger?: winston.Logger, url = REDIS_URL): Promis
       });
       redisClients[url] = new RedisClient(redisClient, globalNamespace);
     } catch (err) {
-      await disconnectRedisClient(redisClient, logger);
       delete redisClients[url];
+      await disconnectRedisClient(redisClient, logger);
       logger?.debug({
         at: "RedisUtils#getRedis",
         message: `Failed to connect to redis server at ${url}.`,
@@ -145,6 +145,9 @@ export async function disconnectRedisClients(logger?: winston.Logger): Promise<v
       message: "Disconnecting from redis server.",
       url,
     };
+    // We should delete the client from our cache object before
+    // we disconnect from redis.
+    delete redisClients[url];
     // We don't want to throw an error if we can't disconnect from redis.
     // We can log the error and continue.
     try {
@@ -153,10 +156,6 @@ export async function disconnectRedisClients(logger?: winston.Logger): Promise<v
     } catch (e) {
       logParams["success"] = false;
       logParams["error"] = e;
-    } finally {
-      // No matter what we need to eject this from
-      // our memory cache object.
-      delete redisClients[url];
     }
     logger?.debug(logParams);
   }
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -71,8 +71,9 @@ const redisClients: { [url: string]: RedisClient } = {};
 
 export async function getRedis(logger?: winston.Logger, url = REDIS_URL): Promise<RedisClient | undefined> {
   if (!redisClients[url]) {
+    let redisClient: _RedisClient | undefined = undefined;
     try {
-      const redisClient = createClient({ url });
+      redisClient = createClient({ url });
       await redisClient.connect();
       logger?.debug({
         at: "RedisUtils#getRedis",
@@ -86,6 +87,9 @@ export async function getRedis(logger?: winston.Logger, url = REDIS_URL): Promis
         message: `Failed to connect to redis server at ${url}.`,
         error: String(err),
       });
+      if (redisClient) {
+        await redisClient.disconnect();
+      }
     }
   }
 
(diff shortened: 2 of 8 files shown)


src/dataworker/index.ts
```diff
@@ -180,8 +180,7 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Wallet)
         break;
       }
     }
-  } catch (error) {
-    await disconnectRedisClient(logger);
-    throw error;
+  } finally {
+    await disconnectRedisClients(logger);
```
james-a-morris, line 184: Instead of catching to re-throw, we can simply use `finally` which has the same effect.

src/utils/RedisUtils.ts
```diff
@@ -132,12 +136,25 @@ export async function getDeposit(key: string, redisClient: RedisClient): Promise
   }
 }
 
-export async function disconnectRedisClient(logger?: winston.Logger): Promise<void> {
-  const redisClient = await getRedis(logger);
-  if (redisClient !== undefined) {
-    // todo understand why redisClient isn't GCed automagically.
-    logger.debug({ at: "disconnectRedisClient", message: "Disconnecting from redis server." });
-    await redisClient.disconnect();
+export async function disconnectRedisClients(logger?: winston.Logger): Promise<void> {
```
james-a-morris, line 139: We should be terminating **all** redis clients from this

nicholaspai (approved):

pxrl:
@james-a-morris I tested my local relayer-v2 instance from a few days ago, and it has only 1 redis connection per run. So the behaviour of having multiple parallel instances is new, probably per the recent SDK bump. I'd just like to be sure that we actually want multiple parallel connections, because the single shared instance has worked without complaint since it was introduced.

pxrl (commented):
@james-a-morris Can we double check that this PR is actually resolving the "redis connection explosion" that we saw last week? I suspect the high number of connections observed was solely due to the incidental placement of the log message within `RedisUtils.getRedis()`. Per the below code, we'd throw an exception on `logger` being undefined before saving the `RedisClient` reference in `redisClients`. Therefore, next time we went looking for a redis instance, we'd create a new one. If the log message was placed after the `redisClients[url] = ...` statement then I don't think we'd have seen the same connection storm.

2023-10-16-121820_664x341_scrot


One additional node: I'm wary of some cross-pollination between `src/caching/RedisCache.ts` and `src/utils/RedisUtils.ts`. These two files seem to import from (and call into) each other, and it's not really clear to me how these are intended to be used anymore. Since `RedisCache.ts` implements the global namespace, I suspect that we actually  want it to be the sole "entry point" for calling code. At the moment we seem to instantiate the Redis instance directly via `RedisUtils` as well, and it seems messy to me that `getRedisCache()` exists in `RedisUtils.ts`.

`RedisCache.ts`:
- Imports `RedisClient` and `getRedis()` from `RedisUtils.ts`.
- `RedisCache::instantiate()` calls `RedisUtils.getRedis()`.

`RedisUtils.ts`:
- Imports `RedisCache` from `RedisCache.ts`.

james-a-morris:
> @james-a-morris Can we double check that this PR is actually resolving the "redis connection explosion" that we saw last week? I suspect the high number of connections observed was solely due to the incidental placement of the log message within `RedisUtils.getRedis()`. Per the below code, we'd throw an exception on `logger` being undefined before saving the `RedisClient` reference in `redisClients`. Therefore, next time we went looking for a redis instance, we'd create a new one. If the log message was placed after the `redisClients[url] = ...` statement then I don't think we'd have seen the same connection storm.
> 
> 2023-10-16-121820_664x341_scrot
> 
> One additional node: I'm wary of some cross-pollination between `src/caching/RedisCache.ts` and `src/utils/RedisUtils.ts`. These two files seem to import from (and call into) each other, and it's not really clear to me how these are intended to be used anymore. Since `RedisCache.ts` implements the global namespace, I suspect that we actually want it to be the sole "entry point" for calling code. At the moment we seem to instantiate the Redis instance directly via `RedisUtils` as well, and it seems messy to me that `getRedisCache()` exists in `RedisUtils.ts`.
> 
> `RedisCache.ts`:
> 
> * Imports `RedisClient` and `getRedis()` from `RedisUtils.ts`.
> * `RedisCache::instantiate()` calls `RedisUtils.getRedis()`.
> 
> `RedisUtils.ts`:
> 
> * Imports `RedisCache` from `RedisCache.ts`.

This PR is a cleanup of our existing logic. The changes that fix this connection storm are #1001 and #1002

src/utils/RedisUtils.ts
```diff
@@ -85,6 +90,9 @@ export async function getRedis(logger?: winston.Logger, url = REDIS_URL): Promis
           message: `Failed to connect to redis server at ${url}.`,
         });
       }
+      if (redisClient) {
+        await redisClient.disconnect();
+      }
```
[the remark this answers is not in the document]
james-a-morris: I'm of two minds here. 

1. I agree that Redis/the Redis client is incentivized to handle closing an incorrectly instantiated socket. In doing so, we should not have to worry about disconnecting.
2. We should assume that Redis/the Redis client is a simple wrapper with no real guarantees. This code is meant to avoid the case where the socket doesn't get closed. If Redis closes the socket, this line of code is a no-op. But, if it doesn't, this will disconnect a potentially open descriptor. In addition, this helps make the code more intuitive in the sense that we're being very explicit as to how we want to handle an error path with `.connect()`

I'm happy to go in either direction, but the rest of this code is a tech debt patch to how we close our redis instances. The fixes that should get us back to prod are in #1001 and #1002

james-a-morris:
@pxrl resolved offline

pxrl (approved):

james-a-morris:
@pxrl @nicholaspai re-requested b/c we have a new change to how we disconnect clients.

nicholaspai (approved):

nicholaspai (approved):

pxrl (approved):

src/utils/RedisUtils.ts
```diff
@@ -77,6 +82,8 @@ export async function getRedis(logger?: winston.Logger, url = REDIS_URL): Promis
       });
       redisClients[url] = new RedisClient(redisClient, globalNamespace);
     } catch (err) {
+      await disconnectRedisClient(redisClient, logger);
+      delete redisClients[url];
```
mrice32, line 86: nit: I would flip these and delete it before disconnecting:
1. To avoid concurrency issues where it's used by someone else while being disconnected.
2. In case the disconnect fails for some reason and you never reach the deletion.

mrice32 (approved):
Looks great! Only one minor comment

src/utils/RedisUtils.ts
james-a-morris: Fair point - I'll do just that

james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/1000/commits/e196fcc249c353bde686c164ce09011658fd5786

We can do it in both places we attempt to remove from the object cache.

PR #1001, opened 2023-10-14
fix: let debugger gracefully not execute if undefined

james-a-morris:
This will resolve the failure we were seeing when we connected to Redis, then immediately failed. 

A note: there are no checks for this and it should be a compilation error to attempt to reference a function on a potentially undefined type.

An error sample that is now resolved:
```
2023-10-14 18:45:34 [debug]: {
  "at": "RedisUtils#getRedis",
  "message": "Connected to redis server at redis://xxxxxxxxxx successfully!",
  "dbSize": 37236,
  "bot-identifier": "xxxxxxxxxx"
}
2023-10-14 18:45:34 [debug]: {
  "at": "RedisUtils#getRedis",
  "message": "Failed to connect to redis server at redis://xxxxxxxxxx",
  "bot-identifier": "xxxxxxxxxx"
}
```

### DIFF
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -41,8 +41,8 @@ export class RedisClient {
       // EX: Expire key after expirySeconds.
       await this.client.set(key, val, { EX: expirySeconds });
     } else {
-      if (expirySeconds <= 0 && this.logger) {
-        this.logger.warn({
+      if (expirySeconds <= 0) {
+        this.logger?.warn({
           at: "RedisClient#setRedisKey",
           message: `Tried to set key ${key} with expirySeconds = ${expirySeconds}. This shouldn't be allowed.`,
         });
--- a/src/utils/RedisUtils.ts
+++ b/src/utils/RedisUtils.ts
@@ -20,7 +20,7 @@ export class RedisClient {
     private readonly namespace?: string,
     private readonly logger?: winston.Logger
   ) {
-    this?.logger.debug({
+    this.logger?.debug({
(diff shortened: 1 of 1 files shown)


src/utils/RedisUtils.ts
```diff
@@ -20,7 +20,7 @@ export class RedisClient {
     private readonly namespace?: string,
     private readonly logger?: winston.Logger
   ) {
-    this.logger.debug({
+    this?.logger.debug({
```
james-a-morris, line 23: This is the bug. The rest below are for consistency

nicholaspai (commented):
Why would there ever not be a logger defined?

james-a-morris:
> Why would there ever not be a logger defined?

https://github.com/across-protocol/relayer-v2/blob/7cf1afd09a30ef27916f6a1e22487a76904d91ba/src/utils/BlockUtils.ts#L29-L59

It's not passed in the `getRedis()` call within `getBlockForTimestamp` and that's a fairly core function

nicholaspai (approved):
Nice. There are other logger calls in this client that should also have the ? Operator by the way

james-a-morris:
> Nice. There are other logger calls in this client that should also have the ? Operator by the way

These are handled by the surrounding if statement, but we can add them just to be sure.

james-a-morris:
> 1e21bde

Updated for clarity: https://github.com/across-protocol/relayer-v2/pull/1001/commits/1e21bde76f1b34f47a4b84cf2f5701d8660af27f

pxrl (approved):

PR #1002, opened 2023-10-14
improve: make api stripping more robust

james-a-morris:
We are relying on a regular expression that's not entirely resistant to different network protocols. Let's use the built in `URL` class to pick out the `origin` field of the URL.

PR #1003, opened 2023-10-15
improve(ProviderUtils): Remove log about cache provider

nicholaspai:


james-a-morris (commented):
This is addressed here: https://github.com/across-protocol/relayer-v2/pull/1002

pxrl:
> This is addressed here: #1002

Aren't the two PRs proposing different things?

nicholaspai:
> > This is addressed here: #1002
> 
> Aren't the two PRs proposing different things?

they are. I am mostly indifferent but I think this log is overall not helpful and the regex hardcoded string looks like a source of future bugs so I propose removing this log. The other PR should still add `getOriginFromURL` but the rest of the logic would be moot from this pr

PR #1013, opened 2023-10-17
improve(relayer): Add filter for deposits whose quote blocks are in the future

nicholaspai:


src/utils/FillUtils.ts
```diff
@@ -273,7 +275,9 @@ export async function getUnfilledDeposits(
         originClient.getDepositsForDestinationChain(destinationChain);
 
       const unfilledDepositsForDestinationChain = depositsForDestinationChain
-        .filter((deposit) => deposit.blockNumber >= earliestBlockNumber)
+        .filter(
+          (deposit) => deposit.blockNumber >= earliestBlockNumber && deposit.quoteBlockNumber <= latestHubPoolBlock
```
james-a-morris, line 279: We check the condition `deposit.quoteBlockNumber <= latestHubPoolBlock` in the `SpokePool`'s `_isEarlyDeposit()` function.

This check may be better wrapped within the SpokePoolClient itself:

```
export class SpokePoolClient extends clients.SpokePoolClient {
  _isEarlyDeposit(depositEvent: FundsDepositedEvent, currentTime: number): boolean {
    const hubCurrentTime = this.hubPoolClient?.currentTime;
    if (!isDefined(hubCurrentTime)) {
      throw new Error("HubPoolClient's currentTime is not defined");
    }
    const quoteTime = depositEvent.args.quoteTimestamp;
    return quoteTime > currentTime || quoteTime > hubCurrentTime || deposit.blockNumber < deploymentBlock
  }
}
```

PR #1021, opened 2023-10-20
refactor: Simplify Relayer BasicFill test cases

pxrl:
Factor out a repetitive test pattern into the beforeAll() call.


nicholaspai (approved):
Thanks for this

PR #1026, opened 2023-10-20
feat: Dynamic profitability for deposits with messages

pxrl:


src/relayer/Relayer.ts
```diff
@@ -539,11 +539,16 @@ export class Relayer {
       : destinationChainId;
 
     const refundFee = this.computeRefundFee(version, deposit);
-    const { profitable, nativeGasCost } = profitClient.isFillProfitable(deposit, fillAmount, refundFee, hubPoolToken);
+    const { profitable, nativeGasCost } = await profitClient.isFillProfitable(
```
nicholaspai, line 542: Should we make this async call in parallel with the one above it?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: `Relayer.computeRefundFee()` is a synchronous call; we could wrap it async, but it'd probably run to completion before `ProfitClient.isFillProfitable()` was executed. Also, for now, `Relayer.computeRefundFee()` just returns a hardcoded `toBN(0)` because we're pre-UBA.

src/clients/ProfitClient.ts
```diff
@@ -455,9 +469,9 @@ export class ProfitClient {
         depositId,
         depositor: TEST_RECIPIENT,
         recipient: TEST_RECIPIENT,
-        originToken: "", // not relevant
+        originToken: "", // Not verified by the SpokePool.
```
james-a-morris, line 472: Thoughts on removing this entirely and typecasting the resultant object?

src/clients/ProfitClient.ts
```diff
@@ -181,21 +181,30 @@ export class ProfitClient {
...
     }
 
-    return ethersConstants.Zero; // @todo: Return uint256max in a follow-up change.
+    const { relayerAddress, relayerFeeQueries } = this;
+    try {
+      const gasCost = await relayerFeeQueries[chainId].getGasCosts(deposit, fillAmount, relayerAddress);
+      return toBN(gasCost); // BigNumberish -> BigNumber
+    } catch {
+      return bnZero;
+    }
```
james-a-morris, line 190: Are we trying to suppress the `catch` entirely here? Thoughts on adding a debugging log

pxrl: We already log generally that gas estimation failed, but I have updated the PR to log with warning here: https://github.com/across-protocol/relayer-v2/pull/1026/commits/6fb190bbcdb52b627ea58a2d05471487ffd70ac1

Log level warning rather than debug because it was specifically identified that it should be loud so we can see and debug these issues in prod.

src/clients/ProfitClient.ts
pxrl: That saves 2 lines of code but opens up for new errors if we change the type definition for Deposit, since they wouldn't necessarily be caught here. I'd probably prefer to keep it as-is - wdyt?

PR #1032, opened 2023-10-23
improve: Upgrade to sdk-v2 version with optimized caching logic

nicholaspai:
Should see speed ups in dataworker by:
- reducing time to cache utilization in HubPoolClient
- use redis cache for block finder getBlockForTimestamp functions

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -13,7 +13,8 @@ export class HubPoolClient extends clients.HubPoolClient {
     deploymentBlock?: number,
     chainId = 1,
     eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    cachingMechanism?: interfaces.CachingMechanismInterface
+    cachingMechanism?: interfaces.CachingMechanismInterface,
+    timeToCache?: number
   ) {
     super(
       logger,
@@ -25,6 +26,7 @@ export class HubPoolClient extends clients.HubPoolClient {
       {
         ignoredHubExecutedBundles: IGNORED_HUB_EXECUTED_BUNDLES,
         ignoredHubProposedBundles: IGNORED_HUB_PROPOSED_BUNDLES,
+        timeToCache,
       },
       cachingMechanism
     );
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -43,7 +43,7 @@ export class CommonConfig {
     this.version = ACROSS_BOT_VERSION ?? "unknown";
 
     this.timeToCache = Number(HUB_POOL_TIME_TO_CACHE ?? 60 * 60); // 1 hour by default.
-    if (this.timeToCache < 0) {
+    if (Number.isNaN(this.timeToCache) || this.timeToCache < 0) {
       throw new Error("Invalid default caching safe lag");
     }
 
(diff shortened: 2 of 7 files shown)


src/utils/BlockUtils.ts
```diff
@@ -30,29 +31,5 @@ export async function getBlockForTimestamp(
...
-
-    // Expire key after 90 days.
-    if (shouldCache(timestamp, currentBlock.timestamp)) {
-      await setRedisKey(key, blockNumber.toString(), redisClient, 60 * 60 * 24 * 90);
-    }
-    return blockNumber;
-  } else {
-    return parseInt(result);
-  }
+  return getCachedBlockForTimestamp(chainId, timestamp, blockFinder ?? (await getBlockFinder(chainId)));
```
pxrl, line 34: Not a showstopper, but resolving `blockFinder` this way is a bit dense.

pxrl (approved):
Looks OK to me, pending others' comments.

src/clients/HubPoolClient.ts
```diff
@@ -12,11 +12,13 @@ export class HubPoolClient extends clients.HubPoolClient {
     configStoreClient: clients.AcrossConfigStoreClient,
     deploymentBlock?: number,
     chainId = 1,
-    eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 }
+    eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
+    timeToCache = 60 * 60
```
mrice32, line 16:
```suggestion
    timeToCache?: number
```

Why not just let this be undefined and use the default from the superclass?

src/utils/BlockUtils.ts
mrice32: Agreed, I would put the blockfinder resolution on a separate line above.

james-a-morris, line 34:
```suggestion
  return utils.getCachedBlockForTimestamp(chainId, timestamp, blockFinder ?? (await getBlockFinder(chainId)));
```

src/utils/BlockUtils.ts
```diff
@@ -1,4 +1,5 @@
-import { BlockFinder, getProvider, getRedis, isDefined, setRedisKey, shouldCache } from "./";
+import { BlockFinder, getProvider, isDefined } from "./";
+import { getCachedBlockForTimestamp } from "@across-protocol/sdk-v2";
```
james-a-morris:
```suggestion
import { utils } from "@across-protocol/sdk-v2";
```

mrice32 (approved):
Looks good!

src/common/Config.ts
```diff
@@ -36,10 +37,16 @@ export class CommonConfig {
       SPOKE_POOL_CHAINS_OVERRIDE,
       ACROSS_BOT_VERSION,
       ACROSS_MAX_CONFIG_VERSION,
+      HUB_POOL_TIME_TO_CACHE,
     } = env;
 
     this.version = ACROSS_BOT_VERSION ?? "unknown";
 
+    this.timeToCache = Number(HUB_POOL_TIME_TO_CACHE ?? 60 * 60); // 1 hour by default.
+    if (this.timeToCache < 0) {
```
james-a-morris, line 46:
```suggestion
    if (Number.isNaN(this.timeToCache) || this.timeToCache < 0) {
```

nicholaspai: why do you think this is necessary? this seems redundant

james-a-morris: We should probably check for `Number.isNaN()` since `this.timeToCache < 0` fails if `this.timeToCache` is not a number

james-a-morris (approved):

PR #1034, opened 2023-10-23
feat: add basic response validation to the RetryProvider

mrice32:
This adds a simple method to make sure we don't accept a null or undefined response. In the future, we can enhance this logic by validating that it contains the fields, etc that we expect.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -426,6 +426,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       logger.debug({
         at: "ProviderUtils",
         message: "Provider returned invalid response",
+        provider: getOriginFromURL(provider.connection.url),
         method,
         params,
         response,
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -431,7 +431,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
         params,
         response,
       });
-      throw new Error(`Response failed validation for method ${method}, params ${params}, response ${response}`);
+      throw new Error("Response failed validation");
     }
     return response;
   }
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -415,13 +415,24 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     return isDefined(response);
   }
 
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

james-a-morris (approved):
LGTM - I wonder if we can start to move this away from the promise.then.catch syntax and normalize it with the majority of the codebase

src/utils/ProviderUtils.ts
```diff
@@ -410,10 +410,24 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
 
+  _validateResponse(method: string, params: Array<any>, response: any): boolean {
+    // Basic validation logic to start.
+    return isDefined(response);
+  }
+
+  _sendAndValidate(provider: ethers.providers.StaticJsonRpcProvider, method: string, params: Array<any>): Promise<any> {
+    return provider.send(method, params).then((response) => {
+      if (!this._validateResponse(method, params, response)) {
+        throw new Error(`Response failed validation for method ${method}, params ${params}, response ${response}`);
```
pxrl, line 421: Is it worth logging something here, rather than just bundling it all into the exception message? If we log it with some structure then it might be easier to parse/inspect when something goes wrong.

mrice32: Done. Would you like me to remove these fields from the error?

mrice32:
> LGTM - I wonder if we can start to move this away from the promise.then.catch syntax and normalize it with the majority of the codebase

Good point. In one case the .then syntax is probably the simplest implementation, since it's used to build up a promise chain.

But I removed it in the new code.

PTAL

src/utils/ProviderUtils.ts
```diff
@@ -410,10 +410,35 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+    method: string,
+    params: Array<any>
+  ): Promise<any> {
+    const response = await provider.send(method, params);
+    if (!this._validateResponse(method, params, response)) {
+      // Not a warning to avoid spam since this could trigger a lot.
+      logger.debug({
+        at: "ProviderUtils",
+        message: "Provider returned invalid response",
+        method,
```
pxrl, line 429:
```suggestion
        provider: getOriginFromUrl(provider.connection.url),
        method,
```

pxrl (approved):

james-a-morris (approved):
Sounds good. I agree that this is probably the most ideal usage of the bluebird-esque promise structure.

src/utils/ProviderUtils.ts
mrice32: Done

PR #1035, opened 2023-10-24
improve(spokepool): Dynamic relayerFeePct pricing on deposits

pxrl:
This will be useful for verifying messaging.

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -1,12 +1,22 @@
+import axios, { isAxiosError } from "axios";
 import minimist from "minimist";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { ExpandedERC20__factory as ERC20 } from "@across-protocol/contracts-v2";
 import { LogDescription } from "@ethersproject/abi";
 import { Contract, ethers, Wallet } from "ethers";
-import { groupBy } from "lodash";
+import { groupBy as _groupBy } from "lodash";
 import { config } from "dotenv";
-import { getNetworkName, getSigner, resolveTokenSymbols } from "../src/utils";
+import { BigNumber, formatFeePct, getNetworkName, getSigner, isDefined, resolveTokenSymbols, toBN } from "../src/utils";
 import * as utils from "./utils";
 
+type relayerFeeQuery = {
+  token: string;
+  originChainId: number;
+  destinationChainId: number;
+  amount: BigNumber;
+};
+
+const { fixedPointAdjustment: fixedPoint } = sdkUtils;
 const { MaxUint256, Zero } = ethers.constants;
 const { isAddress } = ethers.utils;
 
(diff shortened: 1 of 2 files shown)


pxrl:
Example usage:

```
$ yarn deposit --from 42161 --to 324 --token usdc --amount 5 --wallet secret --message "0xc123"
yarn run v1.22.19
$ yarn ts-node ./scripts/spokepool.ts deposit --from 42161 --to 324 --token usdc --amount 5 --wallet secret --message 0xc123
$ dev/across-protocol/relayer-v2/node_modules/.bin/ts-node ./scripts/spokepool.ts deposit --from 42161 --to 324 --token usdc --amount 5 --wallet secret --message 0xc123
Relayer fee quote for 5.00 USDC 42161 -> 324: 9.73 % (0.4867 USDC) (y/n) y
Submitting USDC deposit on Arbitrum: 0x0cfc799e06af2f1f0d9f7723e5502403d198831dc7e43a4c3bb6a3fdae093f4b.
Deposit # 1184319 on Arbitrum:
        tokenSymbol        : USDC
        amount             : 5000000
        originChainId      : 42161
        destinationChainId : 324
        relayerFeePct      : 97352200500000000
        depositId          : 1184319
        quoteTimestamp     : 1698150781
        originToken        : 0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8
        recipient          : 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D
        depositor          : 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D
        message            : 0xc123

Done in 17.51s.
```

nicholaspai (approved):
LGTM

scripts/spokepool.ts
```diff
@@ -47,10 +57,54 @@ function printFill(log: LogDescription): void {
...
 }
 
+async function getRelayerFeePct(request: relayerFeeQuery, timeout = 3000): Promise<BigNumber> {
+  const host = "across.to";
+  const path = "api/suggested-fees";
+  const url = `https://${host}/${path}`;
+
+  try {
+    const quote = await axios.get(url, { timeout, params: request });
+    return quote?.data["relayFeePct"];
```
mrice32, line 67: nit: why the ? -- won't it error anyway if quote is undefined because the next part of the chain will fail?

Also do you want it to error if the data doesn't contain the "relayFeePct" field? Right now, it will just return undefined I think.

scripts/spokepool.ts
```diff
@@ -47,10 +57,54 @@ function printFill(log: LogDescription): void {
...
+async function getRelayerQuote(
+  fromChainId: number,
+  toChainId: number,
+  token: utils.ERC20,
+  amount: BigNumber
+): Promise<BigNumber> {
+  const tokenFormatter = sdkUtils.createFormatFunction(2, 4, false, token.decimals);
+  let relayerFeePct = Zero;
+  let quoteAccepted = false;
+  do {
```
mrice32, line 85: What's the use case of this "accept a quote" looping functionality?

mrice32 (commented):
Looks great! Only two minor comments

james-a-morris:
@pxrl could we not use this? https://github.com/across-protocol/sdk-v2/tree/master/src/apiClient

james-a-morris (commented):
https://github.com/across-protocol/relayer-v2/pull/1035#issuecomment-1777647422

scripts/spokepool.ts
pxrl: Once the user outsources the relayerFeePct to the API, it just seems sensible to me that they should be prompted to confirm accept the quote before proceeding. We have historically seen some mis-quoting where the relayerFeePct was set relatively high on a high amount, and that's risky if the script just blazes away and submits the txn.

It can be bypassed by supplying the relayerFeePct at runtime, in which case `suggested-fees` won't be queried.

It's perhaps a fair argument that the user should also be prompted to accept the LP fee... I hadn't prioritised that because it's generally a lot more stable than the relayer fee and is determined centrally by the protocol.

scripts/spokepool.ts
pxrl: Good call on both points :+1:

pxrl:
> @pxrl could we not use this? https://github.com/across-protocol/sdk-v2/tree/master/src/apiClient

Sure - (discussed offline) pending some updates so that it catches the 4xx error responses and supplies them back, I think that'd be an improvement :+1:

pxrl:
> > @pxrl could we not use this? https://github.com/across-protocol/sdk-v2/tree/master/src/apiClient
> 
> Sure - (discussed offline) pending some updates so that it catches the 4xx error responses and supplies them back, I think that'd be an improvement 👍

Will circle back to this, because I'd like to get the benefit of this feature in master now, and the apiClient in sdk-v2 needs some updates and requires an SDK version bump.

scripts/spokepool.ts
pxrl: Updated: https://github.com/across-protocol/relayer-v2/pull/1035/commits/ddee49aa43b37e129daa06c083ffc59f26eefac9

pxrl: Updated: https://github.com/across-protocol/relayer-v2/pull/1035/commits/ddee49aa43b37e129daa06c083ffc59f26eefac9

scripts/spokepool.ts
pxrl: Happy to revisit this if we can find a better way to handle it :+1:

PR #1036, opened 2023-10-24
fix: Use RL relayer address for basic gas cost estimation

pxrl:
It's invalid to use the current relayer's EOA because that address may not hold USDC on all chains. Instead, default to the main RL address.

The relayer's own address will be used for dynamic message-based simulation.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -20,7 +20,7 @@ import { Deposit, DepositWithBlock, L1Token, SpokePoolClientsByChain } from "../
 import { HubPoolClient } from ".";
 
 const { formatEther } = ethersUtils;
-const { EMPTY_MESSAGE } = sdkConsts;
+const { EMPTY_MESSAGE, DEFAULT_SIMULATED_RELAYER_ADDRESS: TEST_RELAYER } = sdkConsts;
 const { bnOne, bnUint32Max, fixedPointAdjustment: fixedPoint, isMessageEmpty, resolveDepositMessage } = sdkUtils;
 
 // We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):

james-a-morris:
LGTM.

My only question: will setting the `relayer`, ' depositor`, and `recipient` to the same addresses affect gas computation? It may be worth taking a quick look.

src/clients/ProfitClient.ts
```diff
@@ -436,10 +428,14 @@ export class ProfitClient {
...
+
+      // @dev The relayer _can not_ be the recipient, since the SpokePool short-circuits the ERC20 transfer
+      // and consumes less gas. Instead, just use the main RL address as the simulated relayer, since it has
+      // all supported tokens and approvals in place on all chains.
       const deposit: Deposit = {
         depositId,
-        depositor: TEST_RECIPIENT,
-        recipient: TEST_RECIPIENT,
+        depositor: relayerAddress,
+        recipient: relayerAddress,
```
nicholaspai, line 438: This function is only used for empty message deposits right?

src/clients/ProfitClient.ts
```diff
@@ -452,7 +448,7 @@ export class ProfitClient {
       };
 
       // An extra toBN cast is needed as the provider returns a different BigNumber type.
-      const gasCost = await relayerFeeQueries[destinationChainId].getGasCosts(deposit, fillAmount, relayerAddress);
+      const gasCost = await relayerFeeQueries[destinationChainId].getGasCosts(deposit, fillAmount, TEST_RELAYER);
```
nicholaspai, line 451: Per the PR description, shouldn't this be TEST_RECIPIENT if the message is hardcoded to empty?

pxrl: Yep, good catch. It was an error to have removed that.

src/clients/ProfitClient.ts
pxrl: Yep - this just replicates the pre-existing behaviour of pre-computing the gas cost based on a small USDC fill.

pxrl:
> LGTM.
> 
> My only question: will setting the `relayer`, ' depositor`, and `recipient` to the same addresses affect gas computation? It may be worth taking a quick look.

There's no issue for the depositor address since the destination SpokePool doesn't really care about it (beyond hashing it for fill tracking). it's just important that relayer != recipient, because in that case the SpokePool skips the ERC20 transfer.

nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #1037, opened 2023-10-24
improve: Suppress RPC provider key from error messages

pxrl:


src/utils/ProviderUtils.ts
pxrl, line 1: fwiw I am surprised how many instances we had of this in the code :thinking:

nicholaspai (approved):
damn nice catch...

nicholaspai:
> damn nice catch...

Wish we could add a linting rule to prevent this somehow

@james-a-morris any ideas?

mrice32 (approved):

mrice32:
> > damn nice catch...
> 
> Wish we could add a linting rule to prevent this somehow
> 
> @james-a-morris any ideas?

Good idea on linting this!

src/utils/ProviderUtils.ts
```diff
@@ -72,8 +72,14 @@ class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
 
 const defaultTimeout = 60 * 1000;
 
+// @dev To avoid accidentally leaking RPC keys in log messages, resolve the RPC provider hostname
+// centrally. There should be no instances of `provider.connection.url` in log messages or errors.
+function getProviderHostname(provider: ethers.providers.StaticJsonRpcProvider): string {
+  return new URL(provider.connection.url).hostname;
+}
```
james-a-morris, line 79: We have `getOriginFromUrl( )` - should we introduce this second one? We can probably overload or make a similar function called `getOriginFromProvider(provider: ...)`

src/utils/ProviderUtils.ts
```diff
@@ -72,8 +72,14 @@ class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
 
 const defaultTimeout = 60 * 1000;
 
+// @dev To avoid accidentally leaking RPC keys in log messages, resolve the RPC provider hostname
+// centrally. There should be no instances of `provider.connection.url` in log messages or errors.
+function getProviderHostname(provider: ethers.providers.StaticJsonRpcProvider): string {
+  return new URL(provider.connection.url).hostname;
```
james-a-morris, line 78: The `hostname` will clip away the network protocol & potential port. Using `origin` will futureproof us if we ever want to intermingle differing protocols or separate ports on the same host.

james-a-morris:
> Member

cc @pxrl RE: https://github.com/across-protocol/relayer-v2/pull/1038

I think we should put all usages to `connection.url` in one file that's separated from `ProviderUtils.ts` - a file where we accept the usage of referencing `provider.connection.url`. I've created a CI check so that we can isolate these calls and fail if they are referenced.

PR #1038, opened 2023-10-24
feat: avoid directly using

james-a-morris:
This includes a new test in CircleCI that looks for a direct reference to `provider.connection.url`. We _almost never_ need to reference this path except to print logs. We should do that in one location centrally and fail a CI build if we encounter this throughout the rest of the code.

Note: this is not meant to be a true security check for leaked secrets. There are many ways to bypass this check. However, for most normal actors who are using the code in the style that we are used to seeing, this CI build should flag on accidental pre-exposure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,21 @@ jobs:
       - run:
           name: Run lint
           command: yarn lint
+  no-leaked-logs:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - checkout
+      - run:
+          name: Search for connection.url
+          command: |
+            if find . -type f -not -path "./node_modules/*" -not -path "./src/utils/NetworkUtils.ts" -exec grep -H "connection.url" {} +; then
+              echo "Found 'connection.url'. Failing the test."
+              exit 1
+            else
+              echo "'connection.url' not found."
+              exit 0
+            fi
 workflows:
   version: 2.1
   build_and_test:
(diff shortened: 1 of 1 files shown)


james-a-morris:
looping in @evaldofelipe because he's a CI wizard

.circleci/config.yml
```diff
@@ -53,6 +53,21 @@ jobs:
...
           command: yarn lint
+  no-leaked-logs:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - checkout
+      - run:
+          name: Search for connection.url
+          command: |
+            if find . -type f -not -path "./node_modules/*" -not -path "./src/utils/NetworkUtils.ts" -exec grep -H "connection.url" {} +; then
```
pxrl, line 64: How about something like `grep -q 'connection.url' src test` ?

BTW, the file to exclude is `src/utils/ProviderUtils.ts` - that's where we use `provider.connection.url`.

mrice32 (approved):
Good idea! LGTM once @pxrl is happy with the solution!

.circleci/config.yml
james-a-morris: Regarding the use of `grep`, I agree that it is more efficient and I like it.

However, I have some concerns about directly referencing `provider.connection.url` within a function in  `ProviderUtils.ts`. 

Our previous issues have shown that this is a problem file, so I propose creating a separate file that only exports a function called `getOriginFromProvider()`. This way, we can minimize the number of references to `connection.url` and allow this check in the file that's most likely to benefit from it: ProviderUtils.ts

PR #1039, opened 2023-10-24
improve: use SDK for verifying `isMessageEmpty`

james-a-morris:
Utilize the sdk implementation of `isMessageEmpty`

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1455,25 +1455,8 @@ export class Dataworker {
     // Ignore slow fill leaves for deposits with messages as these messages might be very expensive to execute.
     // The original depositor can always execute these and pay for the gas themselves.
     const leaves = _leaves.filter(({ payoutAdjustmentPct, relayData }) => {
-      // We don't have access to any speed up requests, so we don't have the full picture
-      // just looking at the relay data. Let's enrich it with the latest speed up requests.
-      const depositWithSpeedup = client.appendMaxSpeedUpSignatureToDeposit(
-        // We don't have access to the full deposit, so we just use the relay data. It has
-        // all the information we need to resolve the speed up parameters. Per the function,
-        // we really just need the deposit ID, the depositor, and relayerFeePct
-        {
-          depositId: relayData.depositId,
-          depositor: relayData.depositor,
-          relayerFeePct: relayData.relayerFeePct,
-          message: relayData.message, // Pass this in for the next function
-        } as DepositWithBlock
-      );
-      // This function retrieves either the updated message or the original message if no speed up request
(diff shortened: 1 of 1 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1454,8 +1454,9 @@ export class Dataworker {
   ): Promise<void> {
     // Ignore slow fill leaves for deposits with messages as these messages might be very expensive to execute.
     // The original depositor can always execute these and pay for the gas themselves.
-    const leaves = _leaves.filter(({ relayData, payoutAdjustmentPct }) => {
-      if (relayData.message !== "0x") {
+    const leaves = _leaves.filter(({ payoutAdjustmentPct, relayData }) => {
+      // If there is a message, we ignore the leaf and log an error.
+      if (!sdk.utils.isMessageEmpty(relayData.message)) {
```
pxrl, line 1459: Placeholder for discussion about whether the dataworker should ever produce a slow fill based on an updated deposit. If not, it might produce a slow fill that's unable to be executed if the recipient/message combination is invalid.

mrice32 (approved):

src/dataworker/Dataworker.ts
nicholaspai: The slow fill `message` needs to be equal to the original `message` in the deposit. Currently the SpokePool doesn't provide an interface that someone can use to execute a slow fill with an updated message.

nicholaspai: https://github.com/across-protocol/relayer-v2/pull/1039#discussion_r1370766660

agreed with this comment. One way we could support this would be to add a `_verifyUpdateDepositMessage` call to the `executeSlowLeaf` function.

src/dataworker/Dataworker.ts
```diff
@@ -1454,8 +1454,9 @@ export class Dataworker {
...
         this.logger.error({
```
nicholaspai, line 1460: As discussed offline, let's make this `warn` so that if a slow fill gets through with a message it doesn't keep Paging

mrice32: +1 good point

james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1039/commits/648f9bd2d2c492661331f74d1b90ba4ff2c4acfd

src/dataworker/Dataworker.ts
james-a-morris: Sounds good. This is something to track as a ticket. I'm not sure it would be good to serve in this PR

pxrl (approved):

PR #1042, opened 2023-10-25
feat: Suppress zero fills on deposits with messages

pxrl:


pxrl:
todo: follow up with a small test to verify no zero-fills.

nicholaspai (commented):
Should we also check that the recipient is a contract or is that overkill? Without this check we might get some false negatives but it shouldn't happen often in prod

nicholaspai (approved):

pxrl:
> Should we also check that the recipient is a contract or is that overkill? Without this check we might get some false negatives but it shouldn't happen often in prod

Hm, so you're thinking about the scenario where the deposit has a message attached, and the recipient is an EOA, so we'd actually be OK with initiating a slow fill for that?

src/relayer/Relayer.ts
```diff
@@ -352,6 +352,14 @@ export class Relayer {
    * @param deposit Deposit object to zero-fill.
    */
   zeroFillDeposit(deposit: Deposit): void {
+    if (!isMessageEmpty(deposit.message)) {
```
nicholaspai, line 355: Can you add a comment that if the depositor has sped up their message and zero'd their message, we still won't partial fill their deposit, but we can now full fill it

james-a-morris, line 355: Good call

james-a-morris (approved):

src/relayer/Relayer.ts
pxrl: Updated: https://github.com/across-protocol/relayer-v2/pull/1042/commits/99e2b65d80c7e17365eb3f49bff3df62ee7eed7b

PR #1043, opened 2023-10-25
fix(relayer): Add config to skip rebalancing

nicholaspai:


pxrl (approved):
This looks fine, but I wonder if it'd be better to opt-in to rebalancing, rather than opting out of it. Having to negate the config option in order to do something seems cleaner to me. But it's only a personal nit.

mrice32 (approved):

nicholaspai:
> This looks fine, but I wonder if it'd be better to opt-in to rebalancing, rather than opting out of it. Having to negate the config option in order to do something seems cleaner to me. But it's only a personal nit.

Yeah but i do think this is consistent with how we send relays by default and opt in to skipping them.

These shoud all be reversed

PR #1044, opened 2023-10-25
fix: ignore eth_getTransactionReceipt when checking response validity

mrice32:


james-a-morris (approved):
Subtle - good fix!

pxrl (approved):
LGTM. For our future forgetful selves it might be interesting to record exactly why we made this change - i.e. that ethers seems to re-poll on null responses until it gets an actual transaction receipt, and by concluding failure early we seem to be causing premature retries and rotation through the configured rpc providers. wdyt?

nicholaspai (approved):

mrice32:
> LGTM. For our future forgetful selves it might be interesting to record exactly why we made this change - i.e. that ethers seems to re-poll on null responses until it gets an actual transaction receipt, and by concluding failure early we seem to be causing premature retries and rotation through the configured rpc providers. wdyt?

Done -- added a comment!

pxrl (approved):

PR #1045, opened 2023-10-25
improve: Support relayer self-filling despite insufficient balance

pxrl:
The relayer doesn't need to have have enough of the token to complete a                                                                                                                        
fill when its own address is the recipient of the transfer.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -259,7 +259,8 @@ export class Relayer {
         continue;
       }
 
-      if (tokenClient.hasBalanceForFill(deposit, unfilledAmount)) {
+      const selfRelay = [depositor, recipient].every((address) => address === this.relayerAddress);
+      if (tokenClient.hasBalanceForFill(deposit, unfilledAmount) && !selfRelay) {
         // The pre-computed realizedLpFeePct is for the pre-UBA fee model. Update it to the UBA fee model if necessary.
         // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
         deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit);
@@ -275,7 +276,7 @@ export class Relayer {
         } else {
           profitClient.captureUnprofitableFill(deposit, unfilledAmount, gasCost);
         }
-      } else if ([depositor, recipient].every((address) => address === this.relayerAddress)) {
+      } else if (selfRelay) {
         // A relayer can fill its own deposit without an ERC20 transfer. Only bypass profitability requirements if the
         // relayer is both the depositor and the recipient, because a deposit on a cheap SpokePool chain could cause
         // expensive fills on (for example) mainnet.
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

src/relayer/Relayer.ts
```diff
@@ -275,6 +275,11 @@ export class Relayer {
         } else {
           profitClient.captureUnprofitableFill(deposit, unfilledAmount, gasCost);
         }
+      } else if (depositor === this.relayerAddress && recipient === this.relayerAddress) {
```
james-a-morris, line 278: Thoughts on making this `[depositor, recipient].includes(this.relayerAddress)` just to make it a bit more concise?

src/relayer/Relayer.ts
```diff
@@ -275,6 +275,11 @@ export class Relayer {
...
+        // A relayer can fill its own deposit without an ERC20 transfer. Only bypass profitability requirements if the
+        // relayer is both the depositor and the recipient, because a deposit on a cheap SpokePool chain could cause
+        // expensive fills on (for example) mainnet.
+        this.fillRelay(deposit, unfilledAmount, destinationChainId);
```
james-a-morris, line 282: QQ: this may require one less transfer but are we guaranteed to have enough to make this txn?

src/relayer/Relayer.ts
```diff
@@ -275,6 +275,11 @@ export class Relayer {
         } else {
           profitClient.captureUnprofitableFill(deposit, unfilledAmount, gasCost);
         }
+      } else if (depositor === this.relayerAddress && recipient === this.relayerAddress) {
+        // A relayer can fill its own deposit without an ERC20 transfer. Only bypass profitability requirements if the
+        // relayer is both the depositor and the recipient, because a deposit on a cheap SpokePool chain could cause
+        // expensive fills on (for example) mainnet.
```
mrice32, line 281: So the depositor check is to avoid spam? Just making sure I understand

mrice32 (approved):
Looks good!

src/relayer/Relayer.ts
pxrl: Correct - otherwise someone else could make a small deposit on Arbitrum, and could trick the relayer into self-filling at a high gas price on (for example) mainnet. It seems like that could provide leverage for burning a relayer's gas.

src/relayer/Relayer.ts
pxrl: I love using `Array.includes()`, but I don't think it's useful in this case. We specifically want to ensure that both depositor and recipient is the relayer, whereas this check will only verify that one of them was.

pxrl: Settled on `Array.every()` in https://github.com/across-protocol/relayer-v2/pull/1045/commits/fafc6ae2edd7b397daee4c2f7a351d458527fa24 - I think this is what we were looking for :+1:

PR #1046, opened 2023-10-26
improve(ProviderUtils): Temporarily ignore block size field

pxrl:
Alchemy have been disagreeing with other providers on Arbitrum for a few days now; it causes higher RPC usage and unnecessary log noise. Disable it for now, with the option of reverting in future (if they fix it) or retaining it indefinitely, since we don't actually rely on this field.


james-a-morris (approved):

james-a-morris:
LGTM - thoughts on making this an ENV variable?

nicholaspai:
> LGTM - thoughts on making this an ENV variable?

I don't think we should make this an env var

PR #1055, opened 2023-10-30
refactor(ProfitClient): Overhaul gas token resolution

pxrl:
Specify the gas token for a chain by its symbol, and then dynamically resolve the underlying token on mainnet. Also, remove some hardcoding around the lookup of MATIC for Polygon. This is part of some cleanup to make it easier to run the relayer on a testnet.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -469,7 +469,7 @@ export class ProfitClient {
       const destinationToken =
         destinationChainId === hubPoolClient.chainId
           ? hubToken
-          : hubPoolClient.getDestinationTokenForL1Token(testSymbol, destinationChainId);
+          : hubPoolClient.getDestinationTokenForL1Token(hubToken, destinationChainId);
       assert(isDefined(destinationToken), `Chain ${destinationChainId} SpokePool is not configured for ${testSymbol}`);
 
       // @dev The relayer _can not_ be the recipient, since the SpokePool short-circuits the ERC20 transfer
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,4 +1,4 @@
-import { random } from "lodash";
+import { random, uniq } from "lodash";
 import { Provider } from "@ethersproject/abstract-provider";
 import { utils as ethersUtils } from "ethers";
 import {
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):
This makes sense +1

mrice32 (approved):

src/clients/ProfitClient.ts
```diff
@@ -420,12 +429,12 @@ export class ProfitClient {
...
     );
 
-    // Also include MATIC in the price queries as we need it for gas cost calculation.
-    l1Tokens[MATIC] = {
-      address: MATIC,
-      symbol: "MATIC",
-      decimals: 18,
-    };
+    // Also ensure all gas tokens are included in the lookup.
+    uniq(Object.values(GAS_TOKEN_BY_CHAIN_ID)).map((symbol) => {
```
james-a-morris, line 433: Thoughts on avoiding lodash? Could we instead use `new Set(Object.values(GAS_TOKEN_BY_CHAIN_ID))`

james-a-morris (commented):
Left a quick comment on lodash. To my understanding this is actually a large dependency we should look to remove.

src/clients/ProfitClient.ts
pxrl: Yeah I have used that before, but was never keen on all the boilerplate that it requires. `uniq()` just seemed nicer, but I wasn't clear on the bloat that lodash implied.

We do have it in a few places in this codebase, so it'll take some effort to hoist off of it. But I'm fine with not adding to that burden :+1:

pxrl: Addressed in this commit: https://github.com/across-protocol/relayer-v2/pull/1055/commits/c233afcdedbdcbe3896621333829d5e84f192094

I ended up skipping the part where we filter for unique gas token symbols. The code is quite simple and we only support a handful of chains, so the overhead of producing a unique array was probably greater than the brute force approach of trying every chain.

nicholaspai (approved):
Good refactor

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -104,6 +65,50 @@ describe("ProfitClient: Consider relay profit", () => {
   let hubPoolClient: MockHubPoolClient, profitClient: MockProfitClient;
   const message = sdkConstants.EMPTY_MESSAGE; // Empty message by default.
 
+  const randomiseGasCost = (chainId: number): BigNumber => {
+    // Randomise the gas token price (usd)
+    const gasToken = profitClient.resolveGasToken(chainId);
+    const gasPriceUsd = toBNWei(Math.random().toPrecision(10));
```
james-a-morris, line 71: Can we get toPrecision(decimals)

james-a-morris (approved):
LGTM - thanks for pulling as much lodash out as possible

test/ProfitClient.ConsiderProfitability.ts
pxrl: Done: https://github.com/across-protocol/relayer-v2/pull/1055/commits/97bcd4e21553da5381ecbc0d01cc0ef3e3ae4f1c. Used `.toFixed()` instead, since it's closer to what we want.

PR #1056, opened 2023-10-30
chore: Use sdk-v2 getABI()

pxrl:
This replaces the version from contracts-node and permits contracts-node to be dropped as a direct dependency.


mrice32 (approved):

mrice32 (approved):
Great removal!

PR #1057, opened 2023-10-31
chore: Update default signer to secret

pxrl:
This changes the default method of private key derivation from `mnemonic` to `secret`. Secret is preferred because it allows the user to store the private key separately from their .env file, which might otherwise be edited frequently or exposed via screen sharing.

While here, loosen the regex for the private key so that the leading 0x becomes optional. Anything that doesn't match the private key regex is assumed to be a mnemonic.


pxrl:
tbd - update docs to describe the use of `secret`.

pxrl:
Just reviving this PR now. The docs have been updated to describe how to configure secrets: https://docs.across.to/v/developer-docs/~/changes/vwVpZdrWiNKNakj8ftlf/developers/running-a-relayer

Also some screenshots:

secret-1
secret-2

src/utils/SignerUtils.ts
```diff
@@ -114,7 +114,7 @@ async function getSecretSigner(): Promise<Signer> {
   try {
     secret = await readFile(SECRET, { encoding: "utf8" });
     secret = secret.trim().replace("\n", "");
-    return /^0x[0-9a-f]{64}$/.test(secret) ? new Wallet(secret) : Wallet.fromMnemonic(secret);
+    return /^(0x)?[0-9a-f]{64}$/.test(secret) ? new Wallet(secret) : Wallet.fromMnemonic(secret);
```
mrice32, line 117: Is this making the 0x optional?

mrice32 (approved):
LGTM!

src/utils/SignerUtils.ts
pxrl: Yep! The 0x was initially enforced, but during testing with a third-party relayer we scratched our heads as to why it wasn't working until realising they'd pasted in a key without the leading 0x. ethers is fine with either variant, so I figured it was OK to avoid imposing an artificial constraint here.

PR #1058, opened 2023-11-02
chore: Drop Across API timeout from 60s to 3s

pxrl:
Most API queries should resolve in under 1 second, but there are occasions where we log that we hit the 60s timeout. This is a major slowdown for the bots. Since we now rotate to the next available route, if the first query doesn't resolve shortly after it was sent, we should just fail quickly and rotate instead.


nicholaspai (approved):
I agree with this. I think the vercel timeout is 1min

PR #1059, opened 2023-11-03
feat(ProfitClient): Support padding native gas units

pxrl:
This change introduces a new configurable RELAYER_GAS_PADDING, which is applied by the ProfitClient to raw gas estimates before they are fed into profitability calculations. This is effectively a way of accounting for marginally worse-than-estimated gas consumption when a simulated fill is ultimately executed on-chain.

There is some unfortunate overlap with the pre-existing RELAYER_GAS_MULTIPLIER, which is currently used to discount the estimated fill cost to account for gas price volatilility (i.e. to allow prioritisation of service over profitability at the margins).

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -38,6 +38,8 @@ const {
   resolveDepositMessage,
 } = sdkUtils;
 
+type TransactionCostEstimate = sdkUtils.TransactionCostEstimate;
+
 // We use wrapped ERC-20 versions instead of the native tokens such as ETH, MATIC for ease of computing prices.
 export const MATIC = TOKEN_SYMBOLS_MAP.MATIC.addresses[CHAIN_IDs.MAINNET];
 export const USDC = TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET];
@@ -50,9 +52,11 @@ export type FillProfit = {
   tokenPriceUsd: BigNumber; // Resolved USD price of the bridged token.
   fillAmountUsd: BigNumber; // Amount of the bridged token being filled.
   grossRelayerFeeUsd: BigNumber; // USD value of the relay fee paid by the user.
-  nativeGasCost: BigNumber; // Cost of completing the fill in the native gas token.
-  gasMultiplier: BigNumber; // Multiplier to apply to nativeGasCost as padding or discount
-  gasPriceUsd: BigNumber; // Price paid per unit of gas in USD.
+  nativeGasCost: BigNumber; // Cost of completing the fill in the units of gas.
+  tokenGasCost: BigNumber; // Cost of completing the fill in the relevant gas token.
+  gasPadding: BigNumber; // Positive padding applied to nativeGasCost and tokenGasCost before profitability.
+  gasMultiplier: BigNumber; // Multiplier applied to token-only fill cost estimates before profitability.
+  gasTokenPriceUsd: BigNumber; // Price paid per unit of gas the gas token in USD.
   gasCostUsd: BigNumber; // Estimated cost of completing the fill in USD.
   refundFeeUsd: BigNumber; // Estimated relayer refund fee on the refund chain.
   relayerCapitalUsd: BigNumber; // Amount to be sent by the relayer in USD.
@@ -64,6 +68,7 @@ export type FillProfit = {
 type UnprofitableFill = {
   deposit: DepositWithBlock;
   fillAmount: BigNumber;
+  gasCost: BigNumber;
   nativeGasCost: BigNumber;
 };
 
@@ -95,7 +100,7 @@ const TEST_RECIPIENT = "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
 // Note: the type here assumes that all of these classes take the same constructor parameters.
 const QUERY_HANDLERS: {
   [chainId: number]: new (
-    ...args: ConstructorParameters<typeof relayFeeCalculator.EthereumQueries>
+    ...args: ConstructorParameters<typeof relayFeeCalculator.BaseQueries>
   ) => relayFeeCalculator.QueryInterface;
 } = {
   1: relayFeeCalculator.EthereumQueries,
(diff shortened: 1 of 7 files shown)


pxrl:
nb. This PR depends on sdk-v2 version 0.17.11, which is yet to be produced - so building will fail until that bump can be bundled into this PR.

src/clients/ProfitClient.ts
```diff
@@ -95,7 +100,7 @@ const TEST_RECIPIENT = "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
 // Note: the type here assumes that all of these classes take the same constructor parameters.
 const QUERY_HANDLERS: {
   [chainId: number]: new (
-    ...args: ConstructorParameters<typeof relayFeeCalculator.EthereumQueries>
+    ...args: ConstructorParameters<typeof relayFeeCalculator.BaseQueries>
```
pxrl, line 103: This is just an incidental change that James made possible by a recent update to sdk-v2.

src/clients/ProfitClient.ts
```diff
@@ -139,14 +144,21 @@ export class ProfitClient {
     readonly relayerAddress: string,
     readonly defaultMinRelayerFeePct = toBNWei(constants.RELAYER_MIN_FEE_PCT),
     readonly debugProfitability = false,
-    protected gasMultiplier = toBNWei(1)
+    protected gasMultiplier = toBNWei(constants.DEFAULT_RELAYER_GAS_MULTIPLIER),
```
pxrl, line 147: Per discussion, we imagine that `gasMultiplier` might ultimately be removed, because it tends to complicate this picture. It may instead be simpler to just control `RELAYER_MIN_RELAY_FEE_PCT` directly if the operator wants to discount gas. That configurable already supports per-chain configuration, so it can be tuned for chains that have higher gas volatility, whereas `gasMultiplier` / `RELAYER_GAS_MULTIPLIER` is global.

src/clients/ProfitClient.ts
```diff
@@ -179,19 +191,18 @@ export class ProfitClient {
...
   // @todo: Factor in the gas cost of submitting the RefundRequest on alt refund chains.
-  async getTotalGasCost(deposit: Deposit, fillAmount = deposit.amount): Promise<BigNumber> {
+  async getTotalGasCost(deposit: Deposit, fillAmount = deposit.amount): Promise<TransactionCostEstimate> {
     const { destinationChainId: chainId } = deposit;
 
     // If there's no attached message, gas consumption from previous fills can be used in most cases.
     // @todo: Simulate this per-token in future, because some ERC20s consume more gas.
     if (isMessageEmpty(resolveDepositMessage(deposit)) && isDefined(this.totalGasCosts[chainId])) {
-      return toBN(this.totalGasCosts[chainId]);
+      return this.totalGasCosts[chainId];
```
pxrl, line 200: Removed `toBN()` because it's unnecessary.

src/common/Constants.ts
```diff
@@ -138,7 +138,8 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   421613: 0,
 };
 
-export const DEFAULT_RELAYER_GAS_MULTIPLIER = 1.2;
+export const DEFAULT_RELAYER_GAS_PADDING = ".15"; // Padding on token- and message-based relayer fill gas estimates.
```
pxrl, line 141: This means we'll pad gas cost estimates by 15%.

src/clients/ProfitClient.ts
```diff
@@ -201,34 +212,47 @@ export class ProfitClient {
...
+    // Pad these estimates before computing profitability to allow execution headroom and reduce the chance of an OoG.
+    nativeGasCost = nativeGasCost.mul(this.gasPadding).div(fixedPoint);
+    tokenGasCost = tokenGasCost.mul(this.gasPadding).div(fixedPoint);
 
-    if (gasPriceUsd.lte(0) || nativeGasCost.lte(0)) {
-      const err = gasPriceUsd.lte(0) ? "gas price" : "gas consumption";
-      throw new Error(`Unable to compute gas cost (${err} unknown)`);
+    // Gas estimates for token-only fills are stable and reliable. Allow these to be scaled up or down via the
+    // configured gasMultiplier. Do not scale the nativeGasCost, since it might be used to set the transaction gasLimit.
+    // @todo Consider phasing this out and relying solely on the minimum profitability config.
```
nicholaspai, line 245: +1

nicholaspai: It definitely feels like the gasMultiplier is superfluous and we should aim to remove it. It seems like this PR sets up well for a follow up to remove it, FWIW

src/clients/ProfitClient.ts
```diff
@@ -139,14 +144,21 @@ export class ProfitClient {
...
   ) {
-    // Require 1% <= gasMultiplier <= 400%
+    // Require 0% <= gasPadding <= 200%
     assert(
-      this.gasMultiplier.gte(toBNWei("0.00")) && this.gasMultiplier.lte(toBNWei(4)),
-      `Gas multiplier out of range (${this.gasMultiplier})`
+      this.gasPadding.gte(toBNWei("0")) && this.gasPadding.lte(toBNWei(2)),
+      `Gas padding out of range (${this.gasPadding})`
     );
+    this.gasPadding = toBNWei("1").add(gasPadding);
```
nicholaspai, line 155: Why configure gasPadding to be specified without the leading "1" whereas `gasMultiplier` has a leading 1? Would this cause accidental user errors?

pxrl: Given that gasPadding should never go below 0, it seems more intuitive to set it as a raw percentage (i.e. I want to pad my gas estimates by 0.15 pr 15%), rather than a multiplier (1.15).

I do agree clashes with the `RELAYER_GAS_MULTIPLIER` configurable, but I'm wondering whether that might be OK given that we think `RELAYER_GAS_MULTIPLIER` could be deprecated in future.

nicholaspai: Ok I think that's fine

nicholaspai (approved):

PR #1062, opened 2023-11-07
feat(MultiCallerClient): gasLimit plumbing

pxrl:
This change modifies the MultiCallerClient such that it will be more respectful of an enqueued transaction that has had a gasLimit applied.

There is one edge case that's not handled here: when an enqueued transaction with gasLimit is bundled together with other transactions that do not have a gasLimit, the gasLimit on the individual transaction reverts to undefined. This is because it's not possible to accurately set a gasLimit on the bundle without simulating each transaction on the fly, and that's actually what we try to avoid by bundling in the first place.

The caller can avoid this by ensuring that all enqueued transactions are simulated in advance. The relayer itself should implement this once some residual issues with higher-than-average ERC20 gas consumption on specific tokens has been resolved.

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -1,7 +1,10 @@
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { BigNumber } from "ethers";
 import { DEFAULT_MULTICALL_CHUNK_SIZE, DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE, Multicall2Call } from "../common";
 import {
   winston,
   getNetworkName,
+  isDefined,
   isPromiseFulfilled,
   getTarget,
   TransactionResponse,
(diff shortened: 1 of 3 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -162,9 +163,11 @@ export async function getGasPrice(
 }
 
 export async function willSucceed(transaction: AugmentedTransaction): Promise<TransactionSimulationResult> {
-  if (transaction.canFailInSimulation) {
+  // If the transaction already has a gasLimit, it should have been simulated in advance.
```
pxrl, line 166: There are some scenarios where this will probably lead to failed txns, simply due to lag in the relayer between when a txn is simulated/enqueued, and when it is submitted. I believe that this late-stage simulation does actually catch some of these issues.

The alternative is to just simulate it _again_ and ensure that the estimated gas consumption still matches.

src/clients/MultiCallerClient.ts
```diff
@@ -288,40 +291,49 @@ export class MultiCallerClient {
...
+
+      // Aggregate the individual gasLimits. If a transaction does not have a gasLimit defined then it has not been
+      // simulated. In this case, drop the aggregation and revert to undefined to force estimation on submission.
+      gasLimit = isDefined(gasLimit) && isDefined(txn.gasLimit)
+        ? gasLimit.add(txn.gasLimit)
+        : undefined;
     });
 
-    // There should not be any duplicate call data blobs within this array. If there are there is likely an error.
-    callData = [...new Set(callData)];
```
pxrl, line 312: Incidental/drive-by removal - if we had this I'd rather know about it, than suppress it by filtering for unique objects. It also quietly limits what the caller can do, and it would be confusing to debug on the off chance that the caller was actually intentionally trying to include duplicate txns.

src/clients/MultiCallerClient.ts
```diff
@@ -288,40 +291,49 @@ export class MultiCallerClient {
...
-    let callData = transactions.map((txn, idx) => {
       mrkdwn.push(`\n  *txn. ${idx + 1}:* ${txn.message ?? "No message"}: ${txn.mrkdwn ?? "No markdown"}`);
-      return txn.contract.interface.encodeFunctionData(txn.method, txn.args);
+      callData.push(txn.contract.interface.encodeFunctionData(txn.method, txn.args));
+
+      // Aggregate the individual gasLimits. If a transaction does not have a gasLimit defined then it has not been
+      // simulated. In this case, drop the aggregation and revert to undefined to force estimation on submission.
+      gasLimit = isDefined(gasLimit) && isDefined(txn.gasLimit)
+        ? gasLimit.add(txn.gasLimit)
+        : undefined;
```
pxrl, line 320: This all looks scarier than it is - there was previously 2 loops over `transactions`, and now there is only one loop that handles both checks. This loop now also aggregates the `gasLimit` if it is defined across all input transactions, otherwise it reverts the gasLimit to undefined so that the transaction will be estimated prior to submission.

src/clients/MultiCallerClient.ts
nicholaspai: That's fair

src/clients/MultiCallerClient.ts
```diff
@@ -288,40 +291,47 @@ export class MultiCallerClient {
...
+      gasLimit,
     });
 
     return {
       chainId,
       contract,
       method: "multicall",
-      args: [callData],
+      args: callData,
+      gasLimit,
```
nicholaspai, line 334: Was this the source of any issues? This is a rly subtle but nice catch

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/1062#discussion_r1385658674

This is a nice catch. I'm curious how you caught it and whether it resolved any longstanding mysterious bugs

pxrl:
This has some unforseen breakage - pulling back to draft.

src/clients/MultiCallerClient.ts
pxrl: tbh I'm not sure it actually was an issue, and changing it might break some things. Will have to circle back to that :eyes:

pxrl: Yeah...was actually a mistake to remove it. Have restored it and all the tests work OK again.

PR #1065, opened 2023-11-09
chore: Bump Across dependencies for Optimism Görli

pxrl:
This includes new deployments for Optimism-Goerli and wECO on test chains.


james-a-morris (approved):
Good call - I don't think there's any regressions per my integration testing.

src/common/Constants.ts
```diff
@@ -194,12 +200,16 @@ export const PROVIDER_CACHE_TTL_MODIFIER = 0.15;
 // Multicall3 Constants:
 export const multicall3Addresses = {
   1: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  5: "0xcA11bde05977b3631167028862bE2a173976CA11",
   10: "0xcA11bde05977b3631167028862bE2a173976CA11",
   137: "0xcA11bde05977b3631167028862bE2a173976CA11",
   288: "0xcA11bde05977b3631167028862bE2a173976CA11",
+  324: "0xcA11bde05977b3631167028862bE2a173976CA11",
```
mrice32, line 206: Is this right? zksync has a different address derivation process, so a contract should never have the same address on zksync, right?

https://era.zksync.io/docs/reference/troubleshooting/faq.html#can-someone-claim-the-address-i-have-for-my-contract-in-other-evm-networks-in-zksync-era

mrice32 (commented):
Looks great! One question

src/common/Constants.ts
pxrl: :skull: RIP

Great catch! You are of course correct. I checked on the multicall3 github that zkSync was supported, but didn't verify the address.

pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/pull/1065/commits/17319464a42cf2d44b820f3b9cb01a823d43bb12

Addresses pulled from multicall3 deployments.json: https://github.com/mds1/multicall/blob/main/deployments.json

PR #1066, opened 2023-11-09
feat(rebalancer): Add toggle to send rebalances.

nicholaspai:
Can be used to split up relayer functionality into two bots:
- fast bot that sends relays and doesn't rebalance
- slow bot that runs entire relayer logic but doesn't execute anything and does rebalance

### DIFF
--- a/src/relayer/RelayerConfig.ts
+++ b/src/relayer/RelayerConfig.ts
@@ -13,6 +13,7 @@ export class RelayerConfig extends CommonConfig {
   readonly skipRelays: boolean;
   readonly skipRebalancing: boolean;
   readonly sendingRelaysEnabled: boolean;
+  readonly sendingRebalancesEnabled: boolean;
   readonly sendingMessageRelaysEnabled: boolean;
   readonly sendingSlowRelaysEnabled: boolean;
   readonly sendingRefundRequestsEnabled: boolean;
@@ -47,6 +48,7 @@ export class RelayerConfig extends CommonConfig {
       RELAYER_INVENTORY_CONFIG,
       RELAYER_TOKENS,
       SEND_RELAYS,
+      SEND_REBALANCES,
       SEND_MESSAGE_RELAYS,
       SKIP_RELAYS,
       SKIP_REBALANCING,
@@ -138,6 +140,7 @@ export class RelayerConfig extends CommonConfig {
     this.debugProfitability = DEBUG_PROFITABILITY === "true";
     this.relayerGasMultiplier = toBNWei(RELAYER_GAS_MULTIPLIER || Constants.DEFAULT_RELAYER_GAS_MULTIPLIER);
     this.sendingRelaysEnabled = SEND_RELAYS === "true";
+    this.sendingRebalancesEnabled = SEND_REBALANCES === "true";
     this.sendingMessageRelaysEnabled = SEND_MESSAGE_RELAYS === "true";
     this.skipRelays = SKIP_RELAYS === "true";
     this.skipRebalancing = SKIP_REBALANCING === "true";
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -117,7 +117,7 @@ export async function constructRelayerClients(
     adapterManager,
     crossChainTransferClient,
     config.bundleRefundLookback,
-    !config.sendingRelaysEnabled
+    !config.sendingRebalancesEnabled
   );
 
   return { ...commonClients, spokePoolClients, ubaClient, tokenClient, profitClient, inventoryClient, acrossApiClient };


src/relayer/RelayerConfig.ts
```diff
@@ -49,6 +50,7 @@ export class RelayerConfig extends CommonConfig {
       RELAYER_INVENTORY_CONFIG,
       RELAYER_TOKENS,
       SEND_RELAYS,
+      SEND_REBALANCES,
```
james-a-morris, line 53: Thoughts on adding this to the sample ENV?

james-a-morris (commented):
LGTM - left a comment

james-a-morris:
> Can be used to split up relayer functionality into two bots:
> 
> * fast bot that sends relays and doesn't rebalance
> * slow bot that runs entire relayer logic but doesn't execute anything and does rebalance

OOC - was this spurred by anything specific? I think it will be effective as the number of relays to process + the number of chains to rebalance increases

nicholaspai:
> > Can be used to split up relayer functionality into two bots:
> > 
> > * fast bot that sends relays and doesn't rebalance
> > * slow bot that runs entire relayer logic but doesn't execute anything and does rebalance
> 
> OOC - was this spurred by anything specific? I think it will be effective as the number of relays to process + the number of chains to rebalance increases

Yup so this was spurred by the need for the `rebalancer` to run the relayer logic (specifically `checkForUnfilledDepositsAndFill`) to load all "unfilled deposits" and account for any that it doesn't have balance for in the `TokenClient`. See this [line](https://github.com/across-protocol/relayer-v2/blob/master/src/relayer/Relayer.ts#L286).

Without this, the inventory rebalancer won't rebalance for outstanding deposits it can't fill. Instead, it will only rebalance when accounting for current balance.

So, the goal of this PR is to enable two modes of the relayer to run in parallel:
- Relayer with `SEND_RELAYS=true SKIP_REBALANCING=true` that only sends relays and skips the inventory management
- Rebalancer with `SEND_RELAYS=false SEND_REBALANCING=true` that simulates the relayer logic including loading token shortfalls into memory and then actually executes the inventory management

src/relayer/RelayerConfig.ts
```diff
@@ -141,6 +143,7 @@ export class RelayerConfig extends CommonConfig {
     this.relayerGasPadding = toBNWei(RELAYER_GAS_PADDING || Constants.DEFAULT_RELAYER_GAS_PADDING);
     this.relayerGasMultiplier = toBNWei(RELAYER_GAS_MULTIPLIER || Constants.DEFAULT_RELAYER_GAS_MULTIPLIER);
     this.sendingRelaysEnabled = SEND_RELAYS === "true";
+    this.sendingRebalancesEnabled = SEND_REBALANCES === "true";
```
pxrl, line 146: I think we need to look at this a bit more, because we'd have `SEND_REBALANCES` and `SKIP_REBALANCES`, but they're not necessarily mutually exclusive, so it makes the configuration combinations difficult to follow.

nicholaspai: Yeah the idea is you can use `SKIP_REBALANCES` to decide whether to enter into the inventory manager logic and then `SEND_REBALANCES` to actually execute them. I agree this is not very clean but its meant to mirror the SEND_RELAYS and SKIP_RELAYS variable naming. Can we clean them up in a diff pr? or do you have ideas

james-a-morris (commented):
I think this is fine if you need to get this in. 

Although - we're adding a lot of flags related to enabling/disabling functionality. Would there be any buy-in in enumerating the relayer's functionality so we can collapse these flags?

Rough Example:
ENABLED="relayer:no-rebalance,slow-relay"
ENABLED="dataworker:no-finalizer,executor,disputor"
...

james-a-morris (approved):
Approving for the reasons above ^

nicholaspai:
> I think this is fine if you need to get this in. 
> 
> 
> 
> Although - we're adding a lot of flags related to enabling/disabling functionality. Would there be any buy-in in enumerating the relayer's functionality so we can collapse these flags?
> 
> 
> 
> Rough Example:
> 
> ENABLED="relayer:no-rebalance,slow-relay"
> 
> ENABLED="dataworker:no-finalizer,executor,disputor"
> 
> ...

For obvious reasons I think we want to collapse these variables but I don't necessarily agree with this example. Is there a better way to do it?

nicholaspai:
> > I think this is fine if you need to get this in. 
> > 
> > 
> > 
> > Although - we're adding a lot of flags related to enabling/disabling functionality. Would there be any buy-in in enumerating the relayer's functionality so we can collapse these flags?
> > 
> > 
> > 
> > Rough Example:
> > 
> > ENABLED="relayer:no-rebalance,slow-relay"
> > 
> > ENABLED="dataworker:no-finalizer,executor,disputor"
> > 
> > ...
> 
> For obvious reasons I think we want to collapse these variables but I don't necessarily agree with this example. Is there a better way to do it?

Really what we should do is create a new /inventory-manager folder with its own index.ts as the source of these confusing config names is that the inventory manager and relayer are coupled. 

But I still think this PR is worth merging so we can get the more refined tuning option now

pxrl (approved):
I think it's preferable to get this in now so that we restore proactive rebalancing, and then schedule some follow-up work to make some adjustments as part of the upcoming tech debt cycle.

src/relayer/RelayerConfig.ts
pxrl: OK for now; we can revisit it in the tech debt cycle :+1:

PR #1073, opened 2023-11-14
refactor(Dataworker): Replace ProfitClient with PriceClient

pxrl:
The Dataworker only relies indirectly on the ProfitClient for its ability to query token prices. This imposes some test/mock burden on the ProfitClient. This change hoists the ProfitClient out of the Dataworker and instead subs in the underlying PriceClient, which the ProfitClient also uses directly.

As a side benefit, the proposer is currently the only component dependent on token prices, so token price lookups can be relegated to the code that actually needs it.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -409,7 +409,7 @@ export class Dataworker {
       // Exit early if volume of pool rebalance leaves exceeds USD threshold. Volume includes netSendAmounts only since
       // that is the actual amount sent over bridges. This also mitigates the chance that a RelayerRefundLeaf is
       // published but its refund currency isn't sent over the bridge in a PoolRebalanceLeaf.
-      const totalUsdRefund = PoolRebalanceUtils.computePoolRebalanceUsdVolume(
+      const totalUsdRefund = await PoolRebalanceUtils.computePoolRebalanceUsdVolume(
         rootBundleData.poolRebalanceLeaves,
         this.clients
       );
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -1,4 +1,4 @@
-import { typechain } from "@across-protocol/sdk-v2";
+import { typechain, utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { ConfigStoreClient, HubPoolClient, SpokePoolClient } from "../clients";
 import { Clients } from "../common";
 import * as interfaces from "../interfaces";
@@ -29,6 +29,8 @@ import {
   winston,
 } from "../utils";
 import { DataworkerClients } from "./DataworkerClientHelper";
+
+const { bnZero, fixedPointAdjustment: fixedPoint } = sdkUtils;
 
 export function updateRunningBalance(
   runningBalances: interfaces.RunningBalances,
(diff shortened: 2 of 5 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -409,7 +409,7 @@ export class Dataworker {
       // Exit early if volume of pool rebalance leaves exceeds USD threshold. Volume includes netSendAmounts only since
       // that is the actual amount sent over bridges. This also mitigates the chance that a RelayerRefundLeaf is
       // published but its refund currency isn't sent over the bridge in a PoolRebalanceLeaf.
-      const totalUsdRefund = PoolRebalanceUtils.computePoolRebalanceUsdVolume(
+      const totalUsdRefund = await PoolRebalanceUtils.computePoolRebalanceUsdVolume(
```
nicholaspai, line 412: How fast does this call take to run?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -174,19 +176,39 @@ export function addSlowFillsToRunningBalances(
...
       const netSendAmount = poolRebalanceLeaf.netSendAmounts[index];
-      const volume = netSendAmount.abs();
-      const tokenPriceInUsd = clients.profitClient.getPriceOfToken(l1Token);
-      const volumeInUsd = volume.mul(tokenPriceInUsd).div(toBNWei(1));
-      const l1TokenInfo = clients.hubPoolClient.getTokenInfoForL1Token(l1Token);
-      const volumeInUsdScaled = volumeInUsd.mul(toBN(10).pow(18 - l1TokenInfo.decimals));
+      const volume = netSendAmount.abs().mul(bn10.pow(18 - decimals)); // Scale volume to 18 decimals.
 
-      return sum.add(volumeInUsdScaled);
+      const usdVolume = volume.mul(usdTokenPrice).div(fixedPoint);
```
nicholaspai, line 208: usdPrice is always in 18 decimal precision right?

nicholaspai (approved):
This makes sense, I left some questions. We're exchanging the time taken to update `profitClient.update()` upfront with the same time taken to load prices in the dataworker proposer flow later. This should result in speed ups for all bots except for the proposer when it wants to actually propose a bundle. This seems theoretically strictly better!

src/dataworker/Dataworker.ts
pxrl: Logs indicate that we normally update token prices within about 100 - 200 ms. I'd expect this change to impose about the same delay on `computePoolRebalanceUsdVolume()`.

src/dataworker/PoolRebalanceUtils.ts
pxrl: Correct - the PriceClient returns with 0 scaling, and the result is scaled up to 18 decimals irrespective of the token itself.

tokenPrices

pxrl: Just to do a quick sanity check, I ran the proposer locally with a forced bundle range that matches that of the latest proposal. The cumulative USD volumes were within about 500 USD of each other. This is to be expected given small token price fluctuations between when it was proposed and when I tested. The proposal itself had about 250k USDC in refunds to the RL relayer alone on mainnet, so if there was a decimals comparison issue then it would definitely have shown up in a significantly larger delta.

bundle-remote
bundle-local

PR #1074, opened 2023-11-14
refactor(ProfitClient): Support token lookup by symbol

pxrl:
This is a follow-up change to allow the ProfitClient to work primarily with token symbols, rather than their underlying addresses. Supporting price lookup by symbol eases the prod/testnet transition for the relayer.


nicholaspai (approved):
Good QoL improvement

PR #1076, opened 2023-11-14
fix(ProfitClient): Support non-unique token addresses

pxrl:
In cases like ETH/WETH, the underlying mainnet contract address used to resolve the token price is the same. Therefore, mapping from contract address to L1Token doesn't work very well for this purpose. Invert the token mapping so that the token symbol is unique, allowing for multiple symbols to map to a single address. This also helps in the case of test tokens that are temporarily mapped to an existing token.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -470,9 +470,7 @@ export class ProfitClient {
     this.enabledChainIds.forEach((chainId) => {
       const symbol = getNativeTokenSymbol(chainId);
       if (!isDefined(tokens[symbol])) {
-        const { addresses } = TOKEN_SYMBOLS_MAP[symbol];
-        const address = addresses[1];
-        tokens[symbol] = address;
+        tokens[symbol] = TOKEN_SYMBOLS_MAP[symbol].addresses[1];
       }
     });
 
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -456,40 +456,42 @@ export class ProfitClient {
   }
 
   protected async updateTokenPrices(): Promise<void> {
-    // Generate list of tokens to retrieve.
-    const l1Tokens: { [k: string]: L1Token } = Object.fromEntries(
+    // Generate list of tokens to retrieve. Map by symbol because tokens like
+    // ETH/WETH refer to the same mainnet contract address.
+    const tokens: { [_symbol: string]: string } = Object.fromEntries(
       this.hubPoolClient.getL1Tokens().map(({ symbol }) => {
-        const { decimals, addresses } = TOKEN_SYMBOLS_MAP[symbol];
+        const { addresses } = TOKEN_SYMBOLS_MAP[symbol];
         const address = addresses[1];
-        return [address, { symbol, decimals, address }];
+        return [symbol, address];
       })
     );
 
     // Also ensure all gas tokens are included in the lookup.
     this.enabledChainIds.forEach((chainId) => {
       const symbol = getNativeTokenSymbol(chainId);
-      const { decimals, addresses } = TOKEN_SYMBOLS_MAP[symbol];
+      const addresses = TOKEN_SYMBOLS_MAP[symbol];
       const address = addresses[1];
-      l1Tokens[address] ??= { symbol, decimals, address };
+      tokens[symbol] ??= address;
     });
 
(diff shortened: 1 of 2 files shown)


src/clients/ProfitClient.ts
```diff
@@ -456,40 +456,42 @@ export class ProfitClient {
...
     );
 
     // Also ensure all gas tokens are included in the lookup.
     this.enabledChainIds.forEach((chainId) => {
       const symbol = getNativeTokenSymbol(chainId);
-      const { decimals, addresses } = TOKEN_SYMBOLS_MAP[symbol];
+      const addresses = TOKEN_SYMBOLS_MAP[symbol];
       const address = addresses[1];
-      l1Tokens[address] ??= { symbol, decimals, address };
+      tokens[symbol] ??= address;
```
nicholaspai, line 474: So for ETH, this sets the address to which one?

pxrl: 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2. This is as defined in constants-v2: https://github.com/across-protocol/constants-v2/blob/e0327ae4d6312f6cfa01ca08a0077d8efbf00e13/src/constants.ts#L87

nicholaspai (approved):
Much cleaner now

src/clients/ProfitClient.ts
```diff
@@ -456,40 +456,42 @@ export class ProfitClient {
...
     // Pre-populate any new addresses.
-    Object.values(l1Tokens).forEach(({ symbol, address }) => {
+    Object.entries(tokens).forEach(([symbol, address]) => {
       this.tokenSymbolMap[symbol] ??= address;
       this.tokenPrices[address] ??= bnZero;
     });
 
     try {
-      const tokenPrices = await this.priceClient.getPricesByAddress(Object.keys(l1Tokens), "usd");
+      const tokenAddrs = Array.from(new Set(Object.values(tokens)));
```
james-a-morris, line 486: Thoughts: This is used widely in the code, and it may be beneficial to finally create a util function `remove duplicates` that does just this.

src/clients/ProfitClient.ts
```diff
@@ -456,40 +456,42 @@ export class ProfitClient {
...
 
     // Also ensure all gas tokens are included in the lookup.
     this.enabledChainIds.forEach((chainId) => {
       const symbol = getNativeTokenSymbol(chainId);
-      const { decimals, addresses } = TOKEN_SYMBOLS_MAP[symbol];
-      const address = addresses[1];
-      l1Tokens[address] ??= { symbol, decimals, address };
+      if (!isDefined(tokens[symbol])) {
+        tokens[symbol] = TOKEN_SYMBOLS_MAP[symbol].addresses[1];
+      }
```
james-a-morris, line 474:
```suggestion
      tokens[symbol] ??= TOKEN_SYMBOLS_MAP[symbol].addresses[1];
```

src/clients/ProfitClient.ts
```diff
@@ -456,40 +456,42 @@ export class ProfitClient {
...
-    const l1Tokens: { [k: string]: L1Token } = Object.fromEntries(
+    // Generate list of tokens to retrieve. Map by symbol because tokens like
+    // ETH/WETH refer to the same mainnet contract address.
+    const tokens: { [_symbol: string]: string } = Object.fromEntries(
       this.hubPoolClient.getL1Tokens().map(({ symbol }) => {
-        const { decimals, addresses } = TOKEN_SYMBOLS_MAP[symbol];
+        const { addresses } = TOKEN_SYMBOLS_MAP[symbol];
         const address = addresses[1];
-        return [address, { symbol, decimals, address }];
+        return [symbol, address];
```
james-a-morris, line 465: Out of curioisity - is there a reason we removed the `decimals` from this object?

james-a-morris: Hmm nevermind - the example of `WBTC` / `tBTC` would already break this assumption

pxrl: `decimals` was never actually used before, and `address` was duplicated as both the key and the value in the Object mapping. We don't support tBTC, but USDC is anyway 6 decimals and is heavily bridged, so it's another good sanity check.

src/clients/ProfitClient.ts
pxrl: Yeah, can pencil that in for the SDK too.

pxrl: https://github.com/across-protocol/sdk-v2/pull/462

PR #1077, opened 2023-11-14
improve: Update scripts w/ centralised chain defs

pxrl:
The motivation for this was adding support for Optimism Goerli.

### DIFF
--- a/scripts/hubpool.ts
+++ b/scripts/hubpool.ts
@@ -1,6 +1,6 @@
 import minimist from "minimist";
 import { WETH9__factory as WETH9 } from "@across-protocol/contracts-v2";
-import { constants as sdkConsts, utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { constants as sdkConsts } from "@across-protocol/sdk-v2";
 import { BigNumber, ethers, Wallet } from "ethers";
 import { config } from "dotenv";
 import { getNetworkName, getSigner } from "../src/utils";
@@ -180,9 +180,11 @@ async function search(args: Record<string, number | string>, _signer: Wallet): P
       configStore.globalConfig(CHAIN_ID_INDICES, { blockTag: blockNumber }),
     ]);
 
-    const chainIds = _chainIds.length > 0
-      ? JSON.parse(_chainIds.replaceAll('"', ""))
-      : PROTOCOL_DEFAULT_CHAIN_ID_INDICES; // Production chain IDs; will be incorrect on Görli.
+    // If the ConfigStore doesn't have CHAIN_ID_INDICES defined at the relevant block, sub in
+    // the implicit initial value. This is only applicable to production and will be incorrect on Görli.
+    // Görli will soon be deprecated, at which point it won't be relevant anyway.
+    const chainIds =
+      _chainIds.length > 0 ? JSON.parse(_chainIds.replaceAll('"', "")) : PROTOCOL_DEFAULT_CHAIN_ID_INDICES;
 
     const args = hubPool.interface.parseLog({ data, topics }).args;
     const eventArgs = Object.keys(args).filter((key) => isNaN(Number(key)));
--- a/scripts/hubpool.ts
+++ b/scripts/hubpool.ts
@@ -6,7 +6,7 @@ import { config } from "dotenv";
 import { getNetworkName, getSigner } from "../src/utils";
 import * as utils from "./utils";
 
-const { PRODUCTION_CHAIN_IDS, TESTNET_CHAIN_IDS } = sdkConsts;
+const { PROTOCOL_DEFAULT_CHAIN_ID_INDICES } = sdkConsts;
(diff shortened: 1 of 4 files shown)


pxrl:
fwiw this needs an SDK bump, which should occur soon.

scripts/hubpool.ts
```diff
@@ -171,14 +173,14 @@ async function search(args: Record<string, number | string>, _signer: Wallet): P
 
   const events = await hubPool.queryFilter(filter, fromBlock, toBlock);
   const CHAIN_ID_INDICES = ethers.utils.formatBytes32String("CHAIN_ID_INDICES");
+  const DEFAULT_CHAIN_IDS = sdkUtils.chainIsProd(chainId) ? PRODUCTION_CHAIN_IDS : TESTNET_CHAIN_IDS;
```
james-a-morris, line 176: Should we include an enabled-list for the chains? These are all available tokens and the user may want to only search on certain chains.

scripts/utils.ts
```diff
@@ -72,14 +71,10 @@ export function resolveToken(token: string, chainId: number): ERC20 {
...
-    const ok = knownChainIds.includes(chainId);
-    if (!ok) {
-      console.log(`Invalid chain ID: ${chainId}`);
-    }
-    return ok;
-  });
+  return (
+    chainIds.every((chainId) => sdkUtils.chainIsProd(chainId)) ||
+    chainIds.every((chainId) => sdkUtils.chainIsTestnet(chainId))
+  );
```
james-a-morris, line 77: This will make the list `O(2N)`. Could we get away with the following?
```
  return 
    chainIds.every((chainId) => sdkUtils.chainIsProd(chainId)) || sdkUtils.chainIsTestnet(chainId));
```

or even, if we wanted to be more efficient:
```
  return 
    !chainIds.some((chainId) => !(sdkUtils.chainIsProd(chainId)  || sdkUtils.chainIsTestnet(chainId)));
```

Although the above ^ would return false for the empty list.

james-a-morris (commented):
Left a comment about the scripts

scripts/utils.ts
```diff
@@ -10,13 +12,10 @@ export type ERC20 = {
...
 
-export const testChains = [5, 280, 80001, 421613];
-export const chains = [1, 10, 137, 324, 8453, 42161];
-
 // Public RPC endpoints to be used if preferred providers are not defined in the environment.
 const fallbackProviders: { [chainId: number]: string } = {
-  1: "https://eth.llamarpc.com",
-  5: "https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161",
+  [CHAIN_IDs.MAINNET]: "https://eth.llamarpc.com",
+  [CHAIN_IDs.GOERLI]: "https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161",
```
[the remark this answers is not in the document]
pxrl: It's not necessary, since we only have HUbPool deployments on these chains.

scripts/utils.ts
pxrl: I think the main argument against that is that it permits prod and test chainIds to be mixed in the same array, but that's not a valid combination.

But just to pull on this one a bit anyway, are you sure it's O(2N) ? If the user supplies a prod chainId then it should evaluate the first `Array.every()` call completely, and then return, without looking at the second `Array.every()` statement.

If the user supplies a tesnet chainId then the first `Array.every()` should fail on the first chainId (`chainIsProd(chainId) => false`), and then it should rotate to (`chainIsTestnet(chainId) => ...).

In any case, the array of chains is only single-digit elements in length and we don't call this code a lot in a tight loop, so even if it did require a lot of loops, I'd probably still prefer to be strict rather than fast.

scripts/hubpool.ts
pxrl: Just for context, this code is basically scraping events emitted by the HubPool, and the the applicable `CHAIN_ID_INDICES` is also fetched for the relevant block so it can be printed for context. So it's really only intended to be a factual representation of what happened, and that can be post-filtered. Does that make sense?

scripts/utils.ts
james-a-morris: Fair enough! And that's a good point about not wanting to mix the elements. 

Last question in this vein: would it make sense to make `isChainProd(...chainIds[])` so that we can pass one or more elements into it?

mrice32 (approved):
LGTM!

scripts/utils.ts
pxrl: We can definitely opt to do it, but I don't really see the benefit and am inclined not to do it (now) since it involves another sdk bump. The list of chain IDs that we supply to these functions is only single-digit length, so there's basically zero inefficiencies to be saved by doing this. At the same time, it could be annoying to have to supply an array when only one element is being evaluated.

PR #1078, opened 2023-11-14
fix: Sync constants imports

pxrl:
Export directly from src/utils/index.ts so that we don't end up with stale references to contracts-v2.

### DIFF
--- a/src/clients/bridges/op-stack/optimism/OptimismAdapter.ts
+++ b/src/clients/bridges/op-stack/optimism/OptimismAdapter.ts
@@ -1,9 +1,7 @@
-import { winston } from "../../../../utils";
+import { winston, TOKEN_SYMBOLS_MAP } from "../../../../utils";
 import { SpokePoolClient } from "../../..";
 import { BaseAdapter } from "../..";
-import { constants } from "@across-protocol/sdk-v2";
 import { OpStackAdapter } from "../OpStackAdapter";
-const { TOKEN_SYMBOLS_MAP } = constants;
 import { DaiOptimismBridge } from "./DaiOptimismBridge";
 import { SnxOptimismBridge } from "./SnxOptimismBridge";
 
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -1,4 +1,4 @@
-import { TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2";
+import { TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
 import { BigNumber, ethers } from ".";
 
 export function compareAddresses(addressA: string, addressB: string): 1 | -1 | 0 {
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -31,6 +31,8 @@ export { config } from "dotenv";
 export { replaceAddressCase } from "@uma/common";
 export { Logger } from "@uma/financial-templates-lib";
 
+export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
+
 // TypeChain exports used in the bot.
 export {
   getContractInfoFromAddress,
--- a/test/AdapterManager.SendTokensCrossChain.ts
+++ b/test/AdapterManager.SendTokensCrossChain.ts
@@ -2,7 +2,7 @@ import * as zksync from "zksync-web3";
 import { SpokePoolClient } from "../src/clients";
 import { AdapterManager } from "../src/clients/bridges"; // Tested
 import { CONTRACT_ADDRESSES } from "../src/common";
-import { bnToHex, getL2TokenAddresses, toBNWei } from "../src/utils";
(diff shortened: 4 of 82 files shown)


src/utils/AddressUtils.ts
pxrl, line 1: For files inside `src/utils` I have retained importing directoy from `constants-v2` because we don't want circular dependencies by importing from `src/utils/index.ts`. I tried to import and re-export from `src/utils/AddressUtils.ts` but ended up in tsc purgatory and ran out of patience. If anyone has any tips I'm happy to update it.

test/AdapterManager.SendTokensCrossChain.ts
pxrl, line 1: This file contains a few related driveby fixes that I idenfitied whilst here.

src/utils/AddressUtils.ts
nicholaspai: sounds good

src/clients/bridges/op-stack/optimism/OptimismAdapter.ts
```diff
@@ -1,9 +1,7 @@
-import { winston } from "../../../../utils";
+import { winston, TOKEN_SYMBOLS_MAP } from "../../../../utils";
 import { SpokePoolClient } from "../../..";
```
james-a-morris, line 2: Completely out of scope, but this relative import is nuts. It could easily lead to circular dependencies because we're importing from an ancestor `index.ts` file.

src/utils/index.ts
```diff
@@ -31,6 +31,8 @@ export { config } from "dotenv";
 export { replaceAddressCase } from "@uma/common";
 export { Logger } from "@uma/financial-templates-lib";
 
+export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
```
james-a-morris, line 34: Thoughts on placing this high on the import chain in a separate file called `constants.ts`?

Reasoning: we already have this schema in `./utils/SDKUtils.ts` that exports. It would be consistent to do the same for the constants-v2 repo.

test/AdapterManager.SendTokensCrossChain.ts
```diff
@@ -48,12 +47,12 @@ let l1BaseBridge: FakeContract;
...
-  weth: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
-  dai: "0x6B175474E89094C44Da98b954EedeAC495271d0F",
-  wbtc: "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
-  snx: "0xC011a73ee8576Fb46F5E1c5751cA3B9Fe0af2a6F",
-};
+  usdc: TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET],
+  weth: TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET],
+  dai: TOKEN_SYMBOLS_MAP.DAI.addresses[CHAIN_IDs.MAINNET],
+  wbtc: TOKEN_SYMBOLS_MAP.WBTC.addresses[CHAIN_IDs.MAINNET],
+  snx: TOKEN_SYMBOLS_MAP.SNX.addresses[CHAIN_IDs.MAINNET],
```
james-a-morris, line 54: Nit: I'm not a huge fan of the reuse of `TOKEN_SYMBOLS_MAP[{SYMBOL}].addresses[CHAIN_IDs.MAINNET]` but nothing actionable.

james-a-morris (commented):
This looks like a 1:1 refactoring. Correct me if I'm wrong, but it doesn't look like any business logic has been changed?

Only question: is there a reason we're not branched against master?

src/clients/bridges/op-stack/optimism/OptimismAdapter.ts
james-a-morris: This may be worthy of adding to our tech debt cycle @nicholaspai

test/AdapterManager.SendTokensCrossChain.ts
pxrl: Yeah. I also think it could be done more efficiently, but didn't want to spend the time on it because it'd involved changing even more of this code. It works OK, but I thought it was preferable to avoid replicating official addresses, which is why I subbed this in.

pxrl:
> This looks like a 1:1 refactoring. Correct me if I'm wrong, but it doesn't look like any business logic has been changed?

Correct - 1:1, no change in business logic.

> Only question: is there a reason we're not branched against master?

It depends on another PR that I have open, but I didn't want to pollute that with this plumbing.

src/utils/AddressUtils.ts
mrice32: +1

mrice32 (approved):
LGTM!

src/clients/bridges/op-stack/optimism/OptimismAdapter.ts
pxrl: I added a Linear ticket for it :+1:

src/utils/index.ts
pxrl: I added a Linear ticket for it, bundled with the point above :+1:

PR #1082, opened 2023-11-17
fix(zksync-finalizer): Identify correct log index of ZkSync withdrawal when there are multiple withdrawals in same L2 transaction

nicholaspai:
The current code doesn't handle the case where there are multiple `TokensBridged` events within a single ZkSync transaction. This is because it always returns the `withdrawalIdx = 0` for all batched `TokensBridged`.

I believe this is because the `.reverse()` always causes the subsequent `.find()` to return index 0.

This PR simplifies the logic a lot and simply:
- Find all `L1MessageSentEvents` preceding the `TokensBridged` event.
- Return `withdrawalIdx` equal to the # of above preceding `L1MessageSentEvents`. This should return a unique withdrawal index per TokensBridged

### DIFF
--- a/src/finalizer/utils/zkSync.ts
+++ b/src/finalizer/utils/zkSync.ts
@@ -125,14 +125,19 @@ async function filterMessageLogs(
     const { transactionHash, logIndex } = tokenBridged;
     const txnReceipt = txnReceipts[transactionHash];
 
-    // Search backwards from the TokensBridged log index for the corresponding L1MessageSent event.
+    // Search backwards from the TokensBridged log index for all L1MessageSent events. The
+    // last L1MessageSent event is the one that corresponds to the withdrawal.
     // @dev Array.findLast() would be an improvement but tsc doesn't currently allow it.
-    const txnLogs = txnReceipt.logs.slice(0, logIndex).reverse();
-    const withdrawal = txnLogs.find((log) => log.topics[0] === l1MessageSent);
+    const l1MessagesSentPrecedingWithdrawal = txnReceipt.logs.filter(
+      (log) => log.topics[0] === l1MessageSent && log.logIndex <= logIndex
+    );
+    if (l1MessagesSentPrecedingWithdrawal.length === 0) {
+      throw new Error("Found a ZkSync TokensBridged event without any preceding L1MessageSent events");
+    }
 
     // @dev withdrawalIdx is the "withdrawal number" within the transaction, _not_ the index of the log.
-    const l1MessagesSent = txnReceipt.logs.filter((log) => log.topics[0] === l1MessageSent);
-    const withdrawalIdx = l1MessagesSent.indexOf(withdrawal);
+    // The withdrawal index for this withdrawal is therefore the number of L1MessageSent events that precede it.
+    const withdrawalIdx = l1MessagesSentPrecedingWithdrawal.length - 1;
     return { ...tokenBridged, withdrawalIdx };
   });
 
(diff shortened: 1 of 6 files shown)


src/finalizer/utils/zkSync.ts
```diff
@@ -125,14 +125,19 @@ async function filterMessageLogs(
...
-    // Search backwards from the TokensBridged log index for the corresponding L1MessageSent event.
+    // Search backwards from the TokensBridged log index for all L1MessageSent events. The
+    // last L1MessageSent event is the one that corresponds to the withdrawal.
     // @dev Array.findLast() would be an improvement but tsc doesn't currently allow it.
-    const txnLogs = txnReceipt.logs.slice(0, logIndex).reverse();
-    const withdrawal = txnLogs.find((log) => log.topics[0] === l1MessageSent);
+    const l1MessagesSentPrecedingWithdrawal = txnReceipt.logs.filter(
+      (log) => log.topics[0] === l1MessageSent && log.logIndex <= logIndex
+    );
+    if (l1MessagesSentPrecedingWithdrawal.length === 0) {
```
james-a-morris, line 134: Has this happened often during your tests? We may want to log here

src/finalizer/utils/zkSync.ts
```diff
@@ -125,14 +125,19 @@ async function filterMessageLogs(
     const { transactionHash, logIndex } = tokenBridged;
     const txnReceipt = txnReceipts[transactionHash];
 
-    // Search backwards from the TokensBridged log index for the corresponding L1MessageSent event.
+    // Search backwards from the TokensBridged log index for all L1MessageSent events. The
+    // last L1MessageSent event is the one that corresponds to the withdrawal.
     // @dev Array.findLast() would be an improvement but tsc doesn't currently allow it.
```
james-a-morris, line 130: Do we still need this if we're not reversing

src/finalizer/utils/zkSync.ts
```diff
@@ -125,14 +125,19 @@ async function filterMessageLogs(
...
-    const txnLogs = txnReceipt.logs.slice(0, logIndex).reverse();
-    const withdrawal = txnLogs.find((log) => log.topics[0] === l1MessageSent);
+    const l1MessagesSentPrecedingWithdrawal = txnReceipt.logs.filter(
```
james-a-morris, line 131: Could we get away with using `utils.findIndex()` to avoid iterating through the entire array and spawning a new one?

james-a-morris (commented):
Left a few comments

src/finalizer/utils/zkSync.ts
nicholaspai: Good point removing

mrice32 (commented):
Looks great! Only minor comments

src/utils/EventUtils.ts
```diff
@@ -219,6 +219,25 @@ export function isEventOlder<T extends SortableEvent>(ex: T, ey: T): boolean {
...
+ * Returns an array with the same length as the passed in Event array where each index is assigned a new index
+ * that states its relative position to other events with the same transaction hash. If two or more of the input
+ * events have the same transaction hash, they will be assigned unique indices starting at 0 and counting up based
+ * on the order of events passed in.
+ * @param events
+ */
+export function getUniqueLogIndex(events: { transactionHash: string }[]): number[] {
+  const uniqueTokenhashes = {};
+  const logIndexesForMessage = [];
+  for (const event of events) {
```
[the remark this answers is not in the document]
nicholaspai: I think its up to the caller to sort them. Whatever order they are passed in will be the order that the withdrawal indices are assigned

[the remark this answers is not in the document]
james-a-morris: > I think its up to the caller to sort them. Whatever order they are passed in will be the order that the withdrawal indices are assigned

We may want to add that in a doc comment ^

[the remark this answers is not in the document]
nicholaspai: its already there

PR #1084, opened 2023-11-20
improve(relayer): Don't zero-fill when relayer expects to rebalance to fast fill deposit

nicholaspai:
I think this can address the failure case where we submit a zero fill and send an inventory rebalance to an L2 in the same Relayer run when we're short balance there.

This PR introduces a new logical branch to the relayer when it sees a deposit that it can't fast full on the destination chain:
- First capture the token shortfall in the TokenClient. This is important so that the inventory client can accurately determine if it will need to rebalance.
- Load all possible rebalances: this is based on the relayer balances per chain, the outstanding cross chain transfers, and any shortfalls
- See if there is a rebalance that would fast fill this deposit.
- If there is, then check if the new destination chain balance plus this rebalance would be enough for the relayer to fast fill the deposit
- If so, then we shouldn't zero fill and we should let the relayer fast fill. **This way the relayer gets to choose its repayment chain after sending a full fill, as opposed to being forced to take repayment on the destination for a partial fill.**

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -22,7 +22,7 @@ import { CONTRACT_ADDRESSES } from "../common";
 
 type TokenDistributionPerL1Token = { [l1Token: string]: { [chainId: number]: BigNumber } };
 
-type Rebalance = {
+export type Rebalance = {
   chainId: number;
   l1Token: string;
   thresholdPct: BigNumber;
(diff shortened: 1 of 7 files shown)


src/clients/InventoryClient.ts
```diff
@@ -243,20 +253,55 @@ export class InventoryClient {
...
-      thresholdPct: BigNumber;
-      targetPct: BigNumber;
-      currentAllocPct: BigNumber;
-      balance: BigNumber;
-      cumulativeBalance: BigNumber;
-      amount: BigNumber;
-    };
     type ExecutedRebalance = Rebalance & { hash: string };
 
     const rebalancesRequired: Rebalance[] = [];
```
pxrl, line 307: I think this needs to be removed.

src/clients/InventoryClient.ts
```diff
@@ -270,41 +315,7 @@ export class InventoryClient {
       const tokenDistributionPerL1Token = this.getTokenDistributionPerL1Token();
```
pxrl, line 315: Tangential to this change, but I don't really see why it should be wrapped in a `try...catch` clause. Should we just remove that?

nicholaspai: yeah agreed that's a bad pattern to have such a large try-catch

nicholaspai: However, removing try-catch leads to a much larger diff so maybe we hold off

mrice32 (commented):
LGTM!

PR #1085, opened 2023-11-21
feat: Sync latest blocks across clients

nicholaspai:
# Motivation

We see the following situation which leads to inefficient use of relayer capital
- Relayer runs every 2 mins. Its likely that relayer runs overlap and multiple instances are running simultaneously.
- There is a pending deposit on chain X that is very large and requires the relayer to send additional funds to X to fill it.
- Relayer chooses not to zero fill the deposit because it will have enough funds post-rebalance to fast fill the deposit. Run 1 sends a inventory rebalance txn to send tokens to chain X.
- Relayer 2's [`SpokePoolClient.update`](https://github.com/across-protocol/relayer-v2/blob/master/src/relayer/RelayerClientHelper.ts#L140) starts before run 1's inventory rebalance txn is sent so it doesn't see the inventory rebalance event. This update can take a long time. Relayer 2 then updates the TokenClient, which does see that its balance has decreased on mainnet due to the inventory rebalance. Relayer 2 then sees the deposit but doesn't see the cross-chain rebalance AND doesn't think it has enough balance anymore to move its inventory to X, so it sends the zero fill.
- The zero fill then forces the relayer on a subsequent run to fast fill the deposit with its inventory rebalance and take repayment on the destination chain X as opposed to mainnet.

# Implementation
Syncs the `latest` block queried for the following clients:
- SpokePoolClient
- CrossChainTransferClient
- HubPoolClient
- ConfigStoreClient

This way the events queried are queried up to the same end block per chain.

Additional queries to `ERC20.balanceOf` and `provider.getBalance` to query token and ETH balances are still made with an implied `latest` block tag in the following cases:
- checking ETH balance when deciding whether to unwrap ETH in the `inventoryClient.unwrapWeth()` function
- Checking that ERC20 balances haven't changed from `TokenClient.getBalance` before executing rebalances in `inventoryClient.rebalanceInventory`
- Dataworker checking ETH balance before executing pool rebalance and relayer refund leaves

Fixes #ACX-1716

### DIFF
--- a/src/clients/TokenClient.ts
+++ b/src/clients/TokenClient.ts
@@ -42,13 +42,6 @@ export class TokenClient {
     return this.tokenData[chainId][token].balance;
   }
 
-  getAllowanceOnChain(chainId: number, token: string): BigNumber {
-    if (!this._hasTokenPairData(chainId, token)) {
-      return toBN(0);
-    }
-    return this.tokenData[chainId][token].allowance;
-  }
-
   decrementLocalBalance(chainId: number, token: string, amount: BigNumber): void {
     this.tokenData[chainId][token].balance = this.tokenData[chainId][token].balance.sub(amount);
   }
--- a/src/clients/TokenClient.ts
+++ b/src/clients/TokenClient.ts
@@ -219,11 +219,14 @@ export class TokenClient {
       .getAllOriginTokens()
       .map((address) => new Contract(address, ERC20.abi, spokePoolClient.spokePool.signer));
 
+    const blockTag = spokePoolClient.eventSearchConfig.toBlock ?? "latest";
     const tokenData = Object.fromEntries(
       await Promise.all(
         tokens.map(async (token) => {
-          const balance: BigNumber = await token.balanceOf(this.relayerAddress);
-          const allowance: BigNumber = await token.allowance(this.relayerAddress, spokePoolClient.spokePool.address);
+          const balance: BigNumber = await token.balanceOf(this.relayerAddress, { blockTag });
+          const allowance: BigNumber = await token.allowance(this.relayerAddress, spokePoolClient.spokePool.address, {
+            blockTag,
+          });
 
           return [token.address, { balance, allowance }];
         })
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -86,7 +86,6 @@ export abstract class BaseAdapter {
 
   // Note: this must be called after the SpokePoolClients are updated.
   getUpdatedSearchConfigs(): { l1SearchConfig: EventSearchConfig; l2SearchConfig: EventSearchConfig } {
-    // Update search range based on the latest data from corresponding SpokePoolClients' search ranges.
     const l1LatestBlock = this.spokePoolClients[this.hubChainId].latestBlockNumber;
     const l2LatestBlock = this.spokePoolClients[this.chainId].latestBlockNumber;
     if (l1LatestBlock === 0 || l2LatestBlock === 0) {
@@ -96,17 +95,11 @@ export abstract class BaseAdapter {
     return {
       l1SearchConfig: {
         ...this.baseL1SearchConfig,
-        fromBlock: this.baseL1SearchConfig.toBlock
-          ? this.baseL1SearchConfig.toBlock + 1
-          : this.baseL1SearchConfig.fromBlock,
-        toBlock: l1LatestBlock,
+        toBlock: this.baseL1SearchConfig?.toBlock ?? l1LatestBlock,
       },
       l2SearchConfig: {
         ...this.baseL2SearchConfig,
-        fromBlock: this.baseL2SearchConfig.toBlock
-          ? this.baseL2SearchConfig.toBlock + 1
-          : this.baseL2SearchConfig.fromBlock,
-        toBlock: l2LatestBlock,
+        toBlock: this.baseL2SearchConfig?.toBlock ?? l2LatestBlock,
       },
     };
   }
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -142,14 +142,6 @@ export async function constructSpokePoolClientsWithStartBlocks(
     );
   }
 
-  logger.debug({
-    at: "ClientHelper#constructSpokePoolClientsWithStartBlocks",
-    message: "Enabled chains in block range",
-    startBlocks,
-    toBlockOverride,
-    enabledChains,
-  });
-
   // Set up Spoke signers and connect them to spoke pool contract objects:
   const spokePoolSigners = await getSpokePoolSigners(baseSigner, enabledChains);
   const spokePools = await Promise.all(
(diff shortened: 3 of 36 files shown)


src/clients/TokenClient.ts
```diff
@@ -42,13 +42,6 @@ export class TokenClient {
     return this.tokenData[chainId][token].balance;
   }
 
-  getAllowanceOnChain(chainId: number, token: string): BigNumber {
```
nicholaspai, line 45: Opportunistically removing this function that is not used anywhere. This limits the error surface involved when reading `allowance` from this client that is possible out of date

src/relayer/RelayerClientHelper.ts
```diff
@@ -127,8 +127,6 @@ export async function constructRelayerClients(
 export async function updateRelayerClients(clients: RelayerClients, config: RelayerConfig): Promise<void> {
   // SpokePoolClient client requires up to date HubPoolClient and ConfigStore client.
   const { configStoreClient, spokePoolClients, ubaClient } = clients;
-
-  await configStoreClient.update();
```
pxrl, line 131: Is this already handled somewhere else?

src/clients/TokenClient.ts
```diff
@@ -219,11 +212,14 @@ export class TokenClient {
       .getAllOriginTokens()
       .map((address) => new Contract(address, ERC20.abi, spokePoolClient.spokePool.signer));
 
+    const blockTag = spokePoolClient.eventSearchConfig.toBlock ?? "latest";
```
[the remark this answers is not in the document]
pxrl: My bad - I missed that you'd addressed this in the PR description.

src/common/ClientHelper.ts
```diff
@@ -142,14 +142,6 @@ export async function constructSpokePoolClientsWithStartBlocks(
     );
   }
 
-  logger.debug({
-    at: "ClientHelper#constructSpokePoolClientsWithStartBlocks",
-    message: "Enabled chains in block range",
-    startBlocks,
-    toBlockOverride,
-    enabledChains,
-  });
```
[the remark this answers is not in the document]
nicholaspai: We can retain it I thought it was just noisy

src/common/ClientHelper.ts
```diff
@@ -165,7 +157,24 @@ export async function constructSpokePoolClientsWithStartBlocks(
     })
   );
 
-  return getSpokePoolClientsForContract(logger, hubPoolClient, config, spokePools, startBlocks, toBlockOverride);
```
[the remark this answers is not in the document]
nicholaspai: I think we might want to remove the `toBlockOverride` as we never use it right?

src/relayer/RelayerClientHelper.ts
nicholaspai: yeah in  `ClientHelper.updateClients` which gets called by `RelayerClientHelper.constructRelayerClients`

src/common/ClientHelper.ts
```diff
@@ -181,8 +202,15 @@ export function getSpokePoolClientsForContract(
   config: CommonConfig,
   spokePools: { chainId: number; contract: Contract; registrationBlock: number }[],
   fromBlocks: { [chainId: number]: number },
-  toBlocks: { [chainId: number]: number } = {}
+  toBlocks: { [chainId: number]: number }
```
nicholaspai, line 205: This is a safety feature: force caller to supply toBlocks to make this function easier to reason about

src/common/ClientHelper.ts
```diff
@@ -222,11 +256,13 @@ export async function constructClients(
...
 ): Promise<Clients> {
-  const hubSigner = baseSigner.connect(await getProvider(config.hubPoolChainId, logger));
+  const hubPoolProvider = await getProvider(config.hubPoolChainId, logger);
+  const hubSigner = baseSigner.connect(hubPoolProvider);
+  const latestMainnetBlock = await hubPoolProvider.getBlockNumber();
 
   const rateModelClientSearchSettings = {
     fromBlock: Number(getDeploymentBlockNumber("AcrossConfigStore", config.hubPoolChainId)),
-    toBlock: undefined,
+    toBlock: config.toBlockOverride[config.hubPoolChainId] ?? latestMainnetBlock,
```
nicholaspai, line 265: This was arguably a bug. HubPoolClient/ConfigStoreClient should also be allowing caller to override to blocks the same way that the SpokePoolClients can be overridden

src/common/ClientHelper.ts
```diff
@@ -192,9 +220,15 @@ export function getSpokePoolClientsForContract(
         registrationBlock,
       });
     }
+    if (!isDefined(toBlocks[chainId])) {
```
mrice32, line 223: +1!

src/common/ClientHelper.ts
mrice32: +1 -- this PR removes a lot of unhappy paths that make everything harder to analyze. These small tweaks help the type enforce more guarantees.

src/common/ClientHelper.ts
```diff
@@ -165,7 +165,28 @@ export async function constructSpokePoolClientsWithStartBlocks(
     })
   );
 
-  return getSpokePoolClientsForContract(logger, hubPoolClient, config, spokePools, startBlocks, toBlockOverride);
+  // Explicitly set toBlocks for all chains so we can re-use them in other clients to make sure they all query
+  // state to the same "latest" block per chain.
+  const latestBlocksForChain: Record<number, number> = Object.fromEntries(
```
mrice32, line 170: This could be a dumb question, but why do this here instead of doing it at a higher level where we can pass the results _both_ to the spoke clients, the hub client, and the other clients?

It just seems odd that we're reading from the hubPoolClient to inform the values, when it seems like this function (if performed earlier) could _supply_ the info to the HubPoolClient constructor instead of reading from it.

src/clients/bridges/BaseAdapter.ts
```diff
@@ -96,17 +95,11 @@ export abstract class BaseAdapter {
     return {
       l1SearchConfig: {
         ...this.baseL1SearchConfig,
-        fromBlock: this.baseL1SearchConfig.toBlock
-          ? this.baseL1SearchConfig.toBlock + 1
-          : this.baseL1SearchConfig.fromBlock,
-        toBlock: l1LatestBlock,
```
mrice32, line 102: Why was this removed? I don't see where this functionality is replicated elsewhere. Was it no longer needed?

src/clients/TokenClient.ts
mrice32: +1 removing dead code is always good!

mrice32 (commented):
Looks great! I think this contains a lot of good tweaks, reducing the number of paths the code can flow through. A few minor questions!

src/clients/bridges/BaseAdapter.ts
nicholaspai: This function is supposed to return a `from/toBlock` for L1 and L2. Currently, it sets `fromBlock` = `this.baseL1SearchConfig.fromBlock` because `this.baseL1SearchConfig.toBlock` is pretty much always `undefined`. This is a weird conditional  that I wanted to simplify so I believe its equivalent to just return the `baseL1SearchConfig`.

In total, I think I'm duplicating the logic but making it more readable and also making sure that the `toBlock` returned is equal to the eventSearchConfig.toBlock of the spoke pool clients

src/common/ClientHelper.ts
nicholaspai: The main reason is that we need to first figure out which chains we're going to create `spokeSigners` for, and then load latest blocks for all chains. I think we can refactor this though pretty easily I'll take a crack at it

nicholaspai: Ah yes, ok so this is an optimization. In order to figure out which chains we need to create spoke pool signers for, to subsequently query `latest` blocks, we dynamically try to load the enabled chains from the config store client. This requires the config store client to be updated so the pattern in all the bots is to call `constructClients` to create the config store client (amongst other clients), update it, then create spoke pool clients.

We could change this, but it would require keeping a hard coded listing somewhere of all chains we enable rather than being able to load them dynamically from the config store client. I think we actually used to maintain a hardcoded listing (i.e. a `CHAIN_ID_INDICES` default list) but we found this tedious to maintain so we removed it.

src/common/ClientHelper.ts
```diff
@@ -239,10 +275,8 @@ export async function constructClients(
   );
 
   const hubPoolClientSearchSettings = {
+    ...rateModelClientSearchSettings,
     fromBlock: Number(getDeploymentBlockNumber("HubPool", config.hubPoolChainId)),
-    toBlock: undefined, // Important that we set this to `undefined` to always look up latest HubPool events such as
-    // ProposeRootBundle in order to match a bundle block evaluation block range with a pending root bundle.
```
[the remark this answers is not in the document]
pxrl: Should we not set `toBlock: latestMainnetBlock` instead ?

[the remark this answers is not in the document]
nicholaspai: We do, we actually inherit it from `... rateModelClientSearchSettings`

PR #1088, opened 2023-11-27
improve(InventoryClient): Safer dereferencing & use bnZero

pxrl:
The relayer now calls into the InventoryClient in new ways, and this can backfire if the relayer doesn't have an inventory config defined.

While here, sub out instances of toBN(0) in favour of the predefined bnZero. This is more memory efficient and saves creating disposable BNs.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1,4 +1,5 @@
 import { TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import {
   BigNumber,
   winston,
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

src/clients/InventoryClient.ts
```diff
@@ -20,6 +21,8 @@ import { Deposit, FillsToRefund, InventoryConfig } from "../interfaces";
 import lodash from "lodash";
 import { CONTRACT_ADDRESSES } from "../common";
 
+const { bnZero } = sdkUtils;
```
james-a-morris, line 24: We have a `utils/SDKUtils.ts` file for exporting from the sdk-v2 repo. Thoughts on moving this there?

james-a-morris (approved):
Looks great. Left one nit that I'd like to see before we merge

mrice32 (approved):
LGTM!

src/clients/InventoryClient.ts
pxrl: I'll follow up with a separate PR for this, since we use it in many places.

PR #1089, opened 2023-11-27
fix: Supply cache instance to getCachedBlockForTimestamp()

pxrl:
Without this we don't cache the timestamp -> blocknumber mappings, so we rely on the underlying RPC caching to retain the responses to the eth_getBlock queries that we issue. The are two reasons this is less desirable:
- Cached RPC queries are typically ejected much more regularly (default 1 hour, vs 1 week for cached block timestamps).
- Timestamp to block number resolution usually takes a different search path on each query, so there is a high chance of RPC cache misses even when trying to resolve the same timestamp.

This is one of the underlying reasons for the elevated RPC consumption of late.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -64,11 +64,13 @@ export async function constructSpokePoolClientsWithLookback(
   }
 
   const hubPoolChainId = hubPoolClient.chainId;
-  const currentTime = getCurrentTime();
+  const lookback = getCurrentTime() - initialLookBackOverride;
 
   // Use the first block that we'll query on mainnet to figure out which chains were enabled between then
   // and the the latest mainnet block. These chains were enabled via the ConfigStore.
-  const fromBlock_1 = await getBlockForTimestamp(hubPoolChainId, currentTime - initialLookBackOverride);
+  const blockFinder = undefined;
+  const redis = await getRedisCache(logger);
+  const fromBlock_1 = await getBlockForTimestamp(hubPoolChainId, lookback, blockFinder, redis);
   const enabledChains = getEnabledChainsInBlockRange(configStoreClient, config.spokePoolChainsOverride, fromBlock_1);
 
   // Get full list of fromBlocks now for chains that are enabled. This way we don't send RPC requests to
@@ -79,7 +81,7 @@ export async function constructSpokePoolClientsWithLookback(
         if (chainId === 1) {
           return [chainId, fromBlock_1];
         } else {
-          return [chainId, await getBlockForTimestamp(chainId, currentTime - initialLookBackOverride)];
+          return [chainId, await getBlockForTimestamp(chainId, lookback, blockFinder, redis)];
         }
       })
     )
@@ -150,6 +152,9 @@ export async function constructSpokePoolClientsWithStartBlocks(
     enabledChains,
   });
 
+  const blockFinder = undefined;
+  const redis = await getRedisCache(logger);
+
   // Set up Spoke signers and connect them to spoke pool contract objects:
   const spokePoolSigners = await getSpokePoolSigners(baseSigner, enabledChains);
   const spokePools = await Promise.all(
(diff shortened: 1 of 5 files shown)


src/common/ClientHelper.ts
```diff
@@ -64,11 +64,13 @@ export async function constructSpokePoolClientsWithLookback(
   }
 
   const hubPoolChainId = hubPoolClient.chainId;
-  const currentTime = getCurrentTime();
+  const lookback = getCurrentTime() - initialLookBackOverride;
 
   // Use the first block that we'll query on mainnet to figure out which chains were enabled between then
   // and the the latest mainnet block. These chains were enabled via the ConfigStore.
-  const fromBlock_1 = await getBlockForTimestamp(hubPoolChainId, currentTime - initialLookBackOverride);
+  const blockFinder = undefined;
```
pxrl, line 71: Just for reference, this `blockFinder = undefined` pattern will be cleaned up in a follow-up commit, but that needs some rearranging of furniture in the SDK.

src/common/ClientHelper.ts
```diff
@@ -64,11 +64,13 @@ export async function constructSpokePoolClientsWithLookback(
...
+  const redis = await getRedisCache(logger);
+  const fromBlock_1 = await getBlockForTimestamp(hubPoolChainId, lookback, blockFinder, redis);
```
nicholaspai, line 73: How does this `getBlockForTimestamp` call work? It seems that the third arg should be `_blockFinder?` and there is no fourth for redis. Is there an SDK package bump missing in this PR?

src/utils/BlockUtils.ts
```diff
@@ -29,8 +32,10 @@ export async function getBlockFinder(chainId: number): Promise<BlockFinder> {
 export async function getBlockForTimestamp(
   chainId: number,
   timestamp: number,
-  _blockFinder?: BlockFinder
+  blockFinder?: BlockFinder,
+  redisCache?: interfaces.CachingMechanismInterface
 ): Promise<number> {
-  const blockFinder = _blockFinder ?? (await getBlockFinder(chainId));
-  return utils.getCachedBlockForTimestamp(chainId, timestamp, blockFinder);
+  blockFinder ??= await getBlockFinder(chainId);
```
nicholaspai, line 38: Do you think `blockFinder` and `redisCache` are better off setting at the global level of this file, i.e. outside of any function? Or should we, as your comment suggests, just remove this function entirely and leave the calling code responsible for creating a cached instance of `blockFinder` and `redisCache`?

nicholaspai (approved):
This looks good logically but the setting of `blockFinder` doesn't seem very intuitive to me, are we better off removing the function from `BlockUtils` in favor of calling the SDK directly?

src/utils/BlockUtils.ts
james-a-morris: @nicholaspai not sure if it sheds light, but we technically have the ability to set multiple `redisCache` clients in the relayer. This was existing logic that had to be dealt with in previous PRs. Although we only use one in practice, it may lead to further bugs down the road to assume only one exists.

james-a-morris: Then again, we already call it via `getBlockFinder` so is there a reason we'd want to pass a specific `blockFinder` rather than rely on the call to `getBlockFinder` here? AFAIR `getBlockFinder` maintains its own internal cache so we shouldn't expend any more resources than a record lookup if we'd previously called `getBlockFinder(chainId)` before

src/utils/BlockUtils.ts
```diff
@@ -29,8 +32,10 @@ export async function getBlockFinder(chainId: number): Promise<BlockFinder> {
...
+  redisCache ??= await getRedisCache();
```
james-a-morris, line 39: Similar question here: is there any specific case where we'd be willing to fallback to our `getRedisCache()` rather than pass an instance? We may be able to get away with not passing the `redisCache` client in the input.

james-a-morris (commented):
Couple of comments. Otherwise LGTM

src/common/ClientHelper.ts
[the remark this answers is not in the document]
pxrl: `blockFinder` is explicitly initialised to `undefined`, so `getBlockForTimestamp()` subs in its own value.

The reason I'm specifically naming it here is because I have a pet peeve against passing `undefined` in directly on a positional argument - it's prone to blowing up if the ordering of input parameters is changed.

But yeah - I think we should circle back to this and clean it up - I don't see any reason why the `BlockFInder` instance should be supplied by the caller (and we never use it), so I'm in favour of just removing it.

src/utils/BlockUtils.ts
pxrl: The only reason I'm passing it in via the caller is that `getRedisCache()` accepts a logger instance, and that's not available in `getBlockForTimestamp()`. We can change that; I just want this in ASAP now because it's causing inflated RPC consumption.

PR #1092, opened 2023-11-28
feat(Relayer): Refactor filtering out deposits to fill and add blacklisted-depositor rule

nicholaspai:


src/relayer/Relayer.ts
```diff
@@ -118,7 +118,85 @@ export class Relayer {
     // Flush any pre-existing enqueued transactions that might not have been executed.
     this.clients.multiCallerClient.clearTransactionQueue();
 
-    const unfilledDeposits = await this.getUnfilledDeposits();
+    // Fetch unfilled deposits and filter out deposits upfront before we compute the minimum deposit confirmation
+    // per chain, which is based on the deposit volume we could fill.
+    const unfilledDeposits = (await this.getUnfilledDeposits()).filter(({ deposit, unfilledAmount, invalidFills }) => {
```
nicholaspai, line 123: In this section, several filters are copied from below and added here. The new blacklisted-depositor check is added, and no other rules are added/removed

pxrl: Just wanted to check - are you aware of the existing filtering that occurs within `this.getUnfilledDeposits()`? Up there we already filter on:
- Route enabled yes/no
- Deposit version supported yes/no
- Origin/destination chain supported yes/no

wdyt about collecting all the filtering there?

src/relayer/RelayerConfig.ts
```diff
@@ -75,6 +78,9 @@ export class RelayerConfig extends CommonConfig {
     this.slowDepositors = SLOW_DEPOSITORS
       ? JSON.parse(SLOW_DEPOSITORS).map((depositor) => ethers.utils.getAddress(depositor))
       : [];
+    this.blacklistedDepositors = BLACKLISTED_DEPOSITORS
+      ? JSON.parse(BLACKLISTED_DEPOSITORS).map((depositor) => ethers.utils.getAddress(depositor))
+      : [];
```
pxrl, line 83:
```suggestion
    this.blacklistedDepositors = JSON.parse(BLACKLISTED_DEPOSITORS ?? "[]")
      .map((depositor) => ethers.utils.getAddress(depositor));
```

pxrl (approved):

src/relayer/Relayer.ts
nicholaspai: Good point, refactored a3ec993f6dd54b1aa66feb0a6640b4b0dafb3c53

src/relayer/Relayer.ts
```diff
@@ -39,54 +38,120 @@ export class Relayer {
...
+        });
+        return false;
       }
 
-      return this.routeEnabled(originChainId, destinationChainId) ? "supportedDeposits" : "ignoredDeposits";
-    });
+      // Skip deposits with quoteTimestamp in the future (impossible to know HubPool utilization => LP fee cannot be computed).
+      if (quoteTimestamp + quoteTimeBuffer > hubPoolClient.currentTime) {
+        return false;
+      }
```
pxrl, line 69: wdyt about elimintating the `quoteTimeBuffer` here? It's intended to protect against re-orgs, but it's ignorant of per-chain finality guarantees. There is also some overlap with the MDCs configuration, and I think the MDCs config is a more granular way of expressing re-org risk tolerance.

src/relayer/Relayer.ts
```diff
@@ -39,54 +38,120 @@ export class Relayer {
...
-        at: "Relayer::getUnfilledDeposits",
-        message: "Skipping deposits that are not supported by this relayer version.",
-        latestVersionSupported: maxVersion,
-        latestInConfigStore: configStoreClient.getConfigStoreVersionForTimestamp(),
-        deposits,
-      });
-    }
+      // Skip any L1 tokens that are not specified in the config.
+      // If relayerTokens is an empty list, we'll assume that all tokens are supported.
+      const l1Token = hubPoolClient.getL1TokenInfoForL2Token(originToken, originChainId);
```
pxrl, line 73: Since this involves resolving a token mapping, wdyt about deferring it to one of the later stages? i.e. if we order the filtering by least expensive to most expensive we should minimise the overall work that the relayer needs to do per deposit.

src/relayer/Relayer.ts
```diff
@@ -39,54 +38,120 @@ export class Relayer {
...
+      if (
+        relayerTokens.length > 0 &&
+        !relayerTokens.includes(l1Token.address) &&
+        !relayerTokens.includes(l1Token.address.toLowerCase())
```
pxrl, line 77: `relayerTokens` should already be stored in checksum case so we can probably skip the `toLowerCase()` clause.

relayerTokens

src/relayer/Relayer.ts
```diff
@@ -112,13 +177,14 @@ export class Relayer {
   async checkForUnfilledDepositsAndFill(sendSlowRelays = true): Promise<void> {
     // Fetch all unfilled deposits, order by total earnable fee.
     const { config } = this;
-    const { acrossApiClient, hubPoolClient, profitClient, spokePoolClients, tokenClient, inventoryClient } =
-      this.clients;
+    const { hubPoolClient, profitClient, spokePoolClients, tokenClient, inventoryClient } = this.clients;
 
     // Flush any pre-existing enqueued transactions that might not have been executed.
     this.clients.multiCallerClient.clearTransactionQueue();
```
pxrl, line 183: Opportunistic reflow - I wanted to do this previously but it pushed line 180 over 120 chars and forced it to become 7 lines. Now we can keep it all on 1 line.

```suggestion
    const { hubPoolClient, multiCallerClient, profitClient, spokePoolClients, tokenClient, inventoryClient } = this.clients;

    // Flush any pre-existing enqueued transactions that might not have been executed.
    multiCallerClient.clearTransactionQueue();
```

pxrl (commented):
This is looking really good and cleans up the code a lot. I think we can fine-tune the  filtering a bit so that the earliest filters do the majority of the work, but otherwise this looks great and really simplifies the logic.

src/relayer/Relayer.ts
nicholaspai: hmm I think that's ok

nicholaspai: ee8c524c9a4c6db6239f53554a5a587600801294

src/relayer/Relayer.ts
nicholaspai: cff4601a0c737f9b6201c405a4876d420b6e0e00

src/relayer/Relayer.ts
nicholaspai: cff4601a0c737f9b6201c405a4876d420b6e0e00

src/relayer/Relayer.ts
```diff
@@ -39,54 +38,120 @@ export class Relayer {
...
+          at: "Relayer::getUnfilledDeposits",
+          message: "Skipping deposit that is not supported by this relayer version.",
+          latestVersionSupported: maxVersion,
+          latestInConfigStore: configStoreClient.getConfigStoreVersionForTimestamp(),
+          deposit,
+        });
+        return false;
+      }
+
+      if (!this.routeEnabled(originChainId, destinationChainId)) {
```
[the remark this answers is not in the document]
nicholaspai: cff4601a0c737f9b6201c405a4876d420b6e0e00

pxrl (approved):
LGTM!

PR #1095, opened 2023-11-30
improve(finalizer): Use common OP stack detector

pxrl:


nicholaspai (approved):

src/finalizer/utils/opStack.ts
```diff
@@ -1,6 +1,7 @@
 import assert from "assert";
 import { groupBy } from "lodash";
 import * as optimismSDK from "@eth-optimism/sdk";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
```
james-a-morris, line 4: This is fine, but we have this file that we should be importing from: `src/utils/SDKUtils.ts`

Since it exists, we should have future imports from the SDK routed through this.

pxrl: Thanks for the nudge - I keep forgetting to do that. Fixed: https://github.com/across-protocol/relayer-v2/pull/1095/commits/5de914f1c8c895ff6fe42cc023acd95fa3331e4e

PR #1097, opened 2023-12-01
fix(relayer): Reorder deposit filtering

pxrl:
It's important to filter on supported tokens before trying to evaluate whether the fill amount is within the maximum limit.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -55,6 +55,18 @@ export class Relayer {
       const { quoteTimestamp, depositId, depositor, originChainId, destinationChainId, originToken, amount } = deposit;
       const destinationChain = getNetworkName(destinationChainId);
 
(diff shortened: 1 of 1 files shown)


pxrl:
The motivation for this change is that an SNX deposit occurred and the RL relayer was trying to resolve a limit for it without having realised that SNX is not defined in `RELAYER_TOKENS`. Per the implementation of `AcrossApiClient.getLimit()`, the token didn't exist in the limits and so an exception was thrown. This effectively prevented the relayer from processing fills for a short period of time, until the deposit was filled by the Synthetix relayer.

I have also wondered whether we should just adjust `AcrossApiClient.getLimit()` to return 0 instead of throwing in the event that an unknown token is queried, but haven't (yet) concluded that we should go that way because it anyway makes sense to filter on token before limit.

mrice32 (approved):

PR #1099, opened 2023-12-03
refactor(relayer): Don't require ubaClient before UBA

pxrl:
The ubaClient currently imposes a requirement that the relayer has a SpokePoolClient instance for each supported chain, even if the relayer has no particular interest in that chain. This prevents relay operators from opting out of specific chains and requires them to have an RPC provider defined for all supported chains. This is undesirable for scenarios like the SNX relayer, where the only supported chains are mainnet and Optimism.

This change defers ubaClient instantiation until the relayer actually indicates support for the UBA via a configuration change, and places appropriate guards around the use of the ubaClient to ensure it is defined.


nicholaspai (approved):
This is a good QoL update. LGTM I agree with the algorithm

PR #1102, opened 2023-12-05
fix: fix the eth_getLogs field ignore

mrice32:
This fixes an issue where we weren't actually dropping the field from the array elements of the eth_getlogs response. We were operating on the array, itself, which doesn't contain the fields we are trying to remove.


pxrl (approved):

nicholaspai (approved):
So is there a bug with the existing compareObjects function? I'm not sure I understand this fix

mrice32:
@nicholaspai yeah, only in the eth_getLogs case. eth_getLogs returns an array. We are trying to ignore "transactionLogIndex" in the array itself rather than the individual logs within the array. We need to go into each object within the response array to ignore the field.

PR #1105, opened 2023-12-08
refactor: Reduce use of disposable BigNumber instances

pxrl:
ethers provides predefined BNs for 0 and 1, and we import these via the Across SDK. Roll these out as far and wide as possible, so as to reduce the unnecessary instantiation of new short-lived BNs. This will admittedly only have a small impact on memory and CPU consumption, but is still worth doing (especially in cases where the BNs were previously instantiated in a tight loop).


nicholaspai (approved):

james-a-morris (approved):
LGTM. 

I'm not sure we'll see any real benefit to this. Instantiating these variables is incredibly lightweight, especially parsing the one character "0" or "1"

PR #1106, opened 2023-12-09
feat: Support instantiating void signers

pxrl:
tldr there are various scripts that don't need a real signer, so don't require a private key or mnemonic to be loaded in these cases. This makes life easier for people who just want to pull down the repo and run a script (i.e. validateRootBundle.ts).

The ethers Wallet type inherits from Signer, but it imposes additional requirements like having a private key. By deferring to the Signer type, we also permit void signers. This is nice for scenarios where a private key is not needed, which is in fact the vast majority of the usage in the bot.

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -300,8 +300,13 @@ async function run(argv: string[]): Promise<boolean> {
   const args = minimist(argv.slice(1), opts);
 
   config();
-  const keyType = ["deposit", "fill"].includes(argv[0]) ? args.wallet : "void";
-  const signer = await getSigner({ keyType, cleanEnv: true });
+  let signer: Signer;
+  try {
+    const keyType = ["deposit", "fill"].includes(argv[0]) ? args.wallet : "void";
+    const signer = await getSigner({ keyType, cleanEnv: true });
+  } catch (err) {
+    usage(args.wallet); // no return
+  }
 
   switch (argv[0]) {
     case "deposit":
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -300,20 +300,17 @@ async function run(argv: string[]): Promise<boolean> {
   const args = minimist(argv.slice(1), opts);
 
   config();
+  const keyType = ["deposit", "fill"].includes(argv[0]) ? args.wallet : "void";
+  const signer = await getSigner({ keyType, cleanEnv: true });
 
-  let signer: Signer;
   switch (argv[0]) {
     case "deposit":
-      signer = await getSigner({ keyType: args.wallet, cleanEnv: true });
       return await deposit(args, signer);
     case "dump":
-      signer = await getSigner({ keyType: "void", cleanEnv: true });
       return await dumpConfig(args, signer);
     case "fetch":
-      signer = await getSigner({ keyType: "void", cleanEnv: true });
       return await fetchTxn(args, signer);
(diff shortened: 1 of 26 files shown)


scripts/spokepool.ts
```diff
@@ -301,21 +301,19 @@ async function run(argv: string[]): Promise<boolean> {
...
-    usage(args.wallet); // no return
-  }
-
+  let signer: Signer;
   switch (argv[0]) {
     case "deposit":
+      signer = await getSigner({ keyType: args.wallet, cleanEnv: true });
       return await deposit(args, signer);
     case "dump":
+      signer = await getSigner({ keyType: "void", cleanEnv: true });
```
nicholaspai, line 310: Can we refactor this bit?

nicholaspai (approved):

scripts/spokepool.ts
pxrl: Yeah this part is messy in this change; I have a cleanup in mind for that.

scripts/spokepool.ts
```diff
@@ -301,21 +301,19 @@ async function run(argv: string[]): Promise<boolean> {
...
-  try {
-    signer = await getSigner({ keyType: args.wallet, cleanEnv: true });
-  } catch (err) {
-    usage(args.wallet); // no return
-  }
-
+  let signer: Signer;
   switch (argv[0]) {
     case "deposit":
+      signer = await getSigner({ keyType: args.wallet, cleanEnv: true });
```
james-a-morris, line 307: Couldn't we create an object of: 

```
const map = {
    deposit: { fn: deposit, keyType: "wallet" },
    dump: { fn: dumpConfig, keyType: "void" },
    ...
}
```

Then do:
```
    const result = map[argv[0]];
    if(result) {
          return result.fn(args, await getSigner({ keyType: result.keyType, cleanEnv: true}));
    } else {
         usage();
    }
```

scripts/zkSyncDemo.ts
```diff
@@ -92,7 +92,7 @@ export async function run(): Promise<void> {
       token,
       toBN(args.amount),
       recipient,
-      baseSigner.address,
+      await baseSigner.getAddress(),
```
james-a-morris, line 95: Could we reuse this above since it looks like `connectedSigner` & `baseSigner` are derived from the same address.

james-a-morris:
@pxrl GitHub seems to be down atm so I can't approve but LGTM. I left minor nits, but nothing critical that would need to be handled prior to merging.

Great work on this refactor.

scripts/zkSyncDemo.ts
pxrl: Good catch - implemented: https://github.com/across-protocol/relayer-v2/pull/1106/commits/a8ff336ba838cd33e500083b97d329ce7a0f463f

scripts/spokepool.ts
pxrl: Simplified here: https://github.com/across-protocol/relayer-v2/pull/1106/commits/9243dc172ad40fc6fb17442bb4ec9a0776bb18c7

scripts/spokepool.ts
pxrl: Simplified here: https://github.com/across-protocol/relayer-v2/commit/9243dc172ad40fc6fb17442bb4ec9a0776bb18c7

PR #1108, opened 2023-12-12
fix(Dataworker): Set SpokePool end blocks more conservatively

pxrl:
This change ensures that Dataworker proposal block range identification is set realistically relative to the event search config for each SpokePoolClient.

HubPoolClient and SpokePoolClient instantiation was recently updated such that their search configs are now synchronised, and determined close to startup. Previously, the SpokePoolClient's search config would be determined much later because it had to wait for the initial HubPoolClient update to complete, and that can span several mainnet blocks. This meant that there was typically a 0 delta between the `toBlock` search config for each SpokePoolClient and its recorded `latestBlockNumber`. After the recent update, these values are now divergent.

This divergence can result in a situation where each SpokePoolClient may not have sufficient data to correctly match each fill with a deposit, and this has resulted in several disputes over the past few days.

Fixes ACX-1755

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -24,7 +24,6 @@ import {
   formatFeePct,
   getFillDataForSlowFillFromPreviousRootBundle,
   getRefund,
-  isDefined,
   shortenHexString,
   shortenHexStrings,
   toBN,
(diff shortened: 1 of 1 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -475,8 +476,14 @@ export function getWidestPossibleExpectedBlockRange(
...
   // filled during the challenge period.
   const latestPossibleBundleEndBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
-    (chainId: number, index) =>
-      spokeClients[chainId] && Math.max(spokeClients[chainId].latestBlockNumber - endBlockBuffers[index], 0)
+    (chainId: number, index) => {
+      const spokePoolClient = spokeClients[chainId];
+      if (!isDefined(spokePoolClient)) {
+        return undefined;
+      }
+      const toBlock = spokePoolClient.eventSearchConfig.toBlock ?? spokePoolClient.latestBlockNumber;
```
pxrl, line 484: For transparency and context, the fallback to `spokePoolClient.latestBlockNumber` is "needed" for compatibility with existing tests. In production we set `toBlock` explicitly and so we would never fall back to `latestBlockNumber`. In test `toBlock` is not set explicitly and there are repeated calls to `SpokePoolClient.update()`, which actually breaks when `toBlock` is set because in that case it is never incremented.

I think we should revisit the tests to ensure that the SpokePoolClient instantiation is consistent with what is done in production, but that seems to require a larger refactoring of the tests.

pxrl: I was able to simplify this a lot by reverting to `SpokePoolClient.latestBlockSearched`, so this is a non-issue now.

nicholaspai:
Thoughts on using this PR? https://github.com/across-protocol/sdk-v2/pull/471

nicholaspai (approved):

PR #1109, opened 2023-12-12
chore(ProviderUtils): Drop timeout 60s -> 10s

pxrl:
The 60s timeout is a source of latent notification storms during provider outages. This often results in a high number of long-running bot instances.

Improve notification time and recovery time by dropping the timeouts.


james-a-morris (approved):
Should we push this number further down? I'd imagine with the sub-second response times of most providers, we could get away with 5 seconds

pxrl:
> Should we push this number further down? I'd imagine with the sub-second response times of most providers, we could get away with 5 seconds

I'd prefer to do it incrementally based on the feedback from having this in prod - 60s -> 10s is already a pretty huge jump.

PR #1112, opened 2023-12-13
fix: Use disputer address in validateRootBundle script

pxrl:
The validateRootBundle script inherits a lot of core functionality, including the TokenClient. The TokenClient automatically seeks to put the necessary token approvals in place for handling the ABT token, but that's not strictly required for the validateRootBundle script since it never actually makes any transactions. Work around this by specifying the proposer address when instantiating the VoidSigner instance.

### DIFF
--- a/src/scripts/validateRootBundle.ts
+++ b/src/scripts/validateRootBundle.ts
@@ -202,7 +202,10 @@ export async function validate(_logger: winston.Logger, baseSigner: Signer): Pro
 
 export async function run(_logger: winston.Logger): Promise<void> {
   try {
-    const baseSigner = await getSigner({ keyType: "void", cleanEnv: true });
+    // This script inherits the TokenClient, and it attempts to update token approvals.
+    // The disputer bot already has the necessary token approvals in place, so use its address.
+    const roAddress = "0xf7bac63fc7ceacf0589f25454ecf5c2ce904997c";
+    const baseSigner = await getSigner({ keyType: "void", cleanEnv: true, roAddress });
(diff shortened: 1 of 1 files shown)


src/scripts/validateRootBundle.ts
```diff
@@ -202,7 +202,10 @@ export async function validate(_logger: winston.Logger, baseSigner: Signer): Pro
 
 export async function run(_logger: winston.Logger): Promise<void> {
   try {
-    const baseSigner = await getSigner({ keyType: "void", cleanEnv: true });
+    // This script inherits the TokenClient, and it attempts to update token approvals.
+    // The disputer bot already has the necessary token approvals in place, so use its address.
+    const roAddress = "0xf7bAc63fc7CEaCf0589F25454Ecf5C2ce904997c";
```
nicholaspai, line 207: roAddress?

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/1112/files#r1424732096

wat does this variable name mean

src/scripts/validateRootBundle.ts
pxrl: ro -> read-only. We could also call it "voidSigner" or something. Naming things is hard.

pxrl: Ranmed to `voidSigner` here: https://github.com/across-protocol/relayer-v2/pull/1112/commits/e2f8f29c27090b0e456d094c9b63fccfb802c7a2

PR #1113, opened 2023-12-13
fix: Increase RPC timeout (...again)

pxrl:
Am still seeing spurious timeouts on the monitor.


pxrl:
Alex proposed the same PR instead.

PR #1115, opened 2023-12-13
fix: Increase RPC timeout

amateima:


src/utils/ProviderUtils.ts
```diff
@@ -70,7 +70,7 @@ class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
   }
 }
 
-const defaultTimeout = 25 * 1000;
+const defaultTimeout = 40 * 1000;
```
pxrl, line 73: Reverting to 60 because being alerted by the monitor bot is ++annoying.
```suggestion
const defaultTimeout = 60 * 1000;
```

pxrl (approved):

PR #1116, opened 2023-12-13
fix(ClientHelper): Sync SpokePoolClient search configs

pxrl:
Using `latestBlockNumber` leads to drift between mainnet and the respective spoke chains due to the time delay implied by the HubPoolClient update.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -177,6 +177,10 @@ export async function constructSpokePoolClientsWithStartBlocks(
 
   // Explicitly set toBlocks for all chains so we can re-use them in other clients to make sure they all query
   // state to the same "latest" block per chain.
+  if (hubPoolClient.eventSearchConfig.toBlock === undefined) {
+    throw new Error("HubPoolClient eventSearchConfig.toBlock is undefined");
+  }
+  const hubPoolBlock = await hubPoolClient.hubPool.provider.getBlock(hubPoolClient.eventSearchConfig.toBlock);
   const latestBlocksForChain: Record<number, number> = Object.fromEntries(
     await Promise.all(
       enabledChains.map(async (chainId) => {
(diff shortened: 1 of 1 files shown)


src/common/ClientHelper.ts
```diff
@@ -185,12 +189,10 @@ export async function constructSpokePoolClientsWithStartBlocks(
...
         }
         if (chainId === hubPoolClient.chainId) {
-          if (hubPoolClient.eventSearchConfig.toBlock === undefined) {
-            throw new Error("HubPoolClient eventSearchConfig.toBlock is undefined");
-          }
-          return [chainId, hubPoolClient.eventSearchConfig.toBlock];
+          return [chainId, hubPoolBlock.number];
         } else {
-          return [chainId, await spokePoolSigners[chainId].provider.getBlockNumber()];
+          const toBlock = await getBlockForTimestamp(chainId, hubPoolBlock.timestamp, blockFinder, redis);
```
pxrl, line 194: fwiw there are other ways we could address this issue, but this change is very targeted and doesn't involve changing/refactoring any surrounding code. Because the block timestamp will be very recent the BlockFinder will resolve it very quickly.

mrice32: What happens if the spoke chain is _behind_ the hub chain in terms of timestamp? I assume we'll just end up with latest, right (the current behavior)?

mrice32 (approved):
LGTM! Just one question: https://github.com/across-protocol/relayer-v2/pull/1116#discussion_r1425397639

nicholaspai (approved):
What are the consequences where hub pool client is behind the spoke pool client?

src/common/ClientHelper.ts
pxrl: Yeah - the blockFinder resolves future timestamps to the head of the chain, so it would resolve to latest. This is fine in case of a halted chain, but not so great if a chain is genuinely lagging.

pxrl:
> What are the consequences where hub pool client is behind the spoke pool client?

That's a deadlock scenario, since the hub and spoke architecture implies that the spokes can never be finalised ahead of the hub. For example, since quoteTimestamps on spokes are resolved to a block timestamp on the hub, it's impossible to resolve LP fees when the hub is lagging.

PR #1119, opened 2023-12-16
feat: :sparkles: search for deposit id by chain

james-a-morris:
We can now search an arbitrary deposit id on a given chain to find all of its information. This has been very helpful to resolve deposits when looking at fills.

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -308,6 +308,13 @@ async function fetchDepositId(args: Record<string, number | string>): Promise<bo
 
   const deposit = await spokeClient.findDeposit(depositId, destinationChain, depositor);
 
+  if (!deposit) {
+    console.log(
+      `No deposit found for depositId ${depositId} on chain ${originChainId}. Recommended to increase lookback time`
+    );
+    return false;
+  }
+
   console.log(
     [
       `Deposit #${depositId} on ${getNetworkName(originChainId)}:`,
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -273,7 +273,7 @@ async function fetchDepositId(args: Record<string, number | string>): Promise<bo
   const depositor = String(args.depositor);
   const hours = Number(args.hours) || 24;
 
-  if (!utils.validateChainIds([originChainId]) || Number.isNaN(depositId) || depositId < 0) {
+  if (!utils.validateChainIds([originChainId, destinationChainId]) || !Number.isInteger(depositId) || depositId < 0) {
     usage(); // no return
   }
   const provider = new ethers.providers.StaticJsonRpcProvider(utils.getProviderUrl(originChainId));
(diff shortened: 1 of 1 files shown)


scripts/spokepool.ts
```diff
@@ -255,6 +255,75 @@ async function fetchTxn(args: Record<string, number | string>, _signer: Signer):
   return true;
 }
 
+async function fetchDepositId(args: Record<string, number | string>): Promise<boolean> {
+  const chainId = Number(args.chainId);
+  const depositId = Number(args.depositId);
+
+  if (!utils.validateChainIds([chainId]) || Number.isNaN(depositId) || depositId < 0) {
```
nicholaspai, line 262: Should we replace isNaN with isInteger? or something like that

scripts/spokepool.ts
```diff
@@ -255,6 +266,70 @@ async function fetchTxn(args: Record<string, number | string>, _signer: Signer):
...
+      [originChainId]: latestBlockOnChain.number,
+    },
+    [originChainId]
+  );
+
+  const spokeClient = spokePoolClients[originChainId];
+
+  await spokeClient.update();
+
+  const deposit = await spokeClient.findDeposit(depositId, destinationChain, depositor);
```
nicholaspai, line 309: nit: if undefined, should we throw a detailed error recommending caller to set longer `hours`?

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/1119/files#r1428917153

scripts/spokepool.ts
james-a-morris: Good call - done

scripts/spokepool.ts
james-a-morris: Good call - done

pxrl:
This is a really useful feature that we've needed for a long time. I'm not a big fan of introducing the kitchen sink HubPoolClient and SpokePoolClient dependencies into this script because it's intended to be small and lightweight, and adding them starts to bog it down. I think there's a cleaner way of achieving this without adding them by just relying on the filtering functions of the SpokePool contract instance. I also think it'd be better placed as part of the existing `fetch` command - i.e. `fetch --chainId [--txnHash | --depositId]`.

pxrl:
> This is a really useful feature that we've needed for a long time. I'm not a big fan of introducing the kitchen sink HubPoolClient and SpokePoolClient dependencies into this script because it's intended to be small and lightweight, and adding them starts to bog it down. I think there's a cleaner way of achieving this without adding them by just relying on the filtering functions of the SpokePool contract instance. I also think it'd be better placed as part of the existing `fetch` command - i.e. `fetch --chainId [--txnHash | --depositId]`.

fwiw I took this PR and hacked on it a bit to show an example: https://github.com/across-protocol/relayer-v2/pull/1122

I think the above is cleaner in a few ways, primarily because it doesn't lump this script with having to know about the HubPool and SpokePoolClient types.

One counter-argument might be that it can limit the set of providers that can be used because the block range for the filter query is so large. It works fine with Infura and Arbitrum though, and I'm probably OK with that given that it keeps the code lighter.

mrice32 (commented):
@james-a-morris @pxrl can we close one of the PRs in favor of the other?

For reference, this is the other one: https://github.com/across-protocol/relayer-v2/pull/1122

PR #1122, opened 2023-12-17
feat: search for deposit id by chain

pxrl:
We can now search an arbitrary deposit id on a given chain to find all of its information. This has been very helpful to resolve deposits when looking at fills.

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -6,8 +6,19 @@ import { Contract, ethers, Signer } from "ethers";
 import { LogDescription } from "@ethersproject/abi";
 import { constants as sdkConsts, utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { ExpandedERC20__factory as ERC20 } from "@across-protocol/contracts-v2";
-import { BigNumber, formatFeePct, getNetworkName, getSigner, isDefined, resolveTokenSymbols, toBN } from "../src/utils";
+import {
+  BigNumber,
+  formatFeePct,
+  getDeploymentBlockNumber,
+  getNetworkName,
+  getSigner,
+  isDefined,
+  resolveTokenSymbols,
+  toBN,
+} from "../src/utils";
 import * as utils from "./utils";
+
+type Log = ethers.providers.Log;
 
 type relayerFeeQuery = {
   originChainId: number;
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

scripts/spokepool.ts
```diff
@@ -244,6 +258,29 @@ async function fetchTxn(args: Record<string, number | string>, _signer: Signer):
...
+  return { deposits, fills };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+async function fetchTxn(args: Record<string, number | string>, _signer: Signer): Promise<boolean> {
+  const { txnHash } = args;
+  const chainId = Number(args.chainId);
+
+  if (!utils.validateChainIds([chainId])) {
+    usage(); // no return
```
mrice32, line 270: nit: rather than annotating these unreachable paths/returns, I think from a ts perspective, returning the value is clearer, and making the usage function return type `never`. This can be done by adding `return process.exit(...);` in that function.

`never` just tells ts that the path where never is generated never happens (because this return will never actually happen).

```ts
...
  if (!utils.validateChainIds([chainId])) {
    return usage();
  }
...
// Right now, this function is typed as returning boolean, but I don't think booleans are ever returned.
function usage(badInput?: string): never {
  ...
  return process.exit(badInput === undefined ? 0 : 9);
}
```

scripts/spokepool.ts
```diff
@@ -215,23 +226,26 @@ async function dumpConfig(args: Record<string, number | string>, _signer: Signer
...
+  if (isNaN(depositId)) {
+    throw new Error("No depositId specified");
   }
 
+  const { chainId } = await spokePool.provider.getNetwork();
+  const deploymentBlockNumber = getDeploymentBlockNumber("SpokePool", chainId);
+  const latestBlockNumber = await spokePool.provider.getBlockNumber();
+  console.log(`Searching for depositId ${depositId} between ${deploymentBlockNumber} and ${latestBlockNumber}.`);
+  const filter = spokePool.filters.FundsDeposited(null, null, null, null, depositId);
+  return await spokePool.queryFilter(filter, deploymentBlockNumber, latestBlockNumber);
```
mrice32, line 240: Wouldn't this not work on a lot of chains because the block range (and the number of outputs) would usually be huge? Or am I missing something?

mrice32 (commented):
Looks great! One minor type nit and one question

scripts/spokepool.ts
pxrl: It works fine on the major providers, but it does have issues on free or low-tier providers.

The proposed solution is to update the sdk's function that narrows the block range for a deposit ID. At the moment it's dependent on a SpokePoolClient instance, but we deliberately don't instantiate that in this script to avoid the dependency and update overheads associated with it. The PR is here: https://github.com/across-protocol/sdk-v2/pull/476. It's currently snagged on a test case that relies on mocked SpokePoolClient behaviour, but that will be fixed.

So for now I'll just get this in as-is because it adds functionality that we currently don't have, and will acknowledge in a comment that we need to add the block narrowing feature from the SDK.

scripts/spokepool.ts
pxrl: Good call; the node.js docs actually give pretty good guidance on how to handle this, so I've implemented the recommended behaviour in this PR: https://github.com/across-protocol/relayer-v2/pull/1130

exit

scripts/spokepool.ts
pxrl: Documented here: https://github.com/across-protocol/relayer-v2/pull/1122/commits/e205d019bbb9092d8162cc2092af52cfd9869a5d

I'll merge it with this constraint and will follow up with the narrowing PR once it's merged into the sdk.

PR #1125, opened 2023-12-19
refactor: Minor updates post SDK-v2 bump

pxrl:
This includes the recent SDK change to remove `latestBlockNumber` from the {Hub,Spoke}PoolClient implementations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.8",
     "@across-protocol/contracts-v2": "2.4.7",
-    "@across-protocol/sdk-v2": "0.20.3",
+    "@across-protocol/sdk-v2": "0.20.5",
     "@arbitrum/sdk": "^3.1.3",
     "@defi-wonderland/smock": "^2.3.5",
     "@eth-optimism/sdk": "^3.1.0",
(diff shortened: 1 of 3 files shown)


pxrl:
nb. I applied `do not merge` because this PR needs careful checking.

pxrl:
Note also overlap with this PR: https://github.com/across-protocol/relayer-v2/pull/1124/files

I'll cross-check to see whether there are any differences.

pxrl:
I'll repurpose this to include some of the minor refactoring that isn't covered by Nick's PR.

src/dataworker/Dataworker.ts
```diff
@@ -1934,12 +1936,7 @@ export class Dataworker {
             return false;
           }
           const followingBlockNumber =
-            hubPoolClient.getFollowingRootBundle(bundle)?.blockNumber || hubPoolClient.latestBlockSearched;
-
-          if (followingBlockNumber === undefined) {
-            return false;
-          }
```
mrice32, line 1941: Does this change behavior?

mrice32 (approved):

src/dataworker/Dataworker.ts
pxrl: It shouldn't be possible, since that would require that the HubPoolClient instance was never updated. I can't see how we'd end up this deep in the code without having done that. For instance, this code lives in a loop that iterates over parsed root bundles, so if we haven't done an update then we'd never enter into the loop because there'd be nothing to iterate over.

As background, `latestBlockSearched` was previously defined as `number | undefined`, but is now defined as `latestBlockSearched = 0`. So to retain 100% matching logic, we'd check `if (followingBlockNumber === 0)`, but in practice I think the check was only necessary to silence tsc "possibly undefined" warnings.

PR #1127, opened 2023-12-20
fix(ClientHelper): Use latest searched HubPoolClient block instead of eventSearchConfig.toBlock

nicholaspai:
https://github.com/across-protocol/sdk-v2/pull/475 resets the eventSearchConfig.toBlock to undefined after the update so we shouldn't read it directly

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -177,10 +177,7 @@ export async function constructSpokePoolClientsWithStartBlocks(
 
   // Explicitly set toBlocks for all chains so we can re-use them in other clients to make sure they all query
   // state to the same "latest" block per chain.
-  if (hubPoolClient.eventSearchConfig.toBlock === undefined) {
-    throw new Error("HubPoolClient eventSearchConfig.toBlock is undefined");
-  }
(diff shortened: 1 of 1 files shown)


src/common/ClientHelper.ts
```diff
@@ -177,10 +177,7 @@ export async function constructSpokePoolClientsWithStartBlocks(
 
   // Explicitly set toBlocks for all chains so we can re-use them in other clients to make sure they all query
   // state to the same "latest" block per chain.
-  if (hubPoolClient.eventSearchConfig.toBlock === undefined) {
-    throw new Error("HubPoolClient eventSearchConfig.toBlock is undefined");
-  }
-  const hubPoolBlock = await hubPoolClient.hubPool.provider.getBlock(hubPoolClient.eventSearchConfig.toBlock);
+  const hubPoolBlock = await hubPoolClient.hubPool.provider.getBlock(hubPoolClient.latestBlockSearched);
```
james-a-morris, line 180: nit: we may want to think about changing `latestBlockSearched` -> `latestBlockNumberSearched` since it's a number and not a block

james-a-morris (approved):

src/common/ClientHelper.ts
nicholaspai: hmm but is that not implied? not sure i agree but i'm open to a future PR

PR #1128, opened 2023-12-20
feat(multicaller): Enable MulticallerClient user to specify which transactions should never be sent as a batch txn

nicholaspai:
Unbatched txns can be forced to be sent from EOA rather than through Multicall3 contract

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -2127,6 +2127,9 @@ export class Dataworker {
           // If simulating mainnet execution, can fail as it may require funds to be sent from
           // pool rebalance leaf.
           canFailInSimulation: leaf.chainId === this.clients.hubPoolClient.chainId,
+          // @dev: If txn is to be sumbmitted to networkId=137, do not batch with other txns as this function call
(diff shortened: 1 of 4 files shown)


james-a-morris (commented):
Just a few questions to better understand this PR

src/dataworker/Dataworker.ts
```diff
@@ -2127,6 +2127,9 @@ export class Dataworker {
           // If simulating mainnet execution, can fail as it may require funds to be sent from
           // pool rebalance leaf.
           canFailInSimulation: leaf.chainId === this.clients.hubPoolClient.chainId,
+          // @dev: If txn is to be sumbmitted to networkId=137, do not batch with other txns as this function call
+          // must be executed from EOA.
+          unbatched: Number(chainId) === 137,
```
pxrl, line 2132: This might need an additional `import { CHAIN_IDs } from "@across-protocol/constants-v2";

```suggestion
          unbatched: Number(chainId) === CHAIN_IDs.POLYGON,
```

pxrl (approved):
This seems OK and required surprisingly few changes :+1:

PR #1130, opened 2023-12-22
refactor: Improve spokepool script error handling

pxrl:
Follow the node.js guidance by setting process.exitCode rather than exiting directly. This aligns with what was already implemented in the HubPool script.

### DIFF
--- a/scripts/hubpool.ts
+++ b/scripts/hubpool.ts
@@ -243,16 +243,17 @@ async function run(argv: string[]): Promise<number> {
 
   config();
 
+  const cmd = argv[0];
   let signer: Signer;
   try {
-    const keyType = ["dispute"].includes(argv[0]) ? args.wallet : "void";
+    const keyType = ["dispute"].includes(cmd) ? args.wallet : "void";
     signer = await getSigner({ keyType, cleanEnv: true });
   } catch (err) {
     return usage(args.wallet) ? NODE_SUCCESS : NODE_INPUT_ERR;
   }
 
   let result: boolean;
-  switch (argv[0]) {
+  switch (cmd) {
     case "dispute":
       result = await dispute(args, signer);
       break;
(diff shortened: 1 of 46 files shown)


pxrl:
Just to flag, this was initially prompted by @mrice32 feedback here: https://github.com/across-protocol/relayer-v2/pull/1122#pullrequestreview-1791533176

mrice32 (approved):
LGTM!

scripts/spokepool.ts
```diff
@@ -342,39 +342,39 @@ async function run(argv: string[]): Promise<boolean> {
...
+  return result ? NODE_SUCCESS : NODE_APP_ERR;
 }
 
 if (require.main === module) {
   run(process.argv.slice(2))
-    .then(async () => {
-      // eslint-disable-next-line no-process-exit
-      process.exit(0);
-    })
+    .then(async (result) => (process.exitCode = result))
```
mrice32, line 375: Returning the exit code does seem simpler, especially if we set the exitCode in exactly one function (which you are doing here).

Would make a minor nit suggestion to just expand this into {} and not use the assignment as an expression.

james-a-morris: +1 the resultant of the `=` operator doesn't feel very intuitive to read.

scripts/hubpool.ts
```diff
@@ -246,34 +243,33 @@ async function run(argv: string[]): Promise<number> {
 
   config();
 
+  const cmd = argv[0];
```
james-a-morris, line 246: Nice break of `argv[0]` into a variable for later use

james-a-morris (approved):

scripts/spokepool.ts
pxrl: Addressed here: https://github.com/across-protocol/relayer-v2/pull/1130/commits/d4493f53d80f2cdd2dd25afeb20fbee38e6573d2

PR #1139, opened 2024-01-05
feat: Dynamically adjust proposal block ranges

pxrl:
This commit introduces dynamic adjustment of the block ranges that the proposer will propose. The intention of this change is to make the proposer more robust to incomplete or inconsistent data being supplied by one or more RPC providers.

Dynamic adjustment is enacted in the event that any of the following conditions are detected:
 - An invalid fill traces to a missing deposit (no deposit found for the specified depositId).
 - A deposit is known to be filled, without any associated FilledRelay event.

In the first case, this results in narrowing of the block range on both the origin and destination chains. For the latter, only the destination chain is narrowed.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -1,8 +1,10 @@
+import assert from "assert";
 import * as _ from "lodash";
 import {
   DepositWithBlock,
   FillsToRefund,
   FillWithBlock,
+  InvalidFill,
   ProposedRootBundle,
   RefundRequestWithBlock,
   UnfilledDeposit,
@@ -38,6 +40,7 @@ type DataCacheValue = {
   unfilledDeposits: UnfilledDeposit[];
   fillsToRefund: FillsToRefund;
   allValidFills: FillWithBlock[];
+  allInvalidFills: InvalidFill[];
   deposits: DepositWithBlock[];
   earlyDeposits: typechain.FundsDepositedEvent[];
 };
@@ -176,6 +179,7 @@ export class BundleDataClient {
     unfilledDeposits: UnfilledDeposit[];
     fillsToRefund: FillsToRefund;
     allValidFills: FillWithBlock[];
+    allInvalidFills: InvalidFill[];
     deposits: DepositWithBlock[];
     earlyDeposits: typechain.FundsDepositedEvent[];
   }> {
@@ -190,6 +194,7 @@ export class BundleDataClient {
     }
     return this._loadData(blockRangesForChains, spokePoolClients, isUBA, logData);
   }
+
   async _loadData(
     blockRangesForChains: number[][],
     spokePoolClients: { [chainId: number]: SpokePoolClient },
@@ -199,6 +204,7 @@ export class BundleDataClient {
     unfilledDeposits: UnfilledDeposit[];
     fillsToRefund: FillsToRefund;
     allValidFills: FillWithBlock[];
+    allInvalidFills: InvalidFill[];
     deposits: DepositWithBlock[];
     earlyDeposits: typechain.FundsDepositedEvent[];
   }> {
@@ -225,7 +231,7 @@ export class BundleDataClient {
     const allRelayerRefunds: { repaymentChain: number; repaymentToken: string }[] = [];
     const deposits: DepositWithBlock[] = [];
     const allValidFills: FillWithBlock[] = [];
-    const allInvalidFills: FillWithBlock[] = [];
+    const allInvalidFills: InvalidFill[] = [];
     const earlyDeposits: typechain.FundsDepositedEvent[] = [];
 
     // Save refund in-memory for validated fill.
@@ -275,23 +281,27 @@ export class BundleDataClient {
       updateTotalRefundAmount(fillsToRefund, fill, chainToSendRefundTo, repaymentToken);
     };
 
-    const validateFillAndSaveData = async (fill: FillWithBlock, blockRangeForChain: number[]): Promise<void> => {
+    const validateFillAndSaveData = async (fill: FillWithBlock, blockRangeForChain: number[]): Promise<boolean> => {
       const originClient = spokePoolClients[fill.originChainId];
       const matchedDeposit = originClient.getDepositForFill(fill);
       if (matchedDeposit) {
         addRefundForValidFill(fill, matchedDeposit, blockRangeForChain);
-      } else {
-        // Matched deposit for fill was not found in spoke client. This situation should be rare so let's
-        // send some extra RPC requests to blocks older than the spoke client's initial event search config
-        // to find the deposit if it exists.
-        const spokePoolClient = spokePoolClients[fill.originChainId];
-        const historicalDeposit = await queryHistoricalDepositForFill(spokePoolClient, fill);
-        if (historicalDeposit.found) {
-          addRefundForValidFill(fill, historicalDeposit.deposit, blockRangeForChain);
-        } else {
-          allInvalidFills.push(fill);
-        }
+        return true;
       }
+
+      // Matched deposit for fill was not found in spoke client. This situation should be rare so let's
+      // send some extra RPC requests to blocks older than the spoke client's initial event search config
+      // to find the deposit if it exists.
+      const spokePoolClient = spokePoolClients[fill.originChainId];
+      const historicalDeposit = await queryHistoricalDepositForFill(spokePoolClient, fill);
+      if (historicalDeposit.found) {
+        addRefundForValidFill(fill, historicalDeposit.deposit, blockRangeForChain);
+        return true;
+      }
+
+      assert(historicalDeposit.found === false); // Help tsc to narrow the discriminated union type.
+      allInvalidFills.push({ fill, code: historicalDeposit.code, reason: historicalDeposit.reason });
+      return false;
     };
 
     const _isChainDisabled = (chainId: number): boolean => {
@@ -450,7 +460,14 @@ export class BundleDataClient {
       });
     }
 
-    this.loadDataCache[key] = { fillsToRefund, deposits, unfilledDeposits, allValidFills, earlyDeposits };
+    this.loadDataCache[key] = {
+      fillsToRefund,
+      deposits,
+      unfilledDeposits,
+      allValidFills,
+      allInvalidFills,
+      earlyDeposits,
+    };
 
     return this.loadDataFromCache(key);
   }
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1,9 +1,11 @@
+import assert from "assert";
 import { utils as ethersUtils } from "ethers";
 import {
   winston,
   EMPTY_MERKLE_ROOT,
   sortEventsDescending,
   BigNumber,
+  getNetworkName,
   getRefund,
   MerkleTree,
   sortEventsAscending,
@@ -18,6 +20,7 @@ import {
   DepositWithBlock,
   FillsToRefund,
   FillWithBlock,
+  InvalidFill,
   isUbaOutflow,
   outflowIsFill,
   ProposedRootBundle,
@@ -29,6 +32,7 @@ import {
   RunningBalances,
   PoolRebalanceLeaf,
   RelayerRefundLeaf,
+  RelayData,
 } from "../interfaces";
 import { DataworkerClients } from "./DataworkerClientHelper";
 import { SpokePoolClient, UBAClient, BalanceAllocator } from "../clients";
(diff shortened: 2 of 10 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -453,15 +457,167 @@ export class Dataworker {
...
+        if (blockNumber <= endBlock) {
+          // Find the fill immediately preceding this invalid fill.
+          const previousFill = destSpokePoolClient
+            .getFillsWithBlockInRange(startBlock, Math.max(blockNumber - 1, startBlock))
+            .at(-1);
+
+          // Wind back to the bundle end block number to that of the previous fill.
+          const newEndBlock = previousFill?.blockNumber ?? startBlock;
+          updatedBlockRanges[destinationChainId] =
+            newEndBlock > startBlock ? [startBlock, newEndBlock] : [startBlock - 1, startBlock - 1]; // @fix: Must use previous end block!
```
pxrl, line 518: Flagging a todo; prefer not to rely on startBlock - 1 as the inferred previous end block.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +457,167 @@ export class Dataworker {
...
+    const newUnfilledDeposits = deposits.filter(({ originChainId, depositId, amount, blockNumber }) => {
+      if (blockNumber > updatedBlockRanges[originChainId][1]) {
+        return false; // Fill event is already out of scope due to previous narrowing.
+      }
+      return allValidFills.find(
+        (fill) =>
+          fill.depositId === depositId && fill.originChainId === originChainId && fill.totalFilledAmount.eq(amount)
+      )
+        ? false
+        : true;
```
pxrl, line 542: Prettier made sure to uglify this.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +457,167 @@ export class Dataworker {
...
+      const { spokePool, deploymentBlock } = spokePoolClients[deposit.destinationChainId];
+      const blockRange = updatedBlockRanges[deposit.destinationChainId];
+      const endBlock = blockRange[1];
+
+      // @todo: Some tests rely on this; fix!
+      const startBlock = blockRange[0] > deploymentBlock ? blockRange[0] : deploymentBlock;
+
+      // @todo: Beware the scenario where the fill is completed before the deposit, yet the deposit hash matches 100%.
+      // This corner case must be ruled out or mitigated before merge, because it will cause the proposer to throw.
+      return await sdk.utils.findFillBlock(spokePool, deposit as RelayData, startBlock, endBlock);
```
pxrl, line 556: Flagging a todo - if the fill was completed before `startBlock` then `findFillBlock()` will throw, and this will upend the proposer. I'm not sure whether this can happen in prod, but it needs to be ruled out or mitigated.

test/Dataworker.proposeRootBundle.ts
```diff
@@ -1,29 +1,50 @@
...
+  deleteFill(originChainId: number, depositId: number): void {
+    const fills = this.fills[originChainId];
+    const depositIdx = fills
+      .map((fill) => `${fill.originChainId}-${fill.depositId}`)
+      .indexOf(`${originChainId}-${depositId}`);
+    assert(depositIdx !== -1);
+
+    this.fills[originChainId].splice(depositIdx, 1);
+  }
+}
```
pxrl, line 42: I defined this class here because it's bespoke functionality that is (so far) only required in this specific test.

pxrl (commented):
Just adding some context.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +457,167 @@ export class Dataworker {
...
   async Legacy_proposeRootBundle(
     blockRangesForProposal: number[][],
     spokePoolClients: SpokePoolClientsByChain,
     latestMainnetBundleEndBlock: number,
     logData = false
   ): Promise<ProposeRootBundleReturnType> {
     const timerStart = Date.now();
+
+    // Dry-run the proposal over the input block ranges to identify any data inconsistencies.
+    const blockRanges = await this.narrowProposalBlockRanges(blockRangesForProposal, spokePoolClients);
```
pxrl, line 616: todo: This exact path is _also_used for root bundle validation. iIn that case we explicitly do not want to modify the block ranges, and should instead proceed with the block ranges that were specified in the proposal.

In the event that they do differ, the result would be a dispute anyway, so this doesn't actually risk misbehaviour, but it is counter-intuitive and probably undesirable anyway.

pxrl: Should be resolved as at this change: https://github.com/across-protocol/relayer-v2/pull/1139/commits/fb12aa072dd769f12caeb62778c2800bc3a9d567

pxrl:
fwiw am aware of a single test failure associated with this change. It's not obvious to me why it's failing at the moment; I will dig into it shortly.

test/Dataworker.proposeRootBundle.ts
pxrl, line 1: nb. In general, this new feature probably warrants some additional test cases to stress out the implementation. This is especially important as it's not possible to backtest the change against previous bundles.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,164 @@ export class Dataworker {
...
+        const originSpokePoolClient = spokePoolClients[originChainId];
+        let [startBlock, endBlock] = updatedBlockRanges[originChainId];
+
+        // Find the previous known deposit. This may not be immediastely preceding depositId.
+        const previousDeposit = originSpokePoolClient
+          .getDepositsForDestinationChain(destinationChainId)
+          .filter((deposit: DepositWithBlock) => deposit.blockNumber < blockNumber)
+          .at(-1);
+
+        if (previousDeposit?.blockNumber ?? startBlock > startBlock) {
```
nicholaspai, line 496: Is there a mistake on the RHS: start > start?

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,164 @@ export class Dataworker {
...
+            previousBlockRange: [startBlock, endBlock],
+            newBlockRange: updatedBlockRanges[destinationChainId],
+          });
+        }
+      }
+    });
+
+    // Quick sanity check - make sure that the block ranges are coherent. A chain may be soft-paused if it has ongoing
+    // RPC issues (block ranges are frozen at the previous proposal endBlock), so ensure that this is also handled.
+    const finalBlockRanges = chainIds.map((chainId) => updatedBlockRanges[chainId]);
```
nicholaspai, line 596: Can we reuse logic from elsewhere in the dataworker that validates bundle block ranges? I'm sure the disputer dies something like this

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,164 @@ export class Dataworker {
...
+            previousBlockRange: [startBlock, endBlock],
+            newBlockRange: updatedBlockRanges[destinationChainId],
+          });
+        }
+      });
+
+    // For each deposit within the origin chain block range for which no or only partial fill events are found,
+    // verify whether the fill has been completed on the destination chain. If the fill has been made to completion
+    // then this is evidence of potential dropped/suppressed events, so narrow the block range to exclude those blocks.
+    const newUnfilledDeposits = deposits.filter(({ originChainId, depositId, amount, blockNumber }) => {
```
nicholaspai, line 535: Can we potentially skip this check and the following in favor of just the above to reduce the complexity of this PR? IIUC there are multiple checks run here to progressively narrow down the block ranges and it might be easier to test with a broader filter at first

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,164 @@ export class Dataworker {
...
+      isUBA,
+      logData
+    );
+
+    // If invalid fills were detected and they appear to be due to gaps in FundsDeposited events:
+    // - Narrow the origin block range to exclude the missing deposit, AND
+    // - Narrow the destination block range to exclude the invalid fill.
+    allInvalidFills
+      .filter(({ code }) => code === InvalidFill.DepositIdNotFound)
+      .forEach(({ fill: { depositId, originChainId, destinationChainId, blockNumber } }) => {
```
nicholaspai, line 483: Can we sort these by block number by chain to allow us to exit the following map earlier?

nicholaspai (commented):
Left some comments on first read through

src/dataworker/Dataworker.ts
pxrl: I don't think so; `previousDeposit?.blockNumber ?? startBlock` is intended to catch the scenario where the missing deposit is the first depositId in the block range. So we should only go searching for an earlier deposit if there are actually blocks between `previousDeposit?.blockNumber` and `startBlock`.

src/dataworker/Dataworker.ts
pxrl: Yeah I think there is a helper function that does this as well - will sub it in.

src/dataworker/Dataworker.ts
pxrl: I simplified the implementation by factoring out the code that updates the end block number. This allows central checking that the update is sensible, and it subs in the correct end block in the event that the chain has soft-paused: https://github.com/across-protocol/relayer-v2/pull/1139/commits/c716ee87503964a7467ccc10b5f076ac02689584

src/dataworker/Dataworker.ts
pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/commit/c716ee87503964a7467ccc10b5f076ac02689584

src/dataworker/Dataworker.ts
pxrl: Quick status on this one - there is some logic in the disputer, but it's not currently reusable in this function. Some refactoring is needed over there to make the block range checks useful in this context. I'm WIP on that, but it's taken the back seat to USS work.

In general, I think it'd be good to share the validation between the disputer and the proposal, but the checks here are so compact and targeted that it shouldn't necessarily hold this PR back.

src/dataworker/Dataworker.ts
pxrl: I think these checks are important to have in because I'm concerned about scenarios where we only miss `FilledRelay` events, but I can also see some merit to incrementally introducing the functionality. wdyt about splitting it so we first introduce the deposit-only check, and see how that goes in prod?

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,168 @@ export class Dataworker {
...
+
+        // Exclude the missing deposit on the origin chain.
+        const originSpokePoolClient = spokePoolClients[originChainId];
+        let [startBlock, endBlock] = updatedBlockRanges[originChainId];
+
+        // Find the previous known deposit. This may not be immediastely preceding depositId.
+        const previousDeposit = originSpokePoolClient
+          .getDepositsForDestinationChain(destinationChainId)
+          .filter((deposit: DepositWithBlock) => deposit.blockNumber < blockNumber)
+          .at(-1);
```
mrice32, line 513: What if the deposit is from a gap between spoke pool deployments or something? So there actually is no deposit for this id? Will this always return undefined in that case, which would mean no change?

If there are edge cases like that that are handled in a subtle way here, we may want to add a brief comment explaining them.

pxrl: The SpokePoolClient only has the ability to handle a single SpokePool deployment, so in the case that we don't find any preceding deposit then `previousDeposit` resolves to `undefined`. In this case, `Dataworker.updateEndBlock()` detects that the updated `endBlock` is undefined and implements a soft-pause of the chain by proposing over the previous endBlock (as we do for Boba).

I'll update the comment on line 509 to clarify this.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,168 @@ export class Dataworker {
...
+      });
+
+    // For each deposit within the origin chain block range for which no or only partial fill events are found,
+    // verify whether the fill has been completed on the destination chain. If the fill has been made to completion
+    // then this is evidence of potential dropped/suppressed events, so narrow the block range to exclude those blocks.
+    const newUnfilledDeposits = deposits.filter(({ originChainId, depositId, amount, blockNumber }) => {
+      if (blockNumber > updatedBlockRanges[originChainId][1]) {
+        return false; // Fill event is already out of scope due to previous narrowing.
+      }
+      return allValidFills.find(
```
mrice32, line 553: Can we use a spoke pool function or map lookup to get the fill for a deposit rather than looping through all fills on each deposit?

We store by a special hash in the client to avoid this n^2 loop (which has caused speed issues in the past).

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,168 @@ export class Dataworker {
...
+    });
+
+    // Quick sanity check - make sure that the block ranges are coherent. A chain may be soft-paused if it has ongoing
+    // RPC issues (block ranges are frozen at the previous proposal endBlock), so ensure that this is also handled.
+    const finalBlockRanges = chainIds.map((chainId) => updatedBlockRanges[chainId]);
+    const coherentBlockRanges = finalBlockRanges.every(([startBlock, endBlock], idx) => {
+      const [originalStartBlock] = blockRanges[idx];
+      return (
+        (endBlock > startBlock && startBlock === originalStartBlock) ||
+        (startBlock === endBlock && startBlock === originalStartBlock - 1) // soft-pause
```
mrice32, line 605: Why would the new start block be one behind the original start block in this case? And is it okay to leave it that way?

pxrl: `originalStartBlock` is normally `previousEndBlock + 1`, so if `startBlock === endBlock` is true then we're implementing a soft-pause on the chain, and will propose over the range of `[previousEndBlock, previousEndBlock]`. This is consistent with the way that chains are "hard-paused" (i.e. disabled via the ConfigStore).

To be more correct, rather than assuming that `previousEndBlock === originalStartBlock - 1`, this check should resolve the previous ending block via `HubPoolClient.getLatestBundleEndBlockForChain()`.

src/dataworker/Dataworker.ts
```diff
@@ -453,15 +459,168 @@ export class Dataworker {
...
+  }
+
   async Legacy_proposeRootBundle(
     blockRangesForProposal: number[][],
     spokePoolClients: SpokePoolClientsByChain,
     latestMainnetBundleEndBlock: number,
     logData = false
   ): Promise<ProposeRootBundleReturnType> {
     const timerStart = Date.now();
+
```
pxrl, line 620:
```suggestion
```

src/dataworker/Dataworker.ts
nicholaspai: Yeah i like introducing one simper check first and then if that works, adding more rules

src/dataworker/Dataworker.ts
pxrl: I have backed that entire section out (for now) per this thread with Nick: https://github.com/across-protocol/relayer-v2/pull/1139#discussion_r1447422412.

Here's the change: https://github.com/across-protocol/relayer-v2/pull/1139/commits/093d80bb8ef10a60b9a9b4707981c41139f9be85

I did consider accessing these via a map, but tentatively decided against it because I don't think this will actually use much CPU time in practice. Arriving here is conditional on at least one deposit is detected as filled, despite no known FilledRelay event. Then, we iterate over the àllValidFills` array once per instance of "filled-but-missing" event.

So in the normal case it costs nothing, but it does cost CPU in the bad case where a chain/RPC is misbehaving. However, because we've already filtered on the missing deposits and have narrowed the block range accordingly, we should skip over most of those missing fill events anyway. So in practice I think the impact of this would be fairly limited, even though it is technically inefficient to search the `allValidFills` array in this way. Does that make sense?

src/dataworker/Dataworker.ts
pxrl: Have backed out this section of code per this thread: https://github.com/across-protocol/relayer-v2/pull/1139#discussion_r1447422412.

pxrl: I'm not totally confident on this, but I'm not sure that we can actually simply sort and exit earlier. This is because we need to avoid making assumptions about the ordering of fills vs. deposits.

For example, the first invalid fill on a destination chain might correspond to the _last_ missing deposit on the origin chain. Then, some later invalid fill on the destination chain might actually fill an earlier missing deposit on the origin chain. This also seems more likely to occur in the case that the RPCs are serving inconsistent data.

If we group/sort by destinationChainId and destination block number, we might not narrow the originChainId block range correctly. If we group/sort originChainId and origin block number, we might not narrow the destinationChainid block range correctly.

In general, even when it's really bad, the number of missing deposits we typically see is about 20 - 30. In order to identify both the earliest block for both origin and destination chains we might end up looping multiple times over all of the invalid fills.In the case where we only have tens missing events, it seems like it might be cheaper and simpler to just process them one by one.

Full disclosure: I might have overlooked something really obvious here. wdyt?

src/dataworker/Dataworker.ts
pxrl: I've backed it out here: https://github.com/across-protocol/relayer-v2/pull/1139/commits/093d80bb8ef10a60b9a9b4707981c41139f9be85 :+1:

src/dataworker/Dataworker.ts
nicholaspai: this is a good point, that the latest deposit might match with the earliest fill and vice versra

src/dataworker/Dataworker.ts
nicholaspai: what would a bundle block range look like for a new spoke pool address? whether it be a new chain or an updated spoke pool address for an existing chain?

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,105 @@ export class Dataworker {
...
+        chainId
+      );
+
+      endBlock ??= previousEndBlock;
+      assert(
+        endBlock < currentEndBlock,
+        `Invalid block range update for chain ${chainId}: block ${endBlock} >= ${currentEndBlock}`
+      );
+
+      updatedBlockRanges[chainId] =
```
nicholaspai, line 492:
```suggestion
      // If endBlock is equal to or before the currentEndBlock then we are "soft-pausing" the chain by setting the bundle start and end block equal to the previous end block. This effectively tells the dataworker to not progress on that chain.
      updatedBlockRanges[chainId] =
```

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,105 @@ export class Dataworker {
...
+        const originSpokePoolClient = spokePoolClients[originChainId];
+        let [startBlock, endBlock] = updatedBlockRanges[originChainId];
+
+        // Find the previous known deposit. This may resolve a deposit before the immediately preceding depositId.
+        const previousDeposit = originSpokePoolClient
+          .getDepositsForDestinationChain(destinationChainId)
+          .filter((deposit: DepositWithBlock) => deposit.blockNumber < blockNumber)
+          .at(-1);
+
+        updateEndBlock(originChainId, previousDeposit?.blockNumber);
```
nicholaspai, line 514: Can someone grief the dataworker by now sending a fill for an obviously invalid deposit ID like 1 or 9_999_999_999 and always make it narrow its block range?

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,105 @@ export class Dataworker {
...
+          depositId,
+          previousBlockRange: [startBlock, endBlock],
+          newBlockRange: updatedBlockRanges[originChainId],
+        });
+
+        // Update the endBlock on the destination chain to exclude the invalid fill.
+        const destSpokePoolClient = spokePoolClients[destinationChainId];
+        [startBlock, endBlock] = updatedBlockRanges[destinationChainId];
+
+        if (blockNumber <= endBlock) {
```
nicholaspai, line 527: nit: wouldn't `blockNumber`s for invalid fills always be within the bundle block range by definition when calling `loadData`?

nicholaspai (commented):
I think the new implementation is a lot more clear and therefore I left more targeted questions

src/dataworker/Dataworker.ts
pxrl: The `DepositIdNotFound` cause is only set if we expect that the deposit should have been found within the SpokePoolClient's event cache - i.e. the events that should fall within the block range that were just scraped. The relevant code is here: https://github.com/across-protocol/sdk-v2/blob/1ba43981cdb3667670e1f683675a681d43de84bc/src/utils/DepositUtils.ts#L64.

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,105 @@ export class Dataworker {
...
+          // Find the fill immediately preceding this invalid fill.
+          const previousFill = destSpokePoolClient
+            .getFillsWithBlockInRange(startBlock, Math.max(blockNumber - 1, startBlock))
```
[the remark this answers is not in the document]
pxrl: It's OK for it to be invalid, but if it's invalid specifically for `DepositIdNotFound` then it will also be in the `allInvalidFills` array, so we will separately check for that and narrow the block range further if necessary.

src/dataworker/Dataworker.ts
pxrl: Not relevant anymore, since this code has been pulled out of this PR.

src/dataworker/Dataworker.ts
pxrl: Not necessarily, because we source `endBlock` from `updatedBlockRanges`, so it is subject to being iteratively updated within this loop. So if we are examining a fill where the `blockNumber` has already been excluded by narrowing then we should just skip over that.

pxrl: Clarified here: https://github.com/across-protocol/relayer-v2/pull/1139/commits/75599485dd8c23cd0fa6768133b7fd94f30c3f27

src/dataworker/Dataworker.ts
pxrl: The bundlEndBlock is sourced from [HubPoolClient.getLatestBundleEndBlockForChain()](https://github.com/across-protocol/sdk-v2/blob/1ba43981cdb3667670e1f683675a681d43de84bc/src/clients/HubPoolClient.ts#L687), so we inherit its behaviour. For a new deployment on an existing chain, we're bound to continue from the previous bundleEndBlock. This should work as expected.

For a new chain, the chainId index isn't found in the previous bundle, so we default to proposing from 0. This is also as was the case for the activation of zkSync and Base - so we inherit the existing behaviour.

The only change that I can foresee here is that if that initial proposal contains invalid fills, then we'd revert to proposing over [0,0]. I'm not sure whether that's a problem in itself, but it's a pretty extreme edge case. and we'd detect it immediately because activating a new chain implies manual review of the proposal block ranges.

nicholaspai (dismissed):
I'm on board with merging this, closely monitoring the data worker proposals for a few days

OR

adding some more unit tests before we activate this.

We are probably safe to launch this and in the worst case, we'll self-dispute since the disputer does not have this `narrowBlockRanges` logic. And AFAICT the disputer's logic to construct the block ranges from `getWidestPossibleBlockRanges` hasn't changed

src/dataworker/Dataworker.ts
mrice32: It does, it sounds like you're saying this loop will be run rarely so optimization isn't a concern. I wasn't aware of that when I initially read the code, but will take another pass.

src/dataworker/Dataworker.ts
mrice32: Understood. That makes sense.

src/clients/BundleDataClient.ts
```diff
@@ -289,7 +295,8 @@ export class BundleDataClient {
         if (historicalDeposit.found) {
           addRefundForValidFill(fill, historicalDeposit.deposit, blockRangeForChain);
         } else {
-          allInvalidFills.push(fill);
+          assert(historicalDeposit.found === false); // Help tsc to narrow the discriminated union type.
```
mrice32, line 298: This is annoying -- ts really should not require runtime code to force it to _get_ that it must be this type. It won't compile without this line?

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,108 @@ export class Dataworker {
...
+      .forEach(({ fill: { depositId, originChainId, destinationChainId, blockNumber } }) => {
+        const [originChain, destinationChain] = [getNetworkName(originChainId), getNetworkName(destinationChainId)];
+
+        // Exclude the missing deposit on the origin chain.
+        const originSpokePoolClient = spokePoolClients[originChainId];
+        let [startBlock, endBlock] = updatedBlockRanges[originChainId];
+
+        // Find the previous known deposit. This may resolve a deposit before the immediately preceding depositId.
+        const previousDeposit = originSpokePoolClient
+          .getDepositsForDestinationChain(destinationChainId)
```
mrice32, line 512: I don't think these are guaranteed to be sorted, which might mean this deposit would be too early, right?

src/dataworker/Dataworker.ts
mrice32: I think the 0,0 case is remote and doesn't cause an obvious problem, so I wouldn't worry about handling it explicitly.

This all makes sense. I think a comment that says something like:
```
This can return undefined if there is no known preceding deposit in range.
That will result in the chain being soft-paused by setting the endBlock equal to the previous endBlock, making this bundle cover no blocks on that chain.
```

Would be really helpful to the reader.

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,108 @@ export class Dataworker {
...
+        chainIds,
+        this.clients.hubPoolClient.latestBlockSearched,
+        chainId
+      );
+
+      endBlock ??= previousEndBlock;
+      assert(
+        endBlock < currentEndBlock,
+        `Invalid block range update for chain ${chainId}: block ${endBlock} >= ${currentEndBlock}`
+      );
```
mrice32, line 490: Since these updates can come in out of order, I would expect this case to happen, no?

I find a missing deposit at block 7, then later I see a missing deposit at block 12.

Maybe I'm missing the logic that prevents this.

mrice32 (commented):
A few more comments! Sorry, this logic is pretty nuanced

src/clients/BundleDataClient.ts
james-a-morris: OOC @pxrl what is the error that is thrown?

src/dataworker/Dataworker.ts
```diff
@@ -263,10 +266,10 @@ export class Dataworker {
    * of log level
    * @returns Array of blocks ranges to propose for next bundle.
    */
-  _getNextProposalBlockRanges(
+  async _getNextProposalBlockRanges(
```
james-a-morris, line 269: It looks like this function only has one await and it's at the very last line. Thoughts on removing this `async` and returning the promise generated by the call to `this.narrowProposalBlockRanges(blockRangesForProposal, spokePoolClients)`

src/dataworker/Dataworker.ts
```diff
@@ -453,6 +458,108 @@ export class Dataworker {
     }
   }
 
+  async narrowProposalBlockRanges(
+    blockRanges: number[][],
+    spokePoolClients: SpokePoolClientsByChain,
+    logData = true,
+    isUBA = false
```
james-a-morris, line 465: Will this need to be changed for the UBA removal changes?

james-a-morris (commented):
Left a few comments

PR #1140, opened 2024-01-07
fix: Correct zkSync finalizer

pxrl:
A recent update to the way Hub -> Spoke tokens are resolved broke the zkSync finalizer. This only shows up now because it's the first time in several weeks.


nicholaspai (approved):

james-a-morris (commented):
LGTM - but I'm curious how this didn't get caught by `tsc`. Are we omitting these files from the compiler? 

My understanding is that this initial error should not have passed any tests because it shouldn't have built

pxrl:
This one was submitted here: https://github.com/across-protocol/relayer-v2/pull/1141

PR #1141, opened 2024-01-07
fix: Correct zkSync finalizer

dohaki:
A recent update to the way Hub -> Spoke tokens are resolved broke the zkSync finalizer. This only shows up now because it's the first time in several weeks.

Re-proposing https://github.com/across-protocol/relayer-v2/pull/1140


pxrl (approved):
Tested OK and finalized a withdrawal here: https://etherscan.io/tx/0xc9665c6b93871a2caaa2da7689b74ff9321bb3058cc82ff1c6f84b2624886fee

PR #1143, opened 2024-01-09
fix(InventoryClient): Don't try to query l2 token for l1 token on chain that it isn't enabled for

nicholaspai:


src/clients/InventoryClient.ts
```diff
@@ -88,11 +85,15 @@ export class InventoryClient {
     const cumulativeBalance = this.getCumulativeBalance(l1Token);
     const distribution: { [chainId: number]: BigNumber } = {};
     this.getEnabledChains().forEach((chainId) => {
+      // If token doesn't have entry on chain, skip creating an entry for it since we'll likely run into an error
+      // later trying to grab the chain equivalent of the L1 token via the HubPoolClient.
+      // if (this._l1TokenEnabledForChain(l1Token, chainId)) {
```
mrice32, line 90: Why comment?

nicholaspai: oops sorry was testing locally, this is meant to be the fix!

mrice32 (approved):

PR #1147, opened 2024-01-16
improve: deduplicate yarn.lock

james-a-morris:
This PR uses a popular tool to deduplicate our Yarn.lock file. Among other indicators, a duplicate package entry is a package that appears on multiple nodes of the dependency tree with overlapping version ranges but is resolved to different versions.

Package: https://www.npmjs.com/package/yarn-deduplicate
Command: `npx yarn-deduplicate yarn.lock`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11,15 +11,10 @@
     "@uma/common" "^2.17.0"
     hardhat "^2.9.3"
 
-"@across-protocol/constants-v2@1.0.8", "@across-protocol/constants-v2@^1.0.8":
+"@across-protocol/constants-v2@1.0.8", "@across-protocol/constants-v2@^1.0.4", "@across-protocol/constants-v2@^1.0.8":
   version "1.0.8"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -11,16 +11,11 @@
     "@uma/common" "^2.17.0"
     hardhat "^2.9.3"
 
-"@across-protocol/constants-v2@1.0.8", "@across-protocol/constants-v2@^1.0.8":
+"@across-protocol/constants-v2@1.0.8", "@across-protocol/constants-v2@^1.0.4", "@across-protocol/constants-v2@^1.0.8":
```
nicholaspai, line 14: why does this add 1.0.8 twice?

nicholaspai (approved):

yarn.lock
james-a-morris: These ranges are all considered different. 

- `@1.0.8` is pegged to the `1.0.8` version,
- `@^1.0.8` is pegged to `1.0.x` and allowed to take the latest minor version b/c it has a caret (currently at 1.0.8)
- `@^1.0.4` is pegged to `1.0.x` and allowed to take the latest minor version b/c it has a caret (currently at 1.0.8)

pxrl (approved):
Good initiative!

PR #1149, opened 2024-01-17
feat(finalizer): add L2->L1 Scroll claim support

james-a-morris:
This PR adds support for finalizing L2->L1 claims on Scroll. Scroll's L2->L1 communication requires a 2 step claim process - the second part of this process (the claim step) is accomplished here.

This PR also adds support for testing arbitrary new spoke chains through the use of `customSpokeAddresses`.

To test this PR, the following ENV variables were set:
```
CUSTOM_SPOKE_ADDRESSES={"534352":{"address":"0x2d4710f04da90184255782d3715224a6c776955d","registrationBlock":2828344}}
```

Additionally, the following _temporary_ changes were made to set the finalizer to generate valid multicall data:
1. The `ScrollFinalizer` only searched for claims from an EOA instead of claims from the spoke
2. The `ScrollFinalizer` hardcoded the L1 Token Symbol to USDC so that no custom tokens had to be set
     - I.e. All the following `l1TokenSymbol` will read as `USDC` instead of `WETH`.
3. The `constructSpokePoolClientsWithLookback` in `./src/finalizer/index.ts` was called with `enabledChains=[534352]`
4. The `ScrollFinalizer`'s `l1Messenger` was hardcoded to the Sepolia address: `0x50c7d3e7f7c656493D1D76aaa1a836CedfCBB16A`
5. The `ScrollFinalizer`'s call to the official Scroll Bridge History API was hardcoded to the sepolia testnet API

The resulting call produced the expected 2 `callData`/`target`s of:

```
[
  {
    callData: '0xc311b6fc000000000000000000000000481b20a927206af7a754db8b904b052e2781ea270000000000000000000000003da0bf44814cfc678376b331183827215821169500000000000000000000000000000000000000000000000000038d7ea4c6800000000000000000000000000000000000000000000000000000000000000532d700000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000001e000000000000000000000000000000000000000000000000000000000000000e484bd13b0000000000000000000000000fff9976782d46cc05630d1f6ebab18b2324d6b1400000000000000000000000053000000000000000000000000000000000000040000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d0000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d00000000000000000000000000000000000000000000000000038d7ea4c6800000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f3d200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000260fd4b797c60f443fb4877d0a7330c1191a2d4a989d131148fadf927657b89d5c215b0ac30d5f42c1b0d7675876f0cf7e5db4cc7bd5fdd5cd7c771a1594d710a1cbd660ca3e2701c40f63a51a601472c34b3449c0db23ecb55a4a9fb4124640dd621ddb9a356815c3fac1026b6dec5df3124afbadb485c9ba5a3e3398a04b7ba85dc2f42354689517daf941f15538ab343bbc7baaa8ece07caf8cef98f2e09ba700eb01ebfc9ed27500cd4dfc979272d1f0913cc9f66540d7e8005811109e1cf2d566df64372c1f8f2058247c297100760ee5ac0709bd2d5145794430da17349edfa3fe9cb6af2795edb2c5599c3f0798b5ead1b65c123c0e93ff9566a365411b59867cc5f7f196b93bae1e27e6320742445d290f2263827498b54fec539f756af2cff9ea9177d67c985b351e12f41e734588f6a5a159c38bb61861c68df2233c0f9dc3e7fe016e050eff260334f18a5d4fe391d82092319f5964f2e2eb7c1c3a5f8b13a49e282f609c317a833fb8d976d11517c571d1221a265d25af778ecf89299ea86c963fc5b2b5be71a206e04f23533b5a4aa91ab8119f51d9fa50050a25e7ad36d8d119e6c757d5883739a6cb87cafe909210330f425439df8fbddbb308d5c67add7c6caf302256adedf7ab114da0acfe870d449a3a489f781d659e8beccda7bce9f4e8618b6bd2f4132ce798cdc7a60e7e1460a7299e3c6342a579626d27874b09783cef2e7750f7ea24f6090c9ce47f33cf25ca5e16a1207b4a50fda2be1d3b5c807b281e4683cc6d6315cf95b9ade8641defcb32372f1c126e398ef7a1ef973d30ca636d922d10ae577c73bc4fe92699225f30c0c2e9d6727bceb256d',
    target: '0x50c7d3e7f7c656493D1D76aaa1a836CedfCBB16A'
  },
  {
    callData: '0xc311b6fc00000000000000000000000091e8addfe1358aca5314c644312d38237fc1101c0000000000000000000000008a54a2347da2562917304141ab67324615e9866d00000000000000000000000000000000000000000000000000038d7ea4c6800000000000000000000000000000000000000000000000000000000000000532d500000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000001a000000000000000000000000000000000000000000000000000000000000000a48eaac8a30000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d0000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d00000000000000000000000000000000000000000000000000038d7ea4c680000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f3ce00000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000260ce01f781e48a1acacc1374f324bb203264d35c8bd8ee6f12b6b8754002b39431ad3228b676f7d3cd4284a5443f17f1962b36e491b30a40b2405849e597ba5fb5bd660ca3e2701c40f63a51a601472c34b3449c0db23ecb55a4a9fb4124640dd621ddb9a356815c3fac1026b6dec5df3124afbadb485c9ba5a3e3398a04b7ba85dc2f42354689517daf941f15538ab343bbc7baaa8ece07caf8cef98f2e09ba700eb01ebfc9ed27500cd4dfc979272d1f0913cc9f66540d7e8005811109e1cf2d566df64372c1f8f2058247c297100760ee5ac0709bd2d5145794430da17349edfa3fe9cb6af2795edb2c5599c3f0798b5ead1b65c123c0e93ff9566a365411b59867cc5f7f196b93bae1e27e6320742445d290f2263827498b54fec539f756af2cff9ea9177d67c985b351e12f41e734588f6a5a159c38bb61861c68df2233c0f9dc3e7fe016e050eff260334f18a5d4fe391d82092319f5964f2e2eb7c1c3a5f8b13a49e282f609c317a833fb8d976d11517c571d1221a265d25af778ecf89299ea86c963fc5b2b5be71a206e04f23533b5a4aa91ab8119f51d9fa50050a25e7ad36d8d119e6c757d5883739a6cb87cafe909210330f425439df8fbddbb308d5c67add7c6caf302256adedf7ab114da0acfe870d449a3a489f781d659e8beccda7bce9f4e8618b6bd2f4132ce798cdc7a60e7e1460a7299e3c6342a579626d27874b09783cef2e7750f7ea24f6090c9ce47f33cf25ca5e16a1207b4a50fda2be1d3b5c807b281e4683cc6d6315cf95b9ade8641defcb32372f1c126e398ef7a1ef973d30ca636d922d10ae577c73bc4fe92699225f30c0c2e9d6727bceb256d',
    target: '0x50c7d3e7f7c656493D1D76aaa1a836CedfCBB16A'
  }
]
```
Along with these `withdrawals`:
```
[
  {
    l2ChainId: 534352,
    l1TokenSymbol: 'USDC', // Technically WETH
    amount: '1000000000000000',
    type: 'withdrawal'
  },
  {
    l2ChainId: 534352,
    l1TokenSymbol: 'USDC', // Technically GHO (Arbitrary ERC20)
    amount: '1000000000000000',
    type: 'withdrawal'
  }
]
```

A successful result can be seen via this Tenderly simulation: https://www.tdly.co/shared/simulation/57aa2cdf-8212-48fc-8885-016882118083


src/finalizer/index.ts
```diff
@@ -63,6 +66,8 @@ export async function finalize(
     324: oneDaySeconds * 4,
     8453: optimisticRollupFinalizationWindow,
     42161: optimisticRollupFinalizationWindow,
+    534352: oneHourSeconds * 4, // Scroll Mainnet
+    534351: oneHourSeconds * 4, // Scroll Sepolia
```
james-a-morris, line 70: We are relying on the `BridgeHistoryAPI` which doesn't seem to have a window. However, Scroll has a 1hr L2 -> L1 lockout so we should have more than enough padding here.

nicholaspai (approved):
This looks great. So to test the finalizer against unofficial spokes, you only need to set the custom spoke pool address? 

I could also see removing the custom spoke pool logic in favor of making the tester just manually hardcode the spoke pool contract. But if this test config is that easy to set then this is a nice QoL improvement. 

Also in the PR description you note that the finalization is a two step process. How does the first step work? Presumably it's some sort of txn you should send to kick off the 1hr liveness to initiate the outstanding claim?

james-a-morris:
> This looks great. So to test the finalizer against unofficial spokes, you only need to set the custom spoke pool address?
> 
> I could also see removing the custom spoke pool logic in favor of making the tester just manually hardcode the spoke pool contract. But if this test config is that easy to set then this is a nice QoL improvement.
> 
> Also in the PR description you note that the finalization is a two step process. How does the first step work? Presumably it's some sort of txn you should send to kick off the 1hr liveness to initiate the outstanding claim?

The full steps are outlined in the description, but the only changes outside of the chain-specific finalizer that are needed is to set the `enabledChains: [{chainId}]` in the `./src/finalizer/index.ts` and to set the `CUSTOM_SPOKE_ADDRESSES` in the ENV.

It's a two-step process from L2->L1.
1. Execute a call to the Scroll Messenger on L2
    - The SpokePool will complete this step when it attempts to refund via the Scroll bridge
3. After ~1Hr, execute a claim to the Scroll Messenger on L1
    - We are here ^

pxrl:
@james-a-morris I'm not so enthusiastic about putting test functionalities directly in the HubPoolClient that runs in production. I was wondering whether we could instead have a separate TestHubPoolClient that inherits from HubPoolClient, and is instantiated on some other environment? Then we could bundle any test functionality in there, and its use would be gated by a single env var (i.e. TEST_CLIENTS=true) or something. wdyt?

james-a-morris:
> @james-a-morris I'm not so enthusiastic about putting test functionalities directly in the HubPoolClient that runs in production. I was wondering whether we could instead have a separate TestHubPoolClient that inherits from HubPoolClient, and is instantiated on some other environment variable. Then we could bundle any test functionality in there, and its use would be gated by a single env var (i.e. TEST_CLIENTS=true) or something. wdyt?

Ah so you're saying we should upstream it entirely - I like that. I do think this is getting out of scope of the spirit of this PR so I will break out that logic into its own branch and we can discuss there.

pxrl:
> > @james-a-morris I'm not so enthusiastic about putting test functionalities directly in the HubPoolClient that runs in production. I was wondering whether we could instead have a separate TestHubPoolClient that inherits from HubPoolClient, and is instantiated on some other environment variable. Then we could bundle any test functionality in there, and its use would be gated by a single env var (i.e. TEST_CLIENTS=true) or something. wdyt?
> 
> Ah so you're saying we should upstream it entirely - I like that. I do think this is getting out of scope of the spirit of this PR so I will break out that logic into its own branch and we can discuss there.

I'd be fine with it living in relayer-v2 while it's matured, but ultimately it could be upstreamed to sdk-v2. But yeah - good idea to split the custom config part out of this PR so it can go in, because the remainder is ++simple and also LGTM, and we can flesh out the test functionality in another PR :+1+

james-a-morris:
Moved to here for now - I can move it into the SDK at a later point

james-a-morris:
> > > @james-a-morris I'm not so enthusiastic about putting test functionalities directly in the HubPoolClient that runs in production. I was wondering whether we could instead have a separate TestHubPoolClient that inherits from HubPoolClient, and is instantiated on some other environment variable. Then we could bundle any test functionality in there, and its use would be gated by a single env var (i.e. TEST_CLIENTS=true) or something. wdyt?
> > 
> > 
> > Ah so you're saying we should upstream it entirely - I like that. I do think this is getting out of scope of the spirit of this PR so I will break out that logic into its own branch and we can discuss there.
> 
> I'd be fine with it living in relayer-v2 while it's matured, but ultimately it could be upstreamed to sdk-v2. But yeah - good idea to split the custom config part out of this PR so it can go in, because the remainder is ++simple and also LGTM, and we can flesh out the test functionality in another PR :+1+

Moved here (https://github.com/across-protocol/relayer-v2/pull/1151) for now. We can further push it to the SDK at a later time.

james-a-morris:
@pxrl rolled back the changes on this PR

mrice32 (approved):
LGTM!

src/finalizer/index.ts
james-a-morris: By lockout I just mean we don't need to query the chain so this entry isn't as important

PR #1150, opened 2024-01-19
fix: Increase Polygon lookback

pxrl:
The Polygon finalizer instance limits its lookback to head - lookback, which is currently set to 1 day. If the finalizer is down or ineffective for > 1 day then this risks ignoring transactions that would otherwise be able to be finalised. There's very limited benefit to limiting the lookback to only 1 day, so pad it out to 5.


mrice32 (approved):
Will this substantially increase RPC usage or slow the bot down?

src/finalizer/index.ts
```diff
@@ -53,8 +53,8 @@ export async function finalize(
   hubPoolClient: HubPoolClient,
   spokePoolClients: SpokePoolClientsByChain,
   configuredChainIds: number[],
-  optimisticRollupFinalizationWindow: number = 5 * oneDaySeconds,
-  polygonFinalizationWindow: number = oneDaySeconds
+  optimisticRollupFinalizationWindow = 5 * oneDaySeconds,
+  polygonFinalizationWindow = 5 * oneDaySeconds
```
james-a-morris, line 57: Is Polygon often down for more than 1 day? I'm for this because we are already grabbing the last 10 days of data, but I'm curious if this will be more than enough for most outages.

james-a-morris (approved):
LGTM - one question

src/finalizer/index.ts
pxrl: The concern is not about Polygon being down, but the finalizer being down. We have had some cases where an error persisted > 24 hours and that prevented some withdrawals from being finalised. By the time the error was corrected, some had slipped out of the 1 day window.

pxrl:
> Will this substantially increase RPC usage or slow the bot down?

Good question. I haven't quantified it...I expect some increase in RPC consumption, but not anything noticeable in terms of execution time. I'd be surprised if either were substantial.

nicholaspai (approved):

PR #1156, opened 2024-01-23
feat(finalizer): Bi-directional CCTP Finalizations

james-a-morris:
This PR contains the following additions:
- A change to the finalizer to accept non-exclusive L1 execution chains for the multicaller.
- An L1 -> L2 finalization script for CCTP
- An L2 -> L1 finalization script for CCTP

The result of testing produced the following values. All testing was completed between Base Sepolia & ETH Sepolia.

**L1 -> L2**
Transaction to begin the transfer:
https://sepolia.etherscan.io/tx/0x2306f28d2035c4bab2e0abf9027aeb1c93f95929bb56e45ab2fefd53a8639bbe
Multicall/target data that resolved from finalizer script:
```
{
  target: '0x7865fAfC2db2093669d92c0F33AeEF291086BEFD',
  callData: '0x57ecfd280000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000f8000000000000000000000006000000000003ea6f0000000000000000000000009f3b8679c73c2fef8b59b4f3444d4e156fb70aa50000000000000000000000009f3b8679c73c2fef8b59b4f3444d4e156fb70aa50000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001c7d4b196cb0c7b01d743fbc6116a902379c72380000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d00000000000000000000000000000000000000000000000000000000000000010000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d000000000000000000000000000000000000000000000000000000000000000000000000000000820139a8ac53fc0b68456a564b1304731b1014f95d7ffb137f3d0b5f60d478aa4c6d599150c3e61529b6f43977566b72db8aeb987b0b0703f4473e8046b36e41051b5efec5888ee3759213516785976c9f6cf5b03b39eda11d0aa7ef08ad41cc3d6b4074302a17fd5a12a09cd2860f2adc2e6c02b164ff07146c3426a0062e242e911c000000000000000000000000000000000000000000000000000000000000'
}
```
Successful finalizer log:
```
{
  "at": "Finalizer",
  "message": "Finalized unknown deposit on unknown for 1 USDC 🪃",
  "transactionHash": "<unsupported chain/hash 84532:0x98ae88b64a92a707b6127ed69843f3bc1b75273bbc92932d7fc90ffca13be838>}",
  "bot-identifier": "NO_BOT_ID"
}
```
Successful submission of multicall data:
https://sepolia.basescan.org/tx/0x98ae88b64a92a707b6127ed69843f3bc1b75273bbc92932d7fc90ffca13be838

**L2 -> L1**:
Transaction to begin the transfer: https://sepolia.basescan.org/tx/0x86b8c10d5dbcca1bad99437910b99d9835c316cedbe41c71bf2dd6f7a768953e
Simulated Multicall Data from Finalizer:
```
{
  target: '0x7865fAfC2db2093669d92c0F33AeEF291086BEFD',
  callData: '0x57ecfd280000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000f800000000000000060000000000000000000001870000000000000000000000009f3b8679c73c2fef8b59b4f3444d4e156fb70aa50000000000000000000000009f3b8679c73c2fef8b59b4f3444d4e156fb70aa5000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000036cbd53842c5426634e7929541ec2318f3dcf7e0000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d00000000000000000000000000000000000000000000000000000000000000020000000000000000000000009a8f92a830a5cb89a3816e3d267cb7791c16b04d00000000000000000000000000000000000000000000000000000000000000000000000000000082ec741eac799fa72c2983f97ce3d80840efb2c23003302fdc0bdc3b3175a2f01e5fd770d46daa0bc97f54371a4d92e3736ffe3493cdbab0b7608e37ec5e1eaf3d1b97d7e120c517cab44f6dcb1c3dd778bc84e298ac7b0b67c2af83a1999daffe626e488eee4b4680646b93b8e378b5de172ead0cab08e6959ae6a1cb47ae78ec911b000000000000000000000000000000000000000000000000000000000000'
}
```
Successful finalizer log:
```
{
  "at": "Finalizer",
  "message": "Finalized unknown withdrawal on unknown for 2 USDC 🪃",
  "transactionHash": "<unsupported chain/hash 11155111:0x0c864d712e8d8c4e5c76e6f12caef5747bbd6f5073ae374d9f99933c421cc0d4>}",
  "bot-identifier": "NO_BOT_ID"
}
```
Successful submission of multicall data:
https://sepolia.etherscan.io/tx/0x0c864d712e8d8c4e5c76e6f12caef5747bbd6f5073ae374d9f99933c421cc0d4

---

**Note**: all testing has been completed on testnet with a functioning Finalizer. 

Closes ACX-1816 & ACX-1815


src/clients/HubPoolClient.ts
```diff
@@ -46,4 +46,9 @@ export class HubPoolClient extends clients.HubPoolClient {
 
     return await super.computeRealizedLpFeePct(deposit);
   }
+
+  // TODO: Replace when the SDK is bumped
```
nicholaspai, line 50: Can you `yarn link` locally and skip this change?

src/common/Constants.ts
```diff
@@ -225,3 +225,19 @@ export type Multicall2Call = {
 // These are the spokes that can hold both ETH and WETH, so they should be added together when caclulating whether
 // a bundle execution is possible with the funds in the pool.
 export const spokesThatHoldEthAndWeth = [10, 324, 8453];
+
+export const chainIdsToCctpDomains: { [chainId: number]: number } = {
```
nicholaspai, line 229: Can you add a source for these domains so we can update in future?

src/finalizer/index.ts
```diff
@@ -40,14 +46,16 @@ let logger: winston.Logger;
 const oneDaySeconds = 24 * 60 * 60;
 const oneHourSeconds = 60 * 60;
 
-const chainFinalizers: { [chainId: number]: ChainFinalizer } = {
```
nicholaspai, line 43: I think it might be a bit cleaner to leave the `chainFinalizers` as is and then define an `overrideChainFinalizers: { [chainId]: { [token] } }` to make it clear only certain token-chain combos should be finalized specially

src/finalizer/types.ts
```diff
@@ -8,6 +8,7 @@ export type Withdrawal = {
   l1TokenSymbol: string;
   amount: string;
   type: "proof" | "withdrawal";
+  executionChainId: number;
```
nicholaspai, line 11: do you think we should define a separate `Deposit`... that is clearly a reversal of the Withdrawal? I suggest because its kind of weird to have an `l2ChainId` property here and also an `executionChainId`. What does this mean for an L1 to L2 txn we need to finalize? Is L2ChainId == L1 in that case?

Or, we could rename `l2ChainId` to `originationChainId`/`originChainId`/`initiationChainId` to pair with `executionChainId`/`destinationChainId` but I don't want to conflate origin/destination with the core Deposit/Fill keywords.

src/finalizer/utils/cctp.ts
```diff
@@ -0,0 +1,295 @@
...
+async function _resolveCCTPRelatedTxn(
+  receipt: TransactionReceipt,
+  currentChainId: number,
+  targetDestinationChainId: number
+): Promise<DecodedCCTPMessage> {
+  // We need the Event[0] topic to be the MessageSent event.
+  // Note: we could use an interface here but we only need the topic
+  //       and not the entire contract.
+  const cctpEventTopic = ethers.utils.id("MessageSent(bytes)");
+  const log = receipt.logs.find(
```
nicholaspai, line 172: What if there are multiple MessageSent log in the same txn hash? This could happen if we're sending multiple CCTP deposits from the Hub to the Spoke from multiple PoolRebalanceLeaf executed atomically. This I think is possible because we execute all the PoolRebalanceLeaves via Multisender

You might need to do something like this: https://github.com/across-protocol/relayer-v2/blob/master/src/finalizer/utils/arbitrum.ts#L135

src/finalizer/utils/cctp.ts
```diff
@@ -0,0 +1,295 @@
...
+  // We need to decode the message bytes from the log data
+  const messageBytes = ethers.utils.defaultAbiCoder.decode(["bytes"], log.data)[0];
+  // We need to further decode the message bytes to get the sourceDomain, destinationDomain, nonce, and amount
+  const messageBytesArray = ethers.utils.arrayify(messageBytes);
+  const sourceDomain = ethers.utils.hexlify(messageBytesArray.slice(4, 8)); // sourceDomain 4 bytes starting index 4
+  const destinationDomain = ethers.utils.hexlify(messageBytesArray.slice(8, 12)); // destinationDomain 4 bytes starting index 8
+  const nonce = ethers.utils.hexlify(messageBytesArray.slice(12, 20)); // nonce 8 bytes starting index 12
+  const nonceHash = ethers.utils.solidityKeccak256(["uint32", "uint64"], [sourceDomain, nonce]);
+  const messageHash = ethers.utils.keccak256(messageBytes);
+  const amountSent = ethers.utils.hexlify(messageBytesArray.slice(184, 216)); // amount 32 bytes starting index 216 (idx 68 of body after idx 116 which ends the header)
```
nicholaspai, line 189: Is this how CCTP recommends we decode their messages..? Seems convoluted. Is there an ABI we can use to decode or would that just return the `bytes` that there is a special format we need to parse further

james-a-morris: The event is `MessageSent(bytes)`. I haven't found an official SDK of Circle's contracts/event types nor do I believe they have released one. 

They break down the message format here: https://developers.circle.com/stablecoins/docs/message-format

I can add a comment explaining where I found the information to retrieve this.

nicholaspai (commented):
I'm still not convinced that the l1 to l2 `cctp` functions shouldn't be in a new directory `l1ToL2Finalizer` to avoid comingling the existing finalizer code with this other direction code. It doesn't look very clean to me at first glance. Moreover, the `l1ToL2Finalizer` doesn't even use the `SpokePoolClient` functions and the `l2ToL1Finalizer` doesn't use the `HubPoolClient` functions (outside of `chainId`.)

Also, nit but you need to handle multiple logs in the same txn receipt for both directions

src/finalizer/utils/cctp.ts
nicholaspai: a comment would suffice

src/finalizer/index.ts
```diff
@@ -106,22 +113,36 @@ export async function finalize(
...
-    const { callData, withdrawals } = await chainFinalizer(
-      logger,
-      hubSigner,
-      hubPoolClient,
-      client,
-      latestBlockToFinalize
-    );
+    let totalWithdrawalsForChain = 0;
+
+    for (const finalizer of chainFinalizer) {
```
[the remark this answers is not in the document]
james-a-morris: I agree that eventually, we'd want to separate this on tokens as well, i.e., USDC/Non-USDC. But yes, this change would mean that certain chains will have multiple finalizers for USDC until we deploy CCTP and stop bridging USDC through the canonical bridges.

src/finalizer/utils/cctp.ts
james-a-morris: Good call. This is now completed. I've left an explicit log about how this can be the case (and how it is accounted for): https://github.com/across-protocol/relayer-v2/blob/750ca0da3a09e813e247bb69c8a062de6d421c6a/src/utils/CCTPUtils.ts#L31-L39

src/finalizer/types.ts
james-a-morris: Absolutely agree - I've modified `Withdrawal` to be `CrossChainTransfer`. It now contains 3 types: `proof`, `deposit`, `withdrawal`.

src/finalizer/index.ts
james-a-morris: Good call

src/clients/MultiCallerClient.ts
```diff
@@ -98,11 +98,11 @@ export class MultiCallerClient {
     }
   }
 
-  async executeTransactionQueue(simulate = false): Promise<string[]> {
-    // For compatibility with the existing implementation, flatten all txn hashes into a single array.
-    // To be resolved once the legacy implementation is removed and the callers have been updated.
-    const txnHashes: { [chainId: number]: string[] } = await this.executeTxnQueues(simulate);
-    return Object.values(txnHashes).flat();
+  async executeTransactionQueue(simulate = false): Promise<Record<number, string>> {
```
nicholaspai, line 101: MultiCallerClient.ts test file should fail with this change, and probably some of the Relayer unit tests. Just based on a quick look up of this function and where we assume the shape of the returned value

src/clients/MultiCallerClient.ts
```diff
@@ -98,11 +98,11 @@ export class MultiCallerClient {
...
+    // For compatibility with the existing implementation, only resolve the first
+    // txn of each chain.
+    const txnHashesByChain = await this.executeTxnQueues(simulate);
+    return Object.fromEntries(Object.entries(txnHashesByChain).map(([chainId, txnHashes]) => [chainId, txnHashes[0]]));
```
nicholaspai, line 105: actually, what is going on here? My danger instincts are going off with this change

src/finalizer/index.ts
```diff
@@ -245,7 +310,8 @@ export async function constructFinalizerClients(
     commonClients.configStoreClient,
     config,
     baseSigner,
-    config.maxFinalizerLookback
+    config.maxFinalizerLookback,
+    [11155111, 84532]
```
nicholaspai, line 314: should not commit this

src/finalizer/index.ts
```diff
@@ -106,22 +131,50 @@ export async function finalize(
...
+        finalizationsToBatch.push({ txn, crossChainTransfer: crossChainTransfers[idx] });
+      });
+
+      totalWithdrawalsForChain += crossChainTransfers.filter(({ type }) => type === "withdrawal").length;
+      totalDepositsForChain += crossChainTransfers.filter(({ type }) => type === "deposit").length;
+    }
     logger.debug({
       at: "finalize",
-      message: `Found ${withdrawals.length} ${network} withdrawals for finalization.`,
+      message: `Found ${totalWithdrawalsForChain} ${network} transfers (${totalWithdrawalsForChain} withdrawals | ${totalDepositsForChain} deposits ) for finalization.`,
```
nicholaspai, line 160: Should we not also be logging the potential proofs to be sent?

src/utils/CCTPUtils.ts
```diff
@@ -0,0 +1,168 @@
...
+  destinationDomain: number;
+  attestation: string;
+};
+
+/**
+ * Used to map a CCTP domain to a chain id. This is the inverse of chainIdsToCctpDomains.
+ * Note: due to the nature of testnet/mainnet chain ids mapping to the same CCTP domain, we
+ *       actually have a mapping of CCTP Domain -> [chainId].
+ */
+const invertedCctpDomainsToChainIds = Object.entries(chainIdsToCctpDomains).reduce((acc, [chainId, cctpDomain]) => {
```
nicholaspai, line 23: why name this "inverted" when its implied from the naming "cctpDomainsToChainIds"?

nicholaspai (commented):
Some driveby comments

src/finalizer/index.ts
james-a-morris: Good call. Logging it now: https://github.com/across-protocol/relayer-v2/pull/1156/commits/43a2f5a663247db18663552051289bbd1913939d

src/finalizer/index.ts
james-a-morris: Agreed. https://github.com/across-protocol/relayer-v2/pull/1156/commits/8a0087dcd899e57ab24e1c24cb1bb5e9574275a4

src/clients/MultiCallerClient.ts
james-a-morris: This was meant to return the hashes of multicaller transactions by chain. The reason that I changed this logic was because the relayer code only read this value in the finalizer. I've since changed the call to `executeTransactionQueues()` directly so that I can grab the txn hashes by chain. 

It was stashed while I was trying to remove all the test code. 😞 https://github.com/across-protocol/relayer-v2/pull/1156/commits/ba2412c575b916a815ec090b1fdf39f0149157cb

I'd like to circle back to the

src/clients/MultiCallerClient.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1156#discussion_r1468272927

james-a-morris: The most recent change was stashed. It's now reflected in the repo

src/utils/CCTPUtils.ts
james-a-morris: Done. https://github.com/across-protocol/relayer-v2/pull/1156/commits/9b52e03b43f1d5fb18b765cf7edca6d719872d78

I was trying to be explicit but that's fair it's already in the variable name.

nicholaspai (approved):
This looks great, great work

PR #1157, opened 2024-01-24
refactor: Dataworker-specific preparation for v3

pxrl:
This commit introduces forwards-compatible changes to the relayer-v2 repository that will ease the pending types upgrade for Across v3.

No change in behaviour is intended as a result of the change, but it should de-risk the eventual diff by de-cluttering it.


src/clients/InventoryClient.ts
```diff
@@ -185,15 +186,23 @@ export class InventoryClient {
...
       return destinationChainId;
     }
 
-    l1Token ??= this.hubPoolClient.getL1TokenForL2TokenAtBlock(originToken, originChainId);
+    const inputToken = sdkUtils.getDepositInputToken(deposit);
+    const outputToken = sdkUtils.getDepositOutputToken(deposit);
+
+    // The InventoryClient assumes 1:1 equivalency between input and output tokens. If the tokens are not equivalent
+    // then just take refunds on the destination chain. The operator is responsible for balancing inventory.
+    if (!this.hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)) {
```
pxrl, line 202: This is technically a change in behaviour, however it's only theoretical since we've never supported native swaps. I also don't anticipate it being supported in the relayer in the near-term future, but it was important to identify the potential here and ensure that it's handled safely, because the existing logic would otherwise produce undesirable behaviour.

src/clients/ProfitClient.ts
```diff
@@ -203,8 +203,9 @@ export class ProfitClient {
   }
 
   // @todo: Factor in the gas cost of submitting the RefundRequest on alt refund chains.
-  async getTotalGasCost(deposit: Deposit, fillAmount = deposit.amount): Promise<TransactionCostEstimate> {
+  async getTotalGasCost(deposit: Deposit, fillAmount?: BigNumber): Promise<TransactionCostEstimate> {
```
pxrl, line 206: Adjusted because v3Deposits do not have `deposit.amount`, so we need to resolve it dynamically via the `getDepositOutputAmount()` helper. No change in functionality, though.

src/clients/ProfitClient.ts
```diff
@@ -367,8 +370,9 @@ export class ProfitClient {
   getFillAmountInUsd(deposit: Deposit, fillAmount: BigNumber): BigNumber {
     const l1TokenInfo = this.hubPoolClient.getTokenInfoForDeposit(deposit);
     if (!l1TokenInfo) {
+      const inputToken = sdkUtils.getDepositInputToken(deposit);
```
pxrl, line 373: Note: `HubPoolClient.getTokenInfoForDeposit()` selects `originToken`.  This function should probably switch on the type of Deposit (`v2Deposit` | `v3Deposit`), and if `v3Deposit` it should use the outputToken instead.

Second headache: If this should have theoretical support for arbitrary output tokens, `outputToken` might not be represented on mainnet, so the PriceClient would need to support querying arbitrary tokens outside of mainnet. So for now I guess proceed with `inputToken` and identify this update as a potential roadmap item.

src/clients/TokenClient.ts
```diff
@@ -61,11 +62,13 @@ export class TokenClient {
   }
 
   hasBalanceForFill(deposit: Deposit, fillAmount: BigNumber): boolean {
-    return this.getBalance(deposit.destinationChainId, deposit.destinationToken).gte(fillAmount);
+    const outputToken = sdkUtils.getDepositOutputToken(deposit);
+    return this.getBalance(deposit.destinationChainId, outputToken).gte(fillAmount);
   }
 
   hasBalanceForZeroFill(deposit: Deposit): boolean {
```
pxrl, line 69: This is technically not required because we won't zero fill in v3, but using the new utility functions keeps tsc from complaining about the possibility of hitting a v3  deposit without `destinationToken` here.

src/dataworker/Dataworker.ts
```diff
@@ -742,11 +744,32 @@ export class Dataworker {
...
-          // token should be repaid in.
+          // All flows in here are assumed to be valid, so we can use the flow's repayment chain
+          // to pay out the refund. But we need to check which token should be repaid in.
           if (isUbaOutflow(flow)) {
-            const refundToken = outflowIsFill(flow) ? flow.destinationToken : flow.refundToken;
+            let refundToken: string;
+
+            if (outflowIsFill(flow)) {
+              if (sdkUtils.isV2Fill(flow)) {
+                refundToken = flow.destinationToken;
```
pxrl, line 754: This is just tsc-appeasement since we won't be concerned about overlap between v2 and the UBA.

src/dataworker/Dataworker.ts
```diff
@@ -742,11 +744,32 @@ export class Dataworker {
...
+              } else {
+                // If the fill requests a refund on the destination chain, resolve the refundToken from the inputToken.
+                // This handles refunds where the outputToken is not equivalent to the inputToken (i.e. swaps).
+                const { hubPoolClient } = this.clients;
+                const { originChainId, destinationChainId, inputToken: originToken } = flow;
+
+                // @todo: Should resolve this more precisely.
+                const quoteBlockNumber = hubPoolClient.latestBlockSearched;
+                refundToken = hubPoolClient.getL2TokenForDeposit({
```
pxrl, line 763: fwiw, this translates `inputToken` on `originChainId` to its corresponding representation on `destinationChainId`. This is necessary because we don't have any guarantees of equivalency between `inputToken` and `outputToken`, and it's basically just as difficult to verify whether they are equivalent as it is to just resolve it directly from `inputToken`.

src/dataworker/Dataworker.ts
```diff
@@ -742,11 +744,32 @@ export class Dataworker {
...
+            if (outflowIsFill(flow)) {
+              if (sdkUtils.isV2Fill(flow)) {
+                refundToken = flow.destinationToken;
+              } else {
+                // If the fill requests a refund on the destination chain, resolve the refundToken from the inputToken.
+                // This handles refunds where the outputToken is not equivalent to the inputToken (i.e. swaps).
+                const { hubPoolClient } = this.clients;
+                const { originChainId, destinationChainId, inputToken: originToken } = flow;
+
+                // @todo: Should resolve this more precisely.
```
pxrl, line 761:
```suggestion
                // @todo: Should resolve quoteBlockNumber on mainnet more precisely.
```

src/dataworker/Dataworker.ts
```diff
@@ -782,10 +805,8 @@ export class Dataworker {
           destinationChainId,
           this.chainIdListForBundleEvaluationBlockNumbers
         );
-        const tokenSymbol = this.clients.hubPoolClient.getL1TokenInfoForL2Token(
-          deposit.destinationToken,
-          destinationChainId
-        ).symbol;
+        const outputToken = sdkUtils.getDepositOutputToken(deposit);
+        const tokenSymbol = this.clients.hubPoolClient.getL1TokenInfoForL2Token(outputToken, destinationChainId).symbol;
```
pxrl, line 809: nb. we will need filtering on `unfilledDeposits[]` to ensure that we only initiate slow fills for deposits where `inputToken` is equivalent to `outputToken`.

I think the best place to implement this is just to verify it in the SpokePoolClient at the point where slow fill requests are parsed - i.e. we should discard any slow fill requests where the tokens are not equivalent.

src/dataworker/Dataworker.ts
```diff
@@ -1415,6 +1436,9 @@ export class Dataworker {
     // Ignore slow fill leaves for deposits with messages as these messages might be very expensive to execute.
     // The original depositor can always execute these and pay for the gas themselves.
     const leaves = _leaves.filter(({ payoutAdjustmentPct, relayData }) => {
```
pxrl, line 1438: This will probably be refactored later on, such that we enforce `v2SlowFillLeaf` here so that the utility functions are not needed. That's OK since we're relying on contract ABIs here, so fine-grained discrimination of types is implied.

src/dataworker/Dataworker.ts
```diff
@@ -1490,23 +1514,26 @@ export class Dataworker {
...
 
-          // Check if fill was a full fill. If so, execution is unnecessary.
-          if (fill && fill.totalFilledAmount.eq(fill.amount)) {
-            return undefined;
-          }
-
           // If the most recent fill is not found, just make the most conservative assumption: a 0-sized fill.
-          const amountFilled = fill ? fill.totalFilledAmount : BigNumber.from(0);
+          let amountFilled = bnZero;
+          if (isDefined(fill)) {
```
pxrl, line 1519: To be explicit, this could probably check for `isV3Fill(fill)` directly in here, since we know that any valid V3Fill will implicitly be a a full fill. That might give tsc some hints that any remaining `fill` objects are v2-only.

pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/1157/commits/1a7f7b2181a446c00fa7e50ed3fafbbea789a3c0. This allows some of the diff to be reverted.

src/dataworker/Dataworker.ts
```diff
@@ -1423,8 +1447,8 @@ export class Dataworker {
           leafExecutionArgs: [
```
pxrl, line 1447: `leafExecutionArgs` is dependent on whether it's a v2 or v3 slow fill, so this might need to be refactored a little bit. However, it's also not clear whether we need to log all this information anyway - we could probably get away by just logging the originChainId, depositId, and message.

src/dataworker/Dataworker.ts
```diff
@@ -1534,13 +1560,16 @@ export class Dataworker {
...
         `rootBundleId: ${rootBundleId}\n` +
         `slowRelayRoot: ${slowRelayTree.getHexRoot()}\n` +
         `Origin chain: ${relayData.originChainId}\n` +
         `Destination chain:${relayData.destinationChainId}\n` +
         `Deposit Id: ${relayData.depositId}\n` +
-        `amount: ${relayData.amount.toString()}`;
+        `amount: ${outputAmount.toString()}`;
 
       if (submitExecution) {
         this.clients.multiCallerClient.enqueueTransaction({
```
pxrl, line 1575: This is again ABI-dependent, so it's safe to assume we can narrow the relayData type to v2 or v3. This should probably be revisited later.

src/dataworker/DataworkerUtils.ts
```diff
@@ -151,8 +151,8 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
       relayData: {
```
pxrl, line 151: This will need to be revisited in a follow-up change because the form of `relayData` is different for v2 vs v3.

src/dataworker/Dataworker.ts
```diff
@@ -1471,8 +1495,8 @@ export class Dataworker {
           fill.originChainId === relayData.originChainId &&
```
pxrl, line 1495: nb. Since we're comparing against relayData here, we'll need a type-specific comparison in a follow-up commit.

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -620,20 +627,16 @@ export function generateMarkdownForRootBundle(
 
   let slowRelayLeavesPretty = "";
   slowRelayLeaves.forEach((leaf, index) => {
```
pxrl, line 629: This is another section where we probably need separate implementations to handle v2 and v3 individually.

src/relayer/Relayer.ts
```diff
@@ -690,11 +693,12 @@ export class Relayer {
...
     assert(isDefined(ubaClient), "No ubaClient");
 
-    const tokenSymbol = hubPoolClient.getL1TokenInfoForL2Token(deposit.originToken, deposit.originChainId)?.symbol;
+    const inputToken = sdkUtils.getDepositInputToken(deposit);
+    const tokenSymbol = hubPoolClient.getL1TokenInfoForL2Token(inputToken, deposit.originChainId)?.symbol;
     const relayerBalancingFee = ubaClient.computeBalancingFeeForNextRefund(
       deposit.destinationChainId,
       tokenSymbol,
-      deposit.amount
+      sdkUtils.getDepositInputAmount(deposit)
```
pxrl, line 701: Not super relevant since UBA pricing is not active, but the theoretical balancing fee should probably be `inputAmount.sub(realizedLpFeePct)`.

test/Relayer.RefundRequests.ts
pxrl, line 1: 2 test cases were disabled here - they relate to UBA RefundRequest event processing. It's not really possible to handle these correctly at the moment since v2 RefundRequests will never happen, but v3 RefundRequests aren't actually possible at this point in time.

src/clients/InventoryClient.ts
nicholaspai: I think we should use this function in the relayer and make sure to skip any deposits where tokens are not equivalent. Thoughts?

src/clients/ProfitClient.ts
nicholaspai: I think all of the changes we roll out with at first should assume output token and input are equivalent and if they are not we should throw an error or exit prematurely since we can't handle them yet

src/clients/InventoryClient.ts
pxrl: 100% agree. It's not part of this change, but it's definitely part of the relayer-specific upgrades because we have many areas where we're not robust to a different outputToken, and we also lack the ability to express it in configuration.

src/dataworker/Dataworker.ts
nicholaspai: We might just want to remove some uba logic and have the entire dataworker throw if uba is activated for now since it's hard to reason about whether this code will ever get touched if the uba isn't active

src/dataworker/Dataworker.ts
nicholaspai: We could do that. Or we could do it in the dataworker. If we do it in the spoke then we should expose a function that performs the filter rather than refusing to store events into memory

src/dataworker/Dataworker.ts
```diff
@@ -1490,13 +1514,15 @@ export class Dataworker {
...
-          }
-
           // If the most recent fill is not found, just make the most conservative assumption: a 0-sized fill.
-          const amountFilled = fill ? fill.totalFilledAmount : BigNumber.from(0);
+          let amountFilled = bnZero;
+          if (isDefined(fill)) {
+            // If fill was a full fill, execution is unnecessary.
+            amountFilled = sdkUtils.getTotalFilledAmount(fill);
+            if (amountFilled.eq(sdkUtils.getFillOutputAmount(fill)) || sdkUtils.isV3Fill(fill)) {
+              return undefined;
```
nicholaspai, line 1523: We'll have to replace this in a follow up?

src/relayer/Relayer.ts
```diff
@@ -283,7 +277,7 @@ export class Relayer {
         continue;
       }
 
-      const l1Token = hubPoolClient.getL1TokenInfoForL2Token(originToken, originChainId);
+      const l1Token = hubPoolClient.getL1TokenInfoForL2Token(inputToken, originChainId);
```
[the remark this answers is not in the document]
nicholaspai: Agreed. For now we should skip them and log them as a warning since no one should be sending such deposits yet

src/relayer/Relayer.ts
```diff
@@ -581,14 +577,19 @@ export class Relayer {
     return fills;
   }
 
+  // @note RefundRequests can be issued in v2, but will never be acted on. They
```
nicholaspai, line 580: refund requests will be impossible in v3. Just like most of the uba functions. So maybe we should just remove

src/dataworker/Dataworker.ts
pxrl: I think in a v3 world it should be entirely unnecessary, since this is checking for initial partial fills to compute the outstanding balance to send in a slow fill. The `isV3Fill()` just helps tsc to know that any subsequent fill objects will be of type `v2FillWithBlock`, so it's therefore OK to dereference v2 fields directly.

So I think the best solution is to ensure that v2 fills are the only ones that this function ever has to deal with - i.e. partitioning it off at a higher level, and updating the function prototype to require that only `v2FillWithBlock` events are supplied.

src/relayer/Relayer.ts
pxrl: They're obviously not part of the current SpokePool contract, but if we were to roll the UBA out then we'd obviously need to re-add them right?

If we think it's unlikely (or at least distant) then I'm totally onboard with removing this code - I just didn't want to remove stuff that we'd end up re-adding later.

src/clients/ProfitClient.ts
nicholaspai: > This function should probably switch on the type of Deposit (v2Deposit | v3Deposit), and if v3Deposit it should use the outputToken instead.

I agree with this. Its going to ease eventual support for cross-token swaps if we grab the token info for the outputToken

As to the second headache about PriceClient not supporting `outputToken`, will need to add a `catch` statement after the `getPriceOfToken` call below?

src/dataworker/Dataworker.ts
nicholaspai: I do think that by the time you get into the Dataworker `buildSlowLeaf()`-esque functions, you can assume that the `unfilledDeposits[]` passed as input will already have filtered on valid slow fills

src/dataworker/Dataworker.ts
nicholaspai: ah good point. In V3, if the leaf exists in `leavesWithLatestFills` by definition, the slow fill leaf has been replaced by a fast fill

src/relayer/Relayer.ts
nicholaspai: We probably shouldn't remove in this PR to reduce diff but yeah we do plan on removing the `requestRepayment` function and would consider adding it again in the future but it might be very different since the fill/deposit interfaces have changed so much. The existing function just doens't work cleanly for V3

src/relayer/Relayer.ts
nicholaspai: hmm yeah true... worth replacing here?

nicholaspai: or at least leaving a comment but regardless we shouldn't hit this code

src/utils/FillMathUtils.ts
```diff
@@ -11,6 +14,7 @@ export function _getFeeAmount(fillAmount: BigNumber, feePct: BigNumber): BigNumb
 }
 
 export function _getRealizedLpFeeForFill(fill: Fill): BigNumber {
+  assert(sdkUtils.isV2Fill(fill));
```
nicholaspai, line 17: +1

src/relayer/Relayer.ts
pxrl: I bumped the removal up and got it into master, so this is now a non-issue: https://github.com/across-protocol/relayer-v2/pull/1157/commits/8aab1f86be632c00fc3001a838dfa7219b52751f.

test/Relayer.RefundRequests.ts
pxrl: The entire file was ultimately removed: https://github.com/across-protocol/relayer-v2/pull/1157/commits/8aab1f86be632c00fc3001a838dfa7219b52751f.

test/Relayer.RefundRequests.ts
```diff
@@ -283,7 +285,6 @@ describe("Relayer: Request refunds for cross-chain repayments", async function (
     expect(refundRequest.transactionHash).to.equal(transactionHash);
   });
 
-  // @note Temporarily disabled, pending upstream fixes in the SDK.
```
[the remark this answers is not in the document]
pxrl: Nuked: https://github.com/across-protocol/relayer-v2/pull/1157/commits/8aab1f86be632c00fc3001a838dfa7219b52751f.

src/dataworker/DataworkerUtils.ts
```diff
@@ -146,22 +149,8 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
...
-        originChainId: deposit.deposit.originChainId,
-        destinationChainId: deposit.deposit.destinationChainId,
-        realizedLpFeePct: deposit.deposit.realizedLpFeePct,
-        relayerFeePct: deposit.deposit.relayerFeePct,
-        depositId: deposit.deposit.depositId,
-        message: deposit.deposit.message,
-      },
-      payoutAdjustmentPct: deposit?.relayerBalancingFee?.toString() ?? "0",
-    })
+  const slowRelayLeaves: SlowFillLeaf[] = unfilledDeposits.map((deposit: UnfilledDeposit) =>
```
pxrl, line 152: To be explciit, this array consists of mixed v2 and v3 slow fill leaves. Sorting is as per v2 (based on originChainId and deposit Id), so they can be interleaved. The hash function is aware of both v2 and v3 leaves and it applies the correct format based on what it receives at runtime.

src/dataworker/DataworkerUtils.ts
```diff
@@ -181,6 +170,51 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
...
+    },
+    payoutAdjustmentPct: unfilledDeposit.relayerBalancingFee?.toString() ?? "0",
+  };
+}
+
+function buildV3SlowFillLeaf(unfilledDeposit: UnfilledDeposit): v3SlowFillLeaf {
+  const { deposit } = unfilledDeposit;
+  assert(utils.isV3Deposit(deposit));
+
+  const updatedOutputAmount = deposit.updatedOutputAmount ?? deposit.outputAmount;
```
pxrl, line 198: Note: This is not correct - `updatedOutputAmount` in this case is determined by the proposer, not the depositor. The proposer will set this to `deposit.outputAmount * (1 - realizedLpFeePct)`  in order to zero the relayer fee, which is effectively as per v2 for slow fills.

src/dataworker/DataworkerUtils.ts
pxrl: Fixed here: https://github.com/across-protocol/relayer-v2/pull/1157/commits/d74b4c929f090c354d93a36afc3b46313a1dce90.

src/dataworker/Dataworker.ts
```diff
@@ -743,12 +744,31 @@ export class Dataworker {
...
         flowsForChain.forEach(({ flow, balancingFee }) => {
-          // All flows in here are assumed to be valid, so we can use the flow's
-          // repayment chain to pay out the refund. But we need to check which
-          // token should be repaid in.
+          // All flows in here are assumed to be valid, so we can use the flow's repayment chain
+          // to pay out the refund. But we need to check which token should be repaid in.
           if (isUbaOutflow(flow)) {
-            assert(outflowIsFill(flow));
-            const refundToken = flow.destinationToken;
+            let refundToken: string;
```
james-a-morris, line 750: OOC will the UBA code paths be phased out?

src/dataworker/Dataworker.ts
```diff
@@ -1270,7 +1288,7 @@ export class Dataworker {
           message: `Evaluating ${rootBundleRelays.length} historical non-empty slow roots relayed to chain ${chainId}`,
         });
 
-        const slowFillsForChain = client.getFills().filter((fill) => fill.updatableRelayData.isSlowRelay);
+        const slowFillsForChain = client.getFills().filter((fill) => sdkUtils.isSlowFill(fill));
```
james-a-morris, line 1291:
```suggestion
        const slowFillsForChain = client.getFills().filter(sdkUtils.isSlowFill);
```

src/dataworker/DataworkerUtils.ts
```diff
@@ -181,6 +170,51 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
   };
 }
 
+function buildV2SlowFillLeaf(unfilledDeposit: UnfilledDeposit): v2SlowFillLeaf {
+  const { deposit } = unfilledDeposit;
+  assert(utils.isV2Deposit(deposit));
```
james-a-morris, line 175: Thoughts on adding a new `Error` class for `NotV2Deposit` or `NotV3Deposit`? Or, adding text to this assert so that it's clear what's going on if this exception is thrown?

james-a-morris (commented):
Left some comments. Looks great

src/dataworker/DataworkerUtils.ts
nicholaspai: I think we can keep the sorting algorithm the same and apply it to intermixed v2 and v3 leaves, right

src/dataworker/DataworkerUtils.ts
nicholaspai: Do you think we should set the `updatedOutputAmount` correctly here? I would imagine it gets set in `buildSlowRelayTree` which calls into here

src/dataworker/Dataworker.ts
nicholaspai: yes

src/dataworker/DataworkerUtils.ts
pxrl: Yep, agree. It needs to be set upstream; it's only set here temporarily to appease tsc. It'll be fixed as part of this PR.

pxrl: I realised that the `v3SlowFillLeaf` type includes `updatedOutputAmount` directly, so we can just inherit directly from that. Marking this issue as resolved now.

src/dataworker/DataworkerUtils.ts
pxrl: Still TBD whether we can mix the types, but I think the proposed implementation is OK in either case, since in the worst case we can control it at an upper layer by requiring `slowFillLeaves.every(sdkUtils.isV2SlowFillLeaf)` or `slowFillLeaves.every(sdkUtils.isV3SlowFillLeaf)` accordingly.

src/dataworker/Dataworker.ts
pxrl: The implementation has been updated in https://github.com/across-protocol/relayer-v2/pull/1157/commits/847953523985a6e197a655d2935f459f1f3bbdb6. I'm pretty happy with where it's at now.

src/dataworker/Dataworker.ts
pxrl: I updated this PR with type-specific checks. See https://github.com/across-protocol/relayer-v2/pull/1157/commits/847953523985a6e197a655d2935f459f1f3bbdb6.

src/dataworker/Dataworker.ts
pxrl: Has been updated here - so this can be marked resolved now: https://github.com/across-protocol/relayer-v2/pull/1157/commits/847953523985a6e197a655d2935f459f1f3bbdb6.

src/dataworker/Dataworker.ts
pxrl: Updated here - should be OK now: https://github.com/across-protocol/relayer-v2/pull/1157/commits/847953523985a6e197a655d2935f459f1f3bbdb6.

src/dataworker/Dataworker.ts
pxrl: Has been refactored in this PR - see this commit: https://github.com/across-protocol/relayer-v2/pull/1157/commits/847953523985a6e197a655d2935f459f1f3bbdb6.

src/dataworker/DataworkerUtils.ts
nicholaspai: I think we can mix the types and should proceed with implementing them intermixed. Sorry for the mixup

src/clients/ProfitClient.ts
pxrl: For now I think we just filter the input deposit events that are fed to the ProfitClient, such that we ensure input and output tokens are equivalent. That seems simplest - otherwise we need to start making everything robust to disparate output tokens.

src/clients/TokenClient.ts
pxrl: I considered being more strict here and enforcing type `v2Deposit`, but it introduces additional branching to the relayer code. There's not much reason to do this other than to be slightly more correct, since this code will ultimately be removed.

src/dataworker/Dataworker.ts
pxrl: Has been removed in master and the updates were merged in here: https://github.com/across-protocol/relayer-v2/pull/1157/commits/9cb4edba0f54b982d4ae2e755c3adf1c0ce899ea

src/relayer/Relayer.ts
pxrl: This is no longer relevant because the UBA functionality has been flushed per this commit: https://github.com/across-protocol/relayer-v2/pull/1157/commits/9cb4edba0f54b982d4ae2e755c3adf1c0ce899ea.

src/dataworker/Dataworker.ts
pxrl: UBA functionality removed: https://github.com/across-protocol/relayer-v2/pull/1157/commits/9cb4edba0f54b982d4ae2e755c3adf1c0ce899ea

src/dataworker/Dataworker.ts
pxrl: UBA removed: https://github.com/across-protocol/relayer-v2/pull/1157/commits/9cb4edba0f54b982d4ae2e755c3adf1c0ce899ea

src/dataworker/DataworkerUtils.ts
pxrl: @james-a-morris These asserts are only intended to get us over the v2 -> v3 hurdle, where we have a brief period of supporting both event types concurrently. My gut feel is that it's probably overkill to do anything more than a simple assert.

In terms of a text string, what would you propose? We could add something like "Not a v2 deposit", but I think that's already self-evident from the assert statement. So I'm not really sure what we'd add without having to dump a lot of additional context. wdyt?

src/clients/ProfitClient.ts
pxrl: I've added a token equivalency check to the filtering that we apply on unfilled deposits. This is at home with all of the other filtering that we apply. https://github.com/across-protocol/relayer-v2/pull/1157/commits/7219e21c492c202a65c46a42a133bc0285b39348

src/dataworker/PoolRebalanceUtils.ts
pxrl: Addressed here: https://github.com/across-protocol/relayer-v2/pull/1157/commits/07310bc9a202151f47d2b639193194943a0217f6

This section of code dynamically updated the slow fill leaves by adding new members and deleting some existing ones. It did this in order to make them printable into markdown. I've instead opted to create a new object that contains only the members that we want, and it avoids modifying the underlying/input object.

As part of this, I dropped the `relayerFeePct` field since it's not actually relevant to a slow fill - i.e. we drop it to 0, so printing the deposit-time value is arguably a bit misleading.

src/relayer/Relayer.ts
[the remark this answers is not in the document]
pxrl: Per https://github.com/across-protocol/relayer-v2/pull/1157#discussion_r1472294180, deposits with disparate output tokens are filtered out quite early in the relayer loop now. That allows us to assume that inputToken and outputToken are equivalent here. We'll need to circle back later and revise this once v2 is removed.

src/dataworker/Dataworker.ts
```diff
@@ -1127,50 +1127,73 @@ export class Dataworker {
...
+        leavesWithLatestFills.map(async (slowFill) => {
+          const destinationChainId = sdkUtils.getSlowFillLeafChainId(slowFill);
+          if (destinationChainId !== chainId) {
+            throw new Error(`Leaf chainId does not match input chainId (${destinationChainId} != ${chainId})`);
           }
 
           // If the most recent fill is not found, just make the most conservative assumption: a 0-sized fill.
-          const amountFilled = fill ? fill.totalFilledAmount : BigNumber.from(0);
+          let amountFilled = bnZero;
+          const fill = slowFill?.fill;
```
mrice32, line 1179: nit: why the ?

Don't we know that slowFill is defined?

src/dataworker/Dataworker.ts
```diff
@@ -1181,65 +1204,106 @@ export class Dataworker {
...
-          ],
+          chainId,
+          method,
+          args,
           message: "Executed SlowRelayLeaf 🌿!",
           mrkdwn,
           // If mainnet, send through Multicall3 so it can be batched with PoolRebalanceLeaf executions, otherwise
           // SpokePool.multicall() is fine.
-          unpermissioned: Number(chainId) === 1,
+          unpermissioned: chainId === hubChainId,
```
mrice32, line 1254: +1

src/dataworker/Dataworker.ts
```diff
@@ -1181,65 +1204,106 @@ export class Dataworker {
...
+  encodeV3SlowFillLeaf(
+    _slowRelayTree: MerkleTree<SlowFillLeaf>,
+    rootBundleId: number,
+    leaf: v3SlowFillLeaf
+  ): { method: string; args: (number | string[] | v3SlowFillLeaf)[] } {
+    // const { relayData, chainId, updatedOutputAmount } = leaf;
+
+    const method = "executeV3SlowRelayLeaf";
+    // const proof = slowRelayTree.getHexProof({ relayData, chainId, updatedOutputAmount })
+    const proof = ["xxx todo"];
```
pxrl, line 1301: nb. I'd propose to get this in as-is. The code is commented out because tsc is otherwise unhappy about types mismatching. This will be resolved long before we ever actually end up here in production.

mrice32 (commented):
Looks great!

This is really sensitive code, so I want to make another pass tomorrow before ✅

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -620,31 +627,39 @@ export function generateMarkdownForRootBundle(
...
+    if (sdkUtils.isV2SlowFillLeaf(leaf)) {
+      slowFill.destinationToken = convertTokenAddressToSymbol(leaf.relayData.destinationChainId, outputToken);
+      slowFill.amount = outputAmount;
+
+      // v2SlowFill payoutAdjustmentPct is incorrectly defined as a string.
+      // It's unconditionally 0 and will be removed, so just BN it on the fly.
+      slowFill.payoutAdjustmentPct = `${formatFeePct(toBN(leaf.payoutAdjustmentPct))}%`;
+    } else {
+      slowFill.outputToken = outputToken;
+      slowFill.outputAmount = outputAmount;
```
pxrl, line 659: I wonder if this should be derived as `inputAmount * (1e18 - realizedLpFeePct)` instead...

Reason: For v2 deposits we show `realizedLpFeePct` as a % of amount, so the absolute `realizedLpFee` would be determined by multiplying via `realizedLpFeePct * amount`. This does not hold true for v3 slow fills when `outputAmount` is used because it is ex-LP fee.

src/dataworker/Dataworker.ts
pxrl: Yep, fair point! Fixed: https://github.com/across-protocol/relayer-v2/pull/1157/commits/d9809c8b781db91fbe12fa97cdd7418a73aec8b9

pxrl:
> Looks great!
> 
> This is really sensitive code, so I want to make another pass tomorrow before ✅

All good - I'm not thrilled by the number of changes here so I've decided to refocus this one on dataworker functionality. Relayer changes are much simpler and lower risk, and will be made over here: https://github.com/across-protocol/relayer-v2/pull/1171.

src/clients/ProfitClient.ts
pxrl: Relocated to a separate PR: https://github.com/across-protocol/relayer-v2/pull/1171.

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: How does `updatedOutputAmount` fit into this reasoning?

nicholaspai (approved):
How does `updatedOutputAmount` fit into this reasoning?

src/dataworker/PoolRebalanceUtils.ts
pxrl: Had forgotten it was available - since we can rely on token equivalency then the LP fee can be inferred by the spread between inputAmount and outputAmount. Thanks for the nudge!

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -620,31 +622,39 @@ export function generateMarkdownForRootBundle(
...
+      // Fee decimals is always 18. 1e18 = 100% so 1e16 = 1%.
+      realizedLpFeePct: `${formatFeePct(leaf.relayData.realizedLpFeePct)}%`,
+    };
+
+    if (sdkUtils.isV2SlowFillLeaf(leaf)) {
+      slowFill.destinationToken = convertTokenAddressToSymbol(leaf.relayData.destinationChainId, outputToken);
+      slowFill.amount = outputAmount;
+
+      // v2SlowFill payoutAdjustmentPct is incorrectly defined as a string.
+      // It's unconditionally 0 and will be removed, so just BN it on the fly.
```
mrice32, line 650: The type for it is wrong? Should we fix that upstream? Or do we somehow rely on the incorrect type (if so, how does that work)?

src/utils/FillUtils.ts
```diff
@@ -158,7 +158,7 @@ export async function getFillDataForSlowFillFromPreviousRootBundle(
 
   // Find the first fill chronologically for matched deposit for the input fill.
   const allMatchingFills = sortEventsAscending(
-    allValidFills.filter((_fill: FillWithBlock) => sdkUtils.filledSameDeposit(_fill, fill))
+    allValidFills.filter((_fill: FillWithBlock) => sdkUtils.isV2Fill(_fill) && sdkUtils.filledSameDeposit(_fill, fill))
```
mrice32, line 161: Just a quick question -- we don't use getFillDataForSlowFillFromPreviousRootBundle to find the slow fill initiation txn for v3 fills, since that's asserted above? But we still need to find the slow fill request in v3. So we do that via a different method? Am I thinking about that correctly?

mrice32 (approved):
LGTM! Just two questions

src/utils/FillUtils.ts
pxrl: Yep, you're spot on. We do need it; this doesn't currently work with v3 objects. I think this part might need broader refactoring to cleanly handle the v2/v3 split, 

For context, this PR is an attempt to make as many of the "dumb" changes that are prerequisites for v3 support to be introduced. In most cases the changes are nicely compatible with v3, but in cases like this one we'd slam into a brick wall if we tried to feed a v3 event in.

The reason for carving it up like this is that otherwise the modifications start to get contagious and spread throughout the codebase, resulting in a large and difficult PR. But there are probably also line-ball calls in a couple of instances - this might be one of them.

Also fwiw that `isV2Fill()` statement has been removed from the current version of the PR, but the constraint is still there because the input type is specified to be `V2Fill`, so tsc will complain if we try to feed it with a V3 fill object.

src/dataworker/PoolRebalanceUtils.ts
pxrl: It's been defined as a string ever since it was introduced with the UBA work. We seem to use this type as an output - i.e. when building execute transactions, and I guess ethers accepts `BigNumberish` so it also works fine with a string. In that case it's debatable as to whether it's even a bug...I just have an aversion to numbers as strings. In any case, I'm not sure whether we should change it anyway given that it's about to be deprecated. wdyt?

src/utils/FillUtils.ts
pxrl: I used my brain and updated the implementation. The change was simple and it should now handle both v2 and v3 input types. See: https://github.com/across-protocol/relayer-v2/pull/1157/commits/6d4bb8fdb42d040bb3d93d47e30258867a0d2ec8

nicholaspai: To respond to matt's question: 

> But we still need to find the slow fill request in v3. So we do that via a different method

I think we can handle this a bit more cleanly in V3 by detecting [unexecutable slow fills](https://github.com/across-protocol/relayer-v2/pull/1155/files#diff-4d2a319a2341b776c343d9ba4c4772e4fe80e2937347947f4295128d83ff0b8bR510) in the `BundleDataClient`. Then, `BundleDataClient.loadData` can return us a list of `unexecutableSlowFills` that we can use to figure out all the excesses we need to decrement from `runningBalances` in the `Dataworker`

pxrl:
I've tested this pretty extensively:
- Diffed validateRunningBalances.ts between current master and this PR (only timestamps differed).
- All standard test cases pass.
- Validated the past week of root bundle proposals (all OK).

Based on the above, I'll merge it to prod now.

PR #1159, opened 2024-01-25
feat(linea): l2->l1 and l2->l1 finalizer utilities

dohaki:
Closes ACX-1774 and ACX-1815


james-a-morris:
QQ: is the failing related to this PR?

src/finalizer/index.ts
```diff
@@ -161,7 +168,7 @@ export async function finalize(
     }
     logger.debug({
       at: "finalize",
-      message: `Found ${totalWithdrawalsForChain} ${network} transfers (${totalWithdrawalsForChain} withdrawals | ${totalDepositsForChain} deposits | ${totalMiscTxnsForChain} supporting txns ) for finalization.`,
+      message: `Found ${totalWithdrawalsForChain} ${network} transfers (${totalWithdrawalsForChain} withdrawals | ${totalDepositsForChain} deposits | ${totalMiscTxnsForChain} misc txns) for finalization.`,
```
james-a-morris, line 171: +1

src/finalizer/index.ts
```diff
@@ -284,9 +291,11 @@ export async function finalize(
       const { miscReason } = crossChainTransfer;
       const originationNetwork = getNetworkName(originationChainId);
       const destinationNetwork = getNetworkName(destinationChainId);
+      const infoLogMessage =
+        amount && symbol ? `to support a ${originationNetwork} withdrawal of ${amount} ${symbol} 🔜` : "";
       logger.info({
         at: "Finalizer",
-        message: `Submitted ${miscReason} on ${destinationNetwork} to support a ${originationNetwork} withdrawal of ${amount} ${symbol} 🔜`,
+        message: `Submitted ${miscReason} on ${destinationNetwork} ` + infoLogMessage,
```
james-a-morris, line 298:
```suggestion
        message: `Submitted ${miscReason} on ${destinationNetwork}`,
        infoLogMessage,
```
This way we can record in GCP but keep `message` relatively small.

src/finalizer/types.ts
```diff
@@ -7,21 +7,25 @@ import { winston } from "../utils";
...
+      l1TokenSymbol: string;
+      amount: string;
     }
   | {
       type: "misc";
       miscReason: string;
+      // Note: Properties `l1TokenSymbol` and `amount` are optional for type "misc" b/c not all misc transactions will have
+      // these properties, e.g. a call to `relayMessage` of an Adapter contract.
+      l1TokenSymbol?: string;
+      amount?: string;
```
james-a-morris, line 28: OOC which parts would be optional? My understanding is that each event is at least associated with a finalization and that finalization should have an amount/l1TokenSymbol?

src/finalizer/utils/linea/common.ts
```diff
@@ -0,0 +1,81 @@
...
+  srcMessageService: L1MessageServiceContract | L2MessageServiceContract,
+  dstClaimingService: L1ClaimingService | L2MessageServiceContract
+) {
+  /**
+   * Retrieves Linea's MessageSent events for a given transaction hash and enhances them with their status.
+   * Note that there can be multiple MessageSent events per transaction hash.
+   * @param txHash Transaction hash to retrieve MessageSent events for.
+   * @returns Array of MessageSent events with their status.
+   */
+  return async function getMessagesWithStatusByTxHash(
```
james-a-morris, line 36: Nit: do we want this to be a named function? Is there a reason to use this vs. an anonymous fn?

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,193 @@
...
+  const [l1ChainId, hubPoolAddress] = [hubPoolClient.chainId, hubPoolClient.hubPool.address];
+  const l2ChainId = l1ChainId === 1 ? 59144 : 59140;
+  const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
+  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
+  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
+  const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l1Contract, l2Contract);
+
+  // Get Linea's `MessageSent` events originating from HubPool
+  const messageSentEvents = await paginatedEventQuery(
+    // l1Contract.contract.connect(hubPoolClient.hubPool.provider),
```
james-a-morris, line 34: Is this needed?

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,193 @@
...
+    l1Contract.contract,
+    l1Contract.contract.filters.MessageSent(hubPoolAddress, null, null, null, null, null, null),
```
james-a-morris, line 36: The remaining nulls shouldn't be needed
```suggestion
    l1Contract.contract.filters.MessageSent(hubPoolAddress),
```

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,193 @@
...
+    {
+      fromBlock: latestBlockToFinalize,
+      toBlock: hubPoolClient.latestBlockSearched,
+      maxBlockLookBack: 10_000,
```
james-a-morris, line 40: We should probably reference the `CHAIN_MAX_BLOCK_LOOKBACK` for this

james-a-morris (commented):
Left a few comments. Looks amazing

src/finalizer/types.ts
dohaki: Yea maybe we should rename the type to `CrossChainMessage`? Up until now, every finalization was associated with a transfer (L2->L1) and for CCTP (L1->L2 AND L2->L1). On Linea, we also have to finalize L1->L2. This means that we also need to finalize `MessageRelayed` events which are not associated with token transfer. These are admin functions like `setDepositRoute` or similar.

src/finalizer/utils/linea/common.ts
dohaki: Not really, I can make anonymous if you prefer

src/finalizer/utils/linea/l1ToL2.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/4520d04a2e3a3310d83daf49a99f5123933d5ca7

src/finalizer/utils/linea/common.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/4520d04a2e3a3310d83daf49a99f5123933d5ca7

src/finalizer/types.ts
james-a-morris: Ah I see! Yep - I think `CrossChainMessage` would be a great name then

james-a-morris (commented):
Last outstanding question then LGTM https://github.com/across-protocol/relayer-v2/pull/1159#discussion_r1513038189

src/finalizer/types.ts
dohaki: Changed here https://github.com/across-protocol/relayer-v2/pull/1159/commits/21ed6aa1bbddc55ee907908c5a0cd032ddc13c3a

james-a-morris (commented):
LGTM - do we have any finalizations that I can review or simulations?

src/finalizer/utils/linea/l2ToL1.ts
```diff
@@ -0,0 +1,152 @@
...
+  const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
+  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
+  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
+  const l1ClaimingService = lineaSdk.getL1ClaimingService(l1Contract.contractAddress);
+  const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l2Contract, l1ClaimingService);
+
+  // Get src events
+  const l2SrcEvents = spokePoolClient
+    .getTokensBridged()
+    .filter(({ blockNumber }) => blockNumber > latestBlockToFinalize);
```
nicholaspai, line 29: Can we optimize this a bit by making sure the blockNumber is also < latestBlock - 6 hours ? So we're only attempting to finalize events between 6 and 32 hours?

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,192 @@
...
+  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
+  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
+  const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l1Contract, l2Contract);
+
+  // Get Linea's `MessageSent` events originating from HubPool
+  const messageSentEvents = await paginatedEventQuery(
+    l1Contract.contract,
+    l1Contract.contract.filters.MessageSent(hubPoolAddress),
+    {
+      fromBlock: latestBlockToFinalize,
```
nicholaspai, line 37: How long do L1 to L2 messages take to finalize? Do we need a different `latestBlockToFinalize` for linea? I'm also thinking the pattern of passing in the latest block to finalize is kind of broken so feel free to define the lookbacks locally in this file to better optimize the search.

Ive noticed that biggest source of slowdown in the finalizer is checking status via the local SDK for all TokenBridged events that pass this filter, therefore we want to optimize this filter as much as possible: https://github.com/across-protocol/relayer-v2/pull/1284

nicholaspai (commented):
Couple drive-by comments about the lookback window

james-a-morris (approved):
LGTM pending Nick's signoff on the lookbacks

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,192 @@
...
+  const {
+    claimed = [],
+    claimable = [],
+    unknown = [],
+  } = groupBy(mergedMessages, ({ message }) => {
+    return message.status === OnChainMessageStatus.CLAIMED
+      ? "claimed"
+      : message.status === OnChainMessageStatus.CLAIMABLE
+      ? "claimable"
+      : "unknown";
```
mrice32, line 72: tiny nit: I think a switch statement here might be easier to read than a nested ternary.

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Populate txns for claimable messages
+  const populatedTxns = await Promise.all(
+    claimable.map(async ({ message }) => {
+      return l2Contract.contract.populateTransaction.claimMessage(
+        message.messageSender,
+        message.destination,
+        message.fee,
+        message.value,
+        (signer as Wallet).address,
```
mrice32, line 83: nit: any reason not to just call `await getAddress()` to avoid the cast? I suspect there's no additional overhead for a wallet instance since the address is synchronously available

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,192 @@
...
+    target: l2Contract.contractAddress,
+    callData: txn.data,
+  }));
+
+  // Populate cross chain calls for claimable messages
+  const messages = claimable.flatMap(({ adapterEvent }) => {
+    const { name, args } = adapterEvent.parsedLog;
+
+    if (!["TokensRelayed", "MessageRelayed"].includes(name)) {
+      return [];
```
mrice32, line 99: flatMap + returning [] is an interesting strategy for filtering within a map operation. Never seen that before

src/finalizer/utils/linea/l2ToL1.ts
```diff
@@ -0,0 +1,152 @@
...
+  logger: winston.Logger,
+  signer: Signer,
+  hubPoolClient: HubPoolClient,
+  spokePoolClient: SpokePoolClient,
+  latestBlockToFinalize: number
+): Promise<FinalizerPromise> {
+  const [l1ChainId, l2ChainId] = [hubPoolClient.chainId, spokePoolClient.chainId];
+  const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
+  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
+  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
```
mrice32, line 22: It appears that these addresses are unused _unless_ the network is localhost (which we don't support in the initLineaSdk).

I suspect we have the same addresses that they do, but reading this code, you might think changing these lines would change the address used, and from my understanding, it wouldn't.

Does your understanding match mine?

If so, I would suggest dropping these inputs entirely, and we can add an override if we ever offer a localhost option in initLineaSdk. Thoughts?

mrice32 (approved):
Looks good, just minor comments

src/finalizer/utils/linea/l2ToL1.ts
dohaki: AFAICT `latestBlockToFinalize` is passed in from `src/finalizer/index` https://github.com/across-protocol/relayer-v2/blob/0976575dc98463ab938f453c20e0082a353f3a99/src/finalizer/index.ts#L133-L139. For Linea this means that the value is `latestBlockToFinalize = currentTime - 36 hours`.  

Do I understand your suggestion correctly to filter also by the condition `blockNumber < currentTime - 6 hours`?

src/finalizer/utils/linea/l1ToL2.ts
dohaki: It takes ~ 20 mins. I added the correct value to the `finalizationWindows` constant

dohaki: https://github.com/across-protocol/relayer-v2/blob/02faab013f73901703fc78d228fc9e2747a4c59e/src/finalizer/index.ts#L93

src/finalizer/utils/linea/l2ToL1.ts
dohaki: Yes, you are right. Changed

dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/71bf6fa8113b56408d63ae714e02a72abde63147

src/finalizer/utils/linea/l1ToL2.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/71bf6fa8113b56408d63ae714e02a72abde63147

src/finalizer/utils/linea/l1ToL2.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/71bf6fa8113b56408d63ae714e02a72abde63147

src/finalizer/utils/linea/l2ToL1.ts
dohaki: I refactored here https://github.com/across-protocol/relayer-v2/pull/1159/commits/1f6af55d7ff9c820d0dd7200cb58bb4c977a8052

src/finalizer/utils/linea/l1ToL2.ts
dohaki: Refactored here https://github.com/across-protocol/relayer-v2/pull/1159/commits/1f6af55d7ff9c820d0dd7200cb58bb4c977a8052

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -0,0 +1,204 @@
...
+  const l2ChainId = l1ChainId === 1 ? 59144 : 59140;
+  const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
+  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
+  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
+  const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l1Contract, l2Contract);
+
+  // Optimize block range for querying Linea's MessageSent events on L1.
+  // We want to query for events that are at most 1 hour old, but at least 8 hours old
+  // because Linea L1->L2 messages are claimable after ~20 mins.
+  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l1ChainId, 8, 1);
```
nicholaspai, line 39: If we can finalize after 20 mins shouldn't we try searching for blocks that are between 4 hours and 0 hours old?

Between 8 hours and 1 hour doesn't seem optimized since there will be many events younger than 1 hour and older than 20 mins we can finalize and also 8 hours seems too old

src/finalizer/utils/linea/l2ToL1.ts
```diff
@@ -0,0 +1,160 @@
...
+  const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
+  const l2Contract = lineaSdk.getL2Contract();
+  const l1Contract = lineaSdk.getL1Contract();
+  const l1ClaimingService = lineaSdk.getL1ClaimingService(l1Contract.contractAddress);
+  const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l2Contract, l1ClaimingService);
+
+  // Optimize block range for querying relevant source events on L2.
+  // We want to query for events that are at most 6 hours old, but at least 32 hours old
+  // because Linea L2->L1 messages are claimable after 6 - 32 hours
+  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l2ChainId, 32, 6);
```
nicholaspai, line 32: I think we want to set the upper end of the lookback to 48 hours since this filter seems like we'll miss any events older than 32 hours. But if an event takes exactly 32 hours to finalize we'll miss it

nicholaspai (commented):
Comments on the lookbacks

src/finalizer/utils/linea/l2ToL1.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/3f1bca72a54e2e52c128963fd52ad4e985dd4cf2

src/finalizer/utils/linea/l1ToL2.ts
dohaki: https://github.com/across-protocol/relayer-v2/pull/1159/commits/3f1bca72a54e2e52c128963fd52ad4e985dd4cf2

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -34,9 +34,9 @@ export async function lineaL1ToL2Finalizer(
   const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l1Contract, l2Contract);
 
   // Optimize block range for querying Linea's MessageSent events on L1.
-  // We want to query for events that are at most 1 hour old, but at least 8 hours old
+  // We want to query for events that are between 0 and 4 hours old
   // because Linea L1->L2 messages are claimable after ~20 mins.
-  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l1ChainId, 8, 1);
+  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l1ChainId, 4, 0);
```
nicholaspai, line 39: I think we want to be a bit more conservative and leave this at 24 hours to start

src/finalizer/utils/linea/l2ToL1.ts
```diff
@@ -27,9 +27,9 @@ export async function lineaL2ToL1Finalizer(
   const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l2Contract, l1ClaimingService);
 
   // Optimize block range for querying relevant source events on L2.
-  // We want to query for events that are at most 6 hours old, but at least 32 hours old
+  // We want to query for events that are 6 and 48 hours old
   // because Linea L2->L1 messages are claimable after 6 - 32 hours
-  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l2ChainId, 32, 6);
+  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l2ChainId, 48, 6);
```
nicholaspai, line 32: I think we want to change this to 72, 8 to start

nicholaspai (approved):
Some suggestions on constants

PR #1160, opened 2024-01-26
improve: Sanity test for duplicate fills to refund, and opportunistically remove UBA logic in BundleDataClient

nicholaspai:
Duplicate  valid fills shouldn't be possible to send at the contract level since they'll map to the same relay hash and the second will revert, but we might as well add a check here


src/clients/BundleDataClient.ts
```diff
@@ -178,21 +177,11 @@ export class BundleDataClient {
...
-      blockRangesForChains,
-      this.clients.hubPoolClient.chainId,
-      this.chainIdListForBundleEvaluationBlockNumbers
-    )[0];
-    let isUBA = false;
-    if (isUBAActivatedAtBlock(this.clients.hubPoolClient, mainnetStartBlock, this.clients.hubPoolClient.chainId)) {
-      isUBA = true;
-    }
-    return this._loadData(blockRangesForChains, spokePoolClients, isUBA, logData);
+    return this._loadData(blockRangesForChains, spokePoolClients, logData);
```
james-a-morris, line 180: Without the UBA functionality, this logic falls back to a direct call to `this._loadData` as a passthrough. Thoughts on merging the functions into one `loadData` fn?

james-a-morris (commented):
Left a comment regarding `_loadData()` in the `BundleDataClient`

src/clients/BundleDataClient.ts
nicholaspai: will adjust after we get that code into master

src/clients/BundleDataClient.ts
```diff
@@ -233,6 +222,18 @@ export class BundleDataClient {
...
       blockRangeForChain: number[]
     ) => {
+      // Extra check for duplicate fills. These should be blocked at the contract level but might still be included
+      // by the RPC so its worth checking here.
+      if (
+        allValidFills.some(
+          (existingFill) =>
+            existingFill.originChainId === fillWithBlock.originChainId &&
+            existingFill.depositId === fillWithBlock.depositId &&
+            existingFill.totalFilledAmount === fillWithBlock.totalFilledAmount
```
pxrl, line 232: Will probably need to sub in `sdkUtils.getTotalFilledAmount()` here in order to be compatible with v2 and v3 fills.

src/clients/BundleDataClient.ts
```diff
@@ -233,6 +222,18 @@ export class BundleDataClient {
...
+        )
+      ) {
+        return;
```
pxrl, line 235: Instead of `allValidFills.some()`, how about `allValidFills.find()` so we can then log the first matching instance? That would make it easier to troubleshoot this if it ever did occur.

src/clients/BundleDataClient.ts
nicholaspai: 093c3fbc19653b939a72cd24bffb65cdd7c09685

src/clients/BundleDataClient.ts
nicholaspai: 8767b0b03a65be41f6da2ea8e8bde6d21dd84f59

src/clients/BundleDataClient.ts
nicholaspai: 8767b0b03a65be41f6da2ea8e8bde6d21dd84f59

src/clients/BundleDataClient.ts
```diff
@@ -233,6 +209,23 @@ export class BundleDataClient {
       matchedDeposit: DepositWithBlock,
       blockRangeForChain: number[]
     ) => {
+      // Extra check for duplicate fills. These should be blocked at the contract level but might still be included
+      // by the RPC so its worth checking here.
+      const duplicateFill = allValidFills.find(
+        (existingFill) =>
+          existingFill.originChainId === fillWithBlock.originChainId &&
+          existingFill.depositId === fillWithBlock.depositId &&
+          utils.getTotalFilledAmount(existingFill).eq(utils.getTotalFilledAmount(fillWithBlock))
```
mrice32, line 218: Side note: this checks for exact duplicates, but different partial fill divisions could still trick this I believe.

Regardless, I think in v3, this will cease to be a problem since partial fills will go away. We will basically just have to look for duplicates by fill type (slow request vs true fill).

src/clients/BundleDataClient.ts
```diff
@@ -233,6 +209,23 @@ export class BundleDataClient {
       matchedDeposit: DepositWithBlock,
       blockRangeForChain: number[]
     ) => {
+      // Extra check for duplicate fills. These should be blocked at the contract level but might still be included
+      // by the RPC so its worth checking here.
+      const duplicateFill = allValidFills.find(
```
mrice32, line 214: To be clear I'm not sure there's a good answer to this.

This checks for duplicates _in this current event window_, but it doesn't perform a global duplicate check, right?

Can you think of any way to ensure that we haven't repaid this fill in the past (even distant past)? In my mind, it seems like there's no real way to do this without relying on some sort of off-chain persistent bundle indexing service.

mrice32 (approved):
LGTM!

src/clients/BundleDataClient.ts
nicholaspai: yeah that's definitely a concern but maybe out of scope.

nicholaspai: we could perform some short window lookback check if we produce the last N bundles, but that would increase run time so i'm hesitant

PR #1162, opened 2024-01-29
refactor: Remove Relayer UBA functionality

pxrl:
UBA support is hampering updates for v3. Drop it for now; we can re-introduce it later once we're over the v3 hump.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { utils as ethersUtils } from "ethers";
 import { Deposit, DepositWithBlock, L1Token } from "../interfaces";
@@ -283,6 +284,7 @@ export class Relayer {
       const l1Token = hubPoolClient.getL1TokenInfoForL2Token(originToken, originChainId);
       const selfRelay = [depositor, recipient].every((address) => address === this.relayerAddress);
       if (tokenClient.hasBalanceForFill(deposit, unfilledAmount) && !selfRelay) {
+        assert(isDefined(deposit.realizedLpFeePct)); // Sanity check.
         const { repaymentChainId, gasLimit: gasCost } = await this.resolveRepaymentChain(
           version,
           deposit,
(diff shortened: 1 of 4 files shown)


src/relayer/Relayer.ts
```diff
@@ -284,9 +283,6 @@ export class Relayer {
       const l1Token = hubPoolClient.getL1TokenInfoForL2Token(originToken, originChainId);
       const selfRelay = [depositor, recipient].every((address) => address === this.relayerAddress);
       if (tokenClient.hasBalanceForFill(deposit, unfilledAmount) && !selfRelay) {
-        // The pre-computed realizedLpFeePct is for the pre-UBA fee model. Update it to the UBA fee model if necessary.
-        // The SpokePool guarantees the sum of the fees is <= 100% of the deposit amount.
-        deposit.realizedLpFeePct = await this.computeRealizedLpFeePct(version, deposit);
```
nicholaspai, line 289: A deposit will now have to define `realizedLpFeePct` to get to this point in the code right?

src/relayer/Relayer.ts
```diff
@@ -484,7 +480,7 @@ export class Relayer {
       ? await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)
       : destinationChainId;
 
-    const refundFee = this.computeRefundFee(version, deposit);
+    const refundFee = bnZero;
```
nicholaspai, line 483: I'm ok with keeping `refundFee` as an argument to the profitability functions, do you think it would be used in v3?

nicholaspai (approved):
Thx for splitting up dataworker and relayer changes

src/relayer/Relayer.ts
pxrl: Yep - the corresponding sdk PR is here: https://github.com/across-protocol/sdk-v2/pull/517

Per current implementation, we can actually rely on it being there as long as the UBA is not active. With that code being removed then tsc can expect it to be defined again.

src/relayer/Relayer.ts
pxrl: Nah, I don't expect this to be used. I set it to zero here to avoid these changes spreading into the ProfitClient. The ProfitClient cleanup can probably happen simultaneously with the pending v3 updates.

src/relayer/Relayer.ts
pxrl: Just to be paranoid about it, I added an assert to guarantee that we don't end up with an undefined realizedLpFeePct. This can be removed after we bump to 0.20.4.

https://github.com/across-protocol/relayer-v2/pull/1162/commits/5872d93eb8408a4c22c7b0b8f378e186a173bd23

PR #1169, opened 2024-01-30
hotfix(finalizer): update polygon finalizer

james-a-morris:
This PR is aimed at ensuring that all `callData` transactions produce a corresponding `CrossChainTransfer` element.

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -286,7 +286,7 @@ export async function finalize(
       const destinationNetwork = getNetworkName(destinationChainId);
       logger.info({
         at: "Finalizer",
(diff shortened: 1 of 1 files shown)


src/finalizer/index.ts
```diff
@@ -268,14 +268,14 @@ export async function finalize(
...
+    const { transfers = [], misc = [] } = groupBy(
       finalizations.filter(({ crossChainTransfer }) => isDefined(crossChainTransfer)),
       ({ crossChainTransfer: { type } }) => {
-        return type === "proof" ? "proofs" : "transfers";
+        return type === "misc" ? "misc" : "transfers";
       }
     );
 
-    proofs.forEach(
+    misc.forEach(
```
nicholaspai, line 278: Can you make sure the log `message` is appropriate somehow for the different `misc` txns? Should we have the `misc` type have an optional extra string that can be passed to give more details?

nicholaspai (approved):

PR #1171, opened 2024-01-31
refactor: Relayer-specific preparation for v3

pxrl:
This commit introduces forwards-compatible changes to the relayer function, with that will ease the pending types upgrade for Across v3.

No change in behaviour is intended as a result of the change, but it should de-risk the eventual diff by de-cluttering it.

nb. This code has been split out from the original PR https://github.com/across-protocol/relayer-v2/pull/1157.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -51,16 +51,7 @@ export class Relayer {
 
     const maxVersion = configStoreClient.configStoreVersion;
     return unfilledDeposits.filter(({ deposit, version, invalidFills, unfilledAmount }) => {
-      const {
-        quoteTimestamp,
-        depositId,
-        depositor,
-        recipient,
-        originChainId,
-        destinationChainId,
-        originToken,
-        amount,
-      } = deposit;
+      const { quoteTimestamp, depositId, depositor, recipient, originChainId, destinationChainId } = deposit;
       const destinationChain = getNetworkName(destinationChainId);
 
       // If we don't have the latest code to support this deposit, skip it.
@@ -129,7 +120,9 @@ export class Relayer {
       }
 
       // Resolve L1 token and perform additional checks
-      const l1Token = hubPoolClient.getL1TokenInfoForL2Token(originToken, originChainId);
+      // @todo: This is only relevant if inputToken and outputToken are equivalent.
+      const inputToken = sdkUtils.getDepositInputToken(deposit);
+      const l1Token = hubPoolClient.getL1TokenInfoForL2Token(inputToken, originChainId);
 
       // Skip any L1 tokens that are not specified in the config.
       // If relayerTokens is an empty list, we'll assume that all tokens are supported.
@@ -137,6 +130,18 @@ export class Relayer {
         this.logger.debug({
           at: "Relayer::getUnfilledDeposits",
           message: "Skipping deposit for unwhitelisted token",
+          deposit,
+          l1Token,
+        });
+        return false;
+      }
+
+      // Filter out deposits that require native swaps.
+      const outputToken = sdkUtils.getDepositOutputToken(deposit);
+      if (!hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)) {
+        this.logger.debug({
+          at: "Relayer::getUnfilledDeposits",
+          message: "Skipping deposit including native token swap.",
           deposit,
           l1Token,
         });
@@ -154,7 +159,7 @@ export class Relayer {
           limit: acrossApiClient.getLimit(l1Token.address),
           l1Token: l1Token.address,
           depositId,
-          amount,
+          inputAmount: sdkUtils.getDepositInputAmount(deposit),
           unfilledAmount: unfilledAmount.toString(),
           originChainId,
           transactionHash: deposit.transactionHash,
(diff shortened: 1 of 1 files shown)


src/clients/InventoryClient.ts
```diff
@@ -185,15 +186,22 @@ export class InventoryClient {
...
     if (!this.isInventoryManagementEnabled() || destinationChainId === hubChainId) {
       return destinationChainId;
     }
 
-    l1Token ??= this.hubPoolClient.getL1TokenForL2TokenAtBlock(originToken, originChainId);
+    // The InventoryClient assumes 1:1 equivalency between input and output tokens. If the tokens are not equivalent
+    // then just take repayment on the destination chain. The operator is responsible for balancing inventory.
+    const inputToken = sdkUtils.getDepositInputToken(deposit);
+    const outputToken = sdkUtils.getDepositOutputToken(deposit);
+    if (!this.hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)) {
```
nicholaspai, line 201: I think at this stage it might worth throwing an error here so we know if we are trying to take a refund for a fill where the tokens are not equivalent

src/relayer/Relayer.ts
```diff
@@ -143,6 +136,18 @@ export class Relayer {
         return false;
       }
 
+      // Filter out deposits that require native swaps.
+      const outputToken = sdkUtils.getDepositOutputToken(deposit);
+      if (!hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)) {
+        this.logger.debug({
+          at: "Relayer::getUnfilledDeposits",
+          message: "Skipping deposit including native token swap.",
```
nicholaspai, line 144: what is a "native" token swap?

src/clients/InventoryClient.ts
pxrl: Yeah OK. I was thinking that this might cause us to be open to DoS by someone making a deposit with a different outputToken (any tiny amount would suffice) but this logic is downstream of that decision, so it's probably a good call to fail hard if we make it this far.

src/relayer/Relayer.ts
pxrl: My made-up name for an in-protocol swap (i.e. by specifying a different output token). Have we already named this?

src/clients/InventoryClient.ts
pxrl: Updated: https://github.com/across-protocol/relayer-v2/pull/1171/commits/828b4c0d049b16042693fc28efdf38f8c5332f94

mrice32: Yeah, I think we should issue a warning or error log, but not cause the bot to stop. Intentional DoS or just a bad input could take the bot down.

src/relayer/Relayer.ts
mrice32: Not to bikeshead, but native token swap sounds like you're swapping the native token :)

What about relayer swap or in-protocol swap? Or just swap?

mrice32 (commented):
Only blocking comment is the error! Otherwise, LGTM!

src/clients/InventoryClient.ts
```diff
@@ -185,15 +186,28 @@ export class InventoryClient {
...
-    l1Token ??= this.hubPoolClient.getL1TokenForL2TokenAtBlock(originToken, originChainId);
+    // The InventoryClient assumes 1:1 equivalency between input and output tokens. At the moment there is no support
+    // for disparate output tokens, so if one appears here then something is wrong. Throw hard and fast in that case.
+    // In future, fills for disparate output tokens should probably just take refunds on the destination chain and
+    // outsource inventory management to the operator.
+    const inputToken = sdkUtils.getDepositInputToken(deposit);
+    const outputToken = sdkUtils.getDepositOutputToken(deposit);
+    if (!this.hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)) {
+      const [srcChain, dstChain] = [getNetworkName(originChainId), getNetworkName(destinationChainId)];
+      throw new Error(
```
mrice32, line 205: Don't think we should allow this to stop the bot. Prefer warning or error log to alert without blocking the bot from running.

src/clients/InventoryClient.ts
pxrl: I don't think we have any marginal DoS risk here because we are already filtering the array of unfilled deposits based on token equivalency, so we shouldn't actually see any swaps at this late stage. By this point the relayer has already decided it wants to make a fill and that it has sufficient balance for it, so there's no real way to say pre-empt it on that decision other than to throw.

If we were to remove or unintentionally break the token equivalency filtering then we could throw here, but then we'd have an objective error in the bot so I'm probably OK with doing that. wdyt?

src/clients/InventoryClient.ts
pxrl: I unresolved it and left a response: https://github.com/across-protocol/relayer-v2/pull/1171#discussion_r1473768007

src/relayer/Relayer.ts
pxrl: in-protocol :heavy_check_mark:

src/clients/InventoryClient.ts
mrice32: Ah, I see, makes sense! Sorry for the noise -- I missed that in your original comment

mrice32 (approved):

src/clients/InventoryClient.ts
pxrl: All good - ++appreciate the eyes on these changes :pray:

PR #1177, opened 2024-02-02
refactor: Remove redundant type definition

pxrl:
This type is not used anywhere in this repository.


pxrl:
fwiw this is also proposed for removal in sdk-v2 since it's not used there either: https://github.com/across-protocol/sdk-v2/pull/527

nicholaspai (approved):

PR #1179, opened 2024-02-02
chore: Bump sdk-v2

pxrl:
This bump clears a hurdle where some newly introduced types have been slightly renamed. The internal definitions of these types was also reorganised to use RelayData as the core/common part of each type.

### DIFF
--- a/src/interfaces/index.ts
+++ b/src/interfaces/index.ts
@@ -61,12 +61,12 @@ export type TokensBridged = interfaces.TokensBridged;
 export type CachingMechanismInterface = interfaces.CachingMechanismInterface;
 
 // V2 / V3 interfaces
-export type V2Deposit = interfaces.v2Deposit;
-export type V2DepositWithBlock = interfaces.v2DepositWithBlock;
-export type V2SpeedUp = interfaces.v2SpeedUp;
-export type V2Fill = interfaces.v2Fill;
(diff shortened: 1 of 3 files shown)


src/interfaces/index.ts
```diff
@@ -61,12 +61,12 @@ export type TokensBridged = interfaces.TokensBridged;
...
-export type V2Deposit = interfaces.v2Deposit;
-export type V2DepositWithBlock = interfaces.v2DepositWithBlock;
-export type V2SpeedUp = interfaces.v2SpeedUp;
-export type V2Fill = interfaces.v2Fill;
-export type V2FillWithBlock = interfaces.v2FillWithBlock;
-export type V2RelayData = interfaces.v2RelayData;
-export type V3RelayData = interfaces.v3RelayData;
-export type V2SlowFillLeaf = interfaces.v2SlowFillLeaf;
-export type V3SlowFillLeaf = interfaces.v3SlowFillLeaf;
+export type V2Deposit = interfaces.V2Deposit;
```
nicholaspai, line 64: is there a `V3RelayerRefundRoot` type too?

nicholaspai (approved):

src/interfaces/index.ts
pxrl: Not yet - that's still a todo on the remaining PR in sdk-v2. WIll fix it up shortly.

PR #1183, opened 2024-02-06
improve: Remove instance of Dataworker hardcoded chainId

pxrl:
Rather than hardcode a single Arbitrum chain ID, support checking against an array of Arbitrum-stack chains. This will be useful for the Arbitrum Sepolia testnet, but may also be needed for Orbit chains.

While here, zap some 0s.


src/dataworker/Dataworker.ts
```diff
@@ -1470,18 +1472,20 @@ export class Dataworker {
     tree: MerkleTree<PoolRebalanceLeaf>,
     submitExecution: boolean
   ): Promise<void> {
+    const isArbitrum = (chainId: number) => [CHAIN_IDs.ARBITRUM, CHAIN_IDs.ARBITRUM_GOERLI].includes(chainId);
```
pxrl, line 1475: fwiw this is defined locally for the time being. We have an OP-stack equivalent in the sdk so it should probably be migrated there.

pxrl: Proposed here: https://github.com/across-protocol/sdk-v2/pull/538

nicholaspai (approved):

mrice32 (approved):

pxrl:
Have added `do not merge` because I'll wait for an SDK bump before using the new utility function defined over there.

nicholaspai (approved):

PR #1184, opened 2024-02-06
improve: Bump SDK for unified V3 types

pxrl:
This commit introduces the necessary relayer-side updates to accommodate a pending SDK bump.


src/clients/ProfitClient.ts
```diff
@@ -419,7 +419,7 @@ export class ProfitClient {
   }
 
   async isFillProfitable(
```
nicholaspai, line 421: Are we going to make separate v3 functions?

pxrl:
Note - this PR needs the corresponding SDK PR to unionise v2 & v3 types: https://github.com/across-protocol/sdk-v2/pull/501

src/scripts/validateRunningBalances.ts
```diff
@@ -251,32 +252,37 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
-                      f.depositId === slowFillForChain.relayData.depositId
-                  );
 
-                const outputAmount = sdkUtils.getRelayDataOutputAmount(slowFillForChain.relayData);
-                const lastFill = sortEventsDescending(fillsForSameDeposit)[0];
-                const totalFilledAmount = sdkUtils.getTotalFilledAmount(lastFill);
-                const amountSentForSlowFillLeftUnexecuted = outputAmount.sub(totalFilledAmount);
+                let amountSentForSlowFillLeftUnexecuted: BigNumber;
+                if (sdkUtils.isV3SlowFillLeaf(slowFillForChain)) {
+                  amountSentForSlowFillLeftUnexecuted = slowFillForChain.updatedOutputAmount;
```
nicholaspai, line 266: Shouldn't we be checking if a fast fill replaced the slow fill?

src/clients/ProfitClient.ts
pxrl: I think I'll probably minimise that and will aim to use the unified types. This PR basically stubs off any possibility of getting v3 deposits in the relayer, with the aim of rectifying that in subsequent relayer-specific PRs (currently WIP).

src/scripts/validateRunningBalances.ts
pxrl: oof - yes :facepalm:

src/scripts/validateRunningBalances.ts
```diff
@@ -269,14 +271,18 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
                 const lastFill = sortEventsDescending(fillsForSameDeposit)[0];
-                const totalFilledAmount = sdkUtils.getTotalFilledAmount(lastFill);
-                const amountSentForSlowFillLeftUnexecuted = outputAmount.sub(totalFilledAmount);
 
-                if (amountSentForSlowFillLeftUnexecuted.gt(0)) {
-                  const deductionForSlowFill = getRefund(
-                    amountSentForSlowFillLeftUnexecuted,
-                    slowFillForChain.relayData.realizedLpFeePct
-                  );
+                // For v2 slow fills there must be at least one partial fill; for v3 slow fills there _may_ be a fill.
```
nicholaspai, line 275: shoudl we add an `assert` that `lastFill` is always defined if `isV2`?

src/scripts/validateRunningBalances.ts
```diff
@@ -269,14 +271,18 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
 
                 const outputAmount = sdkUtils.getRelayDataOutputAmount(slowFillForChain.relayData);
```
nicholaspai, line 272: This should be the `updatedOutputAmount` for v3 slow fills right?

mrice32 (commented):
Is this ready for review or should we wait on the PR that you referenced?

src/scripts/validateRunningBalances.ts
pxrl: Yep, correct.

src/scripts/validateRunningBalances.ts
```diff
@@ -269,14 +272,18 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
```
nicholaspai, line 273: As discussed offline, this should be `updatedOutputAmount` for V3 and `amount * (1-lpFee)` for V2, fixing a prior bug.

src/scripts/validateRunningBalances.ts
```diff
@@ -269,14 +272,18 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
-                const totalFilledAmount = sdkUtils.getTotalFilledAmount(lastFill);
-                const amountSentForSlowFillLeftUnexecuted = outputAmount.sub(totalFilledAmount);
 
-                if (amountSentForSlowFillLeftUnexecuted.gt(0)) {
-                  const deductionForSlowFill = getRefund(
-                    amountSentForSlowFillLeftUnexecuted,
-                    slowFillForChain.relayData.realizedLpFeePct
-                  );
+                // For v2 slow fills there must be at least one partial fill; for v3 slow fills there _may_ be a fill.
+                const totalFilledAmount =
```
nicholaspai, line 277: I think this also has to be `totalFilledAmount * (1 - lpFeePct)` to be consistent with the change here https://github.com/across-protocol/relayer-v2/pull/1184/files#r1484825717

src/scripts/validateRunningBalances.ts
```diff
@@ -299,28 +306,34 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
+                amountSentForSlowFill = slowFillForChain.updatedOutputAmount;
+              } else {
+                const fillsForSameDeposit = bundleSpokePoolClients[destinationChainId]
+                  .getFillsForOriginChain(slowFillForChain.relayData.originChainId)
+                  .filter((f) => f.depositId === slowFillForChain.relayData.depositId);
+
+                const lastFill = sortEventsDescending(fillsForSameDeposit)[0];
+                const outputAmount = sdkUtils.getRelayDataOutputAmount(slowFillForChain.relayData);
+                const totalFilledAmount = sdkUtils.getTotalFilledAmount(lastFill);
+                amountSentForSlowFill = outputAmount.sub(totalFilledAmount);
```
nicholaspai, line 329: For v2 we probably need to remove the `lpFee` here, which fixes an existing bug. This is how we update the `runningBalances` for slow fills, taking into account the `lpFee`, in the dataworker which is effectively how we determine how many funds to send to the Spoke: https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/PoolRebalanceUtils.ts#L174

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -596,42 +600,43 @@ export function generateMarkdownForRootBundle(
   let slowRelayLeavesPretty = "";
   slowRelayLeaves.forEach((leaf, index) => {
     const outputToken = sdkUtils.getRelayDataOutputToken(leaf.relayData);
-    const outputTokenDecimals = hubPoolClient.getTokenInfo(leaf.relayData.destinationChainId, outputToken).decimals;
+    const destinationChainId = sdkUtils.getSlowFillLeafChainId(leaf);
+    const outputTokenDecimals = hubPoolClient.getTokenInfo(destinationChainId, outputToken).decimals;
 
-    const slowFill: Record<string, string> = {
+    const slowFill: Record<string, number | string> = {
```
nicholaspai, line 606: Should we define a PartialSlowFIll type similiar to a PartialFill/PartialDeposit where we leave out props we can't know when we first query the event like:
- realizedLpFeePct
- payoutAdjustmentPct
- speedUpSignature
- updatedMessage
- updatedRecipient
- newRelayerFeePct
- updatedOutputAmount
- destinationToken

src/scripts/validateRunningBalances.ts
pxrl: I steered us down the wrong path on that discussion. We focused on the determination of whether the deposit was filled or unfilled, and missed that the post-fee amounts are computed afterwards. I've clarified that in this commit: https://github.com/across-protocol/relayer-v2/pull/1184/commits/651f6db6349d74c5fbae6f744cbacb758f9b0889. wdyt?

src/scripts/validateRunningBalances.ts
pxrl: Updated - ptal: https://github.com/across-protocol/relayer-v2/commit/651f6db6349d74c5fbae6f744cbacb758f9b0889

src/scripts/validateRunningBalances.ts
pxrl: I finally realised that it is handled, I was just blind to it :man_facepalming: 

diff

test/TokenClient.BalanceAlowance.ts
```diff
@@ -21,7 +22,29 @@ let owner: SignerWithAddress, spyLogger: winston.Logger;
...
+  [chainId: number]: {
+    [token: string]: { balance: BigNumber; allowance: BigNumber };
+  };
+};
+
 describe("TokenClient: Balance and Allowance", async function () {
+  // @todo: Why is utils.deepEqualsWithBigNumber suddenly broken in this test?
+  // The issue resolves to the use of assert.deepStrictEqual(), which seems to pass through chai to Node's own assert.
+  // The object contents _are_ identical but their addresses differ and it seems to object to that. Why now?!
+  const deepEqualsWithBigNumber = (a: TokenBalanceByChain, b: TokenBalanceByChain): boolean => {
```
pxrl, line 35: @james-a-morris Just calling this change out - for some reason, `deepEqualsWithBigNumber()` suddenly fails in this specific test. I've compared the input objects manually by eye and subsequently with sha256, and they match identically. I _think_ it's incorrectly inferring equality of BigNumber instances because they're obviously stored at different addresses, but I haven't been able to pinpoint the specific change that causes it. Since we're bumping sdk-v2 here I actually think the issue might somehow have its roots there.

Any ideas?

pxrl:
> Is this ready for review or should we wait on the PR that you referenced?

Sorry for the late response on this one; I've just bumped the SDK so it's ready now.

nicholaspai (approved):

test/TokenClient.BalanceAlowance.ts
pxrl: Thanks for the clarification. I reverted to using `expect(a).to.deep.equal(b)`. This seems to work fine: https://github.com/across-protocol/relayer-v2/pull/1184/commits/4fe1fd3a62956bfe727b60dacafbe1f9f42316a6.

src/dataworker/DataworkerUtils.ts
```diff
@@ -190,6 +193,33 @@ function buildV2SlowFillLeaf(unfilledDeposit: UnfilledDeposit): V2SlowFillLeaf {
   };
 }
 
+function buildV3SlowFillLeaf(unfilledDeposit: UnfilledDeposit): V3SlowFillLeaf {
```
mrice32, line 196: Why doesn't this function take a v3 deposit instead of asserting it on the first line?

src/relayer/Relayer.ts
```diff
@@ -203,7 +203,7 @@ export class Relayer {
 
     // Fetch unfilled deposits and filter out deposits upfront before we compute the minimum deposit confirmation
     // per chain, which is based on the deposit volume we could fill.
-    const unfilledDeposits = await this._getUnfilledDeposits();
+    const unfilledDeposits = (await this._getUnfilledDeposits()).filter(({ deposit }) => sdkUtils.isV2Deposit(deposit));
```
mrice32, line 206: No TODO to remove this filter in the future?

src/utils/FillUtils.ts
```diff
@@ -187,9 +188,12 @@ export async function getFillDataForSlowFillFromPreviousRootBundle(
     }
     // Add non-duplicate fills.
     allMatchingFills.push(
-      ...matchingFills.filter(
-        (_fill) => !allMatchingFills.find((existingFill) => existingFill.totalFilledAmount.eq(_fill.totalFilledAmount))
-      )
+      ...matchingFills.filter((_fill) => {
+        const totalFilledAmount = sdkUtils.getTotalFilledAmount(_fill);
+        return !allMatchingFills.find((existingFill) =>
```
mrice32, line 193: Is it possible that a v2 and v3 fill for the same deposit are seen by this code as duplicates of one another?

If so, is that a problem?

I just want to make sure there's no way a user could exploit the way we normalize info across types to force two fills to be seen as duplicates of one another.

mrice32 (approved):
Looks good! Just a few minor questions/comments

src/relayer/Relayer.ts
pxrl: Good catch; it's removed in the PR that immediately follows this one (but ofc you can't see that in this PR alone): https://github.com/across-protocol/relayer-v2/pull/1187/files

src/dataworker/DataworkerUtils.ts
pxrl: The `UnfilledDeposit` type embeds a `Deposit` object (named `deposit`), so per current SDK definition it's not possible to force it to accept a v3 type. Nick has an update to how we handle this type and how we produce slow fill leaves from them in a separate PR.

src/utils/FillUtils.ts
pxrl: If I've understood the question correctly, it shouldn't be possible, because the `allMatchingFills` array is filtered to ensure that all elements are of the same type as the initial `fill`that was passed into this function. The filtering happens around lines 150 - 155. Does that make sense?

pxrl: As a follow up, not also the use of `filledSameDeposit()` for filtering on line 155. `filledSameDeposit()` also embeds a version check to ensure that it only compares relay hashes across like types, so I think we should be fairly well covered in terms of filtering out mismatched fill types.

filledSameDeposit

src/dataworker/DataworkerUtils.ts
```diff
@@ -148,8 +151,8 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
   leaves: SlowFillLeaf[];
   tree: MerkleTree<SlowFillLeaf>;
 } {
-  const slowRelayLeaves: SlowFillLeaf[] = unfilledDeposits.map((deposit: UnfilledDeposit) =>
-    buildV2SlowFillLeaf(deposit)
+  const slowRelayLeaves = unfilledDeposits.map((deposit: UnfilledDeposit) =>
```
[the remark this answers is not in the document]
pxrl: I'm not 100% sure, but tsc isn't complaining about it having been removed. Do you think it's needed?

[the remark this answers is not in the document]
pxrl: Nope - good catch: https://github.com/across-protocol/relayer-v2/pull/1184/commits/efc16dc9f1f5fc88970a62bdf41bd580c398eaeb

src/dataworker/Dataworker.ts
```diff
@@ -1215,9 +1219,7 @@ export class Dataworker {
 
           // Assume we don't need to add balance in the BalanceAllocator to the HubPool because the slow fill's
           // recipient wouldn't be the HubPool in normal circumstances.
-          return success
-            ? { relayData: slowFill.relayData, payoutAdjustmentPct: slowFill.payoutAdjustmentPct }
-            : undefined;
+          return success ? { ...slowFill } : undefined;
```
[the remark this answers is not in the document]
pxrl: Good call :+1: https://github.com/across-protocol/relayer-v2/pull/1184/commits/efc16dc9f1f5fc88970a62bdf41bd580c398eaeb

src/dataworker/Dataworker.ts
```diff
@@ -1175,22 +1175,26 @@ export class Dataworker {
...
-          const fill = slowFill.fill;
-          if (isDefined(fill)) {
-            // If fill was a full fill, execution is unnecessary. V3 fills are full fills by definition.
-            amountFilled = sdkUtils.getTotalFilledAmount(fill);
-            if (amountFilled.eq(sdkUtils.getFillOutputAmount(fill))) {
+          const outputToken = sdkUtils.getRelayDataOutputToken(slowFill.relayData);
+          const outputAmount = sdkUtils.getRelayDataOutputAmount(slowFill.relayData);
+          let amountRequired: BigNumber;
+          if (sdkUtils.isV3SlowFillLeaf(slowFill)) {
+            amountRequired = slowFill.updatedOutputAmount;
```
nicholaspai, line 1182: For V3, you still need to check the filled amount if there is a fill right? Should you not use `sdkUtils.getTotalFilledAmount`?

I implement this a bit differently in my PR https://github.com/across-protocol/relayer-v2/pull/1155/files#diff-c8492eb44587aca5302d0570d80825361ae103e7d5df4643c405a612ec1d240b

src/dataworker/DataworkerUtils.ts
```diff
@@ -148,8 +151,8 @@ export function _buildSlowRelayRoot(unfilledDeposits: UnfilledDeposit[]): {
   leaves: SlowFillLeaf[];
   tree: MerkleTree<SlowFillLeaf>;
 } {
-  const slowRelayLeaves: SlowFillLeaf[] = unfilledDeposits.map((deposit: UnfilledDeposit) =>
-    buildV2SlowFillLeaf(deposit)
+  const slowRelayLeaves = unfilledDeposits.map((deposit) =>
```
nicholaspai, line 154: I'm going to overwrite this in my PR but its fine if this implementation is backwards cmopatible. I define `UnfilledDeposit` to only work with V2 Deposits

nicholaspai: https://github.com/across-protocol/relayer-v2/pull/1155/files#diff-9535349f4b8beb74323e6d78b13634dc9351a05c53fc08785c124cd5a60c8474R210

src/dataworker/DataworkerUtils.ts
nicholaspai: same comment as https://github.com/across-protocol/relayer-v2/pull/1184/files#r1491309422

test/TokenClient.BalanceAlowance.ts
```diff
@@ -96,7 +95,9 @@ describe("TokenClient: Balance and Allowance", async function () {
         [weth_2.address]: { balance: toBNWei(1337), allowance: toBNWei(0) },
       },
     };
-    expect(deepEqualsWithBigNumber(tokenClient.getAllTokenData(), expectedData1)).to.be.true;
+    const tokenData1 = tokenClient.getAllTokenData();
+    expect(tokenData1).to.deep.equal(expectedData1);
```
nicholaspai, line 99: I ran into these issues too. Is this because the BigNumber representation somehow changed in the sdk update?

The issue is that there is a difference in the BigNumber internal properties between the client and the expected objects created in this test. I think one stores the number internally as a hex and the other as a string

pxrl: I think it's somehow triggered by a locally-linked sdk-v2 package, because this issue doesn't seem to arise when the package is sourced remotely. I thought I'd resolved it by making this change, but I've also seen that fail locally.

It does seem linked to the SDK bump and hints that we might have some non-determinism in how the imports are working. It's ++frustrating and I'm not really sure what to do about it at the moment :thinking:

src/utils/FillUtils.ts
mrice32: I see, so we'd never get to this line with two fills that are different versions, so we are only adding non-duplicate fills _of the same type_, since only one version can be valid.

src/relayer/Relayer.ts
mrice32: Gotcha

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
pxrl: Just to close this one out: https://github.com/across-protocol/relayer-v2/pull/1184/commits/dcf7c904e12313f7743e458535b779bd2e5122df

src/dataworker/Dataworker.ts
pxrl: Another good catch. Fixed: https://github.com/across-protocol/relayer-v2/pull/1184/commits/0dbf0ee42a093b8105b74ab192e7969b9be20296

src/dataworker/DataworkerUtils.ts
pxrl: I'll mark this resolved since it'll be modified by the follow-up change by Nick long before we ever see any v3 events.

src/utils/FillUtils.ts
pxrl: Yep, disparate types will be filtered out before we reach this stage. RE non-duplicates, that doesn't seem to have been handled here so I don't anticipate any change in behaviour.

PR #1185, opened 2024-02-07
feat: facilitate testnet relays/dataworker operations

james-a-morris:
This PR is aimed at allowing the relayer-v2 repo to be run on sepolia based testnets.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "node": ">=16.18.0"
   },
   "dependencies": {
-    "@across-protocol/constants-v2": "1.0.8",
+    "@across-protocol/constants-v2": "1.0.9",
     "@across-protocol/contracts-v2": "2.5.0-beta.5",
     "@across-protocol/sdk-v2": "0.21.1",
     "@arbitrum/sdk": "^3.1.3",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "node": ">=16.18.0"
   },
   "dependencies": {
-    "@across-protocol/constants-v2": "1.0.9",
+    "@across-protocol/constants-v2": "1.0.11",
     "@across-protocol/contracts-v2": "2.5.0-beta.5",
     "@across-protocol/sdk-v2": "0.21.1",
     "@arbitrum/sdk": "^3.1.3",
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -37,12 +37,12 @@ const { MaxUint256, Zero } = ethers.constants;
 const { isAddress } = ethers.utils;
 
 function printDeposit(log: LogDescription, originChainId: number): void {
-  const { inputToken } = log.args;
+  const { inputToken, originToken } = log.args;
   const eventArgs = Object.keys(log.args).filter((key) => isNaN(Number(key)));
   const padLeft = eventArgs.reduce((acc, cur) => (cur.length > acc ? cur.length : acc), 0);
 
   const fields = {
-    tokenSymbol: resolveTokenSymbols([inputToken], originChainId)[0],
+    tokenSymbol: resolveTokenSymbols([originToken ?? inputToken], originChainId)[0],
     ...Object.fromEntries(eventArgs.map((key) => [key, log.args[key]])),
   };
   console.log(
@@ -161,24 +161,40 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
     ? toBN(args.relayerFeePct)
     : await getRelayerQuote(fromChainId, toChainId, token, amount, recipient, message);
   const quoteTimestamp = await spokePool.getCurrentTime();
-  const deposit = await (isV2 ? spokePool.depositV2 : spokePool.deposit)(
-    recipient,
-    token.address,
-    amount,
-    toChainId,
-    relayerFeePct,
-    quoteTimestamp,
-    message,
-    maxCount
-  );
-  const { hash: transactionHash } = deposit;
-  console.log(`Submitting ${tokenSymbol} deposit on ${network}: ${transactionHash}`);
-  const receipt = await deposit.wait();
+  try {
+    const deposit = await (isV2 ? spokePool.depositV2 : spokePool.deposit)(
+      recipient,
+      token.address,
+      amount,
+      toChainId,
+      relayerFeePct,
+      quoteTimestamp,
+      message,
+      maxCount
+    );
+    const { hash: transactionHash } = deposit;
+    console.log(`Submitting ${tokenSymbol} deposit on ${network}: ${transactionHash}`);
+    const receipt = await deposit.wait();
 
-  receipt.logs
-    .filter((log) => log.address === spokePool.address)
-    .forEach((log) => printDeposit(spokePool.interface.parseLog(log), fromChainId));
-  return true;
+    receipt.logs
+      .filter((log) => log.address === spokePool.address)
+      .forEach((log) => printDeposit(spokePool.interface.parseLog(log), fromChainId));
+    return true;
+  } catch (err) {
+    console.error("Deposit failed:", err);
+    console.debug("Deposit Parameters: ", {
+      spokeAddress: spokePool.address,
+      recipient,
+      token: token.address,
+      amount,
+      toChainId,
+      relayerFeePct,
+      quoteTimestamp,
+      message,
+      maxCount,
+    });
+    return false;
+  }
 }
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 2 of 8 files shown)


scripts/spokepool.ts
```diff
@@ -161,26 +161,40 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
...
-  console.log(`Submitting ${tokenSymbol} deposit on ${network}: ${transactionHash}.`);
-  const receipt = await deposit.wait();
-
-  receipt.logs
-    .filter((log) => log.address === spokePool.address)
-    .forEach((log) => printDeposit(spokePool.interface.parseLog(log)));
-
-  return true;
+  try {
+    const deposit = await (isV2 ? spokePool.depositV2 : spokePool.deposit)(
```
pxrl, line 165: Is the `isV2` flag intended to toggle existing behaviour or new? We don't have a `depositV2()` SpokePool function, but we do have `depositV3()`. So I'd expect:

`const deposit = await (isV2 ? spokePool.deposit : spokePool.depositV3)`. However, the deposit signatures are not compatible so I don't really see how this would work.

scripts/spokepool.ts
```diff
@@ -329,10 +344,11 @@ async function run(argv: string[]): Promise<number> {
   const fetchDepositOpts = ["chainId", "depositId"];
   const opts = {
     string: ["wallet", ...configOpts, ...depositOpts, ...fetchOpts, ...fillOpts, ...fetchDepositOpts],
-    boolean: ["decimals"], // @dev tbd whether this is good UX or not...may need to change.
+    boolean: ["decimals", "v2"], // @dev tbd whether this is good UX or not...may need to change.
     default: {
-      wallet: "secret",
+      wallet: "privateKey",
```
pxrl, line 349: I'm not keen on this - `secret` is preferred over `privateKey`. I think `mnemonic` and `privateKey` only make sense in pure serverless environments where the operator isn't touching the filesystem, but that's not the use case for this script.

package.json
```diff
@@ -10,7 +10,7 @@
     "node": ">=16.18.0"
   },
   "dependencies": {
-    "@across-protocol/constants-v2": "1.0.8",
+    "@across-protocol/constants-v2": "1.0.9",
```
pxrl, line 13: This can be bumped straight to 1.0.10 :+1:

scripts/spokepool.ts
pxrl: Could this be inferred by the chainId? i.e. if we only have it on Sepolia then could they be a good switch for whether we use this bespoke function?

src/clients/ProfitClient.ts
```diff
@@ -460,11 +464,14 @@ export class ProfitClient {
...
     // ETH/WETH refer to the same mainnet contract address.
     const tokens: { [_symbol: string]: string } = Object.fromEntries(
-      this.hubPoolClient.getL1Tokens().map(({ symbol }) => {
-        const { addresses } = TOKEN_SYMBOLS_MAP[symbol];
-        const address = addresses[1];
-        return [symbol, address];
-      })
+      this.hubPoolClient
+        .getL1Tokens()
+        .filter(({ symbol }) => isDefined(TOKEN_SYMBOLS_MAP[symbol]))
```
nicholaspai, line 469: Should we throw an error if a symbol is missing rather than silently filter it out?

nicholaspai (approved):
This lgtm. I have one question and then want Paul to review because he's working on these files I'm sure

scripts/spokepool.ts
```diff
@@ -146,6 +146,7 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
   const provider = new ethers.providers.StaticJsonRpcProvider(utils.getProviderUrl(fromChainId));
   signer = signer.connect(provider);
   const spokePool = (await utils.getSpokePoolContract(fromChainId)).connect(signer);
+  const mockSpokePool = (await utils.getSpokePoolContract(fromChainId, true)).connect(signer);
```
mrice32, line 149: I think this mock thing is a little confusing. AFAICT, this isn't really a mock, it's a real, production contract, but we're using a mock to get v2 functions.

Can we create a combined spoke pool interface file in the contracts-v2 repo? Then we can use that abi in situations where we want both v2 and v3 functions/events? Or am I missing an additional point of complexity?

src/relayer/Relayer.ts
```diff
@@ -225,7 +225,10 @@ export class Relayer {
     const mdcPerChain = Object.fromEntries(
       Object.entries(unfilledDepositAmountsPerChain).map(([chainId, unfilledAmount]) => {
         const usdThreshold = minimumDepositConfirmationThresholds.find((_usdThreshold) => {
-          return toBNWei(_usdThreshold).gte(unfilledAmount);
+          return (
+            toBNWei(_usdThreshold).gte(unfilledAmount) &&
+            isDefined(config.minDepositConfirmations[_usdThreshold][chainId])
```
mrice32, line 230: Sorry, can you explain why this added check is needed?

mrice32 (commented):
Looks great! Two comments

src/relayer/Relayer.ts
james-a-morris: Sure! The `minimumDepositConfirmationThresholds` is a list of unique USD threshold values across all chains. Without the check, the `const usdThreshold = ...` will resolve to the threshold that matches `toBNWei(_usdThreshold).gte(unfilledAmount)` for the given deposit. Because we're looking at the global thresholds, the `usdThreshold` may resolve to a value that isn't available for the `chainId`. As a result, `usdThreshold` may be defined and we will attempt to resolve a null reference.

```
 return [chainId, config.minDepositConfirmations[usdThreshold ?? "default"][chainId]];
```

scripts/spokepool.ts
james-a-morris: The contracts deployed on testnet are 1:1 with the MockSpokePool ABI. I'd be a bit apprehensive to combine the testnet ABI with the production ABI in `contracts-v2`. Would it be better if I changed the name to something semantically closer to what we're trying to do?

I.e. `testnetSpokePool`

nicholaspai (approved):
i still have 1 oustanding comment then lgtm

src/relayer/Relayer.ts
pxrl: I ran into exactly this issue when testing on Sepolia fwiw; this change resolved it for me.

scripts/spokepool.ts
pxrl: ooc, will we actually need this post the v3 migration? It was very useful for testing and de-risking the migration plan, but once v3 goes live I wouldn't expect us to need this again. Does that make sense?

src/clients/ProfitClient.ts
pxrl: Pre this change it would implicitly throw by trying to dereference undefined. I think it's preferable to filter because it permits the relayer to run even when a deposit defines a token that's not supported (this is especially annoying on testnet). We could choose to log in the event that a token is not resolvable; that might ease debugging in future.

pxrl (approved):
OK by me (subject to the 2 outstanding comment threads).

In general I think we'd probably look to back out the `scripts/spokepool.ts` changes shortly after because I don't expect it to be used anymore.

src/relayer/Relayer.ts
mrice32: Ah okay, thanks for the explanation!

scripts/spokepool.ts
mrice32: Ah, I see, I think this may be an issue of naming. I was originally thinking we were using `isV2` to decide whether to call the v2 or v3 function, but now I see that my initial read was wrong. We basically have the v2 deposit function called deposit in our prod contracts and depositV2 in the mock contract (why is that?). So we're using `isV2` to decide whether we're talking to the mock or not.

Why not call this flag `isMock` or `isTestnet`?

mrice32 (approved):
LGTM! I did find this naming to be a bit odd, though: https://github.com/across-protocol/relayer-v2/pull/1185#discussion_r1497010720

scripts/spokepool.ts
james-a-morris: Yeah that's more valid. @pxrl on this point I do think as long as we have the code in our relayer we should support generating these events. This is a small enough change where we could remove it as part of the V2 cleaning but for now it's helpful in triggering the V2 codepaths.

src/clients/ProfitClient.ts
james-a-morris: We now emit a `warn` if we find any tokens that we expect to be in the hubpool but don't have a corresponding TOKEN_SYMBOLS_MAP entry.

I chose `warn` because any triggering of this warning should be sent on slack & should be corrected quickly.

scripts/spokepool.ts
james-a-morris: @mrice32 technically we can call `V3` contract events from the testnet.

james-a-morris:
@pxrl @mrice32 I think we can pull out the V2 events afterall. These are now mostly (if not fully) deprecated this far out of the V3 launch.

PR #1187, opened 2024-02-08
feat: Support requesting slow fills for v3 deposits

pxrl:
This commit provides a type-sensitive wrapper around Relayer.zeroFillDeposit() and introduces support for the v3 SpokePool.requestSlowFill() function.

It's not currently possible to test submission of v3 slow fill requests, so the test has been skipped for now, pending a bump to contracts-v2. Implementating that test is a prerequisite for merging this change.


src/relayer/Relayer.ts
```diff
@@ -491,6 +528,10 @@ export class Relayer {
       ? await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)
       : destinationChainId;
 
+    if (sdkUtils.isV3Deposit(deposit)) {
+      return { repaymentChainId: undefined, gasLimit: bnUint256Max };
+    }
```
pxrl, line 533: For reference, this is currently needed to keep the ProfitClient from complaining about V3 types. It effectively tells the relayer that no repayment chain could be identified for the deposit because the deposit is unprofitable and should therefore be ignored. This will be removed once the ProfitClient supports V3 types natively.

src/relayer/Relayer.ts
```diff
@@ -391,6 +391,55 @@ export class Relayer {
...
+
+    const formatSlowFillRequestMarkdown = (): string => {
+      const { symbol, decimals } = hubPoolClient.getTokenInfo(destinationChainId, outputToken);
+      const formatter = createFormatFunction(2, 4, false, decimals);
+      const outputAmount = formatter(deposit.outputAmount);
+      const [srcChain, dstChain] = [getNetworkName(originChainId), getNetworkName(destinationChainId)];
+
+      // @todo (future) infer the updated outputAmount by zeroing the relayer fee in order to print the correct amount.
+      return `Reqested slow fill 🐌 of ${outputAmount} ${symbol}` +
```
nicholaspai, line 428:
```suggestion
      return `Requested slow fill 🐌 of ${outputAmount} ${symbol}` +
```

src/relayer/Relayer.ts
nicholaspai: Got it. For V3 we should be able to call `await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)` since there are no partial fills

nicholaspai (approved):
Looks great, will review again after profit client fixes

mrice32 (approved):
LGTM!

PR #1188, opened 2024-02-09
feat: Update to >= contracts-v2@2.5.0

nicholaspai:


contracts/MockSpokePool.sol
pxrl, line 1: We're not overriding `deposit()` to emit `FundsDeposited` here - is that confirmed not required? Perhaps we only need it in the SDK where we're pulling events off the "wire".

pxrl (approved):
This seems straightforward - OK subject to 1 query about the MockSpokePool contract.

contracts/MockSpokePool.sol
nicholaspai: we don't need it to, since we change the test-utils to call `depositV2` on this contract instead

hardhat.config.ts
```diff
@@ -20,12 +20,12 @@ import "@openzeppelin/hardhat-upgrades";
 
 dotenv.config();
 
-const solcVersion = "0.8.18";
+const solcVersion = "0.8.23";
```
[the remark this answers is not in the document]
nicholaspai: i don't think so?

PR #1192, opened 2024-02-13
refactor: Cleanup contracts-v2 exports in test

pxrl:
As was recently done over in the SDK.


pxrl:
fwiw I think we can do a lot more here, but I saw this as the quick and dirty minimum to get control over the exports for now.

nicholaspai (approved):

PR #1194, opened 2024-02-15
refactor: Remove relayerFeePct ordering in relayer

pxrl:
The comment attached to this ordering says that it's done to prioritise the most profitable deposits first. This is only a very coarse way of sorting because it ignores the amount being bridged. In practice it will probably tend to prioritise low-value transactions (which indirectly _may_ be the most profitable due to gas padding).

Sorting this early isn't compatible with v3 deposits because their profitability is determined later in the relayer run. In any case, ordering deposits by profitability doesn't seem to have much benefit in practice because it should only take effect at the absolute margins where the relayer can fill each deposit individually, but not the sum of them. So...remove it for now.


test/Relayer.TokenShortfall.ts
pxrl, line 1: fwiw I'm not a huge fan of this test hardcoding expectations about the ordering of deposit IDs in a log message. I'm tempted to at least truncate the string comparison to "blocking deposits: " so that the ordering isn't specified, and could perhaps instead verify that the multicaller queue has 0 enqueued transactions.

nicholaspai (approved):

PR #1195, opened 2024-02-15
refactor: Reorganise Relayer test for v3 additions

pxrl:
This change makes room for the addition of v3 tests, which have additional beforeEach() steps. The diff for doing this reorg simultaneously with introducing new functionality is very difficult to read, so they're split up to minimise the pain of reviewers.


pxrl:
To be clear - the vast majority of this diff is simply indenting the code. It's really unfortunate that the diff algorithm isn't showing that more cleanly.

nicholaspai (approved):

PR #1196, opened 2024-02-15
feat(relayer): Support fillV3Relay()

pxrl:
Note: This is currently WIP; there's some remaining work to be done in the ProfitClient tests.


src/relayer/Relayer.ts
```diff
@@ -634,19 +670,28 @@ export class Relayer {
...
+
+      if (fillAmount.eq(zeroFillAmount)) {
+        msg += " Has been zero filled due to a token shortfall! This will initiate a slow relay for this deposit.";
+      }
+    } else {
+      const { symbol: outputTokenSymbol, decimals: outputTokenDecimals } =
+        this.clients.hubPoolClient.getTokenInfoForDeposit(deposit);
+      const outputAmount = deposit.outputAmount;
+      const _outputAmount = createFormatFunction(2, 4, false, outputTokenDecimals)(outputAmount.toString());
+      msg += ` and output ${_outputAmount} ${outputTokenSymbol}, with depositor ${depositor}.`;
```
pxrl, line 694: Note: I'd like to add additional fee information here, but it seems low priority for now. It'll be easier to add in once the fillRelay() is removed...

src/relayer/Relayer.ts
```diff
@@ -254,7 +254,7 @@ export class Relayer {
     // If not enough ballance add the shortfall to the shortfall tracker to produce an appropriate log. If the deposit
     // is has no other fills then send a 0 sized fill to initiate a slow relay. If unprofitable then add the
     // unprofitable tx to the unprofitable tx tracker to produce an appropriate log.
-    for (const { deposit, version, unfilledAmount, fillCount } of confirmedUnfilledDeposits) {
+    for (const { deposit, unfilledAmount, fillCount } of confirmedUnfilledDeposits) {
```
pxrl, line 257: todo: Flagging some missing filtering on `confirmedUnfilledDeposits`:
- If deposit.exclusivityDeadline is ahead of the current timestamp (+ some margin?) _and_ `deposit.exclusiveRelayer` is not set to the relayer's own address, drop the deposit.
- If fillDeadline has already elapsed, drop the deposit.

pxrl: Implemented: https://github.com/across-protocol/relayer-v2/pull/1196/commits/895dc30794f61c507cd3f6452c7d848faea72cfa

src/clients/ProfitClient.ts
```diff
@@ -344,7 +366,8 @@ export class ProfitClient {
     const netRelayerFeePct = netRelayerFeeUsd.mul(fixedPoint).div(relayerCapitalUsd);
 
     // If token price or gas price is unknown, assume the relay is unprofitable.
-    const profitable = tokenPriceUsd.gt(0) && gasTokenPriceUsd.gt(0) && netRelayerFeePct.gte(minRelayerFeePct);
+    const profitable =
+      tokenPriceUsd.gt(bnZero) && gasTokenPriceUsd.gt(bnZero) && netRelayerFeePct.gte(minRelayerFeePct);
```
mrice32, line 370: nit: do we need to check tokenPriceUsd.gt(bnZero)? Don't we already check this on 337?

Same for gasTokenPriceUsd, which we check on 261.

src/relayer/Relayer.ts
```diff
@@ -432,20 +445,57 @@ export class Relayer {
...
-        message: "Skipping deposit already filled by this relayer",
+        message: "Skipping deposit already filled by this relayer.",
         originChainId: deposit.originChainId,
         depositId: deposit.depositId,
       });
       return;
     }
+
+    sdkUtils.isV2Deposit(deposit)
+      ? this.fillV2Relay(deposit, sdkUtils.getDepositOutputAmount(deposit), repaymentChainId, gasLimit)
```
mrice32, line 464: Quick q: Doesn't passing in getDepositOutputAmount as the fillAmount cause a behavior change? If this, for instance, was partially filled and we only needed to fill 50%, we would be assuming we'd have to fill the whole thing and decrementing our in-memory token inventory to account for that, right?

To be clear, if my understanding is correct, I think this is totally fine, since true partial fills are exceedingly rare, but just want to acknowledge the behavior change in case it does unintentionally cause an issue.

src/relayer/Relayer.ts
```diff
@@ -522,21 +570,21 @@ export class Relayer {
       ? await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address)
       : destinationChainId;
 
-    if (sdkUtils.isV3Deposit(deposit)) {
-      return { repaymentChainId: undefined, gasLimit: bnUint256Max };
-    }
+    const { realizedLpFeePct } = sdkUtils.isV3Deposit(deposit)
+      ? await hubPoolClient.computeRealizedLpFeePct(deposit)
```
mrice32, line 574: I may be missing something, but it seems like we're not providing a paymentChainId here, right? So won't this return a 0 fee? https://github.com/across-protocol/sdk-v2/blob/79c0e1a24ec1de914ab67b8a5fa3bdf129b4eea8/src/clients/HubPoolClient.ts#L433

Should we be including the preferredChainId in this request? Or is this being accounted for elsewhere?

mrice32 (commented):
Looks great! Just a few comments and questions

src/relayer/Relayer.ts
pxrl: I have a separate change where that goes in. It's a very small & simple update.

src/clients/ProfitClient.ts
pxrl: Good catch - that redundancy has been hidden there for a while :eyes: 

Flushed: https://github.com/across-protocol/relayer-v2/pull/1196/commits/d6490bfc94d9738bcccb6fb24b722aaaf03ffc30

src/relayer/Relayer.ts
pxrl: Well spotted, and you actually caught a related bug. The implementation as it was only updated balances on v2 fills, not v3 fills. I've corrected that in this commit, and have also restored the previously behaviour that passed in unfilledAmount so that we'd update correctly on partial fills: https://github.com/across-protocol/relayer-v2/pull/1196/commits/b7294e8a4f2d13a11e61e3d7ed9e4c87c953d71e

Thanks!

package.json
```diff
@@ -11,8 +11,8 @@
   },
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.8",
-    "@across-protocol/contracts-v2": "2.5.0-beta.5",
```
nicholaspai, line 14: Can this be bumped to a non beta version?

src/clients/ProfitClient.ts
```diff
@@ -365,81 +387,194 @@ export class ProfitClient {
...
+  ): Promise<V3FillProfit> {
+    const { hubPoolClient } = this;
+
+    const inputTokenInfo = hubPoolClient.getL1TokenInfoForL2Token(deposit.inputToken, deposit.originChainId);
+    const inputTokenPriceUsd = this.getPriceOfToken(inputTokenInfo.symbol);
+    const inputTokenScalar = toBNWei(1, 18 - inputTokenInfo.decimals);
+    const scaledInputAmount = deposit.inputAmount.mul(inputTokenScalar);
+    const inputAmountUsd = scaledInputAmount.mul(inputTokenPriceUsd).div(fixedPoint);
+
+    const outputTokenInfo = hubPoolClient.getL1TokenInfoForL2Token(deposit.outputToken, deposit.destinationChainId);
```
nicholaspai, line 409: Can we make sure we early exit and not throw errors if there are any deposits with an output token that we don't know about or can't compute prices for?

src/relayer/Relayer.ts
```diff
@@ -95,6 +87,29 @@ export class Relayer {
...
+          at: "Relayer::getUnfilledDeposits",
+          message: "Ignoring deposit",
+          depositor,
+          recipient,
+        });
+        return false;
+      }
+
+      // Filters specific to v3.
+      if (sdkUtils.isV3Deposit(deposit)) {
```
nicholaspai, line 101: Should we early exit here if input and output tokens aren't equivalent?

src/clients/ProfitClient.ts
pxrl: Yep - there are two very broad protections against this:
1. The ProfitClient wraps its profitability calculations in a try/catch statement, so _any_ computation error there just results in an unprofitable conclusion.
2. The Relayer itself is imposing its own filtering on inputToken & outputToken, such that it filters them out before they even reach the ProfitClient.

The ProfitClient is capable of computing cross-token profitability now, provided that the HubPoolClient knows about the tokens...but it's effectively administratively disabled for the time being.

james-a-morris (commented):
Left a few comments. Mostly curious about

src/relayer/Relayer.ts
pxrl: I relocated the existing check here: https://github.com/across-protocol/relayer-v2/pull/1196/commits/1b8c972bca67252539d9009ac5d3a8a46ecc1d19.

src/clients/ProfitClient.ts
pxrl: Also updated it here: https://github.com/across-protocol/relayer-v2/pull/1196/commits/1b8c972bca67252539d9009ac5d3a8a46ecc1d19.

src/relayer/Relayer.ts
```diff
@@ -232,7 +245,7 @@ export class Relayer {
       })
     );
     this.logger.debug({
-      at: "Relayer",
+      at: "Relayer::checkForUnfilledDepositsAndFill",
```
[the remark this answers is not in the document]
pxrl: We don't have consistency on :: or # in the relayer so I'll probably defer that to another commit.

src/relayer/Relayer.ts
```diff
@@ -592,11 +638,12 @@ export class Relayer {
         const formatFunction = createFormatFunction(2, 4, false, decimals);
         const gasFormatFunction = createFormatFunction(2, 10, false, 18);
         const depositblockExplorerLink = blockExplorerLink(deposit.transactionHash, deposit.originChainId);
+
+        const inputAmount = formatFunction(sdkUtils.getDepositInputAmount(deposit).toString());
+        const relayerFeePct = formatFeePct(sdkUtils.isV2Deposit(deposit) ? deposit.relayerFeePct : bnZero); // @todo
```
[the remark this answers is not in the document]
pxrl: I've updated the comment to clarify, but we don't have the v3 realizedLpFeePct here and it's annoying to make it available down here, so for now it defaults to 0. This is fine since this is only for very loose informational purposes. When v2 gets pulled out there will be opportunity for a refactor here.

src/relayer/Relayer.ts
```diff
@@ -636,19 +683,28 @@ export class Relayer {
...
-      ` relayerFee ${relayerFeePct}% & realizedLpFee ${realizedLpFeePct}%.`;
-    if (fillAmount.eq(zeroFillAmount)) {
-      msg += " Has been zero filled due to a token shortfall! This will initiate a slow relay for this deposit.";
+    const inputAmount = sdkUtils.getDepositInputAmount(deposit);
+    const _inputAmount = createFormatFunction(2, 4, false, decimals)(inputAmount.toString());
+
+    let msg = `Relayed depositId ${deposit.depositId} from ${srcChain} to ${dstChain} of ${_inputAmount} ${symbol}`;
+    if (sdkUtils.isV2Deposit(deposit)) {
+      const _fillAmount = createFormatFunction(2, 4, false, decimals)(fillAmount.toString());
+      const _relayerFeePct = formatFeePct(deposit.relayerFeePct);
```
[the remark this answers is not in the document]
pxrl: It was for consistency with the other variables that are reproduced/reconstructed from the deposit instance, but tbf I do see there is an inconsistency with realizedLpFeePct.

src/clients/ProfitClient.ts
```diff
@@ -365,81 +387,194 @@ export class ProfitClient {
...
+  }
+
   // Return USD amount of fill amount for deposited token, should always return in wei as the units.
   getFillAmountInUsd(deposit: Deposit, fillAmount: BigNumber): BigNumber {
     const l1TokenInfo = this.hubPoolClient.getTokenInfoForDeposit(deposit);
     if (!l1TokenInfo) {
       const inputToken = sdkUtils.getDepositInputToken(deposit);
       throw new Error(
-        `ProfitClient::isFillProfitable missing l1TokenInfo for deposit with origin token: ${inputToken}`
+        `ProfitClient::getFillAmountInUsd missing l1TokenInfo for deposit with origin token: ${inputToken}`
```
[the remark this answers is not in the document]
pxrl: https://github.com/across-protocol/relayer-v2/pull/1196/commits/0ab6d3790956ea218075bbd2b1a65364b1fc84fa

src/clients/ProfitClient.ts
```diff
@@ -343,8 +365,8 @@ export class ProfitClient {
     const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd).sub(refundFeeUsd);
     const netRelayerFeePct = netRelayerFeeUsd.mul(fixedPoint).div(relayerCapitalUsd);
 
-    // If token price or gas price is unknown, assume the relay is unprofitable.
-    const profitable = tokenPriceUsd.gt(0) && gasTokenPriceUsd.gt(0) && netRelayerFeePct.gte(minRelayerFeePct);
+    // If the gas price is unknown, assume the relay is unprofitable.
+    const profitable = netRelayerFeePct.gte(minRelayerFeePct);
```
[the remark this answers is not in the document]
pxrl: Comment is outdated, thanks :+1:

src/relayer/Relayer.ts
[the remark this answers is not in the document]
james-a-morris: Fair - usually the `_` implies that the variable is not used. However, this is only a nit so feel free to not modify it :)

james-a-morris (approved):
LGTM

PR #1197, opened 2024-02-16
feat: add fill script

mrice32:
Add fill script to spokepool utilities.


nicholaspai (approved):

scripts/spokepool.ts
```diff
@@ -183,6 +198,132 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
...
+  const destinationTokenInfo = utils.resolveToken(originTokenInfo.symbol, destinationChainId);
+  const destinationToken = destinationTokenInfo.address;
+  const realizedLpFeePct = await getLpFeePct({
+    originChainId,
+    destinationChainId,
+    token: originToken,
+    amount: amount.toString(),
+    skipAmountLimit: "true",
+    timestamp: Number(quoteTimestamp.toString()),
+  });
```
james-a-morris, line 270: This should fail if the originChainId is a testnet. Can we allow the user to pass an override?

james-a-morris (approved):
LGTM - one question about using this utility with testnets

scripts/spokepool.ts
pxrl: @james-a-morris Do we have a functioning api endpoint for testnet? We could switch on the chain ID(s) to identify whether the fall into Görli (...deprecated), Sepolia or mainnet, and select the quoting endpoint from there.

mrice32:
@pxrl since you've written most of the other scripts, I'm going to wait on merging until you review this. Just want to make sure it aligns with your expectations.

pxrl (approved):
@mrice32 Sorry - these two comments were pending for over a week because I forgot to submit them :\

Aside from that, LGTM. We will fast-follow with a PR to update it once fillRelay is deprecated though :+1:

scripts/spokepool.ts
mrice32: Yeah, I think it might be preferable to always guarantee that this gives you a correct fill rather than allowing overrides just to avoid accidental loss of funds.

This also may be a moot point considering all of this is built for v2, where LP fee is a part of the fill, but now that we're on v3, it no longer is needed. In a follow up, we'll remove the LP fee entirely and support v3 only I suspect.

PR #1198, opened 2024-02-16
feat(proposer): store bundle to arweave

james-a-morris:
This PR adds a persisted state to Arweave based on configurations in the data worker.

Three new environment variables have been added to the dataworker, these include:
1. PERSIST_BUNDLES_TO_ARWEAVE
2. ARWEAVE_WALLET_JWK
3. ARWEAVE_GATEWAY

New JWK wallets can be generated with the following snippet:
```
node -e "require('arweave').init({}).wallets.generate().then(JSON.stringify).then(console.log.bind(console))" > wallet.json
```

For testing, it is recommended to set
```
ARWEAVE_GATEWAY='{ "url": "localhost", "port": 1984, "protocol": "http" }'
```
and use `npx arlocal` to simulate an arweave-like server.

The wallet generated can be funded for use with `arlocal` via the following GET calls:
```
curl http://localhost:1984/mint/{arweave address}/{amountToMint}
curl http://localhost:1984/mine
```

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -503,6 +503,20 @@ export class Dataworker {
       unexecutableSlowFills,
       expiredDepositsToRefundV3,
     } = await this.clients.bundleDataClient.loadData(blockRangesForProposal, spokePoolClients, logData);
+
+    const arweaveBundleStorageHash = await this.clients.arweaveClient.set(
+      {
+        bundleStorageVersion: "1",
+        bundleBlockRanges: blockRangesForProposal,
+        bundleDepositsV3,
+        expiredDepositsToRefundV3,
+        bundleFillsV3,
+        unexecutableSlowFills,
+        bundleSlowFillsV3,
+      },
+      "bundle"
+    );
+
     const allValidFillsInRange = getFillsInRange(
       allValidFills,
       blockRangesForProposal,
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -71,6 +71,15 @@ type RootBundle = {
   tree: MerkleTree<SlowFillLeaf>;
 };
 
+type DataToPersistToDALayerType = {
+  bundleBlockRanges: number[][];
+  bundleDepositsV3: BundleDepositsV3;
+  expiredDepositsToRefundV3: ExpiredDepositsToRefundV3;
+  bundleFillsV3: BundleFillsV3;
+  unexecutableSlowFills: BundleExcessSlowFills;
+  bundleSlowFillsV3: BundleSlowFills;
+};
+
 type ProposeRootBundleReturnType = {
   poolRebalanceLeaves: PoolRebalanceLeaf[];
   poolRebalanceTree: MerkleTree<PoolRebalanceLeaf>;
@@ -78,6 +87,7 @@ type ProposeRootBundleReturnType = {
   relayerRefundTree: MerkleTree<RelayerRefundLeaf>;
   slowFillLeaves: SlowFillLeaf[];
   slowFillTree: MerkleTree<SlowFillLeaf>;
+  dataToPersistToDALayer: DataToPersistToDALayerType;
 };
 
 export type PoolRebalanceRoot = {
@@ -370,7 +380,7 @@ export class Dataworker {
     usdThresholdToSubmitNewBundle?: BigNumber,
     submitProposals = true,
     earliestBlocksInSpokePoolClients: { [chainId: number]: number } = {}
-  ): Promise<void> {
+  ): Promise<DataToPersistToDALayerType> {
     // TODO: Handle the case where we can't get event data or even blockchain data from any chain. This will require
     // some changes to override the bundle block range here, and loadData to skip chains with zero block ranges.
     // For now, we assume that if one blockchain fails to return data, then this entire function will fail. This is a
@@ -482,6 +492,7 @@ export class Dataworker {
         rootBundleData.slowFillTree.getHexRoot()
       );
     }
+    return rootBundleData.dataToPersistToDALayer;
   }
 
   async _proposeRootBundle(
@@ -504,18 +515,18 @@ export class Dataworker {
       expiredDepositsToRefundV3,
     } = await this.clients.bundleDataClient.loadData(blockRangesForProposal, spokePoolClients, logData);
 
-    const arweaveBundleStorageHash = await this.clients.arweaveClient.set(
-      {
-        bundleStorageVersion: "1",
-        bundleBlockRanges: blockRangesForProposal,
-        bundleDepositsV3,
-        expiredDepositsToRefundV3,
-        bundleFillsV3,
-        unexecutableSlowFills,
-        bundleSlowFillsV3,
-      },
-      "bundle"
-    );
+    // Prepare information about what we need to store to
+    // Arweave for the bundle. We will be doing this at a
+    // later point so that we can confirm that this data is
+    // worth storing.
+    const dataToPersistToDALayer = {
+      bundleBlockRanges: blockRangesForProposal,
+      bundleDepositsV3,
+      expiredDepositsToRefundV3,
+      bundleFillsV3,
+      unexecutableSlowFills,
+      bundleSlowFillsV3,
+    };
 
     const allValidFillsInRange = getFillsInRange(
       allValidFills,
@@ -590,6 +601,7 @@ export class Dataworker {
       relayerRefundTree: relayerRefundRoot.tree,
       slowFillLeaves: slowRelayRoot.leaves,
       slowFillTree: slowRelayRoot.tree,
+      dataToPersistToDALayer,
     };
   }
 
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -492,7 +492,7 @@ export class Dataworker {
         rootBundleData.slowFillTree.getHexRoot()
       );
     }
-    return rootBundleData.dataToPersistToDALayer;
+    return submitProposals ? rootBundleData.dataToPersistToDALayer : undefined;
   }
 
   async _proposeRootBundle(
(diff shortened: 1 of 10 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -505,6 +505,20 @@ export class Dataworker {
       unexecutableSlowFills,
       expiredDepositsToRefundV3,
     } = await this.clients.bundleDataClient.loadData(blockRangesForProposal, spokePoolClients, logData);
+
+    const arweaveBundleStorageHash = await this.clients.arweaveClient.set(
```
nicholaspai, line 509: I don't think we should be posting this so early in the proposer run in case the proposer later fails for some reason we will have needlessly spent Arweave txn fees on a bundle  data that we won't be able to use for a valid proposal txn. I recommend moving this as late as possible in the proposal process. Possible in a new statement [here](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/index.ts#L148):

```ts
if (config.proposerEnabled && config.sendingProposalsEnabled) {
        await dataworker.postBundleData(
          bundleData
        );
} 
```

This might require changing the function signature of `proposeRootBundle` to return the bundle data used to post new data.

james-a-morris: Done https://github.com/across-protocol/relayer-v2/pull/1198/commits/503fe60423c9f372dcd12b9e77fecfbb14b93ffb

src/dataworker/Dataworker.ts
```diff
@@ -484,6 +494,7 @@ export class Dataworker {
         rootBundleData.slowFillTree.getHexRoot()
       );
     }
+    return submitProposals ? rootBundleData.dataToPersistToDALayer : undefined;
```
nicholaspai, line 497: Why should this be conditional on the submitProposals flag? I think itd be easier to always return the root bundle data and leave it to the caller to decide how to deal with it

src/dataworker/index.ts
```diff
@@ -107,12 +116,19 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
           config.rootBundleExecutionThreshold,
           config.sendingProposalsEnabled,
           fromBlocks
         );
+        if (
+          config.sendingProposalsEnabled &&
+          isDefined(persistedProposerData) &&
+          Object.keys(persistedProposerData).length > 0
+        ) {
+          dataToPersist = { ...(dataToPersist ?? {}), proposer: persistedProposerData };
```
nicholaspai, line 130: I think it's a bit bulky to make dataToPersist be optional. Instead, assume it's always defined and then post it if the arweave flag is set true

src/dataworker/index.ts
```diff
@@ -146,6 +162,18 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
 
       await clients.multiCallerClient.executeTransactionQueue();
 
+      // Only if we get to this point & we've sent a proposal do we want to persist the data.
+      if (config.persistingBundlesEnabled && isDefined(dataToPersist) && Object.keys(dataToPersist).length > 0) {
```
nicholaspai, line 166: If it's not enabled, can we still simulate the txn posting? This would be useful for seeing what the data would look like, perhaps posted to a testnet, before posting to mainnet.

src/dataworker/DataworkerConfig.ts
```diff
@@ -150,5 +166,22 @@ export class DataworkerConfig extends CommonConfig {
         `dataworkerFastStartBundle=${this.dataworkerFastStartBundle} should be >= dataworkerFastLookbackCount=${this.dataworkerFastLookbackCount}`
       );
     }
+
+    this.persistingBundlesEnabled = PERSIST_BUNDLES_TO_ARWEAVE === "true";
+    assert(
+      !this.persistingBundlesEnabled || this.sendingProposalsEnabled,
```
nicholaspai, line 172: Why should these variables be coupled at the config level? I don't think its infeasible we'd want to post data to arweave but not propose a bundle. Unlikely, sure. But i don't see a benefit to adding a safety check here

src/dataworker/index.ts
nicholaspai, line 166: Why do you think we should refuse to post to arweave if there are no keys in the bundle? Its highly unlikely but technically possible that a bundle block range has NO events and therefore the bundle data is truly empty, yet still a valid bundle.

src/dataworker/Dataworker.ts
james-a-morris: I was trying to avoid persisting data if we didn't have a proposal set. Reverted: https://github.com/across-protocol/relayer-v2/pull/1198/commits/e53427bc779cf4796c8bc771bc939f45d48f1712

src/dataworker/DataworkerConfig.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/1198/commits/ed9a94145cb4066395ee5f94a95d92b38e66f8db

src/dataworker/index.ts
james-a-morris: This check avoids persisting an empty object to Arweave to save on gas costs. However, it's been resolved here: https://github.com/across-protocol/relayer-v2/pull/1198/commits/c3934c99ac0975788f246490e57671b1a20065be

src/dataworker/index.ts
james-a-morris: `dataToPersist` is now defined as `{}` to start. We no longer need the optional.

src/dataworker/index.ts
james-a-morris: Testing can be done via `arlocal` which is a local testnet. You'd need to start via `npx arlocal` and set the `ARWEAVE_GATEWAY` to this url.

Would this be for prod if persistence is disabled?

src/dataworker/index.ts
```diff
@@ -107,12 +108,13 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
 
       if (config.proposerEnabled) {
-        await dataworker.proposeRootBundle(
+        const persistedProposerData = await dataworker.proposeRootBundle(
           spokePoolClients,
           config.rootBundleExecutionThreshold,
           config.sendingProposalsEnabled,
           fromBlocks
         );
+        dataToPersist = { ...dataToPersist, proposer: persistedProposerData };
```
nicholaspai, line 117: Do you anticipate us persisting more data besides what is proposed?

src/dataworker/index.ts
```diff
@@ -146,6 +148,18 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
 
       await clients.multiCallerClient.executeTransactionQueue();
 
+      // If we have data and persisting is enabled, post it to Arweave.
+      if (config.persistingDataEnabled && Object.keys(dataToPersist).length > 0) {
```
nicholaspai, line 152: I think we can remove this keys check for dataToPropose since it'll always have keys right?

src/dataworker/index.ts
```diff
@@ -146,6 +148,18 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+        const hashTxn = await clients.arweaveClient.set(dataToPersist, "proposal");
```
nicholaspai, line 153: Can you explain more what tags will be associated with this txn? Moreover how do you think we will best be able to load historical bundle data posted to arweave?

nicholaspai, line 153: Just had a thought. Should we include the bundle block range as a tag or something unique to this bundle data? This way we can skip this step if there is bundle data for this block range already posted by our account?

This could save us in the case where bundle data is posted but the proposal reverts for some reason. Or in the case of a dispute

src/dataworker/Dataworker.ts
nicholaspai: i think we can handle this in `index.ts`

src/dataworker/index.ts
nicholaspai: hmm is there no testnet to send to? 

Any ways I think this is fine as is

nicholaspai: i don't think we'll ever have an empty object since propose() will always return an object

src/dataworker/Dataworker.ts
james-a-morris: Agreed! I'll close this comment as we're handling this in index.ts

src/dataworker/index.ts
james-a-morris: Sounds good. There's no official testnet but the documentation/community recommends ARLocal

src/dataworker/index.ts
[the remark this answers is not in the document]
james-a-morris: All good! That's fair - if we want Arweave in the critical path then we should move it before the multicall: https://github.com/across-protocol/relayer-v2/pull/1198/commits/e570b0288997e93007a272f5fcb2b863318bde79

james-a-morris: That's a great point. I've added logic for this in the SDK: https://github.com/across-protocol/sdk-v2/pull/562#issue-2142675246.

Once that's merged we can use it here

src/dataworker/index.ts
james-a-morris: I don't expect it in the short term. However, since we're building the object to persist in `index.ts` we should probably leave space.

src/dataworker/index.ts
james-a-morris: Done https://github.com/across-protocol/relayer-v2/pull/1198/commits/e570b0288997e93007a272f5fcb2b863318bde79

src/dataworker/index.ts
james-a-morris: Each txn submitted through the ArweaveClient has an App-Name tag, an App-Version tag, and an optional custom tag. We can load this data via the `getDataByTopic( )` https://github.com/across-protocol/sdk-v2/pull/562

src/dataworker/index.ts
james-a-morris: Done. There is a check for TSC so that it doesn't complain.

src/dataworker/index.ts
nicholaspai: I'm asking because I don't see the usefulness of storing the data at a nested key `proposer`

nicholaspai (approved):

PR #1201, opened 2024-02-20
refactor: Minor InventoryClient simplifications

pxrl:
No functional change intended.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1,4 +1,4 @@
-import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { constants, utils as sdkUtils } from "@across-protocol/sdk-v2";
 import {
   bnZero,
   BigNumber,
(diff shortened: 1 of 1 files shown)


pxrl:
These are just a few tweaks I picked up over recent weeks working on v3.

src/clients/InventoryClient.ts
```diff
@@ -110,11 +113,11 @@ export class InventoryClient {
   getCurrentAllocationPct(l1Token: string, chainId: number): BigNumber {
     // If there is nothing over all chains, return early.
     const cumulativeBalance = this.getCumulativeBalance(l1Token);
-    if (cumulativeBalance.eq(0)) {
+    if (cumulativeBalance.eq(bnZero)) {
       return bnZero;
     }
 
-    const shortfall = this.getTokenShortFall(l1Token, chainId) || bnZero;
```
pxrl, line 117: `this.getTokenShortFall()` returns a BN so I can't see how the or-case could ever be needed.

Digging deeper, this resolves to TokenClient.getTokenShortFallRequirement, which internalises the 0 case. So we can rely on the types here.

```
    getShortfallTotalRequirement(chainId: number, token: string): BigNumber {                                                                                                                                                                                  
      return this.tokenShortfall?.[chainId]?.[token]?.totalRequirement ?? bnZero;                                                                                                                                                                              
    }
```

PR #1204, opened 2024-02-20
[WIP] feat: Update LP fee computation

pxrl:
Permit realizedLpFeePct computation to occur between originChainId and
any arbitrary chain, rather than destinationChainId. This supports more
dynamic fee behaviour.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -12,7 +12,7 @@ import {
   V2DepositWithBlock,
   V2FillWithBlock,
 } from "../interfaces";
-import { SpokePoolClient } from "../clients";
+import { HubPoolClient, LpFeeRequest, SpokePoolClient } from "../clients";
 import {
   winston,
   BigNumber,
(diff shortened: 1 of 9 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -647,92 +672,83 @@ export class BundleDataClient {
         // Keep track of fast fills that replaced slow fills, which we'll use to create "unexecutable" slow fills
         // if the slow fill request was sent in a prior bundle.
         const fastFillsReplacingSlowFills: string[] = [];
-        await utils.forEachAsync(
```
pxrl, line 650: Note: Most of this change is imposed by prettier. It's re-indenting everything because the two arguments to `forEachAsync()` can be grouped on a single line, rather than multiple lines. Scary.

pxrl:
Also note, this PR from sdk-v2 is required. It can be linked locally via yarn, though I am suspicious of some unrelated test failures showing up in that configuration due to some BigNumber shenanigans.

nicholaspai (commented):
https://github.com/across-protocol/relayer-v2/pull/1204/files#r1496443067

pxrl:
Implemented in Nick's PR.

PR #1208, opened 2024-02-21
fix: Update SDK so HubPoolClient.batchCompute handles multiple origin chains

nicholaspai:
I also added a test for this scenario

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -99,6 +99,7 @@ export async function blockRangesAreInvalidForSpokeClients(
       blockRanges,
       chainIdListForBundleEvaluationBlockNumbers
     );
+    assert(Object.keys(endBlockTimestamps).length === Object.values(spokePoolClients).filter(isDefined).length);
   }
   return utils.someAsync(blockRanges, async ([start, end], index) => {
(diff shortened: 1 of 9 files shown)


pxrl (approved):
LGTM.

src/dataworker/DataworkerUtils.ts
```diff
@@ -129,7 +129,7 @@ export async function blockRangesAreInvalidForSpokeClients(
     }
 
     if (endBlockTimestamps !== undefined) {
-      assert(Object.keys(endBlockTimestamps).length === blockRanges.length);
+      assert(Object.keys(endBlockTimestamps).length === Object.values(spokePoolClients).filter(isDefined).length);
```
james-a-morris, line 132: +1. But shouldn't we already filter `null/undefined` `spokePoolClient` instances?

james-a-morris (approved):
LGTM - one comment

src/dataworker/DataworkerUtils.ts
nicholaspai: yeah but that's not what we're checking. Anyways I moved this check up higher https://github.com/across-protocol/relayer-v2/pull/1208/commits/3671dd82e9c6b97f8cd2afe70d32c03a3291fab6

PR #1215, opened 2024-02-21
fix: Dataworker shouldn't call fillDeadlineBuffer for bundles whose start block is before V3 upgrade block

nicholaspai:
Reverts across-protocol/relayer-v2#1214

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -123,8 +123,9 @@ export class Dataworker {
     }
   }
 
-  isV3(): boolean {
-    return sdk.utils.isV3(this.clients.configStoreClient.configStoreVersion);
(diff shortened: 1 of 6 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -347,7 +354,7 @@ export class Dataworker {
         blockRangesForProposal,
         this.chainIdListForBundleEvaluationBlockNumbers,
         earliestBlocksInSpokePoolClients,
-        this.isV3()
+        this.isV3(mainnetBlockRange[0])
```
mrice32, line 357: Won't this mean that if the v3 upgrade happens in the middle of the block range, the bundle will be considered !v3? Does that mean it will ignore v3 events?

mrice32 (commented):
Looks great! One question

PR #1216, opened 2024-02-21
improve(relayer): Check v3 fillstatus before filling

pxrl:


src/relayer/Relayer.ts
```diff
@@ -55,7 +55,7 @@ export class Relayer {
     const unfilledDeposits = await getUnfilledDeposits(spokePoolClients, hubPoolClient, this.config.maxRelayerLookBack);
 
     const maxVersion = configStoreClient.configStoreVersion;
-    return unfilledDeposits.filter(({ deposit, version, invalidFills, unfilledAmount }) => {
+    return sdkUtils.filterAsync(unfilledDeposits, async ({ deposit, version, invalidFills, unfilledAmount }) => {
```
nicholaspai, line 58: Do we need to `await` this?

nicholaspai (approved):
one comment

PR #1220, opened 2024-02-21
fix: Restore unprofitable fill reporting

pxrl:
It was easier to re-add this than first thought.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -601,7 +601,7 @@ export class Relayer {
       profitable,
       nativeGasCost: gasLimit,
       tokenGasCost: gasCost,
-      grossRelayerFeePct: relayerFeePct, // gross relayer fee is equal to total fee minus the lp fee. 
+      grossRelayerFeePct: relayerFeePct, // gross relayer fee is equal to total fee minus the lp fee.
     } = await profitClient.isFillProfitable(deposit, fillAmount, realizedLpFeePct, hubPoolToken);
     // If preferred chain is different from the destination chain and the preferred chain
     // is not profitable, then check if the destination chain is profitable.
@@ -657,10 +657,11 @@ export class Relayer {
         // destination chain, therefore we will fill them using the original preferred chain to maintain
         // inventory assumptions and also quote the original relayer fee pct.
         return {
-          gasLimit,
           repaymentChainId: preferredChainId,
           realizedLpFeePct,
           relayerFeePct,
+          gasCost,
+          gasLimit,
         };
       } else {
         // If preferred chain is not profitable and neither is fallback, then return the original profitability result.
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -307,7 +307,7 @@ export class Relayer {
           realizedLpFeePct,
           relayerFeePct,
           gasLimit: _gasLimit,
-          tokenGasCost: gasCost,
+          gasCost,
         } = await this.resolveRepaymentChain(deposit, unfilledAmount, l1Token);
         if (isDefined(repaymentChainId)) {
           const gasLimit = isMessageEmpty(resolveDepositMessage(deposit)) ? undefined : _gasLimit;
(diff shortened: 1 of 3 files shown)


src/relayer/Relayer.ts
```diff
@@ -652,11 +662,12 @@ export class Relayer {
 
         const inputAmount = formatFunction(sdkUtils.getDepositInputAmount(deposit).toString());
 
-        // @todo Consider whether to add v3 realizedLpFee logging. Use 0 for now.
-        const relayerFeePct = formatFeePct(sdkUtils.isV2Deposit(deposit) ? deposit.relayerFeePct : bnZero);
+        const gasCost = gasFormatFunction(_gasCost.toString());
```
nicholaspai, line 665: Does this fix the issue where gas cost looks like Xe-12 ?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: That was fixed here: https://github.com/across-protocol/relayer-v2/pull/1220/commits/d0ea19fc642e3a31649026c9e69342efacf44205

As it turns out, this was subtly broken back in November when the meaning of "gas cost" shifted from "gas cost in units of gas token" to "gas cost in units of gas".

nicholaspai (commented):
Can you share an example of the slack log?

src/relayer/Relayer.ts
```diff
@@ -638,12 +643,17 @@ export class Relayer {
     Object.keys(unprofitableDeposits).forEach((chainId) => {
       let depositMrkdwn = "";
       Object.keys(unprofitableDeposits[chainId]).forEach((depositId) => {
-        const { deposit, fillAmount, gasCost: _gasCost } = unprofitableDeposits[chainId][depositId];
+        const {
+          deposit,
+          fillAmount,
+          lpFeePct: _lpFeePct,
+          relayerFeePct: _relayerFeePct,
+          gasCost: _gasCost,
```
james-a-morris, line 651: OOC why do we map this?

src/relayer/Relayer.ts
james-a-morris: I believe that the format function is just a nested function around the formatUnits function

james-a-morris, line 665: Would there be any push to rename this to something like: "formattedGasCost"? We use gasCost as a BigNumber elsewhere

src/relayer/Relayer.ts
```diff
@@ -652,11 +662,12 @@ export class Relayer {
...
 
-        // @todo Consider whether to add v3 realizedLpFee logging. Use 0 for now.
-        const relayerFeePct = formatFeePct(sdkUtils.isV2Deposit(deposit) ? deposit.relayerFeePct : bnZero);
+        const gasCost = gasFormatFunction(_gasCost.toString());
+        const relayerFeePct = formatFeePct(_relayerFeePct);
+        const lpFeePct = formatFeePct(_lpFeePct);
         depositMrkdwn +=
           `- DepositId ${deposit.depositId} (tx: ${depositblockExplorerLink}) of amount ${inputAmount} ${symbol}` +
-          ` with a relayerFeePct ${relayerFeePct}% and gas cost ${gasFormatFunction(gasCost)}` +
+          ` with a relayerFeePct ${relayerFeePct}%, lpFeePct ${lpFeePct}, and gas cost ${gasFormatFunction(gasCost)}` +
```
james-a-morris, line 670: Should we use the gasCost variable above?

james-a-morris (commented):
Left some reviews

src/relayer/Relayer.ts
pxrl: `gasCost` is used down at line 774 (i.e. human-readable gasCost as a string, rather than raw gasCost as a BN).

src/relayer/Relayer.ts
pxrl: Not a bad idea :+1:

src/relayer/Relayer.ts
pxrl: Yep; I think that was the result of a bad merge. Fixed here: https://github.com/across-protocol/relayer-v2/pull/1220/commits/891fde892414ceb5aefbd7b93cb1c83241788a8c.

src/relayer/Relayer.ts
pxrl: Updated: https://github.com/across-protocol/relayer-v2/pull/1220/commits/891fde892414ceb5aefbd7b93cb1c83241788a8c.

pxrl:
> Can you share an example of the slack log?

Here's the first instance of it printing the wrong fee. The units printed are raw gas (so ~80k - 250k ballpark), but the formatting function assumes it's in units of gas token.

first

pxrl:
And here's the new logging:

update

PR #1221, opened 2024-02-22
fix(BundleDataClient): Throw away events before bundle end blocks

nicholaspai:
This is currently causing bugs where future events (such as fast fills replacing slow fill requests) are affecting bundle data originally proposed before these events were known

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -596,6 +596,7 @@ export class BundleDataClient {
 
         originClient
           .getDepositsForDestinationChain(destinationChainId)
+          .filter((deposit) => deposit.blockNumber <= originChainBlockRange[1])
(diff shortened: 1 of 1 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -657,6 +658,7 @@ export class BundleDataClient {
         await utils.forEachAsync(
           destinationClient
             .getFillsForOriginChain(originChainId)
+            .filter((fill) => fill.blockNumber <= destinationChainBlockRange[1])
```
pxrl, line 661: Does this mean we'll also propose a slow fill for a slow fill request even if we could technically see that it was subsequently replaced by a fast fill outside of the proposal window?

pxrl (approved):

src/clients/BundleDataClient.ts
nicholaspai: yeah that's what I would want to happen

PR #1223, opened 2024-02-22
improve(SpokePoolClient): Fix re-export

pxrl:
Extending and exporting a the SpokePoolClient class frequently causes tsc warnings to be flagged when running with a locally linked sdk-v2. The errors seem to be anchored in tsc not resolving the inherited parts of the SpokePoolClient correctly - it seems to view the SpokePoolClient as an opaque class that only contains the locally-overridden methods.

Now that it's physically impossible for an early v2 deposit event to occur, remove the override and revert to simply exporting directly.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,4 +1,0 @@
-import { clients } from "@across-protocol/sdk-v2";
-
-export type SpokePoolClient = clients.SpokePoolClient;
-export const { SpokePoolClient } = clients;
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

james-a-morris (approved):
Ah fair! Can we take this one step further and remove the need for this file altogether? I.e. we can place the export in `src/clients`?

pxrl:
> Ah fair! Can we take this one step further and remove the need for this file altogether? I.e. we can place the export in `src/clients`?

Implemented https://github.com/across-protocol/relayer-v2/pull/1223/commits/50d8afb9f3ff979ed8ab8d03d1bdb43d729e773e

PR #1226, opened 2024-02-22
improve: Add nonce-reused error to known list of error reasons

nicholaspai:
Will filter these type of errors out of error-level logs

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -22,8 +22,7 @@ import lodash from "lodash";
 // Use this list of Smart Contract revert reasons to filter out transactions that revert in the
 // Multicaller client's simulations but that we can ignore. Check for exact revert reason instead of using
 // .includes() to partially match reason string in order to not ignore errors thrown by non-contract reverts.
-// For example, a NodeJS error might result in a reason string that includes more than just the contract r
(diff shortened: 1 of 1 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -24,7 +24,13 @@ import lodash from "lodash";
 // .includes() to partially match reason string in order to not ignore errors thrown by non-contract reverts.
 // For example, a NodeJS error might result in a reason string that includes more than just the contract r
 // evert reason.
-export const knownRevertReasons = new Set(["relay filled", "Already claimed", "RelayFilled", "ClaimedMerkleLeaf"]);
+export const knownRevertReasons = new Set([
+  "nonce has already been used",
+  "relay filled",
+  "Already claimed",
+  "RelayFilled",
+  "ClaimedMerkleLeaf",
```
james-a-morris, line 32: OOC should we be making this case-insensitive?

james-a-morris (approved):
LGTM - one question about casing

src/clients/MultiCallerClient.ts
nicholaspai: hmm yeah we should, wanna propose that in separate pr?

PR #1232, opened 2024-02-23
refactor: Update Monitor test to v3

pxrl:
While here, remove the unknown relayer reporting feature. This has been disabled in prod for quite a while, and given that there are so many active relayers now there's no apparent use case for it.

### DIFF
--- a/test/Monitor.ts
+++ b/test/Monitor.ts
@@ -1,3 +1,4 @@
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import {
   BalanceAllocator,
   BundleDataClient,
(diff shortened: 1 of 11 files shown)


test/Monitor.ts
```diff
@@ -161,30 +181,6 @@ describe("Monitor", async function () {
...
-      destinationChainId,
-      amountToDeposit
-    );
-    await buildFillForRepaymentChain(spokePool_2, depositor, deposit, 0.5, constants.destinationChainId);
-
-    await monitorInstance.update();
-    await monitorInstance.checkUnknownRelayers();
-    expect(lastSpyLogIncludes(spy, unknownRelayerMessage)).to.be.true;
-  });
-
```
pxrl, line 187: Removing this test because this is a feature we simply don't need anymore. I'd propose to remove the functionality as part of this PR as well, unless there are any objections.

nicholaspai (approved):

PR #1240, opened 2024-02-26
refactor(relayer): Assume only v3 deposits

pxrl:
Since the v3 launch is now > 1 day old, no operational relayers should be seeing new v2 events. This commit starts the process of stripping out v2 support.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -56,8 +56,8 @@ export class Relayer {
     const unfilledDeposits = await getUnfilledDeposits(spokePoolClients, hubPoolClient, this.config.maxRelayerLookBack);
 
     const maxVersion = configStoreClient.configStoreVersion;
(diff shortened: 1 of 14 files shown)


src/relayer/Relayer.ts
```diff
@@ -84,7 +84,7 @@ export class Relayer {
       }
 
       // Skip deposits with quoteTimestamp in the future (impossible to know HubPool utilization => LP fee cannot be computed).
-      if (quoteTimestamp > hubPoolClient.currentTime) {
+      if (deposit.quoteTimestamp > hubPoolClient.currentTime) {
```
nicholaspai, line 87: Why this change?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: I exchanged it for `inputToken` and `outputToken` up on line 60. If we extract them all then it overflows to > 1 line and becomes one line per variable.

PR #1241, opened 2024-02-26
chore: Drop Dataworker v2 deposit-related events

pxrl:
The Dataworker lookbacks are now short enough that they won't see any FundsDeposited events. RequestedSpeedUpEvents actually seem irrelevant here as well. We can technically still see these, but they're not used for validating fills.


nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #1242, opened 2024-02-27
fix(config): Incorrect quoting on error message

james-a-morris:
This is a precursor to getting espresso configured. It defines out several constants for sepolia itself and fixes a logging bug.


src/common/Constants.ts
```diff
@@ -119,6 +119,7 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
   80001: 10000,
   84531: 10000,
```
pxrl, line 120:
```suggestion
  84531: 10000,
  84532: 10000,
```

src/common/Constants.ts
```diff
@@ -136,6 +137,7 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   80001: 0,
   84531: 0,
```
pxrl, line 138:
```suggestion
  84531: 0,
  84532: 0,
```

pxrl (approved):
Can we also add Base Sepolia?

pxrl:
@james-a-morris I merged in the latest master and it's reduced this diff to correcting a single error in an exception :+1:

PR #1250, opened 2024-03-01
fix: Handle case where there is pool rebalance leaf for chain+token but no refund leaf, only a slow fill leaf

nicholaspai:


src/scripts/validateRunningBalances.ts
```diff
@@ -367,11 +367,23 @@ export async function runScript(_logger: winston.Logger, baseSigner: Signer): Pr
...
             // We shouldn't get here for any bundle since we start with the i-1'th most recent bundle.
             // If so, then a relayed root message might have gotten stuck in a canonical bridge and we will
             // want to know about it.
-            throw new Error(`No relayed root for chain ID ${leaf.chainId} and token ${l2Token}`);
+            throw new Error(
+              `No relayed refund root for chain ID ${leaf.chainId} and token ${l2Token} with netSendAmount ${fromWei(
+                netSendAmount.toString(),
+                decimals
+              )}`
+            );
```
pxrl, line 386:
```suggestion
            const formattedAmount = fromWei(netSendAmount.toString(), decimals);
            throw new Error(
              `No relayed refund root for chain ID ${leaf.chainId} and token ${l2Token} with netSendAmount ${formattedAmount}`
            );
```

pxrl (approved):

PR #1251, opened 2024-03-01
improve: Executor should not execute Polygon slow fills and relayer refunds atomically

nicholaspai:


src/dataworker/index.ts
```diff
@@ -134,12 +134,22 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
-        await dataworker.executeRelayerRefundLeaves(
-          spokePoolClients,
-          balanceAllocator,
-          config.sendingExecutionsEnabled,
-          fromBlocks
-        );
+
+        // The Polygon_SpokePool cannot execute relayer refunds and fills in the same atomic transaction so
+        // skip the following if there are any polygon transactions in the queue.
+        if (clients.multiCallerClient.getQueuedTransactions(137).length > 0) {
```
james-a-morris, line 140: Would this imply that slow relays on polygon block executing relayer refund leaves on all spokes?

nicholaspai: it would yeah good call i have a better solution pushed now https://github.com/across-protocol/relayer-v2/pull/1251/commits/a31d83f1abebd0696f200f3c252fef6fcf30e895

mrice32 (approved):
LGTM!

PR #1252, opened 2024-03-01
feat: add a block distance after which we set no TTL in redis

mrice32:
This changes the behavior of the provider caching system s.t. after a certain per-chain block distance in the past, we add no TTL so the cache stores it forever until it is forcibly evicted by redis's max mem eviction policy.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -194,6 +194,20 @@ export const CHAIN_CACHE_FOLLOW_DISTANCE: { [chainId: number]: number } = {
   11155420: 0,
 };
 
+// This is the block distance at which the bot, by default, stores in redis with no TTL.
+// These are all intended to be roughly 2 days of blocks for each chain.
+// blocks = 172800 / avg_block_time
+export const DEFAULT_NO_TTL_DISTANCE: { [chainId: number]: number } = {
+  1: 14400,
+  10: 86400,
+  137: 86400,
+  288: 86400,
+  324: 172800,
+  8453: 86400,
+  42161: 691200,
+  534352: 57600,
+};
+
 // Reasonable default maxFeePerGas and maxPriorityFeePerGas scalers for each chain.
 export const DEFAULT_GAS_FEE_SCALERS: {
   [chainId: number]: { maxFeePerGasScaler: number; maxPriorityFeePerGasScaler: number };
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -10,6 +10,7 @@ import {
   PROVIDER_CACHE_TTL,
   PROVIDER_CACHE_TTL_MODIFIER as ttl_modifier,
   BLOCK_NUMBER_TTL,
+  DEFAULT_NO_TTL_DISTANCE,
 } from "../common";
 import { delay, getOriginFromURL, Logger } from "./";
 import { compareArrayResultsWithIgnoredKeys, compareResultsAndFilterIgnoredKeys } from "./ObjectUtils";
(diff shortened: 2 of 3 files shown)


nicholaspai (approved):

src/utils/ProviderUtils.ts
```diff
@@ -528,6 +562,12 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
 
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "25");
 
+  const disableNoTtl = process.env["DISABLE_NO_TTL"] === "true";
+
+  const noTtlBlockDistance = process.env[`NO_TTL_BLOCK_DISTANCE_${chainId}`]
+    ? Number(process.env[`NO_TTL_BLOCK_DISTANCE_${chainId}`])
+    : DEFAULT_NO_TTL_DISTANCE[chainId];
```
pxrl, line 569: Minor style nit - IMO preferable to reduce the number of times we have to specify an env var as a string.

```suggestion
  const noTtlBlockDistance = Number(process.env[`NO_TTL_BLOCK_DISTANCE_${chainId}`] ?? DEFAULT_NO_TTL_DISTANCE[chainId]);
```

pxrl (approved):
Optimistic that this will give a decent performance boost in the bots...

src/common/Constants.ts
```diff
@@ -194,6 +194,20 @@ export const CHAIN_CACHE_FOLLOW_DISTANCE: { [chainId: number]: number } = {
...
+  1: 14400,
+  10: 86400,
+  137: 86400,
+  288: 86400,
+  324: 172800,
+  8453: 86400,
+  42161: 691200,
+  534352: 57600,
+};
+
```
james-a-morris, line 210: Do we also want to account for testnets here?

james-a-morris (approved):
One question. Otherwise LGTM

src/utils/ProviderUtils.ts
```diff
@@ -116,6 +123,8 @@ class CacheProvider extends RateLimitedProvider {
   constructor(
     providerCacheNamespace: string,
     readonly redisClient?: RedisClient,
+    // Note: if not provided, this is set to POSITIVE_INFINITY, meaning the TTL is infinite (i.e. no TTL).
+    readonly noTtlBlockDistance: number = Number.POSITIVE_INFINITY,
```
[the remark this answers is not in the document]
mrice32: Probably not. Good call!

src/utils/ProviderUtils.ts
mrice32: Actually, I think I want this to be undefined if _neither_ the env or the default is defined. So I don't think we want to wrap both in Number() -- does that make sense? I should add a comment/type annotation about this.

src/common/Constants.ts
mrice32: My thought was that for testnets we assume by default we don't want to indefinitely cache anything. If you do want that, you can provide the env. Thoughts?

src/utils/ProviderUtils.ts
mrice32: I'm going to go ahead and merge today, but I'll send you a message IRL so you get a chance to come back to this comment.

src/common/Constants.ts
james-a-morris: Fair!

src/utils/ProviderUtils.ts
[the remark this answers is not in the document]
mrice32: Fixed

PR #1253, opened 2024-03-02
fix(ProfitClient): Don't attempt to fill reverting testnet txns

pxrl:
Identified during the hackathon - a reverting deposit was incorrectly being bundled into a multicall and forcing the entire multicall to fail.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -581,7 +581,7 @@ export class ProfitClient {
     }
 
     return {
-      profitable: profitable || this.isTestnet,
+      profitable: profitable || (this.isTestnet && nativeGasCost.lt(uint256Max)),
       nativeGasCost,
       tokenGasCost,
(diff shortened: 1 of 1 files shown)


src/clients/ProfitClient.ts
```diff
@@ -581,7 +581,7 @@ export class ProfitClient {
     }
 
     return {
-      profitable: profitable || this.isTestnet,
+      profitable: profitable || (this.isTestnet && nativeGasCost.lt(uint256Max)),
```
pxrl, line 584: We ignore profitability on testnets in order to prioritise service to hackathon participants, but reverting transactions are bundled into the "unprofitable" category, so they were incorrectly being swept into the resulting multicall. The gas limit is set to `uint256Max` when a simulation reverts, so we can check for that to make sure impossible fills will be dropped.

nicholaspai (approved):

PR #1254, opened 2024-03-02
fix:(BundleDataClient): Fix caching of bundle block timestamps and validating expired older deposits

nicholaspai:
I've noticed that the executor fails to reconstruct bundle data after some time, after originally executing leaves. This led me to think that there are timing-related bugs. I isolated a disagreement about expired deposits, which led to a couple bug discoveries:

two bugs:
- bundle timestamps are not cached using block ranges as key, which leads to issues when running bundle data client.loadData on multiple block ranges, as the executor does
- `fillStatuses()` is called with incorrect `blockTag` because of syntax error

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -141,7 +141,7 @@ function updateBundleSlowFills(
 // @notice Shared client for computing data needed to construct or validate a bundle.
 export class BundleDataClient {
   private loadDataCache: DataCache = {};
-  private bundleTimestampCache: { [chainId: number]: number[] } = {};
+  private bundleTimestampCache: { [blockRangeKey: string]: { [chainId: number]: number[] } } = {};
 
   // eslint-disable-next-line no-useless-constructor
   constructor(
@@ -164,8 +164,10 @@ export class BundleDataClient {
     return _.cloneDeep(this.loadDataCache[key]);
   }
 
-  bundleTimestampsFromCache(): { [chainId: number]: number[] } {
-    return _.cloneDeep(this.bundleTimestampCache);
+  bundleTimestampsFromCache(key: string): undefined | { [chainId: number]: number[] } {
+    if (this.bundleTimestampsFromCache[key]) {
+      return _.cloneDeep(this.bundleTimestampsFromCache[key]);
+    }
   }
 
   async getPendingRefundsFromValidBundles(bundleLookback: number): Promise<CombinedRefunds[]> {
@@ -430,16 +432,15 @@ export class BundleDataClient {
     // determine whether fillDeadlines have expired.
     // @dev Going to leave this in so we can see impact on run-time in prod. This makes (allChainIds.length * 2) RPC
     // calls in parallel.
-    const _cachedBundleTimestamps = this.bundleTimestampsFromCache();
+    const _cachedBundleTimestamps = this.bundleTimestampsFromCache(key);
     let bundleBlockTimestamps: { [chainId: string]: number[] } = {};
-    _cachedBundleTimestamps;
-    if (Object.keys(_cachedBundleTimestamps).length === 0) {
+    if (_cachedBundleTimestamps) {
       bundleBlockTimestamps = await this.getBundleBlockTimestamps(
         this.chainIdListForBundleEvaluationBlockNumbers,
         blockRangesForChains,
         spokePoolClients
       );
-      this.bundleTimestampCache = bundleBlockTimestamps;
+      this.bundleTimestampCache[key] = bundleBlockTimestamps;
       if (logData) {
         this.logger.debug({
           at: "BundleDataClient#loadData",
@@ -631,7 +632,7 @@ export class BundleDataClient {
             // If deposit is in bundle and it has expired, additionally save it as an expired deposit.
             // If deposit is not in the bundle block range, then save it as an older deposit that
             // may have expired.
-            if (deposit.blockNumber <= originChainBlockRange[1] && deposit.blockNumber >= originChainBlockRange[0]) {
+            if (deposit.blockNumber >= originChainBlockRange[0]) {
               // Deposit is a V3 deposit in this origin chain's bundle block range and is not a duplicate.
               updateBundleDepositsV3(bundleDepositsV3, deposit);
               // We don't check that fillDeadline >= bundleBlockTimestamps[destinationChainId][0] because
@@ -641,7 +642,7 @@ export class BundleDataClient {
               if (deposit.fillDeadline < bundleBlockTimestamps[destinationChainId][1]) {
                 expiredBundleDepositHashes.add(relayDataHash);
               }
-            } else if (deposit.blockNumber < originChainBlockRange[0]) {
+            } else {
               olderDepositHashes.add(relayDataHash);
             }
           });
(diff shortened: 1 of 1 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -912,7 +908,7 @@ export class BundleDataClient {
         // by checkings its on-chain fill status.
         const fillStatus: BigNumber = await spokePoolClients[deposit.destinationChainId].spokePool.fillStatuses(
           relayDataHash,
-          { blockTag: destinationChainId[1] }
+          { blockTag: destinationBlockRange[1] }
```
nicholaspai, line 911: bug fix

src/clients/BundleDataClient.ts
```diff
@@ -631,7 +632,7 @@ export class BundleDataClient {
             // If deposit is in bundle and it has expired, additionally save it as an expired deposit.
             // If deposit is not in the bundle block range, then save it as an older deposit that
             // may have expired.
-            if (deposit.blockNumber <= originChainBlockRange[1] && deposit.blockNumber >= originChainBlockRange[0]) {
```
nicholaspai, line 634: opportunistic refactor

src/clients/BundleDataClient.ts
```diff
@@ -141,7 +141,7 @@ function updateBundleSlowFills(
 // @notice Shared client for computing data needed to construct or validate a bundle.
 export class BundleDataClient {
   private loadDataCache: DataCache = {};
-  private bundleTimestampCache: { [chainId: number]: number[] } = {};
+  private bundleTimestampCache: { [blockRangeKey: string]: { [chainId: number]: number[] } } = {};
```
james-a-morris, line 144: QQ: what is `blockRangeKey` meant to be in this case? Is it `CHAIN:START:END`?

src/clients/BundleDataClient.ts
```diff
@@ -164,8 +164,10 @@ export class BundleDataClient {
     return _.cloneDeep(this.loadDataCache[key]);
   }
 
-  bundleTimestampsFromCache(): { [chainId: number]: number[] } {
-    return _.cloneDeep(this.bundleTimestampCache);
+  bundleTimestampsFromCache(key: string): undefined | { [chainId: number]: number[] } {
```
james-a-morris, line 167: Thoughts on adding an explicit `return undefined`? I don't think it adds much value code-wise but it helps understand that the cache could miss

src/clients/BundleDataClient.ts
```diff
@@ -430,16 +432,15 @@ export class BundleDataClient {
...
-    _cachedBundleTimestamps;
-    if (Object.keys(_cachedBundleTimestamps).length === 0) {
+    if (!_cachedBundleTimestamps) {
       bundleBlockTimestamps = await this.getBundleBlockTimestamps(
         this.chainIdListForBundleEvaluationBlockNumbers,
         blockRangesForChains,
         spokePoolClients
       );
-      this.bundleTimestampCache = bundleBlockTimestamps;
+      this.bundleTimestampCache[key] = bundleBlockTimestamps;
```
james-a-morris, line 443: We have a sanitized getter - should we also include a sanitized setter for this map?

src/clients/BundleDataClient.ts
```diff
@@ -631,7 +632,7 @@ export class BundleDataClient {
             // If deposit is in bundle and it has expired, additionally save it as an expired deposit.
             // If deposit is not in the bundle block range, then save it as an older deposit that
             // may have expired.
-            if (deposit.blockNumber <= originChainBlockRange[1] && deposit.blockNumber >= originChainBlockRange[0]) {
+            if (deposit.blockNumber >= originChainBlockRange[0]) {
```
james-a-morris, line 635: We may want to leave a comment that we don't mind deposits that happen in the future w.r.t. the origin chain block range

src/clients/BundleDataClient.ts
```diff
@@ -912,7 +908,7 @@ export class BundleDataClient {
         // by checkings its on-chain fill status.
         const fillStatus: BigNumber = await spokePoolClients[deposit.destinationChainId].spokePool.fillStatuses(
           relayDataHash,
-          { blockTag: destinationChainId[1] }
+          { blockTag: Math.min(destinationBlockRange[1], spokePoolClients[destinationChainId].latestBlockSearched) }
```
james-a-morris, line 911: Good call - will we ever be in a situation where we are in-between a block range?

james-a-morris (commented):
Left a few comments

src/clients/BundleDataClient.ts
nicholaspai: It's the same key as in `DataCache` so I'll use the Record syntax here to align it more

src/clients/BundleDataClient.ts
nicholaspai: that's actually not what's going on here. I removed this because the LHS was already applied in the `filter` on line 619.

src/clients/BundleDataClient.ts
nicholaspai: We never really shoul dbe using this `Math.min` except for test cases. It's very ugly, i'll add a comment

src/clients/BundleDataClient.ts
```diff
@@ -912,7 +913,12 @@ export class BundleDataClient {
         // by checkings its on-chain fill status.
         const fillStatus: BigNumber = await spokePoolClients[deposit.destinationChainId].spokePool.fillStatuses(
```
pxrl, line 914: `sdkUtils.relayFillStatus()` will handle the bounds checking and return a `FillStatus` enum - might be nice to use that (will probably also collapse the code to a single line).

```suggestion
        const fillStatus = await relayFillStatus(
```

src/clients/BundleDataClient.ts
pxrl: The commit message indicates this was a bug, but it doesn't really seem like a direct problem - was the issue that it did something unexpected in test and suppressed another issue?

pxrl (approved):

src/clients/BundleDataClient.ts
nicholaspai: it broke tests yeah because we use weird block ranges in tests where the end block >>> latest block searched

src/clients/BundleDataClient.ts
```diff
@@ -910,29 +911,33 @@ export class BundleDataClient {
...
-          relayDataHash,
-          { blockTag: destinationChainId[1] }
+        const fillStatus = await utils.relayFillStatus(
+          spokePoolClients[destinationChainId].spokePool,
+          deposit,
+          // We can assume that in production
+          // the block ranges passed into this function would never contain blocks where the spoke pool client
+          // hasn't queried. This is because this function will usually be called
+          // in production with block ranges that were validated by
+          // DataworkerUtils.blockRangesAreInvalidForSpokeClients
```
mrice32, line 921: I'm a little confused here. Why do we need to assume this?

mrice32 (commented):
Just one question

src/clients/BundleDataClient.ts
nicholaspai: I dont think we should be proposing any bundles with an end block that we haven't seen yet. This would very likely lead to a dispute. Therefore the proposer refuses to do that when bundles are in the future. I believe the disputer will also dispute any bundle with an end block in the future.

pxrl (approved):
The main fix seems to be the bundle timestamps change; I don't see any scenarios where we'd be overshooting the fillStatus block in production so that seems mostly a workaround for some test difficulties. OK from me on that basis.

PR #1257, opened 2024-03-05
improve: Update eth-optimism sdk to 3.2.x

nicholaspai:
Required to support Fault proofs on OP mainnet. Fault proofs going live on testnet this month so this PR is required to make finalizer work on testnets


james-a-morris (approved):
LGTM assuming all integration tests pass.

All unit tests pass so I'm confident we won't see a regression here

mrice32 (approved):
LGTM!

pxrl (approved):

PR #1258, opened 2024-03-05
refactor: Simplify FillUtils getUnfilledDeposits()

pxrl:
This is an incremental change to drop some unneeded members from the return type, and to return an Object of destination chain => unfilled deposits array. This permits the caller to keep the unfilled deposits segregated, which fits future use cases pretty well.

This is take #2 at this change, since the original commit incorrectly handled the deposit filtering based on the block lookback by applying the wrong "low block" for each chain Id.

### DIFF
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -1,6 +1,6 @@
 import assert from "assert";
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
-import { HubPoolClient } from "../clients";
+import { HubPoolClient, SpokePoolClient } from "../clients";
 import {
   Fill,
   FillsToRefund,
@@ -259,8 +259,6 @@ export function getFillsInRange(
 export type RelayerUnfilledDeposit = {
   deposit: V3DepositWithBlock;
   version: number;
-  unfilledAmount: BigNumber;
-  fillCount: number;
   invalidFills: Fill[];
 };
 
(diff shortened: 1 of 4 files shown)


src/utils/FillUtils.ts
```diff
@@ -273,49 +271,54 @@ export async function getUnfilledDeposits(
...
   const chainIds = Object.values(spokePoolClients).map(({ chainId }) => chainId);
-  let earliestBlockNumbers = Object.values(spokePoolClients).map(({ deploymentBlock }) => deploymentBlock);
+  const originFromBlocks: { [chainId: number]: number } = Object.fromEntries(
+    Object.values(spokePoolClients).map(({ chainId, deploymentBlock }) => [chainId, deploymentBlock])
+  );
 
   if (isDefined(depositLookBack)) {
     const blockFinder = undefined;
     const redis = await getRedisCache();
-    earliestBlockNumbers = await Promise.all(
```
pxrl, line 284: tldr on the previous bug was that the wrong `earliestBlockNumber` was used when applying filtering on the deposits found in each origin SpokePoolClient.

The old code correctly iterated over the SpokePoolClient instances and used the correct index when resolving the lowest block number to use for filtering. The update incorrectly used the destination chain's index when iterating over each origin SpokePoolClient, so the block limit made no sense in that context.

I've updated it now to store the low block in a Record instead. This removes any assumptions about array orderings and size, which anyway do not hold true when using `Array.filter().map()`.

src/utils/FillUtils.ts
```diff
@@ -273,49 +271,54 @@ export async function getUnfilledDeposits(
...
-        const hints = { lowBlock: spokePoolClient.deploymentBlock };
-        return getBlockForTimestamp(spokePoolClient.chainId, timestamp, blockFinder, redis, hints);
-      })
-    );
+    await sdkUtils.forEachAsync(Object.values(spokePoolClients), async (spokePoolClient: SpokePoolClient) => {
+      const { chainId: originChainId, deploymentBlock } = spokePoolClient;
+      const timestamp = spokePoolClient.getCurrentTime() - depositLookBack;
+      const hints = { lowBlock: deploymentBlock };
+      const startBlock = await getBlockForTimestamp(originChainId, timestamp, blockFinder, redis, hints);
+      originFromBlocks[originChainId] = Math.max(deploymentBlock, startBlock);
```
nicholaspai, line 289: Why max?

pxrl: This prevents undershoot of the SpokePool deployment. That in itself isn't harmful, but there's an assert later on when this is used to ensure that the low black falls within the bounds of `[deploymentBlock, latestBlockSearched)`. This helps to prevent the previous issue of the deposit filtering being nonsensical for the given origin chain.

assert

The reason for asserting is that this code isn't super well covered in test because it has a hard coupling to redis and an RPC provider instance identifiable via the `originChainId`. In test this gets mostly fudged because there's actually only one provider for all of the SpokePoolClient instances, and we spoof the chain Id in the mocked SpokePool deployments in HRE. So in practice we actually pass `depositLookBack = undefined` in these tests and the timestamp-to-block resolution is skipped entirely, which is why the first version of this PR was able to cause problems in production.

I've been thinking about how we can ensure better test coverage of this, but I haven't landed on a solution yet.

nicholaspai (approved):

src/utils/FillUtils.ts
```diff
@@ -273,49 +271,54 @@ export async function getUnfilledDeposits(
...
-      unfilledDeposits.push(
-        ...unfilledDepositsForDestinationChain.filter((deposit) => deposit.unfilledAmount.gt(bnZero))
-      );
-    }
+        // Find all unfilled deposits for the current loops originChain -> destinationChain. Note that this also
+        // validates that the deposit is filled "correctly" for the given deposit information. This includes validation
+        // of the all deposit -> relay props, the realizedLpFeePct and the origin->destination token mapping.
+        return originClient
+          .getDepositsForDestinationChain(destinationChainId)
+          .filter((deposit) => deposit.blockNumber >= earliestBlockNumber)
```
mrice32, line 313: This wasn't added or removed, so just asking out of curiosity: why do we explicitly filter by the lookback instead of using the lookback to determine the SpokePoolClient's block range and then just taking all deposits the SpokePool gives us?

mrice32 (approved):
LGTM!

src/utils/FillUtils.ts
pxrl: I think it's probably legacy behaviour to support polling mode, but I could be wrong.

mrice32: Gotcha, fine to keep to not change behavior. Was just curious.

PR #1259, opened 2024-03-05
refactor(ProfitClient): Remove v2 support & update tests

pxrl:
This change removes all v2 support from the ProfitClient.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -62,7 +62,7 @@ export type FillProfit = {
   nativeGasCost: BigNumber; // Cost of completing the fill in the units of gas.
   tokenGasCost: BigNumber; // Cost of completing the fill in the relevant gas token.
   gasPadding: BigNumber; // Positive padding applied to nativeGasCost and tokenGasCost before profitability.
-  gasMultiplier: BigNumber; // Gas multiplier applied to fill cost estimates before profitability.
+  gasMultiplier: BigNumber; // Multiplier applied to token-only fill cost estimates before profitability.
   gasTokenPriceUsd: BigNumber; // Price paid per unit of gas the gas token in USD.
   gasCostUsd: BigNumber; // Estimated cost of completing the fill in USD.
   netRelayerFeePct: BigNumber; // Relayer fee after gas costs as a portion of relayerCapitalUsd.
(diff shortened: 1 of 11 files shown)


test/ProfitClient.ConsiderProfitability.ts
pxrl, line 1: nb. the changes here are pretty significant. I'd like to split this PR into 2, where the new tests are introduced alongside the existing ones without actually changing the ProfitClient - this would give best confidence that there are no regressions here.

I will revisit this tomorrow to see if they can be split apart.

src/clients/ProfitClient.ts
```diff
@@ -437,7 +357,7 @@ export class ProfitClient {
 
     // Determine profitability.
     const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
-    const netRelayerFeePct = inputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(inputAmountUsd) : bnZero;
+    const netRelayerFeePct = inputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(outputAmountUsd) : bnZero;
```
nicholaspai, line 360: Should the LHS be the same as the RHS? So either both are input amount or both are output amount?

pxrl: Yeah, good catch. The effect should be the same since this check is intended to catch when the _token price_ is 0 (and it will be the same price for input and output tokens in current setup), but for consistency it should be `outputToken`.

test/ProfitClient.ConsiderProfitability.ts
pxrl: Have split them apart; the test updates are made in this PR: https://github.com/across-protocol/relayer-v2/pull/1262.

src/clients/ProfitClient.ts
pxrl: I'll mark this comment as resolved because the thread continues over here: https://github.com/across-protocol/relayer-v2/pull/1262#discussion_r1513712909.

nicholaspai (approved):

mrice32 (approved):
LGTM! Easy review because the functions and types were so well organized between v2 and v3. +1

PR #1260, opened 2024-03-05
feat(linea): add default constants

james-a-morris:
Add constant defaults so that Linea can be references by the bots.


src/common/Constants.ts
```diff
@@ -54,10 +55,12 @@ export const DEFAULT_MIN_DEPOSIT_CONFIRMATIONS = {
   324: 0,
   8453: 60,
   42161: 0,
+  59144: 64,
```
nicholaspai, line 58: How did you get this?

nicholaspai (approved):

src/common/Constants.ts
james-a-morris: This was an optimistic value due to Linea being a new chain. I'm not sure how often this chain ReOrgs so I was being cautious.

Should we change this value?

mrice32: @james-a-morris before setting this and other finality-related configuration parameters, we should dig into their docs or ask their team about the likelihood of reorgs and how deep they can be. It's pretty important that we get this right.

mrice32 (commented):
One question on finality

src/common/Constants.ts
james-a-morris: @mrice32 I've updated the parameters. I spoke with Linea and they said that Linea has a soft-finality of 1 block. I've added some padding because we're on a new chain. But, these should reflect the accompanying comments for each constant.

src/common/Constants.ts
```diff
@@ -19,6 +19,7 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
   324: 4 * 24 * 60 * 60,
   8453: 172800, // Same as Optimism.
   42161: 1382400,
+  59144: 86400, // 1 block every 4 seconds - 4 days
```
nicholaspai, line 22: I actually don't think this DATAWORKER_FAST_LOOKBACK variable is used anymore in the code?

james-a-morris: You're right - we use `DATAWORKER_FAST_LOOKBACK_COUNT` instead of this per-chain constant. https://github.com/across-protocol/relayer-v2/pull/1260/commits/894a178504cf8022846a2348abd8766ab1f2c113

src/common/Constants.ts
```diff
@@ -184,11 +195,13 @@ export const CHAIN_CACHE_FOLLOW_DISTANCE: { [chainId: number]: number } = {
   324: 512,
   8453: 120,
   42161: 32,
+  59144: 8, // Linea has a soft-finality of 1 block. This value is padded - but at 4s/block the padding is light (32 seconds)
```
pxrl, line 198: This seems pretty low (nb. the Arbitrum one does too!!!). It'll effectively determine how quickly something ends up in the cache for a given provider. I'd probably argue for being a bit more conservative and setting it to ~5 minutes (75 blocks).

src/dataworker/Dataworker.ts
```diff
@@ -2104,6 +2128,40 @@ export class Dataworker {
...
+    const l2MessageABI = CONTRACT_ADDRESSES[client.chainId]?.l2MessageService?.abi;
+    const l2MessageAddress = CONTRACT_ADDRESSES[client.chainId]?.l2MessageService?.address;
+    if (!isDefined(l2MessageABI) || !isDefined(l2MessageAddress)) {
+      this.logger.error({
+        at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
+        message: "L2MessageService contract ABI or address is not defined for Linea chain!",
+        chainId: client.chainId,
+      });
+      throw new Error("L2MessageService contract ABI or address is not defined for Linea chain!");
+    }
```
pxrl, line 2158: Since these two errors are probably never expected to occur in practice, do you think it would be OK with assert()s? It'd cut this function down by >> 50%.

pxrl (approved):
OK from me, but I do think we should extend the follow distance so that we only cache after a minimum of 5 minutes.

src/dataworker/Dataworker.ts
james-a-morris: Agreed: https://github.com/across-protocol/relayer-v2/pull/1260/commits/57443e76dc5152df6a923a42a6b7b11630b45d20

src/common/Constants.ts
```diff
@@ -141,10 +150,12 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   324: 120, // ~1s/block. ZkSync is a centralized sequencer but is relatively unstable so this is kept higher than 0
   8453: 60, // 2s/block. Same finality profile as Optimism
   42161: 240, // ~0.25s/block. Arbitrum is a centralized sequencer
+  59144: 375, // At 4s/block, 25 mins = 375 blocks.
```
nicholaspai, line 153: The other chains are roughly at 2 mins not 25 mins anymore

src/dataworker/Dataworker.ts
```diff
@@ -1906,6 +1906,14 @@ export class Dataworker {
       return;
     }
     const chainId = client.chainId;
+
+    // If the chain is Linea, then we need to allocate ETH in the call to executeRelayerRefundLeaf. This is currently
+    // unique to the L2 -> L1 relay direction for Linea. We will make this variable generic defaulting to undefined
+    // for other chains.
+    const valueToPassViaPayable = sdkUtils.chainIsLinea(chainId)
```
nicholaspai, line 1913: Is this leftover from another branch?

nicholaspai (approved):
Lgtm but I feel like a lot of the constants settings are arbitrary and have no good justification besides that they are safe. My instinct is they are excessively safe. I'd prefer to have more probabilistic reasoning but this is fine to start with

src/dataworker/Dataworker.ts
james-a-morris: Yes and no - this code change is from https://github.com/across-protocol/relayer-v2/pull/1266 and it was merged against this branch

src/common/Constants.ts
james-a-morris: This was added to align with the pre-rebased changes. I've now updated https://github.com/across-protocol/relayer-v2/pull/1260/commits/0c4d4e85704e800e1e9132d3ef5e5749b2e0a559

src/common/Constants.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1260/commits/08a4e110f3198187b44a5120e05dd3d58f93d2e3

src/common/Constants.ts
mrice32: Okay, so I think we agree on 30 for the default (highest) confirmation blocks here.

mrice32 (approved):

src/common/Constants.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/1260/commits/1a9cd54fa3a60da6b0ead2bbc9e5577fa40b5c85

PR #1261, opened 2024-03-05
feat: enable no TTL caching even when regular provider caching is dis…

mrice32:
This PR makes some updates to the caching system.

The behavior changes are:
1. No TTL caching is enabled _by default even when normal provider caching is disabled_. Note: this is an opinionated change, but I think it's reasonable because old block data (default distance of 2 days) rarely changes, and sophisticated users who are concerned with this tail risk would likely be using advanced caching configuration anyway.
2. Updates the env variable to disable no TTL caching so its name is closer to the other disable env variable. The new env is: `NODE_DISABLE_INFINITE_TTL_PROVIDER_CACHING`.
3. If CHAIN_CACHE_FOLLOW_DISTANCE is not defined, we just default to not caching rather than erroring. Unclear if this is the right call. On one hand, this means it's easier to run this on testnet or other arbitrary chains without hardcoding block distances. On the other hand, it means that if we were to forget to add a new chain to this list, caching would silently be disabled on that chain. Thoughts?

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -14,6 +14,7 @@ import {
 } from "../common";
 import { delay, getOriginFromURL, Logger } from "./";
 import { compareArrayResultsWithIgnoredKeys, compareResultsAndFilterIgnoredKeys } from "./ObjectUtils";
+import { MAINNET_CHAIN_IDs } from "@across-protocol/constants-v2";
 
 const logger = Logger;
 
@@ -552,7 +553,7 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
     NODE_DISABLE_PROVIDER_CACHING,
     NODE_PROVIDER_CACHE_NAMESPACE,
     NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS,
-    NODE_DISABLE_NO_TTL_PROVIDER_CACHING,
+    NODE_DISABLE_INFINITE_TTL_PROVIDER_CACHING,
   } = process.env;
 
   const timeout = Number(process.env[`NODE_TIMEOUT_${chainId}`] || NODE_TIMEOUT || defaultTimeout);
@@ -568,7 +569,7 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
 
   const nodeMaxConcurrency = Number(process.env[`NODE_MAX_CONCURRENCY_${chainId}`] || NODE_MAX_CONCURRENCY || "25");
 
-  const disableNoTtlCaching = NODE_DISABLE_NO_TTL_PROVIDER_CACHING === "true";
+  const disableNoTtlCaching = NODE_DISABLE_INFINITE_TTL_PROVIDER_CACHING === "true";
 
   // Note: if there is no env var override _and_ no default, this will remain undefined and
   // effectively disable indefinite caching of old blocks/keys.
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
> If CHAIN_CACHE_FOLLOW_DISTANCE is not defined, we just default to not caching rather than erroring.

This would be a breaking change if merged without a corresponding bot-config, right? Just FYI

I can see some some merit to making sure we have a standard TTL defined for each chain in the chainIDList. Seems like it'd be pretty hard to debug performance issues when adding a new chain and that chain seems to be slower than the rest. I can imagine us chasing after a "bad RPC" when in fact we're not caching. So I'd mildly prefer to fail earlier

pxrl:
> No TTL caching is enabled by default even when normal provider caching is disabled. Note: this is an opinionated change, but I think it's reasonable because we

Sentence looks incomplete - would be nice to complete it for reference purposes.


> 3\. If CHAIN_CACHE_FOLLOW_DISTANCE is not defined, we just default to not caching rather than erroring. Unclear if this is the right call. On one hand, this means it's easier to run this on testnet or other arbitrary chains without hardcoding block distances. On the other hand, it means that if we were to forget to add a new chain to this list, caching would silently be disabled on that chain. Thoughts?

We can distinguish between `MAINNET_CHAIN_IDs` and `TESTNET_CHAIN_IDs`, so we could implement different behaviour on each: https://github.com/across-protocol/constants-v2/blob/cdd23f62b51dca3d53330687f6e3350c30383cc9/src/constants.ts#L26

It would be really nice if we defaulted to 0 for testnet because it's always a pain to have to manually add them when supporting a new chain. This was most recently the case with the Espresso demo.

src/utils/ProviderUtils.ts
```diff
@@ -547,6 +552,7 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
     NODE_DISABLE_PROVIDER_CACHING,
     NODE_PROVIDER_CACHE_NAMESPACE,
     NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS,
+    NODE_DISABLE_NO_TTL_PROVIDER_CACHING,
```
pxrl, line 555: This variable name is a bit counter-intuitive for me - probably because of the double negative `DISABLE_NO`. But I see why it's nice to have this as opt-out rather than opt-in. Would something like `NODE_DISABLE_INFINITE_TTL_PROVIDER_CACHING` be an OK description?

pxrl (approved):

src/utils/ProviderUtils.ts
mrice32: Yep, can change to that!

mrice32:
> > No TTL caching is enabled by default even when normal provider caching is disabled. Note: this is an opinionated change, but I think it's reasonable because we
> 
> Sentence looks incomplete - would be nice to complete it for reference purposes.

Updated, good catch!

> 
> > 3. If CHAIN_CACHE_FOLLOW_DISTANCE is not defined, we just default to not caching rather than erroring. Unclear if this is the right call. On one hand, this means it's easier to run this on testnet or other arbitrary chains without hardcoding block distances. On the other hand, it means that if we were to forget to add a new chain to this list, caching would silently be disabled on that chain. Thoughts?
> 
> We can distinguish between `MAINNET_CHAIN_IDs` and `TESTNET_CHAIN_IDs`, so we could implement different behaviour on each: https://github.com/across-protocol/constants-v2/blob/cdd23f62b51dca3d53330687f6e3350c30383cc9/src/constants.ts#L26
> 
> It would be really nice if we defaulted to 0 for testnet because it's always a pain to have to manually add them when supporting a new chain. This was most recently the case with the Espresso demo. I think it's good to error hard if it's not defined on a production chain.

Can make this change!

mrice32:
> > If CHAIN_CACHE_FOLLOW_DISTANCE is not defined, we just default to not caching rather than erroring.
> 
> This would be a breaking change if merged without a corresponding bot-config, right? Just FYI
> 
> I can see some some merit to making sure we have a standard TTL defined for each chain in the chainIDList. Seems like it'd be pretty hard to debug performance issues when adding a new chain and that chain seems to be slower than the rest. I can imagine us chasing after a "bad RPC" when in fact we're not caching. So I'd mildly prefer to fail earlier

You and @pxrl both pointed out this preference, so this is probably the right way to go. Updated.

mrice32:
> This would be a breaking change if merged without a corresponding bot-config, right? Just FYI

Agreed!

PR #1262, opened 2024-03-06
refactor(ProfitClient): Update tests for v3

pxrl:
This change upgrades all ProfitClient tests to v3. It's submitted in advance of a subsequent PR to remove all v2 support from the ProfitClient.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -437,7 +437,7 @@ export class ProfitClient {
 
     // Determine profitability.
     const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
-    const netRelayerFeePct = inputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(inputAmountUsd) : bnZero;
+    const netRelayerFeePct = outputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(outputAmountUsd) : bnZero;
 
     // If either token prices are unknown, assume the relay is unprofitable. Force non-equivalent tokens
     // to be unprofitable for now. The relayer may be updated in future to support in-protocol swaps.
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -437,7 +437,9 @@ export class ProfitClient {
 
     // Determine profitability.
     const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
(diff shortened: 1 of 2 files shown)


src/clients/ProfitClient.ts
```diff
@@ -437,7 +437,7 @@ export class ProfitClient {
 
     // Determine profitability.
     const netRelayerFeeUsd = grossRelayerFeeUsd.sub(gasCostUsd);
-    const netRelayerFeePct = inputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(inputAmountUsd) : bnZero;
+    const netRelayerFeePct = outputAmountUsd.gt(bnZero) ? netRelayerFeeUsd.mul(fixedPoint).div(outputAmountUsd) : bnZero;
```
nicholaspai, line 440: Why should we not use inputAmount? The input amount is the amount refunded to the relayer so it seems like the logical denominator no?

pxrl: `inputAmount` includes the LP fee, so it pulls netRelayerFeePct lower than it actually is. We could take `inputAmountUsd - lpFeeUsd - gasCostUsd`

pxrl: I don't think so, but there is some nuance here so I'll add that for context.

`netRelayerFeePct` is effectively trying to isolate the relayer capital fee for the fill, so the profitability logic progressively excludes all other known costs associated with the fill (i.e. LP fee + anticipated gas cost). The raw input amount (`inputAmountUsd`) isn't a good fit in this context because it's not entirely repaid to the relayer and it also includes the destination gas provision, so we'd be comparing the relayer revenue _after costs_ (`netRelayerFeeUsd`) with the raw input amount that includes provision for those costs (`inputAmountUsd`). This would it make `netRelayerFeePct` seem lower than it really is.

The computation of `netRelayerFeeUsd / outputAmountUsd` seems simple to me because it reduces the equation to a question of the direct capital outlay (`outputAmountUsd`) against the net fee to be received for supplying it (after gas cost and LP fee).

Does that make sense? I might have a dumb assumption in there, or maybe there's a simpler way to handle it.

nicholaspai: ok I buy that argument because the relayer is only ever transferring away the `outputAmount` so the capital fee should be based on this

nicholaspai (approved):
LGTM but I prefer adding a quick comment about the choice of `outputAmountUsd`

pxrl:
> LGTM but I prefer adding a quick comment about the choice of `outputAmountUsd`

Fair call - have added a comment here: https://github.com/across-protocol/relayer-v2/pull/1262/commits/befcf39e19f9b42ee182c159784b421e028a0e92.

PR #1264, opened 2024-03-07
feat: Support submitting private transactions

pxrl:
This change introduces the ability to define custom RPC providers to be used when submitting transactions. Custom providers are slotted under the PRIVACY namespace, since that's the anticipated use case for them.

This is held fairly generic within ProviderUtils by scoping the change as a generic namespace, rather than introducing any concept of privacy.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -511,10 +511,10 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
 }
 
 // Global provider cache to avoid creating multiple providers for the same chain.
-const providerCache: { [chainId: number]: RetryProvider } = {};
+const providerCache: Record<string, RetryProvider> = {};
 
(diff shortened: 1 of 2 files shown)


pxrl:
Note: This is the most minimally-invasive way I could figure to introduce this feature. It's a little bit on the side of a kludge, but I expect it should be reliable enough and doesn't pollute the config syntax too badly, so hopefully it's something we can live with.

nicholaspai (commented):
One note is that submitting to private RPCs might change how the txn client or multicaller client reports txn receipts. Can we test this out locally?

pxrl:
> One note is that submitting to private RPCs might change how the txn client or multicaller client reports txn receipts. Can we test this out locally?

I've tested locally on mainnet and Sepolia and have added some additional logging to verify that the new provider is instantiated correctly. So far I haven't seen any difference in the transmission flow - it seems to work exactly as expected.

pxrl:
Closing this for now; I saw too many transactions get blackholed during test.

PR #1265, opened 2024-03-07
improve(ProfitClient): Permit applying gas multipler on message fills

pxrl:
This change permits the relayer to have a separate gas multiplier for fills. This is desirable since fills with messages will typically have much higher gas consumption, so heavily discounting them is quite risky. However, ethers seems to over-estimate quite a bit, so it can be helpful to wind some of that back in some cases.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -278,7 +278,7 @@ export class ProfitClient {
     // @todo Consider phasing this out and relying solely on the minimum profitability config.
     if (isMessageEmpty(resolveDepositMessage(deposit))) {
       tokenGasCost = tokenGasCost.mul(this.gasMultiplier).div(fixedPoint);
-    } else if (isDefined(process.env.RELAYER_GAS_MULTIPLIER_MESSAGES)) {
+    } else if (isDefined(process.env.RELAYER_MESSAGE_GAS_MULTIPLIER)) {
       // @todo: Parse this in RelayerConfig and pass it in via the constructor.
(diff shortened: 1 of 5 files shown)


src/clients/ProfitClient.ts
```diff
@@ -279,6 +279,9 @@ export class ProfitClient {
     // @todo Consider phasing this out and relying solely on the minimum profitability config.
     if (isMessageEmpty(resolveDepositMessage(deposit))) {
       tokenGasCost = tokenGasCost.mul(this.gasMultiplier).div(fixedPoint);
+    } else if (isDefined(process.env.RELAYER_GAS_MULTIPLIER_MESSAGES)) {
```
pxrl, line 282:
```suggestion
    } else if (isDefined(process.env.RELAYER_MESSAGE_GAS_MULTIPLIER)) {
```

src/clients/ProfitClient.ts
```diff
@@ -279,6 +279,9 @@ export class ProfitClient {
     // @todo Consider phasing this out and relying solely on the minimum profitability config.
     if (isMessageEmpty(resolveDepositMessage(deposit))) {
       tokenGasCost = tokenGasCost.mul(this.gasMultiplier).div(fixedPoint);
+    } else if (isDefined(process.env.RELAYER_MESSAGE_GAS_MULTIPLIER)) {
+      // @todo: Parse this in RelayerConfig and pass it in via the constructor.
```
pxrl, line 283: nb. This is just a PoC to demonstrate how message-based gas over-estimation can be compensated for via configuration. This PR will be updated to pass this config in via the ProfitClient constructor before merge, assuming there are no other blockers or alternative proposals.

nicholaspai (approved):

PR #1266, opened 2024-03-07
improve(dataworker): allow linea to execute relay leaves

james-a-morris:
Linea adds a constraint in the L2->L1 direction where we need to pass a payable `minimumWeiFee` to the `executeRelayerRefundLeaf`. 

We need to query this fee & pass it accordingly to the multicaller client.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -2101,16 +2101,17 @@ export class Dataworker {
    * relay fee from the L2Linea Messenger contract.
    * @param leaf The relay leaf to execute. Used in this function to prevent non-Linea chains from calling this method.
    * @returns The amount of ETH required to execute the relay leaf.
+   * @throws If the method is called using a non-linea spoke pool client.
    */
-  _getRequiredEthForLineaRelayLeafExecution(client: SpokePoolClient): Promise<BigNumber | undefined> {
+  _getRequiredEthForLineaRelayLeafExecution(client: SpokePoolClient): Promise<BigNumber> {
+    // You should *only* call this method on Linea chains.
     if (![CHAIN_IDs.LINEA, CHAIN_IDs.LINEA_GOERLI].includes(client.chainId)) {
-      this.logger.warn({
+      this.logger.error({
         at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
         message: "Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!",
         chainId: client.chainId,
       });
-      // There is no fee for non-Linea chains w/r/t Linea so we can return undefined.
-      return undefined;
+      throw new Error("Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!");
     }
     // Resolve the contract entry for the L2Linea Messenger contract.
     const { abi: l2MessageABI, address: l2MessageAddress } = CONTRACT_ADDRESSES[client.chainId].l2MessageService;
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -2113,8 +2113,17 @@ export class Dataworker {
       });
       throw new Error("Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!");
     }
-    // Resolve the contract entry for the L2Linea Messenger contract.
-    const { abi: l2MessageABI, address: l2MessageAddress } = CONTRACT_ADDRESSES[client.chainId].l2MessageService;
+    // Resolve and sanitize the L2MessageService contract ABI and address.
+    const l2MessageABI = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.abi;
+    const l2MessageAddress = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.address;
+    if (!isDefined(l2MessageABI) || !isDefined(l2MessageAddress)) {
+      this.logger.error({
+        at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
+        message: "L2MessageService contract ABI or address is not defined for Linea chain!",
+        chainId: client.chainId,
+      });
+      throw new Error("L2MessageService contract ABI or address is not defined for Linea chain!");
+    }
     // For Linea, the bot needs enough ETH to pay for each L2 -> L1 message.
     const l2MessagerContract = new Contract(l2MessageAddress, l2MessageABI, client.spokePool.provider);
     // Get the latest relay fee from the L2Linea Messenger contract.
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -2112,8 +2112,8 @@ export class Dataworker {
       throw new Error("Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!");
     }
     // Resolve and sanitize the L2MessageService contract ABI and address.
-    const l2MessageABI = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.abi;
-    const l2MessageAddress = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.address;
+    const l2MessageABI = CONTRACT_ADDRESSES[client.chainId]?.l2MessageService?.abi;
+    const l2MessageAddress = CONTRACT_ADDRESSES[client.chainId]?.l2MessageService?.address;
     if (!isDefined(l2MessageABI) || !isDefined(l2MessageAddress)) {
       this.logger.error({
         at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1,5 +1,5 @@
 import assert from "assert";
-import { utils as ethersUtils } from "ethers";
+import { Contract, utils as ethersUtils } from "ethers";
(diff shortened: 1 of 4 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1904,7 +1929,7 @@ export class Dataworker {
           } else {
             // If mainnet leaf, then allocate balance to the HubPool since it will be atomically transferred.
             if (leaf.chainId === this.clients.hubPoolClient.chainId && leaf.amountToReturn.gt(0)) {
-              await balanceAllocator.addUsed(
+              balanceAllocator.addUsed(
```
james-a-morris, line 1932: Nit: we don't need an `await` here since the fn doesn't return a promise

src/dataworker/Dataworker.ts
```diff
@@ -2070,6 +2096,30 @@ export class Dataworker {
...
+      this.logger.warn({
+        at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
+        message: "Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!",
+        chainId: client.chainId,
+      });
+      // There is no fee for non-Linea chains w/r/t Linea so we can return undefined.
+      return undefined;
+    }
+    // Resolve the contract entry for the L2Linea Messenger contract.
+    const { abi: l2MessageABI, address: l2MessageAddress } = CONTRACT_ADDRESSES[client.chainId].l2MessageService;
```
james-a-morris, line 2116: Wasn't sure if we wanted to sanitize the case where l2MessageABI or l2MessageAddress is null. I opted to not check/throw an error since

A: this PR defines the ABI/address for both Linea mainnet/testnet
B: we'd get an error when we call `new Contract( )`
C: it felt overcautious w.r.t. how we reference this struct in other locations of the codebase

src/dataworker/Dataworker.ts
```diff
@@ -1872,6 +1873,15 @@ export class Dataworker {
       return;
     }
     const chainId = client.chainId;
+
+    // If the chain is Linea, then we need to allocate ETH in the call to executeRelayerRefundLeaf. This is currently
+    // unique to the L2 -> L1 relay direction for Linea. We will make this variable generic defaulting to undefined
+    // for other chains.
+    const valueToPassViaPayable =
+      chainId === CHAIN_IDs.LINEA || CHAIN_IDs.LINEA_GOERLI
```
nicholaspai, line 1881: Should we add an SDK constants similar to [isArbitrum[(https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/Dataworker.ts#L1555C24-L1555C39) ?

src/dataworker/Dataworker.ts
```diff
@@ -2070,6 +2096,30 @@ export class Dataworker {
...
 
+  /**
+   * Retrieves the amount of ETH required to execute a Linea relay leaf by querying the latest
+   * relay fee from the L2Linea Messenger contract.
+   * @param leaf The relay leaf to execute. Used in this function to prevent non-Linea chains from calling this method.
+   * @returns The amount of ETH required to execute the relay leaf.
+   */
+  _getRequiredEthForLineaRelayLeafExecution(client: SpokePoolClient): Promise<BigNumber | undefined> {
+    if (![CHAIN_IDs.LINEA, CHAIN_IDs.LINEA_GOERLI].includes(client.chainId)) {
+      this.logger.warn({
```
nicholaspai, line 2107: Shouldnt' this be an error not a warning? This seems like a programmer error if we ever enter here

src/dataworker/Dataworker.ts
nicholaspai: I think at a minimum we need a `?.` syntax. I also prefer throwing an error here

src/dataworker/Dataworker.ts
```diff
@@ -2070,6 +2096,30 @@ export class Dataworker {
     return requiredAmount;
   }
 
+  /**
+   * Retrieves the amount of ETH required to execute a Linea relay leaf by querying the latest
+   * relay fee from the L2Linea Messenger contract.
+   * @param leaf The relay leaf to execute. Used in this function to prevent non-Linea chains from calling this method.
+   * @returns The amount of ETH required to execute the relay leaf.
+   */
+  _getRequiredEthForLineaRelayLeafExecution(client: SpokePoolClient): Promise<BigNumber | undefined> {
```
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1266/commits/ce854057309f4bb2239404c6a7ddd94ced76aaa6

src/dataworker/Dataworker.ts
james-a-morris: Given that we returned undefined I felt it could be classified as a "warn". But, now that we're throwing if the chain is incorrect then `error` aligns much more closely.

james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1266/commits/ce854057309f4bb2239404c6a7ddd94ced76aaa6

src/dataworker/Dataworker.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1266/commits/6ab93dbba4b5c26c429470189b68d7445a3647ca

src/dataworker/Dataworker.ts
james-a-morris: We could - I don't think it would hurt. Waiting for this to go in: https://github.com/across-protocol/sdk-v2/pull/580

src/dataworker/Dataworker.ts
```diff
@@ -2070,6 +2094,40 @@ export class Dataworker {
...
+    if (!sdkUtils.chainIsLinea(client.chainId)) {
+      this.logger.error({
+        at: "Dataworker#_getRequiredEthForLineaRelayLeafExecution",
+        message: "Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!",
+        chainId: client.chainId,
+      });
+      throw new Error("Tried to get required ETH for Linea relay leaf execution on a non-Linea chain!");
+    }
+    // Resolve and sanitize the L2MessageService contract ABI and address.
+    const l2MessageABI = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.abi;
```
nicholaspai, line 2115: are you missing a ? before the first .?

src/dataworker/Dataworker.ts
```diff
@@ -2070,6 +2094,40 @@ export class Dataworker {
...
+    const l2MessageAddress = CONTRACT_ADDRESSES[client.chainId].l2MessageService?.address;
```
nicholaspai, line 2116: are you missing a ? before the first .?

james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1266/commits/4a9351df2b98891b15ccb9a0455f1b547c97a364

src/dataworker/Dataworker.ts
james-a-morris: https://github.com/across-protocol/relayer-v2/pull/1266/commits/4a9351df2b98891b15ccb9a0455f1b547c97a364

src/dataworker/Dataworker.ts
```diff
@@ -1882,13 +1890,28 @@ export class Dataworker {
...
+              chainId: leaf.chainId,
+              tokens: l2TokensToCountTowardsSpokePoolLeafExecutionCapital(leaf.l2TokenAddress, leaf.chainId),
+              holder: client.spokePool.address,
+              amount: totalSent,
+            },
+          ];
+          // If we have to pass ETH via the payable function, then we need to add a balance request for the SpokePool
+          // to ensure that it has enough ETH to send.
+          // NOTE: this is ETH required separately from the amount required to send the tokens
+          if (isDefined(valueToPassViaPayable)) {
```
nicholaspai, line 1904: should this check be instead if `leaf.chainId is Linea` then set `amount = await this._getRequiredEthForLineaRelayLeafExecution(client)`?

nicholaspai (approved):

src/dataworker/Dataworker.ts
```diff
@@ -1882,13 +1890,28 @@ export class Dataworker {
...
-
+          const balanceRequestsToQuery = [
+            {
+              chainId: leaf.chainId,
+              tokens: l2TokensToCountTowardsSpokePoolLeafExecutionCapital(leaf.l2TokenAddress, leaf.chainId),
+              holder: client.spokePool.address,
+              amount: totalSent,
+            },
+          ];
+          // If we have to pass ETH via the payable function, then we need to add a balance request for the SpokePool
```
mrice32, line 1901: Sorry, please remind me: this ETH is coming from a separate balance the spokepool maintains? Or is it passed by the executing wallet to the spoke pool as a part of the transaction?

mrice32 (commented):
Looks good, just one question about the balance checking/request and where those funds are being sourced.

src/dataworker/Dataworker.ts
james-a-morris: Good call. We should actually be using the SpokePool signer's address because we're passing this in as a value. https://github.com/across-protocol/relayer-v2/pull/1266/commits/2371821f560f213497bf6a6ba636d05d07f714cc

mrice32 (approved):
LGTM!

PR #1267, opened 2024-03-07
improve(ProfitClient): Simulate v3 fills

pxrl:
The ProfitClient has, until now, used v2 fill estimates for its profitability calculations. This change upgrades to using v3 fills, which are slightly less expensive and should improve the relayer's perceived profitability when evaluating fills.

This is a pre-requisite for deprecating fillRelay() in the contracts.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,4 +1,3 @@
-import { random } from "lodash";
 import { Provider } from "@ethersproject/abstract-provider";
 import { utils as ethersUtils } from "ethers";
 import {
@@ -11,8 +10,8 @@ import {
 import * as constants from "../common/Constants";
 import {
   assert,
+  bnOne,
   bnZero,
-  bnUint32Max as uint32Max,
   bnUint256Max as uint256Max,
   fixedPointAdjustment as fixedPoint,
   BigNumber,
@@ -27,6 +26,7 @@ import {
   assign,
   CHAIN_IDs,
   TOKEN_SYMBOLS_MAP,
+  ZERO_ADDRESS,
 } from "../utils";
 import { Deposit, DepositWithBlock, L1Token, SpokePoolClientsByChain, V2Deposit, V3Deposit } from "../interfaces";
 import { HubPoolClient } from ".";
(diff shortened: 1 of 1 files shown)


pxrl:
> This change upgrades to using v3 fills, which are slightly less expensive and should improve the relayer's perceived profitability when evaluating fills.

I'll note that in testing, I resolve estimates that are actually a bit higher than what we currently see in prod. The Arbitrum estimates can be disregarded because the gas consumption varies according to the L1 gas price, but mainnet and OP stack estimates should be fairly reproducible. 

In general I think the current prod estimates are below what they should be, since some quick sampling shows v2 fills in the ~110k range, and as high as 140k:
https://etherscan.io/tx/0x906278d896eb8bd942fc89104c40324cd22289c679fabc356cb85dca30610d80
https://etherscan.io/tx/0xbc4173ec233763bdc101111bba4654614a133174f7e357c834f250b247146664
https://etherscan.io/tx/0x20c4028f84f2f97bcb59c12d6af87ef06814760e94f7e442ce00d9610e54b3fe
https://etherscan.io/tx/0xa114b4eded2051cf3d3ef3ec274327357edb6332722a279e5cf2a5aa7143189f
https://etherscan.io/tx/0xdd253aac1ff07051d115c5bc80c25f7cafd9138cd8607b8c7560c6130c65504e

Current prod estimates:
```
totalGasCosts: {
  1: {
    nativeGasCost: "95328"
    tokenGasCost: "5177209736935296"
  },
  10: {
    nativeGasCost: "95586"
    tokenGasCost: "167903901186687"
  },
  137: {
    nativeGasCost: "94687"
    tokenGasCost: "8658632391571694"
  },
  324: {
    nativeGasCost: "1429434"
    tokenGasCost: "142943400000000"
  },
  8453: {
    nativeGasCost: "95610"
    tokenGasCost: "160560708403474"
  },
  42161: {
    nativeGasCost: "2498845"
    tokenGasCost: "249884502498845"
  }
}
```

Updated v3 estimates:

```                                                                                                                                                                                                              
"totalGasCosts": {                                                                                                           
  "1": {                                                                                                                     
    "nativeGasCost": "114988",                                                                                               
    "tokenGasCost": "6218344315608492"                                                                                       
  },                                                                                                                         
  "10": {                                                                                                                    
    "nativeGasCost": "115271",                                                                                               
    "tokenGasCost": "155560689147678"                                                                                        
  },
  "137": {
    "nativeGasCost": "103688",
    "tokenGasCost": "10301685979082472"
  },
  "324": {
   "nativeGasCost": "2266592",
    "tokenGasCost": "226659200000000"
  },
  "8453": {
    "nativeGasCost": "115283",
    "tokenGasCost": "173237177962009"
  },
  "42161": {
    "nativeGasCost": "1887935",
    "tokenGasCost": "188793501887935"
  }
}
```

src/clients/ProfitClient.ts
```diff
@@ -684,40 +683,42 @@ export class ProfitClient {
...
+      depositId: 0,
+      depositor: TEST_RECIPIENT,
+      recipient: TEST_RECIPIENT,
+      inputToken: ZERO_ADDRESS, // Not verified by the SpokePool.
+      inputAmount: outputAmount.add(bnOne),
+      outputToken: "", // SpokePool-specific, overwritten later.
+      outputAmount,
+      originChainId: 0, // Not verified by the SpokePool.
+      destinationChainId: 0, // SpokePool-specific, overwritten later.
+      quoteTimestamp: currentTime - 60,
```
nicholaspai, line 701: Would changing any of these parameters change the logical path that the fillRelay function takes ? I guess we probably want to use the most commmon path as a gas estimate

nicholaspai (approved):
One comment

src/clients/ProfitClient.ts
pxrl: tldr the estimate of 95k was incorrect due to a round-to-zero error resulting from simulating a fill w/ fillAmount 1 unit of token. I've bumped this up to 100 units and the estimates immediately jumped up.

I later found WETH to be over-estimating the cost of the fill, so have additionally proposed to use USDC instead. It now produces estimates in the 130k gas ballpark, and that fits what I see in actual historical fills on mainnet.

The fix is proposed here and I will rebase this PR on top: https://github.com/across-protocol/relayer-v2/pull/1268

PR #1268, opened 2024-03-08
fix(ProfitClient): Correct gas estimation errors

pxrl:
Gas estimation for fills on most chains has been on the low side and has not been representative of the actual cost of filling. This has been traced to the fill amount being set to 1 unit of the destination token, which leads to fee rounding in the SpokePool, in turn producing a lower than expected gas estimate. This is not representative of the true cost of a fill.

Having addressed the first issue, the use of WETH on mainnet destined for a recipient with 0 balance has proven to produce over-estimates for the fill cost. This seems to be due to the marginal cost of initialising a recipient's ETH balance to a non-zero value, with the delta being about 25k gas for doing so (verified by testing against an "unused" EOA vs. one with an ETH balance). WETH additionally incurs the marginal withdrawal (unwrap) expense when sending to an EOA, leading to an eventual estimate of about 140k gas on mainnet. 140k is a bit on the high side because _most_ recipients would be unlikely to have a zero ETH balance, but their ERC20 balances are probably more regularly going to be zero.

When testing with USDC on mainnet, the resulting estimate is instead ~130k gas. This seems like a more reasonable assumption for the ProfitClient to make, based on actual onchain sampling of fills.

tldr: the before/after fill cost for mainnet is:
Before: ~95k gas
After: ~130k gas

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,4 +1,3 @@
-import { random } from "lodash";
 import { Provider } from "@ethersproject/abstract-provider";
 import { utils as ethersUtils } from "ethers";
 import {
@@ -11,8 +10,8 @@ import {
 import * as constants from "../common/Constants";
 import {
   assert,
+  bnOne,
   bnZero,
-  bnUint32Max as uint32Max,
   bnUint256Max as uint256Max,
   fixedPointAdjustment as fixedPoint,
   BigNumber,
@@ -27,6 +26,7 @@ import {
   assign,
   CHAIN_IDs,
   TOKEN_SYMBOLS_MAP,
+  ZERO_ADDRESS,
 } from "../utils";
 import { Deposit, DepositWithBlock, L1Token, SpokePoolClientsByChain, V2Deposit, V3Deposit } from "../interfaces";
 import { HubPoolClient } from ".";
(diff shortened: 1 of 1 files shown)


src/clients/ProfitClient.ts
```diff
@@ -683,13 +682,17 @@ export class ProfitClient {
...
     const depositId = random(uint32Max.toNumber()); // random depositId + "" originToken => ~impossible to collide.
-    const fillAmount = bnOne;
+    const fillAmount = toBN(100); // Avoid rounding to zero but ensure the relayer has sufficient balance to estimate.
+    const feePct = toBNWei("0.0001"); // 1bps
     const quoteTimestamp = getCurrentTime();
 
+    // Prefer USDC on mainnet because it's consistent in terms of gas estimation (no unwrap conditional).
+    // Prefer WETH on testnet because it's more likely to be configured for the destination SpokePool.
+    const [testSymbol, relayer] =
+      this.hubPoolClient.chainId === CHAIN_IDs.MAINNET ? ["WETH", PROD_RELAYER] : ["WETH", TEST_RELAYER];
```
nicholaspai, line 693: Shouldn't the LHS be USDC?

nicholaspai (approved):
https://github.com/across-protocol/relayer-v2/pull/1268#discussion_r1517108354

src/clients/ProfitClient.ts
pxrl: Yep; fixed: https://github.com/across-protocol/relayer-v2/pull/1268/commits/7b39e48943344541c0f019027f394e41ee449466

PR #1269, opened 2024-03-08
improve(ProfitClient): Simulate v3 fills

pxrl:
The ProfitClient has, until now, used v2 fill estimates for its profitability calculations. This change upgrades to using v3 fills, which are very slightly less expensive and should improve the relayer's perceived profitability when evaluating fills.

This is a pre-requisite for deprecating fillRelay() in the contracts.

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -1,4 +1,3 @@
-import { random } from "lodash";
 import { Provider } from "@ethersproject/abstract-provider";
 import { utils as ethersUtils } from "ethers";
 import {
@@ -11,8 +10,8 @@ import {
 import * as constants from "../common/Constants";
 import {
   assert,
+  bnOne,
   bnZero,
-  bnUint32Max as uint32Max,
   bnUint256Max as uint256Max,
   fixedPointAdjustment as fixedPoint,
   BigNumber,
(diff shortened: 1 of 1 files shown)


pxrl:
nb. I had to resubmit this PR from scratch because I hosed up my remote branch. This one is clean.

nicholaspai (approved):

nicholaspai:
> The ProfitClient has, until now, used v2 fill estimates for its profitability calculations. This change upgrades to using v3 fills, which are very slightly less expensive and should improve the relayer's perceived profitability when evaluating fills.
> 
> This is a pre-requisite for deprecating fillRelay() in the contracts.

How different are the gas costs?

pxrl:
> > The ProfitClient has, until now, used v2 fill estimates for its profitability calculations. This change upgrades to using v3 fills, which are very slightly less expensive and should improve the relayer's perceived profitability when evaluating fills.
> > This is a pre-requisite for deprecating fillRelay() in the contracts.
> 
> How different are the gas costs?

The estimates are pretty similar to what we have now (after the fix yesterday), but in practice I think we're usually saving about 5 - 10k gas on v3 fills vs. v2, on mainnet.

pxrl:
> This logic follows our API - I wonder if we should dump the simulation into the SDK as a util

idk - it's already there for the most part so I'm not sure there's much more to save.

PR #1270, opened 2024-03-08
feat(relayer): Support relaying for exclusive depositors or recipients

pxrl:
This feature will be used for third-parties to exclusively relay for their own purposes.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -46,7 +46,8 @@ export class Relayer {
    */
   private async _getUnfilledDeposits(): Promise<RelayerUnfilledDeposit[]> {
     const { configStoreClient, hubPoolClient, spokePoolClients, acrossApiClient } = this.clients;
-    const { relayerTokens, ignoredAddresses, acceptInvalidFills } = this.config;
+    const { relayerTokens, ignoredAddresses, acceptInvalidFills, exclusiveDepositors, exclusiveRecipients } =
+      this.config;
 
     const unfilledDeposits = await getUnfilledDeposits(spokePoolClients, hubPoolClient, this.config.maxRelayerLookBack);
 
@@ -90,6 +91,14 @@ export class Relayer {
           depositor,
           recipient,
         });
+        return false;
+      }
+
+      if (exclusiveDepositors.length > 0 && !exclusiveDepositors.includes(getAddress(depositor))) {
+        return false;
+      }
+
+      if (exclusiveRecipients.length > 0 && !exclusiveRecipients.includes(getAddress(recipient))) {
         return false;
       }
 
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -94,11 +94,11 @@ export class Relayer {
         return false;
       }
 
-      if (exclusiveDepositors.length > 0 && !exclusiveDepositors.includes(getAddress(depositor))) {
+      if (exclusiveDepositors?.length > 0 && !exclusiveDepositors.includes(getAddress(depositor))) {
         return false;
       }
 
-      if (exclusiveRecipients.length > 0 && !exclusiveRecipients.includes(getAddress(recipient))) {
+      if (exclusiveRecipients?.length > 0 && !exclusiveRecipients.includes(getAddress(recipient))) {
         return false;
       }
 
--- a/src/relayer/RelayerConfig.ts
+++ b/src/relayer/RelayerConfig.ts
@@ -3,6 +3,8 @@ import { CommonConfig, ProcessEnv } from "../common";
 import * as Constants from "../common/Constants";
 import { InventoryConfig } from "../interfaces";
 
+const { getAddress } = ethers.utils;
+
 export class RelayerConfig extends CommonConfig {
   readonly inventoryConfig: InventoryConfig;
   readonly debugProfitability: boolean;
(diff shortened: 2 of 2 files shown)


src/relayer/Relayer.ts
```diff
@@ -93,6 +94,14 @@ export class Relayer {
         return false;
       }
 
+      if (exclusiveDepositors.length > 0 && !exclusiveDepositors.includes(getAddress(depositor))) {
+        return false;
+      }
+
+      if (exclusiveRecipients.length > 0 && !exclusiveRecipients.includes(getAddress(recipient))) {
+        return false;
+      }
```
pxrl, line 103: nb. There's a gap here, because the addresses are not chain-dependent. It's implicitly assumed that the operator has limited the set of chains to the set that makes sense, but it would be nice to upgrade the config such that the depositor or recipient addresses could be defined per chain.

nicholaspai (approved):

src/relayer/RelayerConfig.ts
```diff
@@ -61,11 +67,11 @@ export class RelayerConfig extends CommonConfig {
     this.relayerDestinationChains = JSON.parse(RELAYER_DESTINATION_CHAINS ?? "[]");
 
     // Empty means all tokens.
-    this.relayerTokens = RELAYER_TOKENS
-      ? JSON.parse(RELAYER_TOKENS).map((token) => ethers.utils.getAddress(token))
-      : [];
+    this.relayerTokens = RELAYER_TOKENS ? JSON.parse(RELAYER_TOKENS).map((token: string) => getAddress(token)) : [];
+    this.exclusiveDepositors = JSON.parse(RELAYER_EXCLUSIVE_DEPOSITORS ?? "[]").map((addr: string) => getAddress(addr));
+    this.exclusiveRecipients = JSON.parse(RELAYER_EXCLUSIVE_RECIPIENTS ?? "[]").map((addr: string) => getAddress(addr));
```
james-a-morris, line 72: It's possible that the ENV could read an empty string instead of undefined/null

```suggestion
    this.exclusiveDepositors = JSON.parse(RELAYER_EXCLUSIVE_DEPOSITORS || "[]").map((addr: string) => getAddress(addr));
    this.exclusiveRecipients = JSON.parse(RELAYER_EXCLUSIVE_RECIPIENTS || "[]").map((addr: string) => getAddress(addr));
```

src/relayer/RelayerConfig.ts
```diff
@@ -61,11 +67,11 @@ export class RelayerConfig extends CommonConfig {
     this.relayerDestinationChains = JSON.parse(RELAYER_DESTINATION_CHAINS ?? "[]");
 
     // Empty means all tokens.
-    this.relayerTokens = RELAYER_TOKENS
-      ? JSON.parse(RELAYER_TOKENS).map((token) => ethers.utils.getAddress(token))
-      : [];
+    this.relayerTokens = RELAYER_TOKENS ? JSON.parse(RELAYER_TOKENS).map((token: string) => getAddress(token)) : [];
```
james-a-morris, line 70: Nit: follows the below changes
```suggestion
    this.relayerTokens = JSON.parse(RELAYER_TOKENS || "[]").map((token: string) => getAddress(token));
```

src/relayer/RelayerConfig.ts
```diff
@@ -61,11 +67,11 @@ export class RelayerConfig extends CommonConfig {
...
+    this.exclusiveDepositors = JSON.parse(RELAYER_EXCLUSIVE_DEPOSITORS ?? "[]").map((addr: string) => getAddress(addr));
+    this.exclusiveRecipients = JSON.parse(RELAYER_EXCLUSIVE_RECIPIENTS ?? "[]").map((addr: string) => getAddress(addr));
     this.slowDepositors = SLOW_DEPOSITORS
-      ? JSON.parse(SLOW_DEPOSITORS).map((depositor) => ethers.utils.getAddress(depositor))
+      ? JSON.parse(SLOW_DEPOSITORS).map((depositor: string) => getAddress(depositor))
```
james-a-morris, line 74: Can't leave a suggestion but should we also follow the above convention?

src/relayer/Relayer.ts
```diff
@@ -93,6 +94,14 @@ export class Relayer {
         return false;
       }
 
+      if (exclusiveDepositors?.length > 0 && !exclusiveDepositors.includes(getAddress(depositor))) {
```
james-a-morris, line 97: OOC why is this optional? It makes sense why `exclusiveDepositors` could be empty, but it looks like we define it in the config in all cases

src/relayer/Relayer.ts
```diff
@@ -93,6 +94,14 @@ export class Relayer {
...
+        return false;
+      }
+
+      if (exclusiveRecipients?.length > 0 && !exclusiveRecipients.includes(getAddress(recipient))) {
```
james-a-morris, line 101: Similar with ^

james-a-morris (approved):
Left a few comments. Look great

src/relayer/Relayer.ts
pxrl: It's optional because we have quite a few tests that supply an object for the relayer config, and assuming it is defined then forces it to be defined in all of those tests. We have _so_ many config items that it's not feasible to define them all in test, and in practice we don't want to instantiate the actual relayer config because it reads from the env and would affect the tests.

src/relayer/RelayerConfig.ts
pxrl: I'm not sure I see the point of this change - wouldn't `||` make it more likely to unintentionally fall back to an empty array? `RELAYER_EXCLUSIVE_RECIPIENTS` is defined in any way then I'd argue it should be parsed, and if it's set incorrectly then we should fail immediately, rather than silently reverting to an empty array.

pxrl:
I'll park this PR for now. It was driven by a one-time use case, but it's not something that's consistently required, so there's no reason to add it to the codebase. If we need it in future it can be revived.

PR #1271, opened 2024-03-08
improve(relayer): Don't query irrelevant HubPool events

pxrl:
The relayer has no obvious need to query root bundle events, so skip them.

Note: Tests all pass, and I've also tested this on mainnet. It just needs the corresponding SDK-v2 PR to be merged, and a new package to be built (https://github.com/across-protocol/sdk-v2/pull/581).

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -31,6 +31,11 @@ export async function constructRelayerClients(
   const commonClients = await constructClients(logger, config, baseSigner);
   const { configStoreClient, hubPoolClient } = commonClients;
   await updateClients(commonClients, config);
+  await hubPoolClient.update([
+    "CrossChainContractsSet",
+    "L1TokenEnabledForLiquidityProvision",
+    "SetPoolRebalanceRoute",
+  ]);
 
   // If both origin and destination chains are configured, then limit the SpokePoolClients instantiated to the
   // sum of them. Otherwise, do not specify the chains to be instantiated to inherit one SpokePoolClient per
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.11",
     "@across-protocol/contracts-v2": "2.5.0-beta.7",
-    "@across-protocol/sdk-v2": "0.22.6",
+    "@across-protocol/sdk-v2": "0.22.10",
     "@arbitrum/sdk": "^3.1.3",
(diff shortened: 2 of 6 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -31,6 +31,11 @@ export async function constructRelayerClients(
   const commonClients = await constructClients(logger, config, baseSigner);
   const { configStoreClient, hubPoolClient } = commonClients;
   await updateClients(commonClients, config);
+  await hubPoolClient.update([
```
nicholaspai, line 34: So which events are we skipping in the relayer now?

nicholaspai (approved):
One question. It's not clear to me which events we're skipping now so just want to confirm

src/relayer/RelayerClientHelper.ts
pxrl: Ah, when we don't specify the events to query, we inherit the complete set (see below). So we are cutting out:
`ProposeRootBundle`, `RootBundleCanceled`, `RootBundleDisputed`, `RootBundleExecuted`.

```
type HubPoolEvent =                                                                                                                                                                                                              
  | "SetPoolRebalanceRoute"                                                                                                                                                                                                      
  | "L1TokenEnabledForLiquidityProvision"                                                                                                                                                                                        
  | "ProposeRootBundle"                                                                                                                                                                                                          
  | "RootBundleCanceled"                                                                                                                                                                                                         
  | "RootBundleDisputed"                                                                                                                                                                                                         
  | "RootBundleExecuted"                                                                                                                                                                                                         
  | "CrossChainContractsSet";
```

james-a-morris: To poke at the alternative - would it be more expressive to choose routes to _disable_ instead of enable? I.e. we are specifying an allow-list, but would it be more expressive given our use-case to have a disallow-list?

james-a-morris (commented):
https://github.com/across-protocol/relayer-v2/pull/1271#discussion_r1517980098

src/relayer/RelayerClientHelper.ts
nicholaspai: I think its better to specify which to enable, since in the future if we add more events we won't get unexpected behavior

pxrl: The existing behaviour makes most sense for me - if the caller is explicit about which events they want, then that's what they get - otherwise they get the complete set.

PR #1272, opened 2024-03-08
improve(relayer): Drop unused SpokePool events

pxrl:
These events are not required by the relayer, so there's no need to query them.

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -134,8 +134,6 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
     "RequestedV3SlowFill",
     "FilledV3Relay",
     "EnabledDepositRoute",
-    "RelayedRootBundle",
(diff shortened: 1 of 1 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -134,8 +134,6 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
     "RequestedV3SlowFill",
     "FilledV3Relay",
     "EnabledDepositRoute",
-    "RelayedRootBundle",
-    "ExecutedRelayerRefundRoot",
```
pxrl, line 138: tbd on `RelayedRootBundle`, but I have confirmed that `ExecutedRelayerRefundRoot` is not used by the relayer.

pxrl: I can't find any relayer use of `RelayedRootBundle` either.

nicholaspai (approved):

PR #1275, opened 2024-03-09
improve(relayer): Limit HubPoolClient lookback

pxrl:
The HubPoolClient now has logic to override the requested lookback for specific events if it has not previously been updated. This is the same logic that is implemented for the SpokePoolClients. As a result, it's now possible to be much more restrictive on the lookback for the HubPoolClient. The relayer can benefit from this be only querying RootBundle events for the past ~12 hours, which should be sufficient for the InventoryManager to determine repayment chains.

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -28,7 +28,8 @@ export async function constructRelayerClients(
   baseSigner: Signer
 ): Promise<RelayerClients> {
   const signerAddr = await baseSigner.getAddress();
-  const commonClients = await constructClients(logger, config, baseSigner);
+  const hubPoolLookBack = 3600 * 12;
+  const commonClients = await constructClients(logger, config, baseSigner, hubPoolLookBack);
   const { configStoreClient, hubPoolClient } = commonClients;
   await updateClients(commonClients, config);
   await hubPoolClient.update();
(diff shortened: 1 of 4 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -28,7 +28,8 @@ export async function constructRelayerClients(
   baseSigner: Signer
 ): Promise<RelayerClients> {
   const signerAddr = await baseSigner.getAddress();
-  const commonClients = await constructClients(logger, config, baseSigner);
+  const hubPoolLookBack = 3600 * 12;
```
pxrl, line 31: nb. I chose to hardcode this directly where it's used because it doesn't seem like anything we should tune. I didn't necessarily want to park it in `src/common/Constants.ts` because it's so relayer-specific, but maybe that would be the right move anyway.

nicholaspai, line 31:
```suggestion
  // The relayer only uses the HubPoolClient to query bundle refunds for the latest validated bundle and the pending bundle, therefore 12 hours should cover these bundles in almost all cases.
  const hubPoolLookBack = 3600 * 12;
```

nicholaspai (approved):
Just one comment suggestion!

src/relayer/RelayerClientHelper.ts
pxrl: Good call - added: https://github.com/across-protocol/relayer-v2/pull/1275/commits/68065cf815b7546541222a11b8ec37d77b557085. Thanks!

PR #1278, opened 2024-03-10
improve: Hardcode bundle refund lookback for relayer and monitor

nicholaspai:
We never practically need to look back more than 1 bundle, and trying to look back too many bundles leads weird errors in InventoryClient because it tries to compute bundle data using events its missing due to shortened HubPoolClient and SpokePoolClient lookbacks

Also, changes InventoryClient behavior that defaults taking repayment on mainnet if the BundleDataClient fails for some reason. We've seen in practice that taking repayment on mainnet for extended periods of time locks up bridge liquidity, so we'd actually rather take extra repayment on the spokes than on mainnet because we can always manually rebalance using external fast bridges or slow fills, rather than locking it up in 7 day ORU bridges

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -42,6 +42,7 @@ export class InventoryClient {
   private logDisabledManagement = false;
   private readonly scalar: BigNumber;
   private readonly formatWei: ReturnType<typeof createFormatFunction>;
+  private bundleRefundLookback = 1;
 
   constructor(
     readonly relayer: string,
@@ -53,7 +54,6 @@ export class InventoryClient {
     readonly bundleDataClient: BundleDataClient,
     readonly adapterManager: AdapterManager,
     readonly crossChainTransferClient: CrossChainTransferClient,
-    readonly bundleRefundLookback = 2,
     readonly simMode = false
   ) {
     this.scalar = sdkUtils.fixedPointAdjustment;
@@ -157,9 +157,9 @@ export class InventoryClient {
 
   // Return the upcoming refunds (in pending and next bundles) on each chain.
   async getBundleRefunds(l1Token: string): Promise<{ [chainId: string]: BigNumber }> {
-    // Increase virtual balance by pending relayer refunds from the latest valid bundles.
-    // Allow caller to set how many bundles to look back for refunds. The default is set to 2 which means
-    // we'll look back only at the two latest valid bundle unless the caller overrides.
+    // Increase virtual balance by pending relayer refunds from the latest valid bundle and the
+    // upcoming bundle. We can assume that all refunds from the second latest valid bundle have already
+    // been executed.
     const refundsToConsider: CombinedRefunds[] = await this.bundleDataClient.getPendingRefundsFromValidBundles(
       this.bundleRefundLookback
     );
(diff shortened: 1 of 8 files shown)


src/monitor/Monitor.ts
```diff
@@ -558,8 +558,9 @@ export class Monitor {
   }
 
   async updateLatestAndFutureRelayerRefunds(relayerBalanceReport: RelayerBalanceReport): Promise<void> {
+    // We realistically only need to check for refunds for the latest validated bundle.
     const validatedBundleRefunds: CombinedRefunds[] =
-      await this.clients.bundleDataClient.getPendingRefundsFromValidBundles(this.monitorConfig.bundleRefundLookback);
+      await this.clients.bundleDataClient.getPendingRefundsFromValidBundles(1);
```
james-a-morris, line 563: Could we not use `this.monitorConfig.bundleRefundLookback` here? We may need to remove the `private` casting

james-a-morris (commented):
Left some comments regarding the code.

--
Regarding the PR as a whole: were the errors produced by the inventory client spurred by recent changes to improve the speed of the relayer?

src/monitor/Monitor.ts
nicholaspai: it shouldn't exist in the config anymore, the point of this PR is to hardcode it so caller can't override

nicholaspai:
> Left some comments regarding the code.
> 
> -- Regarding the PR as a whole: were the errors produced by the inventory client spurred by recent changes to improve the speed of the relayer?

there are no errors this is jsut an optimization

src/clients/InventoryClient.ts
```diff
@@ -226,16 +226,18 @@ export class InventoryClient {
...
       // Consider any refunds from executed and to-be executed bundles.
       totalRefundsPerChain = await this.getBundleRefunds(l1Token);
     } catch (e) {
       this.log("Failed to get bundle refunds, defaulting refund chain to hub chain");
-      // Fallback to getting refunds on Mainnet if calculating bundle refunds goes wrong.
-      // Inventory management can always rebalance from Mainnet to other chains easily if needed.
-      return hubChainId;
+      // Fallback to ignoring bundle refunds if calculating bundle refunds goes wrong.
+      // This would create issues if there are relatively a lot of upcoming relayer refunds that would affect
+      // the relayer's repayment chain of choice.
```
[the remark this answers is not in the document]
nicholaspai: it already defaults to this value on line 230?

src/monitor/Monitor.ts
james-a-morris: Totally - I was referencing `src/clients/InventoryClient.ts`. 

Followup: thoughts on making this a const variable since it's hardcoded? If we added a `BUNDLE_REFUND_LOOKBACK = 1,`, we could use it in both the inventory client & this monitor code.

nicholaspai: what about defaulting to 1 https://github.com/across-protocol/relayer-v2/pull/1278/commits/f90c93e1c7b5f961c5480e4c4d5636e200265ea0

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
james-a-morris: Fair - this one isn't really that actionable. More of a nit

PR #1280, opened 2024-03-11
fix: InventoryClient should set default values for `totalRefundsPerChain`

nicholaspai:
d35608d1b5522c4f12b0d018c8de173d1cbe8f02 changed behavior of this try-catch such that if catch is called, code continues rather than short circuiting. Therefore we need a default value for totalRefunds≈


mrice32 (approved):
LGTM!

PR #1282, opened 2024-03-11
feat: permanently cache allowances for l1 canonical token bridges once >

nicholaspai:
Allowances should never decrease in practice once set to MAX_UINT so we can save a lot of eth_calls

### DIFF
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -27,7 +27,9 @@ import {
   createFormatFunction,
   BigNumberish,
   TOKEN_SYMBOLS_MAP,
+  getRedisCache,
 } from "../../utils";
+import { utils } from "@across-protocol/sdk-v2";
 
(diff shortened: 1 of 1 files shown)


src/clients/bridges/BaseAdapter.ts
```diff
@@ -107,31 +109,56 @@ export abstract class BaseAdapter {
     return { ...this.spokePoolClients[chainId].eventSearchConfig };
   }
 
+  getAllowanceCacheKey(l1Token: string, targetContract: string): string {
+    return `l1CanonicalTokenBridgeAllowance_${l1Token}_targetContract:${targetContract}`;
```
mrice32, line 113: nit: I think key should include the relayer address too, so two bots running on different addresses but sharing a cache wouldn't break.

mrice32 (approved):
LGTM! One nit

src/clients/bridges/BaseAdapter.ts
nicholaspai: great callout

PR #1283, opened 2024-03-11
improve: Set more reasonable BUNDLE_END_BLOCK_BUFFERS

nicholaspai:
These  values control how much the proposed bundle block ranges follow HEAD on each l2. These should be set as low as possible but also still higher than any reasonable finality risk for the l2. These were 
set higher in the past but setting them lower will result in more frequent bundles being proposed and faster refunds on average for fillers

Also, the reason why I choose to reset the defaults rather than propose a local bot config change is that we want the disputer and proposer to be aligned on these buffers, otherwise there is a slight chance that the disputer thinks the buffer should be longer than the proposer does, resulting in the disputer choosing not to send a dispute until the buffer time has been cleared. Currently the disputer is programmed not to dispute any bundle whose end blocks are greater than the HEAD block but not more than the buffer period, because the disputer is implicitly acknowledging that it is within the possible finality period.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -239,7 +239,10 @@ export class InventoryClient {
         })
       );
     }
-    this.log(`Time taken to get bundle refunds: ${Math.floor(Date.now() - startTime) / 1000}s`, totalRefundsPerChain);
+    this.log(`Time taken to get bundle refunds: ${Math.floor(Date.now() - startTime) / 1000}s`, {
+      l1Token: l1Token,
+      totalRefundsPerChain,
+    });
 
     // Add upcoming refunds going to this destination chain.
     chainVirtualBalanceWithShortfallPostRelay = chainVirtualBalanceWithShortfallPostRelay.add(
@@ -263,6 +266,7 @@ export class InventoryClient {
     const refundChainId = expectedPostRelayAllocation.gt(targetPct) ? hubChainId : destinationChainId;
 
     this.log("Evaluated refund Chain", {
+      l1Token: l1Token,
       chainShortfall,
       chainVirtualBalance,
       chainVirtualBalanceWithShortfall,
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -27,7 +27,9 @@ import {
   createFormatFunction,
   BigNumberish,
   TOKEN_SYMBOLS_MAP,
+  getRedisCache,
 } from "../../utils";
+import { utils } from "@across-protocol/sdk-v2";
 
 import { CONTRACT_ADDRESSES } from "../../common";
 import { OutstandingTransfers, SortableEvent } from "../../interfaces";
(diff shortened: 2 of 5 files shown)


src/clients/bridges/BaseAdapter.ts
```diff
@@ -27,7 +27,9 @@ import {
   createFormatFunction,
   BigNumberish,
   TOKEN_SYMBOLS_MAP,
+  getRedisCache,
```
mrice32, line 30: nit: I think the cache PR got included in this one accidentally FYI

src/common/Constants.ts
```diff
@@ -129,14 +129,17 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
...
+  // the mainnet bundle end block for the next bundle.
+  10: 60, // Optimism finality is inherited from Ethereum's
+  137: 128, // Polygon reorgs often so this number is set larger than the largest observed reorg.
   288: 0, // **UPDATE** 288 is disabled so there should be no buffer.
-  324: 1500, // At 1s/block, 25 mins = 1500 blocks.
-  8453: 750, // At 2s/block, 25 mins = 750 blocks.
-  42161: 300, // At a conservative 1 TPS, 5 mins = 300 seconds = 300 transactions. And 1 block per txn.
+  324: 60, // ZkSync is a centralized sequencer but is relatively unstable so this is kept higher than 0
+  8453: 60, // Same finality profile as Optimism
+  42161: 0, // Arbitrum is a centralized sequencer
```
mrice32, line 142: I'm not sure I agree that some should be set to 0 based on re-orgability.

I thought they were set to ensure that if you see a fill within the block range, the deposit is always visible to you, regardless of the timing of the queries within a reasonable run. For instance, if a deposit is made at time T and the fill is at T + 1, but I query the fill chain 4 seconds after I query the deposit chain, I could have missed the deposit but gotten the fill, meaning I would call the fill invalid, right? 

If you agree with that reasoning, I'd argue that you might want to do ~1 minute or so on every chain to completely rule out query-timing based issues.

Thoughts?

mrice32 (commented):
Just one comment

src/clients/InventoryClient.ts
```diff
@@ -239,7 +239,10 @@ export class InventoryClient {
         })
       );
     }
-    this.log(`Time taken to get bundle refunds: ${Math.floor(Date.now() - startTime) / 1000}s`, totalRefundsPerChain);
+    this.log(`Time taken to get bundle refunds: ${Math.floor(Date.now() - startTime) / 1000}s`, {
```
nicholaspai, line 242: linting caught this

src/common/Constants.ts
nicholaspai: good point, otherwise we'd self-dispute

src/common/Constants.ts
```diff
@@ -129,14 +129,18 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
...
+  1: 10, // 12s/block
+  10: 60, // 2s/block
+  137: 128, // 2s/block. Polygon reorgs often so this number is set larger than the largest observed reorg.
   288: 0, // **UPDATE** 288 is disabled so there should be no buffer.
-  324: 1500, // At 1s/block, 25 mins = 1500 blocks.
-  8453: 750, // At 2s/block, 25 mins = 750 blocks.
-  42161: 300, // At a conservative 1 TPS, 5 mins = 300 seconds = 300 transactions. And 1 block per txn.
+  324: 120, // 2s/block. ZkSync is a centralized sequencer but is relatively unstable so this is kept higher than 0
+  8453: 60, // 2s/block. Same finality profile as Optimism
+  42161: 120, // 1s/block. Arbitrum is a centralized sequencer
```
mrice32, line 143:
```suggestion
  42161: 240, // ~0.25s/block. Arbitrum is a centralized sequencer
```

src/common/Constants.ts
```diff
@@ -129,14 +129,18 @@ export const CHAIN_MAX_BLOCK_LOOKBACK = {
...
-  10: 150, // 2s/block, 5 mins = 300 seconds = 300 transactions. And 1 block per txn.
-  137: 750, // At 2s/block, 25 mins = 25 * 60 / 2 = 750 blocks
+  1: 10, // 12s/block
+  10: 60, // 2s/block
+  137: 128, // 2s/block. Polygon reorgs often so this number is set larger than the largest observed reorg.
   288: 0, // **UPDATE** 288 is disabled so there should be no buffer.
-  324: 1500, // At 1s/block, 25 mins = 1500 blocks.
-  8453: 750, // At 2s/block, 25 mins = 750 blocks.
-  42161: 300, // At a conservative 1 TPS, 5 mins = 300 seconds = 300 transactions. And 1 block per txn.
+  324: 120, // 2s/block. ZkSync is a centralized sequencer but is relatively unstable so this is kept higher than 0
```
mrice32, line 141:
```suggestion
  324: 120, // ~1s/block. ZkSync is a centralized sequencer but is relatively unstable so this is kept higher than 0
```

mrice32 (approved):
Looks great! Two minor suggestions:

One just to correct the block time comment.

The second is because arbitrum is much faster now than it was previously, so the number of blocks should be increased.

PR #1285, opened 2024-03-12
improve(Dataworker): Don't send error-level log if invalid bundle reason is expected to be benign

nicholaspai:
This copies the logic from `validatePendingRootBundle` that also should be used in executePoolRebalanceLeaf

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1424,13 +1424,27 @@ export class Dataworker {
     );
 
     if (!valid) {
-      this.logger.error({
-        at: "Dataworker#executePoolRebalanceLeaves",
-        message: "Found invalid proposal after challenge period!",
-        reason,
-        e: reason,
-        notificationPath: "across-error",
-      });
+      // In the case where the Dataworker config is improperly configured, emit an error level alert so bot runner
+      // can get dataworker running ASAP.
+      if (ERROR_DISPUTE_REASONS.has(reason)) {
+        this.logger.error({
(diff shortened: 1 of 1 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1424,13 +1424,27 @@ export class Dataworker {
...
+          message: "Dataworker configuration error, should resolve itself eventually 😪",
+          reason,
+        });
+      } else {
+        this.logger.error({
+          at: "Dataworker#executePoolRebalanceLeaves",
+          message: "Found invalid proposal after challenge period!",
+          reason,
+        });
+      }
```
james-a-morris, line 1447: Thoughts on reducing the # of conditionals?
```suggestion
    if (IGNORE_DISPUTE_REASONS.has(reason)) {
      this.logger.debug({
        at: "Dataworker#executePoolRebalanceLeaves",
        message:
          "Dataworker configuration error, should resolve itself eventually 😪",
        reason,
      });
    } else {
      this.logger.error({
        at: "Dataworker#executePoolRebalanceLeaves",
        message: ERROR_DISPUTE_REASONS.has(reason)
          ? "Dataworker configuration error needs to be fixed"
          : "Found invalid proposal after challenge period!",
        reason,
      });
    }
```

james-a-morris (approved):
LGTM - left a comment on the # of conditionals since both the first if & the last else use `this.logger.error(...)`

mrice32 (approved):
LGTM!

PR #1286, opened 2024-03-12
improve: Send quorum warning to across-warn channel

nicholaspai:
We should collect warnings that we want to see but don't want to pollute other channels in a new `across-warn` channel

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -440,6 +440,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       logger.debug({
         at: "ProviderUtils",
         message: "Some providers mismatched with the quorum result or failed 🚸",
+        notificationPath: "across-warn",
         method,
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -437,7 +437,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       .filter(([, result]) => compareRpcResults(method, result, quorumResult))
       .map(([provider]) => provider.connection.url);
     if (Object.keys(mismatchedProviders).length > 0 || errors.length > 0) {
-      logger.warn({
+      logger.debug({
```
pxrl, line 440: Agree on it being unsustainably noisy, and I'm not strongly opposed to this. Just to consider alternatives, wdyt about submitting it to a different slack channel so it doesn't pollute the main relayer channel? That's consistent with some other "soft" errors that became annoyingly frequent.

PR #1293, opened 2024-03-14
improve: Change some dataworker tests to use V3 functions instead of V2

nicholaspai:
Helps support removing V2 reliance in Dataworker

### DIFF
--- a/test/Dataworker.executePoolRebalances.ts
+++ b/test/Dataworker.executePoolRebalances.ts
@@ -11,7 +11,6 @@ import {
   Contract,
   FakeContract,
   SignerWithAddress,
-  buildFillForRepaymentChain,
   depositV3,
   ethers,
   expect,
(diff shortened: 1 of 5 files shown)


pxrl (approved):

pxrl:
@nicholaspai I cut the tail end off some of these tests because they previously relied on making two partial fills for the same deposit. We obviously can't do this in v3, so the test would be contrived. We could instead flip it so that we request a slow fill on the first bundle, and then make a full fill to replace it later. But I'm not sure if this makes sense to bother with, since we should already test that elsewhere. wdyt?

nicholaspai:
> @nicholaspai I cut the tail end off some of these tests because they previously relied on making two partial fills for the same deposit. We obviously can't do this in v3, so the test would be contrived. We could instead flip it so that we request a slow fill on the first bundle, and then make a full fill to replace it later. But I'm not sure if this makes sense to bother with, since we should already test that elsewhere. wdyt?

Makes sense. The bundle construction is tested in depth in loadData so these sections are redundant. These tests were made a LONG time ago.

PR #1294, opened 2024-03-14
chore: Drop v2 events from monitor

pxrl:
In advance of bumping contracts-v2.


nicholaspai (approved):

src/monitor/MonitorClientHelper.ts
```diff
@@ -81,9 +81,6 @@ export async function constructMonitorClients(
 
 export async function updateMonitorClients(clients: MonitorClients): Promise<void> {
   await updateSpokePoolClients(clients.spokePoolClients, [
-    "FundsDeposited",
-    "RequestedSpeedUpDeposit",
-    "FilledRelay",
```
pxrl, line 86: Technically we can still see these, but it's very unlikely and not especially interesting to track, so drop it now.

PR #1295, opened 2024-03-14
feat: Add optional secondary simulation option in MulticallerClient

nicholaspai:
Can be used by Relayer to add a second simulation for certain transactions that we expect ethers to not always be correct about. For example, filling deposits with messages sent to Proxy recipients on Base has been succeeding in the ethers txn simulation but failing when actually submitted.

It's unclear whether the recipient proxy or the Base network is at fault but we do know that the Alchemy gas estimator SDK and the ethers gas estimator disagree, with the former catching the failed txn and the latter not.

Resolves ACX-1969


src/clients/MultiCallerClient.ts
```diff
@@ -293,6 +293,7 @@ export class MultiCallerClient {
       contract: multisender,
       method: "aggregate",
       args: [callData],
+      alchemySimulate: transactions.some((txn) => txn.alchemySimulate),
```
nicholaspai, line 296: this will cost one Alchemy SDK async call per batch of txns containing a deposit with a message, which I think is a very low cost

nicholaspai:
@pxrl and I concluded the issue this PR was trying to resolve (one RPCs eth_call was disagreeing with anotehrs) was due to a one-off RPC issue likely arising from the Dencun hardfork

PR #1296, opened 2024-03-15
improve(finalizer): Each finalizer adapter should define its own filtering logic

nicholaspai:
The finalizer's speed bottleneck currently is checking Transaction status for each L2 TokensBridged event, so we should optimize how we filter for these events. Some chains we want to only look for events 
older than some time, for example for ORU's that take 7 days to be ready for finalization. On other chains we want to stop looking for events older than some time, like for Polygon which can finalize in as 
soon as 15 mins and never longer than 1 day. Some chains like Linea and ZkSync take 8 hours or 1 day to finalize but we shouldn't then be looking later than 2 days for those events. Finally, OpStack also 
has a secondary look up to submit proofs to initiate withdrawals, and we shouldn't look more than 1 day for these.

Clearly each chain has its own set of filters so this PR refactors the confusing and increasingly rigid `latestBlockToFinalize` param in favor of defining filters within each adapter's code.


dohaki (approved):
LGTM

PR #1301, opened 2024-03-15
feat: Executor should update exchange rates after executing Ethereum RelayerRefundLeaves in some cases

nicholaspai:
This handles the edge case where we need to update liquid reserves in the HubPool following a relayer refund execution in order to execute non-Ethereum PoolRebalanceLeaves. (Recall the Ethereum PoolRebalanceLeaf is always executed first)


src/dataworker/Dataworker.ts
```diff
@@ -1556,10 +1556,66 @@ export class Dataworker {
...
 
+    // Before executing the other pool rebalance leaves, see if we should update any exchange rates to account for
+    // any tokens returned to the hub pool via the EthereumSpokePool that we'll need to use to execute
+    // any of the remaining pool rebalance leaves.
+    const nonHubChainPoolRebalanceLeaves = unexecutedLeaves.filter((leaf) => leaf.chainId !== hubPoolChainId);
+    await sdkUtils.forEachAsync(nonHubChainPoolRebalanceLeaves, async (leaf) => {
+      await sdkUtils.forEachAsync(leaf.l1Tokens, async (l1Token, idx) => {
+        // If leaf's netSendAmount is negative, then we don't need to updateExchangeRates since the Hub will not
+        // have a liquidity constraint because it won't be sending any tokens.
+        if (leaf.netSendAmounts[idx].gt(0)) {
```
pxrl, line 1567: Could you apply this as a filter stage? That'd remove a level of indentation.

src/dataworker/Dataworker.ts
```diff
@@ -1556,10 +1556,66 @@ export class Dataworker {
...
+            );
+            if (virtualHubPoolBalance.gte(leaf.netSendAmounts[idx])) {
+              this.logger.debug({
+                at: "Dataworker#executePoolRebalanceLeaves",
+                message: `Relayer refund leaf will return enough funds to HubPool to execute PoolRebalanceLeaf, updating exchange rate for ${l1Token}`,
+                currHubPoolBalance,
+                virtualHubPoolBalance,
+                netSendAmount: leaf.netSendAmounts[idx],
+              });
+              if (submitExecution) {
```
pxrl, line 1597: Is the `submitExecution` clause needed? iirc we normally make `multiCallerClient.executeTransactionQueue()` conditional upon it anyway, so it _should_ be fine to populate the queue even if we don't expect to execute it.

src/dataworker/Dataworker.ts
```diff
@@ -1556,10 +1556,66 @@ export class Dataworker {
...
+                const tokenSymbol = this.clients.hubPoolClient.getTokenInfo(hubPoolChainId, l1Token)?.symbol;
```
pxrl, line 1598: Might be nice to bump this higher up, so it can be used in the log message at line 1592.

pxrl (approved):

src/dataworker/Dataworker.ts
nicholaspai: I believe we should re-use whatever execution logic we use in `_updateExchangeRate`

src/dataworker/Dataworker.ts
nicholaspai: I'll early exit instead, its not super easy to filter

src/dataworker/Dataworker.ts
nicholaspai: what would you add to the logic?

nicholaspai:
We should probably add a test for this so I'll wait on merging

src/dataworker/Dataworker.ts
pxrl: Just print the symbol in the log message, rather than the address. It's only a minor point though.

PR #1305, opened 2024-03-18
improve: increase default MDC 

nicholaspai:
These default values should be higher than highest levels we override with in the bot-configs, so this PR matches those values. Particularly on optimism.


pxrl (approved):
+1, it's also important that `DEFAULT_MIN_DEPOSIT_CONFIRMATIONS` is _at least_ as large as the other largest MDC for a specific sum.

PR #1306, opened 2024-03-18
fix(relayer): Reinstate testnet support

pxrl:
The recent change to limit the relayer lookback in production did not consider that root bundles are not proposed every couple of ours on testnet. This rendered the relayer unable to complete a run.


pxrl:
fwiw the testnet relayer has been running successfully against an older commit, so there hasn't been any service disruption.

nicholaspai (approved):

PR #1308, opened 2024-03-19
feat: Smarter exchange rate updates

nicholaspai:
This is from a comment added in `Dataworker.ts` that also explains this PR well:

```
// There are three times that we should look to update the HubPool's liquid reserves:
// 1. First, before we attempt to execute the HubChain PoolRebalance leaves and RelayerRefund leaves.
//    We should see if there are new liquid reserves we need to account for before sending out these
//    netSendAmounts.
// 2. Second, before we attempt to execute the PoolRebalance leaves for the other chains. We should
//    see if there are new liquid reserves we need to account for before sending out these netSendAmounts. This
//    updated liquid reserves balance could be from previous finalizations or any amountToReturn value sent
//    back from the Ethereum RelayerRefundLeaves.
// 3. Third, we haven't updated the exchange rate for an L1 token on a PoolRebalanceLeaf in a while that
//    we're going to execute, so we should batch in an update.
```

The motivation for this PR is to reduce the # of `exchangeRateCurrent()` calls to reduce dataworker costs. This PR should make it such that these calls are only batched with PoolRebalanceLeaf executions and never sent otherwise.


src/dataworker/Dataworker.ts
```diff
@@ -1720,15 +1695,185 @@ export class Dataworker {
...
+      if (currentLiquidReserves.gte(netSendAmounts[idx])) {
+        this.logger.debug({
+          at: "Dataworker#_updateExchangeRatesBeforeExecutingHubChainLeaves",
+          message: `Skipping exchange rate update for ${tokenSymbol} because current liquid reserves > netSendAmount`,
+          currentLiquidReserves,
+          netSendAmount: netSendAmounts[idx],
+          l1Token,
+        });
+        return;
+      }
```
mrice32, line 1736: Great great optimization. +1

src/dataworker/Dataworker.ts
```diff
@@ -1492,11 +1475,28 @@ export class Dataworker {
...
+    let updatedLiquidReserves: Record<string, BigNumber> = {};
+
     // First, execute mainnet pool rebalance leaves. Then try to execute any relayer refund and slow leaves for the
     // expected relayed root hash, then proceed with remaining pool rebalance leaves. This is an optimization that
     // takes advantage of the fact that mainnet transfers between HubPool and SpokePool are atomic.
     const mainnetLeaves = unexecutedLeaves.filter((leaf) => leaf.chainId === hubPoolChainId);
     if (mainnetLeaves.length > 0) {
+      assert(mainnetLeaves.length === 1);
+      updatedLiquidReserves = await this._updateExchangeRatesBeforeExecutingHubChainLeaves(
+        mainnetLeaves[0],
```
mrice32, line 1497: OOC, why only the 0th leaf. If we had multiple (which I know we don't today), would we not want to update for them all?

src/dataworker/Dataworker.ts
```diff
@@ -1737,14 +1882,14 @@ export class Dataworker {
         this.clients.hubPoolClient.currentTime === undefined ||
         this.clients.hubPoolClient.currentTime - lastestFeesCompoundedTime <= 7200 // 2 hours
```
mrice32, line 1883: Two questions:
1. Do you remember why we decided that we needed to regularly update these?
2. Can we do it much less often than every two hours? What about ~5 days or so?

mrice32 (approved):
Code looks great! Only question is around the updates that aren't directly related to an execution need.

src/dataworker/Dataworker.ts
nicholaspai: Ah yes great call there could be multiple

src/dataworker/Dataworker.ts
nicholaspai: https://github.com/across-protocol/contracts-v2/blob/master/contracts/HubPool.sol#L945

nicholaspai: Let's try for 2 days

PR #1311, opened 2024-03-19
fix: Dataworker and BundleDataClient should only use chain ID indices from HubChain bundle start block

nicholaspai:
There  are errors when trying to access block ranges at indices that exist in chain IDs but not in the block ranges list

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -1109,7 +1109,7 @@ export class BundleDataClient {
       (
         await utils.mapAsync(chainIds, async (chainId, index) => {
           const blockRangeForChain = blockRangesForChains[index];
-          if (isChainDisabled(blockRangeForChain)) {
+          if (isChainDisabled(blockRangeForChain) || !blockRangeForChain) {
             return;
           }
           const [_startBlockForChain, _endBlockForChain] = blockRangeForChain;
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -302,9 +302,11 @@ export class BundleDataClient {
       throw new Error("HubPoolClient not updated");
     }
 
-    if (blockRangesForChains.length > this.chainIdListForBundleEvaluationBlockNumbers.length) {
+    const chainIds = this.clients.configStoreClient.getChainIdIndicesForBlock(blockRangesForChains[0][0]);
+
+    if (blockRangesForChains.length > chainIds.length) {
       throw new Error(
-        `Unexpected block range list length of ${blockRangesForChains.length}, should be <= ${this.chainIdListForBundleEvaluationBlockNumbers.length}`
+        `Unexpected block range list length of ${blockRangesForChains.length}, should be <= ${chainIds.length}`
       );
     }
 
@@ -371,7 +373,7 @@ export class BundleDataClient {
         fill,
         this.clients.hubPoolClient,
         blockRangesForChains,
-        this.chainIdListForBundleEvaluationBlockNumbers
+        chainIds
       );
 
       // Fills to refund includes both slow and non-slow fills and they both should increase the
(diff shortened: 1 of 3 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -302,9 +302,11 @@ export class BundleDataClient {
       throw new Error("HubPoolClient not updated");
     }
 
-    if (blockRangesForChains.length > this.chainIdListForBundleEvaluationBlockNumbers.length) {
+    const chainIds = this.clients.configStoreClient.getChainIdIndicesForBlock(blockRangesForChains[0][0]);
+
+    if (blockRangesForChains.length > chainIds.length) {
```
james-a-morris, line 307: Thoughts on sanitizing this against `undefined`?
```suggestion
    if (!isDefined(chainIds) || blockRangesForChains.length > chainIds.length) {
```

src/dataworker/DataworkerUtils.ts
```diff
@@ -100,7 +100,8 @@ export async function blockRangesAreInvalidForSpokeClients(
       blockRanges,
       chainIdListForBundleEvaluationBlockNumbers
     );
-    assert(Object.keys(endBlockTimestamps).length === Object.values(spokePoolClients).filter(isDefined).length);
+    // There should be a spoke pool client instantiated for every bundle timestamp.
+    assert(Object.keys(endBlockTimestamps).every((chainId) => spokePoolClients[chainId] !== undefined));
```
james-a-morris, line 104: Thoughts on switching to a negated `some`? It would potentially halt the loop faster 
```suggestion
    assert(!Object.keys(endBlockTimestamps).some((chainId) => !isDefined(spokePoolClients[chainId])));
```

james-a-morris (commented):
Few first-pass nits. Going to load this in a codespace and review its semantics

src/clients/BundleDataClient.ts
nicholaspai: I think this check is important. At this point we should only be using chain IDs that are valid for this block range. So at this point its only programmer error if we fail this

src/clients/BundleDataClient.ts
```diff
@@ -302,9 +302,11 @@ export class BundleDataClient {
       throw new Error("HubPoolClient not updated");
     }
 
-    if (blockRangesForChains.length > this.chainIdListForBundleEvaluationBlockNumbers.length) {
+    const chainIds = this.clients.configStoreClient.getChainIdIndicesForBlock(blockRangesForChains[0][0]);
```
james-a-morris, line 305: We should probably put a small comment here indicating that `blockRangesForChains[0][0]` is the starting hub chain block.

src/clients/BundleDataClient.ts
```diff
@@ -1109,7 +1079,7 @@ export class BundleDataClient {
       (
         await utils.mapAsync(chainIds, async (chainId, index) => {
           const blockRangeForChain = blockRangesForChains[index];
-          if (isChainDisabled(blockRangeForChain)) {
+          if (!isDefined(blockRangeForChain) || isChainDisabled(blockRangeForChain)) {
```
james-a-morris, line 1082: Will close #1310

mrice32 (approved):
LGTM!

james-a-morris (commented):
Looks good. 

One question: is the intent here to reference the ranges from the start of the bundle regardless of what the user passes as `chainIdListForBundleEvaluationBlockNumbers`? 

I think this makes a lot of sense because, in practice, we instantiate the `BundleDataClient` with the latest ranges from `configStoreClient.getChainIdIndicesForBlock()`.

james-a-morris (approved):
LGTM

src/clients/BundleDataClient.ts
james-a-morris: Closed ^

src/clients/BundleDataClient.ts
```diff
@@ -302,9 +302,11 @@ export class BundleDataClient {
       throw new Error("HubPoolClient not updated");
     }
 
-    if (blockRangesForChains.length > this.chainIdListForBundleEvaluationBlockNumbers.length) {
```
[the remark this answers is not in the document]
nicholaspai: are you talking about the name of the RHS or the LHS?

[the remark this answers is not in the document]
mrice32: LHS

PR #1312, opened 2024-03-19
nit(base-adapter): unneeded conditional check

james-a-morris:
The `some()` function only returns true if a match is found. An empty array calling `.some` is always guaranteed to return a falsy value.

Drive-by comment on my way to getting the Linea Manager

### DIFF
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -276,12 +276,6 @@ export abstract class BaseAdapter {
    */
   isSupportedToken(l1Token: string): l1Token is SupportedL1Token {
     const relevantSymbols = matchTokenSymbol(l1Token, this.hubChainId);
-
-    // If we don't have a symbol for this token, return that the token is not supported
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

src/clients/bridges/BaseAdapter.ts
```diff
@@ -276,12 +276,6 @@ export abstract class BaseAdapter {
    */
   isSupportedToken(l1Token: string): l1Token is SupportedL1Token {
     const relevantSymbols = matchTokenSymbol(l1Token, this.hubChainId);
-
-    // If we don't have a symbol for this token, return that the token is not supported
-    if (relevantSymbols.length === 0) {
-      return false;
-    }
-
```
james-a-morris, line 284: We can squash this logic and assume in the case where `relevantSymbols == []` then `[].some( )` will return false for any value because there's no element in the array that could pass the predicate we provide

pxrl (approved):

PR #1313, opened 2024-03-19
fix(blockRangesAreInvalidForSpokePoolClients): Handle case where bundle range start block is 0

nicholaspai:
This  happens for the first bundle proposed containing a new chain

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -130,7 +130,7 @@ export async function blockRangesAreInvalidForSpokeClients(
     // Note: Math.max the from block with the registration block of the spoke pool to handle the edge case for the first
     // bundle that set its start blocks equal 0.
     const bundleRangeFromBlock = Math.max(spokePoolClient.deploymentBlock, start);
-    if (bundleRangeFromBlock <= latestInvalidBundleStartBlockForChain || end > clientLastBlockQueried) {
+    if (bundleRangeFromBlock < latestInvalidBundleStartBlockForChain || end > clientLastBlockQueried) {
       return true;
     }
 
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -89,7 +89,7 @@ export async function blockRangesAreInvalidForSpokeClients(
   spokePoolClients: Record<number, SpokePoolClient>,
   blockRanges: number[][],
   chainIdListForBundleEvaluationBlockNumbers: number[],
-  latestInvalidBundleStartBlock: { [chainId: number]: number },
+  earliestValidBundleStartBlock: { [chainId: number]: number },
   isV3 = false
 ): Promise<boolean> {
   assert(blockRanges.length === chainIdListForBundleEvaluationBlockNumbers.length);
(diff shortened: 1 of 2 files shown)


src/dataworker/DataworkerUtils.ts
```diff
@@ -122,17 +122,23 @@ export async function blockRangesAreInvalidForSpokeClients(
...
+      latestInvalidBundleStartBlock[chainId] ?? spokePoolClient.deploymentBlock;
+
     // If range start block is less than the earliest spoke pool client we can validate or the range end block
     // is greater than the latest client end block, then ranges are invalid.
-    // Note: Math.max the from block with the deployment block of the spoke pool to handle the edge case for the first
+    // Note: Math.max the from block with the registration block of the spoke pool to handle the edge case for the first
     // bundle that set its start blocks equal 0.
     const bundleRangeFromBlock = Math.max(spokePoolClient.deploymentBlock, start);
-    if (bundleRangeFromBlock <= latestInvalidBundleStartBlock[chainId] || end > clientLastBlockQueried) {
+    if (bundleRangeFromBlock < latestInvalidBundleStartBlockForChain || end > clientLastBlockQueried) {
```
james-a-morris, line 133: Why did we remove the `<=` operator for the `<`?

james-a-morris (commented):
One question

src/dataworker/DataworkerUtils.ts
```diff
@@ -122,17 +122,23 @@ export async function blockRangesAreInvalidForSpokeClients(
```
pxrl, line 122: ooc, should the assignment of `endBlockTimestamps = await getTimestampsForBundleEndBlocks(...)` also be bounded? It looks like it'd resolve to the block timestamp for block 0, which would presumably also produce some undesirable side-effects.

src/dataworker/DataworkerUtils.ts
```diff
@@ -122,17 +122,23 @@ export async function blockRangesAreInvalidForSpokeClients(
 
     const clientLastBlockQueried = spokePoolClient.latestBlockSearched;
 
+    const latestInvalidBundleStartBlockForChain =
+      latestInvalidBundleStartBlock[chainId] ?? spokePoolClient.deploymentBlock;
```
pxrl, line 126: Worth defaulting to `deploymentBlock - 1`, so then all the existing logic still holds OK?

src/dataworker/DataworkerUtils.ts
pxrl: This could be avoided by defaulting `latestInvalidBundleStartBlockForChain` to `spokePoolClient.deploymentBlock - 1` I guess.

nicholaspai: i don't think that's correct though, I should just rename this to earliestValidBundleStartBlock

src/dataworker/DataworkerUtils.ts
nicholaspai: These query end timestamps so we're fine!

pxrl (approved):

james-a-morris (approved):

PR #1314, opened 2024-03-19
feat(linea): add entry for atomic depositor

james-a-morris:
Atomic WETH deposits aren't supported on Linea. We should use the Atomic depositor to unwrap/send WETH to Linea.

An item to note:
1. By only passing the `value` equal to the `amount` we're effectively preventing the AtomicWETHDepositor from allowing automatic executions to Linea. This is how the `LineaChainAdapter` performs its transfer.
2. To allow optional automatic filling, we'd need to be able to pass payable value such that `lineaL1MessageService.send{ value: amount + msg.value }(...)`

### DIFF
--- a/contracts/AtomicWethDepositor.sol
+++ b/contracts/AtomicWethDepositor.sol
@@ -39,7 +39,8 @@ interface LineaL1MessageService {
 
 /**
  * @notice Contract deployed on Ethereum helps relay bots atomically unwrap and bridge WETH over the canonical chain
- * bridges for Optimism, Base, Boba, ZkSync, and Polygon. Needed as these chains only support bridging of ETH, not WETH.
+ * bridges for Optimism, Base, Boba, ZkSync, Linea, and Polygon. Needed as these chains only support bridging of ETH, 
+ * not WETH.
  */
 
 contract AtomicWethDepositor {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

nicholaspai:
> Atomic WETH deposits aren't supported on Linea. We should use the Atomic depositor to unwrap/send WETH to Linea.
> 
> An item to note:
> 
> 1. By only passing the `value` equal to the `amount` we're effectively preventing the AtomicWETHDepositor from allowing automatic executions to Linea. This is how the `LineaChainAdapter` performs its transfer.
> 2. To allow optional automatic filling, we'd need to be able to pass payable value such that `lineaL1MessageService.send{ value: amount + msg.value }(...)`

For posterity sake, why again do we not want to enable automatic executions?

contracts/AtomicWethDepositor.sol
```diff
@@ -69,6 +75,12 @@ contract AtomicWethDepositor {
         polygonL1Bridge.depositEtherFor{ value: amount }(to);
     }
 
+    function bridgeWethToLinea(address to, uint256 amount) public payable {
+        weth.transferFrom(msg.sender, address(this), amount);
+        weth.withdraw(amount);
+        lineaL1MessageService.sendMessage{ value: amount }(to, 0, "");
```
dohaki, line 81: Maybe it makes sense to change this to something like
```suggestion
        lineaL1MessageService.sendMessage{ value: amount + msg.value }(to, msg.value, "");
```
Where `msg.value` is the fee we can optionally set for auto-claim this L1->L2 transfer of ETH. This would give us the flexibility to change our approach in the inventory manager / finalizer bots. Wdyt?

dohaki (approved):

nicholaspai (approved):

PR #1315, opened 2024-03-20
feat: sampled RPC logging

mrice32:
This PR adds a small change to sample RPC requests/responses to be sampled and logged by the provider (including latency). The sample rate is configurable s.t. an operator can produce more or fewer logs depending on their interest (or disable). The default is set to 0.1%.

My hope is that this will enable operators to pipe their logs to dashboarding tools to visualize RPC performance.

Open question: should the sample rate be set to 0 by default to keep things simple for the typical operator who may not care about these metrics?


nicholaspai (approved):
I like defaulting to 0 and making this configurable via the env

src/utils/ProviderUtils.ts
```diff
@@ -59,6 +60,45 @@ class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider {
...
+          success: false,
+          timeElapsed: elapsedTimeS,
+        });
+        throw error;
+      }
+    } else {
+      // Non sample path: no logging or timing, just issue the request.
+      return super.send(method, params);
+    }
+  }
```
pxrl, line 100: Minor style nit, but consider moving the likely case (non-sample path) to the top? Then it can be guarded by the if statement and the rest of the code can be dropped out of it. It'd make it easier to read IMO.

pxrl, line 100: Also a minor suggestion to factor out the common parts of the call to `logger.debug()`.

```suggestion
    if (this.pctRpcCallsLogged > 0 && Math.random() <= this.pctRpcCallsLogged / 100) {
      const loggerArgs = {
          at: "ProviderUtils",
          message: "Provider response sample",
          provider: getOriginFromURL(this.connection.url),
          method,
          params,
          success: true,      
      };
      // In this path we log an rpc response sample.
      // Note: use performance.now() to ensure a purely monotonic clock.
      const startTime = performance.now();
      try {
        const result = await super.send(method, params);
        const elapsedTimeS = (performance.now() - startTime) / 1000;
        logger.debug({ ...loggerArgs, result, timeElapsed: elapsedTimeS });
        });
      } catch (error) {
        // Log errors as well.
        // For now, to keep logs light, don't log the error itself, just propogate and let it be handled higher up.
        const elapsedTimeS = (performance.now() - startTime) / 1000;
        logger.debug({ ...loggerArgs, success: false, timeElapsed: elapsedTimeS });
        throw error;
      }
    } else {
      // Non sample path: no logging or timing, just issue the request.
      return super.send(method, params);
    }
  }
```

pxrl (approved):
Great idea!

src/utils/ProviderUtils.ts

pxrl (approved):

mrice32:
> I like defaulting to 0 and making this configurable via the env

Done

PR #1319, opened 2024-03-20
feat(linea): Adapter Manager Client

james-a-morris:
This PR adds a mostly completed Linea Adapter Manager. Namely - the only function that has been specifically made a no-op is `getOutstandingCrossChainTransfers` but a `// FIXME:` has been added and I will follow up with it before EOD.


nicholaspai (approved):

dohaki (approved):
LGTM

PR #1320, opened 2024-03-20
fix(linea): use default contract addresses l1->l2

dohaki:
We should not override the addresses because we can just use the defaults

### DIFF
--- a/src/finalizer/utils/linea/l1ToL2.ts
+++ b/src/finalizer/utils/linea/l1ToL2.ts
@@ -5,7 +5,7 @@ import { utils, providers } from "ethers";
 import { groupBy } from "lodash";
 
 import { HubPoolClient } from "../../../clients";
-import { CHAIN_MAX_BLOCK_LOOKBACK, CONTRACT_ADDRESSES } from "../../../common";
+import { CHAIN_MAX_BLOCK_LOOKBACK } from "../../../common";
 import { Signer, winston, convertFromWei, TransactionReceipt, paginatedEventQuery } from "../../../utils";
 import { FinalizerPromise, CrossChainMessage } from "../../types";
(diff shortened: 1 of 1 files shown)


src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -29,8 +29,8 @@ export async function lineaL1ToL2Finalizer(
   const [l1ChainId, hubPoolAddress] = [hubPoolClient.chainId, hubPoolClient.hubPool.address];
   const l2ChainId = l1ChainId === 1 ? 59144 : 59140;
   const lineaSdk = initLineaSdk(l1ChainId, l2ChainId);
-  const l2Contract = lineaSdk.getL2Contract(CONTRACT_ADDRESSES[l2ChainId]?.lineaMessageService.address);
-  const l1Contract = lineaSdk.getL1Contract(CONTRACT_ADDRESSES[l1ChainId]?.lineaMessageService.address);
+  const l2Contract = lineaSdk.getL2Contract();
+  const l1Contract = lineaSdk.getL1Contract();
```
james-a-morris, line 33: OOC - did we need to use the SDK? I know for ZKSync we avoided the SDK where possible.

james-a-morris (approved):

src/finalizer/utils/linea/l1ToL2.ts
dohaki: It has some nice utilities for the claiming flow and retrieving the message status. Because for some messages you need to provide a proof and for others you don't.

dohaki: So yea I would prefer to keep the SDK for now

PR #1321, opened 2024-03-20
fix: Relayer should only add chains to CrossChainTransferClient that are supported in AdapterManager

nicholaspai:
This allows relayer to ignore chains that haven't implemented the full inventory manager function set

This is not a bug seen in prod because RELAYER_DESTINATION_CHAINS is set to not include linea


pxrl (approved):
Looks sensible :+1:

PR #1322, opened 2024-03-20
improve(TokenClient): Cache allowance and bondToken eth_call data

nicholaspai:
These values should rarely change. Follows strategy implemented in #1282 in the BaseAdapter (inventory manager)


src/clients/TokenClient.ts
```diff
@@ -180,11 +185,31 @@ export class TokenClient {
...
+      if (cachedBondToken !== null) {
+        bondToken = cachedBondToken;
+      }
+    }
+    if (!bondToken) {
+      [balanceInfo, bondToken] = await Promise.all([
+        Promise.all(
+          Object.values(this.spokePoolClients).map((spokePoolClient) => this.fetchTokenData(spokePoolClient))
+        ),
+        this.hubPoolClient.hubPool.bondToken(),
```
pxrl, line 202: We could leverage the existing RPC caching for this by just querying at a hardcoded block number (or something rounded to a very large number). Then we wouldn't need to have any extra handling for caching the bond token. Might be worth considering.

src/clients/TokenClient.ts
```diff
@@ -210,10 +235,20 @@ export class TokenClient {
...
+    }_${originToken}_${await spokePoolClient.spokePool.signer.getAddress()}_targetContract:${
+      spokePoolClient.spokePool.address
+    }`;
+  }
+
+  getBondTokenCacheKey(): string {
+    return `bondToken_${this.hubPoolClient.hubPool.address}`;
+  }
+
+  async fetchTokenData(spokePoolClient: SpokePoolClient): Promise<FetchTokenDataReturnType> {
```
pxrl, line 250: As above, if the caller can supply a blockTag to this function then we could just pass in a blockTag that's rounded such that it only rotates every x days/weeks/months. In that way we should get implicit RPC caching without needing to introduce redis here.

src/clients/TokenClient.ts
nicholaspai: is it safe to just hardcode a single block number?

pxrl: Caching with infinte TTL is basically the same as hardcoding the block number, but there's the issue of a hardcoded block numer not being portable to testnet...

So perhaps it could be done by rounding the block tag to the nearest _nth_ block? i.e. something like `Math.floor(hubPoolDeploymentBlock, Math.round((hubPoolLatestBlock / 200k) * 200k)`, or something along those lines? That'd give ~4 weeks of caching, but it could be scaled up as necessary.

It just seems like a convenient way of avoiding talking directly to the cache here, which IMO tends to dominate the code a bit.

nicholaspai: Hm this might set the wrong bondToken into the cache forever if set incorrectly but ok this works well for bondToken

src/clients/TokenClient.ts
pxrl: Per discussion, maybe this strategy doesn't work as well for allowances because new tokens are being added on occasion, so it'd need to be refreshed a lot more often.

src/clients/TokenClient.ts
```diff
@@ -183,7 +188,14 @@ export class TokenClient {
 
     const [balanceInfo, bondToken] = await Promise.all([
       Promise.all(Object.values(this.spokePoolClients).map((spokePoolClient) => this.fetchTokenData(spokePoolClient))),
-      this.hubPoolClient.hubPool.bondToken(),
+      // Make eth_call using nearest 100_000th block to take advantage of eth_call caching when blockTag is
+      // specified. This is unexpected to change so we are ok to rarely increase this block tag.
+      this.hubPoolClient.hubPool.bondToken({
+        blockTag: Math.min(
```
pxrl, line 194: I think this needs to be `Math.max()` to avoid always resolving to deploymentBlock.

```suggestion
        blockTag: Math.max(
```

pxrl (approved):
LGTM!

nicholaspai:
While fixing the block tag to the nearest 100_000th block when querying bondToken is nice, it breaks tests and I don't think they're worth fixing. The main motivation for taking advantage of the eth_call specific-blockTag caching is to make the code cleaner but I think this PR's refactor is now minimally invasive

@pxrl

src/clients/TokenClient.ts
```diff
@@ -235,6 +233,56 @@ export class TokenClient {
...
+        return toBN(result);
+      }
+    }
+    const allowance: BigNumber = await token.allowance(this.relayerAddress, spokePoolClient.spokePool.address, {
+      blockTag,
+    });
+    if (redis) {
+      // Save allowance in cache with no TTL as these should never decrement.
+      await redis.set(await this._getAllowanceCacheKey(spokePoolClient, token.address), MAX_SAFE_ALLOWANCE);
+    }
```
mrice32, line 262: Should we check that the allowance is > MAX_SAFE_ALLOWANCE before setting the cache to that value??

src/clients/TokenClient.ts
```diff
@@ -235,6 +233,56 @@ export class TokenClient {
...
+    if (redis) {
+      const cachedBondToken = await redis.get<string>(this._getBondTokenCacheKey());
+      if (cachedBondToken !== null) {
+        return cachedBondToken;
+      }
+    }
+    const bondToken: string = await this.hubPoolClient.hubPool.bondToken();
+    if (redis) {
+      // Save allowance in cache with no TTL as this should never change.
+      await redis.set(this._getBondTokenCacheKey(), bondToken);
```
mrice32, line 281: We should remember to change this cache key if we ever do update the bond token (or other permanently cached values)

mrice32 (commented):
Looks great, just one comment!

src/clients/TokenClient.ts
```diff
@@ -235,6 +233,56 @@ export class TokenClient {
...
+      const result = await redis.get<string>(await this._getAllowanceCacheKey(spokePoolClient, token.address));
+      if (result !== null) {
+        return toBN(result);
+      }
+    }
+    const allowance: BigNumber = await token.allowance(this.relayerAddress, spokePoolClient.spokePool.address, {
+      blockTag,
+    });
+    if (redis) {
+      // Save allowance in cache with no TTL as these should never decrement.
```
pxrl, line 260:
```suggestion
      // Save allowance in cache with no TTL as these should never be exhausted.
```

src/clients/TokenClient.ts
```diff
@@ -235,6 +233,56 @@ export class TokenClient {
...
+    const redis = await getRedisCache(this.logger);
+    if (redis) {
+      const cachedBondToken = await redis.get<string>(this._getBondTokenCacheKey());
+      if (cachedBondToken !== null) {
+        return cachedBondToken;
+      }
+    }
+    const bondToken: string = await this.hubPoolClient.hubPool.bondToken();
+    if (redis) {
+      // Save allowance in cache with no TTL as this should never change.
```
pxrl, line 280:
```suggestion
      // The bond token should not change, and using the wrong bond token will be immediately obvious, so cache with
      // infinite TTL.
```

src/clients/TokenClient.ts
nicholaspai: ah of course

PR #1328, opened 2024-03-20
feat(finalizer): make l1l2 routes generic

james-a-morris:
This PR aims to make the L1->L2 route generic. Adding this PR includes an optional (non-breaking) ENV variable called `L1_L2_FINALIZER_MONITOR_ADDRESS`. 

This will ideally be generalized to CCTP, but for now it's just available on the L1 -> L2 linea direction. Any address included in this search will have its L1 -> L2 messages claimed by the finalizer.

--
Additional changes:
1. Bug Fix: L1 -> L2 for Linea was not being called. This is because we included the entry in the `chainFinalizerOverrides` in slot `[1]` but since there is no corresponding `chainFinalizers` for chain `[1]` the function would never be called.
2. I've cleaned up the imports - ordering them and collapsing multiple imports where possible.


src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -40,6 +43,12 @@ export async function lineaL1ToL2Finalizer(
   const l1Contract = lineaSdk.getL1Contract();
   const getMessagesWithStatusByTxHash = makeGetMessagesWithStatusByTxHash(l1Contract, l2Contract);
 
+  // We always want to make sure that the l1ToL2AddressesToFinalize array contains
+  // the HubPool address, so we can finalize any pending messages sent from the HubPool.
+  if (!l1ToL2AddressesToFinalize.includes(getAddress(hubPoolAddress))) {
```
nicholaspai, line 48: I don't think we want every finalizer instance to finalize these messages. For example, we might want to purposefully ignore these in the faster, inventory-manager finalizer, so that the slower finalizer can include them in its batch to save on gas.

We only want the fast finalizer to finalize time sensitive inventory rebalances

src/finalizer/index.ts
```diff
@@ -59,16 +60,14 @@ const chainFinalizers: { [chainId: number]: ChainFinalizer } = {
  */
 const chainFinalizerOverrides: { [chainId: number]: ChainFinalizer[] } = {
   // Mainnets
-  1: [lineaL1ToL2Finalizer],
```
nicholaspai, line 62: This was working before if we included 1 in the FINALIZER_CHAINS list

src/finalizer/index.ts
```diff
@@ -30,12 +31,12 @@ import {
...
+  lineaL1ToL2Finalizer,
+  lineaL2ToL1Finalizer,
   opStackFinalizer,
   polygonFinalizer,
   scrollFinalizer,
   zkSyncFinalizer,
-  lineaL2ToL1Finalizer,
-  lineaL1ToL2Finalizer,
 } from "./utils";
 const { isDefined } = sdkUtils;
```
nicholaspai, line 42: nit unrelated: should we add an entry for 59144: lineaFinalizer below in chainFinalizers?

nicholaspai (commented):
https://github.com/across-protocol/relayer-v2/pull/1328/files#r1532910340

This looks good but see the above comment

Also,

src/finalizer/utils/linea/l1ToL2.ts
james-a-morris: The `l1ToL2AddressesToFinalize` is optional. If it is undefined in the ENV, it defaults to an empty array and we only finalize for the hub pool.

src/finalizer/index.ts
james-a-morris: Reverted

src/finalizer/index.ts
james-a-morris: We can. It shouldn't change anything functionally but it's probably the better place to include it. https://github.com/across-protocol/relayer-v2/pull/1328/commits/a84f55e302101cd0abfe4d396fdf5af3e7cdc7ee

src/finalizer/utils/linea/l1ToL2.ts
nicholaspai: Actually, scratch my comment. I think we always want l1 to l2 messages finalized as quickly as possible for Linea. This way we send new root bundles after executed PoolRebalanceLeaves to the Spokes ASAP. So lets leave as is

src/finalizer/index.ts
```diff
@@ -64,7 +66,6 @@ const chainFinalizerOverrides: { [chainId: number]: ChainFinalizer[] } = {
   137: [polygonFinalizer, cctpL1toL2Finalizer, cctpL2toL1Finalizer],
   8453: [opStackFinalizer, cctpL1toL2Finalizer, cctpL2toL1Finalizer],
   42161: [arbitrumOneFinalizer, cctpL1toL2Finalizer, cctpL2toL1Finalizer],
-  59144: [lineaL2ToL1Finalizer],
```
nicholaspai, line 67: why remove this?

nicholaspai (approved):

src/finalizer/index.ts
```diff
@@ -353,9 +367,10 @@ export async function runFinalizer(_logger: winston.Logger, baseSigner: Signer):
           commonClients.hubSigner,
           commonClients.hubPoolClient,
           spokePoolClients,
-          process.env.FINALIZER_CHAINS
-            ? JSON.parse(process.env.FINALIZER_CHAINS)
-            : commonClients.configStoreClient.getChainIdIndicesForBlock(),
+          config.chainsToFinalize.length === 0
+            ? commonClients.configStoreClient.getChainIdIndicesForBlock()
```
pxrl: This fails on Boba when `FINALIZER_CHAINS` is not defined.

pxrl: Updated to filter the default set of chain IDs: 4d6cba40dc145da734b523c771fec1a33d2bde6e

pxrl (commented):
I've been able to run this locally and it completes its run OK. I didn't see any finalisations though.

src/finalizer/utils/linea/common.ts
```diff
@@ -107,3 +115,156 @@ export async function getBlockRangeByHoursOffsets(
...
+      l1TokenAddress: string;
+      amount: BigNumber;
+    }
+  | {
+      type: "misc";
+    } {
+  const { _calldata, _value } = event.args;
+  // First check a WETH deposit. A WETH deposit is a message with a positive
+  // value and an empty calldata.
+  if (_calldata === "0x") {
```
nicholaspai, line 135: Should we check if value is positive, per the comment?

nicholaspai (approved):
Have you ran finalizer locally? It should show that there are some already claimed l1 to l2 messages

src/finalizer/index.ts
james-a-morris: It doesn't need to be overridden if we include it in the finalizers above ^

src/finalizer/utils/linea/common.ts
```diff
@@ -107,3 +115,154 @@ export async function getBlockRangeByHoursOffsets(
...
+}
+
+export async function findMessageSentEvents(
+  contract: L1MessageServiceContract | L2MessageServiceContract,
+  l1ToL2AddressesToFinalize: string[],
+  searchConfig: EventSearchConfig
+): Promise<MessageSentEvent[]> {
+  return paginatedEventQuery(
+    contract.contract,
+    (contract.contract as Contract).filters.MessageSent(l1ToL2AddressesToFinalize),
```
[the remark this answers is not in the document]
james-a-morris: `contract.contract` is a typed `ethers.Contract`. The issue is that `MessageSent(...)` doesn't allow array entries for parameters (which perform a logical OR). The typecast is, unfortunately, required to allow this functionality.

src/finalizer/utils/linea/common.ts
```diff
@@ -107,3 +115,154 @@ export async function getBlockRangeByHoursOffsets(
...
+        }))
+        .filter((log) => {
+          // Start with the TokenBridge calldata format.
+          try {
+            const decoded = bridgeContract.interface.decodeFunctionData("completeBridging", log.args._calldata);
+            return (
+              compareAddressesSimple(decoded._recipient, event.args.recipient) && decoded._amount.eq(event.args.amount)
+            );
+          } catch (_e) {
+            // We don't care about this because we have more to check
```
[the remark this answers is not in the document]
james-a-morris: We will theoretically have a multitude of `MessageSent` events - we only want to filter ones that both decode to a corresponding bridge & that match the corresponding parameters

james-a-morris:
> Have you ran finalizer locally? It should show that there are some already claimed l1 to l2 messages

Yes - I've finalized each type of txn https://lineascan.build/tx/0xe12be0fe4875b7327d2d27dca31e529ff0c99f278c3bf77e397f2dea8feb1f7f

src/finalizer/utils/linea/common.ts
nicholaspai: but shouldn't you also check that _value is positive?

james-a-morris: That's a fair check - I don't imagine that an empty call with an empty _value should be available, but this is a fair filter to add

src/finalizer/types.ts
```diff
@@ -36,6 +37,20 @@ export interface ChainFinalizer {
     logger: winston.Logger,
     signer: Signer,
     hubPoolClient: HubPoolClient,
-    spokePoolClient: SpokePoolClient
+    spokePoolClient: SpokePoolClient,
+    l1ToL2AddressesToFinalize: string[]
   ): Promise<FinalizerPromise>;
 }
+
+export interface EnrichedLineaSentMessageEvent {
```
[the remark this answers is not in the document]
james-a-morris: This is a holdover - it has been removed: https://github.com/across-protocol/relayer-v2/pull/1328/commits/3cd06adcb2afe35a9452a78a457eb148db3dfdcc

PR #1329, opened 2024-03-20
refactor(relayer): Defer deposit MDCs check

pxrl:
Rather than duplicating the array of unfilled deposits, just check it as the first conditional in the subsequent loop. This allows the entire function to be subsequently factored away, which will occur in a follow-up commit.


nicholaspai (approved):
Does this result in  a speed up by avoiding an extra loop?

pxrl:
> Does this result in a speed up by avoiding an extra loop?

Probably, but I don't think it's noticeable vs. the other delays we have in this workflow. This change is still just rearranging furniture for a more impactful change that's coming.

PR #1330, opened 2024-03-20
fix(LineaFinalizer): Filter out TokensRelayed events that don't have target = LineaSpoke

nicholaspai:


mrice32 (approved):

pxrl (approved):
This checks out against the ABI you linked :+1:

PR #1333, opened 2024-03-21
fix(relayer): Post-process external inventory config

pxrl:
The recent change to support externally-defined inventory configs neglected to post-process the imported config. It's important to scale each configured amount before taking the config into use.

### DIFF
--- a/src/relayer/RelayerClientHelper.ts
+++ b/src/relayer/RelayerClientHelper.ts
@@ -124,6 +124,7 @@ export async function constructRelayerClients(
       const msg = typeguards.isError(err) ? err.message : (err as Record<string, unknown>)?.code;
       throw new Error(`Inventory config error in ${config.externalInventoryConfig} (${msg ?? "unknown error"})`);
     }
+    config.parseInventoryConfig();
     logger.debug({
       at: "Relayer#constructRelayerClients",
       message: "Updated Inventory config.",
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

src/relayer/RelayerClientHelper.ts
```diff
@@ -124,6 +124,7 @@ export async function constructRelayerClients(
       const msg = typeguards.isError(err) ? err.message : (err as Record<string, unknown>)?.code;
       throw new Error(`Inventory config error in ${config.externalInventoryConfig} (${msg ?? "unknown error"})`);
     }
+    config.parseInventoryConfig();
```
james-a-morris, line 127: Would it make sense to wrap this log into the relay config? I.e. we could check this information in the constructor where we do all other ENV parsing. We can use `readFileSync` to avoid the async call

james-a-morris (commented):
One question on placing the inventory config lookup in the helper rather than in the config class itself. Even an async function to load the `externalInventoryConfigURI` could be a good step

src/relayer/RelayerClientHelper.ts
pxrl: Yeah - the async call is why it didn't go in the RelayerConfig class, but I'd forgotten about `readFileSync()`. That's a much cleaner solution. Will update!

pxrl: @james-a-morris Updated: https://github.com/across-protocol/relayer-v2/pull/1333/commits/ad98a4169bd1760d03853795049138681fa35933

It's much more streamlined now and generally feels like a much cleaner implementation. wdyt?

james-a-morris (approved):
LGTM!

PR #1336, opened 2024-03-21
feat: Cache limits for 5 mins from Across API

nicholaspai:
Lots of relayers running at same time is stressful on API requests

### DIFF
--- a/src/clients/AcrossAPIClient.ts
+++ b/src/clients/AcrossAPIClient.ts
@@ -1,7 +1,16 @@
 import _ from "lodash";
 import axios, { AxiosError } from "axios";
 import { SpokePoolClientsByChain } from "../interfaces";
-import { bnZero, isDefined, winston, BigNumber, getL2TokenAddresses, CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "../utils";
+import {
+  bnZero,
+  isDefined,
+  winston,
+  BigNumber,
+  getL2TokenAddresses,
+  CHAIN_IDs,
+  TOKEN_SYMBOLS_MAP,
+  getRedisCache,
+} from "../utils";
 import { HubPoolClient } from "./HubPoolClient";
 
 export interface DepositLimits {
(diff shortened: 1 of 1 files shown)


src/clients/AcrossAPIClient.ts
```diff
@@ -116,10 +125,24 @@ export class AcrossApiClient {
     const path = "limits";
     const url = `${this.endpoint}/${path}`;
 
+    const redis = await getRedisCache();
     for (const destinationChainId of destinationChainIds) {
       const params = { token: l1Token, destinationChainId, originChainId: 1 };
+      if (redis) {
+        const cachedLimits = await redis.get<string>(`limits_api_${l1Token}_${destinationChainId}`);
+        if (cachedLimits !== null && BigNumber.isBigNumber(cachedLimits)) {
```
mrice32, line 133: won't cachedLimits be a string? So wouldn't isBigNumber return false?

src/clients/AcrossAPIClient.ts
```diff
@@ -116,10 +125,24 @@ export class AcrossApiClient {
...
+          return { maxDeposit: BigNumber.from(cachedLimits) };
+        }
+      }
       try {
         const result = await axios(url, { timeout, params });
+        if (!result?.data?.maxDeposit) {
+          throw new Error("Invalid response from /limits, expected maxDeposit field.");
+        }
```
mrice32, line 141: Two issues:
1. This will get caught by the catch to construct a warning, but that catch assumes it's an axios error, which isn't true here.
2. The catch will mean this turns into a warning. Do we want a warning? Or do we want this to propagate up as a true error?

mrice32 (commented):
Two comments!

src/clients/AcrossAPIClient.ts
nicholaspai: Good call, ill add a separate try-catch for the redis operation

src/clients/AcrossAPIClient.ts
nicholaspai: Good point added an error log here and better catching https://github.com/across-protocol/relayer-v2/pull/1336/commits/054894eec7cc9641651e079d3674cb5349948b58

pxrl (approved):
LGTM - will be nice to reduce some noise in the logs.

PR #1338, opened 2024-03-21
improve(relayer): Batch query deposit fill status

pxrl:
The relayer currently implements late-stage checking of the fill status on a per-deposit basis. This is fairly slow and inefficient, but does successfully protect against an avalanche of failed fill simulations that result from overlapping bot runs.

This change implements batched querying of all deposits for each destination chain. This is significantly more efficient in terms of RPC consumption, and should allow for subsequent "already filled" filtering to occur much faster. It's also a key enabler for some subsequent efficiency improvements.


nicholaspai (approved):
This is a nice idea and implementation. It's ok to query this for each fill since we only do it once per batch, unlike currently where we only check it for unfilled fills but sequentially

src/utils/FillUtils.ts
```diff
@@ -291,34 +293,41 @@ export async function getUnfilledDeposits(
...
+          return originClient
+            .getDepositsForDestinationChain(destinationChainId)
+            .filter((deposit) => deposit.blockNumber >= earliestBlockNumber)
+            .filter(sdkUtils.isV3Deposit<V3DepositWithBlock, V2DepositWithBlock>); // @todo: Remove after v2 deprecated.
+        })
+        .flat();
+
+      // Resolve the latest fill status for each deposit and filter out any that are now filled.
+      const { spokePool } = destinationClient;
+      const fillStatus = await sdkUtils.fillStatusArray(spokePool, deposits);
```
pxrl, line 322: nb. I'm wary that we might pump some huge number of deposits into `fillStatusArray()`, so I will update the `fillStatusArray()` to chunk the requests and submit them in parallel. That gives the best abstraction for the user of the function.

pxrl: Implemented and included as part of this PR: https://github.com/across-protocol/sdk-v2/commit/8315bf4e6ac02ad435f75411ad060501806b1e89.

PR #1339, opened 2024-03-21
feat(relayer): Support ignoring FilledV3Relay events

pxrl:
This commit makes the SpokePoolClient FilledV3Relay query optional, subject to the ACCEPT_INVALID_FILLS env var being defined true. This ensures that the behaviour is opt-in.

This is enabled by the recent change to query the fill status of all outstanding deposits. Since there's no need to submit the LP fee onchain in Across v3, the risks of opting into this setting are significantly reduced, whilst the benefits are:
- Reduced SpokePoolClient update times.
- Reduced overheads in matchind deposits with fills.


nicholaspai (approved):
Nice!

mrice32 (approved):
LGTM!

mrice32 (commented):
Mentioned IRL, but repeating here: will this cause any issues with the relayer determining what refunds it expects in the next bundle?

pxrl:
> Mentioned IRL, but repeating here: will this cause any issues with the relayer determining what refunds it expects in the next bundle?

Really good consideration - it would.

It's a useful feature for third-parties to have so I think it's probably worth having, but I wonder if it should also be conditional on inventory management being disabled. I've marked it `do not merge` for the time being and will think over it for a little while.

nicholaspai:
We can probably re review this again after changing how bundle refunds are computed in the InventoryManager. If we can fetch them from Arweave then this PR can be useful even with InventoryManager enabled

nicholaspai (changes requested):
do not merge yet until bundle data can be fetched without using FilledV3Relay events

nicholaspai (approved):
I think we can evaluate this after local testing now that we load most bundle data from arweave. However, there is still a case where we'd want to comute bundles from events when there is no bundle pending. This happens in the short window (~15 mins max) between last bundle execution and next bundle proposed.

This PR would cause the Inventory Manager to not count any bundle refunds in that time, which probably is not a big deal and worth the speed up if it is signficiant

nicholaspai:
> I think we can evaluate this after local testing now that we load most bundle data from arweave. However, there is still a case where we'd want to comute bundles from events when there is no bundle pending. This happens in the short window (~15 mins max) between last bundle execution and next bundle proposed.
> 
> 
> 
> This PR would cause the Inventory Manager to not count any bundle refunds in that time, which probably is not a big deal and worth the speed up if it is signficiant

I think this pr is now intractable if we're going to go with the bundle refund approximation approach in #1382

pxrl:
> > I think we can evaluate this after local testing now that we load most bundle data from arweave. However, there is still a case where we'd want to comute bundles from events when there is no bundle pending. This happens in the short window (~15 mins max) between last bundle execution and next bundle proposed.
> > This PR would cause the Inventory Manager to not count any bundle refunds in that time, which probably is not a big deal and worth the speed up if it is signficiant
> 
> I think this pr is now intractable if we're going to go with the bundle refund approximation approach in #1382

Yep, agreed. Time for it to be closed.

PR #1340, opened 2024-03-21
improve(relayer): Drop slow fill request queries

pxrl:
By retaining the deposit FillStatus on the destination chain it's possible to skip scraping for RequestedV3SpeedUp events. This reduces the RPC load during SpokePoolClient update.


src/relayer/Relayer.ts
```diff
@@ -262,17 +267,13 @@ export class Relayer {
...
-        this.logger.debug({
-          at: "Relayer",
-          message: "Initiating slow fill for grey listed depositor",
-          depositor,
-        });
-        this.requestSlowFill(deposit);
-      }
-      // Regardless of whether we should send a slow fill or not for this depositor, exit early at this point
-      // so we don't fast fill an already slow filled deposit from the slow fill-only list.
+    if (slowDepositors?.includes(depositor) && fillStatus === FillStatus.Unfilled) {
```
pxrl, line 270: Opinionated driveby change: If the operator has configured `SLOW_DEPOSITORS`, then don't require `SEND_SLOW_RELAYS` to be enabled.

src/relayer/Relayer.ts
```diff
@@ -262,17 +267,13 @@ export class Relayer {
     }
 
     // If depositor is on the slow deposit list, then send a zero fill to initiate a slow relay and return early.
-    if (slowDepositors?.includes(depositor)) {
-      if (sendSlowRelays) {
```
pxrl, line 266: Opinionated driveby change: If the operator has configured SLOW_DEPOSITORS, then don't require SEND_SLOW_RELAYS to be enabled.

nicholaspai: why not? not sure I agree

nicholaspai (approved):
This also seems good

james-a-morris (commented):
Although I agree that a slow depositor should trigger a slow relay every time, I'm not sure we want to remove the configuration to be able to disable slow relays from a given relayer.

pxrl:
> Although I agree that a slow depositor should trigger a slow relay every time, I'm not sure we want to remove the configuration to be able to disable slow relays from a given relayer.

Agree - that's not changed here and it remains as-is :+1: Slow relays is still 100% opt-in.

src/relayer/Relayer.ts
pxrl: The current configuration gives no way to opt-in to slow fill specific addresses and to ignore all others. Meanwhile, if the operator has explicitly defined an array of addresses to fill with slow relays, I see that as an implicit opt-in to slow relays for those addresses, and it shouldn't be required to separately enable slow fills for all other addresses.

This doesn't mean that the relayer would make slow fills for all other deposits - only for addresses identified in `slowDepositors`. The existing `sendSlowRelays` conditional still applies to everything else.

nicholaspai (commented):
I think this will also cause bundle refund construction problematic since we need slow fill requests to compute bundle data correctly

nicholaspai (changes requested):
do not merge yet until bundle data can be fetched without using SlowFillRequest events

pxrl:
> I think this will also cause bundle refund construction problematic since we need slow fill requests to compute bundle data correctly

That's important for the proposer/disputer, but is relevant for the relayer? This PR only affects the relayer's SpokePoolClient update, and deposit update requests don't affect relayer repayments so I can't see any obvious problem with ignoring them.

nicholaspai (approved):
I think we can revisit this now that we load most bundle data from arweave. I also think this probably wasn't an issue affecting bundle refund computation

pxrl:
> I think we can revisit this now that we load most bundle data from arweave. I also think this probably wasn't an issue affecting bundle refund computation

Yeah - I also wasn't able to reason my way to any problem with dropping these events. I'll merge this on Monday :+1:

PR #1341, opened 2024-03-21
fix(relayer): Remove redundant array manipulation

pxrl:
This code was recently touched; it's an error without any apparent side-effects.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -258,9 +258,7 @@ export class Relayer {
     // Fetch unfilled deposits and filter out deposits upfront before we compute the minimum deposit confirmation
     // per chain, which is based on the deposit volume we could fill.
     const unfilledDeposits = await this._getUnfilledDeposits();
-    const allUnfilledDeposits = Object.values(unfilledDeposits)
-      .flat()
-      .map(({ deposit }) => deposit);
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

src/relayer/Relayer.ts
```diff
@@ -258,7 +258,9 @@ export class Relayer {
     // Fetch unfilled deposits and filter out deposits upfront before we compute the minimum deposit confirmation
     // per chain, which is based on the deposit volume we could fill.
     const unfilledDeposits = await this._getUnfilledDeposits();
-    const allUnfilledDeposits = Object.values(unfilledDeposits.map(({ deposit }) => deposit));
+    const allUnfilledDeposits = Object.values(unfilledDeposits)
+      .flat()
```
james-a-morris, line 262: I guess the destructuring of { deposit } left `deposit === undefined`? Good find

james-a-morris (approved):

src/relayer/Relayer.ts
pxrl: It wasn't as bad as I first thought - `unfilledDeposits` is already a flat array, so the extra `Object.values()` was just redundant code.

pxrl:
Closing this because it's only a superficial issue and will be cleaned up in a subsequent commit that already touches this code.

PR #1346, opened 2024-03-22
improve(relayer): Parallel processing for destination chain fills

pxrl:
Relayer fills are currently processed as a flat array. When messages are involved this requires simulation of the fill on the destination chain, thus incurring delay. In these cases, take the opportunity to process other fills whilst awaiting the outcome of the simulation. This is also another step towards containing RPC misbehaviour to the affected chain and preventing contagion to other destinations.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { utils as ethersUtils } from "ethers";
 import { L1Token, V3Deposit, V3DepositWithBlock } from "../interfaces";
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
LGTM

src/relayer/Relayer.ts
```diff
@@ -36,141 +37,154 @@ export class Relayer {
...
-          at: "Relayer::getUnfilledDeposits",
-          message: "Skipping deposit from or to disabled chains.",
-          deposit,
-          enabledOriginChains: this.config.relayerOriginChains,
-          enabledDestinationChains: this.config.relayerDestinationChains,
-        });
-        return false;
-      }
+    // If we don't have the latest code to support this deposit, skip it.
+    if (version > configStoreClient.configStoreVersion) {
```
james-a-morris, line 52:
```suggestion
    if (configStoreClient.configStoreVersion < version) {
```

Nit: could we flip the comparison? They're functionally the same but IMO this is easier to read

src/relayer/Relayer.ts
```diff
@@ -36,141 +37,154 @@ export class Relayer {
...
+    }
 
-      // It would be preferable to use host time since it's more reliably up-to-date, but this creates issues in test.
-      const currentTime = this.clients.spokePoolClients[destinationChainId].getCurrentTime();
-      if (deposit.fillDeadline <= currentTime) {
-        return false;
-      }
+    // Skip any L1 tokens that are not specified in the config.
+    // If relayerTokens is an empty list, we'll assume that all tokens are supported.
+    const l1Token = hubPoolClient.getL1TokenInfoForL2Token(inputToken, originChainId);
```
james-a-morris, line 91: Nit: thoughts on putting this within a check for `relayerTokens.length > 0` first so that we don't have to call it if `relayerTokens.length === 0`?

james-a-morris (approved):
Two nits

src/relayer/Relayer.ts
pxrl: Good idea, but we use `l1Token` later on in some log messages, so it needs to stay available in the broader scope.

src/relayer/Relayer.ts
pxrl: Resolved out of band: https://github.com/across-protocol/relayer-v2/pull/1346/commits/451b67f15b4982c0f54afdf42d5525d65e9eeed4-

PR #1347, opened 2024-03-22
fix(Dataworker): Send msgValue to pay for Linea leaf executions IFF amountToReturn > 0

nicholaspai:
Currently we are sending the fee for every single Linea leaf execution. This ultimately means we are donating the fee to the LP pool because the LineaSpokePool will deposit any ETH in its balance into WETH rather than keep an ETH balance to use for these leaf executions. (In other words, the Linea SpokePool is dumb and doesn't segregate leaf execution fees from LP funds, therefore it is the responsibility of the leaf executor to only send an appropriate amount to pay for fees exactly when those fees are meant to be used).

The damage right now is minimal, totalling ~0.01 ETH for 10 leaf executions.

Also, sending value needlessly on these executions doesn't allow the MulticallerClient to batch these txns, adding extra gas costs


james-a-morris (approved):
+1

mrice32 (approved):
LGTM!

PR #1349, opened 2024-03-22
improve(Dataworker): Make logs more clear when choosing to skip exchange rate updates

nicholaspai:
- Add details to logs
- Fix one bug to skip duplicate tokens in `_updateOldExchangeRates`.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1554,7 +1554,7 @@ export class Dataworker {
     const l1TokensWithPotentiallyOlderUpdate = expectedTrees.poolRebalanceTree.leaves.reduce((l1TokenSet, leaf) => {
       const currLeafL1Tokens = leaf.l1Tokens;
       currLeafL1Tokens.forEach((l1Token) => {
-        if (!l1TokenSet[l1Token] && !updatedL1Tokens.has(l1Token)) {
+        if (!l1TokenSet.includes[l1Token] && !updatedL1Tokens.has(l1Token)) {
           l1TokenSet.push(l1Token);
         }
       });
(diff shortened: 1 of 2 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1554,7 +1554,7 @@ export class Dataworker {
     const l1TokensWithPotentiallyOlderUpdate = expectedTrees.poolRebalanceTree.leaves.reduce((l1TokenSet, leaf) => {
       const currLeafL1Tokens = leaf.l1Tokens;
       currLeafL1Tokens.forEach((l1Token) => {
-        if (!l1TokenSet[l1Token] && !updatedL1Tokens.has(l1Token)) {
+        if (!l1TokenSet.includes[l1Token] && !updatedL1Tokens.has(l1Token)) {
```
james-a-morris, line 1557: OOC `l1TokenSet` is an array - was this previously working or is this a bug fix?

james-a-morris (approved):
Looks good. Nice catch on the variable spelling. One question about the change from [l1Token] to .includes(l1Token)

src/dataworker/Dataworker.ts
mrice32, line 1557:
```suggestion
        if (!l1TokenSet.includes(l1Token) && !updatedL1Tokens.has(l1Token)) {
```

mrice32 (approved):
Only comment is a syntax bug. Otherwise, LGTM!

src/dataworker/Dataworker.ts
nicholaspai: this is a bug fix. it should work now no?

PR #1355, opened 2024-03-25
improve(relayer): Execute destination transactions async

pxrl:
This change updates the relayer to submit transactions as soon as they are ready for each destination chain, rather than waiting until all destination transaction queues are ready. This should reduce the overall execution time for the relayer when it has transactions to submit on multiple chains, and should improve relayer performance on chains with a high chainId.


test/Relayer.BasicFill.ts
```diff
@@ -198,11 +200,7 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
 
       await updateAllClients();
       await relayerInstance.checkForUnfilledDepositsAndFill();
-      expect(lastSpyLogIncludes(spy, "Filling v3 deposit")).to.be.true;
-      expect(multiCallerClient.transactionCount()).to.equal(1); // One transaction, filling the one deposit.
-
-      const tx = await multiCallerClient.executeTransactionQueue();
-      expect(tx.length).to.equal(1); // There should have been exactly one transaction.
+      expect(lastSpyLogIncludes(spy, "Filled v3 deposit")).to.be.true;
```
pxrl, line 203: fwiw I'm not a huge fan of relying on log inspection to determine whether the test passed or failed, but it seems tricky to retain the ability to see the transaction queue when the queues get executed async.

src/clients/MultiCallerClient.ts
```diff
@@ -120,13 +121,17 @@ export class MultiCallerClient {
   }
 
   // For each chain, collate the enqueued transactions and process them in parallel.
-  async executeTxnQueues(simulate = false): Promise<Record<number, string[]>> {
-    const chainIds = [...new Set(Object.keys(this.valueTxns).concat(Object.keys(this.txns)))];
+  async executeTxnQueues(simulate = false, chainIds: number[] = []): Promise<Record<number, string[]>> {
+    if (chainIds.length === 0) {
+      chainIds = sdkUtils.dedupArray([
```
nicholaspai, line 126: Why not keep the Set?

src/relayer/Relayer.ts
```diff
@@ -369,15 +369,21 @@ export class Relayer {
...
+
       await this.evaluateFills(
         unfilledDeposits.map(({ deposit }) => deposit),
         maxBlockNumbers,
         sendSlowRelays
       );
+
+      const destinationChainId = Number(chainId);
+      if (multiCallerClient.getQueuedTransactions(destinationChainId).length > 0) {
+        await multiCallerClient.executeTxnQueues(simulate, [destinationChainId]);
```
nicholaspai, line 385: I'm trying to think if executing these now has any second order effects of affecting our inventory client logic, can you think of any?

src/clients/MultiCallerClient.ts
pxrl: dedupArray() implements that Set dedplication logic behind the scenes: https://github.com/across-protocol/sdk-v2/blob/8315bf4e6ac02ad435f75411ad060501806b1e89/src/utils/ArrayUtils.ts#L7

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: An earlier version of this PR definitely had a potential problem because it tried to run each deposit for each destination chain in parallel, and that could have easily messed up the tokenClient in various ways. As the PR currently stands, all chains are executed in parallel, but each of the deposits for the destination chain are evaluated in sequence, and this should ensure that the tokenClient remains coherent.

Calls into the tokenClient are typically qualified by chainId, so it should be OK to run per-chain updates on that in parallel. It also looks like this is the case in the InventoryClient when evaluating the repayment chain.

pxrl:
Status on this PR: thinking about a better way of verifying in tests; relying solely on log inspection isn't great.

test/Relayer.BasicFill.ts
pxrl: @nicholaspai I found a solution that I'm pretty happy with: `Relayer.checkUnfilledDepositsAndFill()` now returns an object containing an array of txn hashes (or an empty array) for each chain. This allows us to verify whether a txn was submitted on a specific chain or not.

https://github.com/across-protocol/relayer-v2/pull/1355/commits/7d61b8171adf4f874da778c6de1a4dd653bef29b

pxrl:
> Status on this PR: thinking about a better way of verifying in tests; relying solely on log inspection isn't great.

This has been resolved in this commit: https://github.com/across-protocol/relayer-v2/pull/1355/commits/7d61b8171adf4f874da778c6de1a4dd653bef29b

nicholaspai (approved):

PR #1360, opened 2024-03-28
refactor: Drop v2 types

pxrl:
A follow-up change would drop all references to V3 types in favour of the vanilla variants. That hasn't been done here to keep this change on the smaller side.

(nb. requires a new sdk-v2 package so it's expected to fail build)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.14",
     "@across-protocol/contracts-v2": "2.5.4",
-    "@across-protocol/sdk-v2": "0.22.19",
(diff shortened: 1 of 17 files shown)


nicholaspai (approved):

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.14",
     "@across-protocol/contracts-v2": "2.5.4",
-    "@across-protocol/sdk-v2": "0.22.19",
+    "@across-protocol/sdk-v2": "0.23.0",
```
james-a-morris, line 15: Thoughts on bumping to 0.23.1 to include the new fix?

james-a-morris (approved):
One question

PR #1361, opened 2024-03-28
improve(FillUtils): Filter SpokePoolClients by update state

pxrl:
This seems to be almost the only change required to support handling SpokePoolClients that fail to complete their update.

### DIFF
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -1,6 +1,6 @@
 import assert from "assert";
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
-import { HubPoolClient, SpokePoolClient } from "../clients";
+import { HubPoolClient } from "../clients";
 import { Fill, FillStatus, SpokePoolClientsByChain, V2DepositWithBlock, V3DepositWithBlock } from "../interfaces";
 import { getBlockForTimestamp, getRedisCache } from "../utils";
 import { isDefined } from "./";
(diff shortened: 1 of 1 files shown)


src/utils/FillUtils.ts
```diff
@@ -67,25 +67,25 @@ export async function getUnfilledDeposits(
...
+      const hints = { lowBlock: spokePoolClient.deploymentBlock };
       const startBlock = await getBlockForTimestamp(originChainId, timestamp, blockFinder, redis, hints);
-      originFromBlocks[originChainId] = Math.max(deploymentBlock, startBlock);
+      originFromBlocks[originChainId] = Math.max(spokePoolClient.deploymentBlock, startBlock);
     });
   }
 
   // Iterate over each chainId and check for unfilled deposits.
-  await sdkUtils.mapAsync(chainIds, async (destinationChainId) => {
+  await sdkUtils.mapAsync(chainIds, async (destinationChainId: number) => {
```
pxrl, line 88: Driveby change because tsc is complaining that `destinationChainId` implicitly resolves to `any`.

src/utils/FillUtils.ts
```diff
@@ -67,25 +67,27 @@ export async function getUnfilledDeposits(
   depositLookBack?: number
 ): Promise<{ [chainId: number]: RelayerUnfilledDeposit[] }> {
   const unfilledDeposits: { [chainId: number]: RelayerUnfilledDeposit[] } = {};
-  const chainIds = Object.values(spokePoolClients).map(({ chainId }) => chainId);
+  const chainIds = Object.values(spokePoolClients)
+    .filter(({ isUpdated }) => isUpdated)
+    .map(({ chainId }) => chainId);
```
james-a-morris, line 72: Should we enforce the numeric type of chainId? It would implicitly type the `chainIds`.
```suggestion
    .map(({ chainId }) => Number(chainId));
```

james-a-morris (commented):
One nit

src/utils/FillUtils.ts
pxrl: @james-a-morris `SpokePoolClient.chainId` is explicitly defined as a number, so per current implementation it's not necessary and chainIds should be `number[]`. Is the suggestion to protect against `SpokePoolClient.chainId` changing type?

nicholaspai (approved):

PR #1362, opened 2024-03-29
feat(relayer): use arweave bundle to improve performance in relayer

james-a-morris:
This PR adds the ability for the relayer to leverage stored bundle data from Arweave. This data is then leveraged to improve performance when determining an optimal refund chain id.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -161,13 +161,20 @@ export class BundleDataClient {
     this.bundleTimestampCache[key] = timestamps;
   }
 
+  getArweaveClientKey(blockRangesForChains: number[][]): string {
+    return `bundles-${blockRangesForChains}`;
+  }
+
   private async loadPersistedDataFromArweave(
     blockRangesForChains: number[][]
   ): Promise<LoadDataReturnValue | undefined> {
     if (!isDefined(this.clients?.arweaveClient)) {
       return undefined;
     }
-    const persistedData = await this.clients.arweaveClient.getByTopic(`bundles-${blockRangesForChains}`, BundleDataSS);
+    const persistedData = await this.clients.arweaveClient.getByTopic(
+      this.getArweaveClientKey(blockRangesForChains),
+      BundleDataSS
+    );
     // If there is no data or the data is empty, return undefined because we couldn't
     // pull info from the Arweave persistence layer.
     if (!isDefined(persistedData) || persistedData.length < 1) {
@@ -390,14 +397,14 @@ export class BundleDataClient {
         this.logger.debug({
           at: "BundleDataClient#loadData",
           message: "Loaded data from Arweave",
-          blockRangesForChains,
+          blockRangesForChains: this.getArweaveClientKey(blockRangesForChains),
         });
         data = Promise.resolve(arweaveData);
       } else {
         this.logger.debug({
           at: "BundleDataClient#loadData",
           message: "Failed to load data from Arweave",
-          blockRangesForChains,
+          blockRangesForChains: this.getArweaveClientKey(blockRangesForChains),
         });
         data = this._loadData(blockRangesForChains, spokePoolClients, logData);
       }
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -40,6 +40,7 @@ import {
   ExpiredDepositsToRefundV3,
   LoadDataReturnValue,
 } from "../interfaces/BundleData";
+import { BundleDataToPersistToDALayerTypeSS } from "../utils/SuperstructUtils";
 
 type DataCache = Record<string, Promise<LoadDataReturnValue>>;
 
@@ -161,6 +162,69 @@ export class BundleDataClient {
 
   setBundleTimestampsInCache(key: string, timestamps: { [chainId: number]: number[] }): void {
     this.bundleTimestampCache[key] = timestamps;
+  }
+
+  async loadPersistedData(
+    blockRangesForChains: number[][]
+  ): Promise<{ bundleFillsV3: BundleFillsV3; expiredDepositsToRefundV3: ExpiredDepositsToRefundV3 }> {
+    const data = await this.clients.arweaveClient.getByTopic(
+      `bundles-${blockRangesForChains}`,
+      BundleDataToPersistToDALayerTypeSS
+    );
+    // If there is no data or the data is empty, return undefined because we couldn't
+    // pull info from the Arweave persistence layer.
+    if (!isDefined(data) || data.length < 1) {
+      return undefined;
+    }
+    const { bundleFillsV3: _bundleFillsV3, expiredDepositsToRefundV3: _expiredDepositsToRefundV3 } = data[0].data;
+    const bundleFillsV3 = _bundleFillsV3 as unknown as BundleFillsV3;
+    const expiredDepositsToRefundV3 = _expiredDepositsToRefundV3 as unknown as ExpiredDepositsToRefundV3;
+    return { bundleFillsV3, expiredDepositsToRefundV3 };
+  }
+
+  /**
+   * An analog function to getPendingRefundsFromValidBundles(1) that loads persisted data from the Arweave.
+   * @dev This function does not modify the state of the BundleDataClient
+   * @returns The refunds from the last fully executed bundle that has been persisted to the Arweave. Undefined if no
+   *          persisted data is found.
+   */
+  async getPersistedPendingRefundsFromLastValidBundle(): Promise<CombinedRefunds[] | undefined> {
+    // We have only been referencing the latest fully executed root bundle in our
+    // calculations when calling `this.bundleDataClient.getPendingRefundsFromValidBundles()`
+    const bundleToReference = this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(
+      this.clients.hubPoolClient.latestBlockSearched
+    );
+    const bundleEvaluationBlockRanges = getImpliedBundleBlockRanges(
+      this.clients.hubPoolClient,
+      this.clients.configStoreClient,
+      bundleToReference
+    );
+    const persistedData = await this.loadPersistedData(bundleEvaluationBlockRanges);
+    if (persistedData === undefined) {
+      return undefined;
+    }
+    const { bundleFillsV3, expiredDepositsToRefundV3 } = persistedData;
+    return [getRefundsFromBundle(bundleFillsV3, expiredDepositsToRefundV3)];
+  }
+
+  async getPersistedNextBundleRefunds(): Promise<CombinedRefunds> {
+    if (!this.clients.hubPoolClient.hasPendingProposal()) {
+      return undefined;
+    }
+    // If there is a pending proposal, then the most recent proposed root bundle
+    // should be the pending bundle per the hub pool client
+    const bundleToReference = this.clients.hubPoolClient.getLatestProposedRootBundle();
+    const bundleEvaluationBlockRanges = getImpliedBundleBlockRanges(
+      this.clients.hubPoolClient,
+      this.clients.configStoreClient,
+      bundleToReference
+    );
+    const persistedData = await this.loadPersistedData(bundleEvaluationBlockRanges);
+    if (persistedData === undefined) {
+      return undefined;
+    }
+    const { bundleFillsV3, expiredDepositsToRefundV3 } = persistedData;
+    return getRefundsFromBundle(bundleFillsV3, expiredDepositsToRefundV3);
   }
 
   async getPendingRefundsFromValidBundles(bundleLookback = 1): Promise<CombinedRefunds[]> {
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -40,7 +40,7 @@ import {
   ExpiredDepositsToRefundV3,
   LoadDataReturnValue,
 } from "../interfaces/BundleData";
-import { BundleDataToPersistToDALayerTypeSS } from "../utils/SuperstructUtils";
+import { BundleDataSS } from "../utils/SuperstructUtils";
 
 type DataCache = Record<string, Promise<LoadDataReturnValue>>;
 
@@ -168,10 +168,7 @@ export class BundleDataClient {
     if (!isDefined(this.clients?.arweaveClient)) {
       return undefined;
     }
-    const persistedData = await this.clients.arweaveClient.getByTopic(
-      `bundles-${blockRangesForChains}`,
-      BundleDataToPersistToDALayerTypeSS
-    );
+    const persistedData = await this.clients.arweaveClient.getByTopic(`bundles-${blockRangesForChains}`, BundleDataSS);
     // If there is no data or the data is empty, return undefined because we couldn't
     // pull info from the Arweave persistence layer.
     if (!isDefined(persistedData) || persistedData.length < 1) {
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -247,8 +247,7 @@ export class BundleDataClient {
     // have a pending proposal, we can assume that a corresponding Arweave entry exists. In this case, we
     // need to capture the `impliedBlockRange` so that it matches with the tag in the Arweave entry. If no
     // pending bundle exists, we should cast as large of a next bundle range as possible to capture the most
-    // data. We can do this with `getWidestPossibleExpectedBlockRange` which is a superset of the ranges for
-    // `impliedBlockRange`.
+    // data. We can do this with `getWidestPossibleExpectedBlockRange`.
     const futureBundleEvaluationBlockRanges = hubPoolClient.hasPendingProposal()
       ? getImpliedBundleBlockRanges(
           hubPoolClient,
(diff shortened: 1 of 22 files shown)


src/utils/SuperstructUtils.ts
james-a-morris, line 1: The superstruct type to confirm the data we are receiving from Arweave is of the correct type.

src/clients/BundleDataClient.ts
```diff
@@ -163,6 +164,29 @@ export class BundleDataClient {
...
+    if (!isDefined(persistedData) || persistedData.length < 1) {
+      return undefined;
+    }
+    const data = persistedData[0].data;
+    return {
+      bundleFillsV3: data.bundleFillsV3 as unknown as BundleFillsV3,
+      expiredDepositsToRefundV3: data.expiredDepositsToRefundV3 as unknown as ExpiredDepositsToRefundV3,
+      bundleDepositsV3: data.bundleDepositsV3 as unknown as BundleDepositsV3,
+      unexecutableSlowFills: data.unexecutableSlowFills as unknown as BundleExcessSlowFills,
+      bundleSlowFillsV3: data.bundleSlowFillsV3 as unknown as BundleSlowFills,
```
james-a-morris, line 186: Most of the reason we have to do this type cast is that these types are Record<number, Record<string, ...>> and SuperStruct is assuming we're using `Record<string, Record<string, ...>>` which is technically correct because all records in TS are indexed with strings - but fails TSC.

src/utils/SignerUtils.ts
```diff
@@ -121,9 +122,46 @@ async function getSecretSigner(): Promise<Signer> {
...
+      kty: "RSA",
+      e: "AQAB",
+      n: "0",
+      d: "0",
+      p: "0",
+      q: "0",
+      dp: "0",
+      dq: "0",
+      qi: "0",
+    };
```
james-a-morris, line 140: This should be a valid JWK but it will fail if used to `set` any data to arweave.

test/mocks/MockBundleDataClient.ts
james-a-morris, line 1: Added to make the tests work

src/clients/BundleDataClient.ts
nicholaspai: is this potentially dangerous? Does this mean that if the parsed JSON data is in the incorrect shape that we won't know about it until some runtime error later?

src/relayer/RelayerConfig.ts
```diff
@@ -23,6 +23,7 @@ export class RelayerConfig extends CommonConfig {
   readonly relayerMessageGasMultiplier: BigNumber;
   readonly minRelayerFeePct: BigNumber;
   readonly acceptInvalidFills: boolean;
+  readonly retrieveCacheDataFromArweave: boolean;
```
nicholaspai, line 26: How is this used?

src/dataworker/DataworkerClientHelper.ts
```diff
@@ -58,9 +57,11 @@ export async function constructDataworkerClients(
     new defiLlama.PriceFeed(),
   ]);
 
-  // Define the Arweave client.
+  // Define the Arweave client. We need to use a read-write signer for the
+  // dataworker to persist bundle data if `persistingBundleData` is enabled.
+  // Otherwise, we can use a read-only signer.
   const arweaveClient = new caching.ArweaveClient(
-    config.arweaveWalletJWK,
+    getArweaveJWKSigner(config.persistingBundleData ? "read-write" : "readonly"),
```
nicholaspai, line 64: So the idea is for the proposer and eventually the disputer to have read-write keys so that they can post and re-post to Arweave respectively?

nicholaspai (commented):
Really nice PR, left some comments

src/clients/BundleDataClient.ts
```diff
@@ -163,6 +164,29 @@ export class BundleDataClient {
...
+  private async loadPersistedDataFromArweave(blockRangesForChains: number[][]): Promise<LoadDataReturnValue> {
+    if (!isDefined(this.clients?.arweaveClient)) {
+      return undefined;
+    }
+    const persistedData = await this.clients.arweaveClient.getByTopic(
+      `bundles-${blockRangesForChains}`,
+      BundleDataToPersistToDALayerTypeSS
+    );
+    // If there is no data or the data is empty, return undefined because we couldn't
+    // pull info from the Arweave persistence layer.
```
pxrl, line 176: The return type expects `LoadDataReturnValue` which consists of a bunch of subtypes, so returning `undefined` seems counter-intuitive for me. There are also multiple possible causes for returning `undefined` from this function. Thoughts on encapsulating these "failures" with a reason string so the caller can know why they didn't get any persistent data?

src/clients/BundleDataClient.ts
```diff
@@ -216,20 +240,35 @@ export class BundleDataClient {
...
-      getEndBlockBuffers(chainIds, this.blockRangeEndBlockBuffer),
-      this.clients,
-      this.clients.hubPoolClient.latestBlockSearched,
-      this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
-    );
+
+    // We need to find the widest possible range of blocks to search for refunds. Since we aim to leverage
+    // Arweave's DA layer to look for entries, we need to make sure we're using an expected range. If we
+    // have a pending proposal, we can assume that a corresponding Arweave entry exists. In this case, we
+    // need capture the `impliedBlockRange` so that it matches with the tag in the Arweave entry. If no
```
pxrl, line 247:
```suggestion
    // need to capture the `impliedBlockRange` so that it matches with the tag in the Arweave entry. If no
```

src/clients/BundleDataClient.ts
```diff
@@ -216,20 +240,35 @@ export class BundleDataClient {
...
+    // pending bundle exists, we should cast as large of a next bundle range as possible to capture the most
+    // data. We can do this with `getWidestPossibleExpectedBlockRange` which is a superset of the ranges for
+    // `impliedBlockRange`.
```
pxrl, line 250: Is `getWidestPossibleExpectedBlockRange` really a superset? If there's no pending proposal then shouldn't it produce a block range that starts where the previous `impliedBlockRange` ended? Otherwise we'd end up re-proposing over the same block range as the previous proposal (...or I'm just confused about this).

src/clients/BundleDataClient.ts
```diff
@@ -216,20 +240,35 @@ export class BundleDataClient {
...
+          this.clients.hubPoolClient.latestBlockSearched,
+          this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
+        );
     // Refunds that will be processed in the next bundle that will be proposed after the current pending bundle
     // (if any) has been fully executed.
     const { bundleFillsV3, expiredDepositsToRefundV3 } = await this.loadData(
       futureBundleEvaluationBlockRanges,
       this.spokePoolClients,
-      false
+      false,
```
pxrl, line 270: Can we assign these to named booleans so that it's easier to identify what's being set here?

src/clients/BundleDataClient.ts
```diff
@@ -326,12 +365,23 @@ export class BundleDataClient {
...
-    logData = true
+    logData = true,
+    attemptArweaveLoad = false
   ): Promise<LoadDataReturnValue> {
     const key = JSON.stringify(blockRangesForChains);
 
     if (!this.loadDataCache[key]) {
-      this.loadDataCache[key] = this._loadData(blockRangesForChains, spokePoolClients, logData);
+      // We need to await this data to see if we were able to load
+      // data to Arweave.
```
pxrl, line 375:
```suggestion
      // We need to await this data to see if we were able to load data to Arweave.
```

src/dataworker/DataworkerClientHelper.ts
pxrl: Does the disputer need to write anything? Intuitively I'd only expect the proposer to do that.

src/utils/SignerUtils.ts
```diff
@@ -121,9 +122,46 @@ async function getSecretSigner(): Promise<Signer> {
   }
 }
 
+export function getArweaveJWKSigner(keyType: "readonly" | "read-write"): ArweaveWalletJWKInterface {
+  // If the keytype is readonly, we should generate a read-only key
+  // on the fly and return it.
+  if (keyType === "readonly") {
+    // This is a dummy key. It is meant to fail if used.
```
pxrl, line 129: The existing wording seems contradictory to the comment above and implies (for me) that it's a temporary key to be replaced by something else.

```suggestion
    // This is a dummy key. It is meant to fail if used to write to Arweave.
```

src/utils/SignerUtils.ts
```diff
@@ -121,9 +122,46 @@ async function getSecretSigner(): Promise<Signer> {
   }
 }
 
+export function getArweaveJWKSigner(keyType: "readonly" | "read-write"): ArweaveWalletJWKInterface {
```
pxrl, line 125: Could this be hoisted to use `SignerOptions` as above? That'd be nice to retain the use of `clearEnv`, and the other types seem generic enough that it might be useful without needing to make changes to it.

src/utils/SuperstructUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+const V3RelayDataSS = {
+  ...RelayDataCommonSS,
+  inputToken: string(),
+  inputAmount: BigNumberType,
+  outputToken: string(),
+  outputAmount: BigNumberType,
+  fillDeadline: number(),
+  exclusiveRelayer: string(),
+  exclusivityDeadline: number(),
+};
```
pxrl, line 47: These have been re-merged back in the sdk fwiw, so could probably also be done here to eliminate `RelayDataCommonSS`.

src/clients/BundleDataClient.ts
nicholaspai: we might just want to error out and let caller catch

nicholaspai: this way this can be used by someone to repost to arweave in case data is missing

src/clients/BundleDataClient.ts
nicholaspai: agreed

src/utils/SuperstructUtils.ts
james-a-morris: To clarify do you mean the superstruct types are in the SDK?

src/clients/BundleDataClient.ts
james-a-morris: I've updated the type so that they must be positive numeric integers. I've also further refined the sub-dictionary for the keys to be valid ETH addresses.

src/clients/BundleDataClient.ts
```diff
@@ -163,6 +164,29 @@ export class BundleDataClient {
     this.bundleTimestampCache[key] = timestamps;
   }
 
+  private async loadPersistedDataFromArweave(blockRangesForChains: number[][]): Promise<LoadDataReturnValue> {
+    if (!isDefined(this.clients?.arweaveClient)) {
+      return undefined;
+    }
+    const persistedData = await this.clients.arweaveClient.getByTopic(
+      `bundles-${blockRangesForChains}`,
+      BundleDataToPersistToDALayerTypeSS
```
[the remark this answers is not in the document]
james-a-morris: Done

src/dataworker/DataworkerClientHelper.ts
james-a-morris: I'd think we only need the proposer to have access to the read-write. `persistingBundleData` should only be enabled with the proposer.

src/clients/BundleDataClient.ts
james-a-morris: The Arweave client should output logs if it can't find any data. I've updated the return type of the function to include an optional `undefined` type.

Returning an empty value is technically within the scope of the function unless we want to throw any time we can't find potentially missing data (data that hasn't yet persisted).

src/clients/BundleDataClient.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/1362/commits/b52a54a83bc07549f648f58a196fcf6b89d034bc

src/relayer/RelayerConfig.ts
james-a-morris: Currently it is not used. It was a placeholder - would we see a case where we'd not want to reference the arweave data if available?

src/clients/BundleDataClient.ts
nicholaspai: This is how I would roughly describe the heuristic here
- If there is a pending bundle, then there should be an arweave file containing its bundle data so load it
- if there is no pending bundle we need to compute what the next bundle is so we figure out the block ranges using getWidestPossibleExpectedBlockRange and then call loadData with it

src/utils/SuperstructUtils.ts
pxrl: Actually just the split of ´RelayDataCommonSS` and `V3RelayDataSS`. This was done in the SDK when there were still V2 and V3 types. I might be misunderstanding the purpose of it here, but I don't think I see `RelayDataCommonSS` used anywhere else, so I'm wondering if it can be merged directly into `RelayDataSS`.

src/clients/BundleDataClient.ts
```diff
@@ -216,20 +241,37 @@ export class BundleDataClient {
...
-      this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
-    );
+
+    // We need to find the widest possible range of blocks to search for refunds. Since we aim to leverage
+    // Arweave's DA layer to look for entries, we need to make sure we're using an expected range. If we
+    // have a pending proposal, we can assume that a corresponding Arweave entry exists. In this case, we
+    // need to capture the `impliedBlockRange` so that it matches with the tag in the Arweave entry. If no
+    // pending bundle exists, we should cast as large of a next bundle range as possible to capture the most
+    // data. We can do this with `getWidestPossibleExpectedBlockRange` which is a superset of the ranges for
+    // `impliedBlockRange`.
```
pxrl, line 251: Per comments above ^^

```suggestion
    // data. We can do this with `getWidestPossibleExpectedBlockRange`.
```

src/clients/BundleDataClient.ts
james-a-morris: Yep - fair call

src/relayer/RelayerConfig.ts
pxrl: I think third-parties without inventory management might not see the need for it. It is a marginal configuration overhead to add, so it increases friction. It'd be nice to have the option of skipping that IMO.

pxrl (approved):
This looks really good. There are a couple of comment threads to close out but otherwise I'm happy w/ it :+1:

src/clients/BundleDataClient.ts
james-a-morris: A bit of a loop around but we can manually correct for the `string` -> `number` type in just the top level record key: https://github.com/across-protocol/relayer-v2/pull/1362/commits/ef5d8b9c626baf03ad7ab845a7f91da76ca0f12d

src/relayer/RelayerConfig.ts
nicholaspai: third parties just won't set their arweave config so all the arweave will be skipped. Why would anyone set an arweave config but set this value to false? I see this as an unneccessary config and we don't even use it. Over-engineering

src/clients/BundleDataClient.ts
nicholaspai: that's good actually so if the caller hasn't set their arweave config, then this step fails and we fallback to current logic. This would be ideal

src/clients/BundleDataClient.ts
nicholaspai: what do you think is safer?

src/utils/SuperstructUtils.ts
james-a-morris: Done: https://github.com/across-protocol/relayer-v2/pull/1362/commits/869ec3f14b800b5c7c60b0377bbe2907fc9b2d27

james-a-morris:
Signoff from @nicholaspai offline.

src/clients/BundleDataClient.ts
```diff
@@ -160,6 +161,39 @@ export class BundleDataClient {
...
+    const convertTypedStringRecordIntoNumericRecord = <UnderlyingType>(
+      data: Record<string, Record<string, UnderlyingType>>
+    ): Record<number, Record<string, UnderlyingType>> =>
+      Object.keys(data).reduce((acc, chainId) => {
+        acc[Number(chainId)] = data[chainId];
+        return acc;
+      }, {} as Record<number, Record<string, UnderlyingType>>);
+
+    const data = persistedData[0].data;
+    return {
```
nicholaspai, line 188: nit: can we add a debug log somewhere in this function to confirm that we've loaded and parsed correctly from Arweave

nicholaspai (approved):
Just one nit !

src/clients/BundleDataClient.ts
james-a-morris: Looks to be added

src/relayer/RelayerConfig.ts
james-a-morris: Removed ^

PR #1364, opened 2024-03-30
improve(relayer): Batch-compute all expected LP fees

pxrl:
The introduction of v3 temporarily removed the ability to batch compute LP fees on deposit ingestion because the (re)payment chain is no longer determined by the deposit destination chain. Instead, the relayer has been computing LP fees at least once per candidate fill, and often twice in the event that the preferred repayment chain is mainnet.

One of the major downsides of computing LP fees individually is that the BlockFinder must be invoked to resolve the quoteTimestamp to a mainnet block number. There are significant savings when this is batched due to duplicate quoteTimestamp consolidation in the HubPoolClient.

This PR restores batch computation of LP fees by aggregating all outstanding deposits and computing them in bulk. This is performed for both destination chain and mainnet. This permits the relayer to subsequently evaluate profitability on the fly, without needing to yield for LP fee computation. The net effect should thus be that relayer runtimes are more consistent and less reactive to a surge in the number of unfilled deposits.


src/clients/HubPoolClient.ts
```diff
@@ -42,4 +42,8 @@ export class HubPoolClient extends clients.HubPoolClient {
 
     return await super.computeRealizedLpFeePct(deposit);
   }
+
+  async batchComputeRealizedLpFeePct(lpFeeRequests: LpFeeRequest[]): Promise<interfaces.RealizedLpFee[]> {
```
pxrl, line 46: Added for tsc/editor appeasement - otherwise it struggles to resolve the correct return type. Would actually be nice to remove this entire local `HubPoolClient` definition, though...

src/relayer/Relayer.ts
```diff
@@ -465,31 +510,30 @@ export class Relayer {
...
     const originChain = getNetworkName(originChainId);
     const destinationChain = getNetworkName(destinationChainId);
 
     const preferredChainId = await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address);
-    const { realizedLpFeePct } = await hubPoolClient.computeRealizedLpFeePct({
-      ...deposit,
-      paymentChainId: preferredChainId,
-    });
+    const { lpFeePct } = repaymentFees.find(({ paymentChainId }) => paymentChainId === preferredChainId);
+    assert(isDefined(lpFeePct));
```
pxrl, line 529: Temporarily added for testing.

src/relayer/Relayer.ts
```diff
@@ -503,10 +547,10 @@ export class Relayer {
...
         deposit: { originChain, depositId, destinationChain, transactionHash },
       });
-      const { realizedLpFeePct: destinationChainLpFeePct } = await hubPoolClient.computeRealizedLpFeePct({
-        ...deposit,
-        paymentChainId: destinationChainId,
-      });
+      const { lpFeePct: destinationChainLpFeePct } = repaymentFees.find(
+        ({ paymentChainId }) => paymentChainId === destinationChainId
+      );
+      assert(isDefined(lpFeePct));
```
pxrl, line 553: Temporarily added for testing.

src/relayer/Relayer.ts
```diff
@@ -331,17 +347,44 @@ export class Relayer {
...
     }
   }
 
+  // Compute realizedLpFeePct over all
+  async batchComputeLpFees(deposits: V3DepositWithBlock[]): Promise<BatchLPFees> {
+    const { hubPoolClient } = this.clients;
+
+    const lpFeeRequests = deposits
+      .map((deposit) => [
+        { ...deposit, paymentChainId: hubPoolClient.chainId },
```
nicholaspai, line 367: Why create two entries per deposit?

pxrl: This computes the LP fee for taking repayment on mainnet vs. destination chain.

The HubPoolClient can be updated in future to accept an array of payment chain IDs, so then we'd only need 1 deposit and would compute multiple fees on it. The overhead of doing it this way is probably not noticeable, though.

src/relayer/Relayer.ts
```diff
@@ -331,17 +347,44 @@ export class Relayer {
...
+        { ...deposit, paymentChainId: deposit.destinationChainId },
```
pxrl, line 368: This should only be added conditional upon destinationChainId !== hubPoolChainId.

src/relayer/Relayer.ts
```diff
@@ -23,6 +24,16 @@ const { getAddress } = ethersUtils;
...
 const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
 
+type RepaymentFee = { inputAmount: BigNumber; paymentChainId: number; lpFeePct: BigNumber };
+
+type BatchLPFees = {
+  [destinationChainId: number]: {
+    [inputToken: string]: {
+      [quoteTimestamp: number]: RepaymentFee[];
+    };
+  };
```
pxrl, line 34: This isn't robust to the same token address being used on multiple origin chains (as is the case for DAI and BAL on some chains). It might be better to have a flat mapping of `originChainId-originToken-destinationChainId-quoteTimestamp` or something.

src/relayer/Relayer.ts
nicholaspai: good call. Why not do this now?

src/relayer/Relayer.ts
nicholaspai: yeah agreed we should fix this especially since ETH is likely to have the same token on many chains and so will DAI and BAL

nicholaspai (commented):
It seems like this PR is still WIP so pls request a review when ready to merge

src/clients/HubPoolClient.ts
james-a-morris: OOC what is the error code? It looks like we have the same types in this version and in the SDK definition

src/relayer/Relayer.ts
```diff
@@ -23,6 +24,16 @@ const { getAddress } = ethersUtils;
 const { isDepositSpedUp, isMessageEmpty, resolveDepositMessage } = sdkUtils;
 const UNPROFITABLE_DEPOSIT_NOTICE_PERIOD = 60 * 60; // 1 hour
 
+type RepaymentFee = { inputAmount: BigNumber; paymentChainId: number; lpFeePct: BigNumber };
+
+type BatchLPFees = {
+  [destinationChainId: number]: {
+    [inputToken: string]: {
+      [quoteTimestamp: number]: RepaymentFee[];
+    };
```
james-a-morris, line 33: I think we use this paradigm in a couple of locations. We have a dictionary of chain ids where each value is a dictionary of input tokens. 

I wonder if it would make sense to normalize this into a `ChainTokenMapping` - something like:

```
type ChainTokenMapping<Value> = Record<number, Record<string, Value>>;
```

src/relayer/Relayer.ts
james-a-morris: I think it may be prudent to keep this as a sanity check.

src/relayer/Relayer.ts
pxrl: Updated here: https://github.com/across-protocol/relayer-v2/pull/1364/commits/cefc716174e7fb53fa2fdf94c6c2f33613f0aa4b

Array entries are now stored according to the following key: `originChainId-inputToken-quoteTimestamp`. This ensures that the input token is always correctly identified and maps correctly to a slot in the HubPool. This key returns an array of fee objects that consist of `inputAmount`, `paymentChainId` and `lpFeePct`. This combination should be flexible enough that we can correctly identify each deposit, and the structure of the `BatchLPFees` object is now flat and simple,

src/clients/HubPoolClient.ts
pxrl: It's a generic complaint from tsc that `batchComputeRealizedLpFeePct()` doesn't exist on the HubPoolClient. I checked again and it only appears when I have a locally-linked sdk-v2 instance, so I'll revert the change. I'll circle back and see if we can drop the locally-defined `HubPoolClient` though. 

hubPoolClient

pxrl: Dropped: https://github.com/across-protocol/relayer-v2/pull/1364/commits/8d0cba73f2783d002b74c8e6419052804aff4a4b

src/relayer/Relayer.ts
pxrl: Implemented filtering for destination mainnet: https://github.com/across-protocol/relayer-v2/pull/1364/commits/62515ed9d2012b3fe9b5cee2a202208f9fa56ad4.

src/relayer/Relayer.ts
```diff
@@ -330,6 +339,15 @@ export class Relayer {
     }
   }
 
+  /**
+   * For a given deposit, map its relevant attributes to a string to be used as a lookup into the LP fee structure.
+   * @param relayData An object consisting of an originChainId, inputToken and quoteTimestamp.
+   * @returns A string identifying the deposit in a BatchLPFees object.
+   */
+  private getLPFeeKey(relayData: Pick<V3Deposit, "originChainId" | "inputToken" | "quoteTimestamp">): string {
+    return `${relayData.originChainId}-${relayData.inputToken}-${relayData.quoteTimestamp}`;
```
nicholaspai, line 348: shouldn't `repaymentChain` be factored into this?

src/relayer/Relayer.ts
```diff
@@ -491,31 +545,30 @@ export class Relayer {
...
     const { depositId, originChainId, destinationChainId, inputAmount, outputAmount, transactionHash } = deposit;
     const originChain = getNetworkName(originChainId);
     const destinationChain = getNetworkName(destinationChainId);
 
     const preferredChainId = await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address);
-    const { realizedLpFeePct } = await hubPoolClient.computeRealizedLpFeePct({
-      ...deposit,
-      paymentChainId: preferredChainId,
-    });
+    const { lpFeePct } = repaymentFees.find(({ paymentChainId }) => paymentChainId === preferredChainId);
```
nicholaspai, line 563: I don't fully understand this, shouldn't `lpFee` be based on the quote timestamp as well?

src/relayer/Relayer.ts
pxrl: This resolves to an array of `RepaymentFee` types, where each `RepaymentFee` type consists of `inputAmount`, `paymentChainId` and `lpFeePct`. So the idea is that you can identify the array solely from a deposit, and from there can iterate over the array to determine the profitability of taking repayment on each chain.

For now we only query fees for `destinationChainId` and `hubPoolChainId`, but I imagine in future it'll also include taking repayment on `originChainId` since that's best for the protocol.

src/relayer/Relayer.ts
pxrl: The input to `evaluateFill()` and subsequently `resolveRepaymentChain()` (see `repaymentFees` above) is an array of LP fees that were identified for the input deposit based on its `originChainId`, `inputToken` and `quoteTimestamp`. So the deposit `quoteTimestamp` is already factored in here.

src/relayer/Relayer.ts
pxrl: This shouldn't trigger in production, but if it did it'd be a very disruptive event. I'll wind it back to a warning log and will just conclude that the fill is not profitable instead - that'd avoid this impacting other unrelated deposits.

src/relayer/Relayer.ts
pxrl: Looking at it again, it misses a 2nd check on `inputAmount` matching, since it aggregates LP fees by attributes that aren't unique to a single deposit.

nicholaspai: > The input to evaluateFill() and subsequently resolveRepaymentChain() (see repaymentFees above) is an array of LP fees that were identified for the input deposit based on its originChainId, inputToken and quoteTimestamp. So the deposit quoteTimestamp is already factored in here.

right ok this makes sense but when I previously looked at `getLPFeeKey` it did not contain `quoteTimestamp` as a key

nicholaspai (commented):
I think this makes sense, how much speedup can we expect?

src/relayer/Relayer.ts
```diff
@@ -341,15 +355,56 @@ export class Relayer {
...
+   */
+  async batchComputeLpFees(deposits: V3DepositWithBlock[]): Promise<BatchLPFees> {
+    const { hubPoolClient } = this.clients;
+
+    const lpFeeRequests = deposits
+      .map((deposit) => {
+        const request = [{ ...deposit, paymentChainId: deposit.destinationChainId }];
+        if (deposit.destinationChainId !== hubPoolClient.chainId) {
+          request.push({ ...deposit, paymentChainId: hubPoolClient.chainId });
+        }
```
pxrl, line 389: Origin chain repayment can also be added here for ~0 cost.

nicholaspai: As in, early exit?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: Moreso just ensuring that we have the LP fees on hand to support origin chain repayment profitability estimation. As I udnerstand it, we'd need to have all 3 on hand (or 2 if one collides w/ hub chain). This commit now implements that: https://github.com/across-protocol/relayer-v3/pull/1364/commits/0fd568f8a97ab1c6c441c9b222d00869a4bebfbc

PR #1371, opened 2024-04-02
fix(USDT): allow BaseAdapter to support approving USDT

james-a-morris:
USDT on mainnet requires a spender's allowance to be zero prior to approving more value. 

This PR adds a zero'd approval if the token attempting to be approved is USDT.

### DIFF
--- a/src/clients/bridges/BaseAdapter.ts
+++ b/src/clients/bridges/BaseAdapter.ts
@@ -168,12 +168,11 @@ export abstract class BaseAdapter {
     let mrkdwn = "*Approval transactions:* \n";
     for (const { l1Token, targetContract } of tokensToApprove) {
       const { hubChainId } = this;
-      const txnRequests = [];
+      const txs = [];
       if (TOKEN_APPROVALS_TO_FIRST_ZERO[hubChainId]?.includes(l1Token.address)) {
-        txnRequests.push(runTransaction(this.logger, l1Token, "approve", [targetContract, bnZero]));
+        txs.push(await runTransaction(this.logger, l1Token, "approve", [targetContract, bnZero]));
       }
-      txnRequests.push(runTransaction(this.logger, l1Token, "approve", [targetContract, MAX_UINT_VAL]));
-      const txs = await Promise.all(txnRequests);
+      txs.push(await runTransaction(this.logger, l1Token, "approve", [targetContract, MAX_UINT_VAL]));
       const receipts = await Promise.all(txs.map((tx) => tx.wait()));
       const hubNetwork = getNetworkName(hubChainId);
       const spokeNetwork = getNetworkName(this.chainId);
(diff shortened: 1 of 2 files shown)


pxrl:
> USDT on mainnet requires a spender's allowance to be zero prior to approving more value.
> 
> This PR adds a zero'd approval if the token attempting to be approved is USDT.

Is this currently erroring somewhere?

src/common/Constants.ts
```diff
@@ -295,3 +295,13 @@ export const chainIdsToCctpDomains: { [chainId: number]: number } = {
...
 };
+
+/**
+ * A mapping of chain IDs to tokens on that chain which need their allowance
+ * to first be zeroed before setting a new allowance. This is useful for
+ * tokens that have a non-standard approval process.
+ * @dev this is a generalization for USDT on Ethereum. Other tokens may be added
+ */
+export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
+  1: ["0xdAC17F958D2ee523a2206206994597C13D831ec7"],
```
pxrl, line 306: wdyt about referencing the existing constants we have?

```suggestion
     [CHAIN_IDs.MAINNET]: [TOKEN_SYMBOLS_MAP.USDT.addresses[CHAIN_IDs.MAINNET],
```

nicholaspai (approved):

james-a-morris:
> > USDT on mainnet requires a spender's allowance to be zero prior to approving more value.
> > This PR adds a zero'd approval if the token attempting to be approved is USDT.
> 
> Is this currently erroring somewhere?

Yes - we're getting an error in the BaseAdapter when a new EOA needs to approve USDT

src/clients/bridges/BaseAdapter.ts
```diff
@@ -161,22 +160,31 @@ export abstract class BaseAdapter {
...
     for (const { l1Token, targetContract } of tokensToApprove) {
-      const tx = await runTransaction(this.logger, l1Token, "approve", [targetContract, MAX_UINT_VAL]);
-      const receipt = await tx.wait();
       const { hubChainId } = this;
+      const txnRequests = [];
+      if (TOKEN_APPROVALS_TO_FIRST_ZERO[hubChainId]?.includes(l1Token.address)) {
+        txnRequests.push(runTransaction(this.logger, l1Token, "approve", [targetContract, bnZero]));
+      }
+      txnRequests.push(runTransaction(this.logger, l1Token, "approve", [targetContract, MAX_UINT_VAL]));
+      const txs = await Promise.all(txnRequests);
```
pxrl, line 176: Is there any chance of this causing nonce reuse issues? There'll be multiple points along the submission workflow where ethers blocks (i.e. when simulating the transaction to estimate the gasLimit, or potentially to query the transaction count for the EOA). Could this result in the same nonce being used for both transactions prior to submission?

(...I wonder if it's more robust to just submit them in sequence)

james-a-morris: This is fair - and there is also the case to consider that we need to guarantee that the zero-value approval happens first. I've made it sequential.

PR #1373, opened 2024-04-02
improve(FillUtils): Protect getUnfilledDeposits() against RPC error

pxrl:
If the deposit fill status cannot be queried on the destination chain, revert to slower methods to determine the status. This fallback method is less accurate due to timing, but it allows execution to continue and the complete set of resulting deposits with unknown status will likely be reduced by subsequent filtering anyway (i.e. based on supported token, number of deposit confirmations, profitability, ...). The ultimate fill status for the residual deposits will be inferred at the time of fill simulation on the destination chain.

### DIFF
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -74,8 +74,12 @@ export async function getUnfilledDeposits(
       const spokePoolClient = spokePoolClients[originChainId];
       const timestamp = spokePoolClient.getCurrentTime() - depositLookBack;
       const hints = { lowBlock: spokePoolClient.deploymentBlock };
-      const startBlock = await getBlockForTimestamp(originChainId, timestamp, blockFinder, redis, hints);
-      originFromBlocks[originChainId] = Math.max(spokePoolClient.deploymentBlock, startBlock);
+      try {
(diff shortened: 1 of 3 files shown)


src/utils/FillUtils.ts
```diff
@@ -97,13 +97,20 @@ export async function getUnfilledDeposits(
         // Find all unfilled deposits for the current loops originChain -> destinationChain.
         return originClient
           .getDepositsForDestinationChain(destinationChainId)
-          .filter((deposit) => deposit.blockNumber >= earliestBlockNumber);
+          .filter(({ blockNumber }) => blockNumber >= earliestBlockNumber);
```
pxrl, line 100: Driveby change - not directly related, but close enough that it made sense to modify it here.

pxrl:
Some additional context:
Intuitively I want to place this kind of error handling directly in `fillStatusArray()`, but the decision to default to `FillStatus.Unfilled´ is opinionated. It suits the relayer where we prioritise service over absolute correctness, but not the dataworker, since this might (for example) cause it to miss legitimate slow fill requests, or to incorrectly enqueue a refund.

nicholaspai (approved):

PR #1378, opened 2024-04-03
fix: Avoid error where we try to compute refunds for a {l1Token, chainId} combination that doesn't exist

nicholaspai:
For example, when `getBundleRefunds()` is called with `l1Token = USDT address`, then this function fails because USDT doesn't exist on Base, which is one of the enabled chains.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -231,7 +231,6 @@ export class InventoryClient {
       // Consider any refunds from executed and to-be executed bundles.
       totalRefundsPerChain = await this.getBundleRefunds(l1Token);
     } catch (e) {
-      console.error(e);
       this.log("Failed to get bundle refunds, defaulting refund chain to hub chain");
       // Fallback to ignoring bundle refunds if calculating bundle refunds goes wrong.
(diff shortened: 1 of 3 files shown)


pxrl (approved):
Really nice catch!

src/clients/InventoryClient.ts
```diff
@@ -166,13 +166,15 @@ export class InventoryClient {
...
-      this.getEnabledChains().map((chainId) => {
-        const destinationToken = this.getDestinationTokenForL1Token(l1Token, chainId);
-        return [
-          chainId,
-          this.bundleDataClient.getTotalRefund(refundsToConsider, this.relayer, Number(chainId), destinationToken),
-        ];
-      })
+      this.getEnabledChains()
+        .filter((chainId) => this.hubPoolClient.l2TokenEnabledForL1Token(l1Token, chainId))
+        .map((chainId) => {
```
james-a-morris, line 171: Great catch

james-a-morris (approved):
+1 - thoughts on using a reduce to prevent the double loop?

nicholaspai:
> +1 - thoughts on using a reduce to prevent the double loop?

yeah good call

PR #1379, opened 2024-04-03
Revert "fix: Avoid error where we try to compute refunds for a {l1Token, chainId} combination that doesn't exist (#1378)"

pxrl:
This reverts commit cf222c0ac0eb4f934c399b0efb8175d5e1b95767.


pxrl:
Unnecessary

PR #1382, opened 2024-04-04
fix(BundleDataClient): Handle refunds sent after pending bundle

nicholaspai:
- Fixes regression introduced in [#1362](https://github.com/across-protocol/relayer-v2/issues/1362) where refunds sent after proposed bundle are no longer computed and accounted for in `InventoryClient.resolveRepaymentChain` in favor of loading pendling bundle data from Arweave
- Removes `loadData` calls from all InventoryClient flows and replaces it with a function `getApproximateRefundsForBlockRange()` that filters on fills sent by relayer within the bundle block range and assumes all fills are valid which is a fair assumption and greatly speeds up the inventory client run
- Removes try-catch in InventoryClient since we should not expect any failures anymore when computing bundle refunds
- Adds Promise.all and promise caching to speed up overall inventory client call to `getBundleRefunds`
- Adds detailed logging

Fixes ACX-2039

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -177,7 +177,7 @@ export class BundleDataClient {
     if (!isDefined(this.clients?.arweaveClient)) {
       return undefined;
     }
-    const start = Date.now();
+    const start = performance.now();
     const persistedData = await this.clients.arweaveClient.getByTopic(
       this.getArweaveClientKey(blockRangesForChains),
       BundleDataSS
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -29,7 +29,11 @@ import {
   getRefundsFromBundle,
   CombinedRefunds,
 } from "../dataworker/DataworkerUtils";
-import { getWidestPossibleExpectedBlockRange, isChainDisabled } from "../dataworker/PoolRebalanceUtils";
+import {
+  getBlockRangeDelta,
+  getWidestPossibleExpectedBlockRange,
+  isChainDisabled,
+} from "../dataworker/PoolRebalanceUtils";
 import { utils } from "@across-protocol/sdk-v2";
 import {
   BundleDepositsV3,
@@ -173,6 +177,7 @@ export class BundleDataClient {
     if (!isDefined(this.clients?.arweaveClient)) {
       return undefined;
     }
+    const start = Date.now();
     const persistedData = await this.clients.arweaveClient.getByTopic(
       this.getArweaveClientKey(blockRangesForChains),
       BundleDataSS
@@ -194,13 +199,27 @@ export class BundleDataClient {
       }, {} as Record<number, Record<string, UnderlyingType>>);
 
     const data = persistedData[0].data;
-    return {
+    const bundleData = {
       bundleFillsV3: convertTypedStringRecordIntoNumericRecord(data.bundleFillsV3),
       expiredDepositsToRefundV3: convertTypedStringRecordIntoNumericRecord(data.expiredDepositsToRefundV3),
       bundleDepositsV3: convertTypedStringRecordIntoNumericRecord(data.bundleDepositsV3),
       unexecutableSlowFills: convertTypedStringRecordIntoNumericRecord(data.unexecutableSlowFills),
       bundleSlowFillsV3: convertTypedStringRecordIntoNumericRecord(data.bundleSlowFillsV3),
     };
+    this.logger.debug({
+      at: "BundleDataClient#loadPersistedDataFromArweave",
+      message: `Loaded persisted data from Arweave in ${Math.floor(Date.now() - start) / 1000}s.`,
+      blockRanges: JSON.stringify(blockRangesForChains),
+      bundleData: prettyPrintV3SpokePoolEvents(
+        bundleData.bundleDepositsV3,
+        bundleData.bundleFillsV3,
+        [], // Invalid fills are not persisted to Arweave.
+        bundleData.bundleSlowFillsV3,
+        bundleData.expiredDepositsToRefundV3,
+        bundleData.unexecutableSlowFills
+      ),
+    });
+    return bundleData;
   }
 
   async getPendingRefundsFromValidBundles(bundleLookback = 1): Promise<CombinedRefunds[]> {
@@ -250,7 +269,8 @@ export class BundleDataClient {
   // in a valid bundle with all of its leaves executed. This contains refunds from:
   // - Bundles that passed liveness but have not had all of their pool rebalance leaves executed.
   // - Bundles that are pending liveness
-  async getNextBundleRefunds(): Promise<CombinedRefunds> {
+  // - Fills sent after the pending, but not validated, bundle
+  async getNextBundleRefunds(): Promise<CombinedRefunds[]> {
     const hubPoolClient = this.clients.hubPoolClient;
     const nextBundleMainnetStartBlock = hubPoolClient.getNextBundleStartBlockNumber(
       this.chainIdListForBundleEvaluationBlockNumbers,
@@ -258,38 +278,58 @@ export class BundleDataClient {
       hubPoolClient.chainId
     );
     const chainIds = this.clients.configStoreClient.getChainIdIndicesForBlock(nextBundleMainnetStartBlock);
+    const combinedRefunds: CombinedRefunds[] = [];
 
-    // We need to find the widest possible range of blocks to search for refunds. Since we aim to leverage
-    // Arweave's DA layer to look for entries, we need to make sure we're using an expected range. If we
-    // have a pending proposal, we can assume that a corresponding Arweave entry exists. In this case, we
-    // need to capture the `impliedBlockRange` so that it matches with the tag in the Arweave entry. If no
-    // pending bundle exists, we should cast as large of a next bundle range as possible to capture the most
-    // data. We can do this with `getWidestPossibleExpectedBlockRange`.
-    const futureBundleEvaluationBlockRanges = hubPoolClient.hasPendingProposal()
-      ? getImpliedBundleBlockRanges(
-          hubPoolClient,
-          this.clients.configStoreClient,
-          hubPoolClient.getLatestProposedRootBundle()
-        )
-      : getWidestPossibleExpectedBlockRange(
-          chainIds,
-          this.spokePoolClients,
-          getEndBlockBuffers(chainIds, this.blockRangeEndBlockBuffer),
-          this.clients,
-          this.clients.hubPoolClient.latestBlockSearched,
-          this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
-        );
-    // Refunds that will be processed in the next bundle that will be proposed after the current pending bundle
-    // (if any) has been fully executed.
+    // @dev: If spoke pool client is undefined for a chain, then the end block will be null or undefined, which
+    // should be handled gracefully and effectively cause this function to ignore refunds for the chain.
+    let widestBundleBlockRanges = getWidestPossibleExpectedBlockRange(
+      chainIds,
+      this.spokePoolClients,
+      getEndBlockBuffers(chainIds, this.blockRangeEndBlockBuffer),
+      this.clients,
+      this.clients.hubPoolClient.latestBlockSearched,
+      this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
+    );
+
+    // If there is a pending bundle that has not been fully executed, then it should have arweave
+    // data so we can load it from there.
+    if (hubPoolClient.hasPendingProposal()) {
+      const pendingBundleBlockRanges = getImpliedBundleBlockRanges(
+        hubPoolClient,
+        this.clients.configStoreClient,
+        hubPoolClient.getLatestProposedRootBundle()
+      );
+      const logData = false;
+      const attemptToLoadFromArweave = true;
+      const { bundleFillsV3, expiredDepositsToRefundV3 } = await this.loadData(
+        pendingBundleBlockRanges,
+        this.spokePoolClients,
+        logData,
+        attemptToLoadFromArweave
+      );
+      combinedRefunds.push(getRefundsFromBundle(bundleFillsV3, expiredDepositsToRefundV3));
+
+      // Shorten the widestBundleBlockRanges now to not double count the pending bundle blocks.
+      widestBundleBlockRanges = getBlockRangeDelta(
+        pendingBundleBlockRanges,
+        widestBundleBlockRanges,
+        this.spokePoolClients,
+        chainIds
+      );
+    }
+
+    // Next, load all refunds sent after the last bundle proposal.
     const logData = false;
-    const attemptToLoadFromArweave = true;
+    const attemptToLoadFromArweave = false;
     const { bundleFillsV3, expiredDepositsToRefundV3 } = await this.loadData(
-      futureBundleEvaluationBlockRanges,
+      widestBundleBlockRanges,
       this.spokePoolClients,
       logData,
       attemptToLoadFromArweave
     );
-    return getRefundsFromBundle(bundleFillsV3, expiredDepositsToRefundV3);
+    const nextBundleRefunds = getRefundsFromBundle(bundleFillsV3, expiredDepositsToRefundV3);
+    combinedRefunds.push(nextBundleRefunds);
+    return combinedRefunds;
   }
 
   getExecutedRefunds(
@@ -300,6 +340,9 @@ export class BundleDataClient {
       [relayer: string]: BigNumber;
     };
   } {
+    if (!isDefined(spokePoolClient)) {
+      return {};
+    }
     // @dev Search from right to left since there can be multiple root bundles with the same relayer refund root.
     // The caller should take caution if they're trying to use this function to find matching refunds for older
     // root bundles as opposed to more recent ones.
@@ -419,6 +462,7 @@ export class BundleDataClient {
     spokePoolClients: SpokePoolClientsByChain,
     logData = true
   ): Promise<LoadDataReturnValue> {
+    const start = Date.now();
     const key = JSON.stringify(blockRangesForChains);
 
     if (!this.clients.configStoreClient.isUpdated) {
@@ -988,6 +1032,12 @@ export class BundleDataClient {
       });
     }
 
+    this.logger.debug({
+      at: "BundleDataClient#_loadData",
+      message: `Computed bundle data in ${Math.floor(Date.now() - start) / 1000}s.`,
+      blockRangesForChains: JSON.stringify(blockRangesForChains),
+      v3SpokeEventsReadable,
+    });
     return {
       bundleDepositsV3,
       expiredDepositsToRefundV3,
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -251,15 +251,10 @@ export class BundleDataClient {
       bundle
     );
     const attemptToLoadFromArweave = true;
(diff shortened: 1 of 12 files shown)


james-a-morris (commented):
I like this. I've been running into these errors while debugging and these logs will help get a better understanding of where the code is failing

src/clients/InventoryClient.ts
```diff
@@ -225,13 +226,31 @@ export class InventoryClient {
...
     let totalRefundsPerChain: { [chainId: string]: BigNumber } = {};
+    // TODO: Remove this try-catch. getBundleRefunds should never fail in practice. This was originally added
+    // to catch the case where the relayer's lookback is insufficient to compute a bundle but I believe the current
+    // loadData implementation will just miss events in the case where the lookback is too short, rather than fail.
     try {
-      // Consider any refunds from executed and to-be executed bundles.
-      totalRefundsPerChain = await this.getBundleRefunds(l1Token);
+      // Consider any refunds from executed and to-be executed bundles. If bundle data client doesn't return in
+      // time, return an object with zero refunds for all chains.
+      totalRefundsPerChain = await Promise.race([this.getBundleRefunds(l1Token), getBundleRefundTimeout(45 * 60)]);
```
james-a-morris, line 251: Thoughts on using a constant variable instead of hardcoding 45*60? Or even allowing an override by the ENV

nicholaspai:
@james-a-morris @pxrl the goal of this PR was changed to just make refunds more correct. It removes the try-catch also as a side-effect and adds detailed timing logs

src/clients/BundleDataClient.ts
```diff
@@ -173,6 +177,7 @@ export class BundleDataClient {
     if (!isDefined(this.clients?.arweaveClient)) {
       return undefined;
     }
+    const start = Date.now();
```
pxrl, line 180: Matt previously identified that performance.now() is preferable because it uses a monotonic clock.

```suggestion
    const start = pperformance.now();
```

src/clients/BundleDataClient.ts
```diff
@@ -194,13 +199,27 @@ export class BundleDataClient {
...
+    const bundleData = {
       bundleFillsV3: convertTypedStringRecordIntoNumericRecord(data.bundleFillsV3),
       expiredDepositsToRefundV3: convertTypedStringRecordIntoNumericRecord(data.expiredDepositsToRefundV3),
       bundleDepositsV3: convertTypedStringRecordIntoNumericRecord(data.bundleDepositsV3),
       unexecutableSlowFills: convertTypedStringRecordIntoNumericRecord(data.unexecutableSlowFills),
       bundleSlowFillsV3: convertTypedStringRecordIntoNumericRecord(data.bundleSlowFillsV3),
     };
+    this.logger.debug({
+      at: "BundleDataClient#loadPersistedDataFromArweave",
+      message: `Loaded persisted data from Arweave in ${Math.floor(Date.now() - start) / 1000}s.`,
```
pxrl: The rounding function should include the division. nb. we actually have an `msToS()` helper in src/utils.

```suggestion
      message: `Loaded persisted data from Arweave in ${Math.round(performance.now() - start / 1000)}s.`,
```

src/clients/BundleDataClient.ts
```diff
@@ -194,13 +199,27 @@ export class BundleDataClient {
...
+      blockRanges: JSON.stringify(blockRangesForChains),
```
pxrl, line 212: I think the stringify call is probably redunance since js should print numeric arrays correctly anyway.

src/clients/BundleDataClient.ts
```diff
@@ -250,46 +269,67 @@ export class BundleDataClient {
...
+    // @dev: If spoke pool client is undefined for a chain, then the end block will be null or undefined, which
+    // should be handled gracefully and effectively cause this function to ignore refunds for the chain.
+    let widestBundleBlockRanges = getWidestPossibleExpectedBlockRange(
+      chainIds,
+      this.spokePoolClients,
+      getEndBlockBuffers(chainIds, this.blockRangeEndBlockBuffer),
+      this.clients,
+      this.clients.hubPoolClient.latestBlockSearched,
+      this.clients.configStoreClient.getEnabledChains(this.clients.hubPoolClient.latestBlockSearched)
+    );
```
pxrl, line 292: `widestBundleBlockRanges` will almost always be overwritten down at line 313 without ever being read because most of the time there will be a pending proposal. Worth refactoring to avoid that double work?

src/clients/BundleDataClient.ts
```diff
@@ -300,6 +340,9 @@ export class BundleDataClient {
       [relayer: string]: BigNumber;
     };
   } {
+    if (!isDefined(spokePoolClient)) {
+      return {};
+    }
```
pxrl, line 345: IMO it's a bug on the calling side to call this function with an undefined SpokePoolClient...but I'm puzzled as to how that could even occur in the first place.

src/clients/BundleDataClient.ts
```diff
@@ -988,6 +1032,12 @@ export class BundleDataClient {
       });
     }
 
+    this.logger.debug({
+      at: "BundleDataClient#_loadData",
+      message: `Computed bundle data in ${Math.floor(Date.now() - start) / 1000}s.`,
```
pxrl: Same as above RE `Math.round((...) / 1000)`.

src/clients/BundleDataClient.ts
nicholaspai: it does not print them nicely its multiline

src/clients/BundleDataClient.ts
nicholaspai: We need to compute it though anyways and its a tiny cost so I think its fine

src/clients/BundleDataClient.ts
```diff
@@ -250,46 +270,78 @@ export class BundleDataClient {
...
+        unexecutableSlowFills: {},
+      };
+    };
+    const logData = false;
+    const attemptToLoadFromArweave = false;
+    // @dev If the loadData call times out, it will continue to resolve in the background, which provides benefits
+    // for the next call to loadData.
+    const { bundleFillsV3, expiredDepositsToRefundV3 } = await Promise.race([
+      this.loadData(widestBundleBlockRanges, this.spokePoolClients, logData, attemptToLoadFromArweave),
+      loadDataTimeout(Number(process.env.LOAD_DATA_TIMEOUT ?? 20)), // timeout denominated in seconds
```
pxrl, line 340: Will reading `process.env.LOAD_DATA_TIMEOUT` directly in this code have the potential to impact test behaviour now or in future if someone has it defined in their environment? Could it alternatively be passed in as config to the BundleDataClient, or to this `getNextBundleRefunds()` function as an argument?

src/clients/BundleDataClient.ts
```diff
@@ -969,15 +1025,6 @@ export class BundleDataClient {
       expiredDepositsToRefundV3,
       unexecutableSlowFills
     );
-    if (logData) {
```
pxrl, line 972: It looks like there's only one log message still guarded by this boolean, and it's a very small log message...Should we just drop `logData` entirely from this function?

src/clients/InventoryClient.ts
```diff
@@ -225,29 +225,11 @@ export class InventoryClient {
...
-        })
-      );
-    }
-
-    this.log(`Time taken to get bundle refunds: ${Math.floor(Date.now() - startTime) / 1000}s`, {
+    const startTime = performance.now();
+    // Consider any refunds from executed and to-be executed bundles. If bundle data client doesn't return in
+    // time, return an object with zero refunds for all chains.
+    const totalRefundsPerChain: { [chainId: string]: BigNumber } = await this.getBundleRefunds(l1Token);
+    this.log(`Time taken to get bundle refunds: ${Math.floor(performance.now() - startTime) / 1000}s`, {
```
pxrl, line 232: Minor, but `Math.floor()` has no effect on any of these calculations because performance.now() and startTime are both at ms precision. `Math.round()` is probably also a better choice because it'll round to the nearest integer, rather than unconditionally round downwards.

```suggestion
    this.log(`Time taken to get bundle refunds: ${Math.round((performance.now() - startTime) / 1000)}s`, {
```

pxrl (approved):
> This extra computation adds run-time but not a lot since the block ranges queried are small but increasing as the challenge period gets larger

This is a fairly heavy process to be doing in the relayer just to find out the relayer refunds for a single relayer. For future, wdyt about a simplification that just scrapes all fills for the specific relayer after the latest proposal end block, and computes expected repayments as deposit.inputAmount for each fill (i.e. deliberately ignoring the LP fee, since we assume it's small anyway).

It'd be a coarse estimate, but I can't see the margin of error being enough to meaningfully impact repayment choice selection in most cases, and it skips the need to compute everything else in the bundle (most of which isn't relevant for this specific purpose). I think tradeoffs like this can definitely make sense in the relayer, where speed is really the priority.

src/clients/BundleDataClient.ts
pxrl: Yeah - I missed that it was passed as an input argument.

nicholaspai:
> > This extra computation adds run-time but not a lot since the block ranges queried are small but increasing as the challenge period gets larger
> 
> This is a fairly heavy process to be doing in the relayer just to find out the expected repayments for a single relayer. For future, wdyt about a simplification that just scrapes all fills for the specific relayer after the latest proposal end block, and computes expected repayments as deposit.inputAmount for each fill (i.e. deliberately ignoring the LP fee, since we assume it's small anyway).
> 
> It'd be a coarse estimate, but I can't see the margin of error being enough to meaningfully impact repayment choice selection in most cases, and it skips the need to compute everything else in the bundle (most of which isn't relevant for this specific purpose). I think tradeoffs like this can definitely make sense in the relayer, where speed is really the priority.

I think this is a great idea

src/clients/BundleDataClient.ts
nicholaspai: hmm yeah good call, i like passing it into `getNextBundleRefunds`

src/clients/BundleDataClient.ts
nicholaspai: good call, done

nicholaspai:
> > This extra computation adds run-time but not a lot since the block ranges queried are small but increasing as the challenge period gets larger
> 
> This is a fairly heavy process to be doing in the relayer just to find out the expected repayments for a single relayer. For future, wdyt about a simplification that just scrapes all fills for the specific relayer after the latest proposal end block, and computes expected repayments as deposit.inputAmount for each fill (i.e. deliberately ignoring the LP fee, since we assume it's small anyway).
> 
> It'd be a coarse estimate, but I can't see the margin of error being enough to meaningfully impact repayment choice selection in most cases, and it skips the need to compute everything else in the bundle (most of which isn't relevant for this specific purpose). I think tradeoffs like this can definitely make sense in the relayer, where speed is really the priority.

implemented in this commit 5d480ea582b79a181213679c9b92a1e6f5dd5765

src/clients/BundleDataClient.ts
```diff
@@ -215,83 +238,162 @@ export class BundleDataClient {
...
+          const refundAmount = fill.inputAmount;
+          refundsForChain[chainToSendRefundTo] ??= {};
+          refundsForChain[chainToSendRefundTo][repaymentToken] ??= {};
+          if (refundsForChain[chainToSendRefundTo][repaymentToken][fill.relayer]) {
+            refundsForChain[chainToSendRefundTo][repaymentToken][fill.relayer] =
+              refundsForChain[chainToSendRefundTo][repaymentToken][fill.relayer].add(refundAmount);
+          } else {
+            refundsForChain[chainToSendRefundTo][repaymentToken][fill.relayer] = refundAmount;
+          }
+        });
```
pxrl, line 326: Small simplification that deletes a few lines:

```suggestion
          const { relayer, inputAmount: refundAmount } = fill;
          refundsForChain[chainToSendRefundTo] ??= {};
          refundsForChain[chainToSendRefundTo][repaymentToken] ??= {};
          const existingRefundAmount = refundsForChain[chainToSendRefundTo][repaymentToken][relayer] ?? bnZero;
          refundsForChain[chainToSendRefundTo][repaymentToken][relayer] = existingRefundAmount.add(refundAmount);
```

src/clients/InventoryClient.ts
```diff
@@ -154,16 +155,54 @@ export class InventoryClient {
     this.crossChainTransferClient.increaseOutstandingTransfer(this.relayer, l1Token, rebalance, Number(chainId));
   }
 
+  async getAllBundleRefunds(): Promise<CombinedRefunds[]> {
+    const refunds: CombinedRefunds[] = [];
+    // @dev This call can take a long time if bundle data for the latest bundle and the pending bundle is not
+    // available on Arweave.
```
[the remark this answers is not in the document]
nicholaspai: Yeah this is a leftover comment will remove

src/clients/InventoryClient.ts
```diff
@@ -154,16 +155,54 @@ export class InventoryClient {
...
-
-    // Consider refunds from next bundle to be proposed:
-    const nextBundleRefunds = await this.bundleDataClient.getNextBundleRefunds();
-    refundsToConsider.push(nextBundleRefunds);
+    if (this.bundleRefundsPromise) {
+      refundsToConsider = await this.bundleRefundsPromise;
+    } else {
+      this.bundleRefundsPromise = this.getAllBundleRefunds();
+      refundsToConsider = lodash.cloneDeep(await this.bundleRefundsPromise);
+    }
```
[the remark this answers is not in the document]
nicholaspai: As opposed to where?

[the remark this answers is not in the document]
nicholaspai: Added a comment

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -229,6 +230,32 @@ export function getRunningBalanceForL1Token(
...
+): number[][] {
+  return widestBlockRanges.map((blockRange, index) => {
+    // If chain is disabled, return disabled range
+    if (initialBlockRanges[index][0] === initialBlockRanges[index][1]) {
+      return initialBlockRanges[index];
+    }
+    // If no spoke pool client for chain, consider it disabled
+    if (!isDefined(spokePoolClients[enabledChainIds[index]])) {
+      return [initialBlockRanges[index][1], initialBlockRanges[index][1]];
+    }
```
james-a-morris, line 249: This feels like the calling code should handle it & not this `getBlockRangeDelta` function. Thoughts on hoisting it up outside this fn?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -229,6 +230,32 @@ export function getRunningBalanceForL1Token(
...
+    // If pending bundle end block and widest end block are the same, return an empty range since there are no
+    // "new" events to consider for this chain.
```
james-a-morris, line 251: We may also want to add a note that this case also includes the event that the initialBlockRange covers the `widestBlockRange`. It shouldn't happen in practice, but this check is implicitly made.

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -229,6 +230,32 @@ export function getRunningBalanceForL1Token(
   return runningBalance.sub(desiredTransferAmount);
 }
 
+// Return block ranges for blocks after initialBlockRanges and up to widestBlockRanges.
+// If a chain is disabled or doesn't have a spoke pool client, return a range of 0
+export function getBlockRangeDelta(
```
james-a-morris, line 235: This _shouldn't_ happen, but did we want to consider the case where the `initialBlockRanges[x][0] > widestBlockRanges[x][0]`

james-a-morris (commented):
First round of comments - will circle back

src/clients/InventoryClient.ts
```diff
@@ -154,16 +155,54 @@ export class InventoryClient {
...
     // Increase virtual balance by pending relayer refunds from the latest valid bundle and the
     // upcoming bundle. We can assume that all refunds from the second latest valid bundle have already
     // been executed.
-    const refundsToConsider: CombinedRefunds[] = await this.bundleDataClient.getPendingRefundsFromValidBundles();
-
-    // Consider refunds from next bundle to be proposed:
-    const nextBundleRefunds = await this.bundleDataClient.getNextBundleRefunds();
-    refundsToConsider.push(nextBundleRefunds);
+    if (this.bundleRefundsPromise) {
+      refundsToConsider = await this.bundleRefundsPromise;
```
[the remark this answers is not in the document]
nicholaspai: why? Wouldn't that cause other parallel runs to get undefined?

src/dataworker/Dataworker.ts
```diff
@@ -473,7 +473,7 @@ export class Dataworker {
   ): Promise<ProposeRootBundleReturnType> {
     const timerStart = Date.now();
     const { bundleDepositsV3, bundleFillsV3, bundleSlowFillsV3, unexecutableSlowFills, expiredDepositsToRefundV3 } =
-      await this.clients.bundleDataClient.loadData(blockRangesForProposal, spokePoolClients, logData);
+      await this.clients.bundleDataClient.loadData(blockRangesForProposal, spokePoolClients);
```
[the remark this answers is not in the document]
nicholaspai: yes

src/clients/InventoryClient.ts
```diff
@@ -154,16 +155,54 @@ export class InventoryClient {
     this.crossChainTransferClient.increaseOutstandingTransfer(this.relayer, l1Token, rebalance, Number(chainId));
   }
 
+  async getAllBundleRefunds(): Promise<CombinedRefunds[]> {
+    const refunds: CombinedRefunds[] = [];
+    // @dev This call can take a long time if bundle data for the latest bundle and the pending bundle is not
+    // available on Arweave.
+    const [pendingRefunds, nextBundleRefunds] = await Promise.all([
+      this.bundleDataClient.getPendingRefundsFromValidBundles([this.relayer]),
+      this.bundleDataClient.getNextBundleRefunds([this.relayer]),
```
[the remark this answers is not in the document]
nicholaspai: yea that's an ok suggestion. think the spread operator makes it harder to add params in the future but its cleaner for now

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
pxrl: nvm - discussed offline.

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
james-a-morris: That is a fair point, but since this looks to be the only location where we set `this.bundleRefundsPromise,`, did we want to account for the data possibly going stale? Or, because we primarily run this loop-less, would it be assumed that the data won't change in the context of a single run

src/clients/InventoryClient.ts
```diff
@@ -154,16 +155,53 @@ export class InventoryClient {
...
-    // Consider refunds from next bundle to be proposed:
-    const nextBundleRefunds = await this.bundleDataClient.getNextBundleRefunds();
-    refundsToConsider.push(nextBundleRefunds);
+    if (this.bundleRefundsPromise) {
+      refundsToConsider = await this.bundleRefundsPromise;
+    } else {
+      // @dev Save this as a promise so that other parallel calls to this function don't make the same call.
+      this.bundleRefundsPromise = this.getAllBundleRefunds();
+      refundsToConsider = lodash.cloneDeep(await this.bundleRefundsPromise);
+    }
```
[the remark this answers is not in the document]
nicholaspai: hm not sure i like this

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
nicholaspai: I think we can assume we run without a loop here. @pxrl we should revisit this in looping mode

PR #1384, opened 2024-04-04
fix(BundleDataClient): Filter chains based on which spoke pool clients are passed in to loadData

nicholaspai:
Not every spoke pool client will be defined, therefore no need to check if one that doesn't exist is "updated"


mrice32 (approved):
LGTM!

PR #1390, opened 2024-04-05
improve(BundleDataClient): Cache Arweave requests

nicholaspai:
Cache the arweave request as a Promise so that we only make 1 request to Arweave per bundle block range. If these requests are sent in parallel, then the cache should be set to a Promise by the first request to send, and then when this request resolves, all parallel requests can use its response.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -127,6 +127,8 @@ function updateBundleSlowFills(dict: BundleSlowFills, deposit: V3DepositWithBloc
 // @notice Shared client for computing data needed to construct or validate a bundle.
 export class BundleDataClient {
   private loadDataCache: DataCache = {};
+  private arweaveDataCache: Record<string, Promise<LoadDataReturnValue>> = {};
+
   private bundleTimestampCache: Record<string, { [chainId: number]: number[] }> = {};
 
   // eslint-disable-next-line no-useless-constructor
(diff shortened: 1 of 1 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -386,32 +391,23 @@ export class BundleDataClient {
...
-            at: "BundleDataClient#loadData",
-            message: "Loaded data from Arweave",
-            blockRangesForChains: this.getArweaveClientKey(blockRangesForChains),
-          });
-        } else {
-          this.logger.debug({
-            at: "BundleDataClient#loadData",
-            message: "Failed to load data from Arweave",
-            blockRangesForChains: this.getArweaveClientKey(blockRangesForChains),
-          });
```
james-a-morris, line 412: Are we removing the logging?

src/clients/BundleDataClient.ts
```diff
@@ -335,6 +337,9 @@ export class BundleDataClient {
   deductExecutedRefunds(allRefunds: CombinedRefunds, bundleContainingRefunds: ProposedRootBundle): CombinedRefunds {
     for (const chainIdStr of Object.keys(allRefunds)) {
       const chainId = Number(chainIdStr);
+      if (!isDefined(this.spokePoolClients[chainId])) {
+        continue;
+      }
```
james-a-morris, line 342: Is this in: https://github.com/across-protocol/relayer-v2/pull/1391/files

src/clients/BundleDataClient.ts
nicholaspai: yeah they're kind of redundant now since we have higher logs upstream in the sdk

PR #1391, opened 2024-04-05
fix(BundleDataClient): Skip deducting refunds for spoke pool clients that are undefined

nicholaspai:


james-a-morris (approved):
I think this change exists in #1390 but LGTM

PR #1397, opened 2024-04-08
refactor: Remove top-level bot loop

pxrl:
The bot currently has 2 independent checks on POLLING_MODE - once at the top-level `index.ts` and once at `src/relayer/Relayer.ts`. A pending change will update `src/relayer/Relayer.ts` to support exiting gracefully after a SIGHUP request. The upper-layer check in `index.ts` prevents this by determining whether to loop or exit based on `config.pollingDelay`. There doesn't seem to be a strong justification for applying this looping at both levels, so remove the top-level of encapsulation for simplicity. Any bot personalities that require special looping behaviour can implement it locally, and if we see a common pattern emerge then it can bubble up to the top level once more.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,14 +1,6 @@
 import minimist from "minimist";
 import { CommonConfig } from "./src/common";
-import {
-  config,
-  delay,
-  retrieveSignerFromCLIArgs,
-  help,
-  Logger,
-  usage,
-  winston,
-} from "./src/utils";
+import { config, delay, retrieveSignerFromCLIArgs, help, Logger, usage, winston } from "./src/utils";
 import { runRelayer } from "./src/relayer";
 import { runDataworker } from "./src/dataworker";
 import { runMonitor } from "./src/monitor";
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -48,18 +36,21 @@ export async function run(args: { [k: string]: boolean | string }): Promise<void
...
+      // One global signer for use with a specific per-chain provider.
+      // todo: Support a void signer for monitor mode (only) if no wallet was supplied.
+      await cmds[cmd](logger, signer);
+    } catch (error) {
+      logger.error({
+        at: "index",
+        message: "There was an execution error!",
+        reason: error,
+        e: error,
+        error,
```
pxrl, line 49: This was copy/pasted from the existing `processCrash()`. I'm not sure why we'd be reporting `error` as `reason` _and_ `e` :thinking:. iirc some part of the logging subsystem does tend to eat any error objects emitted directly in the log object, though.

nicholaspai (approved):

PR #1399, opened 2024-04-08
feat(relayer): Gracefully stop on SIGHUP

pxrl:
This change permits the operator to request the relayer to stop at the end of its current loop. This is mostly useful for polling mode, where the relayer will loop repeatedly.

While here, add in some per-loop performance reporting and refactor a little bit by pulling some code out of the large try...catch statement.

### DIFF
--- a/src/relayer/index.ts
+++ b/src/relayer/index.ts
@@ -9,7 +9,7 @@ import {
 } from "../utils";
 import { Relayer } from "./Relayer";
 import { RelayerConfig } from "./RelayerConfig";
-import { constructRelayerClients, RelayerClients, updateRelayerClients } from "./RelayerClientHelper";
+import { constructRelayerClients, updateRelayerClients } from "./RelayerClientHelper";
 config();
 let logger: winston.Logger;
 
@@ -21,16 +21,25 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
 
   logger = _logger;
   const config = new RelayerConfig(process.env);
-  let relayerClients: RelayerClients;
 
+  let stop = false;
+  process.on('SIGHUP', () => {
+    logger.debug({
+      at: "run",
+      message: "Received SIGHUP, stopping at end of current loop."
+    });
+    stop = true;
+  });
+
+  logger[startupLogLevel(config)]({ at: "Relayer#index", message: "Relayer started 🏃‍♂️", config, relayerRun });
+  const relayerClients = await constructRelayerClients(logger, config, baseSigner);
+  const relayer = new Relayer(await baseSigner.getAddress(), logger, relayerClients, config);
+
+  let run = 1;
   try {
-    logger[startupLogLevel(config)]({ at: "Relayer#index", message: "Relayer started 🏃‍♂️", config, relayerRun });
-    relayerClients = await constructRelayerClients(logger, config, baseSigner);
-    const relayer = new Relayer(await baseSigner.getAddress(), logger, relayerClients, config);
-
-    logger.debug({ at: "Relayer#index", message: "Relayer components initialized. Starting execution loop" });
-
     for (;;) {
+      logger.debug({ at: "Relayer#index", message: `Starting relayer execution loop ${run}.` });
+      const tLoopStart = performance.now();
       await updateRelayerClients(relayerClients, config);
 
       if (!config.skipRelays) {
@@ -50,7 +59,13 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       relayerClients.profitClient.clearUnprofitableFills();
       relayerClients.tokenClient.clearTokenShortfall();
 
-      if (await processEndPollingLoop(logger, "Relayer", config.pollingDelay)) {
+      const runTime = Math.round((performance.now() - tLoopStart) / 1000);
+      logger.debug({
+        at: "Relayer#index",
+        message: `Completed relayer execution loop ${run++} in ${runTime} seconds.`
+      });
+
+      if (await processEndPollingLoop(logger, "Relayer", config.pollingDelay, stop)) {
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


src/relayer/index.ts
```diff
@@ -21,16 +21,25 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
   logger = _logger;
   const config = new RelayerConfig(process.env);
-  let relayerClients: RelayerClients;
 
-  try {
-    logger[startupLogLevel(config)]({ at: "Relayer#index", message: "Relayer started 🏃‍♂️", config, relayerRun });
-    relayerClients = await constructRelayerClients(logger, config, baseSigner);
-    const relayer = new Relayer(await baseSigner.getAddress(), logger, relayerClients, config);
+  let stop = false;
+  process.on('SIGHUP', () => {
```
pxrl, line 26: nb. This could be conditional on config.pollingDelay > 0.

src/relayer/index.ts
```diff
@@ -21,16 +21,25 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
+    logger.debug({
+      at: "run",
```
pxrl, line 28:
```suggestion
      at: "Relayer#run",
```

src/relayer/index.ts
```diff
@@ -21,16 +21,25 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
 
-    logger.debug({ at: "Relayer#index", message: "Relayer components initialized. Starting execution loop" });
+  logger[startupLogLevel(config)]({ at: "Relayer#index", message: "Relayer started 🏃‍♂️", config, relayerRun });
+  const relayerClients = await constructRelayerClients(logger, config, baseSigner);
+  const relayer = new Relayer(await baseSigner.getAddress(), logger, relayerClients, config);
 
+  let run = 1;
+  try {
     for (;;) {
+      logger.debug({ at: "Relayer#index", message: `Starting relayer execution loop ${run}.` });
```
pxrl, line 41:
```suggestion
      logger.debug({ at: "Relayer#run", message: `Starting relayer execution loop ${run}.` });
```

src/relayer/index.ts
```diff
@@ -21,16 +21,25 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
+  process.on('SIGHUP', () => {
+    logger.debug({
+      at: "run",
+      message: "Received SIGHUP, stopping at end of current loop."
+    });
+    stop = true;
+  });
 
-    logger.debug({ at: "Relayer#index", message: "Relayer components initialized. Starting execution loop" });
+  logger[startupLogLevel(config)]({ at: "Relayer#index", message: "Relayer started 🏃‍♂️", config, relayerRun });
```
pxrl, line 34:
```suggestion
  logger[startupLogLevel(config)]({ at: "Relayer#run", message: "Relayer started 🏃‍♂️", config, relayerRun });
```

src/relayer/index.ts
```diff
@@ -50,7 +59,13 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       relayerClients.profitClient.clearUnprofitableFills();
       relayerClients.tokenClient.clearTokenShortfall();
 
-      if (await processEndPollingLoop(logger, "Relayer", config.pollingDelay)) {
+      const runTime = Math.round((performance.now() - tLoopStart) / 1000);
+      logger.debug({
+        at: "Relayer#index",
```
pxrl, line 64:
```suggestion
        at: "Relayer#run",
```

pxrl (commented):
Minor log inconsistencies

pxrl:
The tldr on this is that the operator can use `kill -HUP <pid>` to send a SIGHUP to the process, and that triggers the callback established in `process.on("SIGHUP", ...)`. This sets the `stop` flag, meaning that the relayer will be permitted to complete its current loop but will gracefully exit after calling `processEndPollingLoop()`.

The motivation for this feature is that it will permit a bot in looping mode to be shut down after some period of time via any arbitrary trigger that can execute a command in the operating environment.

src/relayer/index.ts
```diff
@@ -50,7 +59,13 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
       relayerClients.tokenClient.clearTokenShortfall();
 
-      if (await processEndPollingLoop(logger, "Relayer", config.pollingDelay)) {
+      const runTime = Math.round((performance.now() - tLoopStart) / 1000);
+      logger.debug({
+        at: "Relayer#run",
+        message: `Completed relayer execution loop ${run++} in ${runTime} seconds.`,
+      });
+
+      if (await processEndPollingLoop(logger, "Relayer", config.pollingDelay, stop)) {
```
pxrl, line 68: Calling `processEndPollingLoop()` here has the disadvantage of imposing a delay of `pollingDelay` seconds. In testing I've seen that a HUP delivered during this delay results in one additional loop occurring.

I'll either propose an update to `processEndPollingLoop()` so that it doesn't impose the sleep, or will factor it out and handle the check directly in this file.

nicholaspai (approved):

james-a-morris (commented):
I like this PR - but we do have this polling logic in the finalizer, relayer, monitor, and dataworker. 

Thoughts on hoisting this to `./index.ts`? We could make `stop` be a reentrant variable that exists outside the context of specifically relayer so that we get this logic for free in the other bots

pxrl:
> I like this PR - but we do have this polling logic in the finalizer, relayer, monitor, and dataworker.
> 
> Thoughts on hoisting this to `./index.ts`? We could make `stop` be a reentrant variable that exists outside the context of specifically relayer so that we get this logic for free in the other bots

It could ultimately end up in the top-level index.ts, but in practice we don't run the finalizer, monitor or dataworker in looping mode, so I think it's preferable to just iterate locally within the relayer without having to worry about the other personalities, and find a solution that works well, and then have the option of promoting this to the top level.

james-a-morris (approved):

PR #1400, opened 2024-04-08
feat(InventoryManager): Consider origin chain when computing post-relay allocations

nicholaspai:
Simple change for now to start considering origin chain when considering post-relay allocations. The destination chain is preferred over origin chain but the latter is considered if (1) the origin chain token config is set and (2) the destination chain is over the target allocation

Fixes ACX-2040

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -200,8 +200,7 @@ export class InventoryClient {
     const { originChainId, destinationChainId, inputToken, outputToken, outputAmount } = deposit;
     const hubChainId = this.hubPoolClient.chainId;
 
-    // Always refund on L1 if the transfer is to L1.
-    if (!this.isInventoryManagementEnabled() || destinationChainId === hubChainId) {
+    if (!this.isInventoryManagementEnabled()) {
       return destinationChainId;
     }
 
@@ -217,6 +216,16 @@ export class InventoryClient {
       );
     }
     l1Token ??= this.hubPoolClient.getL1TokenForL2TokenAtBlock(outputToken, destinationChainId);
+    // If there is no inventory config for this token then take refund on destination chain.
+    // If neither destination chain nor origin chain have a configuration for this token, then take refund on the
+    // destination chain.
+    if (
+      this.inventoryConfig.tokenConfig?.[l1Token] === undefined ||
+      (this.inventoryConfig.tokenConfig[l1Token][destinationChainId] === undefined &&
+        this.inventoryConfig.tokenConfig[l1Token][originChainId] === undefined)
+    ) {
+      return destinationChainId;
+    }
 
     const startTime = Date.now();
     let totalRefundsPerChain: { [chainId: string]: BigNumber } = {};
@@ -245,17 +254,6 @@ export class InventoryClient {
       totalRefundsPerChain,
     });
     const cumulativeRefunds = Object.values(totalRefundsPerChain).reduce((acc, curr) => acc.add(curr), bnZero);
-
-    // If there is no inventory config for this token then take refund on destination chain.
-    // If neither destination chain nor origin chain have a configuration for this token, then take refund on the
-    // destination chain.
-    if (
-      this.inventoryConfig.tokenConfig?.[l1Token] === undefined ||
-      (this.inventoryConfig.tokenConfig[l1Token][destinationChainId] === undefined &&
-        this.inventoryConfig.tokenConfig[l1Token][originChainId] === undefined)
-    ) {
-      return destinationChainId;
-    }
     const cumulativeVirtualBalance = this.getCumulativeBalance(l1Token);
 
     // Prioritize destination chain repayment over origin chain repayment but prefer both over
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -255,11 +255,8 @@ export class InventoryClient {
     // If there is no inventory config for this token then take refund on destination chain.
     // If neither destination chain nor origin chain have a configuration for this token, then take refund on the
     // destination chain.
-    if (
-      this.inventoryConfig.tokenConfig?.[l1Token] === undefined ||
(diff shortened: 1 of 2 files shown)


src/clients/InventoryClient.ts
```diff
@@ -269,43 +271,77 @@ export class InventoryClient {
...
-      refundChainId,
-    });
-    // If the allocation is greater than the target then refund on L1. Else, refund on destination chain.
-    return refundChainId;
+    const cumulativeVirtualBalance = this.getCumulativeBalance(l1Token);
+
+    // Prioritize destination chain repayment over origin chain repayment but prefer both over
+    // hub chain repayment if they are under allocated.
+    const chainsToEvaluate = [destinationChainId];
+    // Only evaluate origin chain if its not the hub chain.
```
pxrl, line 280: Is this because of the implicit fallback to hubchain anyway?

src/clients/InventoryClient.ts
```diff
@@ -269,43 +271,77 @@ export class InventoryClient {
...
+
+      const targetPct = toBN(this.inventoryConfig.tokenConfig[l1Token][_chain].targetPct);
+      this.log(
+        `Evaluated taking repayment on ${
+          _chain === originChainId ? "origin" : "destination"
+        } chain ${_chain} for deposit ${deposit.depositId}: ${
+          expectedPostRelayAllocation.lte(targetPct) ? "UNDERALLOCATED ✅" : "OVERALLOCATED ❌"
+        }`,
+        {
+          l1Token: l1Token,
```
james-a-morris, line 322:
```suggestion
          l1Token,
```

src/clients/InventoryClient.ts
james-a-morris: I had this question too since we don't check if the destination chain is the hub chain when we add it to `chainsToEvaluate`

james-a-morris (commented):
One question about how why we unconditionally include the `destinationChainId` in `chainsToEvaluate` and two nits

src/clients/InventoryClient.ts
nicholaspai: yep

nicholaspai: i'll add a comment

test/InventoryClient.RefundChain.ts
```diff
@@ -150,140 +153,300 @@ describe("InventoryClient: Refund chain selection", async function () {
...
+        quoteTimestamp: hubPoolClient.currentTime!,
+        fillDeadline: 0,
+        exclusivityDeadline: 0,
+        exclusiveRelayer: ZERO_ADDRESS,
+      };
+    });
+    it("Correctly decides when to refund based on relay size", async function () {
+      // To start with, consider a simple case where the relayer is filling small relays. The current allocation on the
+      // target chain of Optimism for WETH is 20/140=14.2%. This is above the sum of targetL2Pct of 10% plus 2% buffer.
+      // Construct a small mock deposit of size 1 WETH. Post relay Optimism should have (20-1)/(140-1)=13.6%. This is still
```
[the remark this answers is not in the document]
nicholaspai: ah ok ill change

test/InventoryClient.RefundChain.ts
```diff
@@ -150,140 +153,300 @@ describe("InventoryClient: Refund chain selection", async function () {
...
+      sampleDepositData = {
+        depositId: 0,
+        originChainId: 1,
+        destinationChainId: 10,
+        depositor: owner.address,
+        recipient: owner.address,
+        inputToken: mainnetWeth,
+        inputAmount,
+        outputToken: l2TokensForWeth[10],
+        outputAmount: inputAmount.mul(fixedPoint.sub(relayerFeePct)).div(fixedPoint),
```
pxrl, line 168: The `computeOutputAmount()` utility is also subtracting `realizedLpFeePct` from `inputAmount`. It might not matter since this is only the top-level template.

nicholaspai: yeah i'm hesitant to change this since it would change all tests but I do think we can probably ignore fees i this test to make numbers cleaner

test/InventoryClient.RefundChain.ts
```diff
@@ -150,140 +153,300 @@ describe("InventoryClient: Refund chain selection", async function () {
...
   });
 
-  it("Correctly throws when Deposit tokens are not equivalent", async function () {
-    sampleDepositData.inputAmount = toWei(5);
-    sampleDepositData.outputAmount = await computeOutputAmount(sampleDepositData);
-    expect(await inventoryClient.determineRefundChainId(sampleDepositData)).to.equal(
-      sampleDepositData.destinationChainId
-    );
+  describe("origin chain is not equal to hub chain", function () {
+    it("Both origin and destination chain allocations are below target", async function () {
```
pxrl, line 336: What's the prevailing origin chain for the deposit in this test? I don't see it being set here, so it seems like it'd still be mainnet?

test/InventoryClient.RefundChain.ts
pxrl: Yeah it might be nicer to just assume 0 fees so then input and output amounts match.

test/InventoryClient.RefundChain.ts
nicholaspai: it was mainnet but I've changed it to 137 at the top-level because I think origin being mainnet is a special case

nicholaspai: I'll clean it up to make it easier to understand b694445c2f1a6d96ed7299831d161f4c8ea97c55

test/InventoryClient.RefundChain.ts
nicholaspai: Yeah I just went ahead and fixed this since its annoying a603b2324485752f9ade59b5918b5fd91b8b1542

pxrl (approved):
LGTM! Will be interesting to see how this impacts utilisation.

PR #1401, opened 2024-04-09
refactor(relayer): Drop redundant unfilled deposit lookback filtering

pxrl:
The `getUnfilledDeposits()` function invokes a BlockFinder instance on each chain and proceeds to resolve the lookback time delta to a block. It can be expensive to do this on chains where the average block times are not super consistent, so it sometimes causes longer-than-necessary delays.

All deposits sourced from each origin SpokePool subsequently have their fill status verified via the destination SpokePool contract, so the marginal cost for including additional deposits is quite small.

Note also that deposit lookbacks are still enforced on the SpokePoolClient search config, so looping mode is the only case where this filtering would have been effective. The current direction of looping mode is however to completely refresh the SpokePoolClient on each update, so there shouldn't be any concerns about old deposits appearing repeatedly.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -177,11 +177,7 @@ export class Monitor {
   }
 
   async reportUnfilledDeposits(): Promise<void> {
-    const unfilledDeposits = await getUnfilledDeposits(
-      this.clients.spokePoolClients,
-      this.clients.hubPoolClient,
-      this.monitorConfig.maxRelayerLookBack
-    );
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):
This does seem redundant. Any idea why this filter even exists today?

pxrl:
> This does seem redundant. Any idea why this filter even exists today?

My best guess is that it was for looping mode, where the array of deposits would just grow indefinitely over time.

PR #1403, opened 2024-04-11
improve(Relayer): Allow user to skip early exit for "un-finalized" deposits if in simulation mode

nicholaspai:
I  manually comment out this line a lot when testing the relayer locally because I want to see how the relayer would evaluate a 
deposit's 
profitability and pick a repayment chain without sending a deposit.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -275,7 +275,7 @@ export class Relayer {
       });
       // If we're in simulation mode, skip this early exit so that the user can evaluate
       // the full simulation run.
-      if (this.config.sendingRelaysEnabled)  {
+      if (this.config.sendingRelaysEnabled) {
         return;
       }
     }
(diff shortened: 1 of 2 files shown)


src/relayer/Relayer.ts
```diff
@@ -273,7 +273,11 @@ export class Relayer {
         maxBlockNumber,
         transactionHash: deposit.transactionHash,
       });
-      return;
+      // If we're in simulation mode, skip this early exit so that the user can evaluate
+      // the full simulation run.
+      if (this.config.sendingRelaysEnabled) {
```
pxrl, line 278: Should this be negated?

```suggestion
      if (!this.config.sendingRelaysEnabled) {
```

james-a-morris (approved):
I agree with this change. +1 on the DX

mrice32 (approved):
Only downside is that it's hard to simulate the MCD logic, but I think you can just comment this line out for that test case.

PR #1404, opened 2024-04-11
refactor: Remove top-level bot loop (take two)

pxrl:
This change was previously merged and backed out shortly after because it was seen to result in some bot instances hanging. This is almost certainly because the original commit removed two instances of process.exit(), and replaced them with prococess.exitCode = <X>. There appear to be some instances where the bot can have some dangling promises that take a long time to resolve, or that don't resolve for some reason. Replaying the change, but without removing the process.exit() calls should be OK.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -57,11 +57,10 @@ if (require.main === module) {
 
   const args = minimist(process.argv.slice(2), opts);
 
+  let exitCode = 0;
   run(args)
-    .then(() => {
-      exit(0);
-    })
     .catch(async (error) => {
+      exitCode = 1;
       logger.error({
         at: cmd ?? "unknown process",
         message: "There was an execution error!",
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -80,17 +59,19 @@ if (require.main === module) {
 
   run(args)
     .then(() => {
-      // eslint-disable-next-line no-process-exit
-      process.exit(0);
+      exit(0);
```
james-a-morris, line 62: Thoughts on having an `exitCode` variable with a single `exit()` in the `.finally()` block?

```
let exitCode = 0;
run(args).
  .catch(() => {
      exitCode = 1;
      ...{logging}
  })
  .finally(() => {
     exit(exitCode);
  });
```

james-a-morris (commented):
One nit on the structure

index.ts
pxrl: Good suggestion - implemented: https://github.com/across-protocol/relayer-v2/pull/1404/commits/d2f238f754cb26638c1a0beccf634df4a156f2f3

james-a-morris (approved):
LGTM pending the linter fix

PR #1411, opened 2024-04-15
feat(InventoryClient): Take repayment on chains that are over spoke target

nicholaspai:
This algorithm is based on discussions with Korpi.

The current algorithm for determining refund chain is:
1. if destination relayer balance is under target, take repayment on destination
2. Else if origin relayer balance is under target, take repayment on origin
3. Else mainnet

The proposed changes to the algorithm add two new rules above 1:
1. Sort all 7day chains (i.e. ORU's with 7 day withdrawal liveness) by the amount that their "latest running balances" are over their `spokePoolTarget` and evaluate them in this order. Remove any chains where the excess is < 0.
2. For each of these chains, if the relayer balance is under allocated on this chain, take repayment there
3. If there are no more 7 day chains, then fallback to existing algorithm where we evaluate destination chain and then origin chain relayer balance allocation

The complexity of this PR is computing "latest running balances". The algorithm I'm using is:
- Find running balance for token and chain using latest proposed root bundle containing token and chain
- Add all deposited `inputAmounts` to running balances that occurred after the latest proposal's end block
- Subtract all refunds `inputAmounts` from running balances that occurred after the latest proposal's end block

The last bullet point is obviously risky since it opens the relayer up to being griefed and tricked into taking repayment on the chain if someone sends an invalid fill with a large `inputAmount` on a chain with cheap gas. This is pretty easy to do though I'm not sure why someone one would do it. The fix is a change to the algorithm in `getApproximateRefundsForBlockRange` to match fills and deposits on `inputAmount` and `outputAmount` in addition to originChain and depositId. It still allows invalid fills to be sent but its a good approximation for now

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -19,6 +19,7 @@ import {
   assert,
   fixedPointAdjustment,
   isDefined,
+  toBN,
 } from "../utils";
 import { Clients } from "../common";
 import {
@@ -281,6 +282,9 @@ export class BundleDataClient {
   }
 
   // @dev This helper function should probably be moved to the InventoryClient
+  // @dev This function does not perfectly match fills and deposits. Therefore this calculation can be
+  // griefed by someone sending invalid fills. We compare fill and deposit input and output amount along with
+  // origin and destination chain and deposit ID. This should prevent most practical griefing attacks.
   getApproximateRefundsForBlockRange(
     chainIds: number[],
     blockRanges: number[][],
@@ -294,12 +298,29 @@ export class BundleDataClient {
       const chainIndex = chainIds.indexOf(chainId);
       this.spokePoolClients[chainId]
         .getFills()
-        .filter(
-          (fill) =>
-            filteredRefundAddresses.includes(fill.relayer) &&
+        .filter((fill) => {
+          if (filteredRefundAddresses.length > 0 && !filteredRefundAddresses.includes(fill.relayer)) {
+            return false;
+          }
+          // We match fill and deposit on originChain - depositId - inputAmount so even if a griefer fakes the fill
+          // they will at least have paid the same `outputAmount` as the real deposit, which is going to be
+          // an economic deterrance if the griefer wants to materially affect this refund computation. This is because
+          // the inputAmount must also be large if the outputAmount is going to be large. This seems like a reasonable
+          // compromise between security and performance.
+
+          // If origin spoke pool client isn't defined, we can't validate it.
+          if (this.spokePoolClients[fill.originChainId] === undefined) {
+            return false;
+          }
+          const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
+          const hasMatchingDeposit =
+            matchingDeposit?.inputAmount.eq(fill.inputAmount) && matchingDeposit?.outputAmount.eq(fill.outputAmount);
+          return (
             fill.blockNumber >= blockRanges[chainIndex][0] &&
-            fill.blockNumber <= blockRanges[chainIndex][1]
-        )
+            fill.blockNumber <= blockRanges[chainIndex][1] &&
+            hasMatchingDeposit
+          );
+        })
         .forEach((fill) => {
           const { chainToSendRefundTo, repaymentToken } = getRefundInformationFromFill(
             fill,
@@ -319,6 +340,35 @@ export class BundleDataClient {
         });
     }
     return refundsForChain;
+  }
+
+  getUpcomingDepositAmount(chainId: number, l2Token: string, latestBlockToSearch: number): BigNumber {
+    if (this.spokePoolClients[chainId] === undefined) {
+      return toBN(0);
+    }
+    return this.spokePoolClients[chainId]
+      .getDeposits()
+      .filter((deposit) => deposit.blockNumber > latestBlockToSearch && deposit.inputToken === l2Token)
+      .reduce((acc, deposit) => {
+        return acc.add(deposit.inputAmount);
+      }, toBN(0));
+  }
+
+  async getLatestProposedBundleData(): Promise<{ bundleData: LoadDataReturnValue; blockRanges: number[][] }> {
+    const hubPoolClient = this.clients.hubPoolClient;
+    // If there is pending bundle, return its bundle data from arweave, otherwise return the latest
+    // executed bundle.
+    const bundleBlockRanges = getImpliedBundleBlockRanges(
+      hubPoolClient,
+      this.clients.configStoreClient,
+      hubPoolClient.hasPendingProposal()
+        ? hubPoolClient.getLatestProposedRootBundle()
+        : this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(this.clients.hubPoolClient.latestBlockSearched)
+    );
+    return {
+      blockRanges: bundleBlockRanges,
+      bundleData: await this.loadArweaveData(bundleBlockRanges),
+    };
   }
 
   // @dev This function should probably be moved to the InventoryClient since it bypasses loadData completely now.
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -15,13 +15,15 @@ import {
   AnyObject,
   ERC20,
   TOKEN_SYMBOLS_MAP,
+  formatFeePct,
+  fixedPointAdjustment,
 } from "../utils";
-import { HubPoolClient, TokenClient, BundleDataClient } from ".";
+import { HubPoolClient, TokenClient, BundleDataClient, ConfigStoreClient } from ".";
 import { AdapterManager, CrossChainTransferClient } from "./bridges";
 import { InventoryConfig, V3Deposit } from "../interfaces";
 import lodash from "lodash";
 import { CONTRACT_ADDRESSES } from "../common";
-import { CombinedRefunds } from "../dataworker/DataworkerUtils";
+import { CombinedRefunds, _buildPoolRebalanceRoot } from "../dataworker/DataworkerUtils";
 
 type TokenDistributionPerL1Token = { [l1Token: string]: { [chainId: number]: BigNumber } };
 
@@ -38,11 +40,16 @@ export type Rebalance = {
 
 const { CHAIN_IDs } = constants;
 
+// Deciding not to add to constants because this chain list is specific to this client which has to care about
+// which chains have slow canonical L2-->L1 bridges.
+const SLOW_WITHDRAWAL_CHAINS = [CHAIN_IDs.BASE, CHAIN_IDs.ARBITRUM, CHAIN_IDs.OPTIMISM];
+
 export class InventoryClient {
   private logDisabledManagement = false;
   private readonly scalar: BigNumber;
   private readonly formatWei: ReturnType<typeof createFormatFunction>;
   private bundleRefundsPromise: Promise<CombinedRefunds[]> = undefined;
+  private excessRunningBalanceCache: { [l1Token: string]: { [chainId: number]: BigNumber } } = {};
 
   constructor(
     readonly relayer: string,
@@ -54,7 +61,8 @@ export class InventoryClient {
     readonly bundleDataClient: BundleDataClient,
     readonly adapterManager: AdapterManager,
     readonly crossChainTransferClient: CrossChainTransferClient,
-    readonly simMode = false
+    readonly simMode = false,
+    readonly prioritizeLpUtilization = true
   ) {
     this.scalar = sdkUtils.fixedPointAdjustment;
     this.formatWei = createFormatFunction(2, 4, false, 18);
@@ -158,8 +166,8 @@ export class InventoryClient {
   async getAllBundleRefunds(): Promise<CombinedRefunds[]> {
     const refunds: CombinedRefunds[] = [];
     const [pendingRefunds, nextBundleRefunds] = await Promise.all([
-      this.bundleDataClient.getPendingRefundsFromValidBundles(this.relayer),
-      this.bundleDataClient.getNextBundleRefunds(this.relayer),
+      this.bundleDataClient.getPendingRefundsFromValidBundles(),
+      this.bundleDataClient.getNextBundleRefunds(),
     ]);
     refunds.push(...pendingRefunds, ...nextBundleRefunds);
     this.logger.debug({
@@ -254,7 +262,8 @@ export class InventoryClient {
     l1Token ??= this.hubPoolClient.getL1TokenForL2TokenAtBlock(outputToken, destinationChainId);
     // If there is no inventory config for this token then take refund on destination chain.
     // If neither destination chain nor origin chain have a configuration for this token, then take refund on the
-    // destination chain.
+    // destination chain. This maintains the relayer balance allocation at the risk of increasing utilization. In
+    // production, this should never be undefined.
     const tokenConfig = this.inventoryConfig?.tokenConfig?.[l1Token];
     if (tokenConfig?.[destinationChainId] === undefined && tokenConfig?.[originChainId] === undefined) {
       return destinationChainId;
@@ -271,12 +280,33 @@ export class InventoryClient {
     const cumulativeRefunds = Object.values(totalRefundsPerChain).reduce((acc, curr) => acc.add(curr), bnZero);
     const cumulativeVirtualBalance = this.getCumulativeBalance(l1Token);
 
+    // Build list of chains we want to evaluate for repayment:
+    const chainsToEvaluate = [];
+    // Add optimistic rollups to front of evaluation list because these are chains with long withdrawal periods
+    // that we want to prioritize taking repayment on if the chain is going to end up sending funds back to the
+    // hub in the next root bundle over the slow canonical bridge.
+    // We need to calculate the latest running balance for each optimistic rollup chain.
+    // We'll add the last proposed running balance plus new deposits and refunds.
+    if (this.prioritizeLpUtilization) {
+      const excessRunningBalances = this.excessRunningBalanceCache[l1Token]
+        ? this.excessRunningBalanceCache[l1Token]
+        : await this.getExcessRunningBalances(l1Token);
+      // Sort chains by highest excess percentage over the spoke target, so we can prioritize
+      // taking repayment on chains with the most excess balance.
+      const chainsWithExcessSpokeBalances = SLOW_WITHDRAWAL_CHAINS.filter((chainId) =>
+        excessRunningBalances[chainId].gt(0)
+      ).sort((a, b) => excessRunningBalances[b].sub(excessRunningBalances[a]).toNumber());
+      chainsToEvaluate.push(...chainsWithExcessSpokeBalances);
+    }
+    // Add destination and origin chain if they are not already added.
     // Prioritize destination chain repayment over origin chain repayment but prefer both over
     // hub chain repayment if they are under allocated. We don't include hub chain
     // since its the fallback chain if both destination and origin chain are over allocated.
-    const chainsToEvaluate = [destinationChainId];
-    // Only evaluate origin chain if its not the hub chain.
-    if (originChainId !== hubChainId) {
+    // If destination chain is hub chain, we still want to evaluate it before the origin chain.
+    if (!chainsToEvaluate.includes(destinationChainId)) {
+      chainsToEvaluate.push(destinationChainId);
+    }
+    if (!chainsToEvaluate.includes(originChainId) && originChainId !== hubChainId) {
       chainsToEvaluate.push(originChainId);
     }
     for (const _chain of chainsToEvaluate) {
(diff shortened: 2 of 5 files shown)


src/clients/InventoryClient.ts
```diff
@@ -271,12 +285,35 @@ export class InventoryClient {
...
+    const chainsToEvaluate = [];
+    // Add optimistic rollups to front of evaluation list because these are chains with long withdrawal periods
+    // that we want to prioritize taking repayment on if the chain is going to end up sending funds back to the
+    // hub in the next root bundle over the slow canonical bridge.
+    // We need to calculate the latest running balance for each optimistic rollup chain.
+    // We'll add the last proposed running balance plus new deposits and refunds.
+    if (this.prioritizeLpUtilization) {
+      if (!isDefined(this.excessRunningBalancePromises[l1Token])) {
+        // @dev Save this as a promise so that other parallel calls to this function don't make the same call.
+        this.excessRunningBalancePromises[l1Token] = this.getExcessRunningBalances(l1Token, inputAmount);
```
mrice32, line 298: We shouldn't reuse this value since the function takes in the inputAmount for this particular deposit, right?

Can we just have this function compute the running balance and then add in the input amount outside of this call?

src/clients/InventoryClient.ts
```diff
@@ -271,12 +285,35 @@ export class InventoryClient {
...
+      }
+      const excessRunningBalances = lodash.cloneDeep(await this.excessRunningBalancePromises[l1Token]);
+      // Sort chains by highest excess percentage over the spoke target, so we can prioritize
+      // taking repayment on chains with the most excess balance.
+      const chainsWithExcessSpokeBalances = SLOW_WITHDRAWAL_CHAINS.filter((chainId) =>
+        excessRunningBalances[chainId].gt(0)
+      ).sort((a, b) => (excessRunningBalances[b].gt(excessRunningBalances[a]) ? 1 : -1));
```
mrice32, line 305: Sort has to return 0 if the values are equal so that it's consistent.

We should probably have some utility functions that are just called `bnSort` or similar:
```ts
function bnComparatorDescending(a: BigNumber, b: BigNumber) {
  if (b.gt(a)) return 1;
  else if (a.gt(b)) -1;
  else return 0;
}

function bnComparatorAscending(a, b) {
  if (a.gt(b)) return 1;
  else if (b.gt(a)) -1;
  else return 0;
}
```

Then you can use them here as such:
```ts
sort((a, b) => bnComparatorDescending(excessRunningBalances[a], excessRunningBalances[b]))
```

Reference:

> More formally, the comparator is expected to have the following properties, in order to ensure proper sort behavior:
> - Reflexive: compareFn(a, a) === 0.
> - Anti-symmetric: compareFn(a, b) and compareFn(b, a) must both be 0 or have opposite signs.


From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

src/clients/InventoryClient.ts
```diff
@@ -327,21 +367,108 @@ export class InventoryClient {
...
+        // - plus total refund amount on chain since the latest end block proposed
+        const upcomingDeposits = this.bundleDataClient.getUpcomingDepositAmount(
+          chainId,
+          this.getDestinationTokenForL1Token(l1Token, chainId),
+          blockRange[1]
+        );
+        const allBundleRefunds = lodash.cloneDeep(await this.bundleRefundsPromise);
+        const upcomingRefunds = allBundleRefunds.pop(); // @dev upcoming refunds are always pushed last into this list.
+        const upcomingRefundForChain = Object.values(
+          upcomingRefunds?.[chainId]?.[this.getDestinationTokenForL1Token(l1Token, chainId)] ?? {}
```
mrice32, line 424: When do we expect these to be undefined?

src/clients/InventoryClient.ts
```diff
@@ -327,21 +367,108 @@ export class InventoryClient {
...
+        const targetSpokeBalanceForChain = this.hubPoolClient.configStoreClient.getSpokeTargetBalancesForBlock(
+          l1Token,
+          chainId
+        );
+        // A negative running balance means that the spoke has a balance, so we should compare the absolute value
+        // of the negative running balance with the target. If the running balance is positive, then the hub
+        // owes it funds and its below target so we don't want to take additional repayment.
+        const absLatestRunningBalance = latestRunningBalance.abs();
+        const pctOverTarget = latestRunningBalance.lt(0)
+          ? absLatestRunningBalance.gt(targetSpokeBalanceForChain.target)
```
mrice32, line 442: This is a little tricky to read. Can we && these first two conditions since if either are false we return 0?

src/clients/InventoryClient.ts
```diff
@@ -327,21 +367,108 @@ export class InventoryClient {
...
+            ? targetSpokeBalanceForChain.target.gt(0)
```
mrice32, line 443: So this says that if the target is 0, we just say that we're 100% over target? Am I reading that right? How was that number picked? I think we should definitely put a comment on the function that explains all the edge cases.

mrice32 (commented):
Looks good! A few comments, mostly minor

src/clients/BundleDataClient.ts
```diff
@@ -281,10 +284,13 @@ export class BundleDataClient {
   }
 
   // @dev This helper function should probably be moved to the InventoryClient
+  // @dev This function does not perfectly match fills and deposits. Therefore this calculation can be
```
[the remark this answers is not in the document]
nicholaspai: you're right, the slow down is negligible (<50ms)

src/clients/InventoryClient.ts
nicholaspai: ah yes good call

src/clients/InventoryClient.ts
```diff
@@ -38,11 +41,16 @@ export type Rebalance = {
 
 const { CHAIN_IDs } = constants;
 
+// Deciding not to add to constants because this chain list is specific to this client which has to care about
+// which chains have slow canonical L2-->L1 bridges.
+const SLOW_WITHDRAWAL_CHAINS = [CHAIN_IDs.BASE, CHAIN_IDs.ARBITRUM, CHAIN_IDs.OPTIMISM];
```
pxrl, line 46: In the current PR this is also added in constants fwiw.

src/clients/InventoryClient.ts
nicholaspai: Yeah good call out, I thought equaling to 0 would produce non determinism which I was fine with in this PR but its a good call out. I've added the utility method

src/clients/BundleDataClient.ts
```diff
@@ -294,18 +300,36 @@ export class BundleDataClient {
...
-        )
+        .filter((fill) => {
+          if (fill.blockNumber < blockRanges[chainIndex][0] || fill.blockNumber > blockRanges[chainIndex][1]) {
+            return false;
+          }
+
+          // If the fill was sent by one of the whitelisted addresses, then we can just assume its valid.
+          if (whitelistedRelayerAddresses.includes(fill.relayer)) {
+            return true;
+          }
```
pxrl, line 311: ooc, how would this list of addresses be maintained? Manual update based on analytics? It seems like it could become difficult to stay on top of.

pxrl (commented):
@nicholaspai Have you managed to estimate any performance impact for this? I had been hoping we'd be able to drop the BundleDataClient from the relayer entirely by some refactoring to permit the relayer to source Arweave data independently, and then approximating the pending repayments based on recent fills made. I'm wary of some poor worst-case performance when invoking complex repayment calculations and I think it could hurt our ability to be performant at the 10s fill level. I'm starting to wonder if this stuff is just better handled by an external backend that can be queried on the fly from the relayer, rather than burdening the relayer with having to store and compute all this data.

src/clients/InventoryClient.ts
nicholaspai: If a chain didn't exist in the last bundle of a spoke poool client isn't defined, for example for 59144 then we skip it here https://github.com/across-protocol/relayer-v3/blob/c438249d6ef3b58ba3d9a23d9ce03126395f512d/src/clients/BundleDataClient.ts#L298

nicholaspai:
> @nicholaspai Have you managed to estimate any performance impact for this? I had been hoping we'd be able to drop the BundleDataClient from the relayer entirely by some refactoring to permit the relayer to source Arweave data independently, and then approximating the pending repayments based on recent fills made. I'm wary of some poor worst-case performance when invoking complex repayment calculations and I think it could hurt our ability to be performant at the 10s fill level. I'm starting to wonder if this stuff is just better handled by an external backend that can be queried on the fly from the relayer, rather than burdening the relayer with having to store and compute all this data.

its very fast when running locally with gckms. It adds <1 s of run-time to approximate upcoming refunds. The calls to arweave are done in parallel and take ~1s total also

src/clients/BundleDataClient.ts
nicholaspai: I think its just based on the caller. The caller should always pass in their own address as they could test themselves. It just helps them optimize a bit and early exit for their filled relays.

src/relayer/Relayer.ts
```diff
@@ -570,7 +570,20 @@ export class Relayer {
...
+      hubPoolToken.address
+    );
+    this.logger.debug({
+      at: "Relayer::resolveRepaymentChain",
+      message: `Determined preferred repayment chain ${preferredChainId} for deposit from ${originChain} to ${destinationChain} in ${
+        Math.round(performance.now() - start) / 1000
+      }s.`,
+    });
     const { lpFeePct } = repaymentFees.find(({ paymentChainId }) => paymentChainId === preferredChainId);
     assert(isDefined(lpFeePct));
```
pxrl, line 588: Now that we're computing fees for predetermined chain Ids, we'll need to ensure that `preferredChainId` is actually in that set. This means a change up at line 389. In order to keep it simple I wonder if we should just compute the fee over all supported chains.

Something like this should do it:

```
diff --git a/src/relayer/Relayer.ts b/src/relayer/Relayer.ts
index c8c1783c..0f268e44 100644
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -390,6 +390,12 @@ export class Relayer {
           { ...deposit, paymentChainId: originChainId },
         ];
 
+        if (this.clients.inventoryClient.isInventoryManagementEnabled()) {
+          SLOW_WITHDRAWAL_CHAINS
+            .filter((chainId) => [originChainId, destinationChainId].includes(chainId))
+            .forEach((paymentChainId) => request.push({ ...deposit, paymentChainId }));
+        }
+
         // Optionally also query for HubPool chain repayment if it's not origin or destination.
         if (![originChainId, destinationChainId].includes(hubPoolClient.chainId)) {
           request.push({ ...deposit, paymentChainId: hubPoolClient.chainId });
```

src/clients/BundleDataClient.ts
```diff
@@ -321,13 +343,64 @@ export class BundleDataClient {
...
+  async getLatestProposedBundleData(): Promise<{ bundleData: LoadDataReturnValue; blockRanges: number[][] }> {
+    const hubPoolClient = this.clients.hubPoolClient;
+    // If there is pending bundle, return its bundle data from arweave, otherwise return the latest
+    // executed bundle.
+    const bundleBlockRanges = getImpliedBundleBlockRanges(
+      hubPoolClient,
+      this.clients.configStoreClient,
+      hubPoolClient.hasPendingProposal()
+        ? hubPoolClient.getLatestProposedRootBundle()
+        : this.clients.hubPoolClient.getLatestFullyExecutedRootBundle(this.clients.hubPoolClient.latestBlockSearched)
```
pxrl, line 367:
```suggestion
        : hubPoolClient.getLatestFullyExecutedRootBundle(this.clients.hubPoolClient.latestBlockSearched)
```

src/clients/BundleDataClient.ts
```diff
@@ -321,13 +343,64 @@ export class BundleDataClient {
...
+      .filter((deposit) => deposit.blockNumber > latestBlockToSearch && deposit.inputToken === l2Token)
+      .reduce((acc, deposit) => {
+        return acc.add(deposit.inputAmount);
+      }, toBN(0));
+  }
+
+  async getLatestProposedBundleData(): Promise<{ bundleData: LoadDataReturnValue; blockRanges: number[][] }> {
+    const hubPoolClient = this.clients.hubPoolClient;
+    // If there is pending bundle, return its bundle data from arweave, otherwise return the latest
+    // executed bundle.
```
pxrl, line 361: Is this comment correct? bundleData seems to be always loaded via `this.loadArweaveData()`.

src/clients/BundleDataClient.ts
pxrl: Fair enough :+1:  the wording just gives the impression that it's intended to contain multiple addresses.

src/relayer/Relayer.ts
```diff
@@ -570,7 +570,20 @@ export class Relayer {
     const originChain = getNetworkName(originChainId);
     const destinationChain = getNetworkName(destinationChainId);
 
-    const preferredChainId = await inventoryClient.determineRefundChainId(deposit, hubPoolToken.address);
+    const start = performance.now();
+    // Allow relayer to have up to a 50% excess of balance on the refund chain over the configured target.
+    const relayerTargetMultiplier = toBNWei("1.5");
```
pxrl, line 575: wdyt about making this configurable somehow? idk if we have third-parties using inventory management, but if so then this could be a big turn-off for them.

src/utils/BNUtils.ts
```diff
@@ -0,0 +1,21 @@
+import { BigNumber } from ".";
```
pxrl, line 1: I'm uncertain about how circular dependencies are handled, but I think this _might_ risk creating one. For a file named `BNUtils` we might be better re-exporting BigNumber from here in any case.

src/clients/InventoryClient.ts
```diff
@@ -54,7 +60,8 @@ export class InventoryClient {
     readonly bundleDataClient: BundleDataClient,
     readonly adapterManager: AdapterManager,
     readonly crossChainTransferClient: CrossChainTransferClient,
-    readonly simMode = false
+    readonly simMode = false,
+    readonly prioritizeLpUtilization = true
```
pxrl, line 64: Same as above RE making this configurable - it looks like it's implicitly impossible to opt out of since we seem to rely on the default values here.

src/clients/InventoryClient.ts
```diff
@@ -327,21 +368,192 @@ export class InventoryClient {
...
+          return [
+            chainId,
+            {
+              pct: toBN(0),
+              target,
+              excess,
+              excessPostRelay,
+            },
+          ];
+        }
```
pxrl, line 518: Prettier has made this really sparse and imo hard to read. wdyt about factoring it out somehow so it doesn't get spread apart like this? Alternatively, it seems like only `pct` is varying in each case, so maybe there's a better way to refactor it to avoid all the duplication.

src/relayer/Relayer.ts
nicholaspai: ah yes this is a good call!

nicholaspai: a35993f143225f92e7903f931b4986c1d04f31c7

src/clients/BundleDataClient.ts
nicholaspai: ah good call, edited in this commit which makes it more clear that we should fall back to loadData if arweave data is missing for some reason d448f6159c8eea047c1e630c1960a0ea93c97ae0

src/clients/InventoryClient.ts
nicholaspai: isn't this simple to opt out of since you can set the value in the constructor? this is how I bypass this step in the tests

src/relayer/Relayer.ts
nicholaspai: 273951f511fe5b987514f82bdda4d3c5de043352

src/clients/InventoryClient.ts
nicholaspai: good call: 044835a0bc691a0f75cba5374bf99d7acb5cbe93

test/InventoryClient.RefundChain.ts
```diff
@@ -280,7 +289,7 @@ describe("InventoryClient: Refund chain selection", async function () {
     });
 
     it("token config is not defined", async function () {
-      // Defaults to destination chain
+      // Defaults to mainnet.
```
nicholaspai, line 292: @pxrl @mrice32 FYI what do you think about changing the default repayment chain if there is no `tokenConfig` defined to mainnet? This was a choice I made motivated by simplicity of code, since `determineRefundChain`  currently falls back to mainnet, but it could just as easily be `destinationChain` if no other chains are underallocated , or if the token config is not defined. I think falling back to `mainnet` doesn't hurt utilization more than `destinationChain` does and it seems easier to rebalance out of

src/clients/InventoryClient.ts
mrice32: Makes sense. I originally just assumed the ?s were unnecessary, but I mistread.

src/clients/InventoryClient.ts
```diff
@@ -327,21 +359,166 @@ export class InventoryClient {
...
+        const returnObj = {
+          pct: toBN(0),
+          target,
+          excess,
+          excessPostRelay,
+        };
+        // If target is greater than excess running balance, then pct will
+        // be set to 0. If target is 0 then pct is infinite.
+        if (target.lte(0)) {
+          returnObj.pct = toBN(MAX_UINT_VAL);
```
mrice32, line 488: nit: I think this already is a BigNumber, right?

```ts
returnObj.pct = MAX_UINT_VAL;
```

src/clients/InventoryClient.ts
```diff
@@ -327,21 +359,166 @@ export class InventoryClient {
...
+    });
+    return Object.fromEntries(Object.entries(pcts).map(([k, v]) => [k, v.pct]));
+  }
+
+  async getExcessRunningBalancePcts(
+    l1Token: string,
+    refundAmount: BigNumber
+  ): Promise<{ [chainId: number]: BigNumber }> {
+    if (!isDefined(this.excessRunningBalancePromises[l1Token])) {
+      // @dev Save this as a promise so that other parallel calls to this function don't make the same call.
```
mrice32, line 515: +1

src/clients/InventoryClient.ts
```diff
@@ -327,21 +359,166 @@ export class InventoryClient {
...
+      this.excessRunningBalancePromises[l1Token] = this.getLatestRunningBalances(l1Token);
+    }
+    const excessRunningBalances = lodash.cloneDeep(await this.excessRunningBalancePromises[l1Token]);
+    return await this._getExcessRunningBalancePcts(excessRunningBalances, l1Token, refundAmount);
```
mrice32, line 519: nit: I don't think this function returns a promise or is async
```suggestion
    return this._getExcessRunningBalancePcts(excessRunningBalances, l1Token, refundAmount);
```

src/clients/InventoryClient.ts
```diff
@@ -327,21 +359,166 @@ export class InventoryClient {
...
+        // Approximate latest running balance as last known proposed running balance...
+        // - minus total deposit amount on chain since the latest end block proposed
+        // - plus total refund amount on chain since the latest end block proposed
+        const upcomingDeposits = this.bundleDataClient.getUpcomingDepositAmount(
+          chainId,
+          this.getDestinationTokenForL1Token(l1Token, chainId),
+          blockRange[1]
+        );
+        const allBundleRefunds = lodash.cloneDeep(await this.bundleRefundsPromise);
+        const upcomingRefunds = allBundleRefunds.pop(); // @dev upcoming refunds are always pushed last into this list.
```
mrice32, line 422: This is probably correct, but the structure is a little confusing.

OOC, what are the other elements of the array and why do we only care about the last set of refunds in the list?

mrice32 (approved):
Looks great! Mostly minor questions and comments

test/InventoryClient.RefundChain.ts
mrice32: Just so I understand, this is the case where the relayer specifies no targets or rebalancing behavior at all, right? Wouldn't this just mean that all of their funds would end up on mainnet no matter where they started?

I think destination chain is what most relayers would want if they specify no rebalancing or target balances, so their tokens stay where they are and they do rebalancing manually. Without this default, there would be no way to tell the bot to just leave your funds where they are.

Thoughts?

src/clients/InventoryClient.ts
nicholaspai: The last one is always the "upcoming" refunds. The other ones are (1) refunds in last executed bundle and (2) refunds in proposed pending bundle. I will try to make this more clear

test/InventoryClient.RefundChain.ts
nicholaspai: > Just so I understand, this is the case where the relayer specifies no targets or rebalancing behavior at all, right? Wouldn't this just mean that all of their funds would end up on mainnet no matter where they started?

Yes. Currently, if the relayer specifies no inventory config at all, all repayments are taken on destination chain, which I don't think is the same as saying all funds end up on mainnet no matter where they started.

nicholaspai: > I think destination chain is what most relayers would want if they specify no rebalancing or target balances, so their tokens stay where they are and they do rebalancing manually. Without this default, there would be no way to tell the bot to just leave your funds where they are.

Yeah that's reasonable too, i'll revert

nicholaspai: f695584b813ac870aa611508eeb45b349dab499b

PR #1413, opened 2024-04-15
feat(relayer): Externally-indexed SpokePoolClient

pxrl:
This commit introduces a new relayer-specific SpokePoolClient that's suited to running with a non-zero POLLING_DELAY. The benefits of this change include:
- Significantly faster relayer loop times.
- Relayer performance is decoupled from RPC provider performance.
- Additional robustness to RPC provider errors & misbehaviour.
- Increased RPC speed & efficiency via WebSocket event subscription.
- Increased protection against re-orgs in looping mode.

In this mode, the relayer forks off one indexer process per chain. Each indexer is responsible for indexing events based on a configurable lookback, and subsequently subscribes to the required set of events moving forward. Indexers supply event updates to the relayer after each block update is emitted by the primary WebSocket provider; these events are then retained in the SpokePoolClient to be used on the next SpokePoolClient.update() call.

Because WebSocket providers are used for event subscription, the existing relayer ProviderUtils providers are not used. Instead, basic quorum & finality checks are applied by the indexer before relaying the events to the relayer. There is still a possibility that reorgs occur after events have been relayed to the relayer, so whenever a removal notification is received by the indexer process it is immediately relayed to the relayer so it can remove any stale events as soon as is practical.

With this change it's possible to configure the relayer to be very responsive to new deposit events. In testing it's been possible to make fills reliably with only a minority of collisions. The sample size is small due to a lack of test funds in the dev wallet, so deposit confirmation tuning will be needed to hit the sweet spot of fast fills that don't collide with the very-fast third-party relayers.

This PR can be tested by adjusting the following config items:

- Set `POLLING_DELAY` to a positive number (i.e. `POLLING_DELAY=10`), and
- Set `RELAYER_EXTERNAL_SPOKEPOOL_INDEXER=true`.


src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+  // @todo: Should probably prune blocks and events > 100x finality.
+  providers[0].on("block", (blockNumber) => eventMgr.tick(blockNumber));
+
+  // Setup listeners for each supplied provider.
+  providers.forEach((provider) => {
+    const host = getOriginFromURL(provider.connection.url);
+    filters.forEach((filter) =>
+      spokePool.connect(provider).on(filter, (...rawEvent) => {
+        const event = rawEvent.at(-1);
+        (event.removed ? eventMgr.remove : eventMgr.add).bind(eventMgr)(event, host);
```
pxrl, line 208: Not 100% sure why these methods required `method().bind(eventMgr)(args)` when `eventMgr.tick()` seems to work OK without it. Should confirm that.

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+      })
+    );
+  });
+
+  do {
```
pxrl, line 213: todo:
 - Listen for sighup here and break the loop.
 - Push updates to the parent process from here, rather than directly from `provider.on("block", ...)`. This will allow `SpokePool.currentTime()` to be queried as part of that update.

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+  maxBlockRange?: number; // Maximum block range for paginated getLogs queries.
+  filterArgs?: string[]; // Event-specific filter criteria to apply.
+};
+
+const { NODE_SUCCESS, NODE_APP_ERR } = utils;
+
+let logger: winston.Logger;
+let chain: string;
+
+class EventManager {
```
pxrl, line 41: This class is currently missing tests, but it should be possible to write them fairly quickly given its simple imlpementation. The call to `process.send()` probably needs to be factored out such that it supplies a message to be sent instead.

src/relayer/RelayerClientHelper.ts
```diff
@@ -45,15 +99,26 @@ export async function constructRelayerClients(
...
-    config,
-    baseSigner,
-    config.maxRelayerLookBack,
-    enabledChains
-  );
+  const spokePoolClients =
+    config.pollingDelay > 0 && config.externalIndexer
+      ? Object.fromEntries(
+          await sdkUtils.mapAsync(
+            enabledChains ?? [1, 10, 137, 324, 8453, 42161, 59144], // @todo: fix
```
pxrl, line 106: Major todo here - this is a bodge.

src/relayer/RelayerClientHelper.ts
```diff
@@ -45,15 +99,26 @@ export async function constructRelayerClients(
...
-  const spokePoolClients = await constructSpokePoolClientsWithLookback(
-    logger,
-    hubPoolClient,
-    configStoreClient,
-    config,
-    baseSigner,
-    config.maxRelayerLookBack,
-    enabledChains
-  );
+  const spokePoolClients =
```
pxrl, line 102: fwiw the instantiation of SpokePoolClient instances seems pretty messy now; it should be done cleaner in this PR.

pxrl (commented):
Identifying some todos. I'll focus on fixing these now.

src/relayer/RelayerClientHelper.ts
pxrl: Fixed: https://github.com/across-protocol/relayer-v3/pull/1413/commits/0c842fc5cc88e001bd050aac0c031e472ada45da

src/relayer/RelayerClientHelper.ts
pxrl: Refactored: https://github.com/across-protocol/relayer-v3/pull/1413/commits/0c842fc5cc88e001bd050aac0c031e472ada45da.

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: Break on sighup: https://github.com/across-protocol/relayer-v3/pull/1413/commits/5c4f53ee7ec3149c13062a9d8094fa0ba30825e6

Still need to implement the `getCurrentTime()` call.

src/libexec/RelayerSpokePoolIndexer.ts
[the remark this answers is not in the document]
pxrl: Done: https://github.com/across-protocol/relayer-v3/pull/1413/commits/190c63054b509ce06910b56cb9f533bc5b9cefa1.

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+    const finalised = blockNumber - this.finality;
+    const eventHashes = this.eventHashes[finalised] ?? [];
+
+    // Collect the finalised events for sending.
+    const events = eventHashes
+      .map((eventHash) => this.events[eventHash])
+      .filter((event) => isDefined(event) && finalised >= event.blockNumber)
+      .map(mangleEvent);
+
+    const currentTime = Math.floor(Date.now() / 1000); // @todo
```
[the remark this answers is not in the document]
pxrl: It looks this can be addressed by subscribing to `newHeads` instead: https://github.com/ethers-io/ethers.js/discussions/1951#discussioncomment-4560697

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,388 @@
...
+  const events = ["V3FundsDeposited", "RequestedSpeedUpV3Deposit", "FilledV3Relay"];
+
+  const eventMgr = new EventManager(logger, chainId, finality, quorum);
+  do {
+    let providers: WebSocketProvider[] = [];
+    try {
+      providers = getWSProviders(chainId, quorum);
+      assert(providers.length > 0, `Insufficient providers for ${chain} (required ${quorum} by quorum)`);
+      logger.debug({ at: "RelayerSpokePoolIndexer::run", message: `Starting ${chain} listener.`, events, opts });
+      await listen(eventMgr, spokePool, events, providers, opts);
```
pxrl, line 358: There's a window of opportunity between `scrapeEvents()` and `listen()` where events could be emitted by the SpokePool and would not be caught. This is especially so because some `scrapeEvents()` calls might resolve earlier than others due to varying lookbacks.

A better synchronisation mechanism would be helpful, such that `listen()` could be sure to pick up exactly where `scrapeEvents()` left off. Any solution should be robust to duplicates being emitted, though.

One solution might be to start `listener()` in advance of `scrapeEvents()` and have it buffer events, but only submit them later when the "start" blockNumber is known.

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: getCurrentTime() is no longer necessary per this commit: https://github.com/across-protocol/relayer-v3/pull/1413/commits/5892bc5b7418d12b881fb0df43987bbbe4f1b59b

src/libexec/RelayerSpokePoolIndexer.ts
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/5892bc5b7418d12b881fb0df43987bbbe4f1b59b

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+    return;
+  }
+
+  const message: SpokePoolClientMessage = {
+    blockNumber,
+    currentTime,
+    nEvents: events.length,
+    data: JSON.stringify(sortEventsAscending(events), sdkUtils.jsonReplacerWithBigNumbers),
+  };
+  process.send(JSON.stringify(message));
```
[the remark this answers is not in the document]
pxrl: There's also some upper-limit to the message size. The transport needs to be robust to that.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,189 @@
...
+    const fundsDeposited = eventsToQuery.indexOf("V3FundsDeposited");
+    const [firstDepositId, latestDepositId] = [
+      events[fundsDeposited].at(0)?.args?.depositId ?? this.getDeposits().at(0) ?? 0,
+      events[fundsDeposited].at(-1)?.args?.depositId ?? this.getDeposits().at(-1) ?? 0,
+    ];
+
+    return {
+      success: true,
+      currentTime: this.pendingCurrentTime,
+      oldestTime: this.pendingCurrentTime, // @todo: Cleanup (relayer doesn't care about this).
```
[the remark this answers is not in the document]
pxrl: This might indirectly affect repayment computations via the BundleDataClient though.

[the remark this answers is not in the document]
pxrl: Added support here: https://github.com/across-protocol/relayer-v3/pull/1413/commits/22fab8ca99a8fa28f5206b1dd2287b439db861f2

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,189 @@
...
+
+export type SpokePoolClientMessage = SpokePoolEventsAdded | SpokePoolEventRemoved;
+
+export function isSpokePoolEventsAdded(message: SpokePoolClientMessage): message is SpokePoolEventsAdded {
+  return isDefined((message as SpokePoolEventsAdded).blockNumber);
+}
+
+export function isSpokePoolEventRemoved(message: SpokePoolClientMessage): message is SpokePoolEventRemoved {
+  return isDefined((message as SpokePoolEventRemoved).event);
+}
```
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/048f622713bc6481b5fc46a23a8aa03a3ac7c4e8

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,189 @@
...
+
+    this.chain = getNetworkName(chainId);
+    this.pendingBlockNumber = deploymentBlock;
+    this.pendingCurrentTime = 0;
+    this.pendingEvents = this.queryableEventNames.map(() => []);
+    this.pendingEventsRemoved = [];
+
+    if (isDefined(this.worker)) {
+      this.worker.on("message", (rawMessage: string) => {
+        const message = JSON.parse(rawMessage);
```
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/dda1688d4afa2bc581a51dcb5cbef22cd08c60c9

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,189 @@
...
+    return removed;
+  }
+
+  protected async _update(eventsToQuery: string[]): Promise<clients.SpokePoolUpdate> {
+    // If any events have been removed upstream, remove them first.
+    this.pendingEventsRemoved = this.pendingEventsRemoved.filter(this.removeEvent);
+
+    const events = eventsToQuery.map((eventName) => {
+      const eventIdx = this.queryableEventNames.indexOf(eventName);
+      assert(eventIdx !== -1);
```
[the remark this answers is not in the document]
pxrl: Will leave it as-is for now, since it'd be indicative of a bug if it did occur.

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+  return { ...event, args: sdkUtils.spreadEvent(event.args) };
+}
+
+function getEventFilter(contract: Contract, eventName: string, filterArgs: string[]): EventFilter {
+  const filter = contract.filters[eventName];
+  if (!isDefined(filter)) {
+    throw new Error(`Event ${eventName} not defined for contract`);
+  }
+
+  filterArgs; // @todo
```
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/72e15b3fe4751b5ffb2841e3ccc443dce6fcd863

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+    const eventHashes = (this.eventHashes[event.blockNumber] ??= []);
+    if (eventHashes.indexOf(eventHash) === -1) {
+      eventHashes.push(eventHash);
+    } else {
+      this.events[eventHash] = event;
+    }
+  }
+
+  remove(event: Event, provider: string): void {
+    provider; // @todo: TBD whether this is needed here.
```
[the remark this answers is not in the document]
pxrl: See update: https://github.com/across-protocol/relayer-v3/pull/1413#discussion_r1566378173

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,214 @@
...
+              nEvents,
+              pendingEvents,
+            });
+            return;
+          }
+
+          this.logger.debug({
+            at: "SpokePoolClient#receive",
+            message: `Received ${nEvents} ${this.chain} events from indexer.`,
+          });
```
pxrl, line 119: This message is quite chatty - specially on fast chains like Arbitrum and zkSync. It can probably go away before merge (but is quite useful for debugging).

(In general it'd be nice to retain this at loglevel `debug` and be able to opt-in to it via a runtime flag, but I'd argue that we don't really have a true debug level here since we mostly use it instead of `info`)

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,189 @@
...
+      removed = true;
+      pendingEvents.splice(pendingEventIdx, 1);
+      this.logger.debug({
+        at: "SpokePoolClient#removeEvent",
+        message: `Removed ${getNetworkName(this.chainId)} ${eventName} event for block ${event.blockNumber}.`,
+        transactionHash: event.transactionHash,
+      });
+    }
+
+    // @todo: Back out any events that were previously ingested!
```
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/db42635e54cbcbdf175073a8c53620052bb8d60f

The strategy is just to broadcast the transactionHash and event name(s) affected. In case of a multi-block re-org, event notifications from the RPC provider could be quite spammy. Rather than removing the affected events one by one, just identify the relevant transactions based on their hash and drop them.

A more aggressive strategy could just be to discard all event data for the relevant block number. This might be simpler.

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,481 @@
...
+  do {
+    await setTimeout(delay);
+  } while (!stop);
+}
+
+function getWSProviders(chainId: number, quorum = 1): WebSocketProvider[] {
+  const urls = getNodeUrlList(chainId, quorum, "wss");
+  const providers = urls.map((url: string) => new ethersProviders.WebSocketProvider(url));
+  return providers;
+}
```
pxrl, line 358: To be dropped from this PR after another PR is merged: https://github.com/across-protocol/relayer-v3/pull/1414

src/relayer/index.ts
```diff
@@ -7,13 +19,61 @@ let logger: winston.Logger;
...
+  const sampleOpts = { period, lookback: config.maxRelayerLookBack };
+
+  const chainIds = sdkUtils.dedupArray([...config.relayerOriginChains, ...config.relayerDestinationChains]);
+  assert(chainIds.length > 0); // @todo: Fix to work with undefined chain IDs (default to the complete set).
+
+  return Object.fromEntries(
+    chainIds.map((chainId: number) => {
+      const opts = {
+        ...sampleOpts,
+        finality: chainId === 137 ? 32 : 1,
```
pxrl, line 50: Hardcoding - to be removed. This should be set based on the lowest deposit confirmations identified in `config.minimumDepositConfirmations`.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+
+export function isSpokePoolEventsAdded(message: unknown): message is SpokePoolEventsAdded {
+  return EventsAddedMessage.is(message);
+}
+
+export function isSpokePoolEventRemoved(message: unknown): message is SpokePoolEventRemoved {
+  return EventRemovedMessage.is(message);
+}
+
+export class IndexedSpokePoolClient extends clients.SpokePoolClient {
```
nicholaspai, line 46: Should this be in the SDK..?

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+    if (isDefined(this.worker)) {
+      this.worker.on("message", (rawMessage: string) => {
+        let message: SpokePoolClientMessage;
+        try {
+          message = JSON.parse(rawMessage);
+        } catch (err) {
+          const error = typeguards.isError(err) ? err.message : "unknown error";
+          this.logger.warn({
+            at: "SpokePoolClient#receive",
+            message: "Received malformed message from ${this.chain} indexed.",
```
nicholaspai, line 82:
```suggestion
            message: `Received malformed message from ${this.chain} indexed.`,
```

src/clients/SpokePoolClient.ts
pxrl: Maybe in time, but it's 100% local to the relayer at this point so I think it makes sense to keep it close by. There might be some iterating needed once it goes into prod, so it'd be a pain to have to bump the sdk each time.

src/clients/SpokePoolClient.ts
nicholaspai: Agreed yeah it would be nice to have a log level below debug. Or just a fourth level we can easily filter on in GCP

src/relayer/index.ts
```diff
@@ -7,13 +19,57 @@ let logger: winston.Logger;
...
+        blockRange: config.maxRelayerLookBack[chainId] ?? 5_000,
+      };
+      const chain = getNetworkName(chainId);
+      const child = startWorker("node", config.indexerPath, chainId, opts);
```
nicholaspai, line 50: Is forking in nodejs better than creating separate bots for each chain that we can run separately and have them store data in memory or in a cache? One downside to this approach is complexity and having to manage processes from nodejs which isn't really what its designed for (though you know better, maybe this is a very common use case). The upside is only needing to store in memory.

nicholaspai (commented):
To start, I have left a couple comments on overall structure

james-a-morris (commented):
I've added some comments. 

I have some concerns about implementing support for forked processes and message passing as a bundled executable. Leveraging established frameworks like pm2 for multi-process management, and RabbitMQ or Redis for message passing, might offer us robust solutions with minimal configuration.

src/relayer/index.ts
pxrl: Process management looks like a first-class citizen in node ([docs](https://nodejs.org/api/child_process.html)) and the interfaces provided map very well onto standard posix tooling, so I'm not so concerned about its suitability for this kind of work. It was actually a surprise to me that Node supports custom IPC when the forked processes are also Node-based.

There are some startup overheads with forking that would be bad if we were doing it repeatedly as a part of each loop, but in this case it's a one-time thing on startup. In my testing it happens so quickly that I don't even see the associated log messages for it. Additionally, the advantage of forking is that the lifecycle of each indexer is tied to the parent, so when the parent stops, the indexers go as well. This is really convenient in the gcp environment but also in development - it matches 1:1 with the experience we have today, so when working on the bot you don't really notice that the indexer processes are even there (aside from the extra logging).

This approach definitely addresses some relayer performance issues for me locally vs. the existing implementation, and it should scale OK for the foreseeable future, but will eventually hit scalability issues when the number of chains supported increases by some multiple of current (3 - 5 - 10x ?). The multiprocess approach will work fine in that scenario (and would probably be the only option anyway), but there'd be a lot more context switching between processes and it would probably stress a single host with only 2 cores. We'd also have significant issues in other areas of the relayer and dataworker in this scenario though, so I'm less concerned about this being the bottleneck.

At that point I'd instead look to locate indexer processes on other hosts and would probably revert to JSON-RPC for communication between them. These are mostly low-level details that are abstracted away from the application code though, and I think that by the time we get there the bots will have evolved a lot from their current implementation anyway.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+          if (!Array.isArray(pendingEvents) || pendingEvents.length !== nEvents) {
+            this.logger.warn({
+              at: "SpokePoolClient#receive",
+              message: `Received malformed event update events from ${this.chain} indexer.`,
+              blockNumber,
+              nEvents,
+              pendingEvents,
+            });
+            return;
+          }
```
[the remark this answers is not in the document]
pxrl: Yeah, fair point. The callback doesn't need to be defined in the constructor and can be relocated :+1:.

src/relayer/RelayerClientHelper.ts
```diff
@@ -22,10 +39,45 @@ export interface RelayerClients extends Clients {
...
+  const hints = { lowBlock: getDeploymentBlockNumber("SpokePool", chainId) };
+  const registrationBlock = await getBlockForTimestamp(chainId, time, blockFinder, redis, hints);
+
+  return new IndexedSpokePoolClient(
+    logger,
+    spokePool,
+    hubPoolClient,
+    chainId,
+    registrationBlock,
+    { fromBlock: 0, toBlock: 0, maxBlockLookBack: 5_000 },
```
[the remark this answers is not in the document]
pxrl: Yeah; the SpokePoolClient requires an EventSearchConfig but it doesn't actually get used by this flavour of SpokePoolClient. `maxBlockLookBack` can actually be dropped here, and I can add a comment to explain the rest :+1:.

src/relayer/index.ts
```diff
@@ -7,13 +19,57 @@ let logger: winston.Logger;
...
+type IndexerOpts = {
+  lookback?: number;
+  blockRange?: number;
+};
+
+function startWorker(cmd: string, path: string, chainId: number, opts: IndexerOpts): ChildProcess {
+  const args = Object.entries(opts)
+    .map(([k, v]) => [`--${k}`, `${v}`])
+    .flat();
+  return spawn(cmd, [path, "--chainId", chainId.toString(), ...args], {
```
james-a-morris, line 31: Is this OS agnostic?

src/relayer/index.ts
james-a-morris: While Node's standard library does support forking child processes, I wonder if this might not be the most typical or efficient method for our specific use case. 

Given that we already successfully manage communication between bots using Redis, perhaps we could expand this existing infrastructure to meet our new needs without having to fork additional subprocesses. This approach would not only allow us to continue using well-understood code but also might enhance accessibility and integration for other bots

mrice32: I think, fundamentally, we're trying to allow the relayer to pull data and send transactions to many chains at once. To do this effectively as we grow to support more chains, we can't block on any failures or slowdowns on particular chains. Even if we ran a separate relayer for each chain, that relayer would still have to talk to every other chain to gather the deposits, running into the same problem. Also, as we scale, we'll need to begin multithreading this whole operation as it becomes more compute intensive.

I think forking in nodejs is the best approximation of multithreading you can really do (since each nodejs process is single-threaded).

I think the alternative approach to get the modularization and non-blocking behavior without multithreading would be to do explicit in-process messaging/synchronization to allow event pulling and event processing to happen in parallel to not allow one chain to block others. I think the fork approach is a simpler way to achieve that end result and allows us to do multithreading, giving us more compute resources. So it's not strictly necessary, but I think it's a pretty reasonable approach.

Long term, I see the primary downside of this approach being resource inefficiency. I don't know, but I suspect we'll hit memory limitations at some point that limit our ability to support more than `n` chains in this approach since each process will probably have significant memory usage. Similarly, we may run into issues with networking resources (like ports) because we can't as easily coordinate these separate processes' usage of the network. I'm less worried about compute since I assume the relayer is mostly a memory/network hog, but @pxrl would know better given his testing.

IMO, this is a step in the right direction, allowing us to begin scaling the relayer operation and adding more and different modularization strategies as we go.

src/relayer/index.ts
mrice32: Does this mean we're still evaluating MCD in the parent process, not the child?

mrice32 (commented):
Looks great! A few comments, but overall I think this is a solid approach. Would want to get some metrics (if/when you have them) around memory/cpu/port/networking usage.

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: Nice - thanks for the context. Easy fix: https://github.com/across-protocol/relayer-v3/pull/1413/commits/b365bffb5823f30b285e28c4ac96d125c1b79689

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,388 @@
...
+  const spokePool = await utils.getSpokePoolContract(chainId);
+
+  process.on("SIGHUP", () => {
+    logger.debug({ at: "Relayer#run", message: "Received SIGHUP, stopping..." });
+    stop = true;
+  });
+
+  // EnabledDepositRoutes should always look back over the entire history.
+  // @note: An improvement is on the way...
+  // Note: An event emitted between scrapeEvents() and listen(). @todo: Ensure that there is overlap and dedpulication.
```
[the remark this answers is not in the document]
pxrl: > I do think this is mostly not a huge problem for the _correctness_ of the relayer. The two issues:
> 
> * Dropping deposits (but hopefully the lookback will handle that on the next run?), which will result in a delay.

Yep, the configurable lookback (usually ~2 hours) would catch it on the subsequent run.

> * Missing deposit route updates, which could be a real problem right? But these are fairly rare, predictable, and almost never change an existing mapping.

Agreed. I was thinking specifically about the USDC scenario where the token mappings will be changed on some chains. I think this is probably a non-issue though, because I expect the duration of the transition for each chain to be longer some multiple of a single bot cycle, and during this transition the routes should be disabled first before being re-enabled with new tokens. So even in the off chance that the indexer happened to hit this small window, I don't think it'd open up for any worse exposure to having the wrong token mapping.

As a side note, I want to change the way EnabledDepositRoute events are loaded as a follow-up. I think it makes sense for these events to be snapshotted and available to load via a file in the repo, rather than having to load all the way back to genesis once per SpokePool. The infinite TTL caching does help with this, but it still results in a _lot_ of queries being performed, and it only gets worse for each new chain that's added. This would help to shave a couple of seconds off the SpokePool update time and would significantly reduce the number of concurrent queries issued to Redis.

src/relayer/RelayerClientHelper.ts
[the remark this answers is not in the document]
pxrl: Refactored here such that `EventSearchConfig` is instantiated directly in the constructor - hopefully for minimal confusion: https://github.com/across-protocol/relayer-v3/pull/1413/commits/1509815d0d6c15c9fd7e512f7df10b1350e11ad5.

src/common/Constants.ts
```diff
@@ -312,4 +312,6 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
   ],
 };
 
+export const RELAYER_DEFAULT_INDEXER = "./dist/src/libexec/RelayerSpokePoolIndexer.js";
```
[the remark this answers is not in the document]
pxrl: Updated to add a comment and tweak the naming: https://github.com/across-protocol/relayer-v3/pull/1413/commits/2486713514772e5065dd43c1c799b42236d05ac2

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+
+          this.logger.debug({
+            at: "SpokePoolClient#receive",
+            message: `Received ${nEvents} ${this.chain} events from indexer.`,
+          });
+
+          pendingEvents.forEach((event) => {
+            const eventIdx = this.queryableEventNames.indexOf(event.event);
+            if (eventIdx === -1 || event.removed) {
```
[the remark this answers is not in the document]
pxrl: It's moreso indicative of an event that was (possibly) previously sent by the indexer to the SpokePoolClient having been subsequently removed upstream (i.e. via reorg). The indexer doesn't try to track what it has already sent to the SpokePoolClient, so any `removed` event it receives is just relayed over so the SpokePoolClient can decide whether it needs to flush anything.

src/clients/SpokePoolClient.ts
pxrl: There's unfortunately nothing below `debug` :(

There are [7 severity levels](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1) defined, and we mostly use `error`, `warning`, `informational` and `debug`. I think >= `info` is automatically forwarded to notification channels, which means we only have `debug` left over for local logging. I don't recall seeing `notice` used much or at all, so I think we might be under-using it. An improvement could be to bump our existing `info` logs to `notice` (or above), stop forwarding `info` to notification channels and reserve it for local logging, and then have a runtime debug flag that enables `debug`. We could then retain useful debug messages without having them spam the log output. @mrice32 You'll surely have more context on this - any reason we don't use `notice` ?

```
           Numerical         Severity
             Code

              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages
```

src/relayer/index.ts
pxrl: Correct. The indexer doesn't evaluate the events themselves so it doesn't know about the value of any V3FundsDeposited event. It therefore only applies the _minimum_ deposit confirmations specified at runtime (--finality flag).

The MDC logic in the relayer is quite separate from where the SpokePoolClient is ingesting these events, and the relayer's MDC logic also considers the cumulative commitment for the batch of outstanding fills, rather than each deposit event in isolation. This is to mitigate the scenario where there are a number of deposit events in a given block where they individually fall under an MDC limit but would cumulatively exceed it. 

(nb. It has occurred to me that this cumulative evaluation can be problematic when the relayer doesn't have funds to fill one or more _large_ outstanding deposits that are destined for a slow fill - i.e. it would be bumping its MDC requirements for an origin chain based on a large deposit that it'll not actually attempt to fill due to a token shortfall. I have an idea for how we can avoid that, so that we only compute MDCs on deposits where we actually expect to fill them)

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/14ffbf8201468d4dc87cb3372ec995784f87fd1f

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+        if (!isSpokePoolEventsAdded(message)) {
+          this.logger.warn({
+            at: "SpokePoolClient#receive",
+            message: `Received unrecognised message from ${this.chain} indexer.`,
+            data: message,
+          });
+          return;
+        }
+
+        const { blockNumber, currentTime, oldestTime, nEvents, data } = message;
```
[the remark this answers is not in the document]
pxrl: Refactored: https://github.com/across-protocol/relayer-v3/pull/1413#discussion_r1570595500

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,470 @@
...
+  remove(event: Event, provider: string): void {
+    assert(event.removed);
+
+    const events = this.events[event.blockNumber] ?? [];
+
+    // Filter coarsely on transactionHash, since a reorg should invalidate multiple events within a single transaction hash.
+    const eventIdxs = events
+      .map((event, idx) => ({ ...event, idx }))
+      .filter(({ transactionHash }) => transactionHash === event.transactionHash)
+      .map(({ idx }) => idx);
```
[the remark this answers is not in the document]
pxrl: I reverted to handling it on a per-event basis: https://github.com/across-protocol/relayer-v3/pull/1413/commits/58c077fc437102c81967d4e8980c0e90633e4171

The code is a bit cleaner as a result.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+    });
+
+    eventNames.forEach((eventName) => {
+      const eventIdx = this.queryableEventNames.indexOf(eventName);
+      const pendingEvents = this.pendingEvents[eventIdx];
+
+      // First check for removal from any pending events.
+      const pendingEventIdxs = pendingEvents
+        .map((pending, idx) => ({ ...pending, idx }))
+        .filter((pending) => pending.transactionHash === transactionHash)
```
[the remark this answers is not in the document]
pxrl: Great catch - that's spot on. In a reorg scenario the transaction hash will likely reappear, so it's not a good candidate for this. Because of the quroum system it possibly risked removing valid _replacement_ events (i.e. if a 2nd provider was late in notifying of event removal, it might overwrite one that has already been replaced. So transactionHash would never have worked well here.

I reverted to flushing pending events based on multiple criteria in the event, so we're sure to match the right one: https://github.com/across-protocol/relayer-v3/pull/1413/commits/58c077fc437102c81967d4e8980c0e90633e4171.

src/relayer/index.ts
pxrl: I've only tested it in a Linux environment, but I don't see any reason why it wouldn't be portable across any of the implicitly-targeted platforms (Linux, MacOS, BSD). `child_process.spawn()` is even fully supported under Windows.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+    this.pendingEventsRemoved = [];
+
+    if (isDefined(this.worker)) {
+      this.worker.on("message", (rawMessage: string) => {
+        let message: SpokePoolClientMessage;
+        try {
+          message = JSON.parse(rawMessage);
+        } catch (err) {
+          const error = typeguards.isError(err) ? err.message : "unknown error";
+          this.logger.warn({
```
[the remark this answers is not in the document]
pxrl: I'd normally tend towards logging loudly about it, but I also like the simplicity of just throwing - it keeps the code compact. Implemented: https://github.com/across-protocol/relayer-v3/pull/1413/commits/2ae25923652f0d776db0bb8fbced7d8de30c5c52.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+
+          this.logger.debug({
+            at: "SpokePoolClient#receive",
+            message: `Received ${nEvents} ${this.chain} events from indexer.`,
+          });
+
+          pendingEvents.forEach((event) => {
+            const eventIdx = this.queryableEventNames.indexOf(event.event);
+            if (eventIdx === -1 || event.removed) {
+              this.logger.warn({
```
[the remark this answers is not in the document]
pxrl: Fair call - updated: https://github.com/across-protocol/relayer-v3/pull/1413#discussion_r1570692796

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+  eventNames: array(string()),
+});
+
+export function isSpokePoolEventsAdded(message: unknown): message is SpokePoolEventsAdded {
+  return EventsAddedMessage.is(message);
+}
+
+export function isSpokePoolEventRemoved(message: unknown): message is SpokePoolEventRemoved {
+  return EventRemovedMessage.is(message);
+}
```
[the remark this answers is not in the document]
pxrl: Moved here: https://github.com/across-protocol/relayer-v3/pull/1413/commits/a1d140611accdcc96faecf0c39a1169283536e70

I am fwiw conflicted on this. I see the aim of keeping everything together in one place, but these are exclusively used in this file and aren't intended for general use throughout the codebase, so I it also seems OK keep it all local because it's self-contained and really easy to update.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+      // Back out any events that were previously ingested via update(). This is best-effort and may help to save the
+      // relayer from filling a deposit where it must wait for additional deposit confirmations. Note that this is
+      // _unsafe_ to do ad-hoc, since it may interfere with some ongoing relayer computations relying on the
+      // depositHashes object. If that's an acceptable risk then it might be preferable to simply assert().
+      if (eventName === "V3FundsDeposited") {
+        const depositHashes = Object.values(this.depositHashes)
+          .filter((deposit) => deposit.transactionHash === transactionHash)
+          .map((deposit) => this.getDepositHash(deposit));
+
+        depositHashes.forEach((hash) => delete this.depositHashes[hash]);
```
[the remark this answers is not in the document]
pxrl: > Edit: actually, this seems to only be called during a spoke pool update, not at arbitrary times. Do you see a concrete issue with the next loop having a different event history? Shouldn't the loops be completely independent?

Yeah - there's a comment at at the top that mentions it's only done during _update(), specifically because it otherwise might risk occurring in the middle of an async operation that somehow depends on the length or ordering of the array:

```
    // Back out any events that were previously ingested via update(). This is best-effort and may help to save the                                                                            
    // relayer from filling a deposit where it must wait for additional deposit confirmations. Note that this is                                                                               
    // _unsafe_ to do ad-hoc, since it may interfere with some ongoing relayer computations relying on the                                                                                     
    // depositHashes object. If that's an acceptable risk then it might be preferable to simply assert().
```

I'm confident that the main relayer functions will be fine with this because the loops are independent, as you mention. I'm a bit less certain about functions that retain state across loops - i.e. the InventoryClient. It's moving quickly at the moment and I do still see some occasional hiccups with it running in looping mode. We could alternatively just decide to assert() if we identify this scenario - that's arguably safest, albeit with a service interruption.

src/relayer/index.ts
pxrl: Updated here to scan the minimum deposit confirmation configuration and use the _lowest_ configured dollar threshold as the minimum finality for the indexer to use: https://github.com/across-protocol/relayer-v3/pull/1413/commits/4970709fed1cfe32f9e78c87d2c19b59965d0a1a

src/relayer/index.ts
pxrl: >Long term, I see the primary downside of this approach being resource inefficiency. I don't know, but I suspect we'll hit memory limitations at some point that limit our ability to support more than n chains in this approach since each process will probably have significant memory usage. Similarly, we may run into issues with networking resources (like ports) because we can't as easily coordinate these separate processes' usage of the network. I'm less worried about compute since I assume the relayer is mostly a memory/network hog, but @pxrl would know better given his testing.

I think networking in this case should be no worse than we have currently since we're not really adjusting request load, but are instead just shuffling existing demand over to another process. If we continued to make RPC-heavy requests concurrently from the indexer and relayer instances then that'd be an issue because they'd compete with each other, but I'd be surprised to see that in practice. We'll also see a big benefit when we skip querying EnabledDepositRoute events and instead just load a snapshot of the data.

fwiw, here's a snip from some local testing. I levelled up in top and instructed it to isolate for specific pids, so it only shows the indexer processes here.

node

It looks like each process has a resident memory usage of 250 - 300MB. I've had these processes running for about 30 minutes or so, and none are exceeding 300MB so it seems fairly stable. It's quite a lot when I consider that my first PC had 256MB ram...but on a host w/ 8GB RAM we should be able to support quite a few additional chains before we tap out. The virt allocation of 11.1GB might seem alarming, but it's not representative of actual usage.

Side note: Something that could be fun is to reimplement the indexer process in viem instead. This is totally achievable now that the indexer has a separate instance and isn't bound to use the same tech stack as the main relayer process. viem makes a big deal about being lighter and more efficient than ethers, so it'd be a good test of that claim. It would require some minor adjustments to the types sent between the two processes (currently uses an Ethers Event type) but it'd be a fairly simple change and could potentially save on RAM.

I also see that the indexer processes typically sit with very low CPU util - somewhere between 0.1 and 2%. WebSockets really help here - the events are trickling in over time, so the load of processing them is nicely distributed (probably roughly approximating chain activity and the average block time) and there's little (if any) burstiness once we get beyond the initial lookback scraping. I do have a 12 core CPU, so I get the benefit of fanning the processes out across multiple cores, and the kernel isn't probably isn't having to pre-empt any of these. I have an old dual core laptop that I'd like to test this out on to see how the context switching goes, but I'm pretty optimistic for being able to support a lot more chains than we currently do before we'd need to adjust this setup.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,250 @@
...
+            data: message,
+          });
+          return;
+        }
+
+        const { blockNumber, currentTime, oldestTime, nEvents, data } = message;
+        if (nEvents > 0) {
+          const pendingEvents = JSON.parse(data, sdkUtils.jsonReviverWithBigNumbers);
+          if (!Array.isArray(pendingEvents) || pendingEvents.length !== nEvents) {
+            this.logger.warn({
```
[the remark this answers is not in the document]
pxrl: Updated to assert in line with the other comments :+1: https://github.com/across-protocol/relayer-v3/pull/1413#discussion_r1570691967

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: EventManager tests added here: https://github.com/across-protocol/relayer-v3/pull/1413/commits/a435397dbf8d5ce42168893d0c519fa6f6360b6b

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    // Flush the confirmed events.
+    eventHashes.forEach((eventHash) => delete this.events[eventHash]);
+  }
+
+  hashEvent(event: Event): string {
+    const { event: eventName, blockNumber, blockHash, transactionHash, transactionIndex, logIndex, args } = event;
+    return ethersUtils.id(
+      `${eventName}-${blockNumber}-${blockHash}-${transactionHash}-${transactionIndex}-${logIndex}-${args.join("-")}`
+    );
```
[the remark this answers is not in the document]
pxrl: Yeah I thought it better to include more data than less. I'll leave it as-is for now, since CPU load is absolutely minimal for these processes.

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,225 @@
...
+        pending.logIndex === logIndex &&
+          pending.transactionIndex === transactionIndex &&
+          pending.transactionHash === transactionHash &&
+          pending.blockHash === blockHash;
+      });
+
+    if (isDefined(pendingEvent)) {
+      removed = true;
+
+      // Splice out the events in reverse order.
```
pxrl, line 147: Comment is wrong, since this is only splicing a single entry out.
```suggestion
      // Drop the relevant event.
```

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,318 @@
...
+  finality?: number; // Event finality (in blocks).
+  quorum?: number; // Provider quorum to apply.
+  deploymentBlock: number; // SpokePool deployment block
+  maxBlockRange?: number; // Maximum block range for paginated getLogs queries.
+  filterArgs?: { [event: string]: string[] }; // Event-specific filter criteria to apply.
+};
+
+const { NODE_SUCCESS, NODE_APP_ERR } = utils;
+
+const INDEXER_POLLING_PERIOD = 2000; // ms; time to sleep between checking for exit request via SIGUP.
```
pxrl, line 40:
```suggestion
const INDEXER_POLLING_PERIOD = 2000; // ms; time to sleep between checking for exit request via SIGHUP.
```

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,318 @@
...
+  await providers[0]._subscribe("newHeads", ["newHeads"], ({ number: blockNumber, timestamp: currentTime }) => {
+    [blockNumber, currentTime] = [parseInt(blockNumber), parseInt(currentTime)];
+    const events = eventMgr.tick(blockNumber);
+
+    // Post an update to the parent. Do this irrespective of whether there were new events or not, since there's
+    // information in blockNumber and currentTime alone.
+    postEvents(blockNumber, currentTime, events);
+  });
+
+  // Add a handler for each new instance of a subscribed.
```
pxrl, line 184:
```suggestion
  // Add a handler for each new instance of a subscribed event.
```

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,318 @@
...
+    string: ["relayer"],
+  };
+  const args = minimist(argv, minimistOpts);
+
+  const { chainId, finality = 32, quorum = 1, lookback = 7200, relayer = null, maxBlockRange = 10_000 } = args;
+  assert(Number.isInteger(chainId), "chainId must be numeric ");
+  assert(Number.isInteger(finality), "finality must be numeric ");
+  assert(Number.isInteger(lookback), "lookback must be numeric");
+  assert(Number.isInteger(maxBlockRange), "maxBlockRange must be numeric");
+  assert(!isDefined(relayer) || ethersUtils.isAddress(relayer), "relayer address is invalid");
```
pxrl, line 221:
```suggestion
  const { chainId, finality = 32, quorum = 1, lookback = 7200, relayer = null, maxBlockRange = 10_000 } = args;
  assert(Number.isInteger(chainId), "chainId must be numeric ");
  assert(Number.isInteger(finality), "finality must be numeric ");
  assert(Number.isInteger(quorum), "quorum must be numeric ");  
  assert(Number.isInteger(lookback), "lookback must be numeric");
  assert(Number.isInteger(maxBlockRange), "maxBlockRange must be numeric");
  assert(!isDefined(relayer) || ethersUtils.isAddress(relayer), "relayer address is invalid");
```

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,225 @@
...
+        at: "SpokePoolClient#removeEvent",
+        message: `Removed ${this.chain} ${eventName} event for block ${blockNumber}.`,
+        event,
+      });
+    }
+
+    // Back out any events that were previously ingested via update(). This is best-effort and may help to save the
+    // relayer from filling a deposit where it must wait for additional deposit confirmations. Note that this is
+    // _unsafe_ to do ad-hoc, since it may interfere with some ongoing relayer computations relying on the
+    // depositHashes object. If that's an acceptable risk then it might be preferable to simply assert().
```
nicholaspai, line 160: This definitely seems dangerous and makes any users of this client very hard to reason about. I'm just commenting that I agree but don't know a workaround yet

src/relayer/index.ts
```diff
@@ -7,13 +19,64 @@ let logger: winston.Logger;
...
+
 export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
   const relayerRun = randomNumber();
   const startTime = getCurrentTime();
 
   logger = _logger;
   const config = new RelayerConfig(process.env);
 
+  let workers: { [chainId: number]: ChildProcess };
+  if (config.externalIndexer) {
```
nicholaspai, line 76: Does this mean you can have both externally-indexed clients and looping clients?

nicholaspai (approved):
This is a big PR but directionally it seems correct. I'm going to have to re-review but i don't want to block experimentation. I don't think we should use this in prod yet but its good to test a different EOA with

src/clients/SpokePoolClient.ts
pxrl: One additional thing to note: this is only an issue for events that are re-orged _after_ the indexer's own finality configuration. On Polygon I'm proposing to set that to 32 blocks, so the indexer would hold all Polygon events for 32 blocks before it forwards them to the relayer.

On the optimistic L2s I'm proposing 1 block, but the value is derived from the relayer's MDC configuration, so it's easy to adjust this up and down as required on a per-chain bsis. That's the best way to protect against ingesting bad events.

pxrl:
> This is a big PR but directionally it seems correct. I'm going to have to re-review but i don't want to block experimentation. I don't think we should use this in prod yet but its good to test a different EOA with

Yep, agreed. It's proposed for a secondary EOA with a much smaller pool of funds (but still sufficient to evaluate this).

src/relayer/index.ts
pxrl: Here's another snapshot from a run after some recent updates. CPU utilisation is still very low; peak RAM consumption on a couple of processes has increased to a touch more than ~300MB RAM. This is after > 1 hour of runtime, so we'd be unlikely to see this in production with the relayer runtime capped at 5 minutes.

top

src/clients/SpokePoolClient.ts
```diff
@@ -1,4 +1,225 @@
...
+          at: "SpokePoolClient#removeEvent",
+          message: `Removed 1 pre-ingested ${this.chain} ${eventName} event.`,
+          event,
+        });
+      }
+    } else if (eventName === "EnabledDepositRoute") {
+      // These are hard to back out because they're not stored with transaction information. They should be extremely
+      // rare, but at the margins could risk making an invalid fill based on the resolved outputToken for a deposit
+      // that specifies outputToken 0x0. Simply bail in this case; everything should be OK on the next run.
+      assert(false, "Detected re-org affecting deposit route events.");
```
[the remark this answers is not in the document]
pxrl: Ah, throw() makes more sense in this context :+1: 

Fixed: https://github.com/across-protocol/relayer-v3/pull/1413/commits/ef0f55cf1d70b9309ab663dd4e4eae596d254d77

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -0,0 +1,319 @@
...
+ */
+function removeEvent(event: Event): void {
+  if (!isDefined(process.send)) {
+    return;
+  }
+
+  const message: SpokePoolClientMessage = {
+    event: JSON.stringify(mangleEventArgs(event), sdkUtils.jsonReplacerWithBigNumbers),
+  };
+  process.send(JSON.stringify(message));
```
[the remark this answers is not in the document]
pxrl: There's flexibility here and I expect the interface will evolve from here. I'm probably keen to just go with JSON for now, since it's worked really well in my testing and doesn't require any external dependencies. I definitely think the interface will evolve though, so we shouldn't rule anything out.

PR #1414, opened 2024-04-15
improve(ProviderUtils): Add helper to get WebSocket providers

pxrl:


src/utils/ProviderUtils.ts
```diff
@@ -710,6 +710,11 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
   return provider;
 }
 
+export function getWSProviders(chainId: number, quorum = 1): ethers.providers.WebSocketProvider[] {
+  const urls = getNodeUrlList(chainId, quorum, "wss");
```
james-a-morris, line 714: How does quorum behave in a WebSocket context?

nicholaspai: good q

pxrl: For the most part, it's currently external to the WebSocketProvider. This is because WebSocketProvider event subscription occurs via callbacks so it's not possible to synchronously evaluate two updates. Instead, updates need to be buffered indefinitely and evaluated at some later point. I'm currently working on this via the `EventManager` class over in this PR: https://github.com/across-protocol/relayer-v3/pull/1413

It's probably going to be possible to bundle `EventManager` capabilities into some "QuorumWebSocketProvider" class that internalises quorum, but I think it's too early for that at the moment.

Quorum within the context of `getWSProviders()` simply ensures that there are at least `quorum` URLs defined for the specified chain ID, and if not, it throws. This matches the existing behaviour of `getProvider()`.

mrice32: Was going to ask the same question. Ty for the explanation.

mrice32 (approved):
LGTM!

PR #1415, opened 2024-04-18
improve(CCTP): add addresses for CCTP ABI

james-a-morris:
My work in the IM has overlapped with an open PR in the finalizer that's awaiting final reviews - to avoid committing duplicate definitions, my work can be rebased with this PR upstream


nicholaspai (approved):
I like this thanks for unblocking yourself

PR #1416, opened 2024-04-19
fix(executor): Handle edge case when executing non-hub-chain pool leaves

nicholaspai:
There is one missing case not handled:
- HubChain Leaf for a token has a negative net send amount, so no exchange rate needed
- RelayerLeaf for token has a positive amount to return. This adds virtual balance for the token
- non-HubChain leaf for the token has a positive net send amount. The `liquidReserves + amountToReturn` is still `< netSendAmount`. This throws an error. However, we should be checking if calling `exchangeRateCurrent` on the `liquidReserves` sends liquid reserves above the required amount. This follows the pattern in the other two exchange rate update cases

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1703,8 +1703,17 @@ export class Dataworker {
           });
           return;
         }
-        // @dev: Virtual balance = current balance + any used balance.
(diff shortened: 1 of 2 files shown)


pxrl (approved):

pxrl (approved):
Nice!

PR #1417, opened 2024-04-20
improve(POOL): add pool support & bump deps

james-a-morris:
To add support for PoolTogether in the relayer, we need to update the IM adapters & bump the contracts, constants, & SDK.


nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #1419, opened 2024-04-22
improve: Remove repetitive words

findnature:


james-a-morris (commented):
I think this is a good change to the docs. 

Because this is an external contribution, I will also defer to @nicholaspai and @pxrl

nicholaspai (approved):

PR #1423, opened 2024-04-23
improve(indexer): Bail on parent disconnect

pxrl:


src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -252,6 +252,11 @@ async function run(argv: string[]): Promise<void> {
     stop = true;
   });
 
+  process.on("disconnect", () => {
+    logger.debug({ at: "Relayer#run", message: "Parent disconnected, stopping..." });
```
pxrl, line 256:
```suggestion
    logger.debug({ at: "Relayer::run", message: "Parent disconnected, stopping..." });
```

james-a-morris (approved):
+1 Fair call to prevent any further messages while the whisperer is iterating over its last loop.

PR #1424, opened 2024-04-23
feat(CCTP): add cctp chain adapter to IM

james-a-morris:
This PR adds CCTP support for the Op Stack, Arbitrum, and Polygon adapters in the Inventory Manager. Because CCTP implementations are similar between chains, an intermediary class has been created between the adapter and the `BaseAdapter` to store relevant CCTP logic.

Outstanding Cross-Chain Transfers on live data:
```
2024-04-25 03:05:21 [[34mdebug[39m]: {
  "at": "CrossChainTransferClient",
  "message": "Updated cross chain transfers",
  "outstandingCrossChainTransfers": {
    "10": {
      "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D": {
        "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
          "totalAmount": "10000",
          "depositTxHashes": [
            "0x0e7d41c9258c9fb64c48cbaca84cb36024e8ec64a4a161b3f398eefdd032039b"
          ]
        }
      },
    },
    "137": {
      "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D": {
        "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
          "totalAmount": "10000",
          "depositTxHashes": [
            "0x2be88123779dcc64f696ce4e6d2801573e828d6bc337e56bc5392a120c95e60c"
          ]
        }
      },
    },
    "8453": {
      "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D": {
        "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
          "totalAmount": "10000",
          "depositTxHashes": [
            "0x43beec74c015b5e28261979ed3279aa8db122d015624f4fe0f90ae49afa2e3ca"
          ]
        }
      },
    },
    "42161": {
      "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D": {
        "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
          "totalAmount": "10000",
          "depositTxHashes": [
            "0x696d68ef5ab0d39ee90cccdfd9092aab6ace98e450b8ab64b29aa8769c3bf31b"
          ]
        }
      },
    }
  },
  "bot-identifier": "not-james"
}
```


src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -87,10 +88,15 @@ export class ArbitrumAdapter extends BaseAdapter {
...
 
     const promises: Promise<Event[]>[] = [];
+    const cctpOutstandingTransfersPromise: Record<string, Promise<SortableEvent[]>> = {};
     const validTokens: string[] = [];
     // Fetch bridge events for all monitored addresses.
     for (const monitoredAddress of this.monitoredAddresses) {
       for (const l1Token of availableL1Tokens) {
+        if (this.isL1TokenUsdc(l1Token)) {
+          cctpOutstandingTransfersPromise[monitoredAddress] = this.getOutstandingCctpTransfers(monitoredAddress);
+        }
```
james-a-morris, line 98: As a note - I thought it would be a good idea to also scrape CCTP events alongside other USDC transfer events that are currently being scraped for.

Therefore, CCTP is out of band of the existing logic.

src/clients/bridges/BaseAdapter.ts
```diff
@@ -66,12 +65,11 @@ export abstract class BaseAdapter {
 
   constructor(
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
-    _chainId: number,
+    readonly chainId: number,
```
nicholaspai, line 68: +1

src/utils/CCTPUtils.ts
```diff
@@ -14,6 +16,107 @@ export type DecodedCCTPMessage = {
...
+ * @param contract The CCTP MessageTransmitter contract to call.
+ * @returns Whether or not the message has been processed.
+ */
+export async function hasCCTPMessageBeenProcessed(
+  sourceDomain: number,
+  nonce: number,
+  contract: ethers.Contract
+): Promise<boolean> {
+  const nonceHash = hashCCTPSourceAndNonce(sourceDomain, nonce);
+  const resultingCall = await contract.callStatic.usedNonces(nonceHash);
```
nicholaspai, line 115: nit: maybe define the expected return type of `usedNonces` and cast it here. It looks like its either returning a string or a BN?

src/clients/bridges/CCTPAdapter.ts
```diff
@@ -0,0 +1,155 @@
...
+    return chainIdsToCctpDomains[this.chainId];
+  }
+
+  /**
+   * Check if an L1 token is USDC - this is a requirement to be transferred via CCTP
+   * @param l1Token A Web3 address of a token
+   * @returns Whether or not this token is USDC on Mainnet
+   */
+  protected isL1TokenUsdc(l1Token: string): boolean {
+    return compareAddressesSimple(l1Token, TOKEN_SYMBOLS_MAP.USDC.addresses[this.hubChainId]);
```
nicholaspai, line 50: how does this work once we have both USDC and bridged USDC addresses in TOKEN_SYMBOLS_MAP?

src/clients/bridges/CCTPAdapter.ts
```diff
@@ -0,0 +1,155 @@
...
+    return new Contract(
+      CONTRACT_ADDRESSES[hubChainId].cctpTokenMessenger.address,
+      CONTRACT_ADDRESSES[hubChainId].cctpTokenMessenger.abi,
+      this.getSigner(hubChainId)
+    );
+  }
+
+  protected getL2CCTPTokenMessengerBridge(): Contract {
+    return new Contract(
+      CONTRACT_ADDRESSES[this.chainId].cctpTokenMessenger.address,
```
nicholaspai, line 77: Should all of these accessors in CONTRACT_ADDRESSES have "?" to handle `undefined` or do you want to it to just throw?

src/clients/bridges/CCTPAdapter.ts
```diff
@@ -0,0 +1,155 @@
...
+    const l1BridgeMethod = "depositForBurn";
+    const l1BridgeArgs = [amount, this.l2DestinationDomain, cctpAddressToBytes32(address), l1Token];
+    return this._sendTokenToTargetChain(
+      l1Token,
+      l2Token,
+      amount,
+      l1Bridge,
+      l1BridgeMethod,
+      l1BridgeArgs,
+      2,
```
nicholaspai, line 150: what is 2?

src/clients/bridges/ArbitrumAdapter.ts
nicholaspai: that's a fair implementation. I guess the alternative is to implement a `getOutstandingCrossChainTransfers` function in the CCTP adapter and have each XAdapter merge those results with the existing ones at the top level.

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -110,6 +116,13 @@ export class ArbitrumAdapter extends BaseAdapter {
     }
 
     const results = await Promise.all(promises);
+    const resultingCCTPEvents: Record<string, SortableEvent[]> = Object.fromEntries(
+      await Promise.all(
```
nicholaspai, line 120: Can we combine this promise with the above to reduce run-time slightly?

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -136,10 +137,15 @@ export class PolygonAdapter extends BaseAdapter {
     const availableTokens = l1Tokens.filter(this.isSupportedToken.bind(this));
 
     const promises: Promise<Event[]>[] = [];
+    const cctpOutstandingTransfersPromise: Record<string, Promise<SortableEvent[]>> = {};
```
nicholaspai, line 140: Should this be in the `CCTPAdapter` since all XAdapters need to set this?

src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -46,6 +47,17 @@ export class OpStackAdapter extends BaseAdapter {
       );
     }
 
+    // We should manually override the bridge for USDC to use CCTP.
+    const usdcAddress = TOKEN_SYMBOLS_MAP._USDC.addresses[this.hubChainId];
+    if (usdcAddress) {
+      this.customBridges[usdcAddress] = new UsdcCCTPBridge(
```
nicholaspai, line 53: The way that the OPStack maps `l1Token ==> customBridges` is really clean and makes adding CCTP easier--all we have to do is define the UsdcCCTPBridge custom bridge. Should we try to replicate this in the other chain adapters? I understand this might be a larger overhaul so it can be follow up but its worth considering versus the existing changes where we add additional logic alongside existing non-CCTP adapter logic

nicholaspai (commented):
This is great work! Will review again after unit tests are added

src/clients/bridges/CCTPAdapter.ts
james-a-morris: I think we can throw - this should only be an issue once, if any because the values are hardcoded.

I wonder what the tradeoffs would be to enforce the undefined/null references in TSC

src/clients/bridges/CCTPAdapter.ts
james-a-morris: This is a gas multiplier - I think we should initially set it to 2x and taper down.

james-a-morris:
> This is great work! Will review again after unit tests are added

Updated the tests to include the CCTP adapter.

src/clients/bridges/ArbitrumAdapter.ts
james-a-morris: https://github.com/across-protocol/relayer-v3/pull/1424/commits/be0afb8d37e5dc6b8ad75a81dc4a086ab76a7adc

src/utils/CCTPUtils.ts
james-a-morris: https://github.com/across-protocol/relayer-v3/pull/1424/commits/e3d11979f9710823f64fdb7f298cc8c649b4974b

src/utils/CCTPUtils.ts
```diff
@@ -17,6 +19,107 @@ export type DecodedCCTPMessage = {
...
+ * @param contract The CCTP MessageTransmitter contract to call.
+ * @returns Whether or not the message has been processed.
+ */
+export async function hasCCTPMessageBeenProcessed(
+  sourceDomain: number,
+  nonce: number,
+  contract: ethers.Contract
+): Promise<boolean> {
+  const nonceHash = hashCCTPSourceAndNonce(sourceDomain, nonce);
+  const resultingCall = (await contract.callStatic.usedNonces(nonceHash)) as BigNumber;
```
nicholaspai, line 118: would this be better or is it the same?
```suggestion
  const resultingCall: BigNumber = await contract.callStatic.usedNonces(nonceHash);
```

src/utils/CCTPUtils.ts
nicholaspai: https://github.com/across-protocol/relayer-v3/pull/1424#discussion_r1581091975

src/clients/bridges/CCTPAdapter.ts
nicholaspai: ok can you add comment?

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -161,41 +180,48 @@ export class ArbitrumAdapter extends BaseAdapter {
...
-      l2Token,
-      amount,
-      this.getL1GatewayRouter(),
-      "outboundTransfer",
-      args,
-      gasMultiplier,
-      this.l1SubmitValue,
-      simMode
-    );
+    if (this.isL1TokenUsdc(l1Token) && this.isL2TokenUsdc(l2Token)) {
```
nicholaspai, line 199: should we throw if either `this.isL1TokenUsdc(l1Token)` or `this.isL2TokenUsdc(l2Token)` is true but the other is false?

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -215,33 +234,37 @@ export class PolygonAdapter extends BaseAdapter {
...
     address: string,
     l1Token: string,
     l2Token: string,
     amount: BigNumber,
     simMode = false
   ): Promise<TransactionResponse> {
-    let method = "depositFor";
-    // note that the amount is the bytes 32 encoding of the amount.
-    let args = [address, l1Token, bnToHex(amount)];
+    if (this.isL1TokenUsdc(l1Token) && this.isL2TokenUsdc(l2Token)) {
```
nicholaspai, line 244: same comment as https://github.com/across-protocol/relayer-v3/pull/1424/files#r1581102376

test/AdapterManager.SendTokensCrossChain.ts
```diff
@@ -108,14 +118,13 @@ describe("AdapterManager: Send tokens cross-chain", async function () {
   });
   it("Correctly sends tokens to chain: Optimism", async function () {
     const chainId = CHAIN_IDs.OPTIMISM;
-    //  ERC20 tokens:
+    //  CCTP tokens:
     await adapterManager.sendTokenCrossChain(relayer.address, chainId, mainnetTokens.usdc, amountToSend);
```
nicholaspai, line 122: We should not remove the ERC20 test, right? I think its important we still unit test that `depositERC20` is called with a nonUSDC token

test/AdapterManager.SendTokensCrossChain.ts
```diff
@@ -256,14 +265,13 @@ describe("AdapterManager: Send tokens cross-chain", async function () {
   });
   it("Correctly sends tokens to chain: Base", async function () {
     const chainId = CHAIN_IDs.BASE;
-    //  ERC20 tokens:
+    //  CCTP tokens:
     await adapterManager.sendTokenCrossChain(relayer.address, chainId, mainnetTokens.usdc, amountToSend);
-    expect(l1BaseBridge.depositERC20).to.have.been.calledWith(
```
nicholaspai, line 261: same suggestion as https://github.com/across-protocol/relayer-v3/pull/1424/files#r1581109028

test/AdapterManager.SendTokensCrossChain.ts
```diff
@@ -314,7 +322,6 @@ async function constructChainSpecificFakes() {
   l1AtomicDepositor = await makeFake("atomicDepositor", CONTRACT_ADDRESSES[1].atomicDepositor.address);
 
   // Optimism contracts
-  l1OptimismBridge = await makeFake("ovmStandardBridge_10", CONTRACT_ADDRESSES[1].ovmStandardBridge_10.address);
```
nicholaspai, line 317: is this contract not tested?

test/AdapterManager.SendTokensCrossChain.ts
```diff
@@ -108,14 +118,13 @@ describe("AdapterManager: Send tokens cross-chain", async function () {
   });
   it("Correctly sends tokens to chain: Optimism", async function () {
```
nicholaspai, line 119: Are we missing a Polygon CCTP unit test?

nicholaspai (commented):
Left some more suggestions!

test/AdapterManager.SendTokensCrossChain.ts
james-a-morris: It was tested by the one call to `depositERC20`. This has been brought back in a recent commit

test/AdapterManager.SendTokensCrossChain.ts
james-a-morris: We can use BAL to test the default ERC20 since USDC will now be transmitted via CCTP

test/AdapterManager.SendTokensCrossChain.ts
james-a-morris: https://github.com/across-protocol/relayer-v3/pull/1424#discussion_r1581417425

src/utils/CCTPUtils.ts
james-a-morris: It's not a huge difference. One is an explicit type cast of the variable and the other is an implicit typecast of the variable based on the assignment. I'll change this to `resultingCall: BigNumber` instead.

src/clients/bridges/ArbitrumAdapter.ts
james-a-morris: Good call - we now check for `this.isL1TokenUsdc(l1Token)` and the underlying call to `sendCctpTokenToTargetChain` will throw if `l2Token !== native usdc`

src/clients/bridges/PolygonAdapter.ts
james-a-morris: https://github.com/across-protocol/relayer-v3/pull/1424#discussion_r1581478653

james-a-morris:
@nicholaspai @pxrl added a class `UsdcTokenSplitterBridge.ts` to internally choose between CCTP or the canonical bridge when handling USDC. This is needed to avoid bloating the op stack adapter

src/clients/bridges/CCTPAdapter.ts
james-a-morris: Done

src/clients/bridges/PolygonAdapter.ts
james-a-morris: Yes, but it may make sense to keep this here prior to refactoring the code to look more OPStack Adapter-like (per this comment: https://github.com/across-protocol/relayer-v3/pull/1424#discussion_r1579407919 )

test/AdapterManager.SendTokensCrossChain.ts
nicholaspai: well isn't USDC only tramistted via CCTP if the l2 token is native USDC, otherwise it uses canonical bridge to get USDC.e?

nicholaspai: also sorry where is the ERC20 test, do we need to add BAL somewhere? You could also use WBTC since its already set in this test

nicholaspai (commented):
This almost LGTM but I think you're missing either your latest unit test commits or haven't done them yet

test/AdapterManager.SendTokensCrossChain.ts
james-a-morris: I pushed the most recent tests - this should show both USDC -> USDC.e/USDbC via the canonical bridges & BAL -> BAL via canonical

nicholaspai (approved):

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -159,45 +178,55 @@ export class ArbitrumAdapter extends BaseAdapter {
     // Note that if the token trying to be approved is not configured in this client (i.e. not in the l1Gateways object)
     // then this will pass null into the checkAndSendTokenApprovals. This method gracefully deals with this case.
     const associatedL1Bridges = l1Tokens
-      .map((l1Token) => {
+      .flatMap((l1Token) => {
```
pxrl, line 181: TIL!

src/clients/bridges/op-stack/UsdcCCTPBridge.ts
```diff
@@ -0,0 +1,83 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _fromAddress: string,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _eventConfig: EventSearchConfig
+  ): Promise<Event[]> {
+    // Per the documentation of the BaseAdapter's computeOutstandingCrossChainTransfers method, we can return an empty array here
+    // and only return the relevant outstanding events from queryL1BridgeInitiationEvents.
+    // Relevant link: https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/BaseAdapter.ts#L189
+    return Promise.resolve([]);
+  }
```
pxrl, line 82: It should alternatively work fine to just name the vars in the function body in order to get lint off your back. Might be nice here since there are 3 exclusions.

```suggestion
    l1Token: string,
    fromAddress: string,
    eventConfig: EventSearchConfig
  ): Promise<Event[]> {
    // lint appeasement
    l1Token;
    fromAddress;
    eventConfig;
  
    // Per the documentation of the BaseAdapter's computeOutstandingCrossChainTransfers method, we can return an empty array here
    // and only return the relevant outstanding events from queryL1BridgeInitiationEvents.
    // Relevant link: https://github.com/across-protocol/relayer-v2/blob/master/src/clients/bridges/BaseAdapter.ts#L189
    return Promise.resolve([]);
  }
```

src/utils/CCTPUtils.ts
```diff
@@ -17,6 +19,107 @@ export type DecodedCCTPMessage = {
...
+    initializationTransactions.map(async (event) => {
+      const { nonce, destinationDomain } = event.args;
+      // Ensure that the destination domain matches the target destination domain so that we don't
+      // have any double counting of messages.
+      if (destinationDomain !== targetDestinationDomain) {
+        return undefined;
+      }
+      // Call into the destinationMessageTransmitter contract to determine if the message has been processed
+      // on the destionation chain. We want to make sure the message **hasn't** been processed.
+      const isMessageProcessed = await hasCCTPMessageBeenProcessed(sourceDomain, nonce, destinationMessageTransmitter);
```
pxrl, line 94: wdyt about a follow-on change to aggregate the inputs and then make a single multicall, vs. querying in parallel?

pxrl (approved):
Super clean - really nice work!

src/utils/CCTPUtils.ts
james-a-morris: That could be good follow-on!

src/clients/bridges/op-stack/UsdcCCTPBridge.ts
james-a-morris: Done! https://github.com/across-protocol/relayer-v3/pull/1424/commits/4e94fb823fc9a407eebb3aaab5b64e444a685529

PR #1425, opened 2024-04-23
improve(Dataworker#index): Detect colliding `proposeRootBundle` calls

nicholaspai:
Whenever a proposeRootBundle collides with itself, it sends a PD alert which is noisy.


pxrl:
ooc, is this typically due to overlapping proposer runs?

src/dataworker/index.ts
```diff
@@ -167,6 +167,37 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+      });
+      loopStart = performance.now();
+
+      // The dataworker loop takes a long-time to run, so if the proposer is enabled, run a final check and early
+      // exit if a proposal is already pending.
+      const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
+      if (
+        config.proposerEnabled &&
+        isDefined(bundleDataToPersist) &&
+        pendingProposal.unclaimedPoolRebalanceLeafCount.toString() === "0"
```
pxrl, line 189: Should this be !== 0 ?

pxrl (approved):
LGTM subject to the unclaimedPoolRebalanceLeafCount q.

src/clients/MultiCallerClient.ts
pxrl, line 1: Also, should we try w/ the check on whether a proposal exists first? I suspect we won't need this suppression as a result, so it'd be nice to avoid that if possible.

src/dataworker/index.ts
nicholaspai: Doh

PR #1426, opened 2024-04-23
feat(Relayer): Consider all eligible repayment chains if top one is unprofitable

nicholaspai:
Currently the algorithm is to consider destination chain only if top preferred chain is unprofitable and not equal to destination chain

This PR changes the algo to have `determineRefundChain` to return all eligible repayment chain so that the relayer considers them all before falling back to destination chain, assuming that destination chain wasn't already considered

This PR ultimately will give us better relayer profitability (because we evaluate profitability on multiple repayment chains before falling back to a potentially unprofitable one) but it will add some marginal run-time when evaluating all eligible repayment chains

### DIFF
--- a/src/finalizer/utils/arbitrum.ts
+++ b/src/finalizer/utils/arbitrum.ts
@@ -11,6 +11,8 @@ import {
   getRedisCache,
   getBlockForTimestamp,
   getL1TokenInfo,
+  compareAddressesSimple,
+  TOKEN_SYMBOLS_MAP,
 } from "../../utils";
 import { TokensBridged } from "../../interfaces";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
@@ -139,17 +141,24 @@ async function getAllMessageStatuses(
   // This is important for bridge transactions containing multiple events.
   const logIndexesForMessage = getUniqueLogIndex(tokensBridged);
   return (
-    await Promise.all(
-      tokensBridged.map((e, i) => getMessageOutboxStatusAndProof(logger, e, mainnetSigner, logIndexesForMessage[i]))
+    (
+      await Promise.all(
+        tokensBridged.map((e, i) => getMessageOutboxStatusAndProof(logger, e, mainnetSigner, logIndexesForMessage[i]))
+      )
     )
-  )
-    .map((result, i) => {
-      return {
-        ...result,
-        info: tokensBridged[i],
-      };
-    })
-    .filter((result) => result.message !== undefined);
+      .map((result, i) => {
+        return {
+          ...result,
+          info: tokensBridged[i],
+        };
+      })
+      // USDC withdrawals for Arbitrum should be finalized via the CCTP Finalizer.
+      .filter(
+        (result) =>
+          result.message !== undefined &&
+          !compareAddressesSimple(result.info.l2TokenAddress, TOKEN_SYMBOLS_MAP["_USDC"].addresses[CHAIN_ID])
+      )
+  );
 }
 
 async function getMessageOutboxStatusAndProof(
--- a/src/finalizer/utils/linea/l1ToL2.ts
+++ b/src/finalizer/utils/linea/l1ToL2.ts
@@ -40,9 +40,7 @@ export async function lineaL1ToL2Finalizer(
   );
 
   // Optimize block range for querying Linea's MessageSent events on L1.
-  // We want to conservatively query for events that are between 0 and 24 hours old
-  // because Linea L1->L2 messages are claimable after ~20 mins.
-  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l1ChainId, 24, 0);
+  const { fromBlock, toBlock } = await getBlockRangeByHoursOffsets(l1ChainId, 24 * 7, 0);
   logger.debug({
     at: "Finalizer#LineaL1ToL2Finalizer",
     message: "Linea MessageSent event filter",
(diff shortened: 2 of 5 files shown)


src/relayer/Relayer.ts
```diff
@@ -565,34 +565,84 @@ export class Relayer {
...
+    let preferredChainId: number | undefined;
+    let lpFeePct: BigNumber | undefined;
+    let relayerFeePct: BigNumber | undefined;
+    let gasLimit: BigNumber | undefined;
+    let gasCost: BigNumber | undefined;
+    let profitable = false;
+
+    // Save this for later in case none of the preferred chains are profitable and destination chain is profitable,
+    // then use this to calculate the virtual delta in the gross relayer fee.
+    const topPreferredChainProfitability = await profitClient.isFillProfitable(
```
mrice32, line 599: Might be a bad suggestion, but when coming up with the preferred list, why not just add the destination chain to the end of the list if it's not already in there?

Then, could we remove a few of these big if/elses below?

I get that we would lose some information about whether it was truly a "preferred" chain, but the meaning of preferred is sort of ambiguous anyway (since chains can be preferred for different reasons, no?).

If we did this, I think the flow would just be:
- Search for the first profitable chain in the list.
- If one is found, log it and return it.
- If none are found and you exit the loop, log that and return the first one.

Thoughts?

mrice32 (commented):
Looks good! Just one suggestion on a potential refactor

src/relayer/Relayer.ts
nicholaspai: > Might be a bad suggestion, but when coming up with the preferred list, why not just add the destination chain to the end of the list if it's not already in there?

we already do this actually

The reason why we have the big if-else on line 640 is to not mark a deposit as "unprofitable" and ignore it if the depositor set the relayer fee high enough assuming the relayer would take repayment on the destination chain. This is because the API assumes repayment will be on destination chain and quotes fees accordingly.

If we remove that last part, then we might unfairly penalize honest depositors/users of the API.

> Search for the first profitable chain in the list.
> If one is found, log it and return it.
> If none are found and you exit the loop, log that and return the first one.

The problem with this flow is that the if we always return the first chain as repayment chain then we'll never mark a deposit as unprofitable and just fill everything. I imagine we still want to ignore truly unprofitable deposits (e.g. setting inputAmount == outputAmount)

src/relayer/Relayer.ts
```diff
@@ -565,34 +565,84 @@ export class Relayer {
...
+      deposit,
+      lpFeePcts[0],
+      hubPoolToken,
+      preferredChainIds[0]
+    );
+
+    for (let i = 0; i < preferredChainIds.length; i++) {
```
james-a-morris, line 606: Can this be done in parallel?

nicholaspai: yeah maybe. It means extra computations but better worst-case speed compared to looping through all chai s

src/relayer/Relayer.ts
mrice32: > The reason why we have the big if-else on line 640 is to not mark a deposit as "unprofitable" and ignore it if the depositor set the relayer fee high enough assuming the relayer would take repayment on the destination chain. This is because the API assumes repayment will be on destination chain and quotes fees accordingly.
>
> If we remove that last part, then we might unfairly penalize honest depositors/users of the API.

If all are unprofitable, doesn't that imply that the destination chain is unprofitable since you said the destination chain is already in the list?

> The problem with this flow is that the if we always return the first chain as repayment chain then we'll never mark a deposit as unprofitable and just fill everything. I imagine we still want to ignore truly unprofitable deposits (e.g. setting inputAmount == outputAmount)

Don't we return something in the current algorithm even if the deposit is unprofitable? Can't we do the same thing in the case we exit the loop?

If we exit the loop with no profitable chains, we know that the deposit should not be filled, so we can return whatever we do in the current algo in that case. If we find a profitable chain, we return it.

Maybe the issue here is that the fallback case, we want to use the profitability of the destination chain, but use the refund location of the first result. If that's the core issue, it might be simpler to just encode all of that in the element: profitability details of the destination chain, refund location of the first chain. That might be a more invasive refactor, so I understand if it doesn't make sense.


My only reason for raising this is that I think the repayment logic requirements are getting pretty sophisticated/complex, and I want to push on trying to simplify our logic as much as possible to keep the code from getting too complex to manage.

mrice32 (approved):

src/relayer/Relayer.ts
nicholaspai: > If all are unprofitable, doesn't that imply that the destination chain is unprofitable since you said the destination chain is already in the list?

No, its possible that the destination chain is over-allocated from the inventory management perspective so its not selected as a `preferredChainId`. In this case, we haven't evaluated profitability of the destination chain yet.

mrice32: > > If all are unprofitable, doesn't that imply that the destination chain is unprofitable since you said the destination chain is already in the list?
> 
> No, its possible that the destination chain is over-allocated from the inventory management perspective so its not selected as a `preferredChainId`. In this case, we haven't evaluated profitability of the destination chain yet.

Ah, makes sense. I thought you had said above that we do include the destination chain in the list.

mrice32 (approved):

src/relayer/Relayer.ts
nicholaspai: that's what i'm thinking. For now this PR isolates this logic "repay on X, but evaluate with profitability of Y" so I think it makes it easier to refactor in the future

PR #1429, opened 2024-04-24
feat(Dataworker#index): Catch PoolRebalanceLeaf execution collisions

nicholaspai:
`executePoolRebalanceLeaves` sends `.error` level logs if it tries to execute pool rebalance leaves but gets front-run. SpokeLeaf execution failures do not send `.error` level logs because I believe they are allowed to fail in simulation (except HubChain leaves which get batched with pool leaf executions).

This PR adds a return value to `executePoolRebalanceLeaves` so the index.ts runner can know if any pool leaf executions have been enqueued. If any have been, then run a similar check to https://github.com/across-protocol/relayer-v3/pull/1425 and exit early if the root bundle has already been executed

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -175,8 +175,8 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
         const proposalCollision =
           isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
-        const executorCollision = 
-  pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== poolRebalanceLeafExecutionCount;
+        const executorCollision =
+        pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== poolRebalanceLeafExecutionCount;
         if (proposalCollision || executorCollision) {
           logger[startupLogLevel(config)]({
             at: "Dataworker#index",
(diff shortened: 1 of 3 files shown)


src/dataworker/index.ts
```diff
@@ -167,14 +168,28 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+      // @dev This assumes that this bot is running in a serverless setting
+      // and that there is the possibility of overlapping runs that could cause one run to "front-run" a previous
+      // bot instance.
+      const executeDataworkerTransactions = async () => {
         const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
-        if (isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0") {
+        const proposalCollision =
+          isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
+        // This assumes that all pool leaves get executed together in a single transaction. This should
+        // always be the case, which means that the unclaimed leaf count is always in either exactly two states:
```
pxrl, line 182: I think we have recently seen cases where the mainnet leaf didn't get executed on the first attempt due to a missing sync() call, but the others did.

src/dataworker/index.ts
```diff
@@ -167,14 +168,28 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+        // equal to the maximum unclaimed leaf count or 0.
+        const executorCollision =
+          poolRebalanceLeafExecutionCount > 0 && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() === "0";
```
pxrl, line 185: Could this just be a check for inequality on `poolRebalanceLeafExecutionCount !== pendingProposal.unclaimedPoolRebalanceLeafCount` ? I think then there's no need for assumptions about whether all leaves get executed atomically or not.

pxrl (approved):
Couple of minor comments, but I think this should work OK.

PR #1430, opened 2024-04-24
improve(InventoryClient#determineRefundChainId): Derive L1Token from inputToken instead of outputToken

nicholaspai:
Related to #ACX1967

When determining refund chain, the relayer cares (1) about its balance allocation across L2s and L1 and (2) reducing hub pool utilization.

- cumulative balance across all chains should be computed based on the L1 token linked to the inputToken since this is the token that will be refunded to the relayer.

- hub pool utilization also should be based on the L1 token linked to the inputToken.

I think this is the first PR needed to allow the relayer to determine its refund chain accurately when filling deposits where `inputToken = USDC` and `outputToken = USDC.e`.

Further PR's should address the following:
- [this check for token equivalency](https://github.com/across-protocol/relayer-v3/blob/8888fe2a97c952e3f5f392ff6115f979238aef51/src/clients/InventoryClient.ts#L289) needs to be removed. We might want to temporarily only allow it if the `inputToken = USDC` and the `outputToken = USDC.e[destinationChain]`
- The [post-relay](https://github.com/across-protocol/relayer-v3/blob/8888fe2a97c952e3f5f392ff6115f979238aef51/src/clients/InventoryClient.ts#L359) calculations for chain virtual balance need to be changed if the `outputToken != inputToken`


pxrl (approved):
Yep, this simply won't work post-CCTP when outputToken is USDC.e.

PR #1431, opened 2024-04-24
feat: Script to snapshot SpokePool EnabledDepositRoute events

pxrl:
This script is capable of scraping historical EnabledDepositRoute events from all SpokePools configured via the ConfigStore. It has two modes:

snapshot:
  Scrape all EnabledDepositRoute events and save them to a file.

verify:
  Read a snapshot from a file and verify it against the onchain state as
  at the point the snapshot was taken.

Example:
  yarn ts-node ./scripts/DepositRoutes.ts snapshot [--file <outputFile>]
  yarn ts-node ./scripts/DepositRoutes.ts verify [--file <outputFile>]

The HubPool chain ID can also optionally be supplied via the --chainId option.

The intention of this script is to facilitate periodic snapshots to be taken and stored in the relayer repository. The relayer would then load these snapshots directly from disk, rather than querying back to SpokePool genesis for each chain (which has ever-worse performance). Each snapshot stores the blockNumber at which is was taken on each chain, so the relayer is then able to request SpokePoolClient updates from that point.

Scraping these events manually currently takes ~14 seconds in local testing, even when the Redis cache has been primed with these events. Loading from disk should take < 1 second, and is especially efficient because only the events are retained in the file, whereas querying the RPC provider requires the entire block range to be evaluated.

If this snapshotting process is done monthly or quarterly then it will mitigate the incremental performance degradation that the relayer is constantly subjected to.

### DIFF
--- a/scripts/DepositRoutes.ts
+++ b/scripts/DepositRoutes.ts
@@ -1,0 +1,188 @@
+import minimist from "minimist";
+import { readFileSync, writeFileSync } from "fs";
+import { Event, providers as ethersProviders, utils as ethersUtils } from "ethers";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import {
+  CHAIN_IDs,
+  exit,
+  getDeploymentBlockNumber,
+  getNetworkName,
+  getProvider,
+  isDefined,
+  paginatedEventQuery,
+  sortEventsAscendingInPlace,
+} from "../src/utils";
+import * as utils from "./utils";
+
+const { NODE_SUCCESS, NODE_APP_ERR } = utils;
+
+type BlockTag = ethersProviders.BlockTag;
+
+type DepositRoute = {
+  blockNumber: number;
+  originToken: string;
+  destinationChainId: number;
+  enabled: boolean;
+  transactionHash: string;
+};
+
+type SpokePoolDepositRoutes = {
+  address: string;
+  toBlock: number;
+  depositRoutes: DepositRoute[];
+};
+
+type DepositRoutes = { [chainId: number]: SpokePoolDepositRoutes };
+
+async function getEnabledChainIds(hubChainId: number, blockTag: BlockTag): Promise<number[]> {
+  const provider = await getProvider(hubChainId);
+  const configStore = await utils.getContract(hubChainId, "AcrossConfigStore");
+  const CHAIN_ID_INDICES = ethersUtils.formatBytes32String("CHAIN_ID_INDICES");
+  const _chainIds = await configStore.connect(provider).globalConfig(CHAIN_ID_INDICES, { blockTag });
+  return JSON.parse(_chainIds.replaceAll('"', ""));
+}
+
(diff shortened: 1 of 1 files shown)


pxrl:
See this file for an example of the data that's saved in each snapshot:

[DepositRoutes.json](https://github.com/across-protocol/relayer-v3/files/15098048/DepositRoutes.json)

nicholaspai (commented):
Is this only intended for looping mode or also serverless?

pxrl:
> Is this only intended for looping mode or also serverless?

Both the relayer in serverless/one-shot mode and in looping mode. Scraping these events is currently a major source of slowdown for both relayer modes (less so for looping mode, but it still has an impact). Take a look at these stats coming out of the logs:

queries

Since it takes me ~14 seconds locally, I really wonder if these long update times are indicative of congestion on the interface over to redis, since we often have a handful of bots starting simultaneously and performing SpokePoolClient update at approximately the same point in time.

Also, these specific events are an important prerequisite for a number of blocking calls in the relayer - especially around checking token balances and approvals on L2s. If the events can be read in advance of SpokePoolClient update then we might be able to re-organise those updates so more can be done in parallel.

Specifically in the looping relayer w/ external indexer, I also see that this update does not always complete on some SpokePools prior to the relayer commencing its first loop. This means that the TokenClient isn't aware of its balances for those tokens, so the relayer is unable to make fills on those chains. The effect of this is confined to the destination chains where this update didn't complete though, so it does at least allow the relayer to complete fills on some chains in its first loop.

These snapshots could also be used for the dataworker personalities, but since we tend to be more conservative there and are slightly less sensitive in terms of performance, I don't think it's quite as important.

Also just to clarify about how this would be used by the relayer, I'd propose that the snapshot JSON files are stored in the relayer or sdk repos, and that it'd be periodically updated. The script makes it easy to verify the result, so we can have good confidence that the entries are correct on each update.

An alternative would be to store this _condensed_ event format directly in redis, rather than relying on the underlying RPC cache, which requires _a lot_ of queries on chains with very low block times. In terms of performance I think they'll be very similar; the real benefit comes from not having to scrape all the way back to genesis.

james-a-morris (commented):
What ENV variables need to be defined for this script?

scripts/DepositRoutes.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  return true;
+}
+
+function usage(badInput?: string): boolean {
+  let usageStr = badInput ? `\nUnrecognized input: "${badInput}".\n\n` : "";
+
+  usageStr += `
+    Usage:
+    \tyarn ts-node ./scripts/depositRoutes.ts snapshot [--chainId <hubChainId>] [--file <path-to-file>]
```
mrice32, line 138: In usage or in a comment, would be great to have a list of envs that the user needs to provide to make this work.

mrice32 (approved):
LGTM!

pxrl:
Closing this for now, because we have no immediate use for it.

PR #1432, opened 2024-04-24
improve(APIClient): Retain limits for 60 seconds

pxrl:
The new fast relayer is re-querying /limits on every loop, which is unnecessary and increases the chance of hitting a 3s timeout.


src/clients/AcrossAPIClient.ts
```diff
@@ -108,6 +112,7 @@ export class AcrossApiClient {
       limits: this.limits,
     });
     this.updatedLimits = true;
+    this.updateTimestamp = Math.round(Date.now() / 1000);
```
nicholaspai, line 115: Should this be equal to the value that we compare above?

nicholaspai (approved):

src/clients/AcrossAPIClient.ts
pxrl: Updated: https://github.com/across-protocol/relayer-v3/pull/1432/commits/6cf576b43056bb3a6ffb1a29d3f63e728516008a

src/clients/AcrossAPIClient.ts
```diff
@@ -38,8 +40,10 @@ export class AcrossApiClient {
   }
 
   async update(ignoreLimits: boolean): Promise<void> {
-    if (ignoreLimits) {
-      this.logger.debug({ at: "AcrossAPIClient", message: "Skipping querying /limits" });
+    const now = Math.round(Date.now() / 1000);
```
james-a-morris, line 43: Thoughts on using https://github.com/across-protocol/relayer-v2/blob/master/src/utils/TimeUtils.ts#L1-L7

james-a-morris (approved):
One minor nit

mrice32 (approved):
LGTM!

src/clients/AcrossAPIClient.ts
pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1432/commits/9ec0a09af706fd4313d3dce0b202324dbc627a3b

PR #1435, opened 2024-04-24
fix(Dataworker#index): Don't submit double arweave txns

nicholaspai:
#1425 avoids duplicate proposals but I forgot to add duplicate arweave submission protection

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -154,11 +154,21 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
+      // @dev The dataworker loop takes a long-time to run, so if the proposer is enabled, run a final check and early
+      // exit if a proposal is already pending. Similarly, the executor is enabled and if there are pool rebalance
+      // leaves to be executed but the proposed bundle was already executed, then exit early.
+      const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
+      const proposalCollision =
+        isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
+      const executorCollision =
+        poolRebalanceLeafExecutionCount > 0 &&
+        pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== poolRebalanceLeafExecutionCount.toString();
+
       // Define a helper function to persist the bundle data to the DALayer.
       const persistBundle = async () => {
         // Submit the bundle data to persist to the DALayer if persistingBundleData is enabled.
         // Note: The check for `bundleDataToPersist` is necessary for TSC to be happy.
-        if (config.persistingBundleData && isDefined(bundleDataToPersist)) {
+        if (config.persistingBundleData && !proposalCollision && isDefined(bundleDataToPersist)) {
           await persistDataToArweave(
             clients.arweaveClient,
             bundleDataToPersist,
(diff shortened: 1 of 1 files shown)


src/dataworker/index.ts
```diff
@@ -154,11 +154,21 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
+      // @dev The dataworker loop takes a long-time to run, so if the proposer is enabled, run a final check and early
+      // exit if a proposal is already pending. Similarly, the executor is enabled and if there are pool rebalance
+      // leaves to be executed but the proposed bundle was already executed, then exit early.
+      const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
+      const proposalCollision =
+        isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
```
james-a-morris, line 162: OOC, why are we converting this to a string?

src/dataworker/index.ts
```diff
@@ -154,11 +154,21 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
+      // @dev The dataworker loop takes a long-time to run, so if the proposer is enabled, run a final check and early
+      // exit if a proposal is already pending. Similarly, the executor is enabled and if there are pool rebalance
+      // leaves to be executed but the proposed bundle was already executed, then exit early.
+      const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
```
nicholaspai, line 160: This is copied from below

src/dataworker/index.ts
```diff
@@ -154,11 +154,21 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+      const proposalCollision =
+        isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
+      const executorCollision =
```
james-a-morris, line 163: Do we need to move this conditional outside `executeDataworkerTransactions`? It isn't used in the new logic below

src/dataworker/index.ts
nicholaspai: because its not clear what the type is when grabbing from on-chain. It's probably a number but its not clear unless we blindly cast it. The HubPoolClient casts it as a number so we can probably just do that

james-a-morris: Fair enough! LGTM

james-a-morris (approved):

src/dataworker/index.ts
```diff
@@ -154,11 +154,18 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
+      const pendingProposal = await clients.hubPoolClient.hubPool.rootBundleProposal();
+      const proposalCollision =
+        isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0";
+
       // Define a helper function to persist the bundle data to the DALayer.
       const persistBundle = async () => {
         // Submit the bundle data to persist to the DALayer if persistingBundleData is enabled.
         // Note: The check for `bundleDataToPersist` is necessary for TSC to be happy.
-        if (config.persistingBundleData && isDefined(bundleDataToPersist)) {
+        if (config.persistingBundleData && !proposalCollision && isDefined(bundleDataToPersist)) {
```
james-a-morris, line 168: We could put the check for `pendingProposal.unclaimedPoolRebalanceLeafCount.toString() !== "0"` here to avoid needing to test `isDefined( )` twice

src/dataworker/index.ts
```diff
@@ -154,11 +155,17 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Executor disabled" });
       }
 
+      // @dev The dataworker loop takes a long-time to run, so if the proposer is enabled, run a final check and early
+      // exit if a proposal is already pending. Similarly, the executor is enabled and if there are pool rebalance
+      // leaves to be executed but the proposed bundle was already executed, then exit early.
+      const pendingProposal: PendingRootBundle = await clients.hubPoolClient.hubPool.rootBundleProposal();
+      const proposalCollision = isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount !== 0;
```
james-a-morris, line 162: very minor nit since `unclaimedPoolRebalanceLeafCount` is unsigned
```suggestion
      const proposalCollision = isDefined(bundleDataToPersist) && pendingProposal.unclaimedPoolRebalanceLeafCount > 0;
```

PR #1436, opened 2024-04-24
improve(Relayer): Don't log about shortfalls when not sending relays

nicholaspai:
Identical idea to https://github.com/across-protocol/relayer-v3/pull/1433 to quiet down logs when not sending relays


james-a-morris (approved):
It may be worth adding a utility class function that logs debug if `sendingRelaysEnabled` is false

PR #1438, opened 2024-04-25
feat(Dataworker): Detect duplicate relayer refund and slow fill leaf executions

nicholaspai:
Similar to the strategy in #1429 and #1425, detect duplicate leaf executions right before attempting to execute them

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1165,6 +1165,37 @@ export class Dataworker {
             throw new Error(`Leaf chainId does not match input chainId (${destinationChainId} != ${chainId})`);
           }
 
+          // @dev check if there's been a duplicate leaf execution and if so, then exit early.
+          // Since this function is happening near the end of the dataworker run and leaf executions are
+          // relatively infrequent, the additional RPC latency and cost is acceptable.
+          // @dev Can only filter on indexed events.
+          const eventFilter = client.spokePool.filters.FilledV3Relay(
+            null, // inputToken
+            null, // outputToken
+            null, // inputAmount
+            null, // outputAmount
+            null, // repaymentChainId
+            slowFill.relayData.originChainId,
+            slowFill.relayData.depositId,
+            null, // fillDeadline,
+            null, // exclusivityDeadline,
+            null, // exclusiveRelayer,
+            null, // relayer,
+            null, // depositor,
+            null, // recipient,
+            null, // message
+            null //  relayExecutionInfo
+          );
+          const duplicateEvents = await client.spokePool.queryFilter(eventFilter);
+          if (duplicateEvents.length > 0) {
+            this.logger.debug({
+              at: "Dataworker#executeSlowRelayLeaves",
+              message: "Leaf already executed",
+              duplicateEvents,
+            });
+            return undefined;
+          }
+
           const { outputAmount } = slowFill.relayData;
           const fill = latestFills[idx];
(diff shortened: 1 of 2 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1165,6 +1165,37 @@ export class Dataworker {
...
+          );
+          const duplicateEvents = await client.spokePool.queryFilter(eventFilter);
+          if (duplicateEvents.length > 0) {
+            this.logger.debug({
+              at: "Dataworker#executeSlowRelayLeaves",
+              message: "Leaf already executed",
+              duplicateEvents,
+            });
+            return undefined;
+          }
```
pxrl, line 1197: How about using fillStatus() instead? It'll ensure that the relaydata matches 100%, whereas this filter might throw up false positives in the event that a relayer makes an invalid fill for the same originChainId and depositId combination.

src/dataworker/Dataworker.ts
```diff
@@ -2015,6 +2046,29 @@ export class Dataworker {
...
+          // @dev Can only filter on indexed events.
+          const eventFilter = client.spokePool.filters.ExecutedRelayerRefundRoot(
+            null, // amountToReturn
+            leaf.chainId,
+            null, // refundAmounts
+            rootBundleId,
+            leaf.leafId,
+            null, // l2TokenAddress
+            null, // refundAddresses
+            null // msg.sender of the leaf execution.
```
pxrl, line 2061: Should be OK drop these, since afaik Ethers will just infer that they are null.

src/dataworker/Dataworker.ts
```diff
@@ -1165,6 +1166,19 @@ export class Dataworker {
             throw new Error(`Leaf chainId does not match input chainId (${destinationChainId} != ${chainId})`);
           }
 
+          // @dev check if there's been a duplicate leaf execution and if so, then exit early.
+          // Since this function is happening near the end of the dataworker run and leaf executions are
+          // relatively infrequent, the additional RPC latency and cost is acceptable.
+          const relayDataHash = sdkUtils.getRelayDataHash(slowFill.relayData, destinationChainId);
+          const fillStatus: BigNumber = await client.spokePool.fillStatuses(relayDataHash);
+          if (fillStatus.toNumber() === FillStatus.Filled) {
```
pxrl, line 1174: If you use `relayFillStatus()` from `sdk-v2/utils/SpokeUtils` then it gets the relayData hash handles the conversion to a number for you - so you could just do:

`if (await relayFillStatus(client.spokePool, slowFill.relayData, "latest", destinationchainId) === FillStatus.Filled)`.

pxrl (approved):
Looks good; the check on fillStatuses() can be simplified a bit by using an SDK helper.

src/dataworker/Dataworker.ts
```diff
@@ -2016,6 +2034,29 @@ export class Dataworker {
...
+            null, // amountToReturn
+            leaf.chainId,
+            null, // refundAmounts
+            rootBundleId,
+            leaf.leafId
+          );
+          const duplicateEvents = await client.spokePool.queryFilter(
+            eventFilter,
+            client.latestBlockSearched - (client.eventSearchConfig.maxBlockLookBack ?? 5_000)
+          );
```
james-a-morris, line 2051: Would it make sense to use a call to `paginatedEventQuery` here? Also - is there a situation where the maxblocklookback is not defined?

PR #1440, opened 2024-04-25
fix(test): Prevent querying redis in TokenClient test

pxrl:


pxrl:
fwiw I've been plagued by local test failures in the TokenClient area in recent weeks. They never occur in prod because the TokenClient gracefully falls back if redis is not available. I'm hopeful this will address the issue when redis is available locally.

nicholaspai (approved):

src/clients/TokenClient.ts
```diff
@@ -283,4 +284,8 @@ export class TokenClient {
     }
     return hasData;
   }
+
+  protected async getRedis(): Promise<CachingMechanismInterface | undefined> {
```
mrice32, line 288: Quick question: should we go one level deeper and detect test runs in getRedisCache and return undefined there so that it applies to all tests?

Note: maybe this is hard because there's no way to detect this via the env. Just wanted to mention it in case it would be cleaner to do it at the source.

mrice32 (approved):
LGTM! Just one question/suggestion

src/clients/TokenClient.ts
pxrl: Yeah..that's a good suggestion and would be nice. Do you know if hardhat exports any variables that we could check for? If not, I guess we could inject this into the `yarn test` script target - something like:

`"test": "export RELAYER_TEST=true; hardhat test"`

wdyt?

pxrl: I think your proposal is the most watertight approach - I've proposed it in another PR and will close this one if it gets merged: https://github.com/across-protocol/relayer-v3/pull/1442

PR #1441, opened 2024-04-25
improve(TokenClient): Use HubPool events for querying balances

pxrl:
The TokenClient currently relies on SpokePoolClient.getAllOriginTokens(), which is backed by SpokePoolClient EnabledDepositRoute events. This means that TokenClient update is dependent on SpokePoolClient update having completed. If TokenClient update is initiated before SpokePoolClient update completed, the TokenClient assumes it has 0 balance for any tokens on the relevant chain, and the relayer is unable to make any fills in that run.

The tokens supported by SpokePools are also available via the HubPoolClient. This change updates the TokenClient to source its SpokePool token addresses from the HubPoolClient instead. This permits it to be updated concurrently with the SpokePoolClient and removes the hard dependency of the TokenClient on SpokePoolClient update having completed.

This gives a major benefit to the fast relayer in looping mode with an external indexer, but should also eventually permit the relayer to skip querying EnabledDepositRoute events, which should eliminate a significant source of RPC queries against each chain on startup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "lint-fix": "yarn eslint --fix && yarn prettier --write",
     "prettier": "prettier .",
     "eslint": "eslint .",
-    "test": "hardhat test",
+    "test": "RELAYER_TEST=true hardhat test",
     "build": "tsc --build",
     "watch": "tsc --build --incremental --watch",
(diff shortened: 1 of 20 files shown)


nicholaspai (approved):
I like this a lot. This pr seems like a nice drop in replacement

PR #1442, opened 2024-04-25
fix(test): Prevent Redis from being used in test

pxrl:
Redis has unintentionally been used in the TokenClient tests when tested in a local development environment. It hasn't affected CI because there's no redis cache available there.


mrice32 (approved):
LGTM!

PR #1444, opened 2024-04-25
refactor(relayer): Re-org client updates

pxrl:
The TokenClient is no longer dependent on SpokePoolClient update, so it can occur in parallel. Token approvals can also be handled in parallel with inventory client updates. Other updates with no dependencies can also be started early and awaited late in order to get the benefit of background execution.


src/relayer/RelayerClientHelper.ts
```diff
@@ -203,20 +203,27 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
...
-  // Update the token client first so that inventory client has latest balances.
-  await clients.tokenClient.update();
+  // Update the token client before the inventory client has latest balances.
+  await Promise.all([updateSpokePoolClients(spokePoolClients, spokePoolEvents), clients.tokenClient.update()]);
+
+  const setTokenApprovals = async (): Promise<void> => {
+    if (config.sendingRelaysEnabled) {
+      await clients.tokenClient.setOriginTokenApprovals();
+    }
+  };
```
pxrl, line 222: I just noticed that we also call `inventoryClient.setL1TokenApprovals()` below, and it seems to ripple out to calls on each l2. Currently looking at how `setOriginTokenApprovals()` differs from `setL1TokenApprovals()` :eyes:

src/clients/AcrossAPIClient.ts
```diff
@@ -73,11 +75,15 @@ export class AcrossApiClient {
...
-            return (
-              chainId !== CHAIN_IDs.MAINNET &&
-              mainnetSpokePoolClient.isDepositRouteEnabled(l1Token, chainId) &&
-              Object.keys(this.spokePoolClients).includes(chainId.toString())
-            );
+            try {
+              // Verify that a token mapping exists on the destination chain.
+              hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, chainId); // throws if not found.
+              return (
+                chainId !== hubPoolClient.chainId && Object.keys(this.spokePoolClients).includes(chainId.toString())
```
james-a-morris, line 82: Thoughts on removing `Object.keys` to reduce the O(N) conversion?

```suggestion
                chainId !== hubPoolClient.chainId && isDefined(this.spokePoolClients[chainId])
```

src/clients/TokenClient.ts
```diff
@@ -203,54 +209,53 @@ export class TokenClient {
...
+  async fetchTokenData(
+    chainId: number,
+    hubPoolTokens: string[],
+    signer: Signer
+  ): Promise<Record<string, { balance: BigNumber; allowance: BigNumber }>> {
+    const tokens = hubPoolTokens
+      .map((address) => {
+        try {
+          const spokePoolToken = this.hubPoolClient.getL2TokenForL1TokenAtBlock(address, chainId);
+          return new Contract(spokePoolToken, ERC20.abi, signer);
```
james-a-morris, line 221: It may be worth using the ERC20 class directly - feel free to ignore
```suggestion
          return ERC20.connect(spokePoolToken, signer)
```

pxrl: This is nice - I'll propose it for a separate PR since this one doesn't touch the TokenClient anymore.

src/clients/AcrossAPIClient.ts
pxrl: Good call; though that code was actually from a preceding PR that was refactored before merge; it's not part of the current change anymore.

src/relayer/RelayerClientHelper.ts
```diff
@@ -192,36 +191,35 @@ export async function constructRelayerClients(
...
+  // InventoryClient updates depend on the tokenClient and SpokePoolClients.
+  await Promise.all([clients.tokenClient.update(), updateSpokePoolClients(spokePoolClients, spokePoolEvents)]);
 
-  // We can update the inventory client at the same time as checking for eth wrapping as these do not depend on each other.
   await Promise.all([
-    clients.acrossApiClient.update(config.ignoreLimits),
-    clients.inventoryClient.update(),
-    clients.inventoryClient.wrapL2EthIfAboveThreshold(),
-    clients.inventoryClient.setL1TokenApprovals(),
+    inventoryClient.update(),
```
pxrl, line 217: The inventory client is still dependent on SpokePoolClient update because it uses SpokePoolClient event search settings.

nicholaspai (approved):

src/relayer/RelayerClientHelper.ts
```diff
@@ -192,36 +191,35 @@ export async function constructRelayerClients(
...
     "RequestedSpeedUpV3Deposit",
     "FilledV3Relay",
     "RelayedRootBundle",
     "ExecutedRelayerRefundRoot",
-  ]);
+  ];
+
+  // Start updates w/ no dependencies first, and wait on them last.
+  const profitClientUpdate = profitClient.update();
+  const inputTokenApprovals = config.sendingRelaysEnabled ? tokenClient.setOriginTokenApprovals() : async () => true;
```
mrice32, line 210: I think the types are a bit weird here, but this coincidentally works. Suggested a correction to match the types.

```suggestion
  const inputTokenApprovals = config.sendingRelaysEnabled ? tokenClient.setOriginTokenApprovals() : Promise.resolve();
```

src/relayer/RelayerClientHelper.ts
```diff
@@ -192,36 +191,35 @@ export async function constructRelayerClients(
...
 
-  // We can update the inventory client at the same time as checking for eth wrapping as these do not depend on each other.
   await Promise.all([
-    clients.acrossApiClient.update(config.ignoreLimits),
-    clients.inventoryClient.update(),
-    clients.inventoryClient.wrapL2EthIfAboveThreshold(),
-    clients.inventoryClient.setL1TokenApprovals(),
+    inventoryClient.update(),
+    inventoryClient.wrapL2EthIfAboveThreshold(),
+    inventoryClient.setL1TokenApprovals(), // Approve bridge contracts (if rebalancing enabled)
```
mrice32, line 219: Sort of surprised that these don't depend on the inventory client being updated? Is that really the case?

If so, do they depend on anything else above? Or are they completely parallelizable with everything else?

mrice32 (approved):
LGTM! Just minor comments/questions

src/clients/ProfitClient.ts
```diff
@@ -593,17 +593,23 @@ export class ProfitClient {
...
-        destinationChainId === hubPoolClient.chainId
-          ? hubToken
-          : hubPoolClient.getL2TokenForL1TokenAtBlock(hubToken, destinationChainId);
-      assert(isDefined(outputToken), `Chain ${destinationChainId} SpokePool is not configured for ${testSymbol}`);
-
-      const deposit = { ...sampleDeposit, destinationChainId, outputToken };
-      this.totalGasCosts[destinationChainId] = await this._getTotalGasCost(deposit, relayer);
-    });
+    await sdkUtils.mapAsync(
+      enabledChainIds.filter(
```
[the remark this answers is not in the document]
pxrl: It's not explicitly called out in the description, but it's needed to avoid re-estimating the standard fill cost on subsequent ProfitClient updates. Per existing implementation, the ProfitClient is only updated once per run, whereas now it's updated once per loop. In most cases each subsequent update will effectively be a no-op, but in the event that simulation for a particular chain failed on the first update, it would get an opportunity for retry on the subsequent update.

src/relayer/RelayerClientHelper.ts
pxrl: > Sort of surprised that these don't depend on the inventory client being updated? Is that really the case?

Yeah, it seems to be. It's basically a thin layer of orchestration over the AdapterManager, sourcing the L1 tokens from the HubPoolClient (...so it has a hard dependency on the HubPoolClient being updated). It's possibly something that could be factored out of the InventoryClient from a technical/dependency standpoint, though since it concerns approving L1 bridge contracts I can also see why it's at home in the InventoryClient.

mrice32: The only long-term fix for making these dependency chains more clear/obvious is forcing them into the type system, which is a big project.

There may be downsides with this approach, but my thought is that you could make all these clients immutable and constructable via an async static function that takes all the dependencies in as args (including an older version of the same client that you want to use as a "base" for the update!).

In that setup, your dependencies would be clear, as they would all be args, and it would be impossible to provide a partially or un-updated client to something that depends on it, since there is no such thing as an un-updated client.

pxrl:
This is no longer relevant.

PR #1445, opened 2024-04-26
improve: Don't scrape EnabledDepositRoutes in external indexer

pxrl:
SetPoolRebalanceRoute events are now used to infer SpokePool token addresses, so EnabledDepositRoute events are no longer required. This should enable the fast relayer to consistently resolve its token balances on first loop, and should reduce its average fill times.


nicholaspai (approved):
Can we also do this in the serverless relayer?

PR #1446, opened 2024-04-26
fix(AcrossAPIClient): Reverse limits query direction

pxrl:
When querying limits, the API client current uses mainnet as the origin chain. The buffers applied on deposits originating from mainnet are typically negative, implying that Across can support _more_ volume than is available as liquid reserve in the HubPool. This effectively renders the limits check redundant and could cause the relayer to overcommit the HubPool by incorrectly making a fill.

While here, apply a slight refactor in recognition that API client doesn't need require SpokePoolClient instances anymore, so drop them in favour of a simple chain ID list. This removes the current constraint that mainnet needs to be defined as a destination chain. This adjustment makes the relayer more flexible and should be popular with third-parties.

Fixes ACX-2100

### DIFF
--- a/src/clients/AcrossAPIClient.ts
+++ b/src/clients/AcrossAPIClient.ts
@@ -64,7 +64,7 @@ export class AcrossApiClient {
 
     // /limits
     // Store the max deposit limit for each L1 token. The origin chain can be any supported chain
-    // expect the HubPool chain. This assumes the worst-case bridging delay of SpokePool -> mainnet.
+    // expect the HubPool chain. This assumes the worst-case bridging delay of !mainnet -> mainnet.
     const data = await Promise.all(
       tokensQuery.map((l1Token) => {
         const l2TokenAddresses = getL2TokenAddresses(l1Token);
--- a/src/clients/AcrossAPIClient.ts
+++ b/src/clients/AcrossAPIClient.ts
@@ -1,6 +1,5 @@
 import _ from "lodash";
 import axios, { AxiosError } from "axios";
-import { SpokePoolClientsByChain } from "../interfaces";
 import {
   bnZero,
   isDefined,
@@ -10,6 +9,7 @@ import {
   CHAIN_IDs,
   TOKEN_SYMBOLS_MAP,
   getRedisCache,
+  bnUint256Max as uint256Max,
 } from "../utils";
 import { HubPoolClient } from "./HubPoolClient";
 
@@ -19,7 +19,7 @@ export interface DepositLimits {
 
 export class AcrossApiClient {
   private endpoint = "https://app.across.to/api";
-
+  private chainIds: number[];
   private limits: { [token: string]: BigNumber } = {};
 
   public updatedLimits = false;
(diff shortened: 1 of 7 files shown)


src/clients/AcrossAPIClient.ts
```diff
@@ -28,17 +28,19 @@ export class AcrossApiClient {
...
-      this.tokensQuery = Object.entries(TOKEN_SYMBOLS_MAP).map(([, details]) => details.addresses[CHAIN_IDs.MAINNET]);
+      this.tokensQuery = Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[CHAIN_IDs.MAINNET]);
     }
+
+    this.chainIds = chainIds.filter((chainId) => chainId !== hubPoolClient.chainId);
   }
 
   async update(ignoreLimits: boolean): Promise<void> {
-    if (ignoreLimits) {
+    if (ignoreLimits || this.chainIds.length === 0) {
```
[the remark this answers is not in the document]
nicholaspai: can you add this as a comment

src/relayer/Relayer.ts
```diff
@@ -147,16 +147,17 @@ export class Relayer {
...
 
-    // We query the relayer API to get the deposit limits for different token and destination combinations.
+    // We query the relayer API to get the deposit limits for different token and origin combinations.
     // The relayer should *not* be filling deposits that the HubPool doesn't have liquidity for otherwise the relayer's
     // refund will be stuck for potentially 7 days. Note: Filter for supported tokens first, since the relayer only
     // queries for limits on supported tokens.
     const { inputAmount } = deposit;
-    if (acrossApiClient.updatedLimits && inputAmount.gt(acrossApiClient.getLimit(l1Token.address))) {
+    const limit = acrossApiClient.getLimit(originChainId, l1Token.address);
+    if (acrossApiClient.updatedLimits && inputAmount.gt(limit)) {
```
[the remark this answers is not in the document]
nicholaspai: hmm that's a good idea, and now the inventory manager is refactored to still ensure its balance allocation would be met

src/relayer/RelayerClientHelper.ts
```diff
@@ -119,18 +119,13 @@ export async function constructRelayerClients(
...
-      ? spokePoolClients
-      : Object.fromEntries(
-          Object.keys(spokePoolClients)
-            .filter((chainId) => config.relayerDestinationChains.includes(Number(chainId)))
-            .map((chainId) => [chainId, spokePoolClients[chainId]])
-        );
-
-  const acrossApiClient = new AcrossApiClient(logger, hubPoolClient, destinationSpokePoolClients, config.relayerTokens);
+  // Determine which origin chains to query limits for.
+  const srcChainIds =
```
nicholaspai, line 123: I think this makes sense but I'll write it to convince myself. We only want to get limits for chains that we have configured origin chain spoke pool clients for. This is OK for the Relayer because the relayer would only see deposits from these origin chains, so there is a 1:1 relationship between chains we query limits for and deposits we fill

is that your logic?

nicholaspai (approved):
I think this looks right, but let me know what you think of my comments

src/relayer/RelayerClientHelper.ts
pxrl: Yep. It might not even be necessary to constrain the set of source chains in this way, since the most important thing is that the destination chain _is_ mainnet. Defining more chain IDs here gives marginally more robustness, because in the event that a single query fails, a different chain ID will be used for a retry lookup (i.e. to protect against RPC outages affecting individual chains).

src/clients/AcrossAPIClient.ts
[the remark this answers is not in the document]
pxrl: Done :+1:

src/relayer/RelayerClientHelper.ts
```diff
@@ -119,18 +119,13 @@ export async function constructRelayerClients(
...
+    config.relayerOriginChains.length > 0
+      ? config.relayerOriginChains
+      : Object.values(spokePoolClients).map(({ chainId }) => chainId);
```
james-a-morris, line 126: Should we add a filter to ensure that no spokes are undefined? We do the check elsewhere

```suggestion
      : Object.values(spokePoolClients).filter(isDefined).map(({ chainId }) => chainId);
```

pxrl: Normally we check this when looping over chain IDs and then dereference the SpokePoolClient from an object. In this case the code iterates directly over the values so I don't think it can blow up in the same way, because we never create an entry for an undefined SpokePoolClient.

PR #1450, opened 2024-04-26
improve(ClientHelper): Cache SpokePool registration blocks

pxrl:
The registration block for each SpokePool deployment is searched on each bot run. These values don't change, so cache them after 1 week.


nicholaspai (approved):

src/common/ClientHelper.ts
```diff
@@ -40,6 +40,47 @@ async function getSpokePoolSigners(
...
+  const blockFinder = undefined;
+  const mainnetActivationBlock = hubPoolClient.getSpokePoolActivationBlock(chainId, spokePoolAddr);
+  const { timestamp } = await hubPoolClient.hubPool.provider.getBlock(mainnetActivationBlock);
+  const hints = { lowBlock: getDeploymentBlockNumber("SpokePool", chainId) };
+  const activationBlock = await getBlockForTimestamp(chainId, timestamp, blockFinder, redis, hints);
+
+  const day = 24 * 3600;
+  const minAge = 7 * day;
+  if (isDefined(redis) && getCurrentTime() - timestamp > minAge) {
+    const ttl = Math.round(30 * day - Math.random() * day); // ttl 29 - 30 days.
```
james-a-morris, line 77: Should we always have this set to 29 or 30 days?

mrice32: Sorry, why specify this TTL at all? Why not just let it be stored indefinitely until/if redis prunes for space (using its LRU algo)?

src/relayer/RelayerClientHelper.ts
```diff
@@ -49,25 +42,14 @@ async function indexedSpokePoolClient(
 
   // Set up Spoke signers and connect them to spoke pool contract objects.
   const signer = baseSigner.connect(await getProvider(chainId));
-
-  const blockFinder = undefined;
```
mrice32, line 53: +1 for factoring this out and simplifying this

mrice32 (commented):
Only question is around the TTL for the cache

src/common/ClientHelper.ts
pxrl: Yeah, can do. I have an instinctive drive to force expiry but there's no real requirement for it here...it's a bad habit that I should shake.

pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1450/commits/32412487495c507c0b992ec65a6f7280cb4ab943

src/common/ClientHelper.ts
```diff
@@ -40,6 +40,46 @@ async function getSpokePoolSigners(
...
+): Promise<number> {
+  const spokePoolAddr = hubPoolClient.getSpokePoolForBlock(chainId, blockNumber);
+  const key = `relayer_${chainId}_spokepool_${spokePoolAddr}_activation_block`;
+
+  const redis = await getRedisCache(hubPoolClient.logger);
+  if (isDefined(redis)) {
+    const activationBlock = await redis.get(key);
+    const numericActivationBlock = Number(activationBlock);
+    if (Number.isInteger(numericActivationBlock) && numericActivationBlock > 0) {
+      return Number(activationBlock);
```
pxrl, line 63:
```suggestion
      return numericActivationBlock;
```

PR #1452, opened 2024-04-27
feat(relayer): Allow unequivalent tokens if outputToken = USDC.e and inputToken = USDC

nicholaspai:
Follow up to #1430 and allows inventory manager to handle deposits from native USDC to USDC.e. This PR doesn't actually change anything because pool rebalance routes all reference Bridged USDC, not USDC, so `hubPoolClient.areTokensEquivalent` for any USDC deposit will always return true (unless the outputToken has a different symbol than USDC),

Once the PoolRebalanceRoutes switch over, `validateOutputToken ()` for deposits where inputToken = _USDC and outputToken = USDC.e will return true as well as deposits where inputToken and outputToken both equal _USDC, while `hubPoolClient.areTokensEquivalent()` will only return true for the former and false for the latter.

Fixes ACX-1967


src/clients/InventoryClient.ts
```diff
@@ -260,6 +261,35 @@ export class InventoryClient {
...
+    if (equivalentTokens) {
+      return true;
+    }
+
+    // Return true if input token is USDC and output token is Bridged USDC.
+    const isInputTokenUSDC = compareAddressesSimple(inputToken, TOKEN_SYMBOLS_MAP["_USDC"].addresses[originChainId]);
+    const isOutputTokenBridgedUSDC = compareAddressesSimple(
+      outputToken,
+      TOKEN_SYMBOLS_MAP["USDC.e"].addresses[destinationChainId]
+    );
```
pxrl, line 289: I think this could blow up if `USDC.e` isn't defined for `destinationChainId`, because `compareAddressesSimple()` doesn't handle any of the inputs being `undefined`.

src/clients/InventoryClient.ts
```diff
@@ -260,6 +261,35 @@ export class InventoryClient {
...
+    if (equivalentTokens) {
+      return true;
+    }
+
+    // Return true if input token is USDC and output token is Bridged USDC.
+    const isInputTokenUSDC = compareAddressesSimple(inputToken, TOKEN_SYMBOLS_MAP["_USDC"].addresses[originChainId]);
+    const isOutputTokenBridgedUSDC = compareAddressesSimple(
+      outputToken,
+      TOKEN_SYMBOLS_MAP["USDC.e"].addresses[destinationChainId]
```
james-a-morris, line 288: We should also account for Base's case listed in `TOKEN_SYMBOLS_MAP.USDbC`.

Source: https://github.com/across-protocol/constants-v3/blob/82d841c2311cb7cc7ac1a3c089a8bef417911e72/src/tokens.ts#L184-L206

src/clients/InventoryClient.ts
nicholaspai: good point. Thoughts on something like this? eae9c9761d9cde72cc337f370096fcf5c5d07c43

src/clients/InventoryClient.ts
nicholaspai: is this too simple or should we instead iterate through all USDbC and USDC.e addresses? 508d350956f1c3fa70f822581838c3bbadcc8523

src/clients/InventoryClient.ts
```diff
@@ -260,6 +261,39 @@ export class InventoryClient {
...
+      destinationChainId
+    );
+    if (equivalentTokens) {
+      return true;
+    }
+
+    // Return true if input token is USDC and output token is Bridged USDC.
+    const isInputTokenUSDC =
+      TOKEN_SYMBOLS_MAP["_USDC"].addresses?.[originChainId] &&
+      compareAddressesSimple(inputToken, TOKEN_SYMBOLS_MAP["_USDC"].addresses[originChainId]);
```
james-a-morris, line 287: We could remove this dual check by allowing `compareAddressesSimple` to take undefined inputs (and return false if any are)

src/clients/InventoryClient.ts
james-a-morris: I think this is a good check. `USDCb` & `USDC.e` are mutually exclusive so as long as we have the destination chain id then we shouldn't need to iterate

src/clients/InventoryClient.ts
nicholaspai: 6abf566fe3f6e784e4284e7677bb9002d7386c8e

nicholaspai:
@james-a-morris I think this comment was on outdated code:

> I think this is a good check. USDCb & USDC.e are mutually exclusive so as long as we have the destination chain id then we shouldn't need to iterate

src/clients/InventoryClient.ts
```diff
@@ -260,6 +261,35 @@ export class InventoryClient {
     return chainIds;
   }
 
+  /**
+   * Returns true if the depositor-specified output token is supported by the this inventory client.
+   * @param deposit V3 Deposit to consider
+   * @returns boolean True if output and input tokens are equivalent or if input token is USDC and output token
```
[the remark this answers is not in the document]
nicholaspai: so my thinking is that these inputs with USDC.e would be marked invalid by the relayer because the input token does not map to a supported L1 token. I think its worth adding a comment that input of USDC.e is not supported in this client, wdyt?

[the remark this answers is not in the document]
mrice32: Yep, I think that makes sense! So the inventory client's input should _always_ be USDC, but output can be either. ty for the explanation.

mrice32 (approved):

PR #1454, opened 2024-04-28
WIP: feat: Count ETH with WETH balances in InventoryClient

nicholaspai:
Addresses duplicate rebalancing issue where rebalancer doesn't see ETH being wrapped into WETH in time before sending redundant rebalance of ETH to L2

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -75,7 +75,8 @@ export class InventoryClient {
       .reduce((acc, curr) => acc.add(curr), bnZero);
   }
 
-  // Get the balance of a given l1 token on a target chain, considering any outstanding cross chain transfers as a virtual balance on that chain.
+  // Get the balance of a given l1 token on a target chain, considering any outstanding cross chain transfers as a
+  // virtual balance on that chain, and counting ETH and WETH balances together.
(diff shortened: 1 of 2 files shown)


nicholaspai:
Note to self: can use this token client getNativeBalance method wherever we call GetBalance in the repo

nicholaspai:
Make sure that counting eth only factors in when computing allocation percentages to inform rebalances. We don't want it to be counted in the available weth to be rebalanced though as we can't send eth the same way as we can weth

pxrl:
This one could potentially be tricky, since we don't want to rebalance if ETH + WETH > threshold, but in case of a small WETH shortfall causing a non-fill, we might actually want to ignore the ETH :\ I haven't figured out a clean solution for that.

nicholaspai:
Only count native balance on chains that use eth unlike polygon

nicholaspai:
> This one could potentially be tricky, since we don't want to rebalance if ETH + WETH > threshold, but in case of a small WETH shortfall causing a non-fill, we might actually want to ignore the ETH :\ I haven't figured out a clean solution for that.

One idea would be to change the cross chain outstanding transfers logic to only mark cross chain transfers as finalized (for chains that bridge eth not weth) after the eth gets wrapped. But I'm not sure there's a good event to track

PR #1457, opened 2024-04-29
fix(InventoryClient): Map outstanding crosschain transfers to correct relayer address

nicholaspai:
## Summary

The rebalancer is designed to [wrap any excess ETH into WETH](https://github.com/across-protocol/relayer-v3/blob/dd5e20ec4538c8c3e7bfb923187f521f92d85882/src/relayer/RelayerClientHelper.ts#L216) before [updating the `TokenClient`](https://github.com/across-protocol/relayer-v3/blob/dd5e20ec4538c8c3e7bfb923187f521f92d85882/src/relayer/RelayerClientHelper.ts#L223) and eventually proceeding to the [rebalancing inventory step](https://github.com/across-protocol/relayer-v3/blob/dd5e20ec4538c8c3e7bfb923187f521f92d85882/src/relayer/index.ts#L118). However, in practice we've seen that the rebalancer is not able to detect these wrapped ETH events in time to prevent a duplicate rebalance.

This PR adds safety to the OP-stack and ZkSync adapters where we bridge ETH via the AtomicDepositor and expect to receive ETH on the L2 side and subsequently expect to wrap it into WETH. We now only mark a deposit as finalized in the inventory client when the L2 deposit event is followed by a wrap event.

This algorithm can clearly fall prey to false positives where an unrelated ETH unwrap leads to the rebalancer falsely detecting that an L1 to L2 WETH transfer has finalized. However, for the most part, WETH wraps should only be triggered by the rebalancer, and even if wrapping ETH was unrelated to the specific L1 to L2 transfer we care about, it should still convert the ETH into WETH balance which is what we care about and would prevent WETH duplicate cross chain transfers.

~UPDATE: This solution didn't work since we also use the adapters in the Monitor to track stuck rebalances from Hub to Spoke. These WETH transfers would originate from the Hub and be wrapped by the Spoke so tracking an unwrapping from the Spoke's address wouldn't work if the transfer as sent from the Hub. We'd need a larger change to how we choose which `monitoredAddresses` to track.~

We use this approach for EOA xchain transfers only.

This PR can be viewed as an assorted set of inventory manager adapter-related bug fixes.

In addition, I've  moved the `tokenClient.update()` as late as possible so that the rebalancer has a higher chance of capturing unwrapped WETH into its balance before rebalancing duplicate txns.

## Fixes

- This PR fixes a bug in the ZkSyncAdapter where the `initiatedQueryResult` for WETH were not being properly loaded into the cross chain outstanding transfers object. This caused the inventory manager to NEVER identify outstanding zksync deposits, leading to possible duplicate rebalances until the cross chain transferred ETH was wrapped into WETH on the L2 side.
- This PR fixes a bug in the Optimism Adapter where outstanding ETH cross chain transfer amounts are being credited to the atomic depositor address and therefore don't factor into the optimism/base relayer address. This similarly causes issues where the relayer's balance will look off by the amount of ETH transferred until it gets wrapped into WETH on the L2 side and would get picked up by the TokenClient at that point
~- This PR short circuits the `monitoredAddress` loop in each of the chain-specific getOutstandingCrossChainTransfer functions because its never necessary to query events for both the `atomicWethDepositor` and the `relayer` address.~
- This PR removes `atomicWethDepositor` as an injected `monitoredAddress` in the `RelayerClientHelper` by changing how the `adapters/op-stack/WethBridge` handles tracking WETH transfers
- Fix Arbitrum and Polygon adapter bugs where loop/math bugs were causing outstanding transfers to be improperly associated with the wrong address. This is only an issue in the Monitor when there are multiple `monitoredAddresses` added before the ones we really care about, the Hub and Spoke pool addresses for use in the "stuck rebalances" monitor


src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -86,16 +90,14 @@ export class OpStackAdapter extends BaseAdapter {
...
-            assign(
-              this.l2DepositFinalizedEvents,
-              [monitoredAddress, l1Token],
-              depositFinalizedResults.map(processEvent)
-            );
+            // If l1Token is WETH then always map the transfer information to the relayer/signer address, not the
+            // atomic weth depositor contract address which is the `monitoredAddress` used to catch the
+            // transfer events. The following event filters are designed only to catch transfers initiated by an EOA on
+            // L1 sending WETH via the AtomicWethDepositor and receiving ETH on the L2 side at their EOA.
+            const relayerAddress =
```
nicholaspai, line 97: This is the only chain adapter where we only care about the atomic weth depositor and not the relayer address. This is because the L1 event `ETHDepositInitiated` and the L2 event  `DepositFinalized` both index on the `fromAddress` which is the AtomicWethDepositor... Perhaps there's a better way to refactor this code besides passing in both the atomicWethDepositor address and relayer address into each chain adapter and then within the adapter deciding which addresses to choose.

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -107,7 +127,8 @@ export class ZKSyncAdapter extends BaseAdapter {
         assign(
           this.l1DepositInitiatedEvents,
           [address, l1TokenAddress],
-          initiatedQueryResult.map(processEvent).filter((e) => e?.l1Token && e.l1Token === l1TokenAddress)
+          // If token is WETH, the event doesn't have a token or l1Token param.
+          initiatedQueryResult.map(processEvent).filter((e) => isWeth || e.l1Token === l1TokenAddress)
```
nicholaspai, line 131: This was a bug described in the PR description. WETH initiatedQueryResult's never filtered here successfully because `l1Token` is never defined for the WETH event. Therefore we never saw outstanding transfers for WETH

src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -86,16 +94,14 @@ export class OpStackAdapter extends BaseAdapter {
               bridge.queryL2BridgeFinalizationEvents(l1Token, monitoredAddress, l2SearchConfig),
             ]);
 
-            assign(
-              this.l1DepositInitiatedEvents,
-              [monitoredAddress, l1Token],
```
nicholaspai, line 91: Currently we are associating outstanding transfers with the `atomicDepositorAddress` causing similar outcomes to the bug in https://github.com/across-protocol/relayer-v3/pull/1457/files#r1582509701

src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -60,6 +64,10 @@ export class OpStackAdapter extends BaseAdapter {
     );
   }
 
+  getL1Weth(): string {
+    return TOKEN_SYMBOLS_MAP.WETH.addresses[this.hubChainId];
+  }
+
```
pxrl, line 70: Worth sticking this in `BaseAdapter` ? Seems like it could be useful more broadly than OP.

nicholaspai: I just realized its already there as `this.wethAddress` so i'll remove this func

src/clients/bridges/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * l2EthDepositEvents will match with the same l2WrapEvent.
+ */
+export function matchL2EthDepositAndWrapEvents(l2EthDepositEvents: Event[], _l2WrapEvents: Event[]): Event[] {
+  let l2WrapEvents = [..._l2WrapEvents]; // deep-copy because we're going to modify this in-place.
+  return l2EthDepositEvents.filter((l2EthDepositEvent: Event) => {
+    // Search from left to right to find the first following wrap event.
+    const followingWrapEvent = l2WrapEvents.find((wrapEvent) => wrapEvent.blockNumber > l2EthDepositEvent.blockNumber);
+    // Delete the wrap event from the l2 wrap events array to avoid duplicate processing.
+    if (followingWrapEvent) {
+      l2WrapEvents = l2WrapEvents.splice(l2WrapEvents.indexOf(followingWrapEvent));
```
pxrl, line 28: `Array.splice()` modifies the array in place and returns the element that was spliced out - is that the intention here? The comment at line 26 gives the opposite impression - that an event should be dropped from the array. For that I think this should work:

```suggestion
      l2WrapEvents.splice(l2WrapEvents.indexOf(followingWrapEvent), 1);
```

pxrl: I'm also not sure how well `Array.indexOf()` works on complex types - it might be safer to use `Array.findIndex()` to get the index, and then supply that to `Array.splice()`.

nicholaspai: hmm let me try

nicholaspai: fa050e803225010338fde009c8ecbe4547ec36c5

src/clients/bridges/AdapterManager.ts
```diff
@@ -25,23 +25,31 @@ export class AdapterManager {
     if (!spokePoolClients) {
       return;
     }
+    // @dev We only need to monitor the atomic weth depositor contract address when tracking outstanding crosschain
+    // transfers for certain chains. Most chains we can simply monitor the relayer's EOA address. One notable
+    // exception is OP stack chains, where we need to monitor the atomic depositor contract address because the
+    // crosschain transfer events do not allow us to filter on the "recipient" address, which would be the L2
+    // relayer EOA.
+    const monitoredAddressesExclusingAtomicDepositor = this.monitoredAddresses.filter(
```
mrice32, line 33:
```suggestion
    const monitoredAddressesExcludingAtomicDepositor = this.monitoredAddresses.filter(
```

src/clients/bridges/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+ * wrapping ETH is only expected to be done by this relayer, then this is a very accurate proxy for deciding
+ * when WETH cross chain transfers have finalized into the relayer's L2 WETH inventory.
+ * @dev This function is used in the WethBridge class in the OP stack and the ZkSyncAdapter.
+ * @param l2EthDepositEvents List of L2 DepositFinalized events emitted when the EOA receives ETH on L2.
+ * @param _l2WrapEvents List of L2 Wrap events emitted when the EOA wraps ETH into WETH on L2.
+ * @returns List of l2EthDepositEvents followed by a l2WrapEvent. None of the
+ * l2EthDepositEvents will match with the same l2WrapEvent.
+ */
+export function matchL2EthDepositAndWrapEvents(l2EthDepositEvents: Event[], _l2WrapEvents: Event[]): Event[] {
+  const l2WrapEvents = [..._l2WrapEvents]; // deep-copy because we're going to modify this in-place.
```
mrice32, line 22: nit: deep copy would be performing a deep copy of each event object as well. I think a better description of this operation is a clone of the array.
```suggestion
  const l2WrapEvents = [..._l2WrapEvents]; // clone array because we're going to modify this in-place.
```

src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -79,23 +82,29 @@ export class OpStackAdapter extends BaseAdapter {
...
-            assign(
-              this.l2DepositFinalizedEvents,
-              [monitoredAddress, l1Token],
-              depositFinalizedResults.map(processEvent)
-            );
+            // If l1Token is WETH then always map the transfer information to the relayer/signer address, not the
+            // atomic weth depositor contract address which is the `monitoredAddress` used to catch the
+            // transfer events. The following event filters are designed only to catch transfers initiated by an EOA on
+            // L1 sending WETH via the AtomicWethDepositor and receiving ETH on the L2 side at their EOA.
+            const relayerAddress = isWeth ? await this.getSigner(this.chainId).getAddress() : monitoredAddress;
```
mrice32, line 105: Will this assumption cause issues in the reporter or monitor, where the signer address is irrelevant?

mrice32 (commented):
Looks good! A few minor questions/comments

mrice32 (commented):
Also this is a great bug fix! Previously, were we just getting away with this because if you don't rebalance often, the wrap happens before the next run?

src/clients/bridges/op-stack/OpStackAdapter.ts
nicholaspai: good point, i'm going to refactor this a bit so we don't need to monitor the atomicDepositorAddress anymore

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -107,20 +106,19 @@ export class ArbitrumAdapter extends BaseAdapter {
...
         );
-        validTokens.push(l1Token);
       }
     }
 
     const results = await Promise.all(promises);
 
     // 2 events per token.
-    const numEventsPerMonitoredAddress = 2 * validTokens.length;
+    const numEventsPerMonitoredAddress = 2 * availableL1Tokens.length;
```
nicholaspai, line 115: This is a bug fix. `validTokens` is pushed once per l1 token per monitored address, so setting this equal to `validTokens.length` means returning the entire `results` array. This code hasn't been touched in 12+ months so I'm not sure it ever worked...

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -107,20 +106,19 @@ export class ArbitrumAdapter extends BaseAdapter {
...
 
     // Segregate the events list by monitored address.
     const resultsByMonitoredAddress = Object.fromEntries(
       this.monitoredAddresses.map((monitoredAddress, index) => {
         const start = index * numEventsPerMonitoredAddress;
-        return [monitoredAddress, results.slice(start, start + numEventsPerMonitoredAddress + 1)];
```
nicholaspai, line 123: Bug fix, this adds 1 more element than we want

src/clients/bridges/LineaAdapter.ts
```diff
@@ -157,19 +157,18 @@ export class LineaAdapter extends BaseAdapter {
...
-          // 1. Get all initiated MessageSent events from the L1MessageService where the 'from' address is
-          //    the AtomicDepositor and the 'to' address is the user's address.
+          // 1. Get all initiated MessageSent events from the L1MessageService where the 'to' address is the 
+          //    user's address.
           // 2. Pipe the resulting _messageHash argument from step 1 into the MessageClaimed event filter
           // 3. For each MessageSent, match the _messageHash to the _messageHash in the MessageClaimed event
           //    any unmatched MessageSent events are considered outstanding transfers.
           const initiatedQueryResult = await paginatedEventQuery(
             l1MessageService,
-            l1MessageService.filters.MessageSent(atomicDepositor.address, address),
```
nicholaspai, line 172: Filtering on `atomicDepositor.address` means this adapter cannot be used by the Monitor to track any rebalances from Hub to Spoke, so until now we haven't been watching Linea rebalances.

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -136,7 +137,6 @@ export class PolygonAdapter extends BaseAdapter {
     const availableTokens = l1Tokens.filter(this.isSupportedToken.bind(this));
 
     const promises: Promise<Event[]>[] = [];
-    const validTokens: SupportedL1Token[] = [];
```
nicholaspai, line 139: Same bugs in this file as in `Arbitrum_Adapter`

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -78,10 +76,10 @@ export class ZKSyncAdapter extends BaseAdapter {
               l1SearchConfig
             ),
 
-            // Filter on transfers between aliased AtomicDepositor address and l2Receiver
+            // Filter on transfers to l2Receiver
             paginatedEventQuery(
               l2EthContract,
-              l2EthContract.filters.Transfer(aliasedAtomicWethDepositor, address),
```
nicholaspai, line 84: Same issue as in `LineaAdapter` where filtering on the atomic weth depositor address means we can't track HubPool rebalances.

nicholaspai: Removing this filtered param means we're going to see A LOT of transfers into the relayer's EOA but the `computeOutstandingTransfers` call at the end of this function will only keep those events that match with an initiatedQueryResult

src/monitor/Monitor.ts
```diff
@@ -525,19 +525,33 @@ export class Monitor {
...
       for (const l1Token of allL1Tokens) {
+        // Outstanding transfers are mapped to either the spoke pool or the hub pool, depending on which
+        // chain events are queried. Some only allow us to index on the fromAddress, the L1 originator or the 
+        // HubPool, while others only allow us to index on the toAddress, the L2 recipient or the SpokePool.
         const transferBalance = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
           spokePoolAddress,
           chainId,
           l1Token.address
-        );
+        ).add(this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
```
nicholaspai: This was a bug before because outstanding transfers for certain chains where only the `fromAddress` can be tracked, like for Optimism transfers involving WETH, will be associated with the HubPool's address, not the spoke pool's

src/relayer/RelayerClientHelper.ts
```diff
@@ -215,13 +211,6 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
     clients.inventoryClient.update(),
     clients.inventoryClient.wrapL2EthIfAboveThreshold(),
     clients.inventoryClient.setL1TokenApprovals(),
+    config.sendingRelaysEnabled ? clients.tokenClient.setOriginTokenApprovals() : Promise.resolve(),
   ]);
-
-  // Update the token client after the inventory client has done its wrapping of L2 ETH to ensure latest WETH ballance.
-  // The token client needs route data, so wait for update before checking approvals.
-  clients.tokenClient.clearTokenData();
```
nicholaspai, line 222: I moved the token client update to later: right before the relayer and right before the rebalancer runs. This reduces chance we miss any events

src/clients/bridges/op-stack/WethBridge.ts
```diff
@@ -45,23 +45,50 @@ export class WethBridge implements OpStackBridge {
...
     l1Token: string,
     fromAddress: string,
     eventConfig: EventSearchConfig
   ): Promise<Event[]> {
-    return paginatedEventQuery(this.l1Bridge, this.l1Bridge.filters.ETHDepositInitiated(fromAddress), eventConfig);
+    // We need to be smart about the filtering here because the ETHDepositInitiated event does not
+    // index on the `toAddress` which is the `fromAddress` that we pass in here and the address we want
+    // to actually filter on. So we make some simplifying assumptions:
+    // - For our tracking purposes, the ETHDepositInitiated `fromAddress` will be the
+    //   AtomicDepositor if the fromAddress is an EOA.
```
nicholaspai, line 57: The reasoning behind supporting this complexity is that we use this adapter to check for relayer rebalances in the `rebalancer` and for stuck hub pool pool rebalances in the `monitor`

src/relayer/index.ts
```diff
@@ -106,6 +106,9 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       await updateRelayerClients(relayerClients, config);
 
       if (!config.skipRelays) {
+        // Since the above spoke pool updates are slow, refresh token client before sending rebalances now:
+        relayerClients.tokenClient.clearTokenData();
```
nicholaspai, line 110: I think this work will build nicely off of https://github.com/across-protocol/relayer-v3/pull/1444

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -130,7 +128,11 @@ export class ArbitrumAdapter extends BaseAdapter {
       // The logic below takes the results from the promises and spreads them into the l1DepositInitiatedEvents and
       // l2DepositFinalizedEvents state from the BaseAdapter.
       eventsToProcess.forEach((result, index) => {
-        const l1Token = validTokens[Math.floor(index / 2)];
+        if (eventsToProcess.length === 0) {
+          return;
+        }
+        assert(eventsToProcess.length % 2 === 0, "Events list length should be even");
+        const l1Token = availableL1Tokens[Math.floor(index / 2)];
```
james-a-morris, line 135: We can remove the Floor check because our length is even
```suggestion
        const l1Token = availableL1Tokens[index / 2];
```

src/clients/bridges/PolygonAdapter.ts
```diff
@@ -169,28 +169,31 @@ export class PolygonAdapter extends BaseAdapter {
...
     // Process events for each monitored address.
     for (const monitoredAddress of this.monitoredAddresses) {
       const eventsToProcess = resultsByMonitoredAddress[monitoredAddress];
       eventsToProcess.forEach((result, index) => {
-        const l1Token = validTokens[Math.floor(index / 2)];
+        if (eventsToProcess.length === 0) {
+          return;
+        }
+        assert(eventsToProcess.length % 2 === 0, "Events list length should be even");
+        const l1Token = availableTokens[Math.floor(index / 2)];
```
james-a-morris, line 196: Similar to above

src/clients/bridges/LineaAdapter.ts
james-a-morris: I agree with this, but the `MessageSent` event will be emitted on the ERC20 bridges as well so this may lead to double counting.

nicholaspai: hm but we match off the events using the `amount` prop so would we still double count?

nicholaspai: as an example, here's a ERC20 bridge: https://etherscan.io/tx/0x99609b08ae575644d780c109ecc9f889003610736040bc43a4ff77e428712258#eventlog

and here's a WETH bridge: https://etherscan.io/tx/0xc0015d1a25da5a15125525217cdf7b8d1bb68916d54d1794761bfbb305bde140#eventlog

Seems like the ERC20 bridge sets the `_to` address equal to the L2 token bridge, i guess because to send ERC20's, the token bridges are sending messages between each other, and the L2 token bridge releases the tokens. While with ETH, the message goes directly to the recipient EOA with the passed in `msg.value`

james-a-morris: LGTM as long as this passes integration tests - but we do emit MessageSent events whenever we interact with the Linea message bridge.

nicholaspai: I think the _toAddress will be different for ERC20 transfers. It will be the L2 token bridge not the EOA recipient

src/clients/bridges/op-stack/WethBridge.ts
```diff
@@ -45,23 +58,86 @@ export class WethBridge implements OpStackBridge {
...
     eventConfig: EventSearchConfig
   ): Promise<Event[]> {
-    return paginatedEventQuery(
-      this.l2Bridge,
-      this.l2Bridge.filters.DepositFinalized(ZERO_ADDRESS, undefined, fromAddress),
-      eventConfig
-    );
+    // Check if the sender is a contract on the L1 network.
+    const isContract = (await this.l1Bridge.provider.getCode(fromAddress)) !== "0x";
+    if (!isContract) {
```
pxrl, line 99: This check has been repeated a few times - worth centralising the logic? We do fwiw have `isContractDeployedToAddress()` in the sdk/utils.

src/relayer/RelayerClientHelper.ts
pxrl: I've also touched this code in another PR: https://github.com/across-protocol/relayer-v3/pull/1444

I have fwiw seen that TokenClient update can hang for a while, and that really hurts relayer performance. I'd hoped we could kick it off at the beginning and await it in the end, since it's not actually dependent on SpokePoolClient update anymore.

src/relayer/index.ts
pxrl: Nice!

pxrl (approved):
There's an open action from Matt but OK from me on this once that's resolved.

src/clients/bridges/op-stack/OpStackAdapter.ts
nicholaspai: this was refactored and we don't read the signer anymore, only the provider in various places which should always be available

src/clients/bridges/LineaAdapter.ts
```diff
@@ -201,15 +194,34 @@ export class LineaAdapter extends BaseAdapter {
...
 
-          // Define the initialized and finalized event filters for the L1 and L2 bridges
+          // Define the initialized and finalized event filters for the L1 and L2 bridges. We only filter
+          // on the recipient so that the filters work both to track Hub-->Spoke transfers and EOA transfers, and
+          // because some filters like ReceivedFromOtherLayer only index the recipient.
           const [filterL1, filterL2] = isUsdc
-            ? [l1Bridge.filters.Deposited(address, null, address), l2Bridge.filters.ReceivedFromOtherLayer(address)]
-            : [l1Bridge.filters.BridgingInitiated(address, null, l1Token), l2Bridge.filters.BridgingFinalized(l1Token)];
+            ? [
+                l1Bridge.filters.Deposited(null /* depositor */, null, address /* to */),
```
nicholaspai, line 202: The filters for `Deposited` and `BridgingInitiated` have to be changed.
- `Deposited` can't use the same `address` for sender and receiver in the case of tracking the HubPool or SpokePool address
- `recipient` isn't filtered in `BridgingInitiated` and we should not filter on `sender`, otherwise we won't be able track the SpokePool which we need to do to get the `ReceivedFromOtherLayer` filter working. The only way to filter on sender in `BridgingInitiated` is to add a switch: if `isUsdc`, then track SpokePool only since we can only filter on `recipient` in `ReceivedFromOtherLayer`, otherwise track HubPool only.

PR #1460, opened 2024-04-30
improve(RelayerConfig): Permit symbol-based token config

pxrl:
This change permits the InventoryConfig to be specified in either HubPool token address or token symbol. Token symbol is preferred because it's simpler to configure and reduces the chance of mistakes.

During config, any symbols are resolved to their underlying token addresses on the HubPool chain, so no subsequent changes are required for any accesses into the post-processed inventory config.


src/relayer/RelayerConfig.ts
```diff
@@ -145,11 +146,22 @@ export class RelayerConfig extends CommonConfig {
...
-      Object.keys(inventoryConfig?.tokenConfig ?? {}).forEach((l1Token) => {
-        Object.keys(inventoryConfig.tokenConfig[l1Token]).forEach((chainId) => {
-          const tokenConfig = inventoryConfig.tokenConfig[l1Token][chainId];
+      const rawTokenConfigs = inventoryConfig?.tokenConfig ?? {};
+      const tokenConfigs = (inventoryConfig.tokenConfig = {});
+      Object.keys(rawTokenConfigs).forEach((l1Token) => {
+        // If the l1Token is a symbol, resolve the correct address.
+        const effectiveL1Token = ethersUtils.isAddress(l1Token)
+          ? l1Token
+          : TOKEN_SYMBOLS_MAP[l1Token].addresses[this.hubPoolChainId];
```
nicholaspai, line 155: Should we try to throw an early error if l1Token is not a valid symbol?

james-a-morris: Agreed with this. Would it be possible to assert that `ethersUtils.isAddress(l1Token) || isDefined(TOKEN_SYMBOLS_MAP[l1Token])`

pxrl: Implemented: https://github.com/across-protocol/relayer-v3/pull/1460/commits/69252073672ec712e8304c3113580bdbac3113b7

PR #1461, opened 2024-04-30
feat: Lookup outputToken symbol using TOKEN_SYMBOLS_MAP

nicholaspai:
Currently we map a deposit's outputToken to an L1TokenInfo object via the HubPoolClient. This will no longer work once outputTokens are supported that are not mapped to the HubPool via PoolRebalanceRoutes.

This PR replaces the hubPoolClient. getL1TokenInfoForL2Token call wherever the outputToken is supplied with a `getL1TokenInfo` call that looks into TOKEN_SYMBOLS_MAP

We replace this in a few spots:
- Relayer.handleUnprofitableFill which logs about unprofitable deposits.
- Relayer.handleTokenShortfall which maps shortfalls with outputTokens
- ProfitClient which looks up token prices of output tokens
- Monitor which logs about unknown transfers and unfilled deposits using the transferred token and output token respectively, bth of which can't be assumed to be in TOKEN_SYMBOLS_MAP

In addition, this PR adds a hard-coded symbol remapping in the ProfitClient that forces price equivalency between USDC.e, USDbC and USDC symbols, which will be required to support USDC -- USDC.e/USDbC deposits post-CCTP migration.

Fixes ACX-2103


src/clients/ProfitClient.ts
```diff
@@ -112,6 +113,13 @@ const QUERY_HANDLERS: {
   11155420: relayFeeCalculator.OptimismSepoliaQueries,
 };
 
+// Hard-coded mapping of token symbols that should be treated as having equivalent
+// prices. The right-hand side should map to a token symbol in TOKEN_SYMBOLS_MAP.
+const TOKEN_EQUIVALENCE_REMAPPING: { [symbol: string]: string } = {
+  ["USDC.e"]: "USDC",
+  ["USDbC"]: "USDC",
```
pxrl, line 120: Do you think this could work as a mapping of HubPoolToken -> Array? i.e.

`"USDC": ["USDC.e", "USDbC"]`

src/clients/ProfitClient.ts
```diff
@@ -338,7 +361,20 @@ export class ProfitClient {
...
+    let outputTokenInfo: L1Token;
+    // If the output token and the input token are equivalent, then we can look up the token info
+    // via the HubPoolClient since the output token is mapped via PoolRebalanceRoute to the HubPool.
+    // If not, then we should look up outputToken in the TOKEN_SYMBOLS_MAP for the destination chain.
+    const matchingTokens =
+      TOKEN_SYMBOLS_MAP[inputTokenInfo.symbol]?.addresses[deposit.destinationChainId] === deposit.outputToken;
+    if (matchingTokens) {
+      outputTokenInfo = hubPoolClient.getL1TokenInfoForL2Token(deposit.outputToken, deposit.destinationChainId);
+    } else {
+      // This function will throw if the token is not found in the TOKEN_SYMBOLS_MAP for the destination chain.
```
pxrl, line 375: fwiw, if profitability computation throws then we just assume the fill is unprofitable, so it's not necessarily a problem to throw here :+1:

src/clients/HubPoolClient.ts
```diff
@@ -33,6 +34,10 @@ export class HubPoolClient extends clients.HubPoolClient {
     );
   }
 
+  getL1TokenInfoForAddress(tokenAddress: string, chain: number): L1Token {
```
nicholaspai, line 37: I moved the util function into this class to make it easy to mock for tests. Am open to opinions but calling `getL1TokenInfo` directly led to lots of tests failing and its hard to mock an isolated util function

src/clients/ProfitClient.ts
nicholaspai: I don't think that's faster though since we're usually going to be looking up into this mapping with the `outputToken` and trying to find out what it should map to. Speed obviously doesn't matter but I think this structure gives us the most scalability

src/clients/ProfitClient.ts
nicholaspai: I do worry about silent errors though and we'll end up just seeing lots of "unprofitable deposits"

src/monitor/Monitor.ts
```diff
@@ -201,7 +201,7 @@ export class Monitor {
       const chainId = parseInt(chainIdStr);
       mrkdwn += `*Destination: ${getNetworkName(chainId)}*\n`;
       for (const tokenAddress of Object.keys(amountByToken)) {
-        const tokenInfo = this.clients.hubPoolClient.getL1TokenInfoForL2Token(tokenAddress, chainId);
+        const tokenInfo = this.clients.hubPoolClient.getL1TokenInfoForAddress(tokenAddress, chainId);
```
nicholaspai, line 204: Example of a `getL1TokenInfoForL2Token` that we can't rely on anymore because the `tokenAddress` is based on `outputToken`

src/clients/ProfitClient.ts
pxrl: The ProfitClient propagates the underlying error through to the "fill is unprofitable" message, so it is possible to see it in even the slack logs.

I don't really see how we can handle it otherwise, since if it threw "hard" then any depositor could take the relayer down by making a deposit with some random dog coin as the outputToken.

src/clients/InventoryClient.ts
```diff
@@ -215,7 +215,7 @@ export class InventoryClient {
         if (!this.hubPoolClient.l2TokenEnabledForL1Token(l1Token, chainId)) {
           refunds[chainId] = toBN(0);
         } else {
-          const destinationToken = this.getDestinationTokenForL1Token(l1Token, chainId);
+          const destinationToken = this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, Number(chainId));
```
nicholaspai, line 218: As prep for upcoming inventory client changes in #1465 we should replace any calls with an internal `this.getDestinationTokenForL1Token` call with a call to the HubPoolClient to resolve an L2 token using pool rebalance routes. This should be replaced only when computing refunds (which always use pool rebalance route supported tokens) and running balances.

src/clients/ProfitClient.ts
```diff
@@ -367,14 +402,9 @@ export class ProfitClient {
       ? netRelayerFeeUsd.mul(fixedPoint).div(outputAmountUsd)
       : bnZero;
 
-    // If either token prices are unknown, assume the relay is unprofitable. Force non-equivalent tokens
-    // to be unprofitable for now. The relayer may be updated in future to support in-protocol swaps.
-    const equivalentTokens = outputTokenInfo.address === inputTokenInfo.address;
+    // If either token prices are unknown, assume the relay is unprofitable.
     const profitable =
-      equivalentTokens &&
```
nicholaspai, line 374: Here we remove the `equivalentToken` check since we now can handle unequivalent tokens if the output token is USDC.e/USDbC while inputtoken is USDC.

If output token is not in TOKEN_SYMBOLS_MAP then the price will be 0 so this `profitable` will be false

src/clients/ProfitClient.ts
```diff
@@ -112,6 +112,13 @@ const QUERY_HANDLERS: {
   11155420: relayFeeCalculator.OptimismSepoliaQueries,
 };
 
+// Hard-coded mapping of token symbols that should be treated as having equivalent
+// prices. The right-hand side should map to a token symbol in TOKEN_SYMBOLS_MAP.
+const TOKEN_EQUIVALENCE_REMAPPING: { [symbol: string]: string } = {
```
james-a-morris, line 117: Would this make sense to go into `constants-v2` since we can use this elsewhere?

src/clients/HubPoolClient.ts
```diff
@@ -33,6 +34,10 @@ export class HubPoolClient extends clients.HubPoolClient {
     );
   }
 
+  getTokenInfoForAddress(tokenAddress: string, chain: number): L1Token {
+    return getTokenInfo(tokenAddress, chain);
+  }
```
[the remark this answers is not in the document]
james-a-morris: We can add an optional mapping parameter but this may be worth keeping if it's intentional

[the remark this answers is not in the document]
nicholaspai: yeah what do you think is easier to. maintain going forward? i kind of dislike optional params.. but its subjective

src/clients/ProfitClient.ts
nicholaspai: https://github.com/across-protocol/constants-v3/pull/33

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
pxrl: There's also a clash here with the upstream HubPoolClient `getXforYAtBlock()` methods - the distinction of when to use one or the other is subtle, so I think it'd be easy to use the wrong function in some contexts.

pxrl (approved):

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
james-a-morris: We can keep it here - I think that's fine

PR #1463, opened 2024-04-30
fix(test): Prime multicall in Dataworker test

pxrl:
The use of smock in the Dataworker executePoolRebalances test means that multicall output needs to be predefined. This is needed in advance of an sdk-v2 update, which uses multicall() in HubPoolClient.update().


nicholaspai (approved):
How has this been working until now?

pxrl:
> How has this been working until now?

We haven't bumped SDK yet, so it hasn't (and isn't yet) required. It's a harmless change to introduce now though, since the test always has to specify the expected multicall output prior to calling it, so if it's not read then it'll always be overwritten anyway.

PR #1464, opened 2024-04-30
feat: add run-id compatible logger and replace FTL with logger package

mrice32:
This PR does two things:

1. It pulls in the new logger, which should allow all Across bots to use the new `run-identifier` [feature](https://github.com/UMAprotocol/protocol/pull/4745) added in the upstream `@uma/logger` package.
2. It replaces the `@uma/financial-templates-lib` package with the `@uma/logger` package, since we only use the re-exported logger portion of the FTL package.


nicholaspai (approved):
Let's try it !

bmzig (approved):
Sounds good to me.

PR #1465, opened 2024-04-30
feat(InventoryClient): Support 1:many HubPool mappings

pxrl:
This update to the InventoryClient permits multiple tokens to map to a single HubPool token. This will be used with USDC/CCTP, where the existing USDC deployment will map to both native USDC and bridged USDC on the various non-mainnet chains supported by Across. See below for an example of how this is configured. Note that the symbol _USDC is the temporary label for native USDC.

```
  "USDC": {
    "USDC.e": {
      "10": { "targetPct": 8, "thresholdPct": 4 },
      "137": { "targetPct": 4, "thresholdPct": 2 },
      "324": { "targetPct": 8, "thresholdPct": 4 },
      "42161": { "targetPct": 8, "thresholdPct": 4 },
      "59144": { "targetPct": 5, "thresholdPct": 2 }
    },
    "USDbC": {
      "8453": { "targetPct": 5, "thresholdPct": 2 }
    },
    "_USDC": {
      "137": { "targetPct": 4, "thresholdPct": 2 }
    }
  },
```


src/clients/InventoryClient.ts
```diff
@@ -134,13 +153,25 @@ export class InventoryClient {
 
   // Find how short a given chain is for a desired L1Token.
   getTokenShortFall(l1Token: string, chainId: number): BigNumber {
-    return this.tokenClient.getShortfallTotalRequirement(chainId, this.getDestinationTokenForL1Token(l1Token, chainId));
+    return this.getDestinationTokensForL1Token(l1Token, chainId)
+      .map((token) => this.tokenClient.getShortfallTotalRequirement(chainId, token))
+      .reduce((acc, curr) => acc.add(curr), bnZero);
   }
```
pxrl, line 159: Thinking about this one a little bit, I think the InventoryClient should mostly be ignorant that two "equivalent" tokens might share the same chain id, and it should probably just view all chain/token combinations as a flat distribution of balances. That would mean that this specific update should be undone, but it also means that it needs additional information to distinguish in the case that multiple tokens on chainId map onto l1Token.

src/clients/InventoryClient.ts
```diff
@@ -83,9 +99,12 @@ export class InventoryClient {
       return bnZero;
     }
 
+    const balances = this.getDestinationTokensForL1Token(l1Token, chainId).map(
+      (token) => this.tokenClient.getBalance(Number(chainId), token) || bnZero
```
nicholaspai, line 103: Note: we're also going to have to update the tokenClient to look up this same set of tokens. Currently it looks up all l2 tokens associated with l1 tokens the hub pool client knows about, so it wouldn't know about these "extra" tokens. Any ideas for how to implement that?

pxrl: The TokenClient knows about these now, per this change: https://github.com/across-protocol/relayer-v3/commit/e3e6809140554375d008745dd98afbc8717b3026 It's a USDC-specific implementation, but that's all we need for now. Once the pieces fall into place for CCTP I think we'll be able to spot some opportunities refactoring to make it more cohesive.

src/clients/InventoryClient.ts
```diff
@@ -105,13 +135,18 @@ export class InventoryClient {
...
-            .div(cumulativeBalance);
+        distribution[chainId] ??= {};
+
+        if (cumulativeBalance.eq(bnZero)) {
+          return;
         }
+
+        const l2Tokens = this.getDestinationTokensForL1Token(l1Token, chainId);
+        l2Tokens.forEach((l2Token) => {
+          // THe effective balance is the current balance + inbound bridge transfers.
```
pxrl, line 146:
```suggestion
          // The effective balance is the current balance + inbound bridge transfers.
```

src/clients/InventoryClient.ts
```diff
@@ -105,53 +144,90 @@ export class InventoryClient {
...
   // Find how short a given chain is for a desired L1Token.
   getTokenShortFall(l1Token: string, chainId: number): BigNumber {
-    return this.tokenClient.getShortfallTotalRequirement(chainId, this.getDestinationTokenForL1Token(l1Token, chainId));
+    return this.getDestinationTokensForL1Token(l1Token, chainId)
+      .map((token) => this.tokenClient.getShortfallTotalRequirement(chainId, token))
+      .reduce((acc, curr) => acc.add(curr), bnZero);
   }
 
-  getDestinationTokenForL1Token(l1Token: string, chainId: number | string): string {
+  getRepaymentTokenForL1Token(l1Token: string, chainId: number | string): string | undefined {
```
pxrl, line 199: @nicholaspai I think it makes sense to refocus this method on repayment token, rather than destination token. wdyt?

test/InventoryClient.InventoryRebalance.ts
```diff
@@ -31,50 +30,58 @@ let owner: SignerWithAddress, spy: sinon.SinonSpy, spyLogger: winston.Logger;
...
 // Construct an initial distribution that keeps these values within the above thresholds.
 const initialAllocation = {
-  1: { [mainnetWeth]: toWei(100), [mainnetUsdc]: toMegaWei(10000) }, // seed 100 WETH and 10000 USDC on Mainnet
-  10: { [mainnetWeth]: toWei(20), [mainnetUsdc]: toMegaWei(2000) }, // seed 20 WETH and 2000 USDC on Optimism
-  137: { [mainnetWeth]: toWei(10), [mainnetUsdc]: toMegaWei(1000) }, // seed 10 WETH and 1000 USDC on Polygon
-  42161: { [mainnetWeth]: toWei(10), [mainnetUsdc]: toMegaWei(1000) }, // seed 10 WETH and 1000 USDC on Arbitrum
+  [MAINNET]: { [mainnetWeth]: toWei(100), [mainnetUsdc]: toMegaWei(10000) }, // seed 100 WETH and 10000 USDC
+  [OPTIMISM]: { [mainnetWeth]: toWei(20), [mainnetUsdc]: toMegaWei(2000) }, // seed 20 WETH and 2000 USDC
+  [POLYGON]: { [mainnetWeth]: toWei(10), [mainnetUsdc]: toMegaWei(1000) }, // seed 10 WETH and 1000 USDC
+  [BASE]: { [mainnetWeth]: toWei(10), [mainnetUsdc]: toMegaWei(1000) }, // seed 10 WETH and 1000 USDC
```
pxrl, line 79: Base has been added to the initial token allocations in order to test USDbC balance checking. This has forced a few tests to be updated because all the allocation percentages have been affected.

test/InventoryClient.InventoryRebalance.ts
```diff
@@ -121,25 +128,21 @@ describe("InventoryClient: Rebalancing inventory", async function () {
   it("Accessors work as expected", async function () {
     expect(inventoryClient.getEnabledChains()).to.deep.equal(enabledChainIds);
     expect(inventoryClient.getL1Tokens()).to.deep.equal(Object.keys(inventoryConfig.tokenConfig));
-    expect(inventoryClient.getEnabledL2Chains()).to.deep.equal([10, 137, 42161]);
+    expect(inventoryClient.getEnabledL2Chains()).to.deep.equal([OPTIMISM, POLYGON, BASE, ARBITRUM]);
 
-    expect(inventoryClient.getCumulativeBalance(mainnetWeth)).to.equal(initialWethTotal);
-    expect(inventoryClient.getCumulativeBalance(mainnetUsdc)).to.equal(initialUsdcTotal);
+    expect(inventoryClient.getCumulativeBalance(mainnetWeth).eq(initialWethTotal)).to.be.true;
+    expect(inventoryClient.getCumulativeBalance(mainnetUsdc).eq(initialUsdcTotal)).to.be.true;
```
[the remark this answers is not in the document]
pxrl: Factored out to this PR: https://github.com/across-protocol/relayer-v3/pull/1486

src/clients/InventoryClient.ts
nicholaspai: Yeah that makes sense

src/clients/InventoryClient.ts
```diff
@@ -105,53 +144,90 @@ export class InventoryClient {
...
     // return getL2TokenInfo(l1Token, chainId).address
-    return this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, Number(chainId));
+    try {
+      return this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, Number(chainId));
+    } catch {
+      return undefined;
+    }
+  }
+
+  getDestinationTokensForL1Token(l1Token: string, chainId: number | string): string[] {
```
nicholaspai, line 214: I think we might want to rename this to make it clearer that it's looking up the token map using the user supplied inventory config but the fallback is the repayment token mapping

src/interfaces/InventoryManagement.ts
```diff
@@ -25,3 +32,9 @@ export interface InventoryConfig {
   };
   wrapEtherThreshold: BigNumber;
 }
+
+export function isAliasConfig(config: ChainTokenConfig | ChainTokenInventory): config is ChainTokenInventory {
```
nicholaspai, line 36: Why did you choose this name?

src/interfaces/InventoryManagement.ts
```diff
@@ -1,20 +1,27 @@
...
-        thresholdPct: BigNumber; // Threshold, below which, we will execute a rebalance.
-        unwrapWethThreshold?: BigNumber; // Threshold for ETH on this chain to trigger WETH unwrapping to maintain
-        // ETH balance
-        unwrapWethTarget?: BigNumber; // Amount of WETH to unwrap to refill ETH balance. Unused if unwrapWethThreshold
-        // is undefined.
-      };
-    };
-  };
+  // tokenConfig can map to a single token allocation, or a set of allocations that all map to the same HubPool token.
+  tokenConfig: { [l1Token: string]: ChainTokenConfig } | { [l1Token: string]: ChainTokenInventory };
```
nicholaspai, line 23: Can you add a comment to what example shapes of this inventory config would be? It's so deep now it's hard to visualize as easily

src/interfaces/InventoryManagement.ts
pxrl: naming

Naming is hard :) I'm onboard to find something better.

src/interfaces/InventoryManagement.ts
pxrl: Done: https://github.com/across-protocol/relayer-v3/pull/1465/commits/20ad61070de8ab076669657db1730f4b2373b2e8

src/clients/InventoryClient.ts
pxrl: Updated here: https://github.com/across-protocol/relayer-v3/pull/1465/commits/2e1ce4c9aab4efbed42c4b2abaf33586c4087f7f

src/clients/InventoryClient.ts
```diff
@@ -71,32 +72,70 @@ export class InventoryClient {
...
+        .reduce((acc, curr) => acc.add(curr), bnZero);
     }
 
-    // If the chain does not have this token (EG BOBA on Optimism) then 0.
-    const balance =
-      this.tokenClient.getBalance(chainId, this.getDestinationTokenForL1Token(l1Token, chainId)) || bnZero;
-
-    // Consider any L1->L2 transfers that are currently pending in the canonical bridge.
+    // @todo: This will resolve all outstanding transfers for the L1 token, but it should be filtered for _only_
+    // transfers that apply to the specific L2 token. This needs to be fixed in the crossChainTransferClient
```
nicholaspai, line 128: Is there a pr for this upcoming fix?

nicholaspai (approved):
excited to try this!

mrice32 (commented):
Looks great! Only a few small questions/comments

src/clients/InventoryClient.ts
```diff
@@ -105,53 +144,98 @@ export class InventoryClient {
...
     // this function would either return multiple L1Token objects or we'd need to pass in a symbol/l2TokenAddress.
 
     // return getL2TokenInfo(l1Token, chainId).address
-    return this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, Number(chainId));
+    try {
+      return this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, Number(chainId));
+    } catch {
+      return undefined;
+    }
+  }
```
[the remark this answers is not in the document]
pxrl: +1 agree on catch-alls. Rather than emit a custom error, I'd prefer to update the HubPoolClient such that it doesn't throw if the l2 token can't be resolved. Instead, it'd return `undefined` and the caller would be able to decide how to handle that. It's been on my todo list for a while to make this change, so this is a good prompt (unless there are any objections on that).

src/clients/InventoryClient.ts
```diff
@@ -105,53 +144,98 @@ export class InventoryClient {
...
   getTokenDistributionPerL1Token(): TokenDistributionPerL1Token {
     const distributionPerL1Token: TokenDistributionPerL1Token = {};
     this.getL1Tokens().forEach((l1Token) => (distributionPerL1Token[l1Token] = this.getChainDistribution(l1Token)));
     return distributionPerL1Token;
   }
 
   // Get the balance of a given token on a given chain, including shortfalls and any pending cross chain transfers.
-  getCurrentAllocationPct(l1Token: string, chainId: number): BigNumber {
+  getCurrentAllocationPct(l1Token: string, chainId: number, l2Token?: string): BigNumber {
+    l2Token ??= this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, chainId);
```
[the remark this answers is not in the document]
pxrl: Good question...I circled back and realised that the fallback resolution of l2Token isn't even necessary and should never be used, so I've removed it: https://github.com/across-protocol/relayer-v3/pull/1465/commits/4bdc0305fa92744d476b2c1e46dc268afea87b82

src/clients/InventoryClient.ts
```diff
@@ -105,53 +144,98 @@ export class InventoryClient {
       // If token doesn't have entry on chain, skip creating an entry for it since we'll likely run into an error
       // later trying to grab the chain equivalent of the L1 token via the HubPoolClient.
       if (chainId === this.hubPoolClient.chainId || this._l1TokenEnabledForChain(l1Token, chainId)) {
-        const l2Token = this.getDestinationTokenForL1Token(l1Token, chainId);
-        if (cumulativeBalance.gt(bnZero)) {
-          distribution[chainId] ??= {};
-          distribution[chainId][l2Token] = this.getBalanceOnChainForL1Token(chainId, l1Token)
-            .mul(this.scalar)
-            .div(cumulativeBalance);
+        distribution[chainId] ??= {};
```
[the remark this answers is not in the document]
pxrl: Nice catch: https://github.com/across-protocol/relayer-v3/pull/1465/commits/4886cf2c2e8547e3c0f736fda261821f246a4849

src/clients/InventoryClient.ts
pxrl: Yep - https://github.com/across-protocol/relayer-v3/pull/1469

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
mrice32: Agreed, +1 on returning undefined.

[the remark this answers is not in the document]
mrice32: I think this should be done as soon as we can, but it doesn't need to block this from going in. LGTM

mrice32 (approved):
LGTM!

https://github.com/across-protocol/relayer-v3/pull/1465#discussion_r1596255086

PR #1466, opened 2024-05-01
feat(MultiCallClient): add tryMulticall for relayer specific functions

bmzig:
Once spoke pools accommodate `tryMulticall`, the relayer should be able to use this function when bundling functions which are more likely to revert. There are two main considerations:
- When should we distinguish `tryMulticall` from `multicall`? As of current, `tryMulticall` is used when **only** relay-specific functions (`fillV3Relay` and `requestV3SlowFill`) are being aggregated. Given this parameterization, we should never use `tryMulticall` to simulate a transaction, as this would require extra time to parse the return data.
- How should we handle the return data for actual transactions? When there was only `multicall`, a single failed call in the batch caused the entire transaction to revert. Now, transactions do not revert and instead report back whether or not each call was successful. As of writing, this is TBD.


src/clients/MultiCallerClient.ts
```diff
@@ -342,6 +342,11 @@ export class MultiCallerClient {
       // simulated. In this case, drop the aggregation and revert to undefined to force estimation on submission.
       gasLimit = isDefined(gasLimit) && isDefined(txn.gasLimit) ? gasLimit.add(txn.gasLimit) : undefined;
     });
+    const multicallMethod = !simulate
+      ? transactions.every((txn) => ["fillV3Relay", "requestV3SlowFill"].includes(txn.method))
+        ? "tryMulticall"
+        : "multicall"
+      : "multicall";
```
pxrl, line 349: Thoughts on consolidating the logic a bit? I think this should be equivalent.

```suggestion
    const method = !simulate && transactions.every(({ method }) => ["fillV3Relay", "requestV3SlowFill"].includes(method))
        ? "tryMulticall"
        : "multicall";
```

src/clients/MultiCallerClient.ts
```diff
@@ -354,7 +359,7 @@ export class MultiCallerClient {
     return {
       chainId,
       contract,
-      method: "multicall",
+      method: multicallMethod,
```
pxrl, line 362: If you use `method` instead of `multicallMethod` then you can drop the assignment.

```suggestion
      method,
```

test/MultiCallerClient.ts
```diff
@@ -602,4 +602,73 @@ describe("MultiCallerClient", async function () {
...
+        } as unknown as Contract,
+        method: spokePoolMethod,
+        args: [],
+        message: "",
+        mrkdwn: "",
+      };
+      multicallTxns.push(sampleTxn);
+    }
+    const txnQueue: AugmentedTransaction[] = await multiCaller.buildMultiCallBundles(multicallTxns);
+    expect(txnQueue[0].method).to.equal("multicall");
```
pxrl, line 649: Is it worth also checking the length of `txnQueue` here? Alternatively, could do something like:

`txnQueue.forEach(({ method }) => expect(method).to.equal("multicall"));`

src/clients/MultiCallerClient.ts
```diff
@@ -425,14 +431,14 @@ export class MultiCallerClient {
...
+          return txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk, simulate) : txnChunk[0];
         })
         .flat();
     } else {
       // We can support sending multiple transactions to different contracts via an external multisender
       // contract.
       const multicallerTxnBundle = getTxnChunks(multicallerTxns)
-        .map((txnChunk) => (txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk) : txnChunk[0]))
+        .map((txnChunk) => (txnChunk.length > 1 ? this.buildMultiCallBundle(txnChunk, simulate) : txnChunk[0]))
         .flat();
```
pxrl, line 442: In this case the multicall implementation invoked is the external mds multicall3 variant. Is it still necessary to pass in the `simulate` flag here?

pxrl:
This looks good! I like that it's generally a light touch on the MultiCallerClient implementation. There might be some debate about whether to hardcode SpokePool function names directly in the code, but there is already precedent for that here so it's not inconsistent with the existing patterns in this file.

> Given this parameterization, we should never use tryMulticall to simulate a transaction, as this would require extra time to parse the return data.

I think it'd actually be faster to parse the return data because the alternative is we split out each transaction and resubmit them to the RPC provider in paralell for individual simulation. That's relatively expensive in terms of RPC credits and incurs an additional variable delay as we wait for it to resolve.

> How should we handle the return data for actual transactions? When there was only multicall, a single failed call in the batch caused the entire transaction to revert. Now, transactions do not revert and instead report back whether or not each call was successful. As of writing, this is TBD.

It's usually the case that we just want to know that it was confirmed, rather than reading some output from the transaction. But as a rule we typically don't even wait for them to be confirmed. That can fwiw occasionally cause issues, as occurred on Polygon yesterday where a stuck transaction caused no fills to be made for a period of time.

src/clients/MultiCallerClient.ts
bmzig: To be clear, this only executes when we are aggregating a set of multicalls to be executed on the multicall3 contract? In this case, it would make sense to pass in simulate as false, so that we use `tryMulticall` regardless? It looks like aggregate on multicall3 reverts if a single call fails, so this would run into a similar issue if we enforced the use of `multicall`.

pxrl: > To be clear, this only executes when we are aggregating a set of multicalls to be executed on the multicall3 contract? In this case, it would make sense to pass in simulate as false, so that we use tryMulticall regardless? It looks like aggregate on multicall3 reverts if a single call fails, so this would run into a similar issue if we enforced the use of multicall.

Sorry, my comment was wrong! `buildMultiCallBundle()` uses the internal Spoke/Hub multicall, and `buildMultiSenderBundle()` uses the external multicall3 implementation.

src/clients/MultiCallerClient.ts
```diff
@@ -214,12 +214,35 @@ export class MultiCallerClient {
...
       });
-      batchTxns[idx].gasLimit = succeed ? transaction.gasLimit : undefined;
+      if (transaction.method === "tryMulticall") {
+        const succeededTxnRequests: AugmentedTransaction[] = [];
+        data.forEach(({ success }, subIdx) => {
+          if (success) {
+            const txnIdx =
+              this.chunkSize == undefined
+                ? this.chunkSize[chainId] * idx + subIdx
+                : DEFAULT_MULTICALL_CHUNK_SIZE * idx + subIdx;
```
pxrl, line 224: Something like this instead?

```suggestion
            const txnIdx = (this.chunkSize[chainId] ?? DEFAULT_MULTICALL_CHUNK_SIZE) * idx + subIdx;
```

src/clients/MultiCallerClient.ts
```diff
@@ -342,6 +365,9 @@ export class MultiCallerClient {
       // simulated. In this case, drop the aggregation and revert to undefined to force estimation on submission.
       gasLimit = isDefined(gasLimit) && isDefined(txn.gasLimit) ? gasLimit.add(txn.gasLimit) : undefined;
     });
+    const method = transactions.every((txn) => ["fillV3Relay", "requestV3SlowFill"].includes(txn.method))
```
pxrl, line 368: Opportunity: You could consider checking the Contract interface to verify that tryMulticall exists :+1:

src/clients/MultiCallerClient.ts
bmzig: Done. https://github.com/across-protocol/relayer-v3/pull/1466/commits/ad5ff97c0c7b5ffa62684f45d0a80f74f3f9766c

src/clients/MultiCallerClient.ts
```diff
@@ -214,12 +214,35 @@ export class MultiCallerClient {
         batchTxn: { ...transaction, contract: transaction.contract.address },
         reason,
       });
-      batchTxns[idx].gasLimit = succeed ? transaction.gasLimit : undefined;
+      if (transaction.method === "tryMulticall") {
+        const succeededTxnRequests: AugmentedTransaction[] = [];
+        data.forEach(({ success }, subIdx) => {
```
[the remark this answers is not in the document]
bmzig: Data will not be populated in that case (since the `Result` struct is only defined in TransactionUtils when the transaction does not revert), but in that case, the `batchesAllSucceeded` bool will just be false, so all transactions will be simulated individually regardless.

src/clients/MultiCallerClient.ts
```diff
@@ -214,12 +214,32 @@ export class MultiCallerClient {
...
+          if (success) {
+            const txnIdx = (this.chunkSize[chainId] ?? DEFAULT_MULTICALL_CHUNK_SIZE) * idx + subIdx;
+            succeededTxnRequests.push(txns[txnIdx]);
+          }
+        });
+        if (succeededTxnRequests.length != data.length) {
+          batchTxns.splice(idx, 1);
+          txnRequestsToRebuild.push(...succeededTxnRequests);
+        }
```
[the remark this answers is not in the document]
bmzig: I see. I'll keep those guidelines in mind from now on. As far as the early return value goes, if `batchesAllSucceed` is false, all transactions are simulated individually, so there is no speed up benefit here. I think there are a bunch of considerations here; maybe it's best to just use `tryMulticall` for all functions in the spoke pool regardless?

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
mrice32: If it's undefined, won't that cause an exception to be thrown? I think we might want to check that it's defined before calling forEach on it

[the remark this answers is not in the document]
bmzig: You're correct -- it's fixed here https://github.com/across-protocol/relayer-v3/pull/1466/commits/1d74d3e8aa8712d085a7cbf45dcf89d76dfb1214

james-a-morris (commented):
Will circle back on this when the tests clear

bmzig:
I'm closing this because we have another PR here https://github.com/across-protocol/relayer/pull/1676 which should give a smaller diff for the same results.

PR #1467, opened 2024-05-01
feat(TokenClient): Support querying additional USDC deployments

pxrl:
This is the minimum change required to unconditionally query USDC aliases on non-mainnet chains.

Post-CCTP I could imagine that we return to refactor this, such that we have a more common/central handling of token aliases. That would also be useful if Tether ever does something similar.

### DIFF
--- a/src/clients/TokenClient.ts
+++ b/src/clients/TokenClient.ts
@@ -231,7 +231,7 @@ export class TokenClient {
         const { symbol } = hubPoolClient.getL1Tokens().find((hubPoolToken) => hubPoolToken.address === address);
         if (symbol === "USDC") {
           // At the moment, constants-v3 defines native usdc as _USDC.
-          const usdcAliases = ["_USDC"]; // After constants-v3 update: ["USDC.e", "USDbC"]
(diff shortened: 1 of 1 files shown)


src/clients/TokenClient.ts
```diff
@@ -214,16 +216,31 @@ export class TokenClient {
...
         } catch {
-          return undefined;
+          // No known deployment for this token on the SpokePool.
+        }
+
+        // If the HubPool token is USDC then it might map to multiple tokens on the destination chain.
+        const { symbol } = hubPoolClient.getL1Tokens().find((hubPoolToken) => hubPoolToken.address === address);
+        if (symbol === "USDC") {
+          // At the moment, constants-v3 defines native usdc as _USDC.
+          const usdcAliases = ["_USDC"]; // After constants-v3 update: ["USDC.e", "USDbC"]
```
nicholaspai, line 234: Why not add those other symbols in now?

nicholaspai (approved):

src/clients/TokenClient.ts
pxrl: Fair call - this will be a bit more forwards-compatible with the constants-v3 update. Done: https://github.com/across-protocol/relayer-v3/pull/1467/commits/4a034fbd6eee955dc4f967e4693b3eaeb72b5e45

PR #1469, opened 2024-05-01
refactor(CrossChainTransferClient): Support unique L2 tokens

pxrl:
This seems like a necessary refactor in order for rebalances initiated by the InventoryClient to be tracked in a world of 1:many token mappings. It's not complete; it requires a corresponding change in the AdapterManager class, such that the l2Token address is returned as part of `AdapterManager.getOutstandingCrossChainTokenTransferAmount()`.


src/interfaces/index.ts
```diff
@@ -8,7 +9,16 @@ export * from "./Report";
 export * from "./Arweave";
 
 // Bridge interfaces
-export type OutstandingTransfers = interfaces.OutstandingTransfers;
```
pxrl, line 11: Relocated from sdk-v2 because there are no other users of it in the SDK; it seems like it should be defined locally within this repo.

src/clients/InventoryClient.ts
```diff
@@ -630,9 +637,11 @@ export class InventoryClient {
           const amount = deltaPct.mul(cumulativeBalance).div(this.scalar);
           const balance = this.tokenClient.getBalance(1, l1Token);
           // Divide by scalar because allocation percent was multiplied by it to avoid rounding errors.
+          const l2Token = this.getDestinationTokenForL1Token(l1Token, chainId);
```
pxrl, line 640: This is temporary; a follow-on change will ensure that the l2Token is tracked properly, rather than inferred from the l1Token.

nicholaspai:
This PR will make it easier to call the new `getOutstandingCrossChainTransferAmount` from the Relayer, in the lone instance it does so https://github.com/across-protocol/relayer-v3/pull/1479

james-a-morris:
As a note: we currently don't have entries for ZKSync. We'll need to:
1. change the spoke contract for ZKSync to include the CCTPAdapter
2. update the ZKSync adapter contract to include CCTP
3. include the CCTP paths in the relayer/ZKSync adapter

james-a-morris:
@nicholaspai @pxrl if I could get eyes on this - this is the merged #1475 + a second PR. @nicholaspai I sent over a gist + a corresponding GCP log that should point that this looks to be working with some confidence.

james-a-morris (commented):
I shouldn't approve this because I wrote a big chunk of it

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -78,6 +77,7 @@ export class ZKSyncAdapter extends BaseAdapter {
 
         // Resolve whether the token is WETH or not.
         const isWeth = this.isWeth(l1TokenAddress);
+        const l2Token = this.resolveL2TokenAddress(l1TokenAddress, false); // CCTP doesn't exist on ZkSync.
```
nicholaspai, line 80: We should bump `constants-v3` to make sure the zksync entry is removed from _USDC

nicholaspai (commented):
LGTM but there are potentially some missing changes to users of the CrossChainTransferClient interface:
- `getOutstandingCrossChainTransferAmount` is used in the Monitor, Relayer, and InventoryClient
- `getOutstandingCrossChainTransferTxs` is used in the Monitor

These probably went undetected because the new interface only adds optional params but there might be bugs/unexpected consequences if we don't specify the new `l2Token` param. I recommend either making this last param required to as to avoid this issue or patch the downstream calls

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: +1 https://github.com/across-protocol/relayer-v3/pull/1469/commits/eeb8641d3dbb32304c16d465aacb9dda3f1b6eaa

james-a-morris:
> LGTM but there are potentially some missing changes to users of the CrossChainTransferClient interface:
> 
> * `getOutstandingCrossChainTransferAmount` is used in the Monitor, Relayer, and InventoryClient
> * `getOutstandingCrossChainTransferTxs` is used in the Monitor
> 
> These probably went undetected because the new interface only adds optional params but there might be bugs/unexpected consequences if we don't specify the new `l2Token` param. I recommend either making this last param required to as to avoid this issue or patch the downstream calls

I think patching would be ideal - I think there could be some valid use for wanting the total outstanding balance of an L1 token

james-a-morris (approved):

PR #1470, opened 2024-05-02
refactor(InventoryClient): Drop unnecessary token resolutions

pxrl:
The WETH address is known separately from the HubPoolClient so its token address can be queried directly. This eliminates a few uses of getDestinationTokenForL1Token().


nicholaspai (approved):

nicholaspai:
I do something very similar in here https://github.com/across-protocol/relayer-v3/pull/1461/files but I call a utility function to call into token symbols map constant

james-a-morris (approved):

PR #1471, opened 2024-05-02
refactor(InventoryClient): Simplify possible rebalances check

pxrl:
InventoryClient._getPossibleRebalances() accepts a TokenDistributionPerL1Token type, but actually only uses the set of L1 tokens defined. This can be sourced elsewhere in order to simplify the implementation and allow TokenDistributionPerL1Token to be updated.

Note that this commit doesn't attempt to make any functional changes to the InventoryClient - it's moreso decoupling the code so that different parts can be updated to support multiple L2 tokens.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -610,7 +610,7 @@ export class InventoryClient {
     const rebalancesRequired: Rebalance[] = [];
 
     // First, compute the rebalances that we would do assuming we have sufficient tokens on L1.
-    for (const l1Token of Object.keys(this.getL1Tokens())) {
+    for (const l1Token of this.getL1Tokens()) {
       const cumulativeBalance = this.getCumulativeBalance(l1Token);
       if (cumulativeBalance.eq(bnZero)) {
         continue;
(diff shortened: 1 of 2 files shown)


src/clients/InventoryClient.ts
```diff
@@ -95,15 +97,18 @@ export class InventoryClient {
...
+  getChainDistribution(l1Token: string): { [chainId: number]: TokenDistribution } {
     const cumulativeBalance = this.getCumulativeBalance(l1Token);
-    const distribution: { [chainId: number]: BigNumber } = {};
+    const distribution: { [chainId: number]: TokenDistribution } = {};
+
     this.getEnabledChains().forEach((chainId) => {
       // If token doesn't have entry on chain, skip creating an entry for it since we'll likely run into an error
       // later trying to grab the chain equivalent of the L1 token via the HubPoolClient.
       if (chainId === this.hubPoolClient.chainId || this._l1TokenEnabledForChain(l1Token, chainId)) {
+        const l2Token = this.getDestinationTokenForL1Token(l1Token, chainId);
```
pxrl, line 108: This will later be updated to handle multiple l2 tokens that map to a single l1 token.

nicholaspai (approved):

james-a-morris (approved):

james-a-morris:
LGTM pending all passing tests

PR #1473, opened 2024-05-03
hotfix(adapter): correct length of parallel arrays

james-a-morris:
When we attempt to approve tokens, we have to be sure that the l1Tokens and their associated bridges have the same length. The CCTP change made the bridge list longer than the L1 token list.

### DIFF
--- a/src/clients/bridges/ArbitrumAdapter.ts
+++ b/src/clients/bridges/ArbitrumAdapter.ts
@@ -174,6 +174,8 @@ export class ArbitrumAdapter extends CCTPAdapter {
   }
 
   async checkTokenApprovals(address: string, l1Tokens: string[]): Promise<void> {
+    const l1TokenListToApprove = [];
+
     // Note we send the approvals to the L1 Bridge but actually send outbound transfers to the L1 Gateway Router.
     // Note that if the token trying to be approved is not configured in this client (i.e. not in the l1Gateways object)
     // then this will pass null into the checkAndSendTokenApprovals. This method gracefully deals with this case.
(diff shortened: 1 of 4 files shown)


src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -187,10 +189,15 @@ export class ArbitrumAdapter extends CCTPAdapter {
...
+
+        // Push the l1 token to the list of tokens to approve N times, where N is the number of bridges.
+        // I.e. the arrays have to be parallel.
+        l1TokenListToApprove.push(...Array(bridgeAddresses.length).fill(l1Token));
+
         return bridgeAddresses;
       })
       .filter(isDefined);
-    await this.checkAndSendTokenApprovals(address, l1Tokens, associatedL1Bridges);
+    await this.checkAndSendTokenApprovals(address, l1TokenListToApprove, associatedL1Bridges);
```
nicholaspai, line 200: can we assert in BaseAdapter. checkAndSendTokenApprovals that these arrays are equal length?

james-a-morris: Good call

nicholaspai (approved):
https://github.com/across-protocol/relayer-v3/pull/1473/files#r1588596219

src/clients/bridges/ArbitrumAdapter.ts
james-a-morris: https://github.com/across-protocol/relayer-v3/pull/1473/commits/07c45248c673f65c94c323327ab37125f76b7a04

PR #1477, opened 2024-05-03
fix(bridges): Rename l1Gateway to l1Gateways and return string[] always

nicholaspai:
## OpStackAdapter fix:
- If `bridges` is type string, we can't call .length on it otherwise we'll inadvertently use the char count of the string

## PolygonAdapter fix:
- Must add WETH to l1Token list before early exiting


james-a-morris (approved):
Tested locally with a similar bot config to prod. Works LGTM

PR #1479, opened 2024-05-04
improve: Make it clear when we want the L2 versus L1 token info when querying TOKEN_SYMBOLS_MAP

nicholaspai:
Currently we sometimes use the function to grab L2 token info rather than L1 token info.

It doesn't really have any impact because we mostly want just the `symbol` and `decimals` but there is one place in the `Relayer` where we do want the actual L1TokenAddress. I've changed this to call a new function `getL1TokenInfo` instead of `getTokenInfo` which we can use along with the l2 token address already given to query outstanding cross chain transfers correctly following #1469

`getL1TokenInfo` assumes that each token symbols map entry has an L1 token address so this PR also bumps the constants-v3 version that has l1 token addresses for all USDC variants

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "node": ">=16.18.0"
   },
   "dependencies": {
-    "@across-protocol/constants-v2": "1.0.21",
+    "@across-protocol/constants-v2": "1.0.22",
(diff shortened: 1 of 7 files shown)


mrice32 (approved):
LGTM!

PR #1481, opened 2024-05-05
fix(OpStackAdapter): Filter out available l1 tokens when computing outstanding transfers

nicholaspai:
We forgot to include this filter line that we have in other adapters.

### DIFF
--- a/src/clients/bridges/op-stack/OpStackAdapter.ts
+++ b/src/clients/bridges/op-stack/OpStackAdapter.ts
@@ -74,7 +74,7 @@ export class OpStackAdapter extends BaseAdapter {
 
   async getOutstandingCrossChainTransfers(l1Tokens: string[]): Promise<OutstandingTransfers> {
     const { l1SearchConfig, l2SearchConfig } = this.getUpdatedSearchConfigs();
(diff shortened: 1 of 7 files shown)


src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -74,6 +74,7 @@ export class OpStackAdapter extends BaseAdapter {
 
   async getOutstandingCrossChainTransfers(l1Tokens: string[]): Promise<OutstandingTransfers> {
     const { l1SearchConfig, l2SearchConfig } = this.getUpdatedSearchConfigs();
+    const availableL1Tokens = l1Tokens.filter(this.isSupportedToken.bind(this));
```
james-a-morris, line 77: Do we need the bind?

nicholaspai: i don't think so but i think good to stay consistent

james-a-morris: Right but `bind` changes the object scope

james-a-morris (approved):
+1 as long as `bind` isn't an issue

PR #1484, opened 2024-05-06
improve: internal chain adapter filtering

james-a-morris:


nicholaspai (commented):
Can we just remove the `bind` call and replace with this filterSupportedTokens?

nicholaspai (approved):

PR #1487, opened 2024-05-07
improve(relayer): Extend deposit confirmation validation

pxrl:
This change makes the relayer more responsive to unfilled deposits without compromising on safety.

It achieves this by filtering each deposit individually for the minumum number of required confirmation requirements before computing the batch deposit confirmation requirements for set of outstanding deposits. This ensures that individual large deposits that have not met the minimum required threshold cannot DoS lower-value deposits that have met their minimums. In hindsight, this logic is completely obvious.

In order to avoid repeatedly sorting and filtering on the deposit confirmation structure, change its format to be a sorted array. This makes it less computationally-expensive (and less annoying) to evaluate MDC requirements.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -13,7 +13,6 @@ import {
   getNetworkName,
   getUnfilledDeposits,
   isDefined,
-  toBNWei,
   winston,
   fixedPointAdjustment,
 } from "../utils";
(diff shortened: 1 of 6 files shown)


src/relayer/Relayer.ts
```diff
@@ -222,35 +240,28 @@ export class Relayer {
...
         agg[deposit.originChainId] = (agg[deposit.originChainId] ?? bnZero).add(unfilledAmountUsd);
         return agg;
       }, {});
 
-    // Sort thresholds in ascending order.
-    const minimumDepositConfirmationThresholds = Object.keys(minDepositConfirmations)
-      .filter((x) => x !== "default")
-      .sort((x, y) => Number(x) - Number(y));
-
     // Set the MDC for each origin chain equal to lowest threshold greater than the unfilled USD deposit amount.
```
pxrl, line 248: We probably want to revisit this approach in a follow-up commit because it still permits a single large deposit to implicitly block other lower-value deposits by raising their deposit confirmation requirements. I'm still thinking about how best to mitigate this safely, but I'd also really like to see the effect of the current change in production because I think it'll probably have the largest impact vs. current fill behaviour.

src/relayer/Relayer.ts
```diff
@@ -222,35 +240,28 @@ export class Relayer {
...
       .filter((deposit) => tokenClient.hasBalanceForFill(deposit))
       .reduce((agg, deposit) => {
-        const unfilledAmountUsd = profitClient.getFillAmountInUsd(deposit, deposit.outputAmount);
+        const unfilledAmountUsd = profitClient.getFillAmountInUsd(deposit);
         agg[deposit.originChainId] = (agg[deposit.originChainId] ?? bnZero).add(unfilledAmountUsd);
         return agg;
       }, {});
 
-    // Sort thresholds in ascending order.
-    const minimumDepositConfirmationThresholds = Object.keys(minDepositConfirmations)
```
nicholaspai, line 231: Why do we need these mdcs per chain anymore if we're checking the deposits individually? Where do we use this?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: This enforces confirmation requirements for the sum of all deposits on the origin chain. It's still important to do this because chains with a high degree of activity will have lots of lower-value deposits that might sum to a significant value.

PR #1491, opened 2024-05-07
improve(ProviderUtils): Add fields we can ignore when requesting eth_getLogs

nicholaspai:
These fields are not set for zksync eth_getLogs responses returned by Alchemy. They are not used in our code so we don't need to compare them for quorum purposes


mrice32 (approved):
LGTM!

PR #1492, opened 2024-05-08
chore: Bump SDK

pxrl:
The ConfigStore, HubPool and SpokePool client classes now implement more tolerant behaviour when requested to update() repeatedly. This will permit the fast relayer instance to cycle faster than 1 mainnet block per loop.


src/clients/SpokePoolClient.ts
```diff
@@ -219,6 +219,7 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
       latestDepositId,
       searchEndBlock: this.pendingBlockNumber,
       events,
+      hasCCTPBridgingEnabled: false, // @todo: Update indexer to query this.
```
pxrl, line 222: Will follow up with this change after syncing w/ @james-a-morris

nicholaspai (approved):

james-a-morris (approved):

PR #1494, opened 2024-05-08
feat: Add toggleUSDC.ts script for Polygon

nicholaspai:
This is a USDC specific script that assumes we want to produce the following calldata

1. Deposit all routes to/from *Bridged* USDC for a `chainId`
2. Enable all routes to the `chainId` where `inputToken` is Bridged USDC
3. Enable all routes from `chainId` where the `inputToken` Native USDC

Note we'll have to change this script when migrating to CCTP for all other chains because at that point, Native USDC will already be enabled on Polygon. We'll have to do something like:

1. Deposit all routes to/from *Bridged* USDC for a `chainId`, unless that `chainId == 137` then disable from *Native* USDC
2. Enable all routes to the `chainId` where `inputToken` is Native USDC
3. Enable all routes from `chainId` where the `inputToken` Native USDC

### DIFF
--- a/scripts/toggleUSDC.ts
+++ b/scripts/toggleUSDC.ts
@@ -1,0 +1,153 @@
+/* eslint-disable no-process-exit */
+import { retrieveSignerFromCLIArgs, winston, Logger, assert, TOKEN_SYMBOLS_MAP } from "../src/utils";
+import { CommonConfig, constructClients } from "../src/common";
+
+// Example run commands:
+
+// Disable USDC.e to and from Polygon
+// $ ts-node ./scripts/toggleUSDC.ts
+// \ --wallet mnemonic
+// \ --chainId 137
+// \ --enable false
+
+// Enable Native USDC to and from Polygon
+// $ ts-node ./scripts/toggleDepositRoute.ts
+// \ --wallet mnemonic
+// \ --chainId 137
+// \ --enable true
+
+import minimist from "minimist";
+const args = minimist(process.argv.slice(2), {
+  string: ["chainId"],
+  boolean: ["enable"],
+});
+
+export function getTokenAddressForChain(l2ChainId: number, isNativeUsdc = false): string {
+  if (isNativeUsdc) {
+    const hasL2TokenEntry = TOKEN_SYMBOLS_MAP["_USDC"].addresses[l2ChainId] !== undefined;
+    if (!hasL2TokenEntry) {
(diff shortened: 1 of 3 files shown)


mrice32:
> 1. Deposit all routes to/from _Bridged_ USDC for a `chainId`

Should this be `Disable all routes...`? Just want to make sure I'm understanding the flow

scripts/toggleUSDC.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export async function run(logger: winston.Logger): Promise<void> {
+  const chainId = Number(args.chainId);
+  const enable = args.enable;
+  assert(chainId && enable !== undefined);
+
+  // If enabling routes, we want to enable Native USDC if it exists, otherwise enable USDC.e.
+  const nativeUsdc = enable ? TOKEN_SYMBOLS_MAP["_USDC"].addresses[chainId] !== undefined : false;
+  const tokenSymbol = getTokenSymbol(chainId, nativeUsdc);
+  const token = getTokenAddressForChain(chainId, enable);
```
mrice32, line 59: Maybe I'm missing something, but why use `enable` here instead of `nativeUsdc` as is done in the call on 58.

Don't we only want native USDC if `nativeUsdc` is true?

mrice32 (approved):
Looks good! Just one nit/question

scripts/toggleUSDC.ts
nicholaspai: yeah well specifically we only want to use the native USDC route if we're enabling. This script is very opinionated and specific in that its supposed to support the following two flows in order:
1. Disable all USDC.e routes involving Polygon
2. Re-enable all USDC.e routes to Polygon and Native USDC from Polygon

scripts/toggleUSDC.ts
```diff
@@ -0,0 +1,153 @@
...
+  const allChainIds = [1, 10, 137, 324, 42161, 8453, 59144];
+
+  // Save all deposit routes involving chain ID that we want to toggle.
+  /**
+     *  function setDepositRoute(
+            uint256 originChainId,
+            uint256 destinationChainId,
+            address originToken,
+            bool depositsEnabled
+        ) public override nonReentrant onlyOwner {
```
james-a-morris, line 83: Is this leftover from referencing the contract code? Maybe we can replace with a permalink

PR #1496, opened 2024-05-08
improve(TokenClient): multicall balance & allowance queries

pxrl:
TokenClient update is occasionally observed to hang for 10s of seconds. It's a surprisingly RPC-intensive method because it issues a parallel RPC query for each supported token on a chain. This doesn't scale well with each new chain or token that's added, so bundle the queries into a multicall. The legacy implementation is retained for chains that don't support multicall3, which is nominally just hardhat in test.

To simplify this change, split out the HubPool BondToken query from the regular update() call. The BondToken part is so specific that it could optionally be pulled out of the TokenClient entirely.


src/clients/TokenClient.ts
```diff
@@ -242,39 +243,55 @@ export class TokenClient {
       })
       .flat();
 
-    const tokenData = Object.fromEntries(
-      await Promise.all(
-        tokens.map(async (token) => {
-          const balance: BigNumber = await token.balanceOf(this.relayerAddress);
-          const allowance = await this._getAllowance(this.spokePoolClients[chainId], token);
-          return [token.address, { balance, allowance }];
+    const multicall3 = async (multicall: Contract) => {
```
pxrl, line 246: nb. will probably factor the multicall3 part out to the SDK. There's similar functionality in the frontend repository so it probably makes sense to consolidate that.

src/clients/TokenClient.ts
```diff
@@ -242,39 +243,55 @@ export class TokenClient {
...
-      if (result !== null) {
-        return toBN(result);
-      }
-    }
-    const allowance: BigNumber = await token.allowance(this.relayerAddress, spokePoolClient.spokePool.address);
-    if (allowance.gte(MAX_SAFE_ALLOWANCE) && redis) {
-      // Save allowance in cache with no TTL as these should be exhausted.
-      await redis.set(key, MAX_SAFE_ALLOWANCE);
-    }
-    return allowance;
```
pxrl, line 277: Caching is dropped here because multicall3 is expected to be available on most/all chains supported by Across, and the fallback logic that queries the allowance is almost certainly only executed in test. The implementation is however backwards compatible with any chain that doesn't support multicall3, so the impact of some future chain not supporting it is pretty marginal.

src/clients/TokenClient.ts
pxrl, line 246: With a little bit of additional refactoring it should be possible to bundle the bondToken() call into the multicall as well, so only 1 RPC call would be issued per chain instead of 2. The benefits of this are probably only a marginal improvement over this current PR though.

src/clients/TokenClient.ts
nicholaspai: Is there any sense in caching the full token-chain  set? Its a big set but i imagine it doesn't change much and we can cache it for like 2-3 mins

pxrl: You mean caching the complete set of allowances & balances? We could, but in looping mode we rely on updating balances on each loop, because we _assume_ that fills made by the relayer are successful, but we never actually verify that. So the only way we have of getting accurate balances is to update the TokenClient.

In terms of the allowance check, it could obviously be cached, but it's basically free to bundle that into the balance check, so we don't really gain anything by caching it.

pxrl:
I've factored out the multicall3 aggregate() call to the SDK. The corresponding PR is here: https://github.com/across-protocol/sdk-v3/pull/644. This PR is now subject to that going in and a new SDK being produced.

src/clients/TokenClient.ts
pxrl: Done: https://github.com/across-protocol/relayer-v3/pull/1496/commits/afdaaa2f3d38a5bdecb7455c327a4a5755c70eac

pxrl: PR: https://github.com/across-protocol/sdk-v3/pull/644

src/clients/TokenClient.ts
```diff
@@ -152,41 +152,102 @@ export class TokenClient {
...
+        }
+
+        // If the HubPool token is USDC then it might map to multiple tokens on the destination chain.
+        if (symbol === "USDC") {
+          // At the moment, constants-v3 defines native usdc as _USDC.
+          const usdcAliases = ["_USDC", "USDC.e", "USDbC"]; // After constants-v3 update: ["USDC.e", "USDbC"]
+          usdcAliases
+            .map((symbol) => TOKEN_SYMBOLS_MAP[symbol]?.addresses[chainId])
+            .filter(isDefined)
+            .forEach((address) => tokenAddrs.push(address));
```
nicholaspai, line 195: Can't you use AddressUtils.getUsdcAddress here?

nicholaspai (approved):
One potential code reuse suggestion otherwise lgtm

src/clients/TokenClient.ts
pxrl: Resolved out of band :+1:

PR #1498, opened 2024-05-09
feat: Bump config store version to 4

nicholaspai:
This is the minimum version required to support new USDC routes. All relayers should bump this version consciously when they are ready to support Native USDC on L2s. This means taking repayments in Native USDC instead of Bridged USDC. This also means that if outputToken = address(0), that the new default output token is Native USDC for routes where Native USDC is enabled

Do not merge this until the first Native USDC pool rebalance route is enabled for Polygon


mrice32 (approved):
LGTM!

pxrl (approved):
+1 happy to see this.

james-a-morris (approved):

PR #1501, opened 2024-05-09
fix(test): Avoid external RPC call

pxrl:
Identified w/ James.


nicholaspai (approved):
Ah nie I saw this too

james-a-morris (approved):

PR #1504, opened 2024-05-10
fix(TransactionUtils): Don't scale maxFeePerGas downwards

pxrl:
If one of the scalers is < 1 then it might cause the gas price to be unintentionally reduced.


src/utils/TransactionUtils.ts
```diff
@@ -158,13 +158,13 @@ export async function getGasPrice(
...
   // Handle chains with legacy pricing.
-  if (feeData.maxPriorityFeePerGas.eq(0)) {
+  if (feeData.maxPriorityFeePerGas.eq(bnZero)) {
     return { gasPrice: scaleByNumber(feeData.maxFeePerGas, priorityScaler) };
   }
 
   // Default to EIP-1559 (type 2) pricing.
   return {
-    maxFeePerGas: scaleByNumber(feeData.maxFeePerGas, priorityScaler * maxFeePerGasScaler),
+    maxFeePerGas: scaleByNumber(feeData.maxFeePerGas, Math.max(priorityScaler * maxFeePerGasScaler, 1)),
```
pxrl, line 167: We could also just use `maxFeePerGasScaler` here, and put the onus on the operator to ensure that the priority and base scalers are set in concert.

james-a-morris (approved):

PR #1506, opened 2024-05-10
fix(InventoryClient): Ignore mainnet in rebalance evaluation

pxrl:
The set of rebalance chainIds was incorrectly changed from `getEnabledL2Chains()` to `getEnabledChains()` in the previous commit.


pxrl:
Here's where this was changed in the previous commit - the impact is pretty obvious.

diff

PR #1508, opened 2024-05-10
improve(logs): distinguish usdc in rebalance logs

james-a-morris:
This should distinguish the USDC tokens when handling rebalances.


nicholaspai:
I think this solution handles more cases, like how the Relayer logs out the output token symbol : https://github.com/across-protocol/relayer-v3/pull/1509

james-a-morris:
Closed in lieu of https://github.com/across-protocol/relayer-v3/pull/1509

PR #1509, opened 2024-05-10
fix: Handle L2 USDC symbol lookup 

nicholaspai:
Handle case where an L2 token for chain ID can map to more than one TOKEN_SYMBOLS_MAP entry. For example, L2 Bridged USDC maps to both the USDC and USDC.e/USDbC entries in TOKEN_SYMBOLS_MAP.

This PR is a temporary patch that handles the known USDC issue where both USDC.e/USDbC contain the same chain-address combinations as the USDC entry. This can be reverted once we remove the USDC entry in the TOKEN_SYMBOLS_MAP.

### DIFF
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -100,6 +100,17 @@ export function getTokenAddressWithCCTP(
   return getTokenAddress(l1Token, hubChainId, l2ChainId);
 }
 
+/**
+ * Get the USDC symbol for the given token address and chain ID.
+ * @param l2Token A Web3 token address (not case sensitive)
+ * @param chainId A chain Id to reference
+ * @returns Either USDC (if native) or USDbC/USDC.e (if bridged) or undefined if the token address is not recognized.
+ */
+export function getUsdcSymbol(l2Token: string, chainId: number): string | undefined {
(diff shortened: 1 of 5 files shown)


james-a-morris (approved):
Fair call +1

src/utils/AddressUtils.ts
```diff
@@ -100,6 +100,17 @@ export function getTokenAddressWithCCTP(
...
 
+/**
+ * Get the USDC symbol for the given token address and chain ID.
+ * @param l2Token A Web3 token address (not case sensitive)
+ * @param chainId A chain Id to reference
+ * @returns Either USDC (if native) or USDbC/USDC.e (if bridged) or undefined if the token address is not recognized.
+ */
+export function getUsdcSymbol(l2Token: string, chainId: number): string | undefined {
+  const compareToken = (token?: string) => isDefined(token) && compareAddressesSimple(l2Token, token);
+  return ["_USDC", "USDbC", "USDC.e"].find((token) => compareToken(TOKEN_SYMBOLS_MAP[token].addresses[chainId]));
```
mrice32, line 111: In case one of those symbols stops being used/populated in the map in the future.
```suggestion
  return ["_USDC", "USDbC", "USDC.e"].find((token) => compareToken(TOKEN_SYMBOLS_MAP[token]?.addresses?.[chainId]));
```

PR #1510, opened 2024-05-10
fix(finalizer): enrich l1 -> l2 addresses

james-a-morris:
This PR aims to push a fix to finalize CCTP events.


nicholaspai (approved):
will querying spoke pool on some of the L2 to L1 transfers throw any weird errors?

james-a-morris:
> will querying spoke pool on some of the L2 to L1 transfers throw any weird errors?

This is only relevant for the L1 -> L2 routes.

PR #1512, opened 2024-05-10
feat(finalizer): specify finalization direction

james-a-morris:
We should probably be able to specify finality direction as we need to have different frequency intervals for l1 & l2 finalizations.


nicholaspai (approved):
I liek this a lot

PR #1518, opened 2024-05-13
fix(finalizer): Handle CCTP "pending" messages

nicholaspai:


src/utils/CCTPUtils.ts
```diff
@@ -250,34 +267,10 @@ async function _resolveCCTPRelatedTxns(
...
-    throw new Error("Failed to generate attestation proof");
-  }
-  // Return the attestation proof since it was generated
-  return attestationResponse.attestation;
+async function generateCCTPAttestationProof(messageHash: string, isMainnet: boolean): Promise<Attestation> {
+  const httpResponse = await axios.get<Attestation>(
+    `https://iris-api${isMainnet ? "" : "-sandbox"}.circle.com/attestations/${messageHash}`
+  );
+  const attestationResponse = httpResponse.data;
+  return attestationResponse;
```
james-a-morris, line 275: This will only be valid on 200 response - we may want to account for the 404 case

src/finalizer/utils/cctp/l1ToL2.ts
```diff
@@ -126,6 +120,7 @@ async function generateMultiCallData(
   messageTransmitter: Contract,
   messages: DecodedCCTPMessage[]
 ): Promise<Multicall2Call[]> {
+  assert(messages.every((message) => isDefined(message.attestation)));
```
james-a-morris, line 123: Should we also define this assert in the `generateDepositData`

james-a-morris (commented):
Good reduction - two comments

src/utils/CCTPUtils.ts
nicholaspai: what's the 404 case though?

james-a-morris: If the attestation doesn't exist yet - should happen if called immediately after the deposit txn is made

nicholaspai: handled offline but this is aactually a 200 case. 404 case is a programmer error

james-a-morris (approved):

PR #1524, opened 2024-05-14
make adapters more generic

james-a-morris:
WIP - not ready for review.


src/clients/bridges/generics/BaseChainAdapter.ts
```diff
@@ -0,0 +1,249 @@
...
+          assign(outstandingTransfers, [monitoredAddress, l1Token, l2Token], {
+            totalAmount: outstandingInitiatedEvents.reduce((acc, event) => acc.add(event.amount), BigNumber.from(0)),
+            depositTxHashes: outstandingInitiatedEvents.map((event) => event.transactionHash),
+          });
+        });
+      });
+    });
+
+    this.baseL1SearchConfig.fromBlock = l1SearchConfig.toBlock + 1;
+    this.baseL1SearchConfig.fromBlock = l2SearchConfig.toBlock + 1;
```
pxrl, line 245:
```suggestion
    this.baseL2SearchConfig.fromBlock = l2SearchConfig.toBlock + 1;
```

src/clients/bridges/generics/BaseChainAdapter.ts
```diff
@@ -0,0 +1,249 @@
...
+  private transactionClient: TransactionClient;
+
+  constructor(
+    private readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    protected readonly chainId: number,
+    protected readonly hubChainId: number,
+    protected readonly monitoredAddresses: string[],
+    protected readonly logger: winston.Logger,
+    protected readonly supportedTokens: SupportedTokenSymbol[],
+    protected readonly bridges: { [l1Token: string]: BaseBrideAdapter },
```
pxrl, line 42:
```suggestion
    protected readonly bridges: { [l1Token: string]: BaseBridgeAdapter },
```
```suggestion
    protected readonly bridges: { [l1Token: string]: BaseBrideAdapter },
```

src/clients/bridges/generics/types.ts
```diff
@@ -0,0 +1,28 @@
+import { BigNumber } from "ethers";
+import { SortableEvent } from "../../../interfaces";
+
+export interface DepositEvent extends SortableEvent {
+  amount: BigNumber;
+  transactionHash: string;
```
pxrl, line 6: transactionHash is already a part of SortableEvent.

```suggestion
```

bmzig: Fixed: https://github.com/across-protocol/relayer/pull/1524/commits/8ea569a6a9b4818c52a7d8f1982de8a3b492e2c3

src/clients/bridges/generics/BaseChainAdapter.ts
bmzig: Fixed: https://github.com/across-protocol/relayer/pull/1524/commits/8ea569a6a9b4818c52a7d8f1982de8a3b492e2c3

src/clients/bridges/generics/BaseChainAdapter.ts
bmzig: Fixed: https://github.com/across-protocol/relayer/pull/1524/commits/8ea569a6a9b4818c52a7d8f1982de8a3b492e2c3

bmzig:
I believe the goal here is to make a single generic adapter and then be able to paramerterize it based on what bridges we pass into the constructor. So for example, the adapter instantiated for Optimism will have the canonical bridge as well as CCTP and the DAI/SNX custom bridges passed in. The adapter logic itself should then just go through what bridges it "owns" and find outstanding transfers etc. 

As of the latest commit, this is not done yet. The main thing is to make sure that all the different types of bridges are set up properly so that James's `BaseChainAdapter` can work its magic. There will definitely be some low-hanging-fruit refactors once all the bridges are set up correctly, too.

src/common/Constants.ts
```diff
@@ -329,6 +329,24 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
   ],
 };
 
+// A mapping of L2 chain IDs to an array of tokens Across supports on that chain.
+export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
+  10: ["DAI", "SNX", "BAL", "ETH", "WETH", "USDC", "POOL", "USDT", "WBTC", "UMA", "ACX"],
+  137: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "BADGER", "POOL", "MATIC"],
```
pxrl, line 335:
```suggestion
  137: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "BADGER", "POOL"],
```

src/common/Constants.ts
```diff
@@ -329,6 +329,24 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+  324: ["USDC", "USDT", "WETH", "WBTC", "DAI"],
+  8453: ["BAL", "DAI", "ETH", "WETH", "USDC", "POOL"],
+  34443: ["ETH", "WETH", "USDC", "USDT", "WBTC"],
+  42161: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BADGER", "BAL", "ACX", "POOL"],
```
pxrl, line 339:
```suggestion
  42161: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "POOL"],
```

src/common/Constants.ts
pxrl, line 1: wdyt about getting this part in earlier? It should help us to remove some of the snowflake OP-stack adapters in src/clients/bridges/op-stack/{base,mode}.

src/clients/bridges/utils.ts
pxrl, line 1: I think we could pull some of this functionality out and submit it sooner in a smaller PR, which would also remove some of the corresponding code from the existing bridge adapters.

src/common/Constants.ts
```diff
@@ -329,6 +329,24 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+  59144: ["USDC", "USDT", "WETH", "WBTC", "DAI"],
```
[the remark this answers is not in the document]
bmzig: Should be addressed here: https://github.com/across-protocol/relayer/pull/1618

src/common/Constants.ts
bmzig: We went over this in a 1-1. This is done in this PR: https://github.com/across-protocol/relayer/pull/1619

bmzig:
I still don't think there is much point to reviewing this now since it's so large. There is still work to be done on formatting the bridges.

src/common/Constants.ts
```diff
@@ -329,6 +343,85 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+    [TOKEN_SYMBOLS_MAP.WETH[1]]: PolygonWethBridge,
+    [TOKEN_SYMBOLS_MAP.USDC[1]]: UsdcTokenSplitterBridge,
+  },
+  324: {},
+  8453: {
+    [TOKEN_SYMBOLS_MAP.USDC[1]]: UsdcTokenSplitterBridge,
+    [TOKEN_SYMBOLS_MAP.WETH[1]]: WethBridge,
+  },
+  34443: {
+    [TOKEN_SYMBOLS_MAP.WETH[1]]: WethBridge,
```
bmzig, line 415: Adding a comment here since I'm pretty confident I have incorrectly identified which custom bridges correspond to which chain. This will need to be changed.

bmzig:
There are some issues with querying deposit/finalization events with the ZkSync Weth bridge, since there are multiple different events to listen for. Also, in general, the `BridgeEvent` interface is a bit too specific, since we can't add things like Linea deposit hashes/token addresses to the event data.

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,102 @@
...
+    return {
+      [this.resolveL2TokenAddress(l1Token)]: events.map(processEvent),
+    };
+  }
+
+  async queryL2BridgeFinalizationEvents(
+    l1Token: string,
+    fromAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
```
pxrl, line 77: This function prototype breaks down for zkSync because it's not guaranteed that the depositor address is the same as the recipient address. This is the case for HubPool -> SpokePool transfers. I suspect we'll need to permit the depositor and receipient addresses separately in the BaseBridgeAdapter.

bmzig: Very true. This is still not fixed.

I think we can get the other chain-specific tests in here as soon as they get merged into master, and the ZkSync one will probably have us address this. In the meantime, I can get started working on the finalizer code.

src/adapter/bridges/BaseBridgeAdapter.ts
```diff
@@ -0,0 +1,50 @@
+import { Contract, BigNumber, EventSearchConfig, Signer, Provider, getTokenAddressWithCCTP } from "../../utils";
+
+export interface BridgeTransactionDetails {
+  readonly contract: Contract;
+  readonly method: string;
+  readonly args: unknown[];
+  readonly value?: BigNumber;
+}
```
pxrl, line 8: I'm wondering if this type is even needed, since it's actually just a subset of the `AugmentedTransaction` type that we already use.

src/adapter/bridges/ArbitrumBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+      };
+    };
+    return {
+      [this.resolveL2TokenAddress(l1Token)]: events.map(processEvent),
+    };
+  }
+
+  async queryL2BridgeFinalizationEvents(
+    l1Token: string,
+    fromAddress: string,
```
pxrl, line 95: Per discussion, I suspect we need this to be `toAddress` (for zkSync at least) but I'm not sure whether it'll potentially break other chains. In this case it looks like we actually need `fromAddress` the filter, which creates a headache...

pxrl (commented):
This is really ambitious, but it's looking solid. Here are some comments from my initial check; I haven't managed to read all the changes yet but will circle back once I have.

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+          const bridges = {};
+          const l2Signer = spokePoolClients[chainId].spokePool.signer;
+
+          SUPPORTED_TOKENS[chainId].map((symbol) => {
+            const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[hubChainId];
+            const bridgeConstructor = CUSTOM_BRIDGE[chainId][l1Token] ?? CANONICAL_BRIDGE[chainId];
+
+            // For bridges like Arbitrum and Polygon, we need to supply additional information about the token
+            // so that we may properly select the l1Gateway
+            if (chainIsMatic(chainId) && !CUSTOM_BRIDGE[chainId][l1Token]) {
```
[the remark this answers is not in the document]
bmzig: Good catch. Fixed here: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+                hubChainId,
+                l1Signer,
+                l2Signer,
+                this.l2TokenForL1Token(l1Token, chainId)
+              );
+            } else if (chainIsArbitrum(chainId)) {
+              bridges[l1Token] = new bridgeConstructor(chainId, hubChainId, l1Signer, l2Signer, l1Token);
+            } else {
+              bridges[l1Token] = new bridgeConstructor(chainId, hubChainId, l1Signer, l2Signer, l1Token);
+            }
```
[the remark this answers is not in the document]
bmzig: Yeah there is no difference it is fixed here: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794.

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+          // TODO: Do something about the gas multiplier
+          this.adapters[chainId] = new BaseChainAdapter(
+            spokePoolClients,
+            chainId,
+            hubChainId,
+            filterMonitoredAddresses(chainId),
+            logger,
+            SUPPORTED_TOKENS[chainId],
+            bridges,
+            DEFAULT_GAS_FEE_SCALERS[chainId]?.maxFeePerGasScaler ?? Number(DEFAULT_RELAYER_GAS_MULTIPLIER)
```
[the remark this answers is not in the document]
bmzig: Fixed: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  supportedChains(): number[] {
+    return Object.keys(this.adapters).map((chainId) => Number(chainId));
+  }
+
+  getOutstandingCrossChainTokenTransferAmount(chainId: number, l1Tokens: string[]): Promise<OutstandingTransfers> {
+    const adapter = this.adapters[chainId];
+    // @dev The adapter should filter out tokens that are not supported by the adapter, but we do it here as well.
+    const adapterSupportedL1Tokens = l1Tokens.filter((token) =>
+      adapter.supportedTokens.includes(this.hubPoolClient.getTokenInfo(CHAIN_IDs.MAINNET, token).symbol)
```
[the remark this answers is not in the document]
bmzig: Also fixed: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+      message: `Getting outstandingCrossChainTransfers for ${chainId}`,
+      adapterSupportedL1Tokens,
+      searchConfigs: adapter.getUpdatedSearchConfigs(),
+    });
+    return this.adapters[chainId].getOutstandingCrossChainTransfers(adapterSupportedL1Tokens);
+  }
+
+  sendTokenCrossChain(
+    address: string,
+    chainId: number | string,
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+  async wrapEthIfAboveThreshold(inventoryConfig: InventoryConfig, simMode = false): Promise<void> {
+    await mapAsync(
+      this.chainsToWrapEtherOn.filter((chainId) => isDefined(this.spokePoolClients[chainId])),
+      async (chainId) => {
+        const wrapThreshold =
+          inventoryConfig?.wrapEtherThresholdPerChain?.[chainId] ?? inventoryConfig.wrapEtherThreshold;
+        const wrapTarget = inventoryConfig?.wrapEtherTargetPerChain?.[chainId] ?? inventoryConfig.wrapEtherTarget;
+        assert(
+          wrapThreshold.gte(wrapTarget),
+          `wrapEtherThreshold ${wrapThreshold.toString()} must be >= wrapEtherTarget ${wrapTarget.toString()}`
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/CrossChainTransferClient.ts
```diff
@@ -0,0 +1,102 @@
...
+    // No specific l2Token specified; return the set of all l1Token transfers to chainId.
+    return Object.values(transfers).flatMap(({ depositTxHashes }) => depositTxHashes);
+  }
+
+  getEnabledChains(): number[] {
+    return this.chainIdList;
+  }
+
+  getEnabledL2Chains(): number[] {
+    return this.getEnabledChains().filter((chainId) => chainId !== 1);
```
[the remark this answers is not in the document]
bmzig: We can. I've just stored it in the class. https://github.com/across-protocol/relayer/pull/1524/commits/3b9b5bb698d6983644d330153bd4c67f6ad48794

src/adapter/CrossChainTransferClient.ts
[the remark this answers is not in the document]
bmzig: It does not. I haven't looked at it yet, but It definitely is a good idea to clean that up while we can.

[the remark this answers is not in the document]
pxrl: If there are no changes required and we can get away without duplicating the existing instance then I'd probably also be inclined to drop it from this PR, and potentially make cleanup/changes to to the existing file in a separate PR. wdyt?

src/adapter/bridges/DaiOptimismBridge.ts
```diff
@@ -0,0 +1,104 @@
...
+  BigNumberish,
+} from "../../utils";
+import { CONTRACT_ADDRESSES } from "../../common";
+import { SortableEvent } from "../../interfaces";
+import { BaseBridgeAdapter, BridgeTransactionDetails, BridgeEvents } from "./BaseBridgeAdapter";
+import { Event } from "ethers";
+
+export class DaiOptimismBridge extends BaseBridgeAdapter {
+  private readonly l1Bridge: Contract;
+  private readonly l2Bridge: Contract;
```
pxrl, line 18: For test we probably want to override these, so it would help if they were `protected` instead of `private readonly`.

src/adapter/bridges/DefaultErc20Bridge.ts
pxrl, line 1: This looks very very close (identical?) to `DaiOptimismBridge`. Do they share the same ABI, with the only difference being a different deployment address?

src/adapter/bridges/UsdcCCTPBridge.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.hubChainId,
+      this.l2chainId,
+      fromAddress
+    );
+
+    return {
+      [this.resolveL2TokenAddress(l1Token)]: events.map(processEvent),
+    };
+  }
+  queryL2BridgeFinalizationEvents(
```
pxrl, line 70:
```suggestion

  queryL2BridgeFinalizationEvents(
```

src/adapter/bridges/WethBridge.ts
pxrl, line 1: This adapter has a fairly generic name and appears to handle the AtomicDepositor, but it's scoped exclusively to OVM chains. It looks like the zkSync variant is named accordingly; should this be named `OVMWethBridge` or something?

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,161 @@
...
+      ? await zksync.utils.estimateDefaultBridgeDepositL2Gas(
+          l1Provider,
+          zkProvider,
+          l1Token,
+          amount,
+          toAddress,
+          toAddress,
+          this.gasPerPubdataLimit
+        )
+      : 2_000_000;
```
pxrl, line 66: This is a pre-existing issue, but returning a number here results in l2GasLimit type awkwardness (union number | BigNumber). We can just assign a BN here to fix that.

```suggestion
      : BigNumber.from(2_000_000);
```

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -0,0 +1,163 @@
...
+      ? await zksync.utils.estimateDefaultBridgeDepositL2Gas(
+          l1Provider,
+          zkProvider,
+          l1Token,
+          amount,
+          toAddress,
+          toAddress,
+          this.gasPerPubdataLimit
+        )
+      : 2_000_000;
```
pxrl, line 68:
```suggestion
      : BigNumber.from(2_000_000);
```

src/adapter/types.ts
pxrl, line 1: I'd probably be inclined to just reuse the existing type definitions we have for these in `src/clients/bridges` (or wherever they currently are). We can relocate these residual files as part of deprecating/removing the existing bridge adapters, and doing it that way helps to reduce the risk of accidentally breaking something.

src/common/Constants.ts
```diff
@@ -345,6 +345,91 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
   ],
 };
 
+// A mapping of L2 chain IDs to an array of tokens Across supports on that chain.
+export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
+  10: ["DAI", "SNX", "BAL", "WETH", "USDC", "POOL", "USDT", "WBTC", "UMA", "ACX"],
+  137: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "POOL"],
+  288: [],
```
pxrl, line 352: ooc, why is Boba showing up again? We haven't supported it for a long time so I'd hoped we could survive without it.

src/adapter/CrossChainTransferClient.ts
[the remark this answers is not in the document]
bmzig: Yes. I think that's a good idea. I removed it: https://github.com/across-protocol/relayer/pull/1524/commits/efb7bdf35f9e57baf5592bd749c86c7c82838fa7

src/common/Constants.ts
```diff
@@ -1,4 +1,20 @@
-import { CHAIN_IDs, TOKEN_SYMBOLS_MAP, ethers } from "../utils";
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP, ethers, Signer, Provider } from "../utils";
+import {
+  BaseBridgeAdapter,
+  DefaultERC20Bridge,
```
pxrl, line 4: Might be nice to scope this import so that it's easy to spot what type of chain it is.

```suggestion
  DefaultERC20Bridge as OVMDefaultERC20Bridge,
```

pxrl (commented):
2nd half of my first pass - I'm really positive about the size of these chain adapters, and I think we can probably squeeze them down even further. I think that's a really good sign.

src/common/Constants.ts
bmzig: Fixed. I specified both this and the Weth bridge to be OpStack[bridge]: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/common/Constants.ts
bmzig: It was included due to us automatically instantiating a spoke pool client for Boba. Regardless, I removed it in this commit: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/types.ts
bmzig: Good call. Types is no more: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/ZKSyncBridge.ts
bmzig: Same as below: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/WethBridge.ts
bmzig: Good call. I renamed this and the `DefaultErc20Bridge` https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/UsdcCCTPBridge.ts
bmzig: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/DefaultErc20Bridge.ts
bmzig: I did not notice this. They do have the same ABI, so I just removed the `DaiOptimismBridge` ABIs and now just use the DefaultErc20Bridge ABIs in `common/ContractAddresses`: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/DaiOptimismBridge.ts
```diff
@@ -0,0 +1,104 @@
...
+        from: string;
+        transactionHash: string;
+      };
+      return {
+        amount: eventSpread["_amount"],
+        to: eventSpread["_to"],
+        from: eventSpread["_from"],
+        transactionHash: eventSpread.transactionHash,
+      };
+    };
```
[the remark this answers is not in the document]
bmzig: Yes. I factored out all these functions in this commit: https://github.com/across-protocol/relayer/pull/1524/commits/efb7bdf35f9e57baf5592bd749c86c7c82838fa7

src/adapter/bridges/DaiOptimismBridge.ts
```diff
@@ -0,0 +1,104 @@
...
+      this.l1Bridge.filters.ERC20DepositInitiated(l1Token, undefined, fromAddress),
+      eventConfig
+    );
+    const processEvent = (event: Event) => {
+      const eventSpread = spreadEventWithBlockNumber(event) as SortableEvent & {
+        amount: BigNumberish;
+        to: string;
+        from: string;
+        transactionHash: string;
+      };
```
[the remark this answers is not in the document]
bmzig: Done on this one (using the second suggestion): https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/DaiOptimismBridge.ts
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/bridges/ArbitrumBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+} from "../../utils";
+import { CONTRACT_ADDRESSES, CUSTOM_ARBITRUM_GATEWAYS } from "../../common";
+import { SortableEvent } from "../../interfaces";
+import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./BaseBridgeAdapter";
+import { Event } from "ethers";
+
+const DEFAULT_ERC20_GATEWAY = {
+  l1: "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC",
+  l2: "0x09e9222E96E7B4AE2a407B98d48e330053351EEe",
+};
```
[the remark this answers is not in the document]
bmzig: For the time being, I renamed this bridge `ArbitrumOneBridge` to emphasize that this is not a generic Arbitrum Orbit bridge. https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020

src/adapter/CrossChainTransferClient.ts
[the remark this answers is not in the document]
bmzig: I actually added it back in here: https://github.com/across-protocol/relayer/pull/1524/commits/5aacd351220a191e4ba57c07e729bccbbdb36020 so that we can do simulations with the generic adapters without having to edit anything in `src/clients/bridges`. I am unresolving so this will be addressed again before we merge.

src/adapter/bridges/BaseBridgeAdapter.ts
bmzig: This is probably just meant for better code structure since we only use it in `constructL1ToL2Txn.` I think it's nice to have but we can also remove it (is there a better way to format the responses of `constructL1ToL2Txn` such that we do not need the interface?)

src/adapter/bridges/UsdcTokenSplitterBridge.ts
```diff
@@ -0,0 +1,94 @@
...
+      this.canonicalBridge.queryL1BridgeInitiationEvents(l1Token, fromAddress, eventConfig),
+    ]);
+    // Reduce the events to a single Object. If there are any duplicate keys, merge the events.
+    return events.reduce((acc, event) => {
+      Object.entries(event).forEach(([l2Token, events]) => {
+        if (l2Token in acc) {
+          acc[l2Token] = acc[l2Token].concat(events);
+        } else {
+          acc[l2Token] = events;
+        }
```
pxrl, line 66: Simplification (...acknowledging that this is pre-existing code):

```suggestion
        acc[l2Token] = (acc[l2Token] ?? []).concat(events);
```

pxrl (commented):
Have made another pass over this. It's great seeing so many bridge adapters at < 100 LoC.

src/adapter/bridges/DefaultErc20Bridge.ts
pxrl: Nice! I've also pulled that into a separate PR here: https://github.com/across-protocol/relayer/pull/1634

pxrl: It'd ultimately be nice to squash the DAI bridge adapter entirely, but for that we need the OP stack adapter to support that its bridge addresses are supplied to the constructor, rather than resolving them based directly from `CONTRACT_ADDRESSES`. We can figure that out in subsequent work :+1:

src/clients/InventoryClient.ts
```diff
@@ -25,7 +25,8 @@ import {
   getUsdcSymbol,
 } from "../utils";
 import { HubPoolClient, TokenClient, BundleDataClient } from ".";
-import { AdapterManager, CrossChainTransferClient } from "./bridges";
+import { AdapterManager, CrossChainTransferClient } from "../adapter";
+// import { AdapterManager, CrossChainTransferClient } from "./bridges";
```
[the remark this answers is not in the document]
bmzig: Yes. I am thinking that we may want to enable using these generic adapters for a lower-risk bot upon merging, like the monitor? I'll keep this open so I don't forget to replace the import.

src/adapter/bridges/UsdcTokenSplitterBridge.ts
bmzig: Fixed; https://github.com/across-protocol/relayer/pull/1524/commits/538402a8c4a932e154e43e6955373dc31d2787cd

src/adapter/bridges/UsdcCCTPBridge.ts
```diff
@@ -0,0 +1,89 @@
...
+  ): Promise<BridgeEvents> {
+    // Lint Appeasement
+    l1Token;
+    fromAddress;
+    eventConfig;
+    assert(compareAddressesSimple(l1Token, TOKEN_SYMBOLS_MAP.USDC.addresses[this.hubChainId]));
+
+    // Per the documentation of the BaseAdapter's computeOutstandingCrossChainTransfers method, we can return an empty array here
+    // and only return the relevant outstanding events from queryL1BridgeInitiationEvents.
+    // Relevant link: https://github.com/across-protocol/relayer/blob/master/src/clients/bridges/BaseAdapter.ts#L189
```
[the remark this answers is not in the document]
bmzig: I just replaced the comment with one without a link: https://github.com/across-protocol/relayer/pull/1524/commits/538402a8c4a932e154e43e6955373dc31d2787cd

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,214 @@
...
+  sendTokenCrossChain(
+    address: string,
+    chainId: number,
+    l1Token: string,
+    amount: BigNumber,
+    simMode = false,
+    l2Token?: string
+  ): Promise<TransactionResponse> {
+    this.logger.debug({ at: "AdapterManager", message: "Sending token cross-chain", chainId, l1Token, amount });
+    l2Token ??= this.l2TokenForL1Token(l1Token, Number(chainId));
```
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/538402a8c4a932e154e43e6955373dc31d2787cd

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,214 @@
...
+              chainId,
+              hubChainId,
+              l1Signer,
+              l2Signer,
+              this.l2TokenForL1Token(l1Token, chainId)
+            );
+          } else {
+            bridges[l1Token] = new bridgeConstructor(chainId, hubChainId, l1Signer, l2Signer, l1Token);
+          }
+        });
```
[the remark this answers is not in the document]
bmzig: I hope this answers the question: The main issue is for the Arbitrum bridge, we derive the gateway address based on what the L1Token is. There are 3 different Arbitrum gateways corresponding to USDT/USDC, WETH, and ERC20s. Those need to be l1 token addresses. For Polygon, we need to know the associated L2 token address since finalization queries look for mints from the zero address from those L2 tokens, so essentially, both Arbitrum and Polygon are special cases of the `BaseChainAdapter`. 

Ultimately, the Arbitrum and Polygon bridges need some information about what token they are bridging. Arbitrum bridges needs info about the L1 address of their token, while Polygon needs info about their L2 Token. I just default to passing in the L1 token for the sake of brevity (no other canonical bridge adapter needs to know what token they are bridging).

src/adapter/bridges/DefaultErc20Bridge.ts
bmzig: Sounds good!

src/adapter/bridges/ArbitrumBridge.ts
bmzig: I've gotten much more confused about this as time passed, since it really looks like we are applying aliases to our atomic depositor and hub pool addresses in our current `ZkSyncAdapter`.

src/adapter/BaseChainAdapter.ts
```diff
@@ -0,0 +1,254 @@
...
+    const { l1SearchConfig, l2SearchConfig } = this.getUpdatedSearchConfigs();
+    const availableL1Tokens = this.filterSupportedTokens(l1Tokens);
+
+    const outstandingTransfers: OutstandingTransfers = {};
+
+    await forEachAsync(this.monitoredAddresses, async (monitoredAddress) => {
+      await forEachAsync(availableL1Tokens, async (l1Token) => {
+        const bridge = this.bridges[l1Token];
+        const [depositInitiatedResults, depositFinalizedResults] = await Promise.all([
+          bridge.queryL1BridgeInitiationEvents(l1Token, monitoredAddress, undefined, l1SearchConfig),
```
bmzig, line 226: If this is a L2 Spoke pool address, will this break the filtering?

bmzig:
All bots _should_ work in simulation mode now.

src/adapter/AdapterManager.ts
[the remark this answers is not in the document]
pxrl: Ah, I see now. I had misread and missed that we were passing in an l2Token to the Polygon adapter!

Is it valid to pass `l1Token` in for the other adapters? I don't see it defined in (for example) the OP stack adapter, so I'm wondering if all of the constructor function signatures match 1:1.

In the case of the Polygon adapter, could we consider resolving the L2 token inside of its constructor? I think it could potentially resolve the l2 token via TOKEN_SYMBOLS_MAP - especially since we're unlikely to need to change any mappings for Polygon in future.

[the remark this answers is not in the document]
bmzig: That's a good call. It's probably best to delegate all "edge case" logic to the bridges so we can keep the AdapterManager clean. It's fixed here:https://github.com/across-protocol/relayer/pull/1524/commits/aea4ba28c257df778c88b2533c517e5ce8c678f6

This also fixes the constructor signatures.

src/common/abi/DaiOptimismBridgeL1.json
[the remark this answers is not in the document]
bmzig: Ok. I added it back in: https://github.com/across-protocol/relayer/pull/1524/commits/0a49a9bd1e1f6cdb7384ab81397b5e725fff9c61

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,75 @@
...
+        // TODO: Do something about the gas multiplier
+        this.adapters[chainId] = new BaseChainAdapter(
+          spokePoolClients,
+          chainId,
+          hubChainId,
+          filterMonitoredAddresses(chainId),
+          logger,
+          SUPPORTED_TOKENS[chainId],
+          bridges,
+          DEFAULT_GAS_FEE_SCALERS[hubChainId]?.maxFeePerGasScaler ?? Number(DEFAULT_RELAYER_GAS_MULTIPLIER)
```
pxrl, line 65: In the pre-existing adapters, this was just hardcoded per implementation. See here for example: https://github.com/across-protocol/relayer/blob/746bda6b77f67e5fe29e32ac41bb0c8b0cde131b/src/clients/bridges/op-stack/OpStackAdapter.ts#L151

The objective of it is to work around gas estimation issues that might occur on mainnet and cause the transaction to revert. Optimism is the only chain where I'm aware of this being an issue, so we default to a multiplier of 1.0 and override to 1.5 for Optimism only.

bmzig: I added a new constant to address that here: https://github.com/across-protocol/relayer/pull/1524/commits/871538abf35f0a0553c2b3eca52074fa5573a89d

src/adapter/bridges/DaiOptimismBridge.ts
```diff
@@ -0,0 +1,76 @@
+import { Contract, BigNumber, paginatedEventQuery, EventSearchConfig, Signer, Provider } from "../../utils";
+import { CONTRACT_ADDRESSES } from "../../common";
+import { BaseBridgeAdapter, BridgeTransactionDetails, BridgeEvents } from "./BaseBridgeAdapter";
+import { processEvent } from "../utils";
+
+export class DaiOptimismBridge extends BaseBridgeAdapter {
```
james-a-morris, line 6: If we extend OpStackDefaultErc20Bridge then we just need the constructor here.

james-a-morris: cc @bmzig

bmzig: Nice catch. Fixed in this one: https://github.com/across-protocol/relayer/pull/1524/commits/3ecd389920e6e39e1f1c617c66c8f0272660a5ef

src/adapter/bridges/OpStackWethBridge.ts
```diff
@@ -0,0 +1,184 @@
...
+  TOKEN_SYMBOLS_MAP,
+} from "../../utils";
+import { CONTRACT_ADDRESSES } from "../../common";
+import { matchL2EthDepositAndWrapEvents, processEvent } from "../utils";
+import { utils } from "@across-protocol/sdk";
+import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./BaseBridgeAdapter";
+
+export class OpStackWethBridge extends BaseBridgeAdapter {
+  protected l1Bridge: Contract;
+  protected l2Bridge: Contract;
```
james-a-morris, line 19: It looks like we have an l1/l2 bridge in every iteration of these child classes. Thoughts on hoisting to the parent?

bmzig: I did something like that here: https://github.com/across-protocol/relayer/pull/1524/commits/a0b8f8c84c572b4f148aaffa20b2e6b73c8c7b43

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  async queryL1BridgeInitiationEvents(
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    const events = await paginatedEventQuery(
+      this.l1Bridge,
+      this.l1Bridge.filters.DepositInitiated(undefined, fromAddress, fromAddress),
```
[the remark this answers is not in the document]
pxrl: nb. I managed to make the zkSync test mostly compatible with this new structure, so if we can get that PR in then I can update this PR to take advantage of those tests.

https://github.com/across-protocol/relayer/pull/1632

src/common/Constants.ts
```diff
@@ -326,6 +343,72 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+  34443: {
+    [TOKEN_SYMBOLS_MAP.WETH.addresses[1]]: OpStackWethBridge,
+  },
+  42161: {
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[1]]: UsdcTokenSplitterBridge,
+  },
+  59144: {
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[1]]: LineaUSDCBridge,
+    [TOKEN_SYMBOLS_MAP.WETH.addresses[1]]: LineaWethBridge,
+  },
```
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/158d307142c42c6d614cbd052713739149ddbdd8

src/common/Constants.ts
```diff
@@ -326,6 +343,72 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+    ): BaseBridgeAdapter;
+  };
+} = {
+  10: OpStackDefaultERC20Bridge,
+  137: PolygonERC20Bridge,
+  324: ZKSyncBridge,
+  8453: OpStackDefaultERC20Bridge,
+  34443: OpStackDefaultERC20Bridge,
+  42161: ArbitrumOneBridge,
+  59144: LineaBridge,
```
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/158d307142c42c6d614cbd052713739149ddbdd8

src/adapter/bridges/ZKSyncBridge.ts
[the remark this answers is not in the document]
bmzig: I still need to fix this since using `toAddress` here means I'll have to tweak the BaseChainAdapter class a little. Getting https://github.com/across-protocol/relayer/pull/1632 in is likely pretty important to guide the development.

src/adapter/utils.ts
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/2d99d9f5161ec52cd7fbbc59612f0eecdc86005a

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,131 @@
...
+      ? await zksync.utils.estimateDefaultBridgeDepositL2Gas(
+          l1Provider,
+          zkProvider,
+          l1Token,
+          amount,
+          toAddress,
+          toAddress,
+          this.gasPerPubdataLimit
+        )
+      : BigNumber.from(2_000_000);
```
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/2d99d9f5161ec52cd7fbbc59612f0eecdc86005a

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,131 @@
...
+ * bridge, we need to assume that the l1 and l2 signers contain
+ * associated providers, since we need to get information about
+ * addresses and gas prices (this is also why `constructL1toL2Txn`
+ * is an async fn).
+ */
+export class ZKSyncBridge extends BaseBridgeAdapter {
+  protected zkSyncMailbox: Contract;
+
+  private readonly gasPerPubdataLimit = zksync.utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT;
+  private readonly l2GasLimit = 2_000_000; // We should dynamically define this.
```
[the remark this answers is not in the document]
bmzig: Done: https://github.com/across-protocol/relayer/pull/1524/commits/2d99d9f5161ec52cd7fbbc59612f0eecdc86005a

src/common/Constants.ts
```diff
@@ -327,6 +344,72 @@ export const TOKEN_APPROVALS_TO_FIRST_ZERO: Record<number, string[]> = {
...
+    [TOKEN_SYMBOLS_MAP.WETH.addresses[1]]: OpStackWethBridge,
+  },
+  42161: {
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[1]]: UsdcTokenSplitterBridge,
+  },
+  59144: {
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[1]]: LineaUSDCBridge,
+    [TOKEN_SYMBOLS_MAP.WETH.addresses[1]]: LineaWethBridge,
+  },
+};
```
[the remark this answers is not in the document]
bmzig: Went ahead and just added Blast: https://github.com/across-protocol/relayer/pull/1524/commits/2d99d9f5161ec52cd7fbbc59612f0eecdc86005a

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,68 @@
...
+          this.spokePoolClients[chainId].spokePool.address === address ||
+          !spokePoolAddresses.includes(address)
+      );
+    };
+
+    const l1Signer = spokePoolClients[hubChainId].spokePool.signer;
+
+    Object.values(CHAIN_IDs)
+      .filter((chainId) => isDefined(this.spokePoolClients[chainId]) && chainId !== hubChainId)
+      .map((chainId) => {
```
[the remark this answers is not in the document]
bmzig: Good call. Saves extra loops: https://github.com/across-protocol/relayer/pull/1524/commits/2d99d9f5161ec52cd7fbbc59612f0eecdc86005a

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,131 @@
...
+    // The l2TransactionBaseCost needs to be included as msg.value to pay for the transaction. its a bit of an
+    // overestimate if the estimatedL1GasPrice and/or l2GasLimit are overestimates, and if its insufficient then the
+    // L1 transaction will revert.
+
+    const estimatedL1GasPrice = l1GasPriceData.maxPriorityFeePerGas.add(l1GasPriceData.maxFeePerGas);
+    const l2TransactionBaseCost = await this.zkSyncMailbox.l2TransactionBaseCost(
+      estimatedL1GasPrice,
+      l2GasLimit,
+      this.gasPerPubdataLimit
+    );
```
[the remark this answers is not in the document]
bmzig: I've started working on "merging" the two PRs together in the sense of getting the work done there into the zksync bridge code. This is done now: https://github.com/across-protocol/relayer/pull/1524/commits/bbca12b159643ba9a36da9ffccff0256235c22cc

src/adapter/bridges/ZKSyncBridge.ts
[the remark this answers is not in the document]
bmzig: Fixed here: https://github.com/across-protocol/relayer/pull/1524/commits/bbca12b159643ba9a36da9ffccff0256235c22cc

src/adapter/BaseChainAdapter.ts
```diff
@@ -0,0 +1,268 @@
...
+    const { l1SearchConfig, l2SearchConfig } = this.getUpdatedSearchConfigs();
+    const availableL1Tokens = this.filterSupportedTokens(l1Tokens);
+
+    const outstandingTransfers: OutstandingTransfers = {};
+
+    await forEachAsync(this.monitoredAddresses, async (monitoredAddress) => {
+      await forEachAsync(availableL1Tokens, async (l1Token) => {
+        const bridge = this.bridges[l1Token];
+        const [depositInitiatedResults, depositFinalizedResults] = await Promise.all([
+          // TODO: Do the toAddress field in these queries need to be aliased for special cases, e.g. zkSync.
```
pxrl, line 239: The recipient should not be aliased for any currently-supported chains, but most (if not all) currently supported chains apply aliasing on the sender address. Each chain can impose its own aliasing scheme so I think it makes sense for aliasing to be applied within the adapter itself.

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -0,0 +1,168 @@
...
+    amount: BigNumber
+  ): Promise<BridgeTransactionDetails> {
+    const l1Provider = this.atomicDepositor.provider;
+    const l2Provider = this.l2Eth.provider;
+
+    let zkProvider;
+    try {
+      zkProvider = zkSyncUtils.convertEthersRPCToZKSyncRPC(l2Provider);
+    } catch (error) {
+      // We do not have a logger in this class
```
pxrl, line 63: We set `RELAYER_TEST=true` in the environment if we're running in test, so it might be an idea to `assert(process.env.RELAYER_TEST === "true")` here.

src/clients/bridges/LineaAdapter.ts
```diff
@@ -32,7 +32,8 @@ export class LineaAdapter extends BaseAdapter {
     const { LINEA } = CHAIN_IDs;
     super(spokePoolClients, LINEA, monitoredAddresses, logger, SUPPORTED_TOKENS[LINEA]);
   }
-  async checkTokenApprovals(address: string, l1Tokens: string[]): Promise<void> {
+  async checkTokenApprovals(l1Tokens: string[]): Promise<void> {
+    const address = await this.getSigner(this.hubChainId).getAddress();
     // Note: Linea has two bridges: one for
```
pxrl, line 37: Dangling comment

src/adapter/BaseChainAdapter.ts
pxrl, line 1: Probable bikeshed: I feel like `src/bridges` and `src/bridges/adapters` or even `src/chains` and `src/chains/adapters` is a more obvious naming for this new directory structure, because `src/adapters` might invite other unrelated but adapter-ish things to be placed there. These aren't strongly held feelings though.

pxrl (commented):
@bmzig I think this is ready to go in once the tests are passing :+1:

src/adapter/BaseChainAdapter.ts
bmzig: I agree that this is a better idea.

src/adapter/bridges/SnxOptimismBridge.ts
```diff
@@ -0,0 +1,73 @@
...
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // @dev For the SnxBridge, only the `toAddress` is indexed on the L2 event so we treat the `fromAddress` as the
+    // toAddress when fetching the L1 event.
+    const events = await paginatedEventQuery(
+      this.getL1Bridge(),
+      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress),
+      eventConfig
+    );
```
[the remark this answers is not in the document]
bmzig: Yeah, that can be reverted. Done here: https://github.com/across-protocol/relayer/pull/1524/commits/eb75450fb8d36579c9134c91d1f02df5b08d8187

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: I think this is a good idea. https://github.com/across-protocol/relayer/pull/1524/commits/eb75450fb8d36579c9134c91d1f02df5b08d8187

src/adapter/bridges/ZKSyncBridge.ts
```diff
@@ -0,0 +1,178 @@
...
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    const isSpokePool = await isContractDeployedToAddress(toAddress, this.l2Bridge.provider);
+    // If the monitored address is the hub pool address, then we automatically assume that
+    // we are tracking a Hub -> Spoke transfer; therefore, we manually set the toAddress.
+    // Otherwise, we make no assumptions of the tracking and make the associated query.
+    const events = isSpokePool
```
bmzig, line 106: I just want to point out that this now does do the same thing as the adapter here https://github.com/across-protocol/relayer/pull/1632. In practice, I would think that making the assumption that all monitored transfers to a spoke pool is from the hub pool is fair.

pxrl: Yeah; this work has clarified for me that I think we should only track recipient addresses in `monitoredAddress`. I _think_ that should be available in all cases, and it would avoid the need to monitor the HubPool's own address, since the adapters can just sub it in on the fly when necessary.

This could be a follow-up refinement to explore, once this current PR is in.

src/adapter/BaseChainAdapter.ts
bmzig: This makes sense. We can do this in a follow up.

src/adapter/AdapterManager.ts
```diff
@@ -0,0 +1,219 @@
...
+      .filter((chainId) => chainId !== hubChainId)
+      .map((chainId) => {
+        if (this.spokePoolClients[chainId] !== undefined) {
+          // First, fetch all the bridges associated with the chain.
+          const bridges = {};
+          const l2Signer = spokePoolClients[chainId].spokePool.signer;
+
+          SUPPORTED_TOKENS[chainId].map((symbol) => {
+            const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[hubChainId];
+            const bridgeConstructor = CUSTOM_BRIDGE[chainId][l1Token] ?? CANONICAL_BRIDGE[chainId];
```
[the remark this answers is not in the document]
bmzig: This is still a problem. We don't have a fix for it in this PR, but it's worth the reminder before we merge this.

pxrl (approved):
@bmzig James never populated a commit message for this PR so it'd be good to add a good description when merging :+1:

PR #1526, opened 2024-05-15
fix(finalizer): Look up L1TokenInfo using TOKEN_SYMBOLS_MAP

nicholaspai:
These l2-->l2 finalizers will all break when trying to finalize an l2 token that is no longer linked to l1 via PoolRebalanceRoutes, like USDC.e for 137,10,8453,42161

We haven't seen this yet because polygon, the only CCTP enabled chain so far, has not withdawn any funds from L2 to L1


james-a-morris (approved):
+1 these token resolutions from the hubpool may be an issue elsewhere

PR #1531, opened 2024-05-16
fix(TokenClient): Use correct signer on direct token queries

pxrl:
The recent change to use multicall incorrectly uses the hub chain signer for remote chain tokens. This results in very confusing and counter-intuitive failures when the remote chain does not support multicall.

Also:
- There's no need to query the HubPoolClient for L2 token addresses when the target chain is the hub chain itself.
- Remove a redundant Promise.all().


bmzig (approved):

mrice32 (approved):
LGTM!

james-a-morris (commented):
LGTM - a bit late

PR #1532, opened 2024-05-16
improve(indexer): More debug logging

pxrl:
In pursuit of a transient error in production. It may be related to 5xx errors on the WebSocketProvider.


james-a-morris (approved):
Would we want to set stop = true in this location? or do we have retry logic

pxrl:
> Would we want to set stop = true in this location? or do we have retry logic

Ultimately yeah - it probably does make sense to do that here. I'm just trying to gather information with this PR though, so there'll be a follow-up change that modifies the logic based on the outcome of this logging.

PR #1533, opened 2024-05-19
fix(reporter): Accumulate L2 token balances mapped to same L1 token address

nicholaspai:
Currently we are only querying L2 tokens mapped to l1 tokens via PoolRebalanceRoutes, which means we are not reporting Bridged USDC balances


mrice32 (approved):
LGTM!

PR #1534, opened 2024-05-19
improve: Replace _USDC with USDC mapping in TOKEN_SYMBOLS_MAP

nicholaspai:
This PR was developed using a locally linked `constants-v3` version so it requires downloading the latest NPM release with the _USDC entry removed to be merged


james-a-morris (commented):
I think we can now merge https://github.com/across-protocol/constants-v3/pull/41

But we may want to do a minor/major version bump because it is breaking

mrice32 (approved):
LGTM!

PR #1535, opened 2024-05-19
fix(Finalizer): Dedup transaction hash list when creating list of transactions to query for CCTP l2->l1 transactions

nicholaspai:
We’re not de-duping the set of txn receipts to query for CCTP l2 to l1 deposits, but we do this in the l1 to l2 direciton.

What’s unique for L2 to L1 is we look for SpokePool.TokensBridged() events but we forget there might be multiple of these in the same block

### DIFF
--- a/src/finalizer/utils/cctp/l2ToL1.ts
+++ b/src/finalizer/utils/cctp/l2ToL1.ts
@@ -6,6 +6,7 @@ import { CONTRACT_ADDRESSES, Multicall2Call, chainIdsToCctpDomains } from "../..
 import {
   Contract,
   Signer,
+  TOKEN_SYMBOLS_MAP,
   assert,
   getBlockForTimestamp,
   getCurrentTime,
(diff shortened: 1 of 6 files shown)


src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -67,14 +68,25 @@ async function resolveRelatedTxnReceipts(
...
+
+  const uniqueTxnHashes = new Set<string>();
+  client
+    .getTokensBridged()
+    .filter(
+      (bridgeEvent) =>
+        bridgeEvent.blockNumber >= latestBlockToFinalize &&
+        bridgeEvent.l2TokenAddress === TOKEN_SYMBOLS_MAP._USDC.addresses[sourceChainId]
+    )
+    .forEach((bridgeEvent) => uniqueTxnHashes.add(bridgeEvent.transactionHash));
```
mrice32, line 82: Would this break if there were 2 CCTP sends in the same origin txn?

src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -67,14 +68,25 @@ async function resolveRelatedTxnReceipts(
...
+  const sourceChainId = client.chainId;
+  // Dedup the txnReceipt list because there might be multiple tokens bridged events in the same txn hash.
+
+  const uniqueTxnHashes = new Set<string>();
+  client
+    .getTokensBridged()
+    .filter(
+      (bridgeEvent) =>
+        bridgeEvent.blockNumber >= latestBlockToFinalize &&
+        bridgeEvent.l2TokenAddress === TOKEN_SYMBOLS_MAP._USDC.addresses[sourceChainId]
```
nicholaspai, line 80: separately, we should apply a simple filter here to narrow the txn receipt list down

src/finalizer/utils/cctp/l2ToL1.ts
nicholaspai: no, `_resolveCCTPRelatedTxns` is designed to filter on only the specific log indices of CCTP sends in a txn hash

src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -67,14 +66,25 @@ async function resolveRelatedTxnReceipts(
...
+    )
+    .forEach((bridgeEvent) => uniqueTxnHashes.add(bridgeEvent.transactionHash));
```
james-a-morris, line 80: Should we make sure this is all in the same casing?

src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -67,14 +66,25 @@ async function resolveRelatedTxnReceipts(
...
+  const sourceChainId = client.chainId;
+  // Dedup the txnReceipt list because there might be multiple tokens bridged events in the same txn hash.
+
+  const uniqueTxnHashes = new Set<string>();
+  client
+    .getTokensBridged()
+    .filter(
+      (bridgeEvent) =>
+        bridgeEvent.blockNumber >= latestBlockToFinalize &&
+        bridgeEvent.l2TokenAddress === TOKEN_SYMBOLS_MAP._USDC.addresses[sourceChainId]
```
pxrl: Could optionally use sdkUtils.dedupArray() instead of filtering on the l2 address.

james-a-morris (approved):

src/finalizer/utils/cctp/l2ToL1.ts
nicholaspai: how does that help? the filter removes non-USDC withdrawals while dedup would not

src/finalizer/utils/cctp/l2ToL1.ts
mrice32: kk makes sense

mrice32 (approved):

PR #1537, opened 2024-05-19
improve(error-handling): finalizer batch

james-a-morris:
We were previously only testing that txns were being simulated in isolation. This PR adds an additional check to make sure that each chain is submitting correct multicall data to the blockchain so that the errors are reported. 

Note: we should be using `tryMulticall` so that we can see which multicall is actually failing, but I think that this is a good fallback candidate in the meantime.


src/finalizer/index.ts
```diff
@@ -275,6 +275,34 @@ export async function finalize(
     return false;
   });
 
+  // Ensure that all transactions would succeed in a batch.
+  const failedChainFinalizations = await sdkUtils.filterAsync(
```
nicholaspai, line 279: I think this makes sense but more and more of this simulate-individually, simulate-as-batch, submit flow is replicated in the `MulticallerClient`. Can we use the `MulticallerClient` somehow for the finalizer?

nicholaspai (commented):
https://github.com/across-protocol/relayer-v3/pull/1537/files#r1606857570

src/finalizer/index.ts
nicholaspai: The benefits of using the `MulticallerClient`:
- error reporting is better, we'd get paged earlier on failures
- if a batch txn fails, then txns are simulated and submitted individually, so this problem this weekend wouldn't have happened

Longer term we should use [`tryAggregate`](https://github.com/mds1/multicall/blob/49ac24ba5d18fb495d6323393da495a4513fc018/src/Multicall3.sol#L60) in the `MulticallerClient` to replace some of the bespoke batch simulation logic to more easily pick out failed txns

james-a-morris:
We can close this out in lieu of a more robust approach using `useAggregate`

PR #1538, opened 2024-05-19
fix(finalizer): polygon

james-a-morris:
`TokenBridge` events that use CCTP to transfer tokens off Polygon are causing `UNKNOWN` errors to be thrown in the GCP logs. The reasoning is that our simulations are trying to simulate `exit()` with bridge events that did not originate on the canonical bridge. 

We can't directly filter out the `TokenBridge` events before the simulation logic because we need to be aware of potentially multiple events in one transaction for the `getUniqueLogIndex` and the subsequent call to `buildPayloadForExit`.

This PR simply makes these CCTP message have an error code of `NON_CANONICAL_BRIDGE` which prevents the message from actually being simulated and being pushed to the `finalizableMessages` array.

--

Note: this bug is mainly causes noise & shouldn't directly impact finalization since we catch this error and log as a `debug`

### DIFF
--- a/src/finalizer/utils/polygon.ts
+++ b/src/finalizer/utils/polygon.ts
@@ -13,6 +13,8 @@ import {
   getRedisCache,
   getBlockForTimestamp,
   getL1TokenInfo,
+  compareAddressesSimple,
+  TOKEN_SYMBOLS_MAP,
 } from "../../utils";
 import { EthersError, TokensBridged } from "../../interfaces";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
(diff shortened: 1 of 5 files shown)


src/finalizer/utils/polygon.ts
```diff
@@ -110,6 +112,12 @@ async function getFinalizableTransactions(
   const exitStatus = await Promise.all(
     checkpointedTokensBridged.map(async (_, i) => {
       const payload = payloads[i];
+      const { chainId, l2TokenAddress } = tokensBridged[i];
+
+      if (compareAddressesSimple(l2TokenAddress, TOKEN_SYMBOLS_MAP._USDC.addresses[chainId])) {
```
nicholaspai, line 117: Can we filter this out earlier in the process?

james-a-morris: We can't directly filter out the TokenBridge events before the simulation logic because we need to be aware of potentially multiple events in one transaction for the getUniqueLogIndex and the subsequent call to buildPayloadForExit.

nicholaspai: can't we do it on line 63?

nicholaspai: ah ok i see your point

nicholaspai:
I added filters for arbitrum and optimism in this PR which you can merge into your branch https://github.com/across-protocol/relayer-v3/pull/1539

src/finalizer/utils/polygon.ts
james-a-morris: It's a bit of a pain because the arrays have to both have the same ending length and need to be context-aware of the index to be passed to `buildPayloadForExit`. I figured the only way to account for that without refactoring a swath of logic would be to consider these as a known error and let the normal logic resume.

nicholaspai: yeah it is, lets fix in a follow up

nicholaspai (approved):
LGTM to fix the immediate issue and follow up with a better fix

PR #1542, opened 2024-05-20
improve(reporter): Add special USDC.e entry

nicholaspai:
Adds some special logic to break out USDC.e balances from USDC. L1 USDC is counted with Native USDC balance.

This is required to be a special case because the balance report uses the L1 token symbol mapped to all L2 tokens as the key, whereas USDC.e and Native USDC on L2s both map to a single USDC token, so we need to invent a "USDC.e" virtual L1 token to keep the structure of the report intact.

A more scalable change would have to change the report structure fundamentally.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -28,6 +28,7 @@ import {
   toBNWei,
   winston,
   TOKEN_SYMBOLS_MAP,
+  compareAddressesSimple,
 } from "../utils";
 
 import { MonitorClients, updateMonitorClients } from "./MonitorClientHelper";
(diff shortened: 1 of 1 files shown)


src/monitor/Monitor.ts
```diff
@@ -216,7 +217,16 @@ export class Monitor {
 
   async reportRelayerBalances(): Promise<void> {
     const relayers = this.monitorConfig.monitoredRelayers;
-    const allL1Tokens = this.clients.hubPoolClient.getL1Tokens();
+    const allL1Tokens = [...this.clients.hubPoolClient.getL1Tokens()]; // @dev deep clone since we modify the
+    // array below and we don't want to modify the HubPoolClient's version
+    // @dev Handle special case for L1 USDC which is mapped to two L2 tokens on some chains, so we can more easily
+    // see L2 Bridged USDC balance versus Native USDC. Add USDC.e right after the USDC element.
+    const indexOfUsdc = allL1Tokens.findIndex(({ symbol }) => symbol === "USDC");
```
james-a-morris, line 224: Will this be impacted by the constants-v2 change that removes _USDC

nicholaspai: it will but that means the consants v2 change will have to include this file

pxrl (approved):
This seems sensible given the alternative involves a lot of rework.

PR #1546, opened 2024-05-22
refactor(relayer): Simplify deposit confirmation config

pxrl:
DEFAULT_MIN_DEPOSIT_CONFIRMATIONS isn't actually used anymore, so drop it, but migrate the chain finality defaults into
MIN_DEPOSIT_CONFIRMATIONS. Also, split out testnet chain IDs to a specific tier. This should make it a bit simpler to update and maintain in future.

While here, drop Goerli MDC configs since they're no longer needed. This change also includes the pending Mode deposit confirmation config from PR #1582.


pxrl:
Background: This was also needed to fix the relayer on testnet.

src/common/Constants.ts
```diff
@@ -32,79 +32,57 @@ export const FINALIZER_TOKENBRIDGE_LOOKBACK = 14 * 24 * 60 * 60;
...
+    8453: 120,
+    34443: 120,
+    42161: 0,
+    59144: 30,
+  },
   1000: {
     1: 32, // Justified block
     10: 60,
     137: 100, // Probabilistically safe level based on historic Polygon reorgs
-    288: 0,
```
nicholaspai, line 77: So what will happen to 280 now?

src/relayer/RelayerConfig.ts
```diff
@@ -268,12 +268,15 @@ export class RelayerConfig extends CommonConfig {
...
     // Append default thresholds as a safe upper-bound.
-    Object.keys(this.minDepositConfirmations).forEach((chainId) =>
-      this.minDepositConfirmations[chainId].push({
-        usdThreshold: bnUint256Max,
-        minConfirmations: Number.MAX_SAFE_INTEGER,
-      })
-    );
+    Object.keys(this.minDepositConfirmations).forEach((chainId) => {
+      const depositConfirmations = this.minDepositConfirmations[chainId];
+      if (depositConfirmations.at(-1).minConfirmations < Number.MAX_SAFE_INTEGER) {
```
nicholaspai, line 273: Never seen this at(-1) syntax before. Is it the same as the "last" element?

src/common/Constants.ts
pxrl: Nothing - this is zkSync Görli and this PR is removing definitions for the Görli chains.

src/relayer/RelayerConfig.ts
pxrl: Yep - it's a shorter version of `depositConfirmations[depositConfirmations.length - 1]`.

src/common/Constants.ts
nicholaspai: sorry i meant 288

pxrl: Hm, yeah it's also nuked on this. Do we think Boba will ever come back? It's easy enough to re-add it if it ever gets enabled again, but until that point it's not required.

src/relayer/RelayerConfig.ts
james-a-morris: As a note - `.at(-1)` does return `Value | undefined` so if we ever flip to check undefined then we will error on this

james-a-morris (commented):
One question https://github.com/across-protocol/relayer-v3/pull/1546#discussion_r1613926446

nicholaspai (approved):

PR #1547, opened 2024-05-22
improve(reporter): Improve USDC.e refunds and outstanding transfer reporting

nicholaspai:
- Refunds for chains where Native USDC is not supported should not have refund entries under USDC but under USDC.e
- Outstanding transfers should split out USDC.e vs USDC


src/monitor/Monitor.ts
```diff
@@ -632,22 +633,51 @@ export class Monitor {
     for (const refunds of nextBundleRefunds) {
       for (const relayer of this.monitorConfig.monitoredRelayers) {
         this.updateRelayerRefunds(refunds, relayerBalanceReport[relayer], relayer, BalanceType.NEXT);
-        this.updateCrossChainTransfers(relayer, relayerBalanceReport[relayer]);
```
nicholaspai, line 635: Not sure why this wasn't split out before but we shouldn't update xchain transfers conditional on there being a refund entry for a relayer

src/monitor/Monitor.ts
```diff
@@ -632,22 +633,51 @@ export class Monitor {
...
-          relayer,
-          chainId,
-          l1Token.address
-        );
+        // Handle special case for USDC which has multiple L2 tokens we might hold in inventory mapped to a single
+        // L1 token.
+        if (l1Token.symbol === "USDC" && chainId !== this.clients.hubPoolClient.chainId) {
+          const bridgedUsdcAddress =
+            TOKEN_SYMBOLS_MAP[chainId === CHAIN_IDs.BASE ? "USDbC" : "USDC.e"].addresses[chainId];
+          const nativeUsdcAddress = TOKEN_SYMBOLS_MAP["_USDC"].addresses[chainId];
```
pxrl, line 652: What if there is no native USDC for the remote chain? i.e. Linea.

pxrl: Ah, I see it's handled later.

pxrl (approved):
Looks OK :+1:

PR #1548, opened 2024-05-22
improve(wethdepositor): make OP-Stack generic

james-a-morris:
A quick pass at making the `AtomicWethDepositor` both ownable and able to modify it's OP stack changes.

Draws to this:
The major draw I see here is that if we want to add a new atomic depositor that isn't an OP stack change then we're going to have to redeploy with all the bridge events.


contracts/AtomicWethDepositor.sol
```diff
@@ -41,50 +47,186 @@ interface LineaL1MessageService {
...
+        ovmChainIdToBridge[chainId] = newBridge;
+        emit OVML1BridgeModified(chainId, address(newBridge), address(oldBridge));
+    }
+
+    /**
+     * @notice A helper function to disable the OvmL1Bridge contract for a given chainId.
+     * @dev This simply sets the OvmL1Bridge contract to the zero address.
+     * @param chainId The chainId of the OVM chain.
+     */
+    function disableOvmL1Bridge(uint256 chainId) public onlyOwner {
```
nicholaspai, line 162: I don't thnk you need to disable it if you assume that someone can just set `newBridge` above to address(0) to effectively disable

james-a-morris: Fair - this was just a convenience function

contracts/AtomicWethDepositor.sol
```diff
@@ -41,50 +47,185 @@ interface LineaL1MessageService {
...
+    mapping(uint256 => OvmL1Bridge) public ovmChainIdToBridge;
+
+    ///////////////////////////////
+    //          Events           //
+    ///////////////////////////////
+
+    /**
+     * @notice An event to track ETH deposits to ZkSync. This event is emitted when a deposit to ZkSync is initiated.
+     */
     event ZkSyncEthDepositInitiated(address indexed from, address indexed to, uint256 amount);
```
nicholaspai, line 92: what is this comment style you're using...? I've never commented individual events before

contracts/AtomicWethDepositor.sol
nicholaspai: i don't see a need for this, instead i would just add a @dev comment to `setOvmL1Bridge` that setting `newBridge = address(0)` disables a bridge

contracts/AtomicWethDepositor.sol
james-a-morris: I was applying natspec comments to all aspects of the code. NatSpec support for events was added via (https://github.com/ethereum/solidity/pull/9011) - I was being verbose

contracts/AtomicWethDepositor.sol
james-a-morris: Removed & commented here https://github.com/across-protocol/relayer-v3/pull/1548/commits/e844357ea38221f3a6f63e8c4d523d672b3f3589

contracts/AtomicWethDepositor.sol
```diff
@@ -41,56 +47,176 @@ interface LineaL1MessageService {
...
+    /**
+     * @notice A hardcoded address for WETH on Ethereum mainnet.
+     */
     Weth public immutable weth = Weth(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
-    OvmL1Bridge public immutable optimismL1Bridge = OvmL1Bridge(0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1);
-    OvmL1Bridge public immutable modeL1Bridge = OvmL1Bridge(0x735aDBbE72226BD52e818E7181953f42E3b0FF21);
-    OvmL1Bridge public immutable bobaL1Bridge = OvmL1Bridge(0xdc1664458d2f0B6090bEa60A8793A4E66c2F1c00);
-    OvmL1Bridge public immutable baseL1Bridge = OvmL1Bridge(0x3154Cf16ccdb4C6d922629664174b904d80F2C35);
+    /**
+     * @notice A hardcoded address for the Polygon L1 bridge on Ethereum mainnet.
```
mrice32, line 62: Note: Natspec isn't required for public variables, as most are self explanatory

contracts/AtomicWethDepositor.sol
```diff
@@ -41,56 +47,176 @@ interface LineaL1MessageService {
...
+     */
+    error InvalidOvmChainId();
+
+    ///////////////////////////////
+    //        Constructor        //
+    ///////////////////////////////
+    /**
+     * @notice Constructs the AtomicWethDepositor contract. Sets the owner to the deployer.
+     */
+    constructor() Ownable() {
```
mrice32, line 125: I think the constructor can be dropped since ownable takes no args.

contracts/AtomicWethDepositor.sol
```diff
@@ -41,56 +47,176 @@ interface LineaL1MessageService {
  * @notice Contract deployed on Ethereum helps relay bots atomically unwrap and bridge WETH over the canonical chain
  * bridges for Optimism, Base, Boba, ZkSync, Linea, and Polygon. Needed as these chains only support bridging of ETH,
  * not WETH.
+ * @dev This contract is ownable so that the owner can update the OvmL1Bridge contracts for new chains.
  */
+contract AtomicWethDepositor is Ownable {
```
mrice32, line 52: Should we make this multicaller as well? Could be nice to have the capability to batch calls.

mrice32 (commented):
Looks great! A few minor qs and comments

contracts/AtomicWethDepositor.sol
james-a-morris: Agreed but Etherscan does have parsing for natspec for public variables. Would it make sense to be explicit?

contracts/AtomicWethDepositor.sol
james-a-morris: Typically this has been one-off calls but we could add it - would this make sense?

contracts/AtomicWethDepositor.sol
james-a-morris: I was having trouble testing in Remix without it - will remove

contracts/AtomicWethDepositor.sol
nicholaspai: let's add it...why not

nicholaspai: I don't think multicaller allows us to multicall with msg.value though so not sure how helpful this would be

contracts/AtomicWethDepositor.sol
nicholaspai: i don't think we need it to be honest, it reduces the diff of this PR also

contracts/AtomicWethDepositor.sol
nicholaspai: verbosity is sometimes a sin, let's keep things simple and not comment events. The precedent I want to avoid is having to comment for every new event which makes contracts very wordy

nicholaspai (commented):
LGTM pending rolling back all the extra verbosity added with comments

james-a-morris:
@nicholaspai I removed all the less standard natspec comments. I kept only comments for function headers, dynamic variables, and to breakup the file.

contracts/AtomicWethDepositor.sol
james-a-morris: https://github.com/across-protocol/relayer/pull/1548#issuecomment-2192022773

contracts/AtomicWethDepositor.sol
mrice32: This contract takes WETH, not ETH, though, right? So that wouldn't apply here, right?

Note: multicaller might also be useful for setting config values, too.

mrice32 (approved):
LGTM! Just one comment

contracts/AtomicWethDepositor.sol
pxrl: Good call - I added it here: https://github.com/across-protocol/relayer/pull/1548/commits/0f33b738a40983f3a66d29f2a0559f10ee228be8

bmzig (approved):
This looks good to me.

contracts/AtomicWethDepositor.sol
pxrl: fwiw, the implementation for MultiCaller is sourced from the UMA repo and is confirmed to be not payable: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/common/implementation/MultiCaller.sol

pxrl (approved):

PR #1551, opened 2024-05-23
improve(relayer): Decouple per-chain fill status checks

pxrl:
As part of identifying unfilled deposits, the relayer currently performs one or more fillStatus multicall queries on each enabled chain. Multicalls are batched at 250 queries per chain, and queries are performed against each chain in parallel. This introduces an undesirable coupling between each of the chains, such that the relayer cannot progress until all queries have resolved. Additionally, there has been a recent uptick in deposit activity in response to an Across campaign. This resulted in upwards of 2k deposits being ingested by the relayer, and inevitably results in multiple multicall batches being produced per destination chain.

This PR reverts to internal SpokePoolClient deposit/fill matching as the initial method of filtering for unfilled deposits. Additionally, it defers the onchain fillStatus query until the point where an RPC delay won't delay fills on other chains. This is effectively trading CPU cycles for reduced need for RPC queries, which should make the relayer's performance less susceptible to demand spikes and RPC issues.


nicholaspai (approved):
This makes sense, defer the fillStatus check until we can parallelize it for each chainId

pxrl:
Anecdotally, I found one case from yesterday where it appears to have taken 13 seconds to resolve the fillStatus queries. This was at the time where there were ~2k deposits on zkSync alone within a 2 hour timeframe, and several hundred deposits on other chains, so it caused quite a few fillStatus multicall queries to be generated.

james-a-morris (approved):

src/monitor/Monitor.ts
```diff
@@ -171,15 +175,27 @@ export class Monitor {
...
-    const unfilledDeposits = await getUnfilledDeposits(this.clients.spokePoolClients, this.clients.hubPoolClient);
+    const { hubPoolClient, spokePoolClients } = this.clients;
+    const unfilledDeposits: Record<number, DepositWithBlock[]> = Object.fromEntries(
+      await mapAsync(
+        Object.values(spokePoolClients).map(({ chainId }) => chainId),
+        async (destinationChainId) => {
+          const deposits = getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient).map(
+            ({ deposit }) => deposit
+          );
+          const fillStatus = await fillStatusArray(spokePoolClients[destinationChainId].spokePool, deposits);
```
nicholaspai, line 186: is this not redundant? doesn't `getUnfilledDeposits` already filter these out?

pxrl: It will filter many out, but `fillStatusArray()` returns the absolute latest state, which is useful in the case where SpokePoolClient update takes a while - i.e. there can be quite a few blocks between `SpokePoolClient.latestBlockSearched` and the head of the chain. So `fillStatusArray()` gives the absolute latest snapshot.

In future this can be eliminated and we will simulate with tryMulticall() instead, which will report on the success of each fill in a single call.

nicholaspai: ok i guess this is ok because the monitor doesn't need to be fast

pxrl: Ah tbf I missed that this comment was against the monitor, but the motivation is the same - it eliminates quite a lot of stale reporting.

src/monitor/Monitor.ts
```diff
@@ -171,15 +175,27 @@ export class Monitor {
   }
 
   async reportUnfilledDeposits(): Promise<void> {
-    const unfilledDeposits = await getUnfilledDeposits(this.clients.spokePoolClients, this.clients.hubPoolClient);
+    const { hubPoolClient, spokePoolClients } = this.clients;
+    const unfilledDeposits: Record<number, DepositWithBlock[]> = Object.fromEntries(
+      await mapAsync(
+        Object.values(spokePoolClients).map(({ chainId }) => chainId),
```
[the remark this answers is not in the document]
james-a-morris: We could directly dereference `{chainId}` in the second map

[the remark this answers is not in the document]
pxrl: I totally missed that. Good call!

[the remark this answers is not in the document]
pxrl: https://github.com/across-protocol/relayer-v3/pull/1551/commits/7b23531f8804f950b3afb58c0e41ed45b27b1eeb

PR #1553, opened 2024-05-23
fix(finalizer): OpStack finalizer filter

nicholaspai:
Filter for events to finalize is incorrectly only counting events older than 14 days old. This PR changes it to look 
at any events older than 7 days old

### DIFF
--- a/src/finalizer/utils/opStack.ts
+++ b/src/finalizer/utils/opStack.ts
@@ -55,21 +55,18 @@ export async function opStackFinalizer(
   // - Don't submit proofs for finalizations older than 1 day
   // - Don't try to withdraw tokens that are not past the 7 day challenge period
   const redis = await getRedisCache(logger);
-  const [earliestBlockToFinalize, latestBlockToProve] = await Promise.all([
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):
+1 plus we save a network call

src/finalizer/utils/opStack.ts
```diff
@@ -55,21 +55,17 @@ export async function opStackFinalizer(
...
-    getBlockForTimestamp(chainId, getCurrentTime() - 7 * 60 * 60 * 24, undefined, redis),
-  ]);
+  const latestBlockToProve = await getBlockForTimestamp(chainId, getCurrentTime() - 7 * 60 * 60 * 24, undefined, redis);
   const { recentTokensBridgedEvents = [], olderTokensBridgedEvents = [] } = groupBy(
     spokePoolClient.getTokensBridged(),
     (e) => {
       if (e.blockNumber >= latestBlockToProve) {
         return "recentTokensBridgedEvents";
-      } else if (e.blockNumber <= earliestBlockToFinalize) {
+      } else if (e.blockNumber <= latestBlockToProve) {
```
mrice32, line 64: Should this just be an else?

mrice32 (approved):

PR #1555, opened 2024-05-23
improve(relayer): Deferred waiting on transaction submission

pxrl:
This change modifies the relayer to avoid getting completely stuck on a hanging/slow transaction submission. It does this by retaining promises for transaction execution, rather than waiting on each chain's promise to resolve. This will mitigate an often-seen issue where transaction simulation turtles hugely due to RPC issues and/or gas estimation complexities.

In the event that transaction execution any chain does stall, it will not be possible to make subsequent fills on that chain. That's OK for now, but will ultimately be addressed by reducing the worst-case blowouts on transaction simulation and execution (tryMulticall() will be a big help here).


src/relayer/Relayer.ts
```diff
@@ -441,15 +443,36 @@ export class Relayer {
...
+    if (isDefined(pendingTxnReceipts[chainId])) {
+      this.logger.info({
+        at: "Relayer::executeFills",
+        message: `${getNetworkName(chainId)} transaction queue has pending fills; skipping...`,
+      });
+      multiCallerClient.clearTransactionQueue(chainId);
+      return [];
+    }
+    pendingTxnReceipts[chainId] = multiCallerClient.executeTxnQueue(chainId, simulate);
+    const txnReceipts = await pendingTxnReceipts[chainId];
```
bmzig, line 461: Why would the relayer not be hung up here? As I understand it, the goal is to fire off a transaction and then continue the relayer loop, and if we end up looping back here, we just won't submit more fills. But doesn't the relayer just get hung up on awaiting the pendingTxnReceipts anyways, so this does not happen?

mrice32: I had the same question +1

pxrl: `executeFills()` will wait, but we don't wait on `executeFills(). See the change down at line 518.

change

bmzig: Makes sense.

bmzig (approved):

src/relayer/Relayer.ts
```diff
@@ -441,15 +443,36 @@ export class Relayer {
...
+      pendingTxnReceipts,
+      clients: { multiCallerClient },
+    } = this;
+
+    if (isDefined(pendingTxnReceipts[chainId])) {
+      this.logger.info({
+        at: "Relayer::executeFills",
+        message: `${getNetworkName(chainId)} transaction queue has pending fills; skipping...`,
+      });
+      multiCallerClient.clearTransactionQueue(chainId);
```
nicholaspai, line 457: Can't this become an issue if `executeFills` is ever called by a different function than `checkForUnfilledDeopsitsAndFill`? Maybe for safety mark this private, or just don't clear the queue?

pxrl: I think we need to flush the queue because otherwise the relayer will re-enqueue additional transactions for the same unfilled deposits on a subsequent loop. I've marked it protected instead: https://github.com/across-protocol/relayer/pull/1555/commits/0daf4b85e3882c32a6c2878637d440598243547c

nicholaspai (approved):

PR #1558, opened 2024-05-24
chore(finalizer): remove lookback constraints

james-a-morris:
I propose we can bound the lookback on `ZKSync` and `Arbitrum` to be withdrawals with the latest block being each respective lockup window and the earliest block being decided by `FINALIZER_MAX_TOKENBRIDGE_LOOKBACK`


nicholaspai (approved):
I'm in agreement. Lgtm

PR #1560, opened 2024-05-24
improve(Monitor): Instantiate clients correctly

pxrl:
Spotted by James.

### DIFF
--- a/src/monitor/index.ts
+++ b/src/monitor/index.ts
@@ -8,14 +8,11 @@ let logger: winston.Logger;
 export async function runMonitor(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
   logger = _logger;
   const config = new MonitorConfig(process.env);
-  let clients;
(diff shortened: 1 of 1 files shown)


src/monitor/index.ts
```diff
@@ -8,14 +8,11 @@ let logger: winston.Logger;
 export async function runMonitor(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
   logger = _logger;
   const config = new MonitorConfig(process.env);
-  let clients;
+  const clients = await constructMonitorClients(config, logger, baseSigner);
+  const acrossMonitor = new Monitor(logger, config, clients);
 
   try {
```
pxrl, line 14: This top-level try statement is still nasty fwiw :\

james-a-morris (approved):
If we wanted to keep the diff small we could just change line 11 to include the type MonitorClients

PR #1563, opened 2024-05-26
improve(Monitor): reformat reportRelayerBalances log

bmzig:
This change makes the logs from the monitor more readable by formatting the reported balances as a table.


nicholaspai:
Can you give an example output? An image works

bmzig:
Before: 
image
After:
image

nicholaspai:
> Before: 
> 
> image
> 
> After:
> 
> image
> 
> 

Hm can this be rendered better? The columns don't line up vertically still

james-a-morris (commented):
Will this report properly in slack?

bmzig:
> > Before:
> > image
> > After:
> > image
> 
> Hm can this be rendered better? The columns don't line up vertically still

So as you know, the slack markdown support is pretty limited. I did not want to *overdo* it and check the length of each balance report to ensure that the columns are aligned, since that just seemed like it was a bit overkill, but that can be done if you think it's not overengineering this. Fwiw, I checked and slack can't render markdown tables or latex matrices.

bmzig:
> Will this report properly in slack?

The screenshots taken are from slack (I don't know if that is what you meant by the question). I did not change any of the actual logic which calculates the balances -- I only made changes to how the output is formatted.

bmzig:
Here is the updated look: 
image

james-a-morris (commented):
Can you provide the updated slack readout?

james-a-morris (commented):
How would this look for tokens where we have balances on the order of millions? Ie WETH

bmzig:
> How would this look for tokens where we have balances on the order of millions? Ie WETH

image
image

It only really matters if all columns are maxed out with >7 figures when we might see some wrapping.

nicholaspai (approved):
Let's try it!

PR #1565, opened 2024-05-28
feat(tsconfig): disable implicit any

james-a-morris:
There have been several errors encountered in the relayer that are the result of tsc allowing `any` types. To ensure we're maintaining/reasoning about type safety throughout the codebase, I propose we enforce `noImplicitAny: true`.


src/common/Constants.ts
```diff
@@ -122,7 +122,7 @@ export const REDIS_URL_DEFAULT = "redis://localhost:6379";
 // Note: a 0 value here leads to an infinite lookback, which would be useful and reduce RPC requests
 // if the RPC provider allows it. This is why the user should override these lookbacks if they are not using
 // Quicknode for example.
-export const CHAIN_MAX_BLOCK_LOOKBACK = {
+export const CHAIN_MAX_BLOCK_LOOKBACK: { [chainId: number]: number } = {
```
nicholaspai, line 125: should this just be Record<number, number>?

james-a-morris: Yep they're the same definition - I figured since this is a constant that we should describe what the key is

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -448,21 +448,26 @@ export function prettyPrintLeaves(
...
       return result;
-    }, {});
+    }, {} as Record<string, unknown>);
     logger.debug({
       at: "Dataworker#propose",
       message: `${logType} leaf #${index}`,
       leaf: prettyLeaf,
-      proof: tree.getHexProof(leaf),
+      proof: tree.getHexProof(
+        leaf as unknown as sdkInterfaces.SlowFillLeaf &
```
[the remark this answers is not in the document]
james-a-morris: Yep - there was some shaping to get tsc working at the end

src/relayer/index.ts
```diff
@@ -43,7 +43,7 @@ function startWorkers(config: RelayerConfig): { [chainId: number]: ChildProcess
     chainIds.map((chainId: number) => {
       // Identify the lowest configured deposit confirmation threshold for use as indexer finality.
       const finality = config.minDepositConfirmations[chainId].at(0)?.minConfirmations ?? 1024;
-      const blockRange = config.maxRelayerLookBack[chainId] ?? 5_000; // 5k is a safe default.
+      const blockRange = config.maxRelayerLookBack ?? 5_000; // 5k is a safe default.
```
nicholaspai, line 46: this is change in logic, right?

nicholaspai (commented):
first pass comments

src/relayer/index.ts
james-a-morris: Correct - this is one of the bugs I found. `maxRelayerLookback` is a number so `maxRelayerLookBack[...]` should be undefined - or honestly it should fail

PR #1566, opened 2024-05-29
improve(Finalizer): Handle Linea SDK errors

nicholaspai:
The Linea SDK MessageServiceContract constructs its own Provider without our retry logic so we try-catch it in order to catch any one-off failures.

An alternative implementation would be to replace the contract object returned by the SDK and create our own but that look a bit more involved, so for now I implement a try-catch to handle these failures


james-a-morris (approved):
This is a good workaround. I think it's worthwhile to revisit this to move away from the Linea SDK

PR #1568, opened 2024-05-29
fix(relayer): Propagate quorum to external indexer

pxrl:
The external indexers have been defaulting to a quorum of 1 because quroum was not explicitly specified.

### DIFF
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -176,9 +176,20 @@ export class EventManager {
     // This is configurable and will almost always be less than chain finality guarantees.
     const finalised = blockNumber - this.finality;
 
-    // Collect the events that met quorum, stripping out the provider information.
+    // Collect the events that met quorum, stripping out the provider information; drop any that didn't.
+    // This can be brittle when finality is low (i.e. 1). @todo: Support querying back over multiple blocks
+    // to account for RPC notification delays.
     const events = (this.events[finalised] ?? [])
-      .filter((event) => this.getEventQuorum(event) >= this.quorum)
+      .filter((event) => {
+        const eventQuorum = this.getEventQuorum(event);
+        if (this.quorum > eventQuorum) {
+          this.logger.debug({
+            at: "EventManager::tick",
+            message: `Dropped ${this.chain} ${event.event} event due to insufficient quorum.`,
+          });
+        }
+        return eventQuorum >= this.quorum;
+      })
       .map(({ providers, ...event }) => event);
 
     // Flush the events that were just submitted.


nicholaspai (approved):

src/relayer/index.ts
```diff
@@ -44,8 +45,9 @@ function startWorkers(config: RelayerConfig): { [chainId: number]: ChildProcess
       // Identify the lowest configured deposit confirmation threshold for use as indexer finality.
       const finality = config.minDepositConfirmations[chainId].at(0)?.minConfirmations ?? 1024;
       const blockRange = config.maxRelayerLookBack[chainId] ?? 5_000; // 5k is a safe default.
+      const quorum = getChainQuorum(chainId);
 
-      const opts = { ...sampleOpts, finality, blockRange };
+      const opts = { ...sampleOpts, finality, blockRange, quorum };
```
mrice32, line 50: Is this opt just used for the ws provider? And the inherited env var is used for the http provider inside of the indexer?

mrice32 (approved):
LGTM! Just one q

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants-v2": "1.0.26",
     "@across-protocol/contracts-v2": "2.5.9",
-    "@across-protocol/sdk-v2": "0.24.2",
+    "@across-protocol/sdk-v2": "0.24.4",
```
pxrl, line 15: Necessary because the SDK lacked a definition for Mode and that was causing the indexer's to be unable to resolve the correct Hub Chain ID.

mrice32 (approved):

src/utils/ProviderUtils.ts
```diff
@@ -579,6 +579,15 @@ export function getCachedProvider(chainId: number, redisEnabled = true): RetryPr
   return providerCache[getProviderCacheKey(chainId, redisEnabled)];
 }
 
+/**
+ * Return the env-defined quorum configured for `chainId`, or 1 if no quorum has been defined.
+ * @param chainId Chain ID to query for quorum.
+ * @returns Applicable quorum.
+ */
+export function getChainQuorum(chainId: number): number {
+  return Number(process.env[`NODE_QUORUM_${chainId}`] || process.env.NODE_QUORUM || "1");
```
james-a-morris, line 588: Is there a  case where this wouldn't be defined?

pxrl: You'd normally expect NODE_QUORUM to be defined, but it's not required because we've always had 1 as a fallback - i.e. this is the same logic we've always had.

PR #1569, opened 2024-05-29
improve(relayer): Ignore deposits for disabled destination chains

pxrl:
Failing to do this can produce a lot of log spam that is ultimately unnecessary.


src/relayer/index.ts
pxrl, line 1: This is an optimistic driveby change.

nicholaspai (approved):

PR #1570, opened 2024-05-29
fix: retry linea in the case of intermittent RPC errors in the SDK

mrice32:


nicholaspai (approved):

nicholaspai:
Can you add the retry wrapper to line 86 in linea/common "makeGetMessagesWithStatusByTxHash()"?

PR #1571, opened 2024-05-29
[WIP] Support looping mode in bridge adapters

pxrl:
This is a WIP PR to make the necessary updates to the current bridge adapters such that the looping relayer can use inventory management.

The bridge adapters have historically just piggybacked on the last SpokePoolClient update's search config. That's not feasible in the world of 1000 chains because it imposes undesirable coupling across the various supported chains, so instead update the bridge BaseAdapter to live its own life.

Note: At present this is completely untested.


pxrl:
Not relevant anymore - dropping this PR.

PR #1574, opened 2024-05-30
fix(RelayerConfig): Read symbol not address

nicholaspai:
Bug found because `unwrapWethThreshold` in rebalancer seemed to be getting ignored, turned out it was never getting parsed because this line was always false


pxrl (approved):
Good catch

PR #1577, opened 2024-05-30
feat(utils): Fix retryAsync with delay function

nicholaspai:
Adds unit tests to verify behavior

### DIFF
--- a/src/utils/RetryUtils.ts
+++ b/src/utils/RetryUtils.ts
@@ -1,13 +1,17 @@
 import { delay } from "./TimeUtils";
 
-export async function retryAsync<T>(fn: () => Promise<T>, numRetries: number, delayS: number): Promise<T> {
-  try {
-    return await fn();
-  } catch (err) {
-    if (numRetries <= 0) {
-      throw err;
-    }
-    await delay(delayS);
-    return retryAsync(fn, numRetries - 1, delayS);
+export function retryAsync<T, U extends unknown[]>(
+  fn: (...args: U) => Promise<T>,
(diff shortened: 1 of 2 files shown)


src/utils/RetryUtils.ts
```diff
@@ -1,17 +1,13 @@
...
-): Promise<T> {
-  let ret = fn(...args);
-  for (let i = numRetries; i < numRetries; i++) {
-    ret = ret.catch(async () => {
-      await delay(delayS);
-      return fn(...args);
-    });
+export async function retryAsync<T>(fn: () => Promise<T>, numRetries: number, delayS: number): Promise<T> {
+  try {
+    return await fn();
```
mrice32, line 5: Why drop the args?

test/RetryUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+let ERROR_COUNTER = 0;
+
+// This function will throw based on the value of an external counter variable, this way
+// we can use it to test a function that fails intermittently. If `errorIndex` is >
+// to ERROR_COUNTER, then `expectedErrorMsg` will be thrown.
+function incrementCounterThrowError(errorIndex: number, returnValue = expectedReturnValue): Promise<number> {
+  const oldCounter = ERROR_COUNTER;
+  ERROR_COUNTER++;
+  if (errorIndex > oldCounter) {
+    throw new Error(expectedErrorMsg);
```
mrice32, line 18: This isn't actually an async error like what we would get from an RPC

mrice32 (approved):

PR #1579, opened 2024-05-31
improve(test): Better depositV3() in test

pxrl:
This permits depositV3() to be called in parallel, which does actually happen in current tests. Also, update the unfilled deposits test to read unfilled deposits correctly.


test/utils/utils.ts
```diff
@@ -289,55 +289,41 @@ export async function depositV3(
...
+        exclusivityDeadline,
+        message
+      ),
   ]);
+  const txnReceipt = await txnResponse.wait();
 
-  const lastEvent = events.at(-1);
-  let args = lastEvent?.args;
-  assert.exists(args);
-  args = args!; // tsc coersion
```
pxrl, line 317: This previously relied on depositV3() being called in isolation, but in practice the tests called it in parallel.

nicholaspai (approved):

PR #1580, opened 2024-05-31
improve(relayer): Persist fillStatus across loops

pxrl:
The looping relayer is forced to drink from the fire hose when determining the set of unfilled deposits. In times of high demand this can lead to a lot of unnecessary looping and hashing of relayData to filter out deposits that have already been filled. This change adds tracking of fill status across loops in order to permit already-filled deposits to be filtered out as early as possible. There should be limited impact on the initial loop (or when POLLING_MODE=0), but it will have an effect on subsequent loops.

### DIFF
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -1,6 +1,6 @@
 import { utils as sdkUtils } from "@across-protocol/sdk-v2";
 import { HubPoolClient } from "../clients";
-import { Fill, SpokePoolClientsByChain, V3DepositWithBlock } from "../interfaces";
+import { Fill, FillStatus, SpokePoolClientsByChain, V3DepositWithBlock } from "../interfaces";
 import { bnZero } from "../utils";
 import { getBlockRangeForChain } from "../dataworker/DataworkerUtils";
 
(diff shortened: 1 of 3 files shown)


src/utils/FillUtils.ts
```diff
@@ -54,15 +54,20 @@ export type RelayerUnfilledDeposit = {
...
 
   // Iterate over each chainId and check for unfilled deposits.
   const deposits = Object.values(spokePoolClients)
     .filter(({ chainId, isUpdated }) => isUpdated && chainId !== destinationChainId)
     .map((spokePoolClient) => spokePoolClient.getDepositsForDestinationChain(destinationChainId))
-    .flat();
+    .flat()
+    .filter((deposit) => {
+      const depositHash = spokePoolClients[deposit.originChainId].getDepositHash(deposit);
+      return (fillStatus[depositHash] ?? FillStatus.Unfilled) !== FillStatus.Filled;
```
pxrl, line 69: To be explicit, filtering here is effective at eliminating subsequent calls Fill <-> Deposit validation for deposits that have already been confirmed to be filled. It was previously seen in the dataworker that this could be computationally expensive, so it seems like a simple and inexpensive way of bypassing a lot of that work (especially when there are a high number of deposits).

pxrl: > I do wonder if we should update the `Deposit` type to include a "lazily-evaluated and subsequently stored" relayData hash - i.e. a getter that can check for its existence in a global table, and compute it on the fly if it doesn't exist. In that way the relayData hash could be computed on demand, but would be retained for later use for the life of the object.

I think I'll propose this for the SDK anyway - it should give a performance benefit for both dataworker and relayer personalities and it won't be difficult or computationally expensive, so it makes sense to do it.

pxrl:
https://github.com/across-protocol/relayer-v3/pull/1580#discussion_r1622310203

PR #1581, opened 2024-05-31
lint: add no-misused promises to linter

bmzig:
The `no-misused-promises` rule enforces unresolved promises to be resolved before used (e.g. in a boolean conditional or in an object). Adding this in the linter will ideally prevent bugs which occur due to handling promises instead of the resolved value.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,7 +36,6 @@ module.exports = {
     "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true }],
     "chai-expect/missing-assertion": 2,
     "no-duplicate-imports": "error",
-    // "require-await": "error",
     "@typescript-eslint/no-floating-promises": ["error"],
     "@typescript-eslint/no-misused-promises": ["error", { checksVoidReturn: false }],
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,8 +36,9 @@ module.exports = {
     "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true }],
     "chai-expect/missing-assertion": 2,
     "no-duplicate-imports": "error",
-    //    "require-await": "error",
+    // "require-await": "error",
     "@typescript-eslint/no-floating-promises": ["error"],
+    "@typescript-eslint/no-misused-promises": ["error", { checksVoidReturn: false }],
   },
   settings: {
     node: {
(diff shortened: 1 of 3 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -2221,7 +2221,7 @@ export class Dataworker {
     // FIXME: Temporary fix to disable root cache rebalancing and to keep the
     //        executor running for tonight (2023-08-28) until we can fix the
     //        root cache rebalancing bug.
-    if (!this.rootCache[key] || process.env.DATAWORKER_DISABLE_REBALANCE_ROOT_CACHE === "true") {
+    if (!(await this.rootCache[key]) || process.env.DATAWORKER_DISABLE_REBALANCE_ROOT_CACHE === "true") {
```
james-a-morris, line 2224: This is meant to be unawaited - we're testing for the existence of the promise

james-a-morris (changes requested):
I think some of these would impact dataworker timings

src/dataworker/Dataworker.ts
bmzig: Is this true for the other changes? This one is fixed here: https://github.com/across-protocol/relayer-v3/pull/1581/commits/f67cd4156a12f8943643d21025bfd481bd9eb022

bmzig: Also fixed here: https://github.com/across-protocol/relayer/pull/1581/commits/b6826bfaa25c5027460cf1dd3d65a2ecbb4146d1

nicholaspai (approved):

nicholaspai:
> I think some of these would impact dataworker timings

Agreed. Sometimes we purposefully don't await promises. We cache the unresolved promise so the promises can execute in the background

.eslintrc.js
```diff
@@ -36,8 +36,9 @@ module.exports = {
     "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true }],
     "chai-expect/missing-assertion": 2,
     "no-duplicate-imports": "error",
-    //    "require-await": "error",
+    // "require-await": "error",
```
james-a-morris, line 39: If this is wrong we can remove it

james-a-morris (approved):

.eslintrc.js
bmzig: https://github.com/across-protocol/relayer/pull/1581/commits/40a8061d238182a95e35a0cfb09af8486ee28ac5

PR #1585, opened 2024-06-03
improve(Dataworker): Exit early in pool rebalance leaf execution if leaves are already executed

nicholaspai:
Simple QoL improvement but avoids false errors due to executor collisions in serverless mode


src/dataworker/Dataworker.ts
```diff
@@ -1534,7 +1545,6 @@ export class Dataworker {
               root: tree.getHexRoot(),
               leafId: leaf.leafId,
               rebalanceChain: leaf.chainId,
-              chainId: hubPoolChainId,
```
james-a-morris, line 1537: Is this removed because it's always mainnet?

james-a-morris (approved):

PR #1586, opened 2024-06-03
feat(Dataworker): Split l2 and l1 executors

nicholaspai:
We can run the pool rebalance leaf executor separately from the refund leaf executor to speed up both bots. It takes ~3 minutes to reconstruct a bundle, so executing the pool rebalance leaf + 2 refund leaf bundles into the past leads to 9 mins of run time just to execute pool rebalance leaves.

### DIFF
--- a/src/dataworker/DataworkerConfig.ts
+++ b/src/dataworker/DataworkerConfig.ts
@@ -10,7 +10,8 @@ export class DataworkerConfig extends CommonConfig {
   // These variables can be toggled to choose whether the bot will go through the dataworker logic.
   readonly disputerEnabled: boolean;
   readonly proposerEnabled: boolean;
-  readonly executorEnabled: boolean;
+  readonly l2ExecutorEnabled: boolean;
+  readonly l1ExecutorEnabled: boolean;
   readonly finalizerEnabled: boolean;
 
   // This variable can be toggled to bypass the proposer logic and always attempt to propose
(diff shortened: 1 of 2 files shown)


src/dataworker/DataworkerConfig.ts
```diff
@@ -85,9 +87,10 @@ export class DataworkerConfig extends CommonConfig {
       : toBNWei("500000");
     this.disputerEnabled = DISPUTER_ENABLED === "true";
     this.proposerEnabled = PROPOSER_ENABLED === "true";
-    this.executorEnabled = EXECUTOR_ENABLED === "true";
-    if (this.executorEnabled) {
-      assert(this.spokeRootsLookbackCount > 0, "must set spokeRootsLookbackCount > 0 if executor enabled");
+    this.l2ExecutorEnabled = L2_EXECUTOR_ENABLED === "true";
+    this.l1ExecutorEnabled = L1_EXECUTOR_ENABLED === "true";
```
james-a-morris, line 91: OOC should this be case agnostic?

src/dataworker/index.ts
```diff
@@ -128,29 +128,33 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
         logger[startupLogLevel(config)]({ at: "Dataworker#index", message: "Proposer disabled" });
       }
 
-      if (config.executorEnabled) {
+      if (config.l2ExecutorEnabled || config.l1ExecutorEnabled) {
```
james-a-morris, line 131: Could we invert this into a guard

src/dataworker/DataworkerConfig.ts
nicholaspai: lets make this change in a separate PR

james-a-morris (approved):

PR #1587, opened 2024-06-03
improve(Dataworker): Speed up BundleDataClient.loadDataFromScratch

nicholaspai:
# Motivation

Loading bundle data is now consistently taking 3-5 minutes given increased volume, likely stemming from increase in total bridge events per bundle

I've narrowed down, using timers, the slowdown to the `loadDataFromScratch` and the `SpokePoolClient.update()` calls which together account for most of the vast majority of the total run time for any dataworker function

# Implementation

- Replacing `utils.getRelayHashFromEvent` implementation takes current bundle construction with ~5000 deposits and fills runtime from 160s to 5s within the `loadDataFromScratch` function.
- Fetching bundle data from Arweave instead of from scratch in the Executor saves executor time
- Added timers for loading bundle data to help track regressions

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -302,7 +302,7 @@ export class BundleDataClient {
           const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
           const hasMatchingDeposit =
             matchingDeposit !== undefined &&
-            utils.getRelayHashFromEvent(fill) === utils.getRelayHashFromEvent(matchingDeposit);
+            this.getRelayHashFromEvent(fill) === this.getRelayHashFromEvent(matchingDeposit);
           return hasMatchingDeposit;
         })
         .forEach((fill) => {
@@ -607,7 +607,7 @@ export class BundleDataClient {
     blockRangesForChains: number[][],
     spokePoolClients: SpokePoolClientsByChain
   ): Promise<LoadDataReturnValue> {
-    const start = performance.now();
+    let start = performance.now();
     const key = JSON.stringify(blockRangesForChains);
 
     if (!this.clients.configStoreClient.isUpdated) {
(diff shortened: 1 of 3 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -1181,6 +1200,14 @@ export class BundleDataClient {
     };
   }
 
+  // Internal function to uniquely identify a bridge event. This is preferred over `SDK.getRelayDataHash` which returns
+  // keccak256 hash of the relay data, which can be used as input into the on-chain `fillStatuses()` function in the
+  // spoke pool contract. However, this internal function is used to uniquely identify a bridging event
+  // for speed since its easier to build a string from the event data than to hash it.
+  private getRelayHashFromEvent(event: V3DepositWithBlock | V3FillWithBlock | SlowFillRequestWithBlock): string {
+    return `${event.depositor}-${event.recipient}-${event.exclusiveRelayer}-${event.inputToken}-${event.outputToken}-${event.inputAmount}-${event.outputAmount}-${event.originChainId}-${event.depositId}-${event.fillDeadline}-${event.exclusivityDeadline}-${event.message}-${event.destinationChainId}`;
+  }
```
pxrl, line 1209: Nice find here. It definitely maps onto observations that an increase in deposit events leads to a notable slowdown in the bots. I'm really curious to know whether keccak is basically unusable on large data arrays, or whether the issue might be because we tend to treat hashes as disposable data and therefore periodically recompute them. I think the latter could still be true but would like to confirm it.

fwiw I started work last week on retaining the keccak hash so that we only need to compute it once, and would retain it thereafter. It _might_ be an option also worth considering, because there are still use cases for computing the actual relay hash so we can query its status via an RPC call: https://github.com/across-protocol/sdk/pull/657

pxrl (approved):
Given the speedup it makes sense to get this in ASAP.

src/clients/BundleDataClient.ts
nicholaspai: Yep I think we need it in the relayer when we do want to call fillStatuses on chain

PR #1588, opened 2024-06-04
improve(relayer): Persist fillStatus across loops

pxrl:
The looping relayer is forced to drink from the fire hose when determining the set of unfilled deposits. In times of high demand this can lead to a lot of unnecessary looping to filter out deposits that have already been filled. This change adds tracking of fill status across loops in order to permit already-filled deposits to be filtered out as early as possible.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -35,6 +35,7 @@ type RepaymentChainProfitability = {
 
 export class Relayer {
   public readonly relayerAddress: string;
+  public readonly fillStatus: { [depositHash: string]: number } = {};
 
   constructor(
     relayerAddress: string,
@@ -225,7 +226,7 @@ export class Relayer {
         .filter(({ chainId }) => relayerDestinationChains?.includes(chainId) ?? true)
         .map(({ chainId: destinationChainId }) => [
           destinationChainId,
-          getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient).filter((deposit) =>
+          getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient, this.fillStatus).filter((deposit) =>
             this.filterDeposit(deposit)
           ),
         ])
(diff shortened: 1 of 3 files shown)


src/utils/FillUtils.ts
```diff
@@ -54,15 +54,20 @@ export type RelayerUnfilledDeposit = {
...
+  fillStatus: { [deposit: string]: number } = {}
 ): RelayerUnfilledDeposit[] {
   const destinationClient = spokePoolClients[destinationChainId];
 
   // Iterate over each chainId and check for unfilled deposits.
   const deposits = Object.values(spokePoolClients)
     .filter(({ chainId, isUpdated }) => isUpdated && chainId !== destinationChainId)
     .map((spokePoolClient) => spokePoolClient.getDepositsForDestinationChain(destinationChainId))
-    .flat();
+    .flat()
```
james-a-morris, line 66: We can use flat map above to reduce the 2N call

src/relayer/Relayer.ts
```diff
@@ -503,7 +504,12 @@ export class Relayer {
 
       const unfilledDeposits = deposits
         .map((deposit, idx) => ({ ...deposit, fillStatus: fillStatus[idx] }))
-        .filter(({ fillStatus }) => fillStatus !== FillStatus.Filled);
+        .filter(({ fillStatus, ...deposit }) => {
```
james-a-morris, line 507: This isn't a flat map - and we're iterating over an unbounded number of unfilled deposits - would it make sense to remove the map up top and replace with a filter that destructures the mapping

src/relayer/Relayer.ts
```diff
@@ -36,6 +36,7 @@ type RepaymentChainProfitability = {
 
 export class Relayer {
   public readonly relayerAddress: string;
+  public readonly fillStatus: { [depositHash: string]: number } = {};
```
james-a-morris, line 39: Should we clear this cache periodically, or trust that the fast relayer's short lifespan makes the memory usage negligible?

james-a-morris (commented):
Looks good - a few comments

src/relayer/Relayer.ts
pxrl: I _think_ we need to retain the map at the top level because otherwise the subsequent array ends up shorter and its index can't be used to read the fillStatus array. Does that make sense?

src/utils/FillUtils.ts
pxrl: Nice - https://github.com/across-protocol/relayer/pull/1588/commits/75f2470e19f3c9bfb3e8e0ed1c1fc3a4fd8826ec

src/relayer/Relayer.ts
pxrl: I think ultimately it would make sense but the SpokePoolClient instances themselves just keep accuring crud, so the relayer must be restarted every x minutes anyway. We'll need to address that as part of a subsequent refactor.

james-a-morris (approved):
Resolved offline 🌴

PR #1590, opened 2024-06-04
improve(relayer): Wait on unresolved fill promises on exit

pxrl:
The recent change to skip waiting on submission promises relies on submission occurring within some fixed time window, otherwise the relayer just exits.


src/relayer/index.ts
```diff
@@ -135,6 +136,8 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       }
     } while (!stop);
   } finally {
+    await Promise.all(Object.values(txnReceipts));
```
nicholaspai, line 139: should this not be in the try rather than the catch?

src/relayer/index.ts
```diff
@@ -135,6 +136,8 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
+
```
james-a-morris, line 140:
```suggestion
    await Promise.allSettled(Object.values(txnReceipts));
```

pxrl: Done: https://github.com/across-protocol/relayer/pull/1590/commits/c78757ea8b2b7d95c646746af6b9c95333115339

nicholaspai: whats the diff

nicholaspai:
https://github.com/across-protocol/relayer/pull/1590#discussion_r1626280993

https://github.com/across-protocol/relayer/pull/1590#discussion_r1626272439

src/relayer/index.ts
pxrl: Yeah, it could be in the try, subject to (!loop). In practice I believe it should be the same though. Any strong preference ?

nicholaspai (approved):

nicholaspai (approved):

james-a-morris (approved):

pxrl:
This needs an update after the underlying commit was backed out so that a more important fix could go in for the executor.

nicholaspai (approved):

src/relayer/index.ts
pxrl: Have relocated this to inside the try clause, but outside of the loop, so it should be an OK compromise now.

PR #1591, opened 2024-06-05
fix(executor): Skip arweave preload

pxrl:
There can be occasional disagreements when reading from Arweave data.


nicholaspai:
Recommend we close in favor of the fix https://github.com/across-protocol/relayer/pull/1593

PR #1593, opened 2024-06-05
fix(Dataworker): Shallow copy refunds object when constructing RelayerRefundRoot

nicholaspai:
## Motivation
We noticed that the Dataworker failed to reconstruct a bundle using Arweave bundle data but could reconstruct the same bundle using fresh data from scratch. This indicated corrupted data posted to Arweave.

When investigating Arweave, we noticed that all the bundleData looked correct except for `bundleFillsV3.refunds` which contained refunds for expired deposits. This is odd and unexpected since refunds should only be filled with fast fill refunds, not expired deposit refunds. This led me to realize that [this function](https://github.com/across-protocol/relayer/blob/13c5f17db14f11789f3fb6322f52183293fdb3cd/src/dataworker/DataworkerUtils.ts#L233) touches both `bundleFillsV3.refunds` and `expiredDeposits`, which led me to then realize that we are mutating the former when examining the latter.

Specifically:
- currently we set `combinedRefunds[repaymentChainId][l2TokenAddress] = refunds` on line 251
- but that's setting to the reference
- so later when we modify `combinedRefunds` on line 267, we set inadvertently `refunds` to contain the expired deposit refunds
- and recall, `refunds` is part of `bundleFillsV3` which is what we post to Arweave
- this explains why the `refunds` object posted to Arweave could contain expired deposit data

## Bug Description
In `DataworkerUtils`, when we construct the `combinedRefunds` object using the `refunds` from `bundleFillsV3` (i.e. the fast fill refunds in a bundle) and the `expiredDepositRefunds` (i.e. refunds for expired deposits), we accidentally modify the former as a reference variable.

This leads to upstream issues in `proposeRootBundle` which returns the bundle data to post to Arweave [here](https://github.com/across-protocol/relayer/blob/13c5f17db14f11789f3fb6322f52183293fdb3cd/src/dataworker/Dataworker.ts#L488). However, the very subtle bug is that the values we save to post to arweave here are the same references we accidentally mutate above.

This bug would only have been possible if we were trying to execute l2 leaves for a bundle containing expired deposit refunds, which only could have happened after #1587 which loads data from Arweave in the Executor, and expired deposits were only included in a bundle recently

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -481,16 +481,13 @@ export class Dataworker {
     // Arweave for the bundle. We will be doing this at a
     // later point so that we can confirm that this data is
     // worth storing.
-    // @dev Deep clone all objects to avoid any potential
-    // reference issues when we pass the loadData return values to
-    // the following root-building functions.
     const dataToPersistToDALayer = {
-      bundleBlockRanges: _.cloneDeep(blockRangesForProposal),
-      bundleDepositsV3: _.cloneDeep(bundleDepositsV3),
-      expiredDepositsToRefundV3: _.cloneDeep(expiredDepositsToRefundV3),
-      bundleFillsV3: _.cloneDeep(bundleFillsV3),
-      unexecutableSlowFills: _.cloneDeep(unexecutableSlowFills),
-      bundleSlowFillsV3: _.cloneDeep(bundleSlowFillsV3),
+      bundleBlockRanges: blockRangesForProposal,
+      bundleDepositsV3,
+      expiredDepositsToRefundV3,
+      bundleFillsV3,
+      unexecutableSlowFills,
+      bundleSlowFillsV3,
     };
     const [, mainnetBundleEndBlock] = blockRangesForProposal[0];
 
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -481,13 +481,16 @@ export class Dataworker {
     // Arweave for the bundle. We will be doing this at a
     // later point so that we can confirm that this data is
     // worth storing.
+    // @dev Deep clone all objects to avoid any potential
+    // reference issues when we pass the loadData return values to
+    // the following root-building functions.
     const dataToPersistToDALayer = {
-      bundleBlockRanges: blockRangesForProposal,
-      bundleDepositsV3,
-      expiredDepositsToRefundV3,
-      bundleFillsV3,
-      unexecutableSlowFills,
-      bundleSlowFillsV3,
+      bundleBlockRanges: [...blockRangesForProposal],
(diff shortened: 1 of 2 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -481,13 +481,16 @@ export class Dataworker {
     // Arweave for the bundle. We will be doing this at a
     // later point so that we can confirm that this data is
     // worth storing.
+    // @dev Deep clone all objects to avoid any potential
```
pxrl, line 484: This is moreso a shallow clone since it'll create new objects from the ones supplied, but their underlying elements will still be the same. So popping, pushing, deleting (etc) will happen in isolation, but modifying the underlying deposit, fill (etc) objects will not.

nicholaspai: how do we deep clone?

src/dataworker/DataworkerUtils.ts
```diff
@@ -247,7 +248,7 @@ export function getRefundsFromBundle(
         return;
       }
       if (combinedRefunds[repaymentChainId][l2TokenAddress] === undefined) {
-        combinedRefunds[repaymentChainId][l2TokenAddress] = refunds;
+        combinedRefunds[repaymentChainId][l2TokenAddress] = _.cloneDeep(refunds);
```
pxrl, line 251: Small nit: We use cloneDeep here, but not on line 256/257. I _think_ it's OK because the issue was that we were making changes to the array, not its component objects. So as long as we're deconstructing with `...refunds` it should be OK.

nicholaspai: https://github.com/across-protocol/relayer/commit/9321275e4d67ff26b3be1a2e441af05fb10d3416

pxrl (approved):
The direct assignment of `refunds` is definitely an issue. This is fairly subtle - very nice catch.

src/dataworker/DataworkerUtils.ts
```diff
@@ -246,14 +246,16 @@ export function getRefundsFromBundle(
...
       if (combinedRefunds[repaymentChainId][l2TokenAddress] === undefined) {
-        combinedRefunds[repaymentChainId][l2TokenAddress] = refunds;
+        combinedRefunds[repaymentChainId][l2TokenAddress] = refundsShallowCopy;
       } else {
         // Each refunds object should have a unique refund address so we can add new ones to the
         // existing dictionary.
         combinedRefunds[repaymentChainId][l2TokenAddress] = {
           ...combinedRefunds[repaymentChainId][l2TokenAddress],
-          ...refunds,
+          ...refundsClone,
```
pxrl, line 258:
```suggestion
          ...refundsShallowCopy,
```

pxrl (approved):

pxrl (approved):

PR #1599, opened 2024-06-07
chore(v20): bump to v20

james-a-morris:
I've seen a roughly 20% speed improvement

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   install:
     docker:
-      - image: cimg/node:16.18.0
+      - image: cimg/node:20.14.0
     working_directory: ~/relayer-v2
     resource_class: medium+
     steps:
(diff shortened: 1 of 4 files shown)


nicholaspai:
> I've seen a roughly 20% speed improvement

should we also bump this in the sdk/contracts?

nicholaspai (approved):

bmzig (approved):

nicholaspai (changes requested):
do we need to change the `engine` in package.json?

james-a-morris:
> do we need to change the `engine` in package.json?

This specifies that the Node version to run this executable should be greater than `>=16.18.0`, or it will throw if we try to run it. Technically, we can run the bots in v16, so it wouldn't need to be changed. I'm going to bump it because we want third party relayers to use v20 moving forward

nicholaspai (approved):

PR #1600, opened 2024-06-07
fix(Finalizer): Update to zksync-ethers instead of zksync-web3

nicholaspai:
This fixes an issue with the finalizer's use of the SDK's `isWithdrawalFinalized()` implementation, which ~seems to no longer work for withdrawals finalized after the v24 zksync upgrade~ no longer works.

 The reason is that the zksync-web3 SDK `isWithdrawalFinalized()` function no longer returns `true` when a withdrawal is finalized. This is counter intuitive but as you can see in the latest implementation (upgraded very recently, right before the v24 upgrade) of the L1ERC20Bridge, the Solidity comment explains how  this behavior was changed. The new way to check the withdrawal finalization status is to query the SharedBridge, which routes to the L1ERC20Bridge/Mailbox. The zksync-ethers SDK takes care of this for us.

(see line 219 in `finalizeWithdrawal()` in the new [L1ERC20Bridge](https://etherscan.io/address/0x8191975d8b0851c7f0740918896cf298c09aa05e#code) implementation)
Screenshot 2024-06-07 at 23 38 52

## Implementation

This PR replaces zksync-web3 (deprecated) with zksync-ethers which nicely is a drop-in replacemet and loads the correct system contracts.

## DO NOT MERGE

Note, this PR should not go in until all ZkSync RPCs are confirmed to support the new upgrade and also all recent withdrawals are all post v24 upgrade. Currently, Quicknode and Alchemy's `zks_getBridgeContracts` do not return the correct `l1SharedDefaultBridge` value (returning it as `null` instead) which would break the `zksync-ethers` call.

```sh
> curl -X POST -H "Content-Type: application/json" \
--data '{"jsonrpc": "2.0", "id": 1, "method": "zks_getBridgeContracts", "params": [  ]}' \
"https://zksync-mainnet.g.alchemy.com/v2/<KEY>"

> {"jsonrpc":"2.0","id":1,"result":{"l1SharedDefaultBridge":null,"l2SharedDefaultBridge":"0x11f943b2c77b743ab90f4a0ae7d5a4e7fca3e102","l1Erc20DefaultBridge":"0x57891966931eb4bb6fb81430e6ce0a03aabde063","l2Erc20DefaultBridge":"0x11f943b2c77b743ab90f4a0ae7d5a4e7fca3e102","l1WethBridge":"0x0000000000000000000000000000000000000000","l2WethBridge":"0x0000000000000000000000000000000000000000"}}
```


nicholaspai:
Looks like QN, Ankr are now up to date so waiting on Alchemy

nicholaspai:
This is all good to merge now that alchemy and QN have updated their zk endpoints

package.json
```diff
@@ -40,7 +40,8 @@
     "superstruct": "^1.0.3",
     "ts-node": "^10.9.1",
     "winston": "^3.10.0",
-    "zksync-web3": "^0.14.3"
+    "zksync-ethers": "^5.7.2",
+    "ascii-table3": "0.9.0"
```
bmzig, line 44: We should remove this dependency.

bmzig (approved):

pxrl (approved):

PR #1601, opened 2024-06-08
fix(Linea): Define BridgingInitiatedV2

pxrl:
Linea have upgraded their token bridge and emit BridgingInitiatedV2 instead of BridgingInitiated. This allows for filtering on receipient to occur, at the expense of amount.

Required before merge:
- Do we ever rely on filtering by amount?
- Do all ERC20s now emit `BridgingInitiatedV2`, or do some still emit `BridgingInitiated`?
- Is this event emitted in both directions?

Side note: I initially thought we'd need the Linea SDK v0.3.0 and had spent some time struggling to update to it. It causes a lot of things in the current Linea adapter implementation to break, and the fixes seem non-trivial. I suspect we might want to re-architect the Linea finalizer to rely less on type unions, because it's painful when Consensys rug the types we rely on.

### DIFF
--- a/src/clients/bridges/LineaAdapter.ts
+++ b/src/clients/bridges/LineaAdapter.ts
@@ -290,7 +290,7 @@ export class LineaAdapter extends BaseAdapter {
   ): Promise<Event[]> {
     const initiatedQueryResult = await paginatedEventQuery(
       l1Bridge,
-      l1Bridge.filters.BridgingInitiated(null /* sender */, null /* recipient, non-indexed must be null */, l1Token),
+      l1Bridge.filters.BridgingInitiatedV2(null /* sender */, null /* recipient, non-indexed must be null */, l1Token),
       l1SearchConfig
     );
     return initiatedQueryResult.filter(({ args }) => args.recipient === monitoredAddress);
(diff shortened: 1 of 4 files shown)


nicholaspai:
Looks like `BridgingFinalized` have also changed: https://lineascan.build/tx/0x853b5767211b563e1ed70d1686802a3ef09d88c3fa4e3aac4ecc5d63ccf0a42f#eventlog

nicholaspai:
I've tested this with DAI and WBTC and it works as well

nicholaspai (approved):

src/clients/bridges/LineaAdapter.ts
```diff
@@ -304,10 +304,15 @@ export class LineaAdapter extends BaseAdapter {
...
+      l2Bridge.filters.BridgingFinalizedV2(
+        l1Token,
+        null /* bridgedToken */,
+        null /* bridgedToken */,
+        monitoredAddress /* recipient */
+      ),
       l2SearchConfig
     );
-    return finalizedQueryResult.filter(({ args }) => args.recipient === monitoredAddress);
+    return finalizedQueryResult;
```
james-a-morris, line 315: We could return this directly without the await above and remove the `async` from the function.

PR #1604, opened 2024-06-10
chore(revert): 1599

james-a-morris:
Reverts across-protocol/relayer#1599


nicholaspai:
We unfortunately cannot support this until the `umaprotocol` node version is also bumped to 20, because of how docker containers are built in gcp using `umaprotocol`

nicholaspai (approved):

PR #1612, opened 2024-06-12
refactor: Drop CCTP from SpokePoolClient update

pxrl:
This was not used in practice and has been removed upstream.


pxrl:
nb. Currently pending a new SDK package & subsequent sync on yarn.lock.

nicholaspai (approved):

PR #1618, opened 2024-06-13
improve(adapters): centralize chain supported tokens

bmzig:
We want to be able to modify a small number of files (e.g. those in `src/common`) in order to support a new chain. A low-hanging fruit which also shrinks the diff in this PR: https://github.com/across-protocol/relayer/pull/1524 will move the tokens supported by a specific route to our constants. 

The motivation for this is to eventually be able to construct a generic adapter which can fetch all of its configuration directly when all it knows is its chain ID.

### DIFF
--- a/src/clients/bridges/ArbitrumAdapter.ts
+++ b/src/clients/bridges/ArbitrumAdapter.ts
@@ -18,7 +18,7 @@ import {
 } from "../../utils";
 import { SpokePoolClient } from "../../clients";
 import { SortableEvent, OutstandingTransfers } from "../../interfaces";
-import { CONTRACT_ADDRESSES } from "../../common";
+import { CONTRACT_ADDRESSES, SUPPORTED_TOKENS } from "../../common";
 import { CCTPAdapter } from "./CCTPAdapter";
 
 // TODO: Move to ../../common/ContractAddresses.ts
@@ -67,18 +67,7 @@ export class ArbitrumAdapter extends CCTPAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     monitoredAddresses: string[]
   ) {
-    super(spokePoolClients, 42161, monitoredAddresses, logger, [
-      "USDC",
-      "USDT",
-      "WETH",
-      "DAI",
-      "WBTC",
-      "UMA",
-      "BADGER",
-      "BAL",
-      "ACX",
-      "POOL",
-    ]);
+    super(spokePoolClients, 42161, monitoredAddresses, logger, SUPPORTED_TOKENS[CHAIN_IDs.ARBITRUM]);
   }
 
   async getOutstandingCrossChainTransfers(l1Tokens: string[]): Promise<OutstandingTransfers> {
--- a/src/clients/bridges/ArbitrumAdapter.ts
+++ b/src/clients/bridges/ArbitrumAdapter.ts
@@ -67,7 +67,8 @@ export class ArbitrumAdapter extends CCTPAdapter {
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
     monitoredAddresses: string[]
   ) {
-    super(spokePoolClients, 42161, monitoredAddresses, logger, SUPPORTED_TOKENS[CHAIN_IDs.ARBITRUM]);
+    const { ARBITRUM } = CHAIN_IDs;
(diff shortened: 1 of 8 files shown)


nicholaspai (commented):
This LGTM, will wait for paul to review

pxrl (approved):
LGTM. We'll need to trim down the testnet supported tokens later, but this is fine for now.

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -67,18 +67,7 @@ export class ArbitrumAdapter extends CCTPAdapter {
...
-      "WETH",
-      "DAI",
-      "WBTC",
-      "UMA",
-      "BADGER",
-      "BAL",
-      "ACX",
-      "POOL",
-    ]);
+    super(spokePoolClients, 42161, monitoredAddresses, logger, SUPPORTED_TOKENS[CHAIN_IDs.ARBITRUM]);
```
pxrl, line 70: It's also tempting to suggest this, but I guess most of this code will go away shortly anyway so it might be redundant.

```suggestion
    const { ARBITRUM } =  CHAIN_IDs;
    super(spokePoolClients, ARBITRUM, monitoredAddresses, logger, SUPPORTED_TOKENS[ARBITRUM]);
```

bmzig: Yeah it probably does not hurt to do this, so I made the change here: https://github.com/across-protocol/relayer/pull/1618/commits/472dc40e9b31b780f308733266d170ac811135d3

src/common/Constants.ts
```diff
@@ -313,6 +313,22 @@ export const chainIdsToCctpDomains: { [chainId: number]: number } = {
   80001: 7, // Polygon PoS Mumbai
 };
 
+export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
+  10: ["DAI", "SNX", "BAL", "ETH", "WETH", "USDC", "POOL", "USDT", "WBTC", "UMA", "ACX"],
```
pxrl, line 317: @bmzig This now specifies ETH and WETH for Optimism, but ETH was not present in the pre-existing OptimismAdapter(directly above this file). Is that intended?

pxrl (approved):
One question, otherwise LGTM.

src/common/Constants.ts
bmzig: That is not intended. Thank you. It is addressed here: https://github.com/across-protocol/relayer/pull/1618/commits/867b3a5f8df6e22be1d6c3bd28455763289033e1

pxrl (approved):
The updated configs all check out for me :+1:

src/common/Constants.ts
```diff
@@ -313,6 +313,22 @@ export const chainIdsToCctpDomains: { [chainId: number]: number } = {
   80001: 7, // Polygon PoS Mumbai
 };
 
+export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
```
pxrl, line 316: One thing that occurred to me - this mapping is exclusive to canonical bridges and does not consider CCTP. We need to either update its structure or add a parallel mapping (or array) for defining which chains are supported by CCTP.

bmzig: I believe that after we add the generic adapters, we can fix this by adding USDC.e to the corresponding chain's arrays and then mapping USDC to the CCTP bridge and USDC.e to the canonical. I won't fix this now since adding in those identifiers might confuse the current infra with an extra USDC entry.

src/common/Constants.ts
```diff
@@ -313,6 +313,22 @@ export const chainIdsToCctpDomains: { [chainId: number]: number } = {
...
+  10: ["DAI", "SNX", "BAL", "WETH", "USDC", "POOL", "USDT", "WBTC", "UMA", "ACX"],
```
james-a-morris, line 317: Could we use `CHAIN_IDS` to reference these keys?

james-a-morris (commented):
Looks good - one comment

src/common/Constants.ts
bmzig: I didn't out of convention; the rest of the `src/common/*` files don't use that mapping. For the sake of uniformity, I would argue that changing one mapping to use `CHAIN_IDs` should mean changing all of them. Are there downsides to doing that (I can't think of any off the top of my head)? If not, the change can be made.

james-a-morris (approved):

PR #1621, opened 2024-06-13
fix(relayer): Skip cross-chain transfer tracking when looping

pxrl:
Checking the logs more closely, it's obvious that some chain adapters repeatedly produce the same copies of deposit hashes. This creates duplicates and will result in delays and odd results in balance projections. A proper fix is needed in the adapters, but this is deferred for now because of the ongoing new-chain automation work.


nicholaspai (commented):
Doesn't this disable IM for the looping relayer?

pxrl:
> Doesn't this disable IM for the looping relayer?

Nope; it just short-circuits CrossChainTransferClient.update(). So we miss any insight into transfers stuck in canonical bridges, but IM will still draw on pending refunds across multiple chains when determining. What will happen is that we will under-report on the virtual balance for each chain in the time period where there is an inbound deposit to that chain.

nicholaspai (approved):

PR #1622, opened 2024-06-13
improve(relayer): Relax future quoteTimestamp constraint

pxrl:
Any users of SpokePool.depositV3Now() inherit the origin chain's block.timestamp as their quoteTimestamp. Any integrators who call deposit() or depositV3() and supply block.timestamp for quoteTimestamp will see the same result. On fast L2s this will almost always be ahead of the latest HubPool timestamp. Per current relayer implementation, these deposits are silently ignored until the quoteTimestamp is behind the HubPool timestamp.

The risk of resolving an inaccurate LP utilisation at the quoteTimestamp is significantly reduced in v3, such that it's probably now not worth being strict on. The residual risk of future quoteTimestamps is really only when there are large gaps in which HubPool utilisation could change dramatically, and even then, the effect should be small (relayer repayment is somewhat less than expected). The SpokePools already enforce that the quoteTimestamp cannot be ahead of the origin chain's timestamp, so the risk of this is only present if there is significant clock drift between chains.

To balance this, permit quoteTimestamps to be up to 2 blocks ahead of the current HubPool block. This allows some flexibility where there is 0 or negligible effect, whilst still protecting against weird outcomes in the event of clock drift.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -482,8 +482,9 @@ export class Relayer {
     const { hubPoolClient, profitClient, spokePoolClients, tokenClient, multiCallerClient } = this.clients;
 
     // Fetch the average block time for mainnet, for later use in evaluating quoteTimestamps.
-    this.hubPoolBlockBuffer ??=
-      HUB_SPOKE_BLOCK_LAG * (await sdkUtils.averageBlockTime(hubPoolClient.hubPool.provider)).average;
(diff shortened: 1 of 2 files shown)


pxrl:
nb. Most quoteTimestamps are set by the FE/API, which always winds back a few minutes from the latest HubPool timestamp. For those deposits there will be no difference as a result of this change, because the quoteTimestamp is always in the past.

nicholaspai (approved):

PR #1623, opened 2024-06-14
test(arb-adapter): add unit tests

dohaki:
Closes ACX-2207

Uses https://github.com/across-protocol/relayer/pull/1631 as base


bmzig:
Do you think you could add tests for USDC? Once we use this with the generic adapter, Arbitrum will use CCTP and its canonical bridge, so I'm thinking that the test should cover both bridges.

dohaki:
> Do you think you could add tests for USDC? Once we use this with the generic adapter, Arbitrum will use CCTP and its canonical bridge, so I'm thinking that the test should cover both bridges.

yea good point. I will add them

james-a-morris (commented):
Any chance we can add some fuzz testing so that we're not accidentally testing any specific size ranges for deposits? I.e. CCTP had an issue that was uncovered because we tried to bridge too many funds within one message.

src/clients/bridges/ArbitrumAdapter.ts
```diff
@@ -236,10 +267,14 @@ export class ArbitrumAdapter extends CCTPAdapter {
...
     threshold: BigNumber,
     target: BigNumber,
-    simMode = false
+    simMode = false,
+    skipChainIdCheck = false
   ): Promise<TransactionResponse | null> {
     const { chainId } = this;
-    assert(42161 === chainId, `chainId ${chainId} is not supported`);
+
+    if (!skipChainIdCheck) {
```
bmzig, line 275: What's the purpose of `skipChainIdCheck`?

bmzig:
Nice. This also LGTM. I'm curious to see how easily this will translate to the generic stuff since it looks like you already are using `getL1DepositInitiatedEvents/getOutstandingCrossChainTransfers` which is super helpful.

dohaki:
> Any chance we can add some fuzz testing so that we're not accidentally testing any specific size ranges for deposits? I.e. CCTP had an issue that was uncovered because we tried to bridge too many funds within one message.

@james-a-morris can you give a rough outline how you would do that in the context of these unit tests that are added as part of this PR? I agree that fuzz testing for the bug you mentioned makes sense but not sure how to achieve that with mocked CCTP contracts... And if that's in the same scope as these unit tests

src/clients/bridges/ArbitrumAdapter.ts
dohaki: Ah yea that's not needed. Removed

pxrl (approved):

dohaki:
> > Any chance we can add some fuzz testing so that we're not accidentally testing any specific size ranges for deposits? I.e. CCTP had an issue that was uncovered because we tried to bridge too many funds within one message.
> 
> @james-a-morris can you give a rough outline how you would do that in the context of these unit tests that are added as part of this PR? I agree that fuzz testing for the bug you mentioned makes sense but not sure how to achieve that with mocked CCTP contracts... And if that's in the same scope as these unit tests

@james-a-morris I will scope this out of this PR. And if we decide to add those tests, will do so in a separate PR

james-a-morris (approved):

PR #1628, opened 2024-06-18
fix(PolygonAdapter): Only approve mainnet token gateways

pxrl:
The l1Bridge contract is used for event scanning but does not directly receive tokens from the relayer's EOA, so it doesn't require any token approval.


pxrl:
Forget this - this seems to be a quirk of the Polygon bridge contracts.

PR #1629, opened 2024-06-18
feat(lite chain): force origin chain repayment

james-a-morris:
This PR is blocked by https://github.com/across-protocol/sdk/pull/662

The goal of this PR is to make sure fills made by the relayer are made with the origin chain id. Since the same line of code was highly correlated, the removal of slow fills for lite chains has also been added.

There is one outstanding piece of information needed, which I am resolving offline. Therefore, I will keep this as WiP


src/relayer/Relayer.ts
```diff
@@ -363,25 +364,50 @@ export class Relayer {
       const { relayerFeePct, gasCost, gasLimit: _gasLimit, lpFeePct: realizedLpFeePct } = repaymentChainProfitability;
       if (isDefined(repaymentChainId)) {
         const gasLimit = isMessageEmpty(resolveDepositMessage(deposit)) ? undefined : _gasLimit;
-        this.fillRelay(deposit, repaymentChainId, realizedLpFeePct, gasLimit);
+        this.fillRelay(deposit, isLiteChain ? originChainId : repaymentChainId, realizedLpFeePct, gasLimit);
```
pxrl, line 367: I think this belongs moreso in `relayer.resolveRepaymentChain()` - it shouldn't be necessary to add it here :+1:

src/relayer/Relayer.ts
```diff
@@ -363,25 +364,50 @@ export class Relayer {
...
 
         // Update local balance to account for the enqueued fill.
         tokenClient.decrementLocalBalance(destinationChainId, deposit.outputToken, deposit.outputAmount);
       } else {
         profitClient.captureUnprofitableFill(deposit, realizedLpFeePct, relayerFeePct, gasCost);
       }
     } else if (selfRelay) {
```
pxrl, line 374: This code is almost never used, so I'd prefer to keep it really small by just updating the existing logic as follows:

```suggestion
    } else if (selfRelay && !isLiteChain) {
```

src/relayer/Relayer.ts
```diff
@@ -363,25 +364,50 @@ export class Relayer {
...
+      // Disable slow fills for lite chains.
+      if (isLiteChain) {
+        this.logger.debug({
+          at: "Relayer::evaluateFil::slowFill",
+          message: `Skipping slow fill for lite chain deposit ${depositId}.`,
+          depositId,
+          depositor,
+          recipient,
+          transactionHash: deposit.transactionHash,
+          isLiteChain,
```
pxrl, line 402: This is implied by the log message itself.

```suggestion
```

src/relayer/Relayer.ts
```diff
@@ -363,25 +364,50 @@ export class Relayer {
...
+        });
```
pxrl, line 403: Thoughts on relocating this into `Relayer.requestSlowFill()` ? Otherwise this "main" relayer function ends up quite long and filled with lots of verbose log messages.

src/clients/InventoryClient.ts
```diff
@@ -404,6 +405,11 @@ export class InventoryClient {
     const { originChainId, destinationChainId, inputToken, outputToken, outputAmount, inputAmount } = deposit;
     const hubChainId = this.hubPoolClient.chainId;
 
+    // Check if the origin chain is currently a lite chain. If it is, then we should only consider the origin chain
+    if (this.hubPoolClient.configStoreClient.isChainLiteChainAtBlock(originChainId)) {
```
nicholaspai, line 409: I'm not sure this belongs in here. @pxrl wdyt? should the lite chain decision be in here or in the relayer? Also james I see you're not apassing in the deposit.quoteTimestamp which is going to be necessary to determine if the deposit origin chain was a lite chain at deposit time

src/relayer/Relayer.ts
nicholaspai: I agree with paul

src/relayer/Relayer.ts
james-a-morris: Good call. I kept the internal conditional just because then we'd have to add an else-if to the else statement below.

src/relayer/Relayer.ts
james-a-morris: Good call

src/clients/InventoryClient.ts
james-a-morris: I'd prefer to keep this here because we want to lock repayment chain possibilities of lite chains to the origin. THis function is used by the relayer's repayment chain id function to determine which possible chains to reason about.

src/relayer/Relayer.ts
```diff
@@ -553,6 +557,16 @@ export class Relayer {
   }
 
   requestSlowFill(deposit: V3Deposit): void {
+    // Verify that this deposit does not originate on a lite chain, since slow fills are not supported for lite chains.
+    if (deposit.originatesFromLiteChain) {
+      this.logger.warn({
```
nicholaspai, line 562:
```suggestion
      this.logger[this.config.sendingRelaysEnabled ? "warn" : "debug"]({
```

nicholaspai: we should try to reduce noise in the rebalancer when these deposits happen

src/relayer/Relayer.ts
```diff
@@ -371,6 +372,9 @@ export class Relayer {
         profitClient.captureUnprofitableFill(deposit, realizedLpFeePct, relayerFeePct, gasCost);
       }
     } else if (selfRelay) {
+      if (originatesFromLiteChain) {
+        return;
```
nicholaspai, line 376: @pxrl isn't this going to be an issue if we ever send a self deposit? We'll never fill it meaning funds will be stuck forever? Why can't we fill these and then get repaid on the origin chain?

src/clients/BundleDataClient.ts
```diff
@@ -836,6 +836,14 @@ export class BundleDataClient {
                 // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
                 // so this fill is validated.
                 v3RelayHashes[relayDataHash].fill = fill;
+
+                // We need to ensure that the repayment chain is the origin chain if the deposit originates
```
nicholaspai, line 840: This change isn't sufficient, as there are several other places where we populate the `bundleFillsV3` object. 

The important object we need to modify is `bundleFillsV3`, which should only be modified in this file by `updateBundleFillsV3()`. Therefore if you search for this function, you'll see that repayment chain ID for a successful fill is only computed by `getRefundInformationFromFill()` but please double check me on this

nicholaspai (changes requested):
The dataworker changes won't work yet and needs some tests, but the relayer looks good though still probably needs tests

src/clients/BundleDataClient.ts
```diff
@@ -1131,7 +1117,8 @@ export class BundleDataClient {
                 fill,
                 this.clients.hubPoolClient,
                 blockRangesForChains,
-                chainIds
+                chainIds,
+                v3RelayHashes[this.getRelayHashFromEvent(fill)].deposit.originatesFromLiteChain
```
nicholaspai, line 1121: what if `deposit.originatesFromLiteChain` isn't defined?

james-a-morris: These are loops over validated fills that require a deposit to be defined. However, we can change it to be something like: https://github.com/across-protocol/relayer/pull/1629/commits/2db63b8bd2fbb96216731a2066913aaf0a5ded17

src/clients/BundleDataClient.ts
james-a-morris: This was the result of a local change not being pushed. This change is now outdated and the new version uses a closer to where the data is updated.

src/relayer/Relayer.ts
pxrl: The funds won't be stuck forever since if not filled then they'd just be refunded on the origin chain. I think the most likely scenario here is that a third-party relayer wants to rebalance out of a lite chain (i.e. lite chain back to mainnet). In that scenario it'd be pretty annoying if their own bot kept filling it and requesting payment on the origin chain.

src/relayer/Relayer.ts
```diff
@@ -553,6 +553,16 @@ export class Relayer {
   }
 
   requestSlowFill(deposit: V3Deposit): void {
+    // Verify that this deposit does not originate on a lite chain, since slow fills are not supported for lite chains.
+    if (deposit.originatesFromLiteChain) {
+      this.logger[this.config.sendingRelaysEnabled ? "warn" : "debug"]({
```
pxrl, line 558: I don't think this should warn - it'll create a lot of noise out of the main relayer for the duration of the lookback (2 hours).

src/relayer/Relayer.ts
```diff
@@ -553,6 +553,16 @@ export class Relayer {
   }
 
   requestSlowFill(deposit: V3Deposit): void {
+    // Verify that this deposit does not originate on a lite chain, since slow fills are not supported for lite chains.
+    if (deposit.originatesFromLiteChain) {
```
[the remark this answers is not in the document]
james-a-morris: Agreed - but I'm not sure this is in the scope of this PR

src/utils/FillUtils.ts
```diff
@@ -8,14 +8,20 @@ export function getRefundInformationFromFill(
   fill: Fill,
   hubPoolClient: HubPoolClient,
   blockRangesForChains: number[][],
-  chainIdListForBundleEvaluationBlockNumbers: number[]
+  chainIdListForBundleEvaluationBlockNumbers: number[],
+  originatesFromLiteChain: boolean
```
nicholaspai, line 12: nit: prefer naming this `fromLiteChain`

test/Dataworker.loadData.ts
```diff
@@ -343,6 +359,69 @@ describe("Dataworker: Load data used in all functions", async function () {
...
+      // Mock the config store client to include the lite chain index.
+      mockConfigStore.updateGlobalConfig(
+        GLOBAL_CONFIG_STORE_KEYS.LITE_CHAIN_ID_INDICES,
+        JSON.stringify([spokePoolClient_1.chainId])
+      );
+      await mockConfigStore.update();
+      // Ensure that our test has the right setup.
+      expect(mockConfigStore.liteChainIndicesUpdates.length).to.equal(1);
+      mockConfigStore.liteChainIndicesUpdates[0].timestamp = 0;
+      expect(repaymentChainId).to.not.eq(spokePoolClient_1.chainId);
```
nicholaspai, line 373:
```suggestion
      expect(repaymentChainId).to.not.eq(originChainId);
```

is this more correct for preserving this test's usefulness?

test/Dataworker.loadData.ts
```diff
@@ -381,6 +460,63 @@ describe("Dataworker: Load data used in all functions", async function () {
...
+      mockConfigStore.updateGlobalConfig(
+        GLOBAL_CONFIG_STORE_KEYS.LITE_CHAIN_ID_INDICES,
+        JSON.stringify([spokePoolClient_1.chainId])
+      );
+      await mockConfigStore.update();
+      // Ensure that our test has the right setup.
+      expect(mockConfigStore.liteChainIndicesUpdates.length).to.equal(1);
+      mockConfigStore.liteChainIndicesUpdates[0].timestamp = depositObject.quoteTimestamp - 1;
+      expect(mockConfigStore.liteChainIndicesUpdates[0].timestamp).to.be.lt(depositObject.quoteTimestamp);
+      expect(repaymentChainId).to.not.eq(spokePoolClient_1.chainId);
```
nicholaspai, line 495:
```suggestion
      expect(repaymentChainId).to.not.eq(originChainId);
```

same suggestion as above

nicholaspai (commented):
This looks good! Question is should we wait for https://github.com/across-protocol/sdk/pull/669 also to be incorporated? I also left some nits pls review

src/relayer/Relayer.ts
nicholaspai: in either case this filler is going to receive funds on the origin chain i don't understand?

src/relayer/Relayer.ts
nicholaspai: i'm good with that ^

nicholaspai: let's change to debug

src/relayer/Relayer.ts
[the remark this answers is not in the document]
nicholaspai: just a reminder i don't want to forget

src/relayer/Relayer.ts
nicholaspai: ok i understand. I think the original implementation where we exit early here is preferable but I didn't understand the original context. Can you add a comment here that we don't want to fast fill any deposit that we send because we should assume that  for these deposits, we would always prefer someone else to fill them since we must take repayment on the origin chain, which would result in a net-no change in capital allocation for our account and force us to pay an LP fee

test/Dataworker.loadData.ts
james-a-morris: https://github.com/across-protocol/relayer/pull/1629/commits/a489074b3e00abe73d27929ddb2c69b953ad86b4

test/Dataworker.loadData.ts
james-a-morris: https://github.com/across-protocol/relayer/pull/1629/commits/a489074b3e00abe73d27929ddb2c69b953ad86b4

src/utils/FillUtils.ts
james-a-morris: https://github.com/across-protocol/relayer/pull/1629/commits/61d5d998dbbb0e5bbe002aedfcf584875f151d2d

src/relayer/Relayer.ts
```diff
@@ -375,7 +375,7 @@ export class Relayer {
       // relayer is both the depositor and the recipient, because a deposit on a cheap SpokePool chain could cause
       // expensive fills on (for example) mainnet.
       const { lpFeePct } = lpFees.find((lpFee) => lpFee.paymentChainId === destinationChainId);
-      this.fillRelay(deposit, destinationChainId, lpFeePct);
+      this.fillRelay(deposit, deposit.originatesFromLiteChain ? deposit.originChainId : destinationChainId, lpFeePct);
```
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/relayer/pull/1629/commits/33012fb800780137497b065ad02a59dfa0d20f4d

src/relayer/Relayer.ts
james-a-morris: https://github.com/across-protocol/relayer/pull/1629/commits/33012fb800780137497b065ad02a59dfa0d20f4d
https://github.com/across-protocol/relayer/pull/1629/commits/42318f204156d570b66f2df6f130cb0574e70b4e

src/clients/BundleDataClient.ts
```diff
@@ -1111,11 +1118,18 @@ export class BundleDataClient {
...
+                  `Associated deposit not found for a validated fill: (Origin Chain ID: ${fill.originChainId}, Deposit ID: ${fill.depositId})`
+                );
+              }
               const { chainToSendRefundTo: paymentChainId } = getRefundInformationFromFill(
                 fill,
                 this.clients.hubPoolClient,
                 blockRangesForChains,
-                chainIds
+                chainIds,
+                v3RelayHashes[this.getRelayHashFromEvent(fill)].deposit.fromLiteChain
```
nicholaspai, line 1132:
```suggestion
                matchedDeposit.fromLiteChain
```

nicholaspai (commented):
Will review again after we merge in #1635

src/clients/InventoryClient.ts
```diff
@@ -404,6 +405,11 @@ export class InventoryClient {
     const { originChainId, destinationChainId, inputToken, outputToken, outputAmount, inputAmount } = deposit;
     const hubChainId = this.hubPoolClient.chainId;
 
+    // Check if the deposit originates from a lite chain.
+    if (deposit.fromLiteChain) {
+      return [originChainId];
```
nicholaspai, line 410: I think we can do better than this. I think if the origin chain is a lite chain we can evaluate the origin chain as a possible repayment chain choice. Then go through the usual under or over allocation decision. 

Typically if a chain is under allocated we will proceed with the repayment chain otherwise default to mainnet. 

However for lite chains, we can protect our inventory by NOT defaulting to mainnet and choosing not to fill this deposit if we hit that fallback case. This way we never get our inventory over allocated on a lite chain.

nicholaspai (changes requested):

src/relayer/Relayer.ts
```diff
@@ -555,6 +567,26 @@ export class Relayer {
   }
 
   requestSlowFill(deposit: V3Deposit): void {
+    // don't request slow fill if origin chain is a lite chain
+    if (deposit.fromLiteChain) {
```
nicholaspai, line 571: Should we merge these if statements?

nicholaspai (approved):
I'd like to see @pxrl review

src/clients/InventoryClient.ts
```diff
@@ -459,6 +462,7 @@ export class InventoryClient {
     // since its the fallback chain if both destination and origin chain are over allocated.
     // If destination chain is hub chain, we still want to evaluate it before the origin chain.
     if (
+      !deposit.fromLiteChain &&
       !chainsToEvaluate.includes(destinationChainId) &&
       this._l1TokenEnabledForChain(l1Token, Number(destinationChainId))
```
pxrl, line 467: Minor: The majority of deposits will not originate from a lite chain so I think it makes sense to evaluate this last.

```suggestion
      !chainsToEvaluate.includes(destinationChainId) &&
      this._l1TokenEnabledForChain(l1Token, Number(destinationChainId)) &&
      !deposit.fromLiteChain
```

src/clients/ProfitClient.ts
```diff
@@ -590,6 +590,8 @@ export class ProfitClient {
       exclusivityDeadline: 0,
       exclusiveRelayer: ZERO_ADDRESS,
       message: EMPTY_MESSAGE,
+      fromLiteChain: false, // TODO: We *may* care about this in the future.
+      toLiteChain: false, // TODO: We *may* care about this in the future.
```
pxrl, line 594: These don't affect gas consumption so we're OK to drop the comments.

```suggestion
      fromLiteChain: false,
      toLiteChain: false,
```

src/relayer/Relayer.ts
```diff
@@ -373,6 +373,18 @@ export class Relayer {
         profitClient.captureUnprofitableFill(deposit, realizedLpFeePct, relayerFeePct, gasCost);
       }
     } else if (selfRelay) {
+      // Prefer exiting early here to avoid fast filling any deposits we send. This approach assumes that we always
+      // prefer someone else to fill the deposits. By doing this, we ensure repayment on the origin chain, resulting
+      // in no net change in capital allocation for our account and avoiding the need to pay an LP fee.
```
pxrl, line 378: I don't think this comment is correct - avoiding self-fills from lite chains will result in the relayer moving its balance from the lite chain to the destination chain.

```suggestion
      // prefer someone else to fill the deposits.
```

src/relayer/Relayer.ts
pxrl: +1 - I think we should optimise compact code on these lite clauses.

src/relayer/Relayer.ts
```diff
@@ -608,6 +640,12 @@ export class Relayer {
       realizedLpFeePct,
     });
 
+    // If a deposit originates from a lite chain, then the repayment chain must be the origin chain.
+    assert(
+      !deposit.fromLiteChain || repaymentChainId === deposit.originChainId,
+      `Lite chain deposits must be filled on their origin chain. Deposit ${deposit.depositId}.`
```
pxrl, line 646: It'd help to identify the origin chain as well, because otherwise there's no context to the deposit ID :+1:

test/Relayer.UnfilledDeposits.ts
```diff
@@ -188,7 +188,7 @@ describe("Relayer: Unfilled Deposits", async function () {
 
     unfilledDeposits = _getAllUnfilledDeposits();
     expect(unfilledDeposits)
-      .excludingEvery(["realizedLpFeePct", "quoteBlockNumber"])
+      .excludingEvery(["realizedLpFeePct", "quoteBlockNumber", "fromLiteChain", "toLiteChain"])
```
pxrl, line 191: ooc, why do these fields need to be ignored? They seem to be required per the Deposit type in the SDK, so I'd have expected them to be defined.

src/clients/BundleDataClient.ts
```diff
@@ -306,11 +306,18 @@ export class BundleDataClient {
           return hasMatchingDeposit;
         })
         .forEach((fill) => {
+          const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
+          if (!isDefined(matchingDeposit?.fromLiteChain)) {
+            throw new Error(
+              `Associated deposit lite-chain flag not found for fill: (Origin Chain ID: ${fill.originChainId}, Deposit ID: ${fill.depositId})`
+            );
```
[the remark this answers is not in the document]
pxrl: Looking above, I can see that we have already filtered by `hasMatchingDeposit`, so I guess we can probably just `assert(isDefined(matchingDeposit))`. wdyt?

(I think this could probably be refactored later but I don't think it makes sense to do that as part of introducing a new feature)

src/clients/BundleDataClient.ts
```diff
@@ -923,6 +930,14 @@ export class BundleDataClient {
                   return;
                 }
 
+                // slow fill requests for deposits from or to lite chains are considered invalid
+                if (
+                  v3RelayHashes[relayDataHash].deposit.fromLiteChain ||
+                  v3RelayHashes[relayDataHash].deposit.toLiteChain
```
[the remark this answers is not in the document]
amateima: > we're concerned about the slow fill being unexecutable on arrival and needing to be pulled back in a subsequent bundle. Is that the reason for this?

Yes, that's the reason

src/clients/BundleDataClient.ts
```diff
@@ -1111,11 +1132,18 @@ export class BundleDataClient {
       validatedBundleV3Fills.length > 0
         ? this.clients.hubPoolClient.batchComputeRealizedLpFeePct(
             validatedBundleV3Fills.map((fill) => {
+              const matchedDeposit = v3RelayHashes[this.getRelayHashFromEvent(fill)].deposit;
+              if (!isDefined(matchedDeposit?.fromLiteChain)) {
+                throw new Error(
+                  `Associated deposit not found for a validated fill: (Origin Chain ID: ${fill.originChainId}, Deposit ID: ${fill.depositId})`
+                );
+              }
```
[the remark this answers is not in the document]
nicholaspai: Agreed this should never hit

src/clients/InventoryClient.ts
```diff
@@ -393,8 +393,11 @@ export class InventoryClient {
    * needed to cover other unfilled deposits in addition to current token balances. Slow withdrawal chains are only
    * selected if the SpokePool's running balance for that chain is over the system's desired target.
    * @dev The HubChain is always evaluated as a fallback option if the inventory management is enabled and all other
-   * chains are over-allocated.
-   * @dev If inventory management is disabled, then destinationChain is used as a default.
+   * chains are over-allocated, unless the origin chain is a lite chain, in which case.
+   * there is no fallback if the origin chain is not an eligible repayment chain.
```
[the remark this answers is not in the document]
nicholaspai: It's ineligible if the origin chain is over allocated from an inventory management perspective

james-a-morris:
@pxrl consolidating comments related to the asserts/throws:

https://github.com/across-protocol/relayer/pull/1629/commits/9857d74d53722a72132722a74ec1c958de784086

src/relayer/Relayer.ts
james-a-morris: Very fair point

test/Relayer.UnfilledDeposits.ts
james-a-morris: They are ignored for the scope of the test. Our harness itself isn't set up to handle these yet, so I figured it was convenient to ignore and test them separately

pxrl (approved):
LGTM!

PR #1630, opened 2024-06-18
refactor(ContractAddresses): reorganize contract ABIs

bmzig:
> The file `src/common/ContractAddresses.ts` associates bridge ABIs 1:1 with a chain ID. This is OK for production, but requires all ABIs to be defined again on testnets. This also scales poorly for the world of 100 or 1000 chains, where a single file would potentially need to contain many many different ABIs and contract definitions.

> The `ContractAddresses` file should be updated to map chain IDs to contract addresses, and should map the contract address to an ABI. 

In this commit we `import` the ABIs from a separate folder so that the JSON can be loaded at compile time.


src/common/abi/Weth.json
pxrl, line 1: It looks like we previously had multiple instances of WETH ABIs, each with different functions defined. This PR now bundles them together - were you able to confirm that they were all a 1:1 match and that this is the common subset supported by all deployments?

pxrl (approved):
LGTM! One question about the WETH ABI, since we seem to have had some bespoke/minimal snowflake defintions.

After we get this merged, it'd be nice to follow up in another PR to reflow the ABIs so that they have a consistent format. I'm partial to the more compact format where each input type is defined on a single line - it was much easier to review than the verbose one.

src/common/abi/Weth.json
bmzig: Yes. In total there were 3 "different" Weth ABIs. There was a normal Weth ABI which had `balanceOf`, `deposit` and `withdraw` which overlapped with the OpStack Weth ABI, which had those three functions as well as a `Deposit` event. Finally, there was a ZkSync Eth and Weth Abi, both of which were identical and contained just a `Transfer` event. The only way the ZkSync ABI differed from the WETH9 ABI was that the input names were different, which should not matter in this context.

pxrl: I think this is OK - let's get this in.

PR #1631, opened 2024-06-19
test(op-stack): refine unit tests

dohaki:
Closes ACX-2206

Refines existing unit tests for "old" OP stack adapters.

### DIFF
--- a/contracts/MockCctpEvents.sol
+++ b/contracts/MockCctpEvents.sol
@@ -1,0 +1,53 @@
+pragma solidity ^0.8.0;
+
+contract CctpTokenMessenger {
+    event DepositForBurn(
+        uint64 indexed nonce,
+        address indexed burnToken,
+        uint256 amount,
+        address indexed depositor,
+        bytes32 mintRecipient,
+        uint32 destinationDomain,
+        bytes32 destinationTokenMessenger,
+        bytes32 destinationCaller
+    );
+    event MintAndWithdraw(address indexed mintRecipient, uint256 amount, address indexed mintToken);
+
+    function emitDepositForBurn(
+        uint64 nonce,
+        address burnToken,
+        uint256 amount,
+        address depositor,
+        bytes32 mintRecipient,
+        uint32 destinationDomain,
+        bytes32 destinationTokenMessenger,
+        bytes32 destinationCaller
+    ) external {
+        emit DepositForBurn(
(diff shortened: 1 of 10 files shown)


contracts/MockCctpEvents.sol
```diff
@@ -0,0 +1,53 @@
...
+}
+
+contract CctpMessageTransmitter {
+    mapping(bytes32 => uint256) public usedNonces;
+
+    function setUsedNonce(bytes32 nonceHash, bool isUsed) external {
+        if (isUsed) {
+            usedNonces[nonceHash] = 1;
+        } else {
+            usedNonces[nonceHash] = 0;
```
bmzig, line 50:
```suggestion
usedNonces[nonceHash] = isUsed ? 1 : 0;
```

bmzig:
Everything else LGTM.

contracts/MockCctpEvents.sol
dohaki: Nice didn't know that Solidity has ternaries by now

bmzig (approved):

src/clients/bridges/op-stack/OpStackAdapter.ts
```diff
@@ -66,7 +66,7 @@ export class OpStackAdapter extends BaseAdapter {
     // which maps to either the CCTP or OVM Standard bridge depending on the request.
     const usdcAddress = TOKEN_SYMBOLS_MAP.USDC.addresses[this.hubChainId];
     const l2NativeUsdcAddress = TOKEN_SYMBOLS_MAP.USDC.addresses[this.chainId];
-    if (usdcAddress && l2NativeUsdcAddress) {
+    if (usdcAddress && l2NativeUsdcAddress && !this.customBridges[usdcAddress]) {
```
pxrl, line 69: I refactored the code above recently, such that `customBridges` is assigned entirely within this constructor, so this is now the only place we actually assign custom USDC. I think it's probably OK to either drop this check, or assert on it on the line above. Either are fine w/ me.

pxrl (approved):

src/clients/bridges/op-stack/OpStackAdapter.ts
dohaki: yea good point. removed it 👍

PR #1632, opened 2024-06-20
test(briges): Add zkSync adapter test

pxrl:
This PR adds tests for the zkSync chain adapter. These tests apply to the existing implementation, but are mostly forwards-compatible with the new unified adapter design. Some refactoring of the existing adapter was needed in order to support test, and that was a useful exercise to better understand some of the challenges that exist in the bridge adapters in general.


src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -123,20 +100,10 @@ export class ZKSyncAdapter extends BaseAdapter {
             finalizedQueryResult = matchL2EthDepositAndWrapEvents(finalizedQueryResult, wrapQueryResult);
           }
         } else {
+          const [from, to] = isSpokePool ? [hubPool.address, address] : [address, address];
```
pxrl, line 103: This should actually fixed Hub -> Spoke transfer tracking, which otherwise would never have worked because this code assumed an EOA transfer.

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -67,53 +65,32 @@ export class ZKSyncAdapter extends BaseAdapter {
...
+            this.queryL1BridgeInitiationEvents(l1TokenAddress, address, address, l1SearchConfig),
+            this.queryL2BridgeFinalizationEvents(l1TokenAddress, address, address, l2SearchConfig),
+          ];
+
+          // For WETH transfers involving an EOA, only count them if a wrap txn followed the L2 deposit finalization.
+          if (!isSpokePool) {
+            const filter = l2WethContract.filters.Transfer(ZERO_ADDRESS, address);
+            queries.push(paginatedEventQuery(l2WethContract, filter, l2SearchConfig));
+          }
+          [initiatedQueryResult, finalizedQueryResult, wrapQueryResult] = await Promise.all(queries);
```
bmzig, line 91: Just to be sure, this works even if the above conditional does not trigger?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -162,6 +129,67 @@ export class ZKSyncAdapter extends BaseAdapter {
...
+  }
+
+  getAddressAlias(l1Address: string): string {
+    return ethersUtils.getAddress(zksync.utils.applyL1ToL2Alias(l1Address));
+  }
+
+  async queryL2BridgeFinalizationEvents(
+    l1Token: string,
+    sender: string | null = null,
+    recipient: string | null = null,
```
bmzig, line 169: I think we should remove the ` | null = null` here and even enforce that the sender/recipient is defined since if they are undefined, (and the l1Token is Weth), we will have the filter be `Transfer(null, null)` which would return a bunch of events and presumably break the bot. (I'd also argue to do this for `queryL1BridgeInitiationEvents`).

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -162,6 +129,67 @@ export class ZKSyncAdapter extends BaseAdapter {
...
+    searchConfig: EventSearchConfig
+  ): Promise<Event[]> {
+    let bridge: Contract;
+    let filter: EventFilter;
+
+    if (this.isWeth(l1Token)) {
+      if (sender === null && recipient !== null) {
```
bmzig, line 176: What's the purpose of defining sender? It does not look like we use it in this function when this conditional is triggered.

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -162,6 +129,67 @@ export class ZKSyncAdapter extends BaseAdapter {
...
+        sender = this.getAddressAlias(
+          (await this.isL2ChainContract(recipient)) ? this.getHubPool().address : this.getAtomicDepositor().address
+        );
+      }
+
+      bridge = this.getL2Eth();
+      filter = bridge.filters.Transfer(null, recipient);
```
bmzig, line 183: Should we have this?

```suggestion
      filter = bridge.filters.Transfer(sender, recipient);
```

src/clients/bridges/ZKSyncAdapter.ts
bmzig: As referenced in another comment, I don't think this can ever trigger since the `if(this.isWeth(l1Token))` implies that this was called from line 83, which sets both the sender and recipient to the same address.

src/clients/bridges/ZKSyncAdapter.ts
pxrl: Yeah this works OK - `wrapQueryResult` is just undefined in that case. This approach is actually inherited from the pre-existing implementation.

src/clients/bridges/ZKSyncAdapter.ts
pxrl: `null, <recipient>` does get used in the zkSync test to verify that the adapter can catch _all_ deposits destined for the recipient. But I agree that it's not actually used by the application code, as you spotted.

I do think there's some value in being able to wildcard one of the inputs, but it's definitely not valid to permit both inputs to be wildcarded.

src/clients/bridges/ZKSyncAdapter.ts
bmzig: This means that if we are monitoring the spoke pool, we will only ever look at transfer events to/from the hub pool (and not from, say, the RL relayer). Just want to verify that that is indeed what we want. I agree that regardless, this is an improvement since we currently wouldn't measure any outstanding ERC20 transfers from hub -> spoke.

bmzig:
I think this makes sense. I'll do another pass, but I think we should get this in.

bmzig (approved):
+1 The tests look solid.

src/clients/bridges/ZKSyncAdapter.ts
pxrl: Correct. In the relayer we only monitor the relayer's own EOA. In the monitor we tend to track the EOAs that we're interested in (nominally 0x428 and 0x07a) and all of the SpokePool addresses, as well as the HubPool address.

I suspect that in future we can probably revisit whether it's necessary to track the HubPool's address. It'd be simpler if we could narrow the definition of a monitored address to be an address on the _remote_ chain.

test/cross-chain-adapters/zkSync.ts
```diff
@@ -0,0 +1,538 @@
...
+    gasPerPubdataLimit;
+
+    return toBN(2_000_000);
+  }
+}
+
+describe("Cross Chain Adapter: zkSync", async function () {
+  const logger = createSpyLogger().spyLogger;
+  const l2TxGasLimit = bnZero;
+  const l2TxGasPerPubdataByte = bnZero;
```
james-a-morris, line 82: Are we storing this so that we can have a typed entry in the function call?

src/clients/bridges/ZKSyncAdapter.ts
```diff
@@ -40,12 +42,8 @@ export class ZKSyncAdapter extends BaseAdapter {
     const { l1SearchConfig, l2SearchConfig } = this.getUpdatedSearchConfigs();
 
     // Resolve the mailbox and bridge contracts for L1 and L2.
-    const l2EthContract = this.getL2Eth();
     const l2WethContract = this.getL2Weth();
-    const atomicWethDepositor = this.getAtomicDepositor();
     const hubPool = this.getHubPool();
-    const l1ERC20Bridge = this.getL1ERC20BridgeContract();
-    const l2ERC20Bridge = this.getL2ERC20BridgeContract();
```
james-a-morris, line 48: Is this removed in lieu of the new logic?

test/cross-chain-adapters/zkSync.ts
pxrl: It's basically just trying to make it self-documenting in order to make it obvious what the input arguments to a function call are (i.e. to avoid `await function(..., bnZero, bnZero)`.

contracts/MockHubPool.sol
```diff
@@ -0,0 +1,11 @@
+// This file contains contracts that can be used to unit test bridge adapters.
```
james-a-morris, line 1: Was this purpose-built or migrated? I believe we have another mock spokepool somewhere else in the org

contracts/MockWETH9.sol
```diff
@@ -0,0 +1,12 @@
...
+
+pragma solidity ^0.8.0;
+
+contract MockWETH9 {
+    event Transfer(address indexed from, address indexed to, uint256 amount);
+
+    function transfer(address from, address to, uint256 amount) external {
+        emit Transfer(from, to, amount);
+    }
+}
```
james-a-morris, line 12: Could we use `WETH9.sol` contract in the contracts?

src/clients/bridges/ZKSyncAdapter.ts
james-a-morris: This tracks @bmzig.

Although, if we wanted to look at L1 -> L2 we'd need to look at calls to the spoke from either the hub, the dataworker, or the atomic weth depositor.

test/cross-chain-adapters/zkSync.ts
```diff
@@ -0,0 +1,538 @@
...
+
+const { MAINNET, ZK_SYNC } = CHAIN_IDs;
+const { USDC, WETH } = TOKEN_SYMBOLS_MAP;
+const l1Weth = WETH.addresses[MAINNET];
+
+let l1Bridge: Contract, l2Bridge: Contract;
+let l2Eth: Contract, l2Weth: Contract;
+let hubPool: Contract, spokePool: Contract;
+
+class zkSyncTestAdapter extends ZKSyncAdapter {
```
james-a-morris, line 26: Nice use of inheritance here +1

james-a-morris (commented):
lgtm - few comments

contracts/MockWETH9.sol
pxrl: Yeah; that would also work. This was just a shortcut to emit events without needing to worry about approvals or balances & transfers. Happy to update it to use a full WETH9 implementation, but I'd prefer to do that as a follow-up PR so that we merge https://github.com/across-protocol/relayer/pull/1524. wdyt?

nicholaspai (approved):

PR #1633, opened 2024-06-20
feat(monitor): Add logs about stuck root bundle relays

nicholaspai:
Even if root bundle relay from L1 to L2 doesn't contain token transfers, we should still warn if something is stuck

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -1,5 +1,5 @@
 import { BalanceAllocator } from "../clients";
-import { spokePoolClientsToProviders } from "../common";
+import { EXPECTED_L1_TO_L2_MESSAGE_TIME, spokePoolClientsToProviders } from "../common";
 import {
   BalanceType,
   BundleAction,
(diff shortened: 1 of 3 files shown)


src/monitor/Monitor.ts
```diff
@@ -569,28 +569,46 @@ export class Monitor {
...
-    }
 
     const allL1Tokens = this.clients.hubPoolClient.getL1Tokens();
     for (const chainId of this.crossChainAdapterSupportedChains) {
+      const gracePeriod = EXPECTED_L1_TO_L2_MESSAGE_TIME[chainId] ?? REBALANCE_FINALIZE_GRACE_PERIOD;
+      // If we're still within the grace period, skip looking for any stuck rebalances.
+      // Again, this would give false negatives for transfers that have been stuck for longer than one bundle if the
+      // current time is within the grace period of last executed bundle. But this is a good trade off for simpler code.
+      const lastFullyExecutedBundleTime = lastFullyExecutedBundle.challengePeriodEndTimestamp;
+      const currentTime = Number(await this.clients.hubPoolClient.hubPool.getCurrentTime());
```
pxrl, line 580:
```suggestion
    const { currentTime } = this.clients.hubPoolClient;

    // If we're still within the grace period, skip looking for any stuck rebalances.
    // Again, this would give false negatives for transfers that have been stuck for longer than one bundle if the
    // current time is within the grace period of last executed bundle. But this is a good trade off for simpler code.
    const lastFullyExecutedBundleTime = lastFullyExecutedBundle.challengePeriodEndTimestamp;
    for (const chainId of this.crossChainAdapterSupportedChains) {
      const gracePeriod = EXPECTED_L1_TO_L2_MESSAGE_TIME[chainId] ?? REBALANCE_FINALIZE_GRACE_PERIOD;
```

pxrl (approved):

PR #1634, opened 2024-06-21
refactor(bridges): Consolidate Optimism custom bridges

pxrl:
This PR internalises the customBridges logic for the OP stack adapter, which simplifies the top-level code in the AdapterManager. Additionally, with the realisation that the DAI bridge ABI looks to be identical to the standard OP stack ERC20 bridge ABI, flush the DAI instance.


dohaki (approved):
LGTM

bmzig (approved):

nicholaspai (approved):
LGTM

pxrl:
I backed out the ABI changes due to the realisation that the DAI bridge is does not completely overlap with the default Optimism bridge - the Optimism bridge also defines support for ETH deposits, so it's not correct to merge them in this way. A better way would be to split out the ERC20 aspects of the ABI and make that common to DAI & general ERC20 deposits, but the benefits are small and it implies more refactoring work.

PR #1635, opened 2024-06-21
feat: disable slow fills for deposits involving lite chains

amateima:
This PR aims to disable slow fills for deposits that need to be filled on a lite chain or originating from a lite chain.

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -931,7 +931,10 @@ export class BundleDataClient {
                 }
 
                 // slow fill requests for deposits from or to lite chains are considered invalid
-                if (v3RelayHashes[relayDataHash].deposit.fromLiteChain || v3RelayHashes[relayDataHash].deposit.toLiteChain) {
+                if (
+                  v3RelayHashes[relayDataHash].deposit.fromLiteChain ||
+                  v3RelayHashes[relayDataHash].deposit.toLiteChain
+                ) {
                   return;
                 }
 
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -930,13 +930,8 @@ export class BundleDataClient {
                   return;
                 }
 
-                // slow fill requests for deposits originating from a lite chain are considered invalid
-                if (v3RelayHashes[relayDataHash].deposit.fromLiteChain) {
-                  return;
-                }
-
-                // slow fills requested on a lite chain are considered invalid
-                if (v3RelayHashes[relayDataHash].deposit.toLiteChain) {
+                // slow fill requests for deposits from or to lite chains are considered invalid
+                if (v3RelayHashes[relayDataHash].deposit.fromLiteChain || v3RelayHashes[relayDataHash].deposit.toLiteChain) {
                   return;
                 }
 
@@ -982,13 +977,8 @@ export class BundleDataClient {
               // historical deposit query is not working as expected.
               assert(this.getRelayHashFromEvent(matchedDeposit) === relayDataHash);
 
-              // slow fill requests for deposits originating from a lite chain are considered invalid
-              if (matchedDeposit.fromLiteChain) {
-                return;
-              }
-
-              // slow fills requested on a lite chain are considered invalid
-              if (matchedDeposit.toLiteChain) {
+              // slow fill requests for deposits from or to lite chains are considered invalid
+              if (matchedDeposit.fromLiteChain || matchedDeposit.toLiteChain) {
                 return;
               }
 
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -981,6 +981,17 @@ export class BundleDataClient {
               // sanity check it here by comparing the full relay hashes. If there's an error here then the
               // historical deposit query is not working as expected.
               assert(this.getRelayHashFromEvent(matchedDeposit) === relayDataHash);
+
+              // slow fill requests for deposits originating from a lite chain are considered invalid
+              if (matchedDeposit.fromLiteChain) {
+                return;
+              }
+
+              // slow fills requested on a lite chain are considered invalid
+              if (matchedDeposit.toLiteChain) {
+                return;
+              }
+
               v3RelayHashes[relayDataHash].deposit = matchedDeposit;
 
               // Note: we don't need to query for a historical fill at this point because a fill
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -930,6 +930,26 @@ export class BundleDataClient {
                   return;
                 }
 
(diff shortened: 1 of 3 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -923,6 +923,26 @@ export class BundleDataClient {
...
+                // slow fills requested on a lite chain are considered invalid
+                if (
+                  this.clients.configStoreClient.isChainLiteChainAtTimestamp(
+                    slowFillRequest.destinationChainId,
+                    v3RelayHashes[relayDataHash].deposit.quoteTimestamp
+                  )
+                ) {
+                  return;
+                }
+
```
amateima, line 945: Not sure if this logic needs to be placed here or before line 903, case in which the `slowFillRequest` key of object `v3RelayHashes[relayDataHash]` would not be populated at all

src/relayer/Relayer.ts
```diff
@@ -564,8 +564,21 @@ export class Relayer {
       return;
     }
 
-    const { hubPoolClient, spokePoolClients, multiCallerClient } = this.clients;
+    const { hubPoolClient, spokePoolClients, multiCallerClient, configStoreClient } = this.clients;
     const { originChainId, destinationChainId, depositId, outputToken } = deposit;
+
+    // don't request slow fill if origin chain is a lite chain
+    if (configStoreClient.isChainLiteChainAtTimestamp(deposit.originChainId, deposit.quoteTimestamp)) {
```
james-a-morris, line 571: We can directly call `deposit.originatesFromLiteChain` - it's flagged in all `DepositV3` objects that originate from a spoke pool

pxrl: @james-a-morris That also means we'll be pushing `originatesFromLiteChain` to Arweave, right? That'd probably warranty a version bump there, unless we suppress it. I'm not sure whether it's actually useful to have it in Arweave so it might be OK to suppress - wdyt?

src/clients/BundleDataClient.ts
nicholaspai: I think it has to follow 910 at least because there we assert the existence of the `deposit` in `v3RelayHashes[relayDataHash]`

src/relayer/Relayer.ts
nicholaspai, line 571: Can we add a debug log here explaining why we're exiting early?

src/relayer/Relayer.ts
```diff
@@ -564,8 +564,21 @@ export class Relayer {
...
+      return;
+    }
+
+    // don't request slow fill if destination is a lite chain
+    if (
+      this.clients.configStoreClient.isChainLiteChainAtTimestamp(deposit.destinationChainId, deposit.quoteTimestamp)
+    ) {
+      return;
```
nicholaspai, line 579: same comment as https://github.com/across-protocol/relayer/pull/1635/files#r1648988787

src/relayer/Relayer.ts
```diff
@@ -566,8 +566,29 @@ export class Relayer {
       return;
     }
 
-    const { hubPoolClient, spokePoolClients, multiCallerClient } = this.clients;
+    const { hubPoolClient, spokePoolClients, multiCallerClient, configStoreClient } = this.clients;
     const { originChainId, destinationChainId, depositId, outputToken } = deposit;
+
+    // don't request slow fill if origin chain is a lite chain
+    if (deposit.fromLiteChain) {
```
nicholaspai, line 573: @james-a-morris @amateima what should we do with merge conflicts between your prs? Should this PR into James' ?

nicholaspai (approved):
This looks great but we need to resolve conflicts with #1629

src/relayer/Relayer.ts
amateima: I was thinking the same, yes

james-a-morris (approved):

src/clients/BundleDataClient.ts
```diff
@@ -930,6 +930,16 @@ export class BundleDataClient {
                   return;
                 }
 
+                // slow fill requests for deposits originating from a lite chain are considered invalid
```
nicholaspai, line 933: We need these same if statements within the if-block at 972

test/Dataworker.loadData.ts
```diff
@@ -1049,6 +1049,76 @@ describe("Dataworker: Load data used in all functions", async function () {
...
+      const data1 = await dataworkerInstance.clients.bundleDataClient.loadData(getDefaultBlockRange(5), {
+        ...spokePoolClients,
+        [originChainId]: spokePoolClient_1,
+        [destinationChainId]: spokePoolClient_2,
+      });
+
+      expect(data1.bundleSlowFillsV3).to.deep.equal({});
+      expect(data1.bundleDepositsV3[originChainId][erc20_1.address].length).to.equal(1);
+    });
+    it("Slow fill request for deposit that isn't eligible for slow fill because destination is lite chain", async function () {
```
nicholaspai, line 1087: We need additional unit tests for slow fill requests where the deposit needs to be fetched from a historical query like https://github.com/across-protocol/relayer/blob/746bda6b77f67e5fe29e32ac41bb0c8b0cde131b/test/Dataworker.loadData.ts#L845

nicholaspai (changes requested):
We are missing a check in the BundleDataClient where the slow fill request matches with an older deposit

test/Dataworker.loadData.ts
amateima: good catch

src/clients/BundleDataClient.ts
```diff
@@ -930,6 +930,16 @@ export class BundleDataClient {
...
+                if (v3RelayHashes[relayDataHash].deposit.fromLiteChain) {
```
nicholaspai, line 934: Should we collapse these two if statements into a single one?

src/clients/BundleDataClient.ts
```diff
@@ -971,6 +981,17 @@ export class BundleDataClient {
               // sanity check it here by comparing the full relay hashes. If there's an error here then the
               // historical deposit query is not working as expected.
               assert(this.getRelayHashFromEvent(matchedDeposit) === relayDataHash);
+
+              // slow fill requests for deposits originating from a lite chain are considered invalid
+              if (matchedDeposit.fromLiteChain) {
```
nicholaspai, line 986: Should we collapse these two if statements?

nicholaspai (approved):
Lgtm but I left some simple nut suggestions

src/clients/BundleDataClient.ts
amateima: Yep. Will do!

src/clients/BundleDataClient.ts
amateima: Yep. Will do!

PR #1636, opened 2024-06-21
fix(AcrossAPIClient): Support liquid-reserves query

pxrl:
The API was recently updated to return a modified definition of maxDeposit, such that it is capped at the largest desirable deposit amount (but not necessarily the largest deposit that Across can support). This change also added a destinationChainId-dependent aspect, which is incompatible with the relayer's existing use of the API where it tries to only query from anywhere -> mainnet to determine limits. This caused flow-on effects in the relayer where it would refuse to make fills that were perfectly OK to fill. This change restores the ability of the relayer to accurately fill based on configured limits.

### DIFF
--- a/src/clients/AcrossAPIClient.ts
+++ b/src/clients/AcrossAPIClient.ts
@@ -143,11 +143,7 @@ export class AcrossApiClient {
       const baseTtl = 300;
       // Apply a random margin to spread expiry over a larger time window.
       const ttl = baseTtl + Math.ceil(_.random(-0.5, 0.5, true) * baseTtl);
-      await redis.set(
-        this.getLimitsCacheKey(l1Tokens),
-        liquidReserves.map((n) => n.toString()).join(","),
-        ttl
-      );
+      await redis.set(this.getLimitsCacheKey(l1Tokens), liquidReserves.map((n) => n.toString()).join(","), ttl);
     }
 
     return liquidReserves;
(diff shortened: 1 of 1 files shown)


src/clients/AcrossAPIClient.ts
```diff
@@ -59,59 +57,24 @@ export class AcrossApiClient {
       throw new Error("HubPoolClient must be updated before AcrossAPIClient");
     }
     const enabledTokens = hubPoolClient.getL1Tokens().map((token) => token.address);
-    const tokensQuery = this.tokensQuery.filter((token) => enabledTokens.includes(token));
+    const tokens = this.tokensQuery.filter((token) => enabledTokens.includes(token));
     this.logger.debug({
       at: "AcrossAPIClient",
-      message: "Querying /limits",
+      message: "Querying /liquidReserves",
```
nicholaspai, line 63:
```suggestion
      message: "Querying /liquid-reserves",
```

src/clients/AcrossAPIClient.ts
```diff
@@ -59,59 +57,24 @@ export class AcrossApiClient {
...
-        this.limits[token] = data[i].maxDeposit;
-      } else {
-        this.logger.debug({
-          at: "AcrossAPIClient",
-          message: "No valid deposit routes for enabled LP token, skipping",
-          token,
-        });
-      }
-    });
+    // /liquidReserves
```
nicholaspai, line 70:
```suggestion
    // /liquid-reserves
```

nicholaspai (approved):
Nice lgtm once endpoint is up!

PR #1639, opened 2024-06-24
feat(lisk): Add constants to src/common

bmzig:
This commit adds default configurations for Lisk. Lisk has essentially the same requirements as other OP Stack chains, so these constants are identical to those of Base/Mode. Notably, Lisk has no Multicall3 deployment (we may need to deploy one ourselves).


src/common/Constants.ts
```diff
@@ -290,7 +299,7 @@ export type Multicall2Call = {
 
 // These are the spokes that can hold both ETH and WETH, so they should be added together when caclulating whether
 // a bundle execution is possible with the funds in the pool.
-export const spokesThatHoldEthAndWeth = [10, 324, 8453, 34443, 59144];
+export const spokesThatHoldEthAndWeth = [10, 324, 1335, 8453, 34443, 59144];
```
pxrl, line 302:
```suggestion
export const spokesThatHoldEthAndWeth = [10, 324, 1135, 8453, 34443, 59144];
```

src/common/Constants.ts
```diff
@@ -317,6 +326,7 @@ export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
   10: ["DAI", "SNX", "BAL", "WETH", "USDC", "POOL", "USDT", "WBTC", "UMA", "ACX"],
   137: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "POOL"],
   324: ["USDC", "USDT", "WETH", "WBTC", "DAI"],
+  1135: ["WETH", "LSK", "USDC", "USDT"],
```
pxrl, line 329: AFAIK USDC isn't available until August, and LSK is still pending in a PR in constants, so we can't use it yet.

(In general we might consider deferring this line, since we also can't resolve token addresses on Lisk yet.)

```suggestion
  1135: ["WETH", "USDT"],
```

src/common/Constants.ts
bmzig: Sorry. Careless mistake. Fixed here: https://github.com/across-protocol/relayer/pull/1639/commits/2f3e078bc64e9c8596cfc2337731575bd0157a02

src/common/Constants.ts
bmzig: Are we not going to support USDC.e? I removed them here: https://github.com/across-protocol/relayer/pull/1639/commits/2f3e078bc64e9c8596cfc2337731575bd0157a02

src/common/ContractAddresses.ts
```diff
@@ -91,6 +91,10 @@ export const CONTRACT_ADDRESSES: {
       address: "0x735aDBbE72226BD52e818E7181953f42E3b0FF21",
       abi: OVM_L1_STANDARD_BRIDGE_ABI,
     },
+    ovmStandardBridge_1135: {
+      address: "0x2658723Bf70c7667De6B25F99fcce13A16D25d08",
```
pxrl, line 95: Confirmed this address here: https://docs.lisk.com/contracts

l1

pxrl (approved):

james-a-morris (approved):
Will we have any testnet values here?

pxrl:
> Will we have any testnet values here?

Good call. @bmzig Lisk testnet is 4202.

src/common/ContractAddresses.ts
```diff
@@ -182,6 +186,19 @@ export const CONTRACT_ADDRESSES: {
...
+    ovmStandardBridge: {
+      address: "0x4200000000000000000000000000000000000010",
+      abi: OVM_L2_STANDARD_BRIDGE_ABI,
+    },
+    weth: {
+      address: "0x4200000000000000000000000000000000000006",
+      abi: WETH_ABI,
+    },
+    eth: {
+      address: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
```
james-a-morris, line 199: OOC is ETH/WETH treated separately?

bmzig: Probably want a confirmation from @nicholaspai here, but it looks like this is an artifact of when ETH was tracked as a token on Optimism and is now just a dead address, which we check for here: https://github.com/across-protocol/relayer/blob/746bda6b77f67e5fe29e32ac41bb0c8b0cde131b/src/dataworker/DataworkerUtils.ts#L568. The balance is zero for our optimism spoke pool.

bmzig: Or @pxrl

pxrl (approved):

src/common/ContractAddresses.ts
pxrl: Yeah, as I recall it's a historical artifact. We've repeated the same pattern on Base and Mode.

PR #1640, opened 2024-06-26
refactor(constants): Use symbolic chain labels

pxrl:
Using symbols ensures that chain IDs can't accidentally be mistyped. While here, flush all Görli references because they're not longer needed. The ordering of entries is unchanged here; in a follow-up PR I will reorder alphabetically.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -12,15 +12,15 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK = 4 * 60 * 60;
 
 // Target ~4 days per chain. Should cover all events needed to construct pending bundle.
 export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
-  1: 28800,
-  10: 172800, // 1 block every 2 seconds after bedrock
-  137: 138240,
-  288: 11520,
-  324: 4 * 24 * 60 * 60,
-  8453: 172800, // Same as Optimism.
-  34443: 172800, // Same as Optimism.
-  42161: 1382400,
-  59144: 115200, // 1 block every 3 seconds
+  [CHAIN_IDs.MAINNET]: 28800,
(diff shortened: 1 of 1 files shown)


src/common/Constants.ts
james-a-morris, line 1: Should we alphabetize this to correspond to your new change? #1641

src/common/Constants.ts
```diff
@@ -261,27 +228,21 @@ export const PROVIDER_CACHE_TTL_MODIFIER = 0.15;
...
-  5: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  280: "0xF9cda624FBC7e059355ce98a31693d299FACd963",
-  420: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  59140: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  80001: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  84531: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  84532: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  421613: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  534351: "0xcA11bde05977b3631167028862bE2a173976CA11",
-  11155111: "0xcA11bde05977b3631167028862bE2a173976CA11",
```
james-a-morris, line 284: Some of these don't line up with the new changes. Did we remove some?

pxrl: All pre-Sepolia testnets were removed since we don't use them anymore (5, 280, 420, 59140, 80001, 84531, 421613).

src/common/Constants.ts
pxrl: Yeah I do want to sort the entries alphabetically but thought it best to split that out into the subsequent PR so as to make it easier to review. Hopefully that's the case :pray:

james-a-morris (approved):

PR #1641, opened 2024-06-26
refactor(constants): Order symbolic chain IDs alphabetically

pxrl:
This establishes a pattern that makes it obvious where new entries should be added. No values were modified as part of this change.


james-a-morris (commented):
This is fine, but we should record this stylization in the README.md

pxrl:
> This is fine, but we should record this stylization in the README.md

We don't really talk about code style in README, but I added a comment to the top of `Constants`. wdyt?

https://github.com/across-protocol/relayer/pull/1641/commits/4539158698628fb72f96119aef7fd06233615c01

src/common/Constants.ts
```diff
@@ -1,5 +1,9 @@
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP, ethers } from "../utils";
 
+/**
+ * Note: When adding new chains, it's preferred to retain the alphabetical ordering of CHAIN_IDs in Object mappings.
```
pxrl, line 4:
```suggestion
 * Note: When adding new chains, it's preferred to retain alphabetical ordering of CHAIN_IDs in Object mappings.
```

james-a-morris (approved):

PR #1643, opened 2024-06-26
feat(AtomicDepositor): update Atomic Depositor with Lisk info

bmzig:
Since Lisk is an OpStack chain, we need to update our Atomic Depositor so that our chain adapters will function properly. I made the small changes on this first commit, and once we review it, I can re-deploy the contract and update this PR with another commit which contains the mainnet deployment information.


james-a-morris (approved):
Could be a good use case for #1548

pxrl (approved):

bmzig:
We will have to approve the atomic depositor address to take Weth before we merge this in.

james-a-morris (approved):
We should look at #1548

bmzig:
> We should look at #1548

We're going to need to update this one again with Blast information. Should we instead just start to push for https://github.com/across-protocol/relayer/pull/1548?

pxrl:
> > We should look at #1548
> 
> We're going to need to update this one again with Blast information. Should we instead just start to push for #1548?

It'd probably be good to get this in in order to test rebalancing to Lisk, and the contract is already deployed so there's no real advantage to discarding it. It definitely makes sense to piggyback on #1548 for Blast though!

bmzig:
> > > We should look at #1548
> > 
> > 
> > We're going to need to update this one again with Blast information. Should we instead just start to push for #1548?
> 
> It'd probably be good to get this in in order to test rebalancing to Lisk, and the contract is already deployed so there's no real advantage to discarding it. It definitely makes sense to piggyback on #1548 for Blast though!

Fair enough.

PR #1644, opened 2024-06-27
fix(ProviderUtils): filter out totalDifficulty from RPC response

bmzig:
We had an issue where Infura's Optimism RPC provider went down, leading to a large number of quorum disputes for `eth_getBlockByNumber` between Alchemy and Quicknode, particularly in the `totalDifficulty` field. This commit should filter out that field so we won't error if RPCs disagree on it.


mrice32 (approved):
LGTM! Great find

PR #1646, opened 2024-06-28
fix(relayer): Revert to USDT fill simulation on Lisk

pxrl:
USDC won't be available on Lisk until later, so revert to USDT until then.


src/clients/ProfitClient.ts
```diff
@@ -593,8 +593,12 @@ export class ProfitClient {
     };
 
     // Pre-fetch total gas costs for relays on enabled chains.
-    const hubToken = TOKEN_SYMBOLS_MAP[testSymbol].addresses[this.hubPoolClient.chainId];
     await sdkUtils.mapAsync(enabledChainIds, async (destinationChainId) => {
+      // USDC is not yet supported on Lisk, so revert to USDT. @todo: Update.
+      testSymbol = [CHAIN_IDs.LISK, CHAIN_IDs.LISK_SEPOLIA].includes(destinationChainId)
+        ? "WETH"
+        : "USDT";
```
pxrl, line 600: Derp

```suggestion
      if (destinationChainId === CHAIN_IDs.LISK) {
        testSymbol = "USDT";
      }
```

src/clients/ProfitClient.ts
```diff
@@ -593,8 +593,12 @@ export class ProfitClient {
     };
 
     // Pre-fetch total gas costs for relays on enabled chains.
-    const hubToken = TOKEN_SYMBOLS_MAP[testSymbol].addresses[this.hubPoolClient.chainId];
     await sdkUtils.mapAsync(enabledChainIds, async (destinationChainId) => {
+      // USDC is not yet supported on Lisk, so revert to USDT. @todo: Update.
+      if (destinationChainId === CHAIN_IDs.LISK) {
+        testSymbol = "USDT";
+      }
```
james-a-morris, line 600: Maybe we can immortalize this in a lookup somewhere? Presumably, we also have to account for this in the testnet?

pxrl: For testnet we only use WETH because it's more ubiquitous, so we should be OK there.

Good call on using a lookup though. I was thinking that we'd only need this for a brief period of time, but we'll have the same issue with Blast, so it does make sure to setup a mapping.

james-a-morris (approved):

PR #1648, opened 2024-07-01
chore: Bump contracts for Lisk defs

pxrl:


pxrl:
One thing to note for this PR: this will import the new SpokePool ABIs. We are still running with the pre-audit deployments in production, but there shouldn't be any side-effects from this because there should be no breaking changes in this update.

james-a-morris (approved):

PR #1653, opened 2024-07-02
fix(APIClient): Dedup L1 tokens

pxrl:
There can be multiple tokens that map to the same L1 token (i.e. USDC, USDC.e, USDbC). Deduplicate this array on instantiation to avoid issuing duplicate requests.


src/clients/AcrossAPIClient.ts
```diff
@@ -35,7 +36,9 @@ export class AcrossApiClient {
     readonly timeout: number = 3000
   ) {
     if (Object.keys(tokensQuery).length === 0) {
-      this.tokensQuery = Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[CHAIN_IDs.MAINNET]);
+      this.tokensQuery = dedupArray(
+        Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[CHAIN_IDs.MAINNET])
```
james-a-morris, line 40: We should normalize the casing here - maybe either checksum case or some upper/lower case variant.

james-a-morris: Otherwise - we could have multiple of the same addresses but just with similar casings

src/clients/AcrossAPIClient.ts
```diff
@@ -37,7 +38,9 @@ export class AcrossApiClient {
     const hubChainId = hubPoolClient.chainId;
     this.endpoint = `https://${hubChainId === CHAIN_IDs.MAINNET ? "app.across.to" : "testnet.across.to"}/api`;
     if (Object.keys(tokensQuery).length === 0) {
-      this.tokensQuery = Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainId]);
+      this.tokensQuery = dedupArray(
+        Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainID])
+      );
```
pxrl, line 43:
```suggestion
      this.tokensQuery = dedupArray(Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainID]));
```

src/clients/AcrossAPIClient.ts
```diff
@@ -37,7 +38,7 @@ export class AcrossApiClient {
     const hubChainId = hubPoolClient.chainId;
     this.endpoint = `https://${hubChainId === CHAIN_IDs.MAINNET ? "app.across.to" : "testnet.across.to"}/api`;
     if (Object.keys(tokensQuery).length === 0) {
-      this.tokensQuery = Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainId]);
+      this.tokensQuery = dedupArray(Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainID]));
```
pxrl, line 41:
```suggestion
      this.tokensQuery = dedupArray(Object.values(TOKEN_SYMBOLS_MAP).map(({ addresses }) => addresses[hubChainId]));
```

nicholaspai (approved):

PR #1657, opened 2024-07-03
refactor(OpStackAdapter): Reuse signers

pxrl:
This shrinks the code a bit and makes it more testnet-friendly. This code will soon be replaced, but the method of retrieving the l2 signer can be reused.


bmzig (approved):

bmzig:
We probably won't even be able to use this on testnet (the conditional won't even trigger since the chainId will not be 10, nor do we have DAI/SNX bridges defined in contractAddresses). This is good too, though, since it will catch those errors caused by passing in incorrect signers to bridges.

james-a-morris (approved):

PR #1660, opened 2024-07-05
feat(blast): add blast finalizer

bmzig:
This should enable finalizations on Blast. We may need to add a special case in the finalizer because of the Blast stablecoin bridge.


src/finalizer/utils/opStack.ts
```diff
@@ -61,7 +61,10 @@ export async function opStackFinalizer(
   // - Don't submit proofs for finalizations older than 1 day
   // - Don't try to withdraw tokens that are not past the 7 day challenge period
   const redis = await getRedisCache(logger);
-  const latestBlockToProve = await getBlockForTimestamp(chainId, getCurrentTime() - 7 * 60 * 60 * 24, undefined, redis);
+  const latestBlockToProve =
+    chainId === CHAIN_IDs.BLAST
+      ? await getBlockForTimestamp(chainId, getCurrentTime() - 14 * 60 * 60 * 24, undefined, redis)
+      : await getBlockForTimestamp(chainId, getCurrentTime() - 7 * 60 * 60 * 24, undefined, redis);
```
pxrl, line 67:
```suggestion
  const challengeWindow = chainId === CHAIN_IDs.BLAST ? 14 : 7; // days
  const latestBlockToProve = await getBlockForTimestamp(chainId, getCurrentTime() - challengeWindow * 3600 * 24, undefined, redis);
```

src/common/Constants.ts
```diff
@@ -368,6 +368,7 @@ export const EXPECTED_L1_TO_L2_MESSAGE_TIME = {
   [CHAIN_IDs.BASE]: 20 * 60,
   [CHAIN_IDs.MODE]: 20 * 60,
   [CHAIN_IDs.BLAST]: 20 * 60,
+
```
pxrl, line 371:
```suggestion
```

pxrl:
@bmzig Is this PR missing the BLAST addresses for the OP class in `src/common/Constants` ?

src/finalizer/utils/opStack.ts
```diff
@@ -61,7 +61,10 @@ export async function opStackFinalizer(
   // - Don't submit proofs for finalizations older than 1 day
   // - Don't try to withdraw tokens that are not past the 7 day challenge period
   const redis = await getRedisCache(logger);
-  const latestBlockToProve = await getBlockForTimestamp(chainId, getCurrentTime() - 7 * 60 * 60 * 24, undefined, redis);
+  const latestBlockToProve =
+    chainId === CHAIN_IDs.BLAST
+      ? await getBlockForTimestamp(chainId, getCurrentTime() - 14 * 60 * 60 * 24, undefined, redis)
```
james-a-morris, line 66: It could be worth setting this in a constructor as a `latestBlockToProve` offset so that we can keep this block of code generic. 

The comment above this becomes outdated with this change btw (line 62)

james-a-morris (commented):
one comment

src/finalizer/utils/opStack.ts
bmzig: Done here: https://github.com/across-protocol/relayer/pull/1660/commits/3488b874af345c5e6f2917de14d2059b5686a064

src/finalizer/utils/opStack.ts
bmzig: I ended up not passing in the value since if I do the logic in the OpStack finalizer, then I can just check the chain ID (since I can safely assume the chain is an OP stack chain). If I were to pass in a `latestBlockToProve` value into the finalizer, I would need to add some extra logic here: https://github.com/across-protocol/relayer/blob/3488b874af345c5e6f2917de14d2059b5686a064/src/finalizer/index.ts#L216 which would probably be messier. Either way, we do plan on making the finalizers generic, so worst case, this won't be in for long.

james-a-morris (commented):
Will this list expand per op stack chain?

src/finalizer/utils/opStack.ts
james-a-morris: Hmm at the very least we should make this a private function or getter variable

james-a-morris (commented):
https://github.com/across-protocol/relayer/pull/1660#discussion_r1674625296

src/finalizer/utils/opStack.ts
bmzig: Done here: https://github.com/across-protocol/relayer/pull/1660/commits/916959b5fa9c2522e4c8032e10a290090b185f38

src/common/Constants.ts
```diff
@@ -385,6 +385,34 @@ export const EXPECTED_L1_TO_L2_MESSAGE_TIME = {
...
+      L2CrossDomainMessenger: "0x4200000000000000000000000000000000000007",
+      L2ToL1MessagePasser: "0x4200000000000000000000000000000000000016",
+      L2StandardBridge: "0x4200000000000000000000000000000000000010",
+      OVM_L1BlockNumber: "0x4200000000000000000000000000000000000013",
+      OVM_L2ToL1MessagePasser: "0x4200000000000000000000000000000000000016",
+      OVM_DeployerWhitelist: "0x4200000000000000000000000000000000000002",
+      OVM_ETH: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
+      OVM_GasPriceOracle: "0x420000000000000000000000000000000000000F",
+      OVM_SequencerFeeVault: "0x4200000000000000000000000000000000000011",
+      WETH: "0x4300000000000000000000000000000000000004",
```
pxrl, line 412: We should probably reuse the existing WETH definition we have for this.

```suggestion
      WETH: TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.BLAST],
```

src/common/Constants.ts
```diff
@@ -385,6 +385,34 @@ export const EXPECTED_L1_TO_L2_MESSAGE_TIME = {
 };
 
 export const OPSTACK_CONTRACT_OVERRIDES = {
```
pxrl, line 387: I was able to verify these contract addresses via the following pages:
https://docs.blast.io/building/contracts
https://docs.optimism.io/chain/addresses#op-mainnet-l2

pxrl (approved):
LGTM, though I think for the WETH address we should probably establish a pattern of inheriting from TOKEN_SYMBOLS_MAP.

src/common/Constants.ts
bmzig: Fixed here: https://github.com/across-protocol/relayer/pull/1660/commits/5f6b127d63a16d20919ecef9305c27ae5da4c31d

src/common/Constants.ts
```diff
@@ -400,17 +400,8 @@ export const OPSTACK_CONTRACT_OVERRIDES = {
...
-      L2StandardBridge: "0x4200000000000000000000000000000000000010",
-      OVM_L1BlockNumber: "0x4200000000000000000000000000000000000013",
-      OVM_L2ToL1MessagePasser: "0x4200000000000000000000000000000000000016",
-      OVM_DeployerWhitelist: "0x4200000000000000000000000000000000000002",
-      OVM_ETH: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
-      OVM_GasPriceOracle: "0x420000000000000000000000000000000000000F",
-      OVM_SequencerFeeVault: "0x4200000000000000000000000000000000000011",
-      WETH: "0x4300000000000000000000000000000000000004",
-      BedrockMessagePasser: "0x4200000000000000000000000000000000000016",
+      ...DEFAULT_L2_CONTRACT_ADDRESSES,
```
pxrl, line 403: Nice!

pxrl:
@bmzig OK for this to go in?

PR #1662, opened 2024-07-08
fix(finalizer): define contract addresses required for Lisk finalizations

bmzig:
The eth-optimism SDK does not have predefined contract addresses for Lisk like it does for all of our other OP Stack chains. This causes finalizations to throw undefined errors. This commit adds in the necessary contract addresses which we then pass into the optimismSDK's `CrossChainMessenger`. We will also need to do something similar for Blast.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -1,4 +1,5 @@
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP, ethers } from "../utils";
+import { DEFAULT_L2_CONTRACT_ADDRESSES } from "@eth-optimism/sdk";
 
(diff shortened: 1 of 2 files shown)


pxrl (approved):
These addresses all check out. Are the 0x0 addresses just specified for type conformance, and was it documented anywhere that they're not required for our use case?

bmzig:
> These addresses all check out. Are the 0x0 addresses just specified for type conformance, and was it documented anywhere that they're not required for our use case?

I essentially just repurposed the existing format found in the eth-optimism sdk, found here: https://www.npmjs.com/package/@eth-optimism/sdk?activeTab=code

PR #1664, opened 2024-07-09
feat(AtomicDepositor): update atomic depositor with Blast

bmzig:
Updates the atomic depositor contract with the Blast standard bridge.


james-a-morris (approved):
lgtm - we should look into #1548

bmzig:
> lgtm - we should look into #1548

I agree. We are only doing this for the time being since the upgrade should probably come independently from adding a new chain. Ideally, we can get that in right after Blast.

pxrl (approved):

pxrl:
Tested OK w/ a test relayer inventory config.

PR #1666, opened 2024-07-10
improve(relayer): Avoid friendly fire when new chains are added

pxrl:
The relayer currently has an annoying constraint that it needs to be updated each time a new chain is added to the config store. This is because the post-update configuration validation is unnecesarily strict. In cases where the relayer has a custom set of origin and destination chains, it actually doesn't make sense to require that configuration is defined for any new chain that has been added.

So:
- When the relayer does not rely on a new chain, don't verify configuration for that chain.
- When the relayer does rely on a new chain, warn about missing config and revert to a conservative default of 5k blocks for eth_getLogs() calls. The warnings should prompt the operator to update their bot or define the necessary config in their environment.

### DIFF
--- a/src/common/ClientHelper.ts
+++ b/src/common/ClientHelper.ts
@@ -357,9 +357,9 @@ export async function constructClients(
 
 // @dev The HubPoolClient is dependent on the state of the ConfigStoreClient,
 //      so update the ConfigStoreClient first.
-export async function updateClients(clients: Clients, config: CommonConfig): Promise<void> {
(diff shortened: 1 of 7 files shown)


nicholaspai (approved):
Do you think this will cause us to ignore ever adding the chain ids to the constant values?

pxrl:
> Do you think this will cause us to ignore ever adding the chain ids to the constant values?

In practice no, since it still blows up for the Dataworker and Monitor personalities when there is missing chain config. The main effect here is that it allows the relayer to live with less-strict config requirements _if_ the operator has opted-in to a specific set of origin and destination chains.

PR #1673, opened 2024-07-11
chore: bump v20

james-a-morris:
This PR immortalizes v20 as the new standard for the relayer repository. It will throw if the version is not set to at least v20 and includes an `.nvmrc` entry for easily using the Node Version Manager to set to `lts/iron`

closes ACX-1819


pxrl (approved):
Nice! Does this mean we're OK with v20 in prod now?

PR #1675, opened 2024-07-11
improve(Constants): Revise OP stack priority fee multiplier

pxrl:
A 1x multiplier inherits the default 1.5 Gwei pricing from ethers and causes OP stack fills to be expensive.


pxrl:
nb. This will also automatically roll out to the executor, which is currently paying > $1 to execute relayer refund leaves on Optimism and Base. That amounts to something in the range of $600 - 900/month per chain, assuming 1 execution per hour.

https://optimistic.etherscan.io/tx/0x7f7c9d1739c2c1cd2b83e39d0157045d7cb2b72145d02da938d86c39af549714

mrice32 (approved):

PR #1676, opened 2024-07-12
feat(relayer): introduce tryMulticall into relayer

bmzig:
This adds the option to set an environment variable `RELAYER_TRY_MULTICALL_CHAINS` to an array of chain ids that, if set, will call `tryMulticall` instead of `multicall` on that chain's corresponding spoke pool.


bmzig:
I'm running the relayer in sim mode and will remove the do not merge tag once I confirm that it is working as intended.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +513,146 @@ export class MultiCallerClient {
...
+  protected override async _executeTxnQueue(
+    chainId: number,
+    txns: AugmentedTransaction[] = [],
+    valueTxns: AugmentedTransaction[] = [],
+    simulate = false
+  ): Promise<TransactionResponse[]> {
+    const nTxns = txns.length + valueTxns.length;
+    if (nTxns === 0) {
+      return [];
+    }
```
pxrl, line 541: I think it's fine to assert that valueTxns.length === 0, and just otherwise ignore the possibility for them to be passed in. I don't see any scenario where we'd actually need value transactions for making fills in the relayer :+1:

pxrl: (if we assume there are no value transactions then I think a lot of the subsequent code should just collapse on itself)

bmzig: Done here: https://github.com/across-protocol/relayer/pull/1676/commits/34963d84081202df3bab1a757eba9f74a63c26da

bmzig:
I can confirm that this will properly use `tryMulticall` for all `RELAYER_TRY_MULTICALL_CHAINS`. I also can confirm that this properly checks whether simulations fail, and if they do fail, will remove them from the transaction queue and log the results. The only part I am not getting coverage in is whether it will properly rebuild the transactions. If we do want to get this merged (this won't be activated unless the environment variable is set, either way), I added a safety assert here https://github.com/across-protocol/relayer/pull/1676/commits/65deafafe660187fbcbf90d9ccdb176701e598a5 so that we will not potentially delete funds (even without the assert we should _theoretically_ be OK since none of the multicall txns are value txns, but it's good to have to potentially surface some bug).

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+    // First try to simulate the transaction as a batch. If the full batch succeeded, then we don't
+    // need to simulate transactions individually. If the batch failed, then we need to
+    // simulate the transactions individually and pick out the successful ones.
+    const bundledTxns = await this.buildMultiCallBundles(txns, this.chunkSize[chainId]);
+    const bundledSimResults = await this.txnClient.simulate(bundledTxns);
+
+    bundledSimResults.forEach(({ succeed, transaction, reason, data }, idx) => {
+      // TryMulticall _should_ always succeed, but either way, we need the return data to be defined so that we can properly
+      // filter out the transactions which failed.
+      if (succeed && isDefined(data.length)) {
```
bmzig, line 571:
```suggestion
      if (succeed && isDefined(data?.length)) {
```

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
 }
+
+export class TryMulticallClient extends MultiCallerClient {
+  constructor(
+    readonly logger: winston.Logger,
+    readonly chunkSize: { [chainId: number]: number } = {},
+    readonly baseSigner?: Signer
+  ) {
+    super(logger, chunkSize, baseSigner);
+  }
```
pxrl, line 525: Given that we don't do anything extra here, could we potentially remove this? I think it should work OK.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+        data,
+      } as RawTransaction;
+    };
+
+    const txnRequestsToSubmit: AugmentedTransaction[] = [];
+    const txnCalldataToRebuild: RawTransaction[] = [];
+
+    // First try to simulate the transaction as a batch. If the full batch succeeded, then we don't
+    // need to simulate transactions individually. If the batch failed, then we need to
+    // simulate the transactions individually and pick out the successful ones.
```
pxrl, line 564: Is this comment accidentally inherited from the existing code? As I understand it, we simulate once and then drop any transactions from the bundle that failed, but there's never any subsequent simulation of individual transactions.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+    const bundledTxns = await this.buildMultiCallBundles(txns, this.chunkSize[chainId]);
+    const bundledSimResults = await this.txnClient.simulate(bundledTxns);
+
+    bundledSimResults.forEach(({ succeed, transaction, reason, data }, idx) => {
+      // TryMulticall _should_ always succeed, but either way, we need the return data to be defined so that we can properly
+      // filter out the transactions which failed.
+      if (succeed && isDefined(data?.length)) {
```
pxrl, line 571: Have you considered inverting this logic, such that it just returns on `!succeed || !isDefined(data?.length)` ? I think that should allow a level of indendation to be removed from this closure.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+        if (succeededTxnCalldata.length !== data.length) {
+          txnCalldataToRebuild.push(
+            ...succeededTxnCalldata.map((calldata) => buildRawTransaction(transaction.contract, calldata))
+          );
+          this.logger.warn({
+            at: "tryMulticallClient#executeChainTxnQueue",
+            message: `Some calls in ${networkName} transaction batch failed!`,
+            batchTxn: { ...transaction, contract: transaction.contract.address },
+            reason,
+          });
```
pxrl, line 594: `warn` will be quite visible in the monitoring channels, but it's not super interesting because this is a condition that we expect to see a lot, and there's no real impact of these simulation failures occurring. I'm not 100% sure whether we need this log message at all, but if we keep it then I think this could safely be a `debug`.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+          // Otherwise, none of the transactions failed, so we can add the bundle to txnRequestsToSubmit.
+          bundledTxns[idx].gasLimit = transaction.gasLimit;
+          txnRequestsToSubmit.push(bundledTxns[idx]);
+
+          // If txn succeeded or the revert reason is known to be benign, then log at debug level.
+          this.logger.debug({
+            at: "tryMulticallClient#executeChainTxnQueue",
+            message: `Successfully simulated ${networkName} transaction batch!`,
+            batchTxn: { ...transaction, contract: transaction.contract.address },
+            reason,
```
pxrl, line 605: Do we expect `reason` to be defined here?

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+      const txnChunks = lodash.chunk(txnCalldataToRebuild, this.chunkSize[chainId] ?? DEFAULT_MULTICALL_CHUNK_SIZE);
+      const rebuildTryMulticall = (txns: RawTransaction[]) => {
+        const mrkdwn: string[] = [];
+        const contract = txns[0].contract;
+        txns.forEach((txn, idx) => {
+          mrkdwn.push(`\n *txn. ${idx + 1}:* ${txn.data ?? "No calldata"}`);
+          if (txn.contract !== contract) {
+            throw new Error(
+              `There was an error when rebuilding a tryMulticall Bundle. Expected to send txn to ${txn.contract} but was sending txn to ${contract}`
+            );
```
pxrl, line 625: We can probably assume we'd never hit this scenario in production, so in the interest of keeping the code small, I'd consider just squashing this to a simple `assert(a === b)` here.

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,146 @@ export class MultiCallerClient {
...
+          }
+        });
+        return {
+          chainId,
+          contract,
+          method: "tryMulticall",
+          args: [...txns.map((txn) => txn.data)],
+          message: "Across tryMulticall transaction",
+          mrkdwn: mrkdwn.join(""),
+        } as AugmentedTransaction;
```
pxrl, line 635: Can we get away without this type assertion? I _think_ it should be OK, but am not 100% sure.

src/relayer/Relayer.ts
```diff
@@ -754,7 +756,10 @@ export class Relayer {
       );
       await this.evaluateFills(unfilledDeposits, lpFees, maxBlockNumbers, sendSlowRelays);
 
-      if (multiCallerClient.getQueuedTransactions(destinationChainId).length > 0) {
+      const pendingTxnAmount = this.config.tryMulticallChains.includes(destinationChainId)
+        ? tryMulticallClient.getQueuedTransactions(destinationChainId).length
+        : multiCallerClient.getQueuedTransactions(destinationChainId).length;
+      if (pendingTxnAmount > 0) {
```
pxrl, line 762: I was confused about `Amount` because it somehow made me think of "value txns" (i.e. txns w/ msg.value > 0). Thoughts on renaming it?

```suggestion
      const pendingTxns = this.config.tryMulticallChains.includes(destinationChainId)
        ? tryMulticallClient.getQueuedTransactions(destinationChainId).length
        : multiCallerClient.getQueuedTransactions(destinationChainId).length;
      if (pendingTxns > 0) {
```

src/relayer/Relayer.ts
```diff
@@ -872,6 +880,9 @@ export class Relayer {
     const mrkdwn = this.constructRelayFilledMrkdwn(deposit, repaymentChainId, realizedLpFeePct);
     const contract = spokePoolClients[deposit.destinationChainId].spokePool;
     const chainId = deposit.destinationChainId;
+    const multiCallerClient = this.config.tryMulticallChains.includes(chainId)
+      ? this.clients.tryMulticallClient
+      : this.clients.multiCallerClient;
```
pxrl, line 885: This is a pattern of usage that repeats a few times; could it make sense to factor it out to a small method on this class? i.e. `Relayer.getMulticaller()` ?

pxrl (commented):
A couple of questions to check on; this is looking good - I'm excited to see it in prod!

src/clients/MultiCallerClient.ts
bmzig: We can. https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: I accidentally left this from when the plan was to also accomodate normal `multicalls`. The comment is wrong. I changed it here: https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: Good call. https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: I was setting it to warn it out of convention since it is an error in the current multicaller. With that being said, I agree, and ended up changing it to debug. https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: Yeah, we can. https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/relayer/Relayer.ts
bmzig: Done here: https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/relayer/Relayer.ts
bmzig: Good call: https://github.com/across-protocol/relayer/pull/1676/commits/e58634681f71aae14b79402c2bee084e6f7c7fb0

src/clients/MultiCallerClient.ts
bmzig: No it shouldn't be defined. I removed it https://github.com/across-protocol/relayer/pull/1676/commits/2f8c4d0cb9956352dfa5315b3bcc786108b4db43

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,138 @@ export class MultiCallerClient {
...
+  }
+
+  protected override async _executeTxnQueue(
+    chainId: number,
+    txns: AugmentedTransaction[] = [],
+    _valueTxns: AugmentedTransaction[] = [],
+    simulate = false
+  ): Promise<TransactionResponse[]> {
+    // Lint Appeasement
+    _valueTxns;
```
pxrl, line 532: Just to be on the safe side, because this would indicate a misuse of the class.

```suggestion
    assert(_valueTxns.length === 0);
```

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,138 @@ export class MultiCallerClient {
...
+      at: "TryMulticallClient#executeTxnQueue",
+      message: `${simulate ? "Simulating" : "Executing"} ${nTxns} transaction(s) on ${networkName}.`,
+    });
+
+    const buildRawTransaction = (contract: Contract, data: string) => {
+      return {
+        contract,
+        data,
+      } as RawTransaction;
+    };
```
pxrl, line 549: idk whether this closure is even needed, but it should be possible to squash it pretty severely whilst still retaining type assurances.

```suggestion
    const buildRawTransaction = (contract: Contract, data: string): RawTransaction => {{ contract, data });
```

src/clients/MultiCallerClient.ts
```diff
@@ -513,3 +514,138 @@ export class MultiCallerClient {
...
+      const succeededTxnCalldata: string[] = [];
+      // We also need to assert that the number of calls which returned data matches the number of calls made in the transaction.
+      assert(transaction.args.length === data.length);
+      // Go through each transaction result of the txns batched in tryMulticall.
+      // If a transaction succeeded, we take note by adding it to succeededTxnCalldata.
+      data.forEach(({ success }, idx) => {
+        if (success) {
+          succeededTxnCalldata.push(transaction.args[idx]);
+        }
+      });
```
pxrl, line 579: Could something like this work?

```suggestion
      // If we make it here, then tryMulticall was simulated and there is a defined return data.
      // Verify that the number of calls which returned data matches the number of calls made in the transaction.
      assert(transaction.args.length === data.length);

      // Filter the calldata array by whether it succeeded in tryMulticall().
      const succeededTxnCalldata = transaction.args.filter((callData, idx) => data[idx].success);
```

src/relayer/Relayer.ts
```diff
@@ -1234,4 +1237,10 @@ export class Relayer {
 
     return msg;
   }
+
+  private getMulticaller(chainId: number) {
```
pxrl, line 1241: Might want a type annotation for this?

```suggestion
  private getMulticaller(chainId: number): MultiCallerClient {
```

test/Relayer.BasicFill.ts
```diff
@@ -156,11 +158,13 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
         multiCallerClient,
         inventoryClient: new MockInventoryClient(null, null, null, null, null, hubPoolClient),
         acrossApiClient: new AcrossApiClient(spyLogger, hubPoolClient, chainIds),
+        tryMulticallClient,
       },
       {
         relayerTokens: [],
         minDepositConfirmations: defaultMinDepositConfirmations,
         sendingRelaysEnabled: true,
+        tryMulticallChains: [],
```
pxrl, line 167: This is a common annoyance in test w/ the RelayerConfig object. One way I've found around it is to just make the application code default to something empty (i.e. []) if the `config.whatever` member is undefined. That at least saves from having to add a bunch of boilerplate to the tests. It'd be nice to figure out something better for this in the long run, such that we don't have to have these bodges in test.

pxrl (approved):
LGTM! I left a few comments, they're mostly just suggestions/observations and can be implemented in follow-up commits. It'll be great to see this go in!

pxrl (approved):

bmzig:
The only reason why I haven't merged this yet is because I am yet to get a scenario where there are some transactions in the multicall bundle which fail and some which succeed. It should be able to pick out the ones which succeed and just rebuild another bundle, but I haven't confirmed this in test yet.

pxrl:
> The only reason why I haven't merged this yet is because I am yet to get a scenario where there are some transactions in the multicall bundle which fail and some which succeed. It should be able to pick out the ones which succeed and just rebuild another bundle, but I haven't confirmed this in test yet.

I think you can test this by removing the `exclusiveRelayer` checks and then submitting a deposit w/ exclusiveRelayer 0x00..01. The relayer should then try to simulate a fill and it will fail. You could complement this with another deposit where the relayer's own address is given exclusivity.

bmzig:
> > The only reason why I haven't merged this yet is because I am yet to get a scenario where there are some transactions in the multicall bundle which fail and some which succeed. It should be able to pick out the ones which succeed and just rebuild another bundle, but I haven't confirmed this in test yet.
> 
> I think you can test this by removing the `exclusiveRelayer` checks and then submitting a deposit w/ exclusiveRelayer 0x00..01. The relayer should then try to simulate a fill and it will fail. You could complement this with another deposit where the relayer's own address is given exclusivity.

There was a problem with how the transaction arguments were set, but that has been fixed. Should be good to go now.

pxrl (approved):

PR #1679, opened 2024-07-15
fix(relayer): Ignore HubPool liquidity on Lite deposits

pxrl:
The relayer doesn't need to respect HubPool limits on these deposits.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -203,7 +203,7 @@ export class Relayer {
     // The relayer should *not* be filling deposits that the HubPool doesn't have liquidity for otherwise the relayer's
     // refund will be stuck for potentially 7 days. Note: Filter for supported tokens first, since the relayer only
     // queries for limits on supported tokens.
-    if (!ignoreLimits) {
+    if (!ignoreLimits && !deposit.fromLiteChain) {
       const { inputAmount } = deposit;
       const limit = acrossApiClient.getLimit(originChainId, l1Token.address);
       if (acrossApiClient.updatedLimits && inputAmount.gt(limit)) {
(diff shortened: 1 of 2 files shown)


src/relayer/Relayer.ts
```diff
@@ -203,7 +203,7 @@ export class Relayer {
     // The relayer should *not* be filling deposits that the HubPool doesn't have liquidity for otherwise the relayer's
     // refund will be stuck for potentially 7 days. Note: Filter for supported tokens first, since the relayer only
     // queries for limits on supported tokens.
-    if (!ignoreLimits) {
+    if (!ignoreLimits && !deposit.fromLiteChain) {
```
pxrl, line 206: nb. These is one very low-probability scenario where this could be an issue. Relayers are permitted to take repayment on Lisk for non-Lisk deposits & fills, so there are occasions where HubPool funds could need to be sent to Lisk. If the HubPool is overcommitted at that point then this could cause a stuck leaf. However, the probability of relayers voluntarily taking repayment on Lite chains (and coinciding with a low-liquidity HubPool situation) seems low enough that we can live with this theoretical risk.

pxrl: Also to add; this is specially to permit LSK to be bridged with ~0 HubPool liquidity. There's 0 risk of an overcommitment happening in that case because the only supported chains are mainnet and Lisk.

This code could be restricted to the LSK token...

james-a-morris (approved):
fair call +1

src/relayer/Relayer.ts
nicholaspai: can we add a comment about this, perhaps in https://github.com/across-protocol/relayer/blob/918f4a7911b311dfa3eb6ceecd7b98d99f9a7fe2/src/common/Constants.ts#L372 that we should be very cautious adding Lite chains to the list of chains we consider for repayment besides origin/destination?

nicholaspai (approved):
one comment

dohaki (approved):

src/relayer/Relayer.ts
pxrl: Added: https://github.com/across-protocol/relayer/pull/1679/commits/65f5049d7307e604ae63473ac2fb75c0dffa2b2c

PR #1680, opened 2024-07-15
feat(InventoryClient): Consider repayment amount in choosing repayment chain choice

nicholaspai:
The current algorithm for deciding which potential repayment chain choice to use does not take into account the `inputAmount` when deciding whether a chain's balance is under or over allocated.

This can lead to taking repayment on some chains that are currently under allocated but the fill is very large and so we swing very extreme to the over allocation status following the repayment


src/clients/InventoryClient.ts
```diff
@@ -490,32 +490,32 @@ export class InventoryClient {
...
       // Add upcoming refunds:
       chainVirtualBalanceWithShortfallPostRelay = chainVirtualBalanceWithShortfallPostRelay.add(
         totalRefundsPerChain[chainId] ?? bnZero
       );
       // To correctly compute the allocation % for this destination chain, we need to add all upcoming refunds for the
       // equivalents of l1Token on all chains.
-      cumulativeVirtualBalanceWithShortfall = cumulativeVirtualBalanceWithShortfall.add(cumulativeRefunds);
-      const cumulativeVirtualBalanceWithShortfallPostRelay = cumulativeVirtualBalanceWithShortfall.sub(outputAmount);
+      const cumulativeVirtualBalanceWithShortfallPostRefunds =
+        cumulativeVirtualBalanceWithShortfall.add(cumulativeRefunds);
```
bmzig, line 512: I think we're missing something like `.add(inputAmount).sub(outputAmount)` here? We're adding the delta between input amount and output amount into the chain balance, but it's not accounted for in the total balance.

nicholaspai: true

bmzig: Yeah that's fair. I could see it getting less and less accurate the more continuous runtime the fast relayer has, but I do agree it likely won't cause a repayment error in practice.

bmzig (approved):

src/clients/InventoryClient.ts
nicholaspai: the continuous shouldn't matter that much since the relayer fee (e.g. input - output amount) is only ignored for this specific relay. On the next relayer run, the relayer's `balanceOf` will get refreshed and account for the LP fee

PR #1683, opened 2024-07-15
improve(InventoryClient): 1x overage on lite chains

pxrl:
Lite chains will have a tendency to creep over target and require manual rebalancing. Try to limit this by clamping the default overage buffer on lite chains to 1x, instead of the existing default 1.5x.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -476,15 +476,17 @@ export class InventoryClient {
       chainsToEvaluate.push(originChainId);
     }
 
+    const liteChainIds = this.hubPoolClient.configStoreClient.getLiteChainIdIndicesForBlock();
+
     const eligibleRefundChains: number[] = [];
     // At this point, all chains to evaluate have defined token configs and are sorted in order of
     // highest priority to take repayment on, assuming the chain is under-allocated.
-    for (const _chain of chainsToEvaluate) {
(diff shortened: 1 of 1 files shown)


src/clients/InventoryClient.ts
```diff
@@ -514,18 +516,19 @@ export class InventoryClient {
...
       assert(
         isDefined(tokenConfig),
-        `No ${outputToken} tokenConfig in the Inventory Config for ${l1Token} on ${_chain} with a repaymentToken of ${repaymentToken}.`
+        `No ${outputToken} tokenConfig for ${l1Token} on ${chainId} with a repaymentToken ${repaymentToken}.`
       );
-      const thresholdPct = toBN(tokenConfig.targetPct)
-        .mul(tokenConfig.targetOverageBuffer ?? toBNWei("1"))
-        .div(fixedPointAdjustment);
+      const targetOverage =
+        tokenConfig.targetOverageBuffer ?? liteChainIds.includes(chainId) ? toBNWei("1") : toBNWei("1.5");
```
nicholaspai, line 525: can you add a comment justifying why we are more sensitive on Lite chains?

nicholaspai (approved):
I think I agree with this but I need to think on it more

src/clients/InventoryClient.ts
pxrl: Additionally, taking voluntary repayment on a lite chain is actually a net cost to the dataworker because it requires tokens to be bridged to the lite chain, so it causes higher gas consumption for the deposit transactions on mainnet.

PR #1684, opened 2024-07-16
feat(Scroll): add scroll constants

bmzig:
This commit adds basic dataworker/relayer constants for Scroll. For the most part, it has similar finalization requirements to Linea (same block time/soft finality). We should consider using https://github.com/across-protocol/relayer/pull/1524 so that we do not need to build a new Scroll adapter.


src/common/ContractAddresses.ts
```diff
@@ -292,6 +298,16 @@ export const CONTRACT_ADDRESSES: {
       address: "0x0000000000000000000000000000000000000000",
     },
   },
+  534352: {
+    scrollGatewayRouter: {
+      address: "0x4C0926FF5252A435FD19e10ED15e5a249Ba19d79",
+      abi: SCROLL_GATEWAY_ROUTER_L2_ABI,
+    },
+    weth: {
+      address: "0x5300000000000000000000000000000000000004",
```
james-a-morris, line 307: Interesting address choice

src/common/Constants.ts
```diff
@@ -27,7 +27,8 @@ export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
   [CHAIN_IDs.MODE]: 172800, // Same as Optimism.
   [CHAIN_IDs.OPTIMISM]: 172800, // 1 block every 2 seconds after bedrock
   [CHAIN_IDs.POLYGON]: 138240,
-  [CHAIN_IDs.ZK_SYNC]: 4 * 24 * 60 * 60,
+  [CHAIN_IDs.SCROLL]: 259200, // 3 * 24 * 60 * 60,
```
james-a-morris, line 30: Is this comment accurate? Scroll seems to have 3 second block periods so 4 days worth of data @ 3s/block would mean 115200 blocks are needed.

bmzig: You are correct. Fixed here: https://github.com/across-protocol/relayer/pull/1684/commits/6aa9ec2b153433b934b6df01b7d8c8951f1e2c95

src/common/Constants.ts
```diff
@@ -157,6 +163,7 @@ export const DEFAULT_CHAIN_MULTICALL_CHUNK_SIZE: { [chainId: number]: number } =
   [CHAIN_IDs.LISK]: 75,
   [CHAIN_IDs.MODE]: 75,
   [CHAIN_IDs.OPTIMISM]: 75,
+  [CHAIN_IDs.LINEA]: 50,
```
pxrl, line 166: `CHAIN_IDs.SCROLL` ?

bmzig: Sometimes the `u` is just clicked instead of the `i`. Fixed here: https://github.com/across-protocol/relayer/pull/1684/commits/ed19ec27c88d4be13c7dbb95d2a3ed935fa459c2

src/common/ContractAddresses.ts
pxrl: I just merged a PR that imports the WETH definitions from the constants repo, so we should be able to drop this now and just inherit from TOKEN_SYMBOLS_MAP :+1: 

https://github.com/across-protocol/relayer/commit/c2f847c90b024d6970b0306654e68968479348ab

You can see we have the same address in constants, fwiw:
scroll

pxrl (approved):
LGTM pending removal of the WETH address :+1:

src/common/ContractAddresses.ts
bmzig: Removed it here: https://github.com/across-protocol/relayer/pull/1684/commits/ba19d9dfe28381ebf40a5ec08f3f4dc4db80762c

PR #1686, opened 2024-07-16
refactor(Constants): Drop per-chain multicall chunk defaults

pxrl:
Only few chains can support chunk sizes of >50 in practice, and the benefits of chunking so many are very limited (especially post-4844). However, large chunk sizes have caused pain during times of high network load (including causing stuck transactions on Linea), and they impose an overhead when setting defaults for new chains. Instead, just revert to a default of 50 and avoid having to mess with this in future.


nicholaspai (approved):
+1

PR #1687, opened 2024-07-17
feat(relayer): Track origin chain fill commitments

pxrl:
This change strengthens the relayer's ability to limit its commitment to an origin chain in accordance with its configured deposit confirmation requirements. On each update, it uses its deposit confirmations configuration to look back at its previous fills and identify how much it has filled for each origin chain. Fill amounts are summed in USD and used to determine the residual amount that can be filled within that configuration, or alternatively, whether an overcommitment to an origin chain has occurred. As each fill (and corresponding deposit) ages it scales progressively through the deposit confirmation requirements and allows for additional fills to be made.

In case of an overcommitment, the relayer will simply back off for that origin chain until the total amount ages enough such that the risk of a re-org is no longer worsened by making additional fills. Because this change is more strict regarding origin chain commitments, it's likely that any relayers of size would need to increase their deposit confirmation requirements in order to avoid deferring a lot of fills.


pxrl:
nb. This is dependent upon an SDK bump; PR here: https://github.com/across-protocol/sdk/pull/686

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,75 @@ export class Relayer {
...
+    let toBlock = originSpoke.latestBlockSearched;
+
+    // For each deposit confirmation tier (lookback), sum all outstanding commitments back to head.
+    const limits = Object.values(mdcs).map(({ minConfirmations, usdThreshold }) => {
+      const fromBlock = Math.max(toBlock - minConfirmations, originSpoke.deploymentBlock);
+      const commitment = this.computeOriginChainCommitment(chainId, fromBlock, toBlock);
+
+      totalCommitment = totalCommitment.add(commitment);
+      const limit = usdThreshold.sub(totalCommitment);
+      toBlock = fromBlock - 1; // Shuffle the range for the next loop.
```
nicholaspai, line 315: why modify toBlock?

src/relayer/Relayer.ts
```diff
@@ -375,11 +445,27 @@ export class Relayer {
       );
       const { relayerFeePct, gasCost, gasLimit: _gasLimit, lpFeePct: realizedLpFeePct } = repaymentChainProfitability;
       if (isDefined(repaymentChainId)) {
+        const fillLimits = this.fillLimits[originChainId];
```
nicholaspai, line 448: nit do you think we should be storing `fillLimits` in the class state or can we load it as neccessary within this function? It seems you only use `fillLimits` here

src/relayer/Relayer.ts
```diff
@@ -375,11 +445,27 @@ export class Relayer {
...
+          return;
+        }
+
         const gasLimit = isMessageEmpty(resolveDepositMessage(deposit)) ? undefined : _gasLimit;
         this.fillRelay(deposit, repaymentChainId, realizedLpFeePct, gasLimit);
 
-        // Update local balance to account for the enqueued fill.
+        // Update local balances & limits to account for the enqueued fill.
         tokenClient.decrementLocalBalance(destinationChainId, deposit.outputToken, deposit.outputAmount);
+        fillLimit.limit = fillLimit.limit.sub(fillAmountUsd);
```
nicholaspai, line 468: don't you need to reduce the limit for all MDC's that are >= the limit breached?

nicholaspai (commented):
Directionally agree with the motivation for this PR

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,75 @@ export class Relayer {
     return true;
   }
 
+  computeOriginChainCommitment(chainId: number, fromBlock: number, toBlock: number): BigNumber {
```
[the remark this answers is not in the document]
pxrl: `Relayer.computeOriginChainLimits()` is used to produce the commitment between blocks specified by the deposit confirmation configuration. This method is wrapped by `Relayer.computeOriginChainCommitment()` to iterately determine the outstanding commitment for each tier of deposit confirmations, and each progressive tier inherits the lower tiers. This avoids recomputing the outstanding commitments at each tier.

src/relayer/Relayer.ts
pxrl: This method iteratively computes the origin chain in chunks determined by the deposit confirmation config, so it's intentionally modifying `toBlock` to avoid recomputing the committed amount over the same block ranges. But the cumulative commitment is inherited at each incrementing tier.

src/relayer/Relayer.ts
pxrl: Yeah; this is a good point. Each tier above this threshold is also affected.

pxrl: Should be fixed by this commit: https://github.com/across-protocol/relayer/pull/1687/commits/ec9c2ac56b45ef8a2b4903d3f0ccbc1c6d0794c0

src/relayer/Relayer.ts
nicholaspai: This seems a bit hacky no? Is there a better way to construct this waterfall structure?

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,76 @@ export class Relayer {
...
+
+  computeOriginChainLimits(chainId: number): { fromBlock: number; limit: BigNumber }[] {
+    const mdcs = this.config.minDepositConfirmations[chainId];
+    const originSpoke = this.clients.spokePoolClients[chainId];
+
+    let totalCommitment = bnZero;
+    let toBlock = originSpoke.latestBlockSearched;
+
+    // For each deposit confirmation tier (lookback), sum all outstanding commitments back to head.
+    const limits = Object.values(mdcs).map(({ minConfirmations, usdThreshold }) => {
```
nicholaspai, line 309: Are mdcs guaranteed to be sorted in ascending order?

src/relayer/Relayer.ts
nicholaspai: Oh ok. I see now. I this algo was a bit dense to me at first

nicholaspai (commented):
Will review again later

src/relayer/Relayer.ts
pxrl: They're sorted by ascending USD threshold when the config is read: https://github.com/across-protocol/relayer/blob/0b64c1897bc735ea507ea5ea0235b96ceb43cbda/src/relayer/RelayerConfig.ts#L268

There's currently no guarantee that the number of deposit confirmations are similarly incrementing with each successive usd threshold. We probably want to enforce that now, because this change will otherwise lead to counter-intuitive behaviour for anyone who has that configured for filtering.

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,96 @@ export class Relayer {
...
+      toBlock = fromBlock - 1; // Shuffle the range for the next loop.
+
+      return { fromBlock, limit };
+    });
+
+    // Warn on the highest overcommitment, if any.
+    for (let i = limits.length - 1; i >= 0; --i) {
+      const { limit, fromBlock } = limits[i];
+      if (limit.lt(bnZero)) {
+        this.logger.warn({
```
pxrl, line 339: The rebalancer currently uses MDC config to avoid rebalancing until unfilled deposits are at least a certain age, so `warn` might be a bit heavy-handed here. `info` might be more appropriate, but will still reliably produce a lot of logs. I think I'll make the log level conditional upon `config.sendingRelaysEnabled` being set.

pxrl: Implemented: https://github.com/across-protocol/relayer/pull/1687/commits/1eadac3bd97f7ab70bd94357ee6915d4017249f7

src/relayer/Relayer.ts
pxrl: Implemented: https://github.com/across-protocol/relayer/pull/1687/commits/089665ef6a95bf9e563e951daa316d4724649ed7

src/relayer/Relayer.ts
```diff
@@ -374,14 +466,34 @@ export class Relayer {
...
+            message: `Skipping ${originChain} deposit ${depositId} due to anticipated origin chain overcommitment.`,
+            fillAmountUsd,
+            limit: fillLimits[limitIdx]?.limit,
+            transactionHash: deposit.transactionHash,
+          });
+          return;
+        }
+
+        // Decrement the applicable limit and all subsequent limits.
+        fillLimits.forEach((fillLimit) => (fillLimit.limit = fillLimit.limit.sub(fillAmountUsd)));
```
pxrl, line 490: Gut feel is that this is too raw to be here and should be factored out behind a small method like `this.decrementFillLimits(...)`.

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,97 @@ export class Relayer {
...
   }
 
+  /**
+   * For a given origin chain block range, sum the USD value of any fills made by this relayer.
+   * @param chainId Origin chain ID to inspect.
+   * @param fromBlock Optional origin chain block number lower bound.
+   * @param toBlock Optional origin chain block number upper bound.
+   * @returns The cumulative USD value of fills made by this relayer for deposits within the specified block range.
+   */
+  computeOriginChainCommitment(chainId: number, fromBlock?: number, toBlock?: number): BigNumber {
```
nicholaspai, line 290: Why make these optional if we always set them explicitly in this PR? Makes it a bit cleaner if we don't have optional params

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,97 @@ export class Relayer {
...
+
+      totalCommitment = totalCommitment.add(commitment);
+      const limit = usdThreshold.sub(totalCommitment);
+      toBlock = fromBlock - 1; // Shuffle the range for the next loop.
+
+      return { fromBlock, limit };
+    });
+
+    // Warn on the highest overcommitment, if any.
+    for (let i = limits.length - 1; i >= 0; --i) {
```
nicholaspai, line 336: Again I think its worth pointing out perhaps as just a comment that limits are sorted in descending order

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,97 @@ export class Relayer {
...
+
+  /**
+   * For all origin chains chains, map the relayer's deposit confirmation requirements to tiered USD amounts that can be
+   * filled by this relayer.
+   * @returns A mapping of chain ID to an array of origin chain fill limits in USD, ordered by origin chain block range.
+   */
+  computeFillLimits(): { [originChainId: number]: { fromBlock: number; limit: BigNumber }[] } {
+    // For each SpokePool reduce the amount available for fills by the amount
+    // previously committed within the origin chain's finality window.
+    const limits = Object.fromEntries(
```
nicholaspai, line 363: This resultant limits struct is worth logging right?

nicholaspai (approved):
Looks good after my comments are resolved and the last test for filtering deposits based on origin chain commitment is added

src/relayer/Relayer.ts
nicholaspai: yeah I was thinking the same thing. Both this call within `filterDeposit` and in `computeOriginChainLimits` should use this helper method

src/relayer/Relayer.ts
pxrl: I enforced it via an assert here: https://github.com/across-protocol/relayer/pull/1687/commits/8eb727f5e92b952fd0f9f7c22940861a2e7eb28e

src/relayer/Relayer.ts
pxrl: There are instances in the test where we query `computeOriginChainCommitment(..., 0, Number.MAX_SAFE_INTEGER)`, so that's one scenario where it would be useful. But it's also fine to be strict and require that the test is more explicit. Updated here: https://github.com/across-protocol/relayer/pull/1687/commits/e27cc4edd85ed9fdd3764ce32ef602159b1f2a5e

src/relayer/Relayer.ts
pxrl: Yeah - good call. Implemented: https://github.com/across-protocol/relayer/pull/1687/commits/2133f1377c064e669404f2a5c4dee68b28a8bdf7

test/Relayer.BasicFill.ts
```diff
@@ -480,6 +482,141 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
       expect(lastSpyLogIncludes(spy, "0 unfilled deposits found.")).to.be.true;
     });
 
+    it("Correctly tracks origin chain fill commitments", async function () {
```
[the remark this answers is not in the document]
pxrl: Done: https://github.com/across-protocol/relayer/pull/1687/commits/b3fd7ce3c698fe94d25d6cc22270c7200f6c91f4

src/relayer/Relayer.ts
pxrl: I refactored a bit here: https://github.com/across-protocol/relayer/pull/1687/commits/1c7115fde2eaadb8765e6c263de83e395d1983eb

test/Relayer.BasicFill.ts
```diff
@@ -476,10 +495,268 @@ describe("Relayer: Check for Unfilled Deposits and Fill", async function () {
...
+        usdThreshold: fillAmount.mul(100).div(100),
+        minConfirmations: n,
+      }));
+      originChainConfirmations.push({
+        usdThreshold: bnUint256Max,
+        minConfirmations: originChainConfirmations.length + 1,
+      });
+
+      // Set MDC such that the deposit is ignored. The profit client will return a fill USD amount of $0,
+      // so we need to set the MDC for the `0` threshold to be large enough such that the deposit would be ignored.
```
pxrl, line 533:
```suggestion
```

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,143 @@ export class Relayer {
...
+   */
+  computeOriginChainLimits(chainId: number): { fromBlock: number; limit: BigNumber }[] {
+    const mdcs = this.config.minDepositConfirmations[chainId];
+    const originSpoke = this.clients.spokePoolClients[chainId];
+
+    let totalCommitment = bnZero;
+    let toBlock = originSpoke.latestBlockSearched;
+
+    // For each deposit confirmation tier (lookback), sum all outstanding commitments back to head.
+    const limits = Object.values(mdcs).map(({ usdThreshold, minConfirmations }) => {
```
mrice32, line 357: Isn't mcds an array? Why not just call map directly? It took me a while to figure out that we weren't making assumptions about ordering of named properties in an object.

pxrl: You're right; it is. Fixed: https://github.com/across-protocol/relayer/pull/1687/commits/828439f3c6e228b239ad9035e96099ca143700ae

src/relayer/Relayer.ts
```diff
@@ -279,6 +280,143 @@ export class Relayer {
...
+  /**
+   * For a given origin chain, determine whether a new fill will overcommit the chain.
+   * @param originChainId Chain ID of origin chain.
+   * @param amount USD amount to evaluate.
+   * @param limitIdx Optional index into fillLimits to narrow the evaluation.
+   * @returns An index into the limits array.
+   */
+  originChainOvercommitted(originChainId: number, amount = bnZero, limitIdx = 0): boolean {
+    const fillLimits = this.fillLimits[originChainId].slice(limitIdx);
+    return fillLimits?.some(({ limit }) => limit.sub(amount).lt(bnZero)) ?? true;
```
mrice32, line 302: nit: Why do we need to test all the limits? Can't we just test the largest limit (most confirmations)? This would be first element in the array I think.

mrice32 (approved):
LGTM!

src/relayer/Relayer.ts
pxrl: It took me a while to figure out the right example for this, but the rationale is...

If we only check the current threshold then this risks overcommitting at the margins. i.e. in a scenario with USD thresholds:
10: 12 blocks
100: 24 blocks
1000: 48 blocks

We might commit 10 USD to 10 blocks out of the first 12, and concurrently make a fill for 100 USD at block 11. That 99 USD fill will be in the 100 USD range for the subsequent 13 blocks, at which point the 10 USD fills will also be incrementally entering that range. Suddenly the relayer has unintentionally overcommitted.

Per the current logic, this puts an immediate pause on any new 10 USD fills being made, because even though the 10 USD tier is empty, the 100 USD tier is beyond its intended capacity and will remain so until the 100 USD fill has been ejected into the 1k USD tier.

This was at least the idea behind it. Do you think it makes sense? It might just be me jumping at shadows.

mrice32: I 100% agree.

mrice32 (approved):
LGTM!

PR #1688, opened 2024-07-18
fix(ProfitClient): Fix fill amount USD computation

pxrl:
The ProfitClient currently assumes that the value of a fill can be determined from the input token price * the output amount of the output token. This is roughly correct for most Across deposits, but falls over when the deposit includes an in-protocol swap to an output token with a different number of decimals.

This was exposed on an Optimism deposit for inputToken USDC, outputToken AERO on Base.

Deposit hash:
  0x4a3febcfb764467d9eae261b3a3938137db383289fe328050a2871eb407e1147


src/clients/ProfitClient.ts
```diff
@@ -390,16 +391,26 @@ export class ProfitClient {
...
-      );
+  getFillAmountInUsd(
+    deposit: Pick<Deposit, "destinationChainId" | "outputToken" | "outputAmount">
+  ): BigNumber | undefined {
+    const { destinationChainId, outputToken, outputAmount } = deposit;
+    let l1Token: L1Token;
+
+    try {
+      l1Token = this.hubPoolClient.getL1TokenInfoForL2Token(outputToken, destinationChainId);
+    } catch {
```
pxrl, line 402: Inb4 Matt to flag that this is a no-no, but is currently required because the HubPoolClient throws if it can't resolve a token. I started a draft PR to fix that a while ago but it stalled; I will pick it up: https://github.com/across-protocol/sdk/pull/643

nicholaspai (approved):
https://github.com/across-protocol/relayer/pull/1688/files#r1684522349

LGTM if my understanding here is correct

src/clients/ProfitClient.ts
pxrl: If we throw then the bot goes down whenever someone specifies and unknown outputToken.

The alternative is to just compute the amount based on the inputToken, where we can rely on having a price, but that ignores the relayer fee and LP fee.

PR #1690, opened 2024-07-18
fix(executor): Force no multicall on Polygon

pxrl:
Relayer refund leaves must not be batched via multicall on Polygon. The existing executor implementation will attempt multicall in cases where there are multiple relayer refund leaves, and this imposes a soft deadlock on the executor. This can be recovered by executing one leaf at a time. A better solution is just to override the defaults for Polygon in the dataworker.


james-a-morris (approved):
Makes sense - but could we maybe put this logic in `constructClients` and expose this override in `config`?

james-a-morris:
After some thought - we're assuming the only info coming from the config is the `multiCallChunkSize` which may not always be the case. Would it make sense to just override the config before we call `constructCommonClients` to be:

{
    ...config,
    { ...config.multiCallChunkSize, [CHAIN_IDs.POLYGON]: 1 }
}

pxrl:
One clarification here - the requirement for multicall in the Polygon SpokePool is more nuanced than I originally thought. I'm not aware of how we could be multicalling executeRelayerREfund and something else, so this may simply have been stuck transaction weirdness. In general though, I don't see any major problem with skipping multicall on Polygon since the transactions themselves are so cheap anyway.

```
    /**                                                                                                                                                                                        
     * @notice Override multicall so that it cannot include executeRelayerRefundLeaf                                                                                                           
     * as one of the calls combined with other public function calls.                                                                                                                          
     * @dev Multicalling a single transaction will always succeed.                                                                                                                             
     * @dev Multicalling execute functions without combining other public function calls will succeed.                                                                                         
     * @dev Multicalling public function calls without combining execute functions will succeed.                                                                                               
     */                                                                                                                                                                                        
    function _validateMulticallData(bytes[] calldata data) internal pure override {                                                                                                            
        bool hasOtherPublicFunctionCall = false;                                                                                                                                               
        bool hasExecutedLeafCall = false;                                                                                                                                                      
        for (uint256 i = 0; i < data.length; i++) {                                                                                                                                            
            bytes4 selector = bytes4(data[i][:4]);                                                                                                                                             
            if (selector == SpokePoolInterface.executeRelayerRefundLeaf.selector) {                                                                                                            
                if (hasOtherPublicFunctionCall) revert MulticallExecuteLeaf();                                                                                                                 
                hasExecutedLeafCall = true;                                                                                                                                                    
            } else {                                                                                                                                                                           
                if (hasExecutedLeafCall) revert MulticallExecuteLeaf();                                                                                                                        
                hasOtherPublicFunctionCall = true;                                                                                                                                             
            }                                                                                                                                                                                  
        }                                                                                                                                                                                      
    }
```

pxrl:
> After some thought - we're assuming the only info coming from the config is the `multiCallChunkSize` which may not always be the case. Would it make sense to just override the config before we call `constructCommonClients` to be:
> 
> { ...config, { ...config.multiCallChunkSize, [CHAIN_IDs.POLYGON]: 1 } }

That's a cleaner suggestion; yeah. This can probably be handled in the dataworker config parser.

pxrl:
I'll merge this in because I've seen stuck transactions again multiple times over the past couple of weeks.

pxrl:
@james-a-morris Just circling back to this one - I applied the change you suggested :+1:

james-a-morris (approved):

pxrl:
Closing this one out now - this issue hasn't been seen for a very long time.

PR #1696, opened 2024-07-22
feat: Create generic AtomicWethDepositor

nicholaspai:
Create generic weth depositor that is designed to only be deployed once and upgradeable whenver there is a new chain we need to add support for.

The motivation is to eliminate any future contract changes in exchange for slight UX degradation where caller (i.e. relayer) needs to perform a bit more upfront computation before calling this contract. For example, look at the removed functionality of `bridgeWethToZkSync` to see how relayer would now need to cmopute transaction base cost before calling this contract.


contracts/AtomicWethDepositor.sol
```diff
@@ -96,107 +45,42 @@ contract AtomicWethDepositor is Ownable, MultiCaller {
...
-            new bytes[](0),
-            refundRecipient
-        );
-        emit ZkSyncEthDepositInitiated(msg.sender, to, amount);
+    function bridgeWeth(uint256 value, address bridge, bytes calldata bridgeCallData) public nonReentrant {
+        _withdrawWeth(value);
+        bytes4 whitelistedFuncSelector = ovmChainIdToBridge[bridge];
+        if (whitelistedFuncSelector != bytes4(bridgeCallData)) revert InvalidBridgeFunction();
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory result) = bridge.call(bridgeCallData);
```
bmzig, line 81: I think the call here needs a value, right? I know at least for zkSync it is for the gas costs of the txn.

bmzig (approved):
Looks good. You may optionally want to change `ovmChainIdToBridge` to just `chainIdToBridge` but we aren't too exposed to this mapping anyways (if at all). Obviously, since we define `receive` we also run the risk of losing eth if we don't properly encode the transactions, but that's also probably not much of a concern.

nicholaspai:
> Looks good. You may optionally want to change `ovmChainIdToBridge` to just `chainIdToBridge` but we aren't too exposed to this mapping anyways (if at all). Obviously, since we define `receive` we also run the risk of losing eth if we don't properly encode the transactions, but that's also probably not much of a concern.

Ah good call. renamed `ovmChainIdToBridge` to `whitelistedBridgeFunctions`

contracts/AtomicWethDepositor.sol
```diff
@@ -3,89 +3,38 @@ pragma solidity ^0.8.0;
...
     ///////////////////////////////
     //          Events           //
     ///////////////////////////////
 
-    event ZkSyncEthDepositInitiated(address indexed from, address indexed to, uint256 amount);
-    event LineaEthDepositInitiated(address indexed from, address indexed to, uint256 amount);
-    event OVMEthDepositInitiated(uint256 indexed chainId, address indexed from, address indexed to, uint256 amount);
-    event PolygonEthDepositInitiated(address indexed from, address indexed to, uint256 amount);
-    event OVML1BridgeModified(uint256 indexed chainId, address indexed newBridge, address oldBridge);
+    event AtomicWethDepositInitiated(address indexed from, address indexed bridge, uint256 amount);
```
james-a-morris, line 31: We do rely on some of these named events and the `bridge` is subject to change over time. Could we consider adding a `chainId` to this so we don't have to maintain a mapping off-chain?

nicholaspai: I think we can use the `bridge` address here to figure out which bridge we're using against an off-chain mapping. I think the issue with the `chainId` is.. where do we get that information from? Currently it would have to be passed in by the weth depositor caller, which means it could be incorrect

nicholaspai: but I do agree this event tracking is a weakness since we can no longer parse out the `to` parameter. Is this a blocker? I think not but curious what you think or if there are other workarounds

contracts/AtomicWethDepositor.sol
```diff
@@ -96,107 +45,42 @@ contract AtomicWethDepositor is Ownable, MultiCaller {
...
-            new bytes[](0),
-            refundRecipient
-        );
-        emit ZkSyncEthDepositInitiated(msg.sender, to, amount);
+    function bridgeWeth(uint256 value, address bridge, bytes calldata bridgeCallData) public nonReentrant {
+        _withdrawWeth(value);
+        bytes4 whitelistedFuncSelector = whitelistedBridgeFunctions[bridge];
+        if (whitelistedFuncSelector != bytes4(bridgeCallData)) revert InvalidBridgeFunction();
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory result) = bridge.call{ value: value }(bridgeCallData);
```
bmzig, line 81: Sorry I missed this too. Do you think it's alright to not whitelist the bridge addresses too? This is just a `call`, so we don't run the risk of somebody destroying the contract, but this means that we will delete eth if we do not have the right address in the bridge field. Also, anybody an make a contract with a whitelisted selector and do whatever they want (though, I don't _think_ that is much of an issue in this context).

nicholaspai: we are inherently whitelisting the bridge addresses right? The `bridge` address must be set as a key by the admin in the `whitelistedBridgeFunctions` mapping.

> this means that we will delete eth if we do not have the right address in the bridge field
I think you can assume the admin will set these correctly since only the admin can update this mapping

> anybody can make a contract with a whitelisted selector and do whatever they want
But not anyone can add a contract address + selector, right?

bmzig: Yes that's right. I guess you could still technically send eth to any address, but that's not an issue since this contract shouldn't hold any eth for more than a block.

bmzig (approved):

contracts/AtomicWethDepositor.sol
```diff
@@ -3,89 +3,38 @@ pragma solidity ^0.8.0;
...
+contract AtomicWethDepositor is Ownable, MultiCaller, Lockable {
+    Weth public immutable WETH = Weth(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
 
     /**
-     * @notice All OVM Chains have an OvmL1Bridge contract that can be called to deposit ETH.
-     * @notice This mapping is a convenience to allow for easy lookup of the OvmL1Bridge contract for a given chainId.
+     * @notice Mapping of function selectors to whitelisted bridge addresses that can be called by this contract.
      */
-    mapping(uint256 => OvmL1Bridge) public ovmChainIdToBridge;
+    mapping(address => bytes4) public whitelistedBridgeFunctions;
```
mrice32, line 25: Related to another comment below: why can't we just map from chainId to a struct with the bridge address _and_ the selector?

This would allow the caller to pass in the chainId and not have to worry about providing the bridge address. Would also allow the events to contain a validated chainId.

mrice32 (commented):
One comment!

contracts/AtomicWethDepositor.sol
nicholaspai: hmm that's a good idea!

nicholaspai: c0293fb842083f52ff3a8d4697a30abd6f31ec43

mrice32 (approved):
LGTM!

PR #1698, opened 2024-07-22
fix(relayer): Restore internal fill status tracking

pxrl:
This is necessary to protect against attempting repeat fills on chains where the transaction confirmation times can span multiple relayer loops. Without this the relayer may try to fill the same deposit multiple times. This was exposed by a third-party relayer running in looping mode on mainnet.


src/relayer/Relayer.ts
```diff
@@ -666,6 +679,8 @@ export class Relayer {
     const contract = spokePoolClients[deposit.destinationChainId].spokePool;
     const chainId = deposit.destinationChainId;
     multiCallerClient.enqueueTransaction({ contract, chainId, method, args, gasLimit, message, mrkdwn });
+
+    this.setFillStatus(deposit, FillStatus.RequestedSlowFill);
```
nicholaspai, line 683: Should this be a different fill status??

pxrl, line 683:
```suggestion
    this.setFillStatus(deposit, FillStatus.Filled);
```

pxrl: Yep; good catch. I've added tests to catch that now and have updated several others to accommodate this change. https://github.com/across-protocol/relayer/pull/1698/commits/1be8e126896581a846cdcf70e054ef1b8074ea96 https://github.com/across-protocol/relayer/pull/1698/commits/4e493ef9051e942dc2cd2fd93a48fd699cc72764 https://github.com/across-protocol/relayer/pull/1698/commits/8ebbeea916cabb08421a3998ad989d0a469a7dc3

nicholaspai (approved):

PR #1699, opened 2024-07-23
feat(scroll): Add Scroll adapter

bmzig:
This instantiates a Scroll Adapter based off of the new generic adapter code. This is a smaller fix since I did not want to modify the Inventory/CrossChainTransfer clients when we would have to revert them back as soon as the generic code replaces the current.

### DIFF
--- a/src/clients/bridges/ScrollAdapter.ts
+++ b/src/clients/bridges/ScrollAdapter.ts
@@ -1,0 +1,32 @@
+import { SUPPORTED_TOKENS, CUSTOM_BRIDGE, CANONICAL_BRIDGE, DEFAULT_GAS_MULTIPLIER } from "../../common";
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP, winston } from "../../utils";
+import { SpokePoolClient } from "../SpokePoolClient";
+import { BaseChainAdapter } from "../../adapter/BaseChainAdapter";
+
+export class ScrollAdapter extends BaseChainAdapter {
+  constructor(
+    logger: winston.Logger,
+    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    monitoredAddresses: string[]
+  ) {
+    const { SCROLL, MAINNET } = CHAIN_IDs;
+    const bridges = {};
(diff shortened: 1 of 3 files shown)


src/clients/bridges/ScrollAdapter.ts
```diff
@@ -0,0 +1,32 @@
...
+  ) {
+    const { SCROLL, MAINNET } = CHAIN_IDs;
+    const bridges = {};
+    const l2Signer = spokePoolClients[SCROLL].spokePool.signer;
+    const l1Signer = spokePoolClients[MAINNET].spokePool.signer;
+    SUPPORTED_TOKENS[SCROLL]?.map((symbol) => {
+      const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[MAINNET];
+      const bridgeConstructor = CUSTOM_BRIDGE[SCROLL][l1Token] ?? CANONICAL_BRIDGE[SCROLL];
+      bridges[l1Token] = new bridgeConstructor(SCROLL, MAINNET, l1Signer, l2Signer, l1Token);
+    });
```
james-a-morris, line 20: Do we do this elsewhere? Could be a candidate for pushing to a helper function.

bmzig: We do it here: https://github.com/across-protocol/relayer/blob/0b64c1897bc735ea507ea5ea0235b96ceb43cbda/src/adapter/AdapterManager.ts#L42 in the new adapter manager. What do you think? Should this be factored out? I do like keeping everything together in the context of the Adapter manager constructor so we keep things easy to follow.

src/clients/bridges/ScrollAdapter.ts
pxrl, line 1: This is pretty clever - nice way to shim the new adapter into the existing adapter manager!

pxrl (approved):

src/clients/bridges/ScrollAdapter.ts
bmzig: I'm going to keep it as is now and then if we do not fully migrate over to the generic adapters after a few more chains, then we will factor this out. This code _should_ not exist anyways come 1-2 months from now.

PR #1708, opened 2024-07-26
fix(generic bridges): use scroll's canonical bridge function

bmzig:
This fixes an improperly set method on Arbitrum (not sending any `msg.value` to the `outboundTransfer` function), as well as removes `ScrollWethBridge` in favor of the `ScrollERC20Bridge`. The latter is done since we can simply bridge WETH to Scroll via the gateway router, which can subsequently be wrapped/unwrapped, so a dedicated bridge to interface with an atomic depositor is not needed.
_______
As an overview. We recently merged in the generic adapters PR https://github.com/across-protocol/relayer/pull/1524, which at the time of writing, is dormant in production, since the generic AdapterManager in src/adapter/ is only activated if we set `RELAYER_USE_GENERIC_ADAPTER` and/or `MONITOR_USE_GENERIC_ADAPTER` in the environment. Now that it is merged, instead of making an entirely new ScrollAdapter, we can make a wrapper for the generic `BaseChainAdapter` and begin integrating the new code into our current structure. Going forward, we should continue to replace the existing specialized adapters (e.g. ArbitrumAdapter, LineaAdapter, ZkSyncAdapter, etc.) with wrappers for `BaseChainAdapter` using the same format found here https://github.com/across-protocol/relayer/blob/5bd2efd9f818a0fe8adfbb638f5c0b70720153e2/src/adapter/AdapterManager.ts#L35. Once all the adapters are replaced with their generic counterparts, we should be able to remove everything in src/clients/bridges/ except for `CrossChainTransferClient`, which will remain unchanged.


nicholaspai:
> (although this shouldn't be preferred since we would receive WETH, not ETH, on Scroll).

I'm confused by this. We should be happy to send WETH, because the bots should know how to unwrap WETH into ETH, right? [This function](https://github.com/across-protocol/relayer/blob/27f04faccaba5bd710121ed798b673b7f84aff95/src/clients/InventoryClient.ts#L922) should be [run here](https://github.com/across-protocol/relayer/blob/27f04faccaba5bd710121ed798b673b7f84aff95/src/relayer/index.ts#L66) before evaluating any WETH rebalances, allowing the bots to always assume it has enough ETH in its balance before evaluating WETH inventory sends.

From the inventory manager's perspective, ETH is treated as gas balance and WETH is an inventory balance

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -40,10 +50,12 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
     l2Token: string,
     amount: BigNumber
   ): Promise<BridgeTransactionDetails> {
+    const { l1Gateway, l2GasLimit, l2GasPrice, transactionSubmissionData, l1SubmitValue } = this;
     return Promise.resolve({
-      contract: this.l1Gateway,
+      contract: l1Gateway,
       method: "outboundTransfer",
```
nicholaspai, line 56: How was this working before without the `value`..? Also how did you derive the l1 submit value?

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -28,10 +43,13 @@ export class ScrollERC20Bridge extends BaseBridgeAdapter {
...
   ): Promise<BridgeTransactionDetails> {
+    const baseFee = await this.getScrollGasPriceOracle().estimateCrossDomainMessageFee(this.l2Gas);
+    const bufferedFee = baseFee.mul(this.feeMultiplier).div(fixedPointAdjustment);
     return Promise.resolve({
       contract: this.getL1Bridge(),
       method: "depositERC20",
       args: [l1Token, toAddress, amount, this.l2Gas],
+      value: bufferedFee,
     });
   }
```
nicholaspai, line 55: Have you tested that `queryL1BridgeInitiationEvents` and `queryL2BridgeFinalizationEvents` work as intended so we don't send duplicate rebalances? or will that be for a follow up?

nicholaspai (commented):
Couple questions, but nice to remove the Scroll(Atomic)WethBridge!

src/adapter/bridges/ArbitrumOneBridge.ts
bmzig: I did not have an eth inventory config defined before, but this code isn't being used in prod yet. I got the value from here https://github.com/across-protocol/relayer/blob/5bd2efd9f818a0fe8adfbb638f5c0b70720153e2/src/clients/bridges/ArbitrumAdapter.ts#L67, which is being used in prod. I am unsure where this number originally came from, though.

nicholaspai: Oh so why isn't this being used in prod yet? It's weird to have two versions of similar functional code here it's confusing to reviewers. I'd rather there be a comment or some explanation of what is going on here

bmzig: I updated the description of this PR to describe more of the motivation for making these changes.

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -28,27 +56,43 @@ export class ScrollERC20Bridge extends BaseBridgeAdapter {
...
 
+  // The deposit/finalize events that Scroll gateways emit do not index the recipient of transfers. This makes cross-chain transfer tracking a bit
+  // tricky since the recipient of a deposit may not be the same as the depositor (e.g. hub -> spoke transfers, or randomEoa -> monitoredEoa transfers).
+  // For both queryL1BridgeInitiationEvents and queryL2BridgeFinalizationEvents, we currently make the assumption that if we are sending to an L2 contract, then
+  // this L2 contract is the spoke pool, and the original depositor is the hub pool. Likewise, if we are sending to an EOA, then the recipient is the same as the depositor.
+  // While this limits our ability to track all types of events (and incurs an extra RPC call), we must do this to prevent an expensive event query
+  // (i.e. ...filters.DepositERC20(l1Token, undefined, undefined), which would return, for example, all WETH deposits to Scroll within the past 10,000 blocks).
```
bmzig, line 74: I want to highlight this since this approach is introducing some tech debt.

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: Added a test based on the Linea test in this commit https://github.com/across-protocol/relayer/pull/1708/commits/822c08e2b9f34e54cf5ddea665338fd8afe1b992. There is an interesting development highlighted here: https://github.com/across-protocol/relayer/pull/1708#discussion_r1695621243

bmzig:
ERC20 transfer tracking is working now.

src/adapter/bridges/ArbitrumOneBridge.ts
pxrl: @nicholaspai We'll work on phasing these new adapters in one at a time, in the same way the Scroll one is being used.

src/adapter/bridges/BaseBridgeAdapter.ts
```diff
@@ -59,7 +60,12 @@ export abstract class BaseBridgeAdapter {
   ): Promise<BridgeEvents>;
 
   protected resolveL2TokenAddress(l1Token: string): string {
-    return getTranslatedTokenAddress(l1Token, this.hubChainId, this.l2chainId, false);
+    return getTranslatedTokenAddress(
+      l1Token,
+      this.hubChainId,
+      this.l2chainId,
+      isDefined(TOKEN_SYMBOLS_MAP.USDC.addresses[this.l2chainId])
```
pxrl, line 67: Hm, is this intended? I think the this flag is really used to distinguish bridged from native USDC, but for Scroll we have USDC that is currently bridged and will transition to native. Just want to be sure that we won't accidentally get it thinking that Scroll is using native.

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -1,25 +1,53 @@
...
+  isContractDeployedToAddress,
+} from "../../utils";
+import { CONTRACT_ADDRESSES, SCROLL_CUSTOM_GATEWAY } from "../../common";
 import { BaseBridgeAdapter, BridgeTransactionDetails, BridgeEvents } from "./BaseBridgeAdapter";
 import { processEvent } from "../utils";
 
+const SCROLL_STANDARD_GATEWAY: { l1: string; l2: string } = {
+  l1: "0xD8A791fE2bE73eb6E6cF1eb0cb3F36adC9B3F8f9",
+  l2: "0xE2b4795039517653c5Ae8C2A9BFdd783b48f447A",
+};
```
pxrl, line 19: Thoughts on whether these addresses belong in `ContractAddresses.ts` ? I'm not sure either way; I know this pattern is used in some other adapters (Polygon and Arbitrum), but there are also comments over there that these addresses should be migrated :shrug:

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -28,27 +56,43 @@ export class ScrollERC20Bridge extends BaseBridgeAdapter {
     l2Token: string,
     amount: BigNumber
   ): Promise<BridgeTransactionDetails> {
+    const baseFee = await this.getScrollGasPriceOracle().estimateCrossDomainMessageFee(this.l2Gas);
+    const bufferedFee = baseFee.mul(this.feeMultiplier).div(fixedPointAdjustment);
```
pxrl, line 60: Did you see whether the excess get refunded to the recipient on the destination chain?

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -28,27 +56,43 @@ export class ScrollERC20Bridge extends BaseBridgeAdapter {
...
     const events = await paginatedEventQuery(
       l1Bridge,
-      l1Bridge.filters.DepositERC20(l1Token, undefined, fromAddress),
+      l1Bridge.filters.DepositERC20(l1Token, undefined, monitoredFromAddress),
       eventConfig
     );
+    // Take all events which are sending an amount greater than 0.
+    const processedEvents = events
+      .map((event) => processEvent(event, "amount", "to", "from"))
+      .filter(({ amount }) => amount > BigNumber.from(0));
```
pxrl, line 93: Prefer to use `bnZero` instead of `BigNumber.from(0)` because the latter just creates a bunch of disposable objects that have ~0 lifetime and need to be GC'ed.

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: The only downside to that is needing some extra logic on `CONTRACT_ADDRESSES` since they are determined by the `l1Token` being bridged. For example, in Arbitrum, DAI/USDC/WETH all have their own gateway, but in `CONTRACT_ADDRESSES`, I can only fetch a contract address from a chainId.

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: I believe the excess gets refunded on mainnet. https://github.com/scroll-tech/scroll-contracts/blob/ca7f0768b6640dc10b19f3d4da3943a87bdf11b1/src/L1/L1ScrollMessenger.sol#L366

src/adapter/bridges/BaseBridgeAdapter.ts
bmzig: I see. I did this since `TOKEN_SYMBOLS_MAP` only has a USDC entry for Scroll and no USDC.e entry, so this will error otherwise. I think this can be fixed sustainably if for all chains where there exists native USDC, we define a USDCTokenSplitterBridge as a custom bridge, and then internally, if there is no CCTP, we just pass through to the canonical.

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: https://github.com/across-protocol/relayer/pull/1708/commits/e9704f6f0dd5b104a9219fe722a7bb7922747ec5

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: CONTRACT_ADDRESSES is keyed by chainId though, I'm not sure I understand bennett's point?

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: Have you confirmed this behavior?

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: what's the tech debt?

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: can you not create multiple entries in CONTRACT_ADDRESSES[chainId] for example scrollStandardBridge, scrollWbtcStandardBridge etc

bmzig: Yes, but if we do that, in `ScrollERC20Bridge` and `ArbitrumOneBridge` we would then need to do something like:
```
if(token is USDC) { l1Bridge = CONTRACT_ADDRESSES[chainId].scrollStandardGateway }
else if (token is WETH) { l1Bridge = CONTRACT_ADDRESSES[chainId].scrollWethGateway }
... // so forth
```
If this is preferred, then the change can be made, but I feel as though the diff would be greater.

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: Yeah, we can see this transaction get refunded 0.0000068 eth from the 0.0000204347691 `msg.value` fee. https://etherscan.io/tx/0x3ddd5416a2eab55cb959a3500d4663c765e44a7418670f86d1356254c67bdfd4#internal

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: Lack of flexibility to monitor contracts and an extra RPC call.

nicholaspai: i don't think the extra RPC call is tech debt, its a tiny cost to incur for bots that don't require great speed (rebalancer and monitor). 

> Lack of flexibility to monitor contracts

Please elaborate, and suggest a solution

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: do you mean WBTC where you write WETH...? I don't understand the issue. We have many chain-token bridges, like ScrollWeth and ScrollERC20

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: I assume that if the monitored address (i.e. `toAddress`) is an L2 contract, then it is the spoke pool, so right away we can't monitor any other L2 contracts. Also, I assume that if `toAddress` is a contract, then all transfers must have come from the hub pool (which may be incorrect in the future). Also, I assume that if `toAddress` is an EOA, then transfers must have come from that same EOA on mainnet, which is not necessarily true. All of this _works_ at this current time, but is isn't too flexible for future development. 

We can fix this by changing how this function https://github.com/across-protocol/relayer/blob/db2e4a470c357d911cf1c1292cddfb838f0f18a8/src/adapter/BaseChainAdapter.ts#L229 deals with monitored addresses, but I'm not sure what that would look like at the moment...

src/adapter/bridges/ScrollERC20Bridge.ts
bmzig: There is only one `ScrollERC20Bridge` class here? The issue is we send tokens to the same gateway, but listen to events from addresses specific to the token. The same is true for Arbitrum. 

The BaseChainAdapter has a defined bridge for each token enabled on Scroll (WBTC, USDT, USDC, and WETH). So it holds a map which has WBTC -> ScrollERC20Bridge, USDC -> ScrollERC20Bridge, USDT -> ScrollERC20Bridge, WETH -> ScrollERC20Bridge. Each bridge needs to query events from a specific gateway, which is not the same for each token. So during construction of the bridge, in order to tell the ScrollERC20Bridge which gateway to query, we would need to check which L1 token for which the bridge is built. That is why we would need to have all those conditionals, since `CONTRACT_ADDRESSES` only filters by chainId, so the constructor would look something like:
```
if (l1Token === USDC) { this.l1Bridge = CONTRACT_ADDRESSES[chainId].scrollUSDCGateway; }
else if (l1Token === WETH) { this.l1Bridge = CONTRACT_ADDRESSES[chainId].scrollWETHGateway; }
// so on
```

src/adapter/bridges/BaseBridgeAdapter.ts
pxrl: Hm; it could work..but I'm also surprised because I managed to bridge USDC over with an earlier version of this PR where this change wasn't added (iirc).

What happens when you bridge USDC with this config ?

bmzig: It turns out that there _was_ an issue bridging USDC. Originally, I swapped it to this since what would happen is we would get a struct like the following: 
```
{
  undefined: [
    {
      amount: BigNumber { value: "1" },
      to: '0x2e9A5da8Db81F449c70934cd8Ea6483c478CA878',
      from: '0x59EC481c57727e8802FAFA4c7c362f843C5A522c',
      l1Token: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
      l2Token: '0x06eFdBFf2a14a7c8E15944D1F4A48F9F95F663A4',
      data: '0x',
      blockNumber: 26,
      transactionIndex: 0,
      logIndex: 0,
      transactionHash: '0x2c41b9bc5eba18b74a1f23461bd75c1a2ab02ab64698a640a0e69e483b3a588c'
    }
  ]
}
```
And we still will get this, so until I make the fix to use the USDC token splitter bridge as a pass through, we will need to have this `isDefined(...)` constraint in place (Scroll is the only generic adapter at the moment, so it should be alright). It turns out that I was also using an incorrect gateway for USDC, which has its own custom gateway so that it can be upgraded once Scroll supports CCTP. The fix is here. I've personally tested rebalancing for all tokens except WBTC, and they are all good. https://github.com/across-protocol/relayer/pull/1708/commits/f8221b530aab7a49847b5552134d090021a2b372

src/adapter/AdapterManager.ts
```diff
@@ -41,7 +41,7 @@ export class GenericAdapterManager extends AdapterManager {
 
         SUPPORTED_TOKENS[chainId]?.map((symbol) => {
           const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[hubChainId];
-          const bridgeConstructor = CUSTOM_BRIDGE[chainId][l1Token] ?? CANONICAL_BRIDGE[chainId];
+          const bridgeConstructor = CUSTOM_BRIDGE[chainId]?.[l1Token] ?? CANONICAL_BRIDGE[chainId];
```
pxrl, line 44: I wonder if we could make this safer by requiring that the chain ID is defined in `CUSTOM_BRIDGE`. This goes against the idea of NCA because it requires an additional change to the code for each new chain and will tend to pollute constants, but it ensures that someone has verified that there are _no_ custom bridges for the chain. This would hopefully avoid the situation where there _is_ a custom bridge, but it wasn't identified and so tokens get nuked.

src/adapter/bridges/BaseBridgeAdapter.ts
pxrl: Yeah; I think this is OK for Scroll because it's resolving the address mapped by the USDC symbol, which is fine because we only have one USDC definition there anyway. It will cause problems for any chain where both native and bridged USDC exist though. I've dropped a todo in there and added a Linear task to ensure we don't miss it. https://github.com/across-protocol/relayer/pull/1708/commits/91279197878823ea8caf1903a3f834b19e7ae666

pxrl: (I have also tested this w/ Scroll and rebalancing worked OK)

src/common/Constants.ts
```diff
@@ -458,6 +454,20 @@ export const CUSTOM_ARBITRUM_GATEWAYS: { [chainId: number]: { l1: string; l2: st
...
+export const SCROLL_CUSTOM_GATEWAY: { [chainId: number]: { l1: string; l2: string } } = {
+  [TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET]]: {
+    l1: "0x7AC440cAe8EB6328de4fA621163a792c1EA9D4fE",
+    l2: "0x7003E7B7186f0E6601203b99F7B8DECBfA391cf9",
+  },
+  [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: {
+    l1: "0xf1AF3b23DE0A5Ca3CAb7261cb0061C0D779A5c7B",
+    l2: "0x33B60d5Dd260d453cAC3782b0bDC01ce84672142",
+  },
+};
```
pxrl, line 469: I confirmed these addresses via the respective L1 and L2 gateways.

It does seem fragile to have to manually specify so many addresses. I'm aware that we do this in other places; I wonder if we should have some runtime checks that query the gateway contracts and verify that the addresses still match.

src/adapter/bridges/ScrollERC20Bridge.ts
pxrl: I'll mark this resolved for now in order to get it merged; we can follow up with a refactor if necessary.

src/adapter/AdapterManager.ts
pxrl: @bmzig For follow-up.

pxrl (approved):
I've tested this OK when rebalancing all 4 supported tokens. Cross-chain transfer tracking worked correctly. There are some items here that could be revisited for a post-merge cleanup, but for now this is OK to go in.

src/adapter/bridges/ScrollERC20Bridge.ts
nicholaspai: Hm I think the tricky thing is that sometimes we want to track Hub to Spoke transfers where the Hub address != Spoke adderss, whereas sometimes we want to track relayer rebalances where the L1 and L2 addresses are the same. Perhaps we can add parameters to  `getOutstandingCrossChainTransfers` to make this easier but we'd still need to deal with the multiple branches within the function. Do you see any way around this?

> Also, I assume that if toAddress is a contract, then all transfers must have come from the hub pool (which may be incorrect in the future)

I think one assumption we can make that may simplify things is if the fromAddress is the hubPool address (which we have access to) then the destination is always going to be the spoke. Otherwise, we can asssume its an EOA.

I think the current logic in other adapters has something like this?

bmzig: I agree in that the solution likely involves some tweaking to `getOutstandingCrossChainTransfers` I also believe you are correct with the assumption that if `fromAddress` is the hubPool, `toAddress` is the spoke, although sometimes we don't [even distinguish this](https://github.com/across-protocol/relayer/blob/0be62319f314f5fbab22043f30cc9035ea06b384/src/clients/bridges/ArbitrumAdapter.ts#L88) and other times we switch our assumption and assume that if the spoke pool is the to `toAddress`, then the hub pool is the `fromAddress` [(source)](https://github.com/across-protocol/relayer/blob/0be62319f314f5fbab22043f30cc9035ea06b384/src/clients/bridges/ZKSyncAdapter.ts#L152). I think we're still brainstorming ideas, but they generally revolve around removing some of the ambiguity from `monitoredAddresses` (e.g. by doing checks if they are known addresses) either beforehand or in `getOutstandingCrossChainTransfers`. @pxrl might have better insight on this, too.

PR #1709, opened 2024-07-26
improve(relayer): swap priority of origin/destination for Lite chain relayer repayments

bmzig:
This issue is especially pronounced with Lite chains such as Lisk, where filling a deposit from mainnet to Lisk may result in unwanted repayment on Lisk, causing the relayer's balance to be perpetually above target. In general, Lite chains cause the relayer to end up with a lot of inventory on the lite chain so we should aim to take repayment away from the lite chain when we get the chance.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -464,9 +464,9 @@ export class InventoryClient {
     // lite chain deposits force repayment on origin, we end up taking lots of repayment on the lite chain so
     // we should take repayment away from the lite chain where possible.
     if (
+      deposit.toLiteChain &&
       !chainsToEvaluate.includes(originChainId) &&
-      this._l1TokenEnabledForChain(l1Token, Number(originChainId)) &&
-      deposit.toLiteChain
+      this._l1TokenEnabledForChain(l1Token, Number(originChainId))
     ) {
       chainsToEvaluate.push(originChainId);
     }
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -460,23 +460,23 @@ export class InventoryClient {
       chainsToEvaluate.push(...chainsWithExcessSpokeBalances);
     }
     // Add destination and origin chain if they are not already added.
-    // Prioritize destination chain repayment over origin chain repayment but prefer both over
+    // Prioritize origin chain repayment over destination chain repayment but prefer both over
     // hub chain repayment if they are under allocated. We don't include hub chain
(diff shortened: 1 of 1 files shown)


nicholaspai:
Nit your latek syntax isn't rendering can you just use the markdown syntax?

nicholaspai:
Why? 

> but with the inclusion of the aforementioned PR, the origin chain will never be considered (unless it is also a slow withdrawal chain)

nicholaspai (commented):
I don't fully understand the motivation for this can you explain more?

src/clients/InventoryClient.ts
```diff
@@ -459,6 +459,15 @@ export class InventoryClient {
         .map(([chainId]) => Number(chainId));
       chainsToEvaluate.push(...chainsWithExcessSpokeBalances);
     }
+    // Add origin chain to take higher priority than destination chain if the destination chain
+    // is a lite chain, which should allow the relayer to take more repayments on mainnet.
```
nicholaspai, line 463:
```suggestion
    // is a lite chain, which should allow the relayer to take more repayments away from the lite chain. Because 
    // lite chain deposits force repayment on origin, we end up taking lots of repayment on the lite chain so
    // we should take repayment away from the lite chain where possible.
```

nicholaspai (approved):

src/clients/InventoryClient.ts
```diff
@@ -459,6 +459,17 @@ export class InventoryClient {
         .map(([chainId]) => Number(chainId));
       chainsToEvaluate.push(...chainsWithExcessSpokeBalances);
     }
+    // Add origin chain to take higher priority than destination chain if the destination chain
+    // is a lite chain, which should allow the relayer to take more repayments away from the lite chain. Because 
+    // lite chain deposits force repayment on origin, we end up taking lots of repayment on the lite chain so
+    // we should take repayment away from the lite chain where possible.
+    if (
+      !chainsToEvaluate.includes(originChainId) &&
+      deposit.toLiteChain &&
```
pxrl, line 468: I think we can flip these two conditionals because in most cases the first will evaluate true but we will bail on the 2nd because we'd expect relatively fewer deposits destined for lite chains relative to the overall deposit flow.

pxrl (approved):

src/clients/InventoryClient.ts
bmzig: https://github.com/across-protocol/relayer/pull/1709/commits/94bcaa9434775ea9eb5004bff524ae5b5786fc48

pxrl (approved):

PR #1714, opened 2024-07-29
improve(adapters): swap Linea adapter with generic adapter

bmzig:
Here is an example of swapping one of our current adapters with a wrapper for a generic one. I believe it would be good to introduce these wrappers one by one. Linea was an arbitrary choice. This can be merged anytime but should probably be done a little bit after Scroll has been live.


src/adapter/bridges/LineaBridge.ts
```diff
@@ -42,7 +42,7 @@ export class LineaBridge extends BaseBridgeAdapter {
   ): Promise<BridgeEvents> {
     const events = await paginatedEventQuery(
       this.getL1Bridge(),
-      this.getL1Bridge().filters.BridgingInitiatedV2(undefined, fromAddress, l1Token),
```
nicholaspai, line 45: this is a very subtle change can you explain why we're switching `from` to `to`? Without any context this looks like a bug

bmzig: A short, hand-wavey explanation is that this was formerly bad naming and this PR fixes it. Since transfers can theoretically be sent from a random address to a spoke pool or EOA from L1, we want to query the inbound L2 transfers and not the outbound L1 transfers (since, for example, outbound L1 transfers may be sent to a different address as is with zkSync, or once again, we might miss an inbound L2 transfer if it was sent from an address we do not monitor). 

A longer justification is that in our adapters, `monitoredAddresses` is tossed around, and we never really distinguish between whether it is an L1 address or an L2 address (which matters in some cases, e.g. spoke pools and zkSync). For example, we call these functions in the adapters here: https://github.com/across-protocol/relayer/blob/db2e4a470c357d911cf1c1292cddfb838f0f18a8/src/adapter/BaseChainAdapter.ts#L239 (both toAddress and fromAddress are the same thing: an address in `this.monitoredAddresses`). Note we also do this in prod (https://github.com/across-protocol/relayer/blob/db2e4a470c357d911cf1c1292cddfb838f0f18a8/src/clients/bridges/LineaAdapter.ts#L149). The intended functionality is to pass in what addresses we need to monitor into the bridge and have the bridge sort everything out for us, so in this case, since I wanted to classify outstanding transfers as funds coming from **any** L1 source into an L2 address, I need to filter events based on the the `toAddress` and not the `fromAddress` (once again, `fromAddress` === `toAddress`, so ultimately this was a naming fix).

src/adapter/bridges/LineaUSDCBridge.ts
```diff
@@ -58,7 +58,7 @@ export class LineaUSDCBridge extends BaseBridgeAdapter {
   ): Promise<BridgeEvents> {
     const events = await paginatedEventQuery(
       this.getL2Bridge(),
-      this.getL2Bridge().filters.ReceivedFromOtherLayer(fromAddress),
+      this.getL2Bridge().filters.ReceivedFromOtherLayer(toAddress, undefined),
```
james-a-morris, line 61:
```suggestion
      this.getL2Bridge().filters.ReceivedFromOtherLayer(toAddress),
```

src/adapter/bridges/LineaWethBridge.ts
```diff
@@ -46,7 +46,7 @@ export class LineaWethBridge extends BaseBridgeAdapter {
   ): Promise<BridgeEvents> {
     const events = await paginatedEventQuery(
       this.getL1Bridge(),
-      this.getL1Bridge().filters.MessageSent(undefined, fromAddress),
+      this.getL1Bridge().filters.MessageSent(undefined, toAddress, undefined),
```
james-a-morris, line 49:
```suggestion
      this.getL1Bridge().filters.MessageSent(undefined, toAddress),
```

src/adapter/bridges/LineaWethBridge.ts
```diff
@@ -66,7 +66,7 @@ export class LineaWethBridge extends BaseBridgeAdapter {
     // queryL1BridgeInitiationEvents in the BaseChainAdapter class.
     const initiatedQueryResult = await paginatedEventQuery(
       this.getL1Bridge(),
-      this.getL1Bridge().filters.MessageSent(undefined, fromAddress),
+      this.getL1Bridge().filters.MessageSent(undefined, toAddress, undefined),
```
james-a-morris, line 69:
```suggestion
      this.getL1Bridge().filters.MessageSent(undefined, toAddress),
```

bmzig:
We should start considering merging this.

james-a-morris (commented):
lgtm pending signoff from @pxrl

bmzig:
I've tested USDT, WBTC (both which use `LineaBridge`), WETH (which uses `LineaWethBridge`), and USDC.e (uses `LineaUSDCBridge`), and verified that cross-chain transfer tracking works.

src/adapter/bridges/LineaBridge.ts
pxrl: I agree with this description. In general I think we should try to nudge towards the reality of monitoredAddresses being only the recipient addresses.

src/clients/bridges/LineaAdapter.ts
```diff
@@ -1,417 +1,32 @@
...
-      };
+    const { LINEA, MAINNET } = CHAIN_IDs;
+    const bridges = {};
+    const l2Signer = spokePoolClients[LINEA].spokePool.signer;
+    const l1Signer = spokePoolClients[MAINNET].spokePool.signer;
+    SUPPORTED_TOKENS[LINEA]?.map((symbol) => {
+      const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[MAINNET];
+      const bridgeConstructor = CUSTOM_BRIDGE[LINEA]?.[l1Token] ?? CANONICAL_BRIDGE[LINEA];
+      bridges[l1Token] = new bridgeConstructor(LINEA, MAINNET, l1Signer, l2Signer, l1Token);
     });
```
pxrl, line 20: For future improvements (NCA-related), we could update this constructor to accept a hubChainId. It should then be able to use that instead of `MAINNET`, and could infer the correct chainId to use (rather than hardcoding `LINEA`). Then we'd be a step closer to supporting this on testnet.

pxrl:
> I've tested USDT, WBTC (both which use `LineaBridge`), WETH (which uses `LineaWethBridge`), and USDC.e (uses `LineaUSDCBridge`), and verified that cross-chain transfer tracking works.

Just to confirm, did this also include initiating the deposits with this new adapter code?

src/clients/bridges/LineaAdapter.ts
bmzig: Fwiw we will do this once all the adapters switch over https://github.com/across-protocol/relayer/blob/5af02ba2100fa1d943052530b251b1cb99a8466b/src/adapter/AdapterManager.ts#L50. If it takes a long time, though, I would understand the value in modifying these constructors beforehand.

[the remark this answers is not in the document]
pxrl: Sorry, this was garbage. It risks calling `Object.fromEntries(undefined)` which can throw. What you had was better; I'd probably swap `.map()` for `.forEach()` though.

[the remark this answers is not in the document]
bmzig: Done here https://github.com/across-protocol/relayer/pull/1714/commits/c2b5dff7689348f0a010667e79817852982bd47f

pxrl (approved):

bmzig:
I've tested this again by rebalancing all tokens and running the monitor. Looks good to go in after this https://github.com/across-protocol/relayer/pull/1768 or something similar gets merged.

src/adapter/bridges/LineaWethBridge.ts
```diff
@@ -62,12 +72,23 @@ export class LineaWethBridge extends BaseBridgeAdapter {
...
   ): Promise<BridgeEvents> {
-    // TODO: This can probably be refactored to save an RPC call since this is called in parallel with
-    // queryL1BridgeInitiationEvents in the BaseChainAdapter class.
+    const l2Provider = this.getL2Bridge().provider;
+    const l1Provider = this.getL1Bridge().provider;
+
+    const fromTimestamp = (await l2Provider.getBlock(eventConfig.fromBlock)).timestamp;
+    const toTimestamp = (await l2Provider.getBlock(eventConfig.toBlock)).timestamp;
+
+    const blockFinder = new BlockFinder(l1Provider);
```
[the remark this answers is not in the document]
bmzig: I did. The issue with that is the block finder returned there fails tests since it fetches a provider for mainnet and not the hardhat test network. We do a bit of mixing in our tests with chain ids, and it's weird to see which chain each provider thinks they are on verses for which one they provide information.

src/adapter/bridges/LineaWethBridge.ts
```diff
@@ -62,12 +72,23 @@ export class LineaWethBridge extends BaseBridgeAdapter {
...
+    const l1FromBlock = await getBlockForTimestamp(this.hubChainId, fromTimestamp, blockFinder);
+    const l1ToBlock = await getBlockForTimestamp(this.hubChainId, toTimestamp, blockFinder);
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/1714/commits/7b7e6aabb772d4df48bba56653569d6a7d203324

src/adapter/bridges/LineaWethBridge.ts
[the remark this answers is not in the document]
pxrl: > it's weird to see which chain each provider thinks they are on verses for which one they provide information.

Yeah - I think this is a limitation imposed by hardhat, whereby it only runs a single chain. This means that all instances on top have to be tricked into thinking that they have different chain IDs. It does create some annoying complexity and inconsistencies.

src/adapter/bridges/LineaWethBridge.ts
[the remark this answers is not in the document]
pxrl: @bmzig Sorry, I gave bad input here. For the BlockFinder specifically I think it's worth serialising them, because otherwise they'll both do duplicate work. The BlockFinder retains block history, so the 2nd run should be a bit faster than the first because it inherits the block timestamps that were identified during the first run.

src/adapter/bridges/LineaWethBridge.ts
```diff
@@ -62,12 +72,27 @@ export class LineaWethBridge extends BaseBridgeAdapter {
...
+    const [fromBlock, toBlock] = await Promise.all([
+      l2Provider.getBlock(eventConfig.fromBlock),
+      l2Provider.getBlock(eventConfig.toBlock),
+    ]);
+
+    const blockFinder = new BlockFinder(l1Provider);
+    const [l1FromBlock, l1ToBlock] = await Promise.all([
+      getBlockForTimestamp(this.hubChainId, fromBlock.timestamp, blockFinder),
+      getBlockForTimestamp(this.hubChainId, toBlock.timestamp, blockFinder),
+    ]);
```
pxrl, line 87:
```suggestion
    const [l1FromBlock, l1ToBlock] = [
      await getBlockForTimestamp(this.hubChainId, fromBlock.timestamp, blockFinder),
      await getBlockForTimestamp(this.hubChainId, toBlock.timestamp, blockFinder),
    ];
```

src/adapter/bridges/LineaWethBridge.ts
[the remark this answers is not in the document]
pxrl: I just reverted that part here: https://github.com/across-protocol/relayer/pull/1714#discussion_r1746873442

pxrl:
@bmzig Just a quick confirmation - does it still test OK on chain after the latest changes?

src/adapter/bridges/LineaWethBridge.ts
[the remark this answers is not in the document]
bmzig: I worked through the block finder thing here. It will now be undefined so it will fallback to `getBlockFinder` in prod, but we can always define it in tests https://github.com/across-protocol/relayer/pull/1714/commits/2ea24eba60fdeda40e60b9a1d6515a5c825d4476

bmzig:
> @bmzig Just a quick confirmation - does it still test OK on chain after the latest changes?

It does work. Should be fine to merge in on Monday.

PR #1716, opened 2024-07-29
improve(relayer): Support deferring chain fills

pxrl:
For certain chains with long average block times, one valid strategy for avoiding collisions is to defer filling until other relayers have had their first shot at making the fill. This can alternatively be solved by running separate relayers, but that imposes quite a lot of overhead for something that can otherwise be simply configured via the env.


src/relayer/Relayer.ts
```diff
@@ -101,6 +101,41 @@ export class Relayer {
...
+    const isExclusive = getAddress(deposit.exclusiveRelayer) === this.relayerAddress;
+    if (deposit.exclusivityDeadline > currentTime && !isExclusive) {
+      return false;
+    }
+
+    // If the operator configured a minimum fill time for a destination chain, ensure that the deposit
+    // is at least that old before filling it. This is mainly useful on chains with long block times,
+    // where there is a high chance of fill collisions in the first blocks after a deposit is made.
+    const minFillTime = this.config.minFillTime[destinationChainId] ?? 0;
+    if (minFillTime > 0 && !isExclusive) {
```
pxrl, line 119: The anticipated case here is that the minFillTime > 0 for mainnet, so for most deposits the overhead of this computation is not incurred.

src/relayer/Relayer.ts
```diff
@@ -101,6 +101,41 @@ export class Relayer {
...
+      const originSpoke = spokePoolClients[originChainId];
+      const timeDelta = originSpoke.getCurrentTime() - originSpoke.getOldestTime();
+      const blockDelta = originSpoke.latestBlockSearched - originSpoke.deploymentBlock;
+      const avgBlockTime = Math.floor(timeDelta / blockDelta);
+      const depositAge = avgBlockTime * (originSpoke.latestBlockSearched - blockNumber);
+
+      if (minFillTime > depositAge) {
+        console.log(`XXXX Skipping mainnet fill for ${srcChain} deposit ${depositId} (${minFillTime} > ${depositAge})`);
```
pxrl, line 127:
```suggestion
```

src/relayer/Relayer.ts
```diff
@@ -101,6 +101,40 @@ export class Relayer {
...
+    }
+
+    // If the operator configured a minimum fill time for a destination chain, ensure that the deposit
+    // is at least that old before filling it. This is mainly useful on chains with long block times,
+    // where there is a high chance of fill collisions in the first blocks after a deposit is made.
+    const minFillTime = this.config.minFillTime?.[destinationChainId] ?? 0;
+    if (minFillTime > 0 && !isExclusive) {
+      const originSpoke = spokePoolClients[originChainId];
+      const timeDelta = originSpoke.getCurrentTime() - originSpoke.getOldestTime();
+      const blockDelta = originSpoke.latestBlockSearched - originSpoke.deploymentBlock;
```
nicholaspai, line 122: Deployment block is pretty early. And is that even that block that corresponds with getOldestTime()? I would think it should be the searchConfig.fromBlock here

src/relayer/Relayer.ts
```diff
@@ -101,6 +101,40 @@ export class Relayer {
...
+      const avgBlockTime = timeDelta / blockDelta;
+      const depositAge = Math.floor(avgBlockTime * (originSpoke.latestBlockSearched - blockNumber));
+
+      if (minFillTime > depositAge) {
```
nicholaspai, line 126: Why can't you compare originClient.getCurrentTime - originClient.deposit.block timestamp?

nicholaspai (commented):
Left a q on the time computation

src/relayer/Relayer.ts
pxrl: We only have the block number, not the actual block, so we don't know what the deposit timestamp was. The only timestamp we have is the quoteTimestamp, but that can span over a 1 hour period. So trying to complute the "average" block time is an effort to avoid the RPC query for blockNumber -> block.

There is an alternative. We could instantiate the BlockFinder and use it. It should be primed with average block times for each chain, but they will expire before the bot runtime ends, so would at some point impose a blocking call to resolve a new average block time (...on the origin chain, which is undesirable from the main logic loop because it can slow it down). But still - I think this might be a better approach because there are subsequent optimisations that can be done to improve on that.

pxrl: Updated to just use `averageBlockTime()` from the SDK. The cache for this should already be primed by having resolved start blocks for each SpokePoolClient and the results are retained for 15 minutes, so there should be no RPC queries required in order to resolve the average time.

src/relayer/Relayer.ts
pxrl: This should be irrelevant after this update: https://github.com/across-protocol/relayer/pull/1716#discussion_r1697451277

nicholaspai (approved):

PR #1719, opened 2024-07-30
fix(relayer): Gracefully handle partial inventory config

pxrl:
Inventory management reacts badly if it tries to resolve a missing inventory config for a particular chain. This makes the bot unnecessarily fragile.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -545,7 +545,7 @@ export class InventoryClient {
         if (chainId === destinationChainId) {
           this.logger.debug({
             at: "InventoryClient#determineRefundChainId",
-            message: `Defaulting to repayment on ${repaymentChain}.`,
+            message: `Will consider to repayment on ${repaymentChain} as destination chain.`,
           });
           eligibleRefundChains.push(chainId);
         }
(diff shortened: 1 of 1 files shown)


src/clients/InventoryClient.ts
```diff
@@ -531,10 +536,16 @@ export class InventoryClient {
...
       // Consider configured buffer for target to allow relayer to support slight overages.
       const tokenConfig = this.getTokenConfig(l1Token, chainId, repaymentToken);
-      assert(
-        isDefined(tokenConfig),
-        `No ${outputToken} tokenConfig for ${l1Token} on ${chainId} with a repaymentToken ${repaymentToken}.`
-      );
+      if (!isDefined(tokenConfig)) {
+        this.logger.debug({
+          at: "InventoryClient#determineRefundChainId",
+          message: `No token config for ${repaymentToken} on ${getNetworkName(chainId)}.`,
```
nicholaspai, line 542: can you add to this log message that you're selecting destination chain by default

nicholaspai (approved):

src/clients/InventoryClient.ts
pxrl: Fair call. It has to be a separate log message; done here: https://github.com/across-protocol/relayer/pull/1719/commits/7446fece5fbe148809ae8bca4011538da4def0d3

PR #1720, opened 2024-07-30
improve(bridges): add special USDC case for the ScrollBridge.

bmzig:
This is a simple fix to the issue we had for resolving a USDC address for scroll, whereby we would include a special case in `resolveL2TokenAddress` in BaseBridgeAdapter. This moves this logic to the Scroll bridge so this conditional won't be triggered by all bridges.

### DIFF
--- a/src/adapter/bridges/ScrollERC20Bridge.ts
+++ b/src/adapter/bridges/ScrollERC20Bridge.ts
@@ -9,6 +9,8 @@ import {
   fixedPointAdjustment,
   isContractDeployedToAddress,
   bnZero,
+  compareAddressesSimple,
+  TOKEN_SYMBOLS_MAP,
 } from "../../utils";
 import { CONTRACT_ADDRESSES, SCROLL_CUSTOM_GATEWAY } from "../../common";
 import { BaseBridgeAdapter, BridgeTransactionDetails, BridgeEvents } from "./BaseBridgeAdapter";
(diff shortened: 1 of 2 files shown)


james-a-morris (commented):
Is this change a refactor?

bmzig:
> Is this change a refactor?

Essentially. It's a fix for the todo brought up in this PR https://github.com/across-protocol/relayer/pull/1708

src/adapter/bridges/ScrollERC20Bridge.ts
```diff
@@ -127,4 +129,11 @@ export class ScrollERC20Bridge extends BaseBridgeAdapter {
   protected getScrollGatewayRouter(): Contract {
     return this.scrollGatewayRouter;
   }
+
+  protected override resolveL2TokenAddress(l1Token: string): string {
+    if (compareAddressesSimple(TOKEN_SYMBOLS_MAP.USDC.addresses[this.hubChainId], l1Token)) {
+      return TOKEN_SYMBOLS_MAP.USDC.addresses[this.l2chainId]; // Scroll only has one USDC token type.
+    }
+    return super.resolveL2TokenAddress(l1Token);
```
james-a-morris, line 137: Good call. Thoughts on
```suggestion
  protected override resolveL2TokenAddress(l1Token: string): string {
  
    if () {
      return TOKEN_SYMBOLS_MAP.USDC.addresses[this.l2chainId]; // Scroll only has one USDC token type.
    }
    return super.resolveL2TokenAddress(l1Token);
```

james-a-morris (approved):

pxrl (approved):
Nice and clean!

PR #1721, opened 2024-07-30
improve(SpokePoolClient): Support restart of failed listener

pxrl:
In the event that a child process terminates improperly, support restarting it with its initial lookback.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -79,6 +79,7 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
    * @returns void
    */
   protected startWorker(): void {
+    // fromBlock is retained for the life of the SpokePoolClient and is reused in case of listener crash.
     const {
       finality,
       eventSearchConfig: { fromBlock, maxBlockLookBack: blockRange },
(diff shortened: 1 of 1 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -100,6 +102,50 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
...
+
+    this.logger.warn({
+      at: "SpokePoolClient#childExit",
+      message: `${this.chain} SpokePool listener exited.`,
+      code,
+      signal
+    });
+
+    // Flush all ingested deposits to protect against re-org.
+    // xxx this probably belongs upstream in a `protected SpokePoolClient.init()` method
```
nicholaspai, line 125: agreed with putting this into `init()` to avoid any missed line items

src/clients/SpokePoolClient.ts
```diff
@@ -100,6 +102,50 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
...
+    // that is called in the constructor, such that it can be re-called by this method to
+    // re-initialise the same defaults.
```
pxrl, line 127: I will upstream this to the SDK.

nicholaspai (approved):
approving for you to merge quickly if needed

pxrl:
> approving for you to merge quickly if needed

:+1: There's no burning need for it right now, but it's going to be important as we add more chains with more questionable websocket reliability. I'll upstream the init method first and then will circle back to this one.

src/clients/SpokePoolClient.ts
```diff
@@ -100,6 +102,50 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
...
 
+  /**
+   * The worker process has exited. Optionally restart it based on the exit code.
+   * See also: https://nodejs.org/api/child_process.html#event-exit
+   * @param code Optional exit code.
+   * @param signal Optional signal resulting in termination.
+   * @returns void
+   */
+  protected childExit(code?: number, signal?: string): void {
+    if (code === 0) {
```
james-a-morris, line 113: Is there any system defined constants (like `EXIT_SUCCESS` in the [stdlib](https://www.man7.org/linux/man-pages/man3/exit.3.html) for C) that we can use instead of `0`?

src/clients/SpokePoolClient.ts
james-a-morris: +1

PR #1722, opened 2024-07-31
improve(finalizer): update scroll finalizer with new api

bmzig:
The current finalizer is outdated since it fetches info from an API which does not exist. This commit updates the URL to use the one which is currently maintained by Scroll.

### DIFF
--- a/src/finalizer/utils/scroll.ts
+++ b/src/finalizer/utils/scroll.ts
@@ -4,17 +4,25 @@ import { TransactionRequest } from "@ethersproject/abstract-provider";
 import axios from "axios";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
 import { CONTRACT_ADDRESSES, Multicall2Call } from "../../common";
-import { Contract, Signer, getBlockForTimestamp, getCurrentTime, getRedisCache, winston, BigNumber } from "../../utils";
+import { Contract, Signer, getBlockForTimestamp, getCurrentTime, getRedisCache, winston } from "../../utils";
 import { FinalizerPromise, CrossChainMessage } from "../types";
 
+// The value in ScrollClaimInfo is 0 unless it is a Weth withdrawal, in which case, value === token_amount. Essentially, `token_amount` is used to format messages properly, while
+// `value` is used to construct valid withdrawal proofs.
 type ScrollClaimInfo = {
   from: string;
   to: string;
-  value: BigNumber;
-  nonce: BigNumber;
+  value: string;
+  nonce: string;
   message: string;
-  proof: string;
-  batch_index: BigNumber;
+  proof: Proof;
+  claimable: boolean;
+  token_amount: string;
+};
+
+type Proof = {
+  batch_index: string;
+  merkle_proof: string;
 };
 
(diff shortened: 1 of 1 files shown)


bmzig:
Marked as do not merge since I'm getting a strange invalid argument error when trying to encode function data at `populateClaimTransaction`.

bmzig:
Ready for review. I tested by finalizing transactions with an EOA.

src/finalizer/utils/scroll.ts
```diff
@@ -68,35 +73,37 @@ export async function scrollFinalizer(
...
+          results: {
+            claim_info: ScrollClaimInfo;
+            l1_token_address: string;
+            token_amounts: string[];
           }[];
         };
       }>(apiUrl, {
         params: {
           address: targetAddress,
+          page_size: 10,
```
nicholaspai, line 96: What if there are more than 10 withdrawals for our target address in a run? Will that be an issue? Or will we capture them on the next run?

src/finalizer/utils/scroll.ts
```diff
@@ -39,15 +47,12 @@ export async function scrollFinalizer(
   // Why are we breaking with the existing pattern--is it faster?
   // Scroll takes up to 4 hours with finalize a withdrawal so lets search
   // up to 12 hours for withdrawals.
-  const lookback = getCurrentTime() - 12 * 60 * 60 * 24;
```
nicholaspai, line 42: Why remove the filter? Just curious if you had a reason

nicholaspai (commented):
couple q's

src/finalizer/utils/scroll.ts
bmzig: Since the API just returns all pending/finalizable withdrawals anyways, I don't think it's necessary or even a good idea to filter based on time. Basically, if there is a  withdrawal that can be finalized, we should try to finalize it.

src/finalizer/utils/scroll.ts
bmzig: I now just request the maximum amount of info in a current page. https://github.com/across-protocol/relayer/pull/1722/commits/38b70d5cfebe1bc324fa644c08f600732e0be8ee. Although something to consider: It looks like the response is formatted such that the withdrawals which can be finalized are at the end, and the withdrawals which are still pending finalization on L1 are at the beginning, so theoretically, if we have >= 100 pending/non-finalizable withdrawals on Scroll, this won't work. I don't think that's an issue, since finalization on Scroll is fast, and I doubt we would have this many withdrawals occur within 4 hours, but maybe there is something we want to do here?

nicholaspai: > if we have >= 100 pending/non-finalizable withdrawals on Scroll, this won't work

Yea and this is realistic, though unlikely.

Can we somehow filter out any withdrawals in the pending state?

bmzig: Should be good now: https://github.com/across-protocol/relayer/pull/1722/commits/917e8e550973c3b0bce202b1675e0b460424df7e

nicholaspai (approved):

PR #1725, opened 2024-08-01
improve(Monitor): Ignore chains without pool rebalance leaves when logging about "stuck" rebalances

nicholaspai:
Fixes ACX-2433

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -564,7 +564,7 @@ export class Monitor {
   // should stay unstuck for longer than one bundle.
   async checkStuckRebalances(): Promise<void> {
     const hubPoolClient = this.clients.hubPoolClient;
-    const { currentTime } = hubPoolClient;
+    const { currentTime, latestBlockSearched } = hubPoolClient;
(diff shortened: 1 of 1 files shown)


bmzig (approved):

src/monitor/Monitor.ts
```diff
@@ -573,7 +573,16 @@ export class Monitor {
     const lastFullyExecutedBundleTime = lastFullyExecutedBundle.challengePeriodEndTimestamp;
 
     const allL1Tokens = this.clients.hubPoolClient.getL1Tokens();
+    const poolRebalanceLeaves = this.clients.hubPoolClient.getExecutedLeavesForRootBundle(
+      lastFullyExecutedBundle,
+      latestBlockSearched
+    );
     for (const chainId of this.crossChainAdapterSupportedChains) {
```
pxrl, line 580: So in this case we'll warn for max ~1 hour, and less as the challenge window comes down. For follow-up, do we want to instead repeat this process over the past few bundles? Or should we instead make it a more visible when this happens?

pxrl (approved):

PR #1726, opened 2024-08-02
improve(finalizer): cache attestation responses from Circle's API

bmzig:
The finalizer sends too many requests to Circle's attestation API, which sometimes causes it to error out. This update will allow the finalizer to cache attestations which it previously queried, which should reduce the number of total queries the finalizer makes to the API over multiple runs.

### DIFF
--- a/src/utils/CCTPUtils.ts
+++ b/src/utils/CCTPUtils.ts
@@ -277,7 +277,8 @@ async function generateCCTPAttestationProof(messageHash: string, isMainnet: bool
       `https://iris-api${isMainnet ? "" : "-sandbox"}.circle.com/attestations/${messageHash}`
     );
     attestationResponse = httpResponse.data;
-    await redisCache.set(messageHash, attestationResponse, 60 * 60 * 24); // Key expires after one day.
+    const randomInterval = () => Math.floor(Math.random() * (4 * 86400)) + 4 * 86400; // 60 * 60 * 4 = 86400 = 1 day.
+    await redisCache.set(messageHash, attestationResponse, randomInterval()); // Cache expires at random second anywhere between 4-7 days.
   }
   return attestationResponse;
 }
--- a/src/utils/CCTPUtils.ts
+++ b/src/utils/CCTPUtils.ts
@@ -8,6 +8,7 @@ import { EventSearchConfig, paginatedEventQuery } from "./EventUtils";
 import { bnZero } from "./SDKUtils";
 import { isDefined } from "./TypeGuards";
 import { getProvider } from "./ProviderUtils";
+import { getRedisCache } from "./RedisUtils";
 
 export type DecodedCCTPMessage = {
   messageHash: string;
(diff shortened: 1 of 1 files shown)


src/utils/CCTPUtils.ts
```diff
@@ -269,9 +270,14 @@ async function _resolveCCTPRelatedTxns(
  * @link https://developers.circle.com/stablecoins/reference/getattestation
  */
 async function generateCCTPAttestationProof(messageHash: string, isMainnet: boolean): Promise<Attestation> {
-  const httpResponse = await axios.get<Attestation>(
-    `https://iris-api${isMainnet ? "" : "-sandbox"}.circle.com/attestations/${messageHash}`
-  );
-  const attestationResponse = httpResponse.data;
+  const redisCache = await getRedisCache();
+  let attestationResponse = await redisCache.get<Attestation>(messageHash);
```
nicholaspai, line 274: Are these guaranteed to be the same for each `messageHash`?

nicholaspai (approved):
Great idea! https://github.com/across-protocol/relayer/pull/1726/files#r1702027947 just want confirmation here

src/utils/CCTPUtils.ts
```diff
@@ -269,9 +270,14 @@ async function _resolveCCTPRelatedTxns(
...
+  if (!isDefined(attestationResponse)) {
+    const httpResponse = await axios.get<Attestation>(
+      `https://iris-api${isMainnet ? "" : "-sandbox"}.circle.com/attestations/${messageHash}`
+    );
+    attestationResponse = httpResponse.data;
+    await redisCache.set(messageHash, attestationResponse, 60 * 60 * 24); // Key expires after one day.
```
nicholaspai, line 280: Thoughts on expanding this TTL to something like 7 days and also making the expiry random over a large time window so we don't get mass ejections each period?

src/utils/CCTPUtils.ts
bmzig: I believe so, since the message hash is derived from the message being sent, which includes a nonce (which is a state variable in the smart contract): https://github.com/circlefin/evm-cctp-contracts/blob/377c9bd813fb86a42d900ae4003599d82aef635a/src/MessageTransmitter.sol#L353. I think this should guarantee that there should be a one to one mapping from message hashes to attestations.

src/utils/CCTPUtils.ts
bmzig: Done here: https://github.com/across-protocol/relayer/pull/1726/commits/2af5df26a5853732d030400cc44f40a0427ce589

james-a-morris (commented):
OOC do we need this? AFAICT we are able to glean whether or not a CCTP message is completed prior to calling the attestation

james-a-morris:
Thoughts on this? https://github.com/across-protocol/relayer/pull/1728

bmzig:
> Thoughts on this? #1728

Looks good to me. I think we should do as much as we can to reduce the number of API calls.

bmzig:
Closing this PR in favor of https://github.com/across-protocol/relayer/pull/1728. Can reopen in the event that we still have issues with rate limits.

PR #1727, opened 2024-08-02
improve(finalizer): cache transaction receipts

bmzig:
Currently, for some chains such as Linea, the number of requests we make to an RPC scales with how many cross chain transfers we make ([example](https://github.com/across-protocol/relayer/blob/8129687f511dd817dfbc3c7c1d1245eee2bdd4b2/src/finalizer/utils/linea/l2ToL1.ts#L46)). Therefore, for cases when we have a large amount of outstanding transfers at once, we could get rate limited by our RPC providers. This PR caches the `eth_getTransactionReceipt` method, which we observed to be the method called during rate limit errors. This required a "hack" since this method has no block range as input, so we cannot classify whether or not it should be stored with/without a TTL, but since we know the context of which this method is called, specifically for currently outstanding transfers, we can assume the receipt will not have to be stored for a long time and can optimistically assign it to have the standard TTL.

### DIFF
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -251,7 +251,7 @@ class CacheProvider extends RateLimitedProvider {
     }
   }
 
-  private async cacheType(method: string, params: Array<any>, result: Array<any>): Promise<CacheType> {
+  private async cacheType(method: string, params: Array<any>, result: any): Promise<CacheType> {
     // Today, we only cache eth_getLogs and eth_call.
     if (method === "eth_getLogs") {
       const [{ fromBlock, toBlock }] = params;
@@ -289,7 +289,7 @@ class CacheProvider extends RateLimitedProvider {
       // The RPC method `eth_getTransactionReceipt` has no block range in its parameters, so we cannot assign a cache type based off of how far back the request looks. Instead,
       // We look at which block the transaction was mined. This should allow us to not cache a transaction which may be susceptible to a reorg.
       // The only parameter is `hash` which is the transaction hash to get a receipt for.
-      const [{ blockNumber }] = result;
+      const { blockNumber } = result;
       const minedBlock = parseInt(blockNumber, 16);
       return this.cacheTypeForBlock(minedBlock);
     } else {
--- a/src/utils/ProviderUtils.ts
+++ b/src/utils/ProviderUtils.ts
@@ -289,7 +289,7 @@ class CacheProvider extends RateLimitedProvider {
       // The RPC method `eth_getTransactionReceipt` has no block range in its parameters, so we cannot assign a cache type based off of how far back the request looks. Instead,
       // We look at which block the transaction was mined. This should allow us to not cache a transaction which may be susceptible to a reorg.
       // The only parameter is `hash` which is the transaction hash to get a receipt for.
-      const { blockNumber } = result;
+      const [{ blockNumber }] = result;
       const minedBlock = parseInt(blockNumber, 16);
       return this.cacheTypeForBlock(minedBlock);
     } else {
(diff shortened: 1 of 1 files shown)


src/utils/ProviderUtils.ts
```diff
@@ -281,6 +285,14 @@ class CacheProvider extends RateLimitedProvider {
 
       // If the block is old enough to cache, cache the call.
       return this.cacheTypeForBlock(blockNumber);
+    } else if (method === "eth_getTransactionReceipt") {
+      // It's possible to be rate-limited by RPCs during finalization runs, specifically for chains which require
```
nicholaspai, line 289: I think this comment is not useful at this level of the code. We are making a choice to cache this result and we should not care exactly how this request is being used. Instead we should just say simply that we are going to assume that we can't easily get more information about the block number of the single parameter `hash` so we are just going to assume that there should be a TTL.

bmzig: You're right. https://github.com/across-protocol/relayer/pull/1727/commits/32f6385f125197fe3c38466edf7dabfa91f2bb56

james-a-morris:
OOC will this cause any unintended effects on our caching bandwidth? My understanding is that we're going to be caching all transaction receipts if we enable it

src/utils/ProviderUtils.ts
```diff
@@ -281,6 +285,9 @@ class CacheProvider extends RateLimitedProvider {
 
       // If the block is old enough to cache, cache the call.
       return this.cacheTypeForBlock(blockNumber);
+    } else if (method === "eth_getTransactionReceipt") {
+      // The RPC method `eth_getTransactionReceipt` has no block range in its parameters, so we cannot assign a cache type based off of how far back the request looks. Here, we assume that there should be a TTL, which means that we will primarily reap the benefits of caching when this method is called to query recent transactions.
+      return CacheType.WITH_TTL;
```
james-a-morris, line 290: Will this cause any unintended effects on our caching bandwidth? My understanding is that we're going to be caching all transaction receipts if we enable it

james-a-morris (commented):
Any chance we can avoid caching and maybe impose some limiting at the call level? Maybe something we could call as a fill-in for `Promise.all`

src/utils/ProviderUtils.ts
nicholaspai: we don't make this call very often though

src/utils/ProviderUtils.ts
```diff
@@ -281,6 +285,9 @@ class CacheProvider extends RateLimitedProvider {
 
       // If the block is old enough to cache, cache the call.
       return this.cacheTypeForBlock(blockNumber);
+    } else if (method === "eth_getTransactionReceipt") {
+      // The RPC method `eth_getTransactionReceipt` has no block range in its parameters, so we cannot assign a cache type based off of how far back the request looks. Here, we assume that there should be a TTL, which means that we will primarily reap the benefits of caching when this method is called to query recent transactions.
```
nicholaspai, line 289:
```suggestion
      // The RPC method `eth_getTransactionReceipt` has no block range in its parameters, so we cannot assign a cache type based off of how far back the request looks. Here, we assume that there should be a TTL, which means that we will primarily reap the benefits of caching when this method is called to query recent transactions.
      // The only parameter is `hash` which is the transaction hash to get a receipt for.
```

nicholaspai (approved):

src/utils/ProviderUtils.ts
bmzig: FWIW we only call `getTransactionReceipts` in the Linea, CCTP, and Arbitrum finalizers. We want the caching for the faster finalizers (mainly Linea). I can see why we don't want to cache Arbitrum receipts, but each slot should match a single hash -> receipt, so this should still be fine.

bmzig:
Paul brought up a good point where since we just assume a TTL and cache, we may cache a transaction which in the future gets invalidated by a reorg. I changed the order of `send` so that it first checks if a request is cached, and if not, sends the request and caches the result. This way, we can use the result to determine the block at which the transaction was mined to properly cache the transaction.

pxrl:
After nudging Infura on this one it seems like this is actually rate-limiting on their internal network that was propagated back to the bot. So the bot requesting transaction receipts hasn't really been the cause of rate-limiting - the aggregate demand for one of their internal hosts was moreso what caused it. Given that I'm probably OK with leaving the existing code as-is.

bmzig:
Closing. See this comment https://github.com/across-protocol/relayer/pull/1727#issuecomment-2270021987

PR #1728, opened 2024-08-02
improve: prevent attestations on finalized messages

james-a-morris:
Prompted by: https://github.com/across-protocol/relayer/pull/1726

We already throw out pending/finalized messages for CCTP. Because of this, we can avoid calling Circle's attestation API unless we know there's a chance we need to use the attestation. I.e. when a message is already finalized (which most of our finalizer runs will be in the scope of)


nicholaspai (approved):
Very. nice LGTM

bmzig (approved):

PR #1734, opened 2024-08-07
improve(any): remove implicit any

james-a-morris:
Preliminary PR prior to migrating any caching provider logic into the SDK

### DIFF
--- a/src/utils/ObjectUtils.ts
+++ b/src/utils/ObjectUtils.ts
@@ -100,7 +100,7 @@ export function count3DDictionaryValues(
  * @param obj
  * @returns Objects with ignored keys removed
  */
-function deleteIgnoredKeys(ignoredKeys: string[], obj: any) {
+function deleteIgnoredKeys(ignoredKeys: string[], obj: Record<string, unknown>) {
   if (!isDefined(obj)) {
     return;
   }
@@ -111,7 +111,11 @@ function deleteIgnoredKeys(ignoredKeys: string[], obj: any) {
   return newObj;
 }
 
-export function compareResultsAndFilterIgnoredKeys(ignoredKeys: string[], _objA: any, _objB: any): boolean {
+export function compareResultsAndFilterIgnoredKeys(
+  ignoredKeys: string[],
+  _objA: Record<string, unknown>,
+  _objB: Record<string, unknown>
+): boolean {
   // Remove ignored keys from copied objects.
   const filteredA = deleteIgnoredKeys(ignoredKeys, _objA);
   const filteredB = deleteIgnoredKeys(ignoredKeys, _objB);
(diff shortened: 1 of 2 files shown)


src/utils/ObjectUtils.ts
```diff
@@ -100,7 +100,7 @@ export function count3DDictionaryValues(
  * @param obj
  * @returns Objects with ignored keys removed
  */
-function deleteIgnoredKeys(ignoredKeys: string[], obj: any) {
+function deleteIgnoredKeys(ignoredKeys: string[], obj: Record<string, unknown>) {
```
james-a-morris, line 103: `Record<string, unknown>` is considered `any type of object`

daywiss (approved):
lgtm

src/utils/ProviderUtils.ts
```diff
@@ -118,12 +117,12 @@ function formatProviderError(provider: ethers.providers.StaticJsonRpcProvider, r
   return `Provider ${provider.connection.url} failed with error: ${rawErrorText}`;
 }
 
-function createSendErrorWithMessage(message: string, sendError: any) {
+function createSendErrorWithMessage(message: string, sendError: unknown) {
```
nicholaspai, line 120: why not make `sendError` of type `Record`?'

nicholaspai (approved):
lgtm, just one clarifying q

src/utils/ObjectUtils.ts
```diff
@@ -120,10 +124,10 @@ export function compareResultsAndFilterIgnoredKeys(ignoredKeys: string[], _objA:
   return lodash.isEqual(filteredA, filteredB);
 }
 
-export function compareArrayResultsWithIgnoredKeys(ignoredKeys: string[], objA: any[], objB: any[]): boolean {
+export function compareArrayResultsWithIgnoredKeys(ignoredKeys: string[], objA: unknown[], objB: unknown[]): boolean {
   // Remove ignored keys from each element of copied arrays.
-  const filteredA = objA?.map((obj) => deleteIgnoredKeys(ignoredKeys, obj));
-  const filteredB = objB?.map((obj) => deleteIgnoredKeys(ignoredKeys, obj));
+  const filteredA = objA?.map((obj) => deleteIgnoredKeys(ignoredKeys, obj as Record<string, unknown>));
+  const filteredB = objB?.map((obj) => deleteIgnoredKeys(ignoredKeys, obj as Record<string, unknown>));
```
pxrl, line 130: ooc, do you think there are ways we can do better than this in future? (i.e. if we chose to tighten this code)

As I see it, this change is really only appeasing ts by shuffling the "object fuzziness" to a different place. I think that's fine in the context of it not making the problem any worse than it already is - I'm just interested to know if you think it can be done better.

src/utils/ProviderUtils.ts
james-a-morris: https://github.com/across-protocol/relayer/pull/1734/commits/3f4812434e8646ecf084c97a2682ea4b4420f5a4 just preference of the first pass - good call

src/utils/ObjectUtils.ts
james-a-morris: We can probably make this more TS-esque with type generics. I opted to not do this as it felt out of the scope of cleaning the code to a place where we can migrate to the SDK.

But I agree - this is a bandaid

pxrl: > I opted to not do this as it felt out of the scope of cleaning the code to a place where we can migrate to the SDK.

100% agree - I think it's totally fair to apply this bandaid here and was just asking out of curiosity.

pxrl (approved):

PR #1736, opened 2024-08-07
improve(adapters): Migrate Arbitrum One to generic adapter format

bmzig:
This is the last of the adapters I will propose for the time being. Migrating Arbitrum is going to be particularly important since we plan on supporting more Arbitrum L2s in the near future.


bmzig:
Tested with WETH, USDC, USDC.e, WBTC, and DAI (these tokens cover every Arbitrum gateway for the tokens we support) and can confirm that the new adapter can properly track outstanding transfers.

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -83,11 +86,13 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
   ): Promise<BridgeEvents> {
     const events = await paginatedEventQuery(
       this.getL2Bridge(),
-      this.getL2Bridge().filters.DepositFinalized(l1Token, fromAddress, undefined),
+      this.getL2Bridge().filters.DepositFinalized(l1Token, fromAddress, toAddress),
```
james-a-morris, line 89: ooc why are we pointing to a specific address now?

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -67,11 +68,13 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
...
       this.getL1Bridge(),
-      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress),
+      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress, toAddress),
       eventConfig
     );
     return {
-      [this.resolveL2TokenAddress(l1Token)]: events.map((event) => processEvent(event, "_amount", "_to", "_from")),
+      [this.resolveL2TokenAddress(l1Token)]: events
+        .map((event) => processEvent(event, "_amount", "_to", "_from"))
+        .filter(({ amount }) => amount.gt(bnZero)),
```
james-a-morris, line 77: Is this spurred by errors?

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -67,11 +68,13 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
   ): Promise<BridgeEvents> {
     const events = await paginatedEventQuery(
       this.getL1Bridge(),
-      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress),
+      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress, toAddress),
```
james-a-morris, line 71: similar to my first comment

james-a-morris (commented):
few questions

src/adapter/bridges/ArbitrumOneBridge.ts
bmzig: This is still something in discussion, but for the most part, the only time `fromAddress != toAddress` in practice is when we are on a chain which aliases contract addresses (in which case the bridge code should have logic which deals with this). Otherwise, the L1 address should be the same as the L2 address. This is formerly undefined since in our tests, we emit events where the recipient is a random address and the sender is a monitored address (e.g. [right here](https://github.com/across-protocol/relayer/blob/8f0b3ea43103ea062d3b2a63dfa5a08f3451b792/test/cross-chain-adapters/Arbitrum.ts#L79), where we enforce that we track deposits which originate from the monitored address and end in a random address), but this test has no real utility in practice, and for the sake of verbosity, I believe we are moving towards specifying precisely the route a transfer has to take in order to be tracked.

src/adapter/bridges/ArbitrumOneBridge.ts
bmzig: Yeah. The query functions return all transfers, even if they are empty, so the `.filter` is to remove them. Once again, this is not really seen in practice (I think?).

src/adapter/bridges/ArbitrumOneBridge.ts
bmzig: I realized this won't work in tracking Hub -> spoke transfers, so I changed it back to only tracking the `toAddress` https://github.com/across-protocol/relayer/pull/1736/commits/2d2fb4c8f7edc761ac638543a5c2480053b01d58

pxrl: @bmzig Has this been re-tested on chain after the change?

src/adapter/bridges/ArbitrumOneBridge.ts
pxrl: Is this filtering spam transactions directed to the relayer's address?

bmzig: Yeah. Realistically this is just burning ETH, so we wouldn't see it in practice. It was in before due to unit tests. Removed here: https://github.com/across-protocol/relayer/pull/1736/commits/f3f65ecd1b14796058e546235d59198cde9ecea3

src/adapter/bridges/ArbitrumOneBridge.ts
bmzig: It has been. I tested all tokens except for POOL and BAL, which should be OK, since they share the same gateway addresses as other ERCs like WBTC, ACX, and UMA, which were tested. Tracking does work, even for USDC -> USDC/USDC.e. I've also tested the monitor with hub -> spoke transfers and those seem alright too.

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -67,11 +67,14 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
...
       this.getL1Bridge(),
-      this.getL1Bridge().filters.DepositInitiated(undefined, fromAddress),
+      this.getL1Bridge().filters.DepositInitiated(undefined, undefined, toAddress),
       eventConfig
     );
+    const filteredEvents = events.filter(({ args }) => args.l1Token === l1Token);
     return {
-      [this.resolveL2TokenAddress(l1Token)]: events.map((event) => processEvent(event, "_amount", "_to", "_from")),
+      [this.resolveL2TokenAddress(l1Token)]: filteredEvents.map((event) =>
+        processEvent(event, "_amount", "_to", "_from")
```
pxrl, line 76: Minor style nit:

```suggestion
      [this.resolveL2TokenAddress(l1Token)]: events
        .filter(({ args }) => args.l1Token === l1Token)
        .map((event) => processEvent(event, "_amount", "_to", "_from")
```

bmzig: https://github.com/across-protocol/relayer/pull/1736/commits/17b7cdd6761bd969d87f074bf67780faca382f98

PR #1737, opened 2024-08-08
chore: refactor provider utils to sdk

james-a-morris:
A follow up to https://github.com/across-protocol/sdk/pull/693


nicholaspai (approved):

src/utils/ProviderUtils.ts
```diff
@@ -1,584 +1,18 @@
...
-
-    // All other calls should use quorum 1 to avoid errors due to sync differences.
-    return 1;
-  }
-}
+export class RetryProvider extends sdkProviders.RetryProvider {}
 
 // Global provider cache to avoid creating multiple providers for the same chain.
-const providerCache: { [chainId: number]: RetryProvider } = {};
+const providerCache: { [chainId: number]: sdkProviders.RetryProvider } = {};
```
pxrl, line 13: ooc, is it not sufficient to stick with `RetryProvider` given that it's defined directly above?

src/utils/ProviderUtils.ts
```diff
@@ -628,9 +66,10 @@ export async function getProvider(chainId: number, logger?: winston.Logger, useC
     NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS,
     NODE_DISABLE_INFINITE_TTL_PROVIDER_CACHING,
     NODE_PCT_RPC_CALLS_LOGGED,
+    PROVIDER_CACHE_TTL,
   } = process.env;
 
-  const timeout = Number(process.env[`NODE_TIMEOUT_${chainId}`] || NODE_TIMEOUT || defaultTimeout);
+  const timeout = Number(process.env[`NODE_TIMEOUT_${chainId}`] || NODE_TIMEOUT || sdkProviders.defaultTimeout);
```
pxrl, line 72: I'm in two minds about this, but it might make sense to retain a local `defaultTimeout` definition to avoid having to bump the SDK to change the default (and have unknown impact on some other user).

pxrl (commented):
Some initial comments - I'll give this a soak test now.

src/utils/ProviderUtils.ts
james-a-morris: After second thought: `defaultTimeout` only exists in this one place - I will put it back in the relayer repo https://github.com/across-protocol/relayer/pull/1737/commits/9bd8e50e343832a23c32eac092b58587c4d034e9

james-a-morris:
approval from @pxrl offline

PR #1738, opened 2024-08-08
improve(adapters): update adapters with new atomic depositor

bmzig:
Pending further review/deployment of https://github.com/across-protocol/relayer/pull/1696


bmzig:
Likely won't be able to get the last test working until the new atomic depositor gets deployed.

PR #1740, opened 2024-08-09
fix: Try to lessen Linea finalizer RPC burden

pxrl:
The Linea SDK implements some unfortunate RPC queries that are not well-suited to being called in parallel. This seems to be resulting in errors being returned by upstream RPC providers. Serialise some queries, with a longer delay on retry, in order to try to lessen the load and reduce the error rate. This will inevitably lead to a longer execution time, but might help to save the bot from throwing.

### DIFF
--- a/src/finalizer/utils/linea/common.ts
+++ b/src/finalizer/utils/linea/common.ts
@@ -84,11 +84,15 @@ export function makeGetMessagesWithStatusByTxHash(
         };
       });
 
-    // The Linea SDK MessageServiceContract constructs its own Provider without our retry logic so we retry each call
-    // twice with a 1 second delay between in case of intermittent RPC failures.
-    const messageStatus = await Promise.all(
-      messages.map((message) => retryAsync(() => dstClaimingService.getMessageStatus(message.messageHash), 2, 1))
-    );
+    // The Linea SDK MessageServiceContract constructs its own Provider without our caching, rate-limiting or retry
+    // logic. In particular, it scrapes logs over 0 -> latest and can heavily load the RPC provider. Query message
+    // up to twice, with a 1 second delay between, in case of intermittent RPC failures.
+    const messageStatus: OnChainMessageStatus[] = [];
+    for (const message of messages) {
+      const status = await retryAsync(() => dstClaimingService.getMessageStatus(message.messageHash), 2, 1);
+      messageStatus.push(status);
+    }
+
     return messages.map((message, index) => ({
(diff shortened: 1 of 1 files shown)


bmzig (approved):

nicholaspai (approved):

james-a-morris:
Side thought: we can just override the rpc provider that is set in the constructor. Here's what the constructor SDK looks like in JS:

```
    constructor(options) {
        this.network = options.network;
        this.mode = options.mode;
        this.l2MessageTreeDepth = options.l2MessageTreeDepth ?? constants_1.DEFAULT_L2_MESSAGE_TREE_DEPTH;
        if (options.mode === "read-write") {
            this.l1SignerPrivateKey = options.l1SignerPrivateKey;
            this.l2SignerPrivateKey = options.l2SignerPrivateKey;
            this.maxFeePerGas = options.feeEstimatorOptions?.maxFeePerGas;
            this.gasFeeEstimationPercentile = options.feeEstimatorOptions?.gasFeeEstimationPercentile;
        }
        this.l1Provider = new ProviderService_1.default(options.l1RpcUrl);
        this.l2Provider = new ProviderService_1.default(options.l2RpcUrl);
    }

```
and `ProviderService` is:
```
import { ethers } from "ethers";
import { JsonRpcProvider } from "@ethersproject/providers";
export default class ProviderService {
    provider: JsonRpcProvider;
    constructor(rpcUrl: string);
    getSigner(privateKey: string): ethers.Signer;
}
```
with it's JS being:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const ethers_1 = require("ethers");
const providers_1 = require("@ethersproject/providers");
class ProviderService {
    constructor(rpcUrl) {
        this.provider = new providers_1.JsonRpcProvider(rpcUrl);
    }
    getSigner(privateKey) {
        try {
            return new ethers_1.Wallet(privateKey, this.provider);
        }
        catch (e) {
            throw new Error("Something went wrong when trying to generate Wallet. Please check your private key and the provider url.");
        }
    }
}
exports.default = ProviderService;
```

We can probably create a class identical to `ProviderService`, create two instances of it and override the state set by the constructor with:
```
    (sdk as unknown as Record<string, unknown>)["l1Provider"] = new ProviderService( <- with our RPC provider)`
```

james-a-morris (commented):
https://github.com/across-protocol/relayer/pull/1740#issuecomment-2289811014

PR #1744, opened 2024-08-12
improve(Monitor): Unwrap WETH on Ethereum when attempting to refill ETH for configured accounts

nicholaspai:
Motivation:
- Seeing lots of warning when trying to refill ETH from one account to another because the sender has lots of WETH but not enough ETH.

Features:
- Add feature in Monitor to automatically unwrap WETH into ETH when trying to refill ETH on Ethereum
- Refactor/rename functions to better support a future where we have different native tokens per chain.

Details
- Found a bug where we incorrectly assume that the `eth` address on Polygon is 0x0, but the MATIC address seems to be: [0x0000000000000000000000000000000000001010](https://polygonscan.com/token/0x0000000000000000000000000000000000001010)
- This has never really been an issue up until now since we make the correct balance calls (provider.getBalance) anyways even though we make faulty assumptions that the polygon eth token is 0x0
- This could have been an issue when unwrapping WMATIC into MATIC on Polygon but we don't do this anywhere in the code

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -480,9 +480,11 @@ export class Monitor {
             deficit
           );
           // If token is gas token, try unwrapping deficit amount of WETH into ETH to have available for refill.
-          // We use this only on Mainnet for now since we're most likely to refill balances on mainnet and we
-          // want to test this in production before rolling it out to all chains.
-          if (!canRefill && chainId === CHAIN_IDs.MAINNET && token === getNativeTokenAddressForChain(chainId)) {
+          if (
+            !canRefill &&
+            token === getNativeTokenAddressForChain(chainId) &&
+            getNativeTokenSymbol(chainId) === "ETH"
+          ) {
             const weth = new Contract(
               TOKEN_SYMBOLS_MAP.WETH.addresses[chainId],
               WETH9.abi,
@@ -497,6 +499,7 @@ export class Monitor {
                 chainId,
                 requiredUnwrapAmount: deficit.toString(),
                 wethBalance,
+                wethAddress: weth.address,
                 ethBalance: currentBalance.toString(),
                 transactionHash: blockExplorerLink(txn.transactionHash, chainId),
               });
@@ -504,14 +507,12 @@ export class Monitor {
             } else {
               this.logger.warn({
                 at: "Monitor#refillBalances",
-                message: "Trying to unwrap WETH balance to use for refilling ETH but not enough WETH to unwrap",
-                from: signerAddress,
-                to: account,
-                balanceTrigger,
-                balanceTarget,
-                requiredUnwrapAmount: deficit.toString(),
-                token,
+                message: `Trying to unwrap WETH balance from ${signerAddress} to use for refilling ETH in ${account} but not enough WETH to unwrap`,
                 chainId,
+                requiredUnwrapAmount: deficit.toString(),
+                wethBalance,
+                wethAddress: weth.address,
+                ethBalance: currentBalance.toString(),
               });
               return;
             }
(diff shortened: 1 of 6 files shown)


src/monitor/MonitorConfig.ts
```diff
@@ -154,9 +154,9 @@ export class MonitorConfig extends CommonConfig {
             parsedWarnThreshold = Number(warnThreshold);
           }
 
-          const isNativeToken = !token || token === "0x0" || token === getEthAddressForChain(chainId);
+          const isNativeToken = !token || token === getNativeTokenAddressForChain(chainId);
```
nicholaspai, line 157: This is one subtle change in this PR. Currently we allow the configuration to have `token=0x0` for any given MONITORED_BALANCE and let that be an instruction to this file to override that token with the chain's native token. I think this is a mistake and we don't use it in production currently, since many chain native token addresses are not 0x0.

src/monitor/Monitor.ts
```diff
@@ -471,12 +473,48 @@ export class Monitor {
...
-          const canRefill = await this.balanceAllocator.requestBalanceAllocation(
+          let canRefill = await this.balanceAllocator.requestBalanceAllocation(
             chainId,
             [token],
             signerAddress,
             deficit
           );
+          // If token is gas token, try unwrapping deficit amount of WETH into ETH to have available for refill.
+          // We use this only on Mainnet for now since we're most likely to refill balances on mainnet and we
+          // want to test this in production before rolling it out to all chains.
```
pxrl, line 484: This is fwiw also occurring on new chains where the executor has only a tiny initial balance and the relayer is responsible for getting it up to target.

src/monitor/Monitor.ts
```diff
@@ -471,12 +473,48 @@ export class Monitor {
...
+          if (!canRefill && chainId === CHAIN_IDs.MAINNET && token === getNativeTokenAddressForChain(chainId)) {
+            const weth = new Contract(
+              TOKEN_SYMBOLS_MAP["WETH"].addresses[chainId],
```
pxrl, line 487:
```suggestion
              TOKEN_SYMBOLS_MAP.WETH.addresses[chainId],
```

src/monitor/Monitor.ts
```diff
@@ -471,12 +473,48 @@ export class Monitor {
...
+                at: "Monitor#refillBalances",
+                message: "Trying to unwrap WETH balance to use for refilling ETH but not enough WETH to unwrap",
+                from: signerAddress,
+                to: account,
+                balanceTrigger,
+                balanceTarget,
+                requiredUnwrapAmount: deficit.toString(),
+                token,
+                chainId,
+              });
```
pxrl, line 515: We could consider returning here, because otherwise we'll just print another warning message shortly after:

```
            this.logger.warn({
              at: "Monitor#refillBalances",
              message: "Cannot refill balance to target",
              from: signerAddress,
              to: account,
              balanceTrigger,
              balanceTarget,
              deficit,
              token,
              chainId,
            });
```

src/monitor/MonitorConfig.ts
pxrl: Do you think we ever had a situation where token was `0x0` ? This is a string comparison, but `0x0` is not the same as a proper 20 byte address, so I feel like this comparison has probably just been superfluous and could never evaluate true.

pxrl (approved):
LGTM - only some minor comments.

src/monitor/MonitorConfig.ts
nicholaspai: true yeah i don't think we ever did and this feels like unused code that also is wrong

src/monitor/Monitor.ts
nicholaspai: ah ok well I can remove the mainnet only restriction?

PR #1745, opened 2024-08-13
feat(Dataworker): Move persistence of bundle data step to Disputer instead of Proposer

nicholaspai:
This PR adds a feature to the disputer which allows it repersist data to arweave in the event the pending bundle is valid, since it also reconstructs valid bundle data from scratch.

This way the disputer can have multiple chances throughout the bundle's challenge period to repersist to arweave.

A future PR would add this ability to the executor

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -591,8 +591,8 @@ export async function persistDataToArweave(
   // If so, we don't need to persist it again.
   const matchingTxns = await client.getByTopic(tag, any());
   if (matchingTxns.length > 0) {
-    logger.info({
-      at: "Dataworker#index",
+    logger.debug({
+      at: "DataworkerUtils#persistDataToArweave",
       message: `Data already exists on Arweave with tag: ${tag}`,
       hash: matchingTxns.map((txn) => txn.hash),
     });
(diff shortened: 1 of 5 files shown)


james-a-morris (commented):
lgtm - this was clever to put in the disputer

james-a-morris (approved):
lgtm - My only outstanding comment would be potentially increasing the gap between proposal & arweave persistence may slow down the unlucky bot run that needs to load the bundle data client from scratch

src/dataworker/index.ts
```diff
@@ -164,62 +162,25 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
...
-          await persistDataToArweave(
-            clients.arweaveClient,
-            bundleDataToPersist,
-            logger,
-            `bundles-${bundleDataToPersist.bundleBlockRanges}`
-          );
-        }
-      };
-
-      const executeDataworkerTransactions = async () => {
```
nicholaspai, line 185: I removed this wrapper function `executeDataworkerTransactions` since we no longer need to call it in parallel with `persistDataToDaLayer`

src/dataworker/DataworkerUtils.ts
```diff
@@ -603,17 +603,17 @@ export async function persistDataToArweave(
...
     logger.info({
-      at: "Dataworker#index",
+      at: "DataworkerUtils#persistDataToArweave",
       message: "Persisted data to Arweave! 💾",
       receipt: `https://arweave.app/tx/${hashTxn}`,
       rawData: `https://arweave.net/${hashTxn}`,
       address,
       balance,
     });
+    const endTime = performance.now();
```
nicholaspai, line 613: nit bug: should only log time if we persist data

src/dataworker/DataworkerUtils.ts
```diff
@@ -591,8 +591,8 @@ export async function persistDataToArweave(
   // If so, we don't need to persist it again.
   const matchingTxns = await client.getByTopic(tag, any());
   if (matchingTxns.length > 0) {
-    logger.info({
-      at: "Dataworker#index",
+    logger.debug({
```
nicholaspai, line 594: should only info log when sending txn

PR #1746, opened 2024-08-14
improve(relayer): only send token approvals when allowance is half of max

bmzig:
Currently, we cannot run the relayer with the void signer, since we store `MAX_SAFE_ALLOWANCE` in cache with no TTL. This means that when we set a new atomic depositor address (or really any new bridge address we need to approve), the main relayer approves the atomic depositor address to spend `MAX_SAFE_ALLOWANCE` WETH and caches that the approvals for the atomic depositor is `MAX_SAFE_ALLOWANCE` indefinitely. Now, once the main relayer bridges weth to some chain. This decreases the atomic depositor's allowance for the main relayer to some number less than `MAX_SAFE_ALLOWANCE`, say. `MAX_SAFE_ALLOWANCE-x`. If I were to run `yarn relay --wallet void`, since I have no cached allowance for the atomic depositor, the bot fetches the main relayer's current allowance for the atomic depositor, which is `MAX_SAFE_ALLOWANCE-x`. Since this is less than `MAX_SAFE_ALLOWANCE`, the relayer tries to send another approval and errors out. This change won't affect the main relayers since after the first approval, they always fetch the allowance from cache. 

Choosing to set the threshold as half of `MAX_SAFE_ALLOWANCE` was arbitrary, so if we want to change it, we can.


src/clients/bridges/BaseAdapter.ts
```diff
@@ -157,7 +157,7 @@ export class BaseAdapter extends BaseChainAdapter {
       .map((allowance, idx) => {
         const token = l1TokenContracts[idx];
         const bridge = l1Bridges[idx];
-        if (token && bridge && allowance.lt(MAX_SAFE_ALLOWANCE)) {
+        if (token && bridge && allowance.lt(toBN(MAX_SAFE_ALLOWANCE).div(2))) {
```
james-a-morris, line 160: Is this `!aboveAllowanceThreshold(allowance)`?

bmzig: It is. https://github.com/across-protocol/relayer/pull/1746/commits/92a3ec7a50d43c936ac1bd393f063e7bd749bc76

james-a-morris (approved):

pxrl (approved):

pxrl:
@bmzig Nice find on this - happy to see this go in!

PR #1751, opened 2024-08-15
fix(constants): Update Multicall3 addresses

pxrl:
Lisk has a Multicall3 deployment, but Redstone does not (...yet).


james-a-morris (approved):
Is this the first chain without a multicaller address?

pxrl:
> Is this the first chain without a multicaller address?

Lisk previously had no deployment, but it did get multicall3 about 2 weeks ago.

I've dug into it a bit and they have a form to request deployments for new chains. We've worked that into the onboarding form such that any new chain is encouraged to request deployment if they don't already have one.

PR #1755, opened 2024-08-16
chore: fix some comments

mountdisk:


src/clients/TokenClient.ts
```diff
@@ -68,7 +68,7 @@ export class TokenClient {
     return this.getBalance(deposit.destinationChainId, deposit.outputToken).gte(deposit.outputAmount);
   }
 
-  // If the relayer tries to execute a relay but does not have enough tokens to fully fill it it will capture the
+  // If the relayer tries to execute a relay but does not have enough tokens to fully fill it will capture the
```
james-a-morris, line 71: The original looks correct here. The first `it` is a pronoun reference to the relay. It's a bit vague so maybe we can change that?

src/finalizer/utils/opStack.ts
```diff
@@ -288,7 +288,7 @@ async function finalizeOptimismMessage(
     if (requestId === undefined || requestId === 0) {
       throw new Error(`Found invalid requestId ${requestId} for Blast withdrawal ${message.event.transactionHash}`);
     }
-    // @dev The hintId parameter plays a role in our insurance mechanism that kicks in in the rare event that
+    // @dev The hintId parameter plays a role in our insurance mechanism that kicks in the rare event that
```
james-a-morris, line 291: The original was correct here. The first `in` is part of the phrase `kicks in`. I'd recommend maybe we use `kicks-in` instead but I don't think we should remove the second `in`.

james-a-morris (changes requested):
two comments.

pxrl:
@mountdisk  I don't think these are comments that need to be changed. Happy to consider other PRs if you find anything else.

PR #1756, opened 2024-08-16
improve(adapters): swap polygon adapter with generic adapter

bmzig:
Pending unit tests and confirmation that cross-chain transfer tracking works fully.


src/adapter/bridges/PolygonERC20Bridge.ts
```diff
@@ -65,12 +65,12 @@ export class PolygonERC20Bridge extends BaseBridgeAdapter {
...
     const events = await paginatedEventQuery(
       this.getL1Bridge(),
-      this.getL1Bridge().filters.LockedERC20(undefined, fromAddress, l1Token),
+      this.getL1Bridge().filters.LockedERC20(undefined, toAddress, l1Token),
       eventConfig
     );
     return {
       [this.resolveL2TokenAddress(l1Token)]: events.map((event) =>
-        processEvent(event, "amount", "depositorReceiver", "depositor")
+        processEvent(event, "amount", "depositReceiver", "depositor")
```
james-a-morris, line 73: ooc why are we updating this event name?

bmzig: The event name was incorrect. https://github.com/across-protocol/relayer/blob/9c4190d2d07a2966475bb7798638bc4f6ce0e392/src/common/abi/PolygonBridge.json#L6

pxrl: I see it is `depositReceiver` in the existing adapter too :+1:

pxrl (commented):
@bmzig This looks good. Did you confirm it works OK with rebalancing?

bmzig:
> @bmzig This looks good. Did you confirm it works OK with rebalancing?

There was a small issue with resolving the wrong USDC address but that was fixed here https://github.com/across-protocol/relayer/pull/1756/commits/d0f88d794bb9060ae6a2778aadbce8e9969df809. Should be good to merge whenever now.

pxrl (approved):

PR #1758, opened 2024-08-16
feat(ProviderUtils): Support generic RPC provider keys

pxrl:
Permit Alchemy and Infura RPC keys to be defined via the env vars:
  RPC_PROVIDER_KEY_ALCHEMY
  RPC_PROVIDER_KEY_INFURA

If an RPC endpoint is defined via `RPC_PROVIDER_<PROVIDER>_<CHAIN-ID>` then it will still be used, but will otherwise fall back to inferring the RPC URL if the provider is Alchemy or Infura _and_ an API key is defined via the RPC_PROVIDER_KEY_<PROVIDER> env var.


nicholaspai (approved):

james-a-morris (approved):
I like this a lot - it should cut down the configs too

PR #1761, opened 2024-08-19
fix(TryMulticallClient): Use highest gasLimit when sending multicall txns

bmzig:
There was an issue in the `TryMulticallClient` where we would not use enough gas to fund a multicall transaction ([example](https://basescan.org/tx/0x07421f7412c34564403fde59841bf1e262cb75bc5b4992da39dbe4a743bef5a8)), causing the multicall to revert. This PR provides a potential fix, where we set the gas limit of the rebuilt transaction to the highest gas limit we obtained in simulation. We do this since we only know the gasLimit for the `tryMulticall` simulation and not the gas expended for any single `fillV3Relay`. Although as far as I know, setting a high gas limit wouldn't really matter in the context of our relayers, it is worth pointing out the shortcomings of this method:

First, when we simulate `tryMulticall`, some individual `fillV3Relay` functions may fail. While we know which transactions fail, we don't know how much gas they contributed to the total gasLimit. The good news is that since they reverted, they contributed relatively less to the predicted gas limit than a successful `fillV3Relay` transaction. In the worst case, we build a multicall bundle of size `n`, and `n-1` transactions fail, causing us to set a gas limit as though we were executing `n-1` failed `fillV3Relay`s and a single sucessful one. 

Also, in practice, we only build a single multicall bundle, so this will work as intended, but another theoretical shortcoming is that if we do have multiple multicall bundles (which implies that there is at least one bundle with the maximum amount of transactions in it), there is a chance that we rebuild successful transactions from both bundles into a single bundle, but the transactions in the second bundle were fewer but more expensive gas wise. This would also cause the transaction to revert, but this is a pretty extreme edge case. 

I have tested this new code with multicalls, and they have succeeded (gas limit is more akin to a multicall transaction):
https://basescan.org/tx/0xd9854c336380486717793dce1c5a0f04ab064f70eab6767601147519c6290e98
https://basescan.org/tx/0xf2f019a30af766b1bf6b87a1347a006a11672871e757342546cfe3579717593b


nicholaspai:
What makes me uneasy about the approach used for this pr in general is that it seems we are guessing and checking in prod. Instead can we use hardhat unit tests to get more confidence about whether this will work?

bmzig:
> What makes me uneasy about the approach used for this pr in general is that it seems we are guessing and checking in prod. Instead can we use hardhat unit tests to get more confidence about whether this will work?

I'm fine with adding tests, but in this case, I don't think there is much of an ability to test how accurate gas estimation will be right? The only way to test this is with a relayer, as shown above. In the `MultiCallerClient`, we have the luxury of just resimulating all transactions if the bundle fails, and doing this gives us a better idea of the gas limit to for each individual `fillV3Relay`. Here, we save a lot of time since we just rebuild a tryMulticall based off of whether or not the internal transaction succeeded, but we don't get the extra information about each individual transaction's gas expenditure.

src/clients/MultiCallerClient.ts
```diff
@@ -603,19 +609,28 @@ export class TryMulticallClient extends MultiCallerClient {
...
+        const uniqueGasLimits = [gasLimit];
         txns.forEach((txn, idx) => {
           mrkdwn.push(`\n *txn. ${idx + 1}:* ${txn.data ?? "No calldata"}`);
           assert(contract === txn.contract);
+          // Sum up all the unique gasLimits. The only time gasLimit != txns[0].gasLimit is when we are consolidating
+          // multiple multicall bundles into a single one.
+          if (!uniqueGasLimits.includes(txn.gasLimit)) {
+            uniqueGasLimits.push(txn.gasLimit);
+            gasLimit = gasLimit.add(txn.gasLimit);
+          }
```
pxrl, line 622: I don't understand the intention behind looking for unique gasLimits. Also, in the event that uniqueness is useful and reliable, I think `Array.includes()` will be checking for equality of the BigNumber objects, but not uniqueness of the actual underlying numeric representation. So I believe you could have two different BigNumber objects containing the same value, and they would both be considered unique.

test/mocks/MockTransactionClient.ts
```diff
@@ -69,4 +78,19 @@ export class MockedTransactionClient extends TransactionClient {
...
+  checkIndividualTransactions(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
+    // We know it's tryMulticall so there is only one argument.
+    const returnData = txn.args[0].map(([result]) => {
+      return {
+        success: result === txnClientPassResult,
+        data: result,
+      };
+    });
+    return {
+      transaction: { ...txn },
```
pxrl, line 91: Does this need to be a new object?

```suggestion
      transaction: txn,
```

bmzig: True. https://github.com/across-protocol/relayer/pull/1761/commits/2a41182e4aac8ecda63b3d399f088efb5826a7c2

src/clients/MultiCallerClient.ts
bmzig: You're right. Since we likely won't ever see real benefits from combining two separate multicall bundles into a single one, and since this adds a decent amount of complexity to the code, I instead am just going to keep the bundles in a vacuum so we can use the simulated gasLimit for each bundle. https://github.com/across-protocol/relayer/pull/1761/commits/2a41182e4aac8ecda63b3d399f088efb5826a7c2

bmzig:
This was tested over some transactions on Optimism with a multicall chunk size of 4 The transactions which had less than 4 multicalls corresponded to calls which had invalid data in them and needed to be rebuilt. All of them had a gasLimit which matched with the simulated value. ([1](https://optimistic.etherscan.io/tx/0xc20e7dbe0335ce9dde8309376f1376b59ca8c80d8692ec068a6bacc40852f530), [2](https://optimistic.etherscan.io/tx/0x86660cf779031f11af39d33b00651d39f05119b29dfe3f74d9dd49d24bee8975), [3](https://optimistic.etherscan.io/tx/0x3b04210085b7ac53cfc0886a6acbda4f888fe06851ae17d054b54ca62e22b980), [4](https://optimistic.etherscan.io/tx/0x21df007e634670bb24ca4df12f05d062321b2395945b20a052bfaf9d9670207b), [5](https://optimistic.etherscan.io/tx/0xe2eaa9e698a7dfc7b7e6b3ca407acef7b47acb8a117d1eccb79dff231c7956d0)).

src/clients/MultiCallerClient.ts
pxrl: Yeah; I think that's probably a reasonable approach - especially now that `tryMulticall` is available so we'll be able to have larger batches without fear of reverting :fire:

src/clients/MultiCallerClient.ts
```diff
@@ -572,7 +576,9 @@ export class TryMulticallClient extends MultiCallerClient {
       // some txns in the bundle must have failed. We take note only of the ones which succeeded.
       if (succeededTxnCalldata.length !== data.length) {
         txnCalldataToRebuild.push(
-          ...succeededTxnCalldata.map((calldata) => buildRawTransaction(transaction.contract, calldata))
+          succeededTxnCalldata.map((calldata) =>
```
nicholaspai, line 579: Why can't you just set `txnCalldataToRebuild =  succeededTxnCalldata.map(...)` instead of `push`-ing?

src/clients/TransactionClient.ts
```diff
@@ -33,6 +33,7 @@ export interface AugmentedTransaction {
 export interface RawTransaction {
   contract: Contract;
   data: string;
+  gasLimit: BigNumber;
```
nicholaspai, line 36: Doesn't this break other code by changing an interface that other functions presumably use?

test/mocks/MockTransactionClient.ts
```diff
@@ -54,10 +57,16 @@ export class MockedTransactionClient extends TransactionClient {
     }
 
     const _nonce = nonce ?? 1;
+    // For testing the TryMulticallClient, we need to know how many transactions there were in the multicall bundle, so
```
nicholaspai, line 60: what do you mean by "for testing" is this logic only added to get unit tests working? I don't love this approach of adding edge case logic to production code

test/mocks/MockTransactionClient.ts
```diff
@@ -28,6 +28,9 @@ export class MockedTransactionClient extends TransactionClient {
   }
 
   protected override async _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
+    if (txn.method === "tryMulticall") {
+      return this.checkIndividualTransactions(txn);
```
nicholaspai, line 32: Why is tryMulticall a special case here? I don't get it, this at least warrants a comment

test/TryMulticallClient.ts
```diff
@@ -0,0 +1,386 @@
...
+            interface: { encodeFunctionData },
+          } as Contract,
+          args: [result],
+          message: `Test transaction on chain ${chainId}`,
+          mrkdwn: `This transaction is expected to ${fail ? "fail" : "pass"} simulation.`,
+        } as AugmentedTransaction;
+        multiCaller.enqueueTransaction(txn);
+      }
+      const expectedTransactionHash = ethers.utils.id(
+        `Across-v2-${spokePool.address}-${"tryMulticall"}-${nTxns - nFailed}`
```
nicholaspai, line 187: I think a better way to check # of txns is to read the # of events that are emitted in the `executeTxnQueue` function. This can easily be done by using a very simple contract that either you write new or something like [this](https://github.com/across-protocol/contracts/blob/95c4f923932d597d3e63449718bba5c674b084eb/contracts/test/MockSpokePoolV2.sol#L24) which emits events on each successful call. It could also increment a global state counter which you can check post-execution.

test/mocks/MockTransactionClient.ts
nicholaspai: https://github.com/across-protocol/relayer/pull/1761/files#r1729214316

test/TryMulticallClient.ts
```diff
@@ -0,0 +1,386 @@
...
+
+      // Verify that the failed simulations were filtered out.
+      expect(multiCaller.simulationFailureCount()).to.equal(fail ? txns.length : 0);
+      multiCaller.clearSimulationFailures();
+    }
+  });
+
+  it("Correctly rebuilds a tryMulticall from successful transaction data", async function () {
+    // Give the contract multicall so that checks in the MultiCallerClient will pass.
+    const spokePool = (await deploySpokePoolWithToken()).spokePool;
```
nicholaspai, line 163: I don't recommend using spoke pool as the contract to test here, its a very complex contract. Just use a much simpler contract

test/TryMulticallClient.ts
nicholaspai: or just use a smock contract and check number of times a function was called

src/clients/MultiCallerClient.ts
```diff
@@ -599,10 +604,10 @@ export class TryMulticallClient extends MultiCallerClient {
       // chunk size. Every transaction should be aimed at the same spoke pool since 1. The tryMulticall client is only
       // instantiated for the relayer, which uses this client only for interfacing with the spoke pool, and 2. This function
       // is called after filtering transactions by chainId, so each individual transaction is a call to a chainId's spoke pool.
-      const txnChunks = lodash.chunk(txnCalldataToRebuild, this.chunkSize[chainId] ?? DEFAULT_MULTICALL_CHUNK_SIZE);
       const rebuildTryMulticall = (txns: RawTransaction[]) => {
         const mrkdwn: string[] = [];
         const contract = txns[0].contract;
+        const gasLimit = txns[0].gasLimit;
```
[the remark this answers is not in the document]
bmzig: Txns is an object containing a contract, a bytes field (the calldata to call) and a gas limit. We grab index zero here since the contract and gas limit will be the same for all transactions in this array (since this array is just picking out the individual transactions in the multicall bundle which succeeded).

src/clients/MultiCallerClient.ts
bmzig: `txnCalldataToRebuild` is a 2d array. So every entry is an array full of the successfull calldata. We can't assign since if there is more than one bundle, then we would always just take the last bundle as calldata to rebuild.

src/clients/TransactionClient.ts
bmzig: This is a new interface introduced in the first tryMulticall commit, so no other functions will use it.

test/mocks/MockTransactionClient.ts
bmzig: AFAIK this hash is used by other tests to ensure that the correct transaction was called at the correct time. This does effectively the same thing except encompasses information needed to test. Either way, I'll just make a contract to emit events instead.

test/mocks/MockTransactionClient.ts
bmzig: The MockTransactionClient only simulates as a success if the argument is "pass". The arguments won't be "pass" for tryMulticall, they are instead an array of "pass"es, so that we can actually simulate the successes of individual txns. This is a special case of the MockTransactionClient, so I though this would be the best inclusion without making a new function or new mocked client.

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
nicholaspai: can we dereference txns[0] higher up? and add a comment? Generally any hardcoded number requires a comment in the code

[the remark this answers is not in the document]
nicholaspai: also why are there multiple transactions in here?

src/clients/MultiCallerClient.ts
nicholaspai: i don't understand why there is more than one bundle, how is that possible?

src/clients/TransactionClient.ts
nicholaspai: should this then be renamed TryMulticallRawTransaction..? How do you know other functions in future won't use it

test/mocks/MockTransactionClient.ts
nicholaspai: i think that's cleaner

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
bmzig: I think this is a good point. I'm going to just store the calldata and a single copy of the contract and gas limit so I'm not storing `n` copies of the same thing. There are multiple transactions though since each transaction is the "sub" transaction in the multicall bundle, so for example, if there were 5 transactions in a tryMulticall bundle, and 3 succeeded, there would be three RawTransactions in txns.

src/clients/TransactionClient.ts
bmzig: Good point. Will also make this comment https://github.com/across-protocol/relayer/pull/1761#discussion_r1729248670 clearer.

src/clients/MultiCallerClient.ts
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

src/clients/TransactionClient.ts
bmzig: https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

test/mocks/MockTransactionClient.ts
bmzig: https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

test/mocks/MockTransactionClient.ts
bmzig: Removed https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

test/TryMulticallClient.ts
bmzig: Done here https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

test/TryMulticallClient.ts
bmzig: Done here https://github.com/across-protocol/relayer/pull/1761/commits/cfa6c8c3a71130c0275b91c46dcfb91846ffe506

nicholaspai (approved):
I think this makes sense now, took me a bit to understand it deeply how its working.

PR #1766, opened 2024-08-20
refactor(relayer): Rearrange external listener loop

pxrl:
Rather than trying to recover a websocket error, just log it and try to keep going. Eventually it will bail; the parent can then be responsible for restarting it if necessary.

### DIFF
--- a/src/libexec/RelayerSpokePoolIndexer.ts
+++ b/src/libexec/RelayerSpokePoolIndexer.ts
@@ -288,44 +288,34 @@ async function run(argv: string[]): Promise<void> {
   // Events to listen for.
   const events = ["V3FundsDeposited", "RequestedSpeedUpV3Deposit", "FilledV3Relay"];
   const eventMgr = new EventManager(logger, chainId, quorum);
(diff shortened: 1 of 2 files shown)


src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -288,44 +288,34 @@ async function run(argv: string[]): Promise<void> {
...
-            at: "RelayerSpokePoolIndexer::run",
-            message: `${chain} provider connection closed.`,
-            provider: getOriginFromURL(provider.connection.url),
-          });
-        });
-      });
-
-      logger.debug({ at: "RelayerSpokePoolIndexer::run", message: `Starting ${chain} listener.`, events, opts });
-      await listen(eventMgr, spokePool, events, providers, opts);
-    } catch (err) {
```
pxrl, line 317: This loop may be contributing to high CPU load in the event of a provider error that's unrecoverable.

nicholaspai (approved):

PR #1770, opened 2024-08-20
feat(Dataworker): Save merkle leaf proofs to Arweave

nicholaspai:
This is a really simple PR because `persistDataToArweave` was implemented so generically.

The disputer saves merkle leaf data to Arweave at the same opportunity that the disputer saves the bundle data to Arweave. This will double our Arweave usage but make proofs available for all to use.

Example data posted: https://ba6piv2daybktn3lp7fnzmzpaa2p5mz5tukzowxa6ohlji7k4fxq.arweave.net/CDz0V0MGAqm3a3_K3LMvADT-sz2dFZda4POOtKPq4W8


pxrl (commented):
This looks good. We are however changing the format of the data submitted to Arweave; do we need to bump the `ARWEAVE_TAG_APP_VERSION` in SDK? cc @james-a-morris

nicholaspai:
> This looks good. We are however changing the format of the data submitted to Arweave; do we need to bump the `ARWEAVE_TAG_APP_VERSION` in SDK? cc @james-a-morris 

Are we? This pr is saving a new file to arweave but not touching the existing file we upload. So now we're uploading 2 files per bundle rather than 1

src/dataworker/Dataworker.ts
```diff
@@ -618,14 +618,46 @@ export class Dataworker {
...
+                ...leaf,
+                proof: expectedTrees.relayerRefundTree.tree.getHexProof(leaf),
+              };
+            }),
+            relayerRefundRoot: expectedTrees.relayerRefundTree.tree.getHexRoot(),
+            slowRelayLeaves: expectedTrees.slowRelayTree.leaves.map((leaf) => {
+              const { relayData, chainId, updatedOutputAmount } = leaf;
+              return {
+                ...leaf,
+                proof: expectedTrees.slowRelayTree.tree.getHexProof({ relayData, chainId, updatedOutputAmount }),
```
pxrl, line 652: Is there incidentally something additional in `leaf` that we don't want to persist to Arweave? Per the interface definition it should be possible to get away without this dereferencing.

```suggestion
              return {
                ...leaf,
                proof: expectedTrees.slowRelayTree.tree.getHexProof(leaf),
```

pxrl:
> Are we? This pr is saving a new file to arweave but not touching the existing file we upload. So now we're uploading 2 files per bundle rather than 1

Sorry, you're totally right. There's no issue with this.

pxrl (approved):

james-a-morris (approved):
lgtm - this is a append-only change so we won't need a structure change

PR #1771, opened 2024-08-20
improve(relayer): Support keepalives over websockets

pxrl:
As newer chains are added, we sometimes see that websockets are occasionally dropped; potentially as a result of low activity on the chain meaning that the websocket sits idle. This commit adds a simple periodic keepalive that will be initiated by each listener process towards each of its configured providers.


src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -291,9 +291,53 @@ async function run(argv: string[]): Promise<void> {
...
-    provider._websocket.on("error", (err) => {
-      const _provider = getOriginFromURL(provider.connection.url);
+    const { _websocket: ws } = provider;
+    const _provider = getOriginFromURL(provider.connection.url);
+    let interval: NodeJS.Timer | undefined;
+    let timeout: NodeJS.Timeout | undefined;
+
+    const cleanup = () => {
+      if (interval) {
+        clearInterval(interval);
```
nicholaspai, line 306: where is `clearInterval/clearTimeout` defined?

nicholaspai (approved):
Just one 1

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: These are built-ins that are packaged with node: https://nodejs.org/api/timers.html#clearintervaltimeout

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -199,7 +201,7 @@ async function listen(
   });
 
   do {
-    await setTimeout(INDEXER_POLLING_PERIOD);
+    await delay(INDEXER_POLLING_PERIOD);
```
james-a-morris, line 204: OOC can we re-use the SDK's delay function or is this separate?

james-a-morris (commented):
one question otherwise this lgtm.

Is there a good/easy way to test these changes?

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: We can use it :+1:

pxrl: https://github.com/across-protocol/relayer/pull/1771/commits/be3a943cce89d9f91c9f22a66f41c48f28cd7a53

PR #1773, opened 2024-08-21
fix(zkSyncWethBridge): enable hub pool tracking

bmzig:
The monitor thought there were stuck rebalance roots for zkSync since the `ZkSyncWethBridge` was unable to track HubPool transfers. This is because the code was based off of an older version [here](https://github.com/across-protocol/relayer/blob/c2f847c90b024d6970b0306654e68968479348ab/src/clients/bridges/ZKSyncAdapter.ts#L74) which has since been updated. This PR makes the changes to allow the monitor to use the hubPool as a monitored address.

### DIFF
--- a/src/adapter/bridges/ZKSyncWethBridge.ts
+++ b/src/adapter/bridges/ZKSyncWethBridge.ts
@@ -103,24 +103,22 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     // is the NewPriorityRequest event which doesn't have any parameters about the 'to' or 'amount' sent.
     // Therefore, we must track the HubPool and assume any transfers we are tracking from contracts are
     // being sent by the HubPool.
-    let events = await paginatedEventQuery(
-      isL2Contract ? hubPool : this.atomicDepositor,
-      isL2Contract
-        ? hubPool.filters.TokensRelayed()
-        : this.atomicDepositor.filters.ZkSyncEthDepositInitiated(fromAddress, toAddress),
-      eventConfig
-    );
+    let events, processedEvents;
     if (isL2Contract) {
-      // The TokensRelayed event does not distinguish between chain or l1Token, so we filter it here.
-      events = events.filter(
-        (e) => compareAddressesSimple(e.args.to, toAddress) && compareAddressesSimple(e.args.l1Token, wethAddress)
+      events = await paginatedEventQuery(hubPool, hubPool.filters.TokensRelayed(), eventConfig);
+      processedEvents = events
+        .filter(
+          (e) => compareAddressesSimple(e.args.to, toAddress) && compareAddressesSimple(e.args.l1Token, wethAddress)
+        )
+        .map((e) => processEvent(e, "amount", "to", "to"));
+      processedEvents.forEach((e) => (e.from = hubPool.address));
+    } else {
+      events = await paginatedEventQuery(
+        this.atomicDepositor,
+        this.atomicDepositor.filters.ZkSyncEthDepositInitiated(fromAddress, toAddress),
+        eventConfig
       );
-    }
-    const processedEvents = events.map((event) =>
-      isL2Contract ? processEvent(event, "amount", "to", "to") : processEvent(event, "_amount", "_to", "from")
-    );
-    if (isL2Contract) {
-      Object.values(processedEvents).forEach((event) => (event.from = hubPool.address));
+      processedEvents = events.map((e) => processEvent(e, "_amount", "_to", "from"));
     }
     return {
       [this.resolveL2TokenAddress(l1Token)]: processedEvents,
@@ -139,25 +137,26 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     }
 
     const isL2Contract = await this.isL2ChainContract(toAddress);
-    let events = await paginatedEventQuery(
-      this.l2Eth,
-      this.l2Eth.filters.Transfer(
-        zksync.utils.applyL1ToL2Alias(isL2Contract ? hubPool.address : this.atomicDepositor.address),
-        toAddress
-      ),
-      eventConfig
-    );
-    // For WETH transfers involving an EOA, only count them if a wrap txn followed the L2 deposit finalization.
-    if (!isL2Contract) {
-      const wrapEvents = await paginatedEventQuery(
-        this.l2Weth,
-        this.l2Weth.filters.Transfer(ZERO_ADDRESS, toAddress),
+    let processedEvents;
+    if (isL2Contract) {
+      processedEvents = await paginatedEventQuery(
+        this.l2Eth,
+        this.l2Eth.filters.Transfer(zksync.utils.applyL1ToL2Alias(hubPool.address), toAddress),
         eventConfig
       );
-      events = matchL2EthDepositAndWrapEvents(events, wrapEvents);
+    } else {
+      const [events, wrapEvents] = await Promise.all([
+        paginatedEventQuery(
+          this.l2Eth,
+          this.l2Eth.filters.Transfer(zksync.utils.applyL1ToL2Alias(this.atomicDepositor.address), toAddress),
+          eventConfig
+        ),
+        paginatedEventQuery(this.l2Weth, this.l2Weth.filters.Transfer(ZERO_ADDRESS, toAddress), eventConfig),
+      ]);
+      processedEvents = matchL2EthDepositAndWrapEvents(events, wrapEvents);
     }
     return {
-      [this.resolveL2TokenAddress(l1Token)]: events.map((event) => processEvent(event, "_amount", "_to", "from")),
+      [this.resolveL2TokenAddress(l1Token)]: processedEvents.map((e) => processEvent(e, "_amount", "_to", "from")),
     };
   }
 
--- a/src/adapter/bridges/ZKSyncWethBridge.ts
+++ b/src/adapter/bridges/ZKSyncWethBridge.ts
@@ -16,6 +16,7 @@ import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./Bas
 import { processEvent } from "../utils";
 import * as zksync from "zksync-ethers";
 import { zkSync as zkSyncUtils } from "../../utils/chains";
+import { matchL2EthDepositAndWrapEvents } from "../../clients/bridges/utils";
 
 /* For both the canonical bridge and the ZkSync Weth bridge (this
  * bridge), we need to assume that the l1 and l2 signers contain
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -111,8 +106,11 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
       eventConfig
     );
     const processedEvents = events.map((event) =>
-      isL2Contract ? processEvent(event, "amount", "to", "l2Token") : processEvent(event, "_amount", "_to", "from")
+      isL2Contract ? processEvent(event, "amount", "to", "to") : processEvent(event, "_amount", "_to", "from")
     );
+    if (isL2Contract) {
+      Object.values(processedEvents).forEach((event) => (event.from = hubPool.address));
```
nicholaspai, line 112: Is there a `from` field on the TokensRelayed event..?

pxrl: What is the original `event.from` that's being replaced here btw?

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -124,14 +122,10 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     toAddress: string,
     eventConfig: EventSearchConfig
   ): Promise<BridgeEvents> {
-    // This fn will also only work to track EOA's or the SpokePool's transfers, so exclude the hub pool
-    // and any L2 contracts that are not the SpokePool.
-    if (fromAddress === this.getHubPool().address) {
-      return Promise.resolve({});
-    }
```
pxrl, line 131: I recognise from the old adapter; do you have a tldr on why this previosuly worked despite that implementation?

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: The `event.from` was some throwaway value. It is the spoke pool. `TokensRelayed` has no `from`: https://github.com/across-protocol/contracts/blob/2b2e62437595108a67460730c3600ae7f2d7d38f/contracts/chain-adapters/interfaces/AdapterInterface.sol#L12

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: I was wrong in thinking this was the issue. This has been added back. https://github.com/across-protocol/relayer/pull/1773/commits/701f2eac6dc9f9da30aeb41fa4b276b62c361690

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -90,29 +91,34 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     toAddress: string,
     eventConfig: EventSearchConfig
   ): Promise<BridgeEvents> {
-    // This fn will only work to track EOA's or the SpokePool's transfers, so exclude the hub pool
-    // and any L2 contracts that are not the SpokePool.
-    if (fromAddress === this.getHubPool().address) {
+    const hubPool = this.getHubPool();
+    if (fromAddress === hubPool.address) {
```
james-a-morris, line 95: We should probably use `compareAddressSimple` for case sensitivity

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -90,29 +91,34 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
...
+    let events = await paginatedEventQuery(
       isL2Contract ? hubPool : this.atomicDepositor,
       isL2Contract
         ? hubPool.filters.TokensRelayed()
         : this.atomicDepositor.filters.ZkSyncEthDepositInitiated(fromAddress, toAddress),
       eventConfig
     );
+    if (isL2Contract) {
+      // The TokensRelayed event does not distinguish between chain or l1Token, so we filter it here.
+      events = events.filter((e) => e.args.to === toAddress && e.args.l1Token === wethAddress);
```
james-a-morris, line 114: similar to above

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -90,29 +91,34 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
...
+    }
     const processedEvents = events.map((event) =>
-      isL2Contract ? processEvent(event, "amount", "to", "l2Token") : processEvent(event, "_amount", "_to", "from")
+      isL2Contract ? processEvent(event, "amount", "to", "to") : processEvent(event, "_amount", "_to", "from")
     );
+    if (isL2Contract) {
+      Object.values(processedEvents).forEach((event) => (event.from = hubPool.address));
```
james-a-morris, line 120: I'm a bit confused here. We're iterating over each and setting `event.from` to the hub pool address? Would it make sense to use `map ((e) => {...e, from: hubPool.address})` and not impact the existing array

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -124,15 +130,13 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     toAddress: string,
     eventConfig: EventSearchConfig
   ): Promise<BridgeEvents> {
-    // This fn will also only work to track EOA's or the SpokePool's transfers, so exclude the hub pool
-    // and any L2 contracts that are not the SpokePool.
-    if (fromAddress === this.getHubPool().address) {
+    const hubPool = this.getHubPool();
+    if (fromAddress === hubPool.address) {
```
james-a-morris, line 134: Similar to above

bmzig: Good call. https://github.com/across-protocol/relayer/pull/1773/commits/e6ec31598093bff828bdbe0a1ac4f6f4345e46ef

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: https://github.com/across-protocol/relayer/pull/1773/commits/e6ec31598093bff828bdbe0a1ac4f6f4345e46ef

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: https://github.com/across-protocol/relayer/pull/1773/commits/e6ec31598093bff828bdbe0a1ac4f6f4345e46ef

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: This is just an edit in place so I'm not sure if we want to map here.

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -90,29 +92,36 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
...
+      events = events.filter(
+        (e) => compareAddressesSimple(e.args.to, toAddress) && compareAddressesSimple(e.args.l1Token, wethAddress)
+      );
+    }
     const processedEvents = events.map((event) =>
-      isL2Contract ? processEvent(event, "amount", "to", "l2Token") : processEvent(event, "_amount", "_to", "from")
+      isL2Contract ? processEvent(event, "amount", "to", "to") : processEvent(event, "_amount", "_to", "from")
     );
+    if (isL2Contract) {
+      Object.values(processedEvents).forEach((event) => (event.from = hubPool.address));
```
nicholaspai, line 123: Was this edit in place in the original clients/bridges code? Feels like code smell that's going to be hard to maintain or lead to subtle bugs in the future

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -141,10 +148,13 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
...
-    if (isL2Contract) {
-      events.concat(
-        await paginatedEventQuery(this.l2Weth, this.l2Weth.filters.Transfer(ZERO_ADDRESS, fromAddress), eventConfig)
+    if (!isL2Contract) {
+      const wrapEvents = await paginatedEventQuery(
+        this.l2Weth,
+        this.l2Weth.filters.Transfer(ZERO_ADDRESS, toAddress),
+        eventConfig
       );
+      events = matchL2EthDepositAndWrapEvents(events, wrapEvents);
```
nicholaspai, line 157: Did we forget to match wrap events when copying code from clients/bridges to adapters..? Confused why this is being added now

nicholaspai (commented):
I really think we need to spend time in a follow up pr eventually and refactor how we track outstanding transfers as the code in these files is not good code. Obviously most of this is copied from older code but there are too many branches to deal with, too many edge cases, and my instinct is this code is going to be very hard to debug in the future when something goes wrong.

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: This is not in the origin clients/bridges code since they don't transform the events into a `BridgeEvent` interface. This field is also never read at the moment outside of this file. Since there is no `from` in the `TokensRelayed` event, so I needed to manually overwrite that field to set it to the hub pool's address.

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: Yes. It was just tracking eth transfers and not the follow up weth transfers.

src/adapter/bridges/ZKSyncWethBridge.ts
pxrl, line 123: For future consideration: All of these statements are conditional upon `isL2Contract`; I wonder if it's worth refactoring it to be something like:

```
if (isL2Contract) {
  // ...
} else {
  // ...
}

File: src/adapter/bridges/ZKSyncWethBridge.ts [hunk 9]
```diff
@@ -90,29 +92,36 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
     toAddress: string,
     eventConfig: EventSearchConfig
   ): Promise<BridgeEvents> {
-    // This fn will only work to track EOA's or the SpokePool's transfers, so exclude the hub pool
-    // and any L2 contracts that are not the SpokePool.
-    if (fromAddress === this.getHubPool().address) {
+    const hubPool = this.getHubPool();
+    if (compareAddressesSimple(fromAddress, hubPool.address)) {
       return Promise.resolve({});
     }
```
pxrl, line 98: I still dream of being able to drop `hubPool` as a monitored address. Maybe one day...

src/adapter/bridges/ZKSyncWethBridge.ts
nicholaspai: Why don't you set this field in the map above so you don't need to iterate through the list twice?

nicholaspai (commented):
one comment

src/adapter/bridges/ZKSyncWethBridge.ts
```diff
@@ -90,29 +92,34 @@ export class ZKSyncWethBridge extends BaseBridgeAdapter {
...
-    );
+    let events, processedEvents;
+    if (isL2Contract) {
+      events = await paginatedEventQuery(hubPool, hubPool.filters.TokensRelayed(), eventConfig);
+      processedEvents = events
+        .filter(
+          (e) => compareAddressesSimple(e.args.to, toAddress) && compareAddressesSimple(e.args.l1Token, wethAddress)
+        )
+        .map((e) => processEvent(e, "amount", "to", "to"));
+      processedEvents.forEach((e) => (e.from = hubPool.address));
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/1773/commits/3dc32d933c1037053b244592971319a3cfd42617

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: https://github.com/across-protocol/relayer/pull/1773/commits/8cfb639ec3a76060b12964f168d8cbc0797bf446

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: Fixed here: https://github.com/across-protocol/relayer/pull/1773/commits/3dc32d933c1037053b244592971319a3cfd42617

src/adapter/bridges/ZKSyncWethBridge.ts
bmzig: Addressed here: https://github.com/across-protocol/relayer/pull/1773/commits/3dc32d933c1037053b244592971319a3cfd42617

PR #1777, opened 2024-08-23
improve(SpokePoolClient): Log on failed listener

pxrl:
In the event that a child process terminates improperly, ensure that it is logged by the relayer. This will be extended in future to support restarting the listener process.


mrice32 (approved):
LGTM!

PR #1779, opened 2024-08-23
improve(relayer): take refunds on origin when filling on a lite chain.

bmzig:
We will sometimes take repayments on on lite chains for routes from non-lite chains to lite chains. This is undesirable since it requires the relayer to manually pull funds back via the canonical bridge. We should prioritize taking repayment on non-lite chains whenever possible, so this PR makes the inventory client always select either a slow repayment chain or the origin chain when filling on a lite chain.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -555,8 +555,7 @@ export class InventoryClient {
       // It's undesirable to accrue excess balances on a Lite chain because the relayer relies on additional deposits
       // destined for that chain in order to offload its excess.
       const { targetOverageBuffer = DEFAULT_TOKEN_OVERAGE } = tokenConfig;
-      const isLiteChain = (deposit.toLiteChain && chainId === destinationChainId) || (deposit.fromLiteChain && chainId === originChainId);
-      const ignoreOverageBuffer = isLiteChain && chainId != originChainId;
+      const ignoreOverageBuffer = (deposit.toLiteChain && chainId === destinationChainId) && chainId != originChainId;
       const effectiveTargetPct = ignoreOverageBuffer ? tokenConfig.targetPct : tokenConfig.targetPct.mul(targetOverageBuffer).div(fixedPointAdjustment);
 
       this.log(
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -555,8 +555,10 @@ export class InventoryClient {
       // It's undesirable to accrue excess balances on a Lite chain because the relayer relies on additional deposits
       // destined for that chain in order to offload its excess.
       const { targetOverageBuffer = DEFAULT_TOKEN_OVERAGE } = tokenConfig;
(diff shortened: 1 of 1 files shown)


src/clients/InventoryClient.ts
```diff
@@ -557,11 +557,12 @@ export class InventoryClient {
       const { targetOverageBuffer = DEFAULT_TOKEN_OVERAGE } = tokenConfig;
       const effectiveTargetPct = tokenConfig.targetPct.mul(targetOverageBuffer).div(fixedPointAdjustment);
 
+      const takeRepayment = expectedPostRelayAllocation.lte(effectiveTargetPct) || (deposit.toLiteChain && chainId === originChainId);
```
pxrl, line 560: When the destination chain is a lite chain and we are evaluating where to take repayment, I think we should use `targetPct` for lite chains, and otherwise `targetPct.mul(targetOverageBuffer).div(fixedPointAdjustment)`.

`targetOverageBuffer` is intended to permit the relayer to hold additional inventory on chains where it helps the protocol, but that's not the case for lite chains because the dataworker explicitly has to bridge tokens to that Spoke in order to make repayment.

pxrl: (Also just to tease this out a bit more, I wondered whether we should be setting `targetOverageBuffer` to 0 for lite chains. This is undesirable because it caps how much volume we can support _out_ of a lite chain. We want origin chain repayments to be able to inflate up to `targetPct * targetOverageBuffer`, but when the destination chain is a lite chain, we want to try to pull its balance back to `targetPct` by taking repayment elsewhere)

src/clients/InventoryClient.ts
```diff
@@ -555,7 +555,9 @@ export class InventoryClient {
       // It's undesirable to accrue excess balances on a Lite chain because the relayer relies on additional deposits
       // destined for that chain in order to offload its excess.
       const { targetOverageBuffer = DEFAULT_TOKEN_OVERAGE } = tokenConfig;
-      const effectiveTargetPct = tokenConfig.targetPct.mul(targetOverageBuffer).div(fixedPointAdjustment);
+      const isLiteChain = (deposit.toLiteChain && chainId === destinationChainId) || (deposit.fromLiteChain && chainId === originChainId);
```
bmzig, line 558: Should we add an `isLiteChain` function in NetworkUtils in the SDK? I think so...

pxrl, line 558: Do we need the 2nd half of the or statement? We negate that case when setting `ignoreOverageBuffer`.

bmzig: True

bmzig: https://github.com/across-protocol/relayer/pull/1779/commits/5bfa4e7929b5fcc9cafd52e2e26ee17b403c576b

pxrl (approved):
LFG!

PR #1781, opened 2024-08-24
improve(relayer): Support dynamic handover in production

pxrl:
The fast relayer can presently have multiple concurrent instances for a short period of time because there is no method of coordinating the "active" relayer. This can cause some collisions when both relayer instances attempt to fill the same transaction(s), and requires that the relayer be stopped via a SIGHUP, which is slightly awkward in the serverless environment.

This change instead uses redis as a coordination mechanism between relayer instances, such that a new relayer will update a specific redis record when it is ready to take over from any previous instance. Likewise, and existing instance will detect when the record is updated and will gracefully exit thereafter.

### DIFF
--- a/src/relayer/index.ts
+++ b/src/relayer/index.ts
@@ -5,6 +5,7 @@ import {
   disconnectRedisClients,
   getCurrentTime,
   getNetworkName,
+  getRedisCache,
   Signer,
   startupLogLevel,
   winston,
@@ -15,6 +16,7 @@ import { constructRelayerClients, updateRelayerClients } from "./RelayerClientHe
 config();
 let logger: winston.Logger;
 
+const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier } = process.env;
 const randomNumber = () => Math.floor(Math.random() * 1_000_000);
 
 export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
@@ -33,6 +35,9 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
     });
     stop = true;
   });
+
+  const redis = await getRedisCache(logger);
+  let setActiveRelayer = false;
 
   // Explicitly don't log ignoredAddresses because it can be huge and can overwhelm log transports.
   const { ignoredAddresses: _ignoredConfig, ...loggedConfig } = config;
@@ -84,6 +89,24 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
           at: "Relayer#run",
           message: `Completed relayer execution loop ${run++} in ${runTime} seconds.`,
         });
+
+        // Signal to any existing relayer that a handover is underway, or alternatively
+        // check for a handover initiated by another (newer) relayer instance.
+        if (botIdentifier && runIdentifier) {
+          const activeRelayer = await redis.get(botIdentifier);
+          if (activeRelayer !== runIdentifier) {
+            if (!setActiveRelayer) {
+              await redis.set(botIdentifier, runIdentifier);
+              setActiveRelayer = true;
+            } else {
+              logger.debug({
+                at: "Relayer#run",
+                message: `Handing over to ${botIdentifier} instance ${activeRelayer}.`
+              });
+              stop = true;
+            }
+          }
+        }
 
         if (!stop && runTime < config.pollingDelay) {
           const delta = config.pollingDelay - runTime;


src/relayer/index.ts
```diff
@@ -85,6 +90,24 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
         });
 
+        // Signal to any existing relayer that a handover is underway, or alternatively
+        // check for a handover initiated by another (newer) relayer instance.
+        if (botIdentifier && runIdentifier) {
+          const activeRelayer = await redis.get(botIdentifier);
+          if (activeRelayer !== runIdentifier) {
+            if (!setActiveRelayer) {
+              await redis.set(botIdentifier, runIdentifier);
+              setActiveRelayer = true;
```
pxrl, line 100: I'd like to additionally make this conditional on the relayer being fully functional on all spokes. At present this is not guaranteed in the initial loops, so there is some benefit to the existing relayer continuing to run in that case. I'm not settled on which metric to use to infer readiness...SpokePoolClient.isUpdated would be ideal but it's set `true` on each SpokePoolClient update even if no events have been received yet.

james-a-morris (commented):
is it possible to make this more generic for any/all bots? Or do we just want to implement this now for the fast relayer

pxrl:
> is it possible to make this more generic for any/all bots? Or do we just want to implement this now for the fast relayer

The relayer in looping mode is currently the only bot that's really suited for this, because it has a natural recurring point to check for handover. The other bots don't yet have this so it's not really clear for me how they'd benefit from it...but in a world where they are more event-based (i.e. precomputing executions and waiting for the right block to execute them) then I think this could be pretty beneficial and would definitely be worth generalising.

src/clients/SpokePoolClient.ts
```diff
@@ -260,6 +260,11 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
   }
 
   protected async _update(eventsToQuery: string[]): Promise<clients.SpokePoolUpdate> {
+
+    if (this.pendingBlockNumber === this.deploymentBlock) {
+      return { success: false, reason: clients.UpdateFailureReason.AlreadyUpdated }; // @todo: Update reason
```
pxrl, line 265: To be changed once https://github.com/across-protocol/sdk/pull/729 is merged.

```suggestion
      return { success: false, reason: clients.UpdateFailureReason.NotReady };
```

src/relayer/index.ts
pxrl: Done: https://github.com/across-protocol/relayer/pull/1781/commits/e69356ec881f8c87b96916b1ac4f173b9302bc95

src/relayer/index.ts
```diff
@@ -25,46 +37,76 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
       const tLoopStart = performance.now();
+      const ready = await relayer.update();
+      const activeRelayer = await redis.get(botIdentifier);
 
-      await relayer.update();
-      txnReceipts = await relayer.checkForUnfilledDepositsAndFill(enableSlowFills, simulate);
-      await relayer.runMaintenance();
+      // If there is another active relayer, allow up to 10 update cycles for this instance to be ready,
+      // then proceed unconditionally to protect against any RPC outages blocking the relayer.
+      if (!ready && activeRelayer && run < 10) {
```
pxrl, line 65: I thought about making the constraint of 10 loops configurable; haven't ruled it out yet, but I'm not sure we'd ever use it and I don't want to add a configurable that will never be tuned in practice.

src/relayer/index.ts
```diff
@@ -25,46 +37,76 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
+        await delay(delta);
+        continue;
+      }
+
+      // Signal to any existing relayer that a handover is underway, or alternatively check for a handover initiated by
+      // another (newer) relayer instance. The active relayer can also be expired, in which case this relayer should
+      // reconfirm its status as the active relayer.
+      if (loop && botIdentifier && runIdentifier) {
+        if (activeRelayer !== runIdentifier) {
+          if (!setActiveRelayer && activeRelayer) {
```
nicholaspai, line 78: How could activeRelayer be undefined if we check it above?

nicholaspai (approved):

src/relayer/index.ts
pxrl: I was playing with the idea of setting active relayer expiry to be << the expected runtime, such that the active relayer would need to re-confirm itself multiple times during a run. That would be nice in the event that an existing relayer goes down and the new one holds off becoming active because it thinks the old one is still there.

I backed it out because there are potential race conditions if the new and old relayer instances overlap when setting this. i.e. if the existing relayer sees that the active relayer reverts to null due to expiry, then that might collide with a new relayer taking over. I backed it out for now because it'd only be useful in rare circumstances; I'll think about how it can be done more robustly (...might require 2 redis records).

https://github.com/across-protocol/relayer/pull/1781/commits/39cb4579e4c794865a678c2a7d13f07ceb9d5057

src/clients/SpokePoolClient.ts
pxrl: Done: https://github.com/across-protocol/relayer/pull/1781/commits/491e94787f4f3ebbe60cf950dc0ff87c27841f2b

PR #1783, opened 2024-08-26
improve(Relayer): Reduce size of overallocated-lite-chain log and send to special channel

nicholaspai:
No reason to log the full deposit object, and send to unprofitable fill channel. You can argue that an over allocated lite chain produces deposits that are unprofitable from a rebalancing perspective to fill

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -916,7 +916,7 @@ export class Relayer {
         message: deposit.fromLiteChain
           ? `Deposit ${depositId} originated from over-allocated lite chain`
           : `Unable to identify a preferred repayment chain for ${originChain} deposit ${depositId}.`,
-        deposit,
+        notificationPath: "across-unprofitable-fills",
       });
       return {
         repaymentChainProfitability: {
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -916,7 +916,7 @@ export class Relayer {
         message: deposit.fromLiteChain
           ? `Deposit ${depositId} originated from over-allocated lite chain`
           : `Unable to identify a preferred repayment chain for ${originChain} deposit ${depositId}.`,
-        deposit,
+        notificationPath: "across-unprofitable-fills",
```
pxrl, line 919: It'll help if we're able to find the deposit in the block explorer, so I think we should log these fields.

```suggestion
        originChainId: deposit.originChainId,
        transactionHash: deposit.transactionHash,
        notificationPath: "across-unprofitable-fills",
```

nicholaspai: https://github.com/across-protocol/relayer/pull/1783/commits/7d41089b5de66f96702dbbc1de52293b74f617c5

james-a-morris (approved):

pxrl (approved):

PR #1785, opened 2024-08-26
fix(BundleDataClient): Don't mark deposits in the future as invalid

nicholaspai:
Bundle data client incorrectly marks a deposit as invalid when its origin chain block number is greater than the origin chain bundle end block
- This bug was discovered after https://github.com/across-protocol/relayer/commit/aa1ec0d4b9fe6f902be298632dd01e786e07c17c was deployed and it marked a deposit as invalid. Ironically, it wasn't a deposit that was too old that was triggered an invalid fill, but a deposit that was in the future that was ignored
- I would label this as a fix for a long-standing bug

Separately, increase lookback required for spoke pool clients to better accomodate situation where a fill is sent well after a deposit was mined but before it expired.
- This is a follow up to https://github.com/across-protocol/relayer/commit/aa1ec0d4b9fe6f902be298632dd01e786e07c17c

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -762,43 +762,40 @@ export class BundleDataClient {
           continue;
         }
 
-        originClient
-          .getDepositsForDestinationChain(destinationChainId)
-          .filter((deposit) => deposit.blockNumber <= originChainBlockRange[1])
-          .forEach((deposit) => {
-            depositCounter++;
-            const relayDataHash = this.getRelayHashFromEvent(deposit);
-            if (v3RelayHashes[relayDataHash]) {
-              // If we've seen this deposit before, then skip this deposit. This can happen if our RPC provider
-              // gives us bad data.
-              return;
-            }
-            // Even if deposit is not in bundle block range, store all deposits we can see in memory in this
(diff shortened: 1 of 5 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -762,43 +762,40 @@ export class BundleDataClient {
           continue;
         }
 
-        originClient
-          .getDepositsForDestinationChain(destinationChainId)
-          .filter((deposit) => deposit.blockNumber <= originChainBlockRange[1])
```
bmzig, line 767: I don't see the difference here. It looks like `olderDepositHashes` and `expiredBundleDepositHashes` will contain the same info after this fix, right?

nicholaspai: The `.filter((deposit) => deposit.blockNumber <= originChainBlockRange[1])` is gone, and now `v3RelayHashes` will contain deposits where `deposit.blockNumber > originChainBlockRange[1]` and then exit shortly after. This way we can now validate fills using the stored deposit

bmzig: I see.

pxrl (approved):

PR #1790, opened 2024-08-28
improve(relayer): Rate-limit repetitive fill-related log messages

bmzig:
We are throwing a lot of `log.warn`s. This PR makes us only log these warns once per new eligible deposit per relayer run.


src/relayer/Relayer.ts
```diff
@@ -1251,4 +1297,13 @@ export class Relayer {
       ? this.clients.tryMulticallClient
       : this.clients.multiCallerClient;
   }
+
+  private getDepositHash(deposit: { depositId: number; originChainId: number }): string {
```
bmzig, line 1301: This function exists in SpokePoolClient.ts in the sdk. Should we move it to a function in utils in the sdk? I don't think we can assume the relayer is holding any spoke pool clients, except for the one on the origin chain (since it was able to observe the deposit there), so otherwise, this can be changed to `this.spokePoolClients[originChainId].getDepositHash(deposit)`.

src/relayer/Relayer.ts
```diff
@@ -180,8 +193,9 @@ export class Relayer {
       return false;
     }
 
+    const logDeposit = this.shouldLogDeposit(deposit);
     if (!this.clients.inventoryClient.validateOutputToken(deposit)) {
-      this.logger[this.config.sendingRelaysEnabled ? "warn" : "debug"]({
+      this.logger[logDeposit ? "warn" : "debug"]({
```
james-a-morris, line 198: I like this in principle, but it hoists the logic of why we're logging into a class function that exists outside this function's scope. I feel like this reduces readability.

src/relayer/Relayer.ts
```diff
@@ -762,12 +786,33 @@ export class Relayer {
     });
 
     // If during the execution run we had shortfalls or unprofitable fills then handle it by producing associated logs.
-    if (tokenClient.anyCapturedShortFallFills()) {
+    const loggableShortfalls = Object.entries(this.depositLogType).filter(
+      ([, logType]) => logType === LogType.Shortfall
+    );
```
james-a-morris, line 791: Are we adding an array iteration to all of these? Does this incur any slowdowns?

src/relayer/Relayer.ts
```diff
@@ -37,10 +37,23 @@ type RepaymentChainProfitability = {
...
+  UnprofitableFill,
+  LoggedUnprofitableFill,
+  Observed,
+}
+
 export class Relayer {
   public readonly relayerAddress: string;
   public readonly fillStatus: { [depositHash: string]: number } = {};
   private pendingTxnReceipts: { [chainId: number]: Promise<TransactionResponse[]> } = {};
+  private depositLogType: { [depositHash: string]: LogType } = {};
```
james-a-morris, line 56: Could this be accomplished with a record of booleans to represent the event being seen before?

james-a-morris (commented):
I'm a bit worried about the added logging complexity and the added array iterations

src/relayer/Relayer.ts
bmzig: The main issue with logging is how we can stop logging the shortfalls. We store the deposit in the token client but only actually keep track of the deposit amount, deposit id, and the destination chain id, which means that when we go inside this function https://github.com/across-protocol/relayer/blob/c7c705fd436e8131938499432e4f7318c65cc152/src/relayer/Relayer.ts#L1106, we don't actually have enough information to get a deposit hash since we need to know both the deposit id and the origin chain id. So either we can refactor the information stored in the inventory client to also keep track of the origin chain id, or we can move the checks outside `handleTokenShortfall()`. As you stated in other comments, this adds runtime complexity. I chose this way at first so I wouldn't need to modify the inventory client, but based off of these comments, that may be the way to go? 

To get back to this question, this can't be a record of booleans unless we change the inventory client since it wouldn't keep enough information about whether or not we logged a shortfall. At minimum, I believe we need a state when the deposit was just observed (i.e. undefined), a state where we can tell the relayer to log the shortfall (e.g. 0), a state where the relayer has logged the shortfall (e.g. 1), and then a state (e.g. 2) for "everything else" (unprofitable fill, observed non-shortfall deposit, etc.). Note "everything else" and the state to log a shortfall can't be combined since once again, we can't derive the deposit hash from within `handleTokenShortfalls()`, so we need to have some special case to only call that if certain conditions about deposits are met.

src/relayer/Relayer.ts
bmzig: See the below comment for this as well. Unless we make changes to the inventory client, I don't think we can avoid this. Although, to be fair, this won't be needed for `handleUnprofitableFills()` since we do have knowledge of the entire deposit in that function.

pxrl: How about we only log about shortfalls on every _n_th loop?

src/relayer/Relayer.ts
pxrl: I'm toying with the idea of precomputing the actual deposit hashes on event ingestion and making it part of the V3FundsDeposited or FilledV3Relay types (i.e. rather than this "hash" variant, which is just a string concatenation of the depositId and originChainId). I think for now we can live this this as-is though.

src/relayer/Relayer.ts
bmzig: That would simplify this significantly. So essentially, pass in the loop number as an argument to `checkForUnfilledDepositsAndFill` and then act accordingly?

bmzig: How about this? https://github.com/across-protocol/relayer/pull/1790/commits/3be14a94f6971883522c08b37fb87fbce5654331

src/relayer/RelayerConfig.ts
```diff
@@ -114,6 +118,7 @@ export class RelayerConfig extends CommonConfig {
     this.minRelayerFeePct = toBNWei(MIN_RELAYER_FEE_PCT || Constants.RELAYER_MIN_FEE_PCT);
 
     this.tryMulticallChains = JSON.parse(RELAYER_TRY_MULTICALL_CHAINS ?? "[]");
+    this.loggingInterval = Number(RELAYER_LOGGING_INTERVAL ?? 25);
```
pxrl, line 121: An interesting variation on this could be to log per x unit of time (i.e. once per minute). That gives us more reliable logging in the face of variable bot loop times.

bmzig: I agree. That probably is more sustainable. Something like this? https://github.com/across-protocol/relayer/pull/1790/commits/4ab6cce81cf1b7a3b61c7a22e75bffe752bb36e8. Logging interval is still there, it's now just time time in seconds before the next shortfall/unprofitable fill log (if available).

src/relayer/RelayerConfig.ts
```diff
@@ -110,6 +114,7 @@ export class RelayerConfig extends CommonConfig {
...
```
pxrl, line 117:
```suggestion
    this.loggingInterval = Number(RELAYER_LOGGING_INTERVAL ?? 30);
```

src/relayer/RelayerConfig.ts
pxrl: I like it!

src/relayer/Relayer.ts
pxrl: We've sidestepped this issue by just rate-limiting the logs. We should be good to go now!

src/relayer/Relayer.ts
pxrl: Has been backed out from this change.

pxrl (approved):

PR #1791, opened 2024-08-29
[WIP] fix(BundleDataClient): Ignore RequestedSlowFill on lite chain deposits

pxrl:


src/clients/BundleDataClient.ts
```diff
@@ -1101,10 +1101,10 @@ export class BundleDataClient {
         if (fillStatus !== FillStatus.Filled) {
           updateExpiredDepositsV3(expiredDepositsToRefundV3, deposit);
         }
-        // If fill status is RequestedSlowFill, then we might need to mark down an unexecutable
-        // slow fill that we're going to replace with an expired deposit refund.
+        // If fill status is RequestedSlowFill and the deposit did not originate on a lite chain, then we might need to
+        // mark down an unexecutable slow fill that we're going to replace with an expired deposit refund.
```
pxrl, line 1105:
```suggestion
        // mark down an unexecutable slow fill that we're going to replace with an expired deposit refund.
```

src/clients/BundleDataClient.ts
```diff
@@ -1101,10 +1101,10 @@ export class BundleDataClient {
...
         // If deposit cannot be slow filled, then exit early.
-        if (fillStatus !== FillStatus.RequestedSlowFill) {
+        if (fillStatus !== FillStatus.RequestedSlowFill || deposit.fromLiteChain) {
```
pxrl, line 1107: One thing to note - I was never able to find a slow fill leaf for this deposit (presumably because we correctly filtered it out due to originating on a lite chain). However, I'm puzzled that we also don't verify the existence of the slow fill leaf as part of this function :thinking:

pxrl:
Closed in favour of https://github.com/across-protocol/relayer/pull/1792

PR #1792, opened 2024-08-29
fix(BundleDataClient): Ignore slow fill requests for lite chains

nicholaspai:
Specifically, patch a couple of branches where we do not currently ignore slow fill requests:
- If a fill replaces a slow fill request for a deposit to or from a lite chain, we should never mark an excess slow fill amount to be returned to the Hub
- If an expired deposit refund supercedes a slow fill request we should never mark an excess slow fill amount to be returned to the Hub

because slow fill requests to or from lite chains are invalid

### DIFF
--- a/src/clients/BundleDataClient.ts
+++ b/src/clients/BundleDataClient.ts
@@ -844,7 +844,12 @@ export class BundleDataClient {
                   // If fill replaced a slow fill request, then mark it as one that might have created an
                   // unexecutable slow fill. We can't know for sure until we check the slow fill request
                   // events.
-                  if (fill.relayExecutionInfo.fillType === FillType.ReplacedSlowFill) {
+                  // slow fill requests for deposits from or to lite chains are considered invalid
+                  if (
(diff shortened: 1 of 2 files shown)


src/clients/BundleDataClient.ts
```diff
@@ -1026,6 +1036,10 @@ export class BundleDataClient {
             fill.relayExecutionInfo.fillType === FillType.ReplacedSlowFill,
             "Fill type should be ReplacedSlowFill."
           );
+          // slow fill requests for deposits from or to lite chains are considered invalid
+          if (deposit.fromLiteChain || deposit.toLiteChain) {
+            return;
+          }
```
pxrl, line 1042: I think we should probably assert that both `fromLiteChain` and `toLiteChain` are false here, since pushing a lite chain event to `fastfillsReplacingSlowFills` should be prevented by the logic at lines 850/851 and 902/903.

pxrl (approved):
LGTM!

PR #1795, opened 2024-08-29
refactor(relayer): Drop SKIP_REBALANCING env configurable

pxrl:
SKIP_REBALANCING and SEND_REBALANCES are opposites, so there's no reason to have support for both.


src/relayer/RelayerClientHelper.ts
```diff
@@ -209,6 +209,7 @@ export async function updateRelayerClients(clients: RelayerClients, config: Rela
   ]);
 
   // Update the token client first so that inventory client has latest balances.
+  clients.tokenClient.clearTokenData();
```
pxrl, line 212: delet

```suggestion
```

nicholaspai (approved):

PR #1798, opened 2024-08-30
feat: Use Viem in finalizer for some OpStack chains

nicholaspai:
## The eth-optimism NPM package is being deprecated
This package is built on `ethers.js` and it is scheduled for [deprecation](https://docs.optimism.io/builders/notices/sdk-deprecation). The finalizer is one of the main packages in this repo that uses external SDK's so the finalizer eventually MUST migrate to the viem version of the optimism SDK. This PR adds OP-stack support for most chains we currently support minus a few that are non-standard.

## Current limitations of Viem:
- ~No easy way to convert ethers Provider to [Viem transport](https://viem.sh/docs/clients/transports/custom), meaning that RPC calls won't go through our custom RetryProvider. The Viem transport does have retry and fallback logic natively but not caching built-in. I think we'll need a larger refactor to support this migration if we wanted to use Viem more widely throughout the codebase. We can probably get away with Viem in the finalizer if we restrict its use only for the few RPC calls required to support finalizer functions (e.g. building transaction args requires querying contracts).~ I defined a custom `request` function to comply with EIP1193.
- This results in some unsightly casts to `any` to get types to match
- ~`getWithdrawalStatus` only supports querying a transaction with a single withdrawal. This isn't very useful for us today since we usually multicall our withdrawals. We need something like [this PR](https://github.com/wevm/viem/pull/2873) in order to use Viem to finalize~ This PR got merged!
- [Viem accounts](https://viem.sh/docs/clients/wallet#json-rpc-accounts) don't seem to be obviously interoperable with ethers Signers

## Suggested roadmap
- ~Use Viem for optimism mainnet, and then a few op stack chains over time.~
- Build generalized Viem modules that could help us install viem into the broader codebase like:
    - converting between Ethers providers/signers and Viem transports/accounts

## Resources
- [viem.sh/op-stack docs](https://viem.sh/op-stack/guides/withdrawals)
- [github](https://github.com/wevm/viem/tree/main/src)


src/finalizer/utils/opStack.ts
```diff
@@ -142,32 +164,166 @@ export async function opStackFinalizer(
...
+  } = {
+    callData: [],
+    withdrawals: [],
+  };
+  if (VIEM_OP_STACK_CHAINS[chainId]) {
+    const hubChainId = chainIsProd(chainId) ? CHAIN_IDs.MAINNET : CHAIN_IDs.SEPOLIA;
+    const publicClientL1 = viem
+      .createPublicClient({
+        chain: chainIsProd(chainId) ? viemChains.mainnet : viemChains.sepolia,
+        transport: viem.http(getCachedProvider(hubChainId, true).providers[0].connection.url),
```
nicholaspai, line 186: this part is annoying: is there a better way to set up a viem.http() object without having to access the ethers.JSONRpcProvider's connection.url prop?

src/finalizer/utils/opStack.ts
```diff
@@ -142,32 +164,166 @@ export async function opStackFinalizer(
...
+      })
+      .extend(publicActionsL1() as any) as any;
```
nicholaspai, line 188: Also really annoying and can't figure out how to get these typings to work, even the viem code itself has some weird castings so maybe its not that unexpected: https://github.com/wevm/viem/blob/80d0b61356f8a5acab2b9e622564e5b6fca53ccc/src/clients/createPublicClient.ts#L89

james-a-morris (commented):
Is this still in flight? My only question is if we'd still look to use viem for this list of chains

src/finalizer/utils/opStack.ts
```diff
@@ -181,32 +210,174 @@ export async function opStackFinalizer(
...
-    hubPoolClient,
-    logger
-  );
+    const withdrawalStatuses: string[] = [];
+    await mapAsync(events, async (event, i) => {
+      // Useful information for event:
+      const l1TokenInfo = getL1TokenInfo(event.l2TokenAddress, chainId);
+      const amountFromWei = convertFromWei(event.amountToReturn.toString(), l1TokenInfo.decimals);
+
+      const receipt = await publicClientL2.getTransactionReceipt({
```
nicholaspai, line 264: I'm pretty sure this RPC call won't go through our cached provider, and I'm unsure how to get our cached provider to act as a proxy for the public client RPC calls.

src/finalizer/utils/opStack.ts
```diff
@@ -181,32 +210,174 @@ export async function opStackFinalizer(
...
+        hash: event.transactionHash as `0x${string}`,
+      });
+      const withdrawal = getWithdrawals(receipt)[logIndexesForMessage[i]];
+      const withdrawalStatus: GetWithdrawalStatusReturnType = await publicClientL1.getWithdrawalStatus({
```
nicholaspai, line 268: Same issue here: https://github.com/across-protocol/relayer/pull/1798/files#r1901126164

src/finalizer/utils/opStack.ts
```diff
@@ -181,32 +210,174 @@ export async function opStackFinalizer(
...
+  };
+  if (VIEM_OP_STACK_CHAINS[chainId]) {
+    const hubChainId = chainIsProd(chainId) ? CHAIN_IDs.MAINNET : CHAIN_IDs.SEPOLIA;
+    // TODO: can't figure out how to get these typecasts to work, despite this code being in the viem.opStack tutorials
+    const publicClientL1 = viem
+      .createPublicClient({
+        chain: chainIsProd(chainId) ? viemChains.mainnet : viemChains.sepolia,
+        transport: viem.http(getCachedProvider(hubChainId, true).providers[0].connection.url),
+      })
+      .extend(publicActionsL1() as any) as unknown as viem.PublicClient & PublicActionsL1;
```
nicholaspai, line 235: This is also really annoying, as I can't get the type of `publicActionsL1` to work with `.extend` and I have to do this ugly cast

nicholaspai:
> Is this still in flight? My only question is if we'd still look to use viem for this list of chains

This something we need to merge in eventually otherwise we'll be stuck on a deprecated SDK that will eventually not work for us. The main hold up IMO is:
- I haven't figured out a clean way to build a Viem `PublicClient` that uses our custom CachingProvider logic (caches certain RPC calls, uses quorum on others, etc.). See https://github.com/across-protocol/relayer/pull/1798/files#r1823028660
- I want to avoid making RPC calls that skirt around our CachingProvider, see https://github.com/across-protocol/relayer/pull/1798/files#r1901126164

bmzig (dismissed):

src/finalizer/utils/opStack.ts
```diff
@@ -182,34 +211,230 @@ export async function opStackFinalizer(
...
+    const amountFromWei = convertFromWei(event.amountToReturn.toString(), l1TokenInfo.decimals);
 
-  return { callData, crossChainMessages: crossChainTransfers };
+    const receipt = await publicClientL2.getTransactionReceipt({
+      hash: event.transactionHash as `0x${string}`,
+    });
+    const withdrawal = getWithdrawals(receipt)[logIndexesForMessage[i]];
+    const withdrawalStatus: GetWithdrawalStatusReturnType = await publicClientL1.getWithdrawalStatus({
+      receipt,
+      chain: undefined,
```
pxrl, line 355: Does `chain` need to be explicitly set to undefined, or can it just be omitted from this and each following call?

```suggestion
```

src/finalizer/utils/opStack.ts
```diff
@@ -182,34 +211,230 @@ export async function opStackFinalizer(
...
+      const { seconds } = await publicClientL1.getTimeToFinalize({
+        chain: undefined,
+        withdrawalHash: withdrawal.withdrawalHash,
+        targetChain: viemOpStackTargetChainParam,
+      });
+      logger.debug({
+        at: `${getNetworkName(chainId)}Finalizer`,
+        message: `Withdrawal ${event.transactionHash} for ${amountFromWei} of ${
+          l1TokenInfo.symbol
+        } is in challenge period for ${seconds / 60 / 60} hours`,
```
pxrl, line 395: Might be nice to avoid a long floating number:

```suggestion
        } is in challenge period for ${Math.floor(seconds / 3600)} hours`,
```

src/finalizer/utils/opStack.ts
```diff
@@ -182,34 +211,230 @@ export async function opStackFinalizer(
...
+        const numProofSubmitters = await crossChainMessenger.numProofSubmitters(withdrawal.withdrawalHash);
+        const proofSubmitter = await crossChainMessenger.proofSubmitters(
+          withdrawal.withdrawalHash,
+          numProofSubmitters - 1
+        );
+        callData = await crossChainMessenger.populateTransaction.finalizeWithdrawalTransactionExternalProof(
+          withdrawal,
+          proofSubmitter
+        );
+      } catch (e) {
```
pxrl, line 415: What kind of exceptions are we expecting here, and do we want to log them anywhere?

src/utils/ProviderUtils.ts
```diff
@@ -185,6 +203,37 @@ export async function getProvider(
...
+ * provider. This allows us to send requests through our RetryProvider that need to be handled by Viem SDK's.
+ */
+export function createViemCustomTransportFromEthersProvider(providerChainId: number): viem.CustomTransport {
+  return viem.custom(
+    {
+      async request({ method, params }) {
+        const provider = getCachedProvider(providerChainId, true);
+        try {
+          return await provider.send(method, params);
+        } catch (error: any) {
```
pxrl, line 217: Does this work with `unknown` instead of `any` ?

```suggestion
        } catch (error: unknown) {
```

src/finalizer/utils/opStack.ts
```diff
@@ -182,34 +211,230 @@ export async function opStackFinalizer(
...
-    crossChainMessenger,
-    hubPoolClient,
-    logger
-  );
+  let callData: Multicall2Call[];
+  let crossChainTransfers: CrossChainMessage[];
+
+  // We are in a hybrid state where ideally we want to use the Viem OpStack SDK for as many chains as possible but
+  // not all chains are easily supported in Viem, so certain chains we'll continue to use the Ethers-based
+  // Optimism SDK.
```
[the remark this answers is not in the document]
nicholaspai: yep i figured out how to get Unichain and Ink to work despite their viem chain definitions missing an `l2OutputOracle`

src/finalizer/utils/opStack.ts
```diff
@@ -182,34 +211,230 @@ export async function opStackFinalizer(
...
 
-  const finalizations = await multicallOptimismFinalizations(
-    chainId,
-    olderTokensBridgedEvents,
-    crossChainMessenger,
-    hubPoolClient,
-    logger
+  const crossChainMessenger = new Contract(
+    VIEM_OP_STACK_CHAINS[chainId].contracts.portal[hubChainId].address,
+    OPStackPortalL1,
```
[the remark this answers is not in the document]
nicholaspai: its not there: https://github.com/wevm/viem/blob/main/src/chains/definitions/unichain.ts

I also like being able to control our ABI's locally

src/finalizer/utils/opStack.ts
nicholaspai: I'm gonna remove this try-catch, it seems dangerous. We can instead check the portal version on-chain to decide which Portal ABI to call: 47fe2d83ac23b0c7ceb7be7c39464eb39f7edb05

src/finalizer/utils/opStack.ts
nicholaspai: I can't get it to compile without explicitly setting chain to undefined. [here's](https://github.com/wevm/viem/blob/c19ba8d7c572aa8ea19c8616799d1db8675cf11d/src/op-stack/actions/getWithdrawalStatus.ts#L50) the types, maybe you have other ideas? The viem types are pretty hard to work with

nicholaspai: https://github.com/across-protocol/relayer/pull/1798/commits/566317e7bdf834c5021f70e9a2ad5da84e92701a

got it working by removing the `extend()`. I think this doesn't work well and there are definitely type mismatches between the `viem` and `viem/op-stack` packages

src/finalizer/utils/opStack.ts
jxom: i don't think you really need this "cached provider" as Viem Clients themself don't perform any [instantiation side-effects](https://github.com/ethers-io/ethers.js/blob/main/src.ts/providers/abstract-provider.ts#L479-L490) like Ethers.

bmzig (approved):

bmzig (approved):

src/finalizer/utils/opStack.ts
nicholaspai: @jxom first of all thank you for the review, I really appreciate it! Secondly, perhaps there's a misunderstanding of how our cached provider works. It uses Redis to cache results of RPC requests, so if we send an `eth_getTransactionReceipt` query, then we cache the full response in Redis and load that the next time we make that query with the same input. We need to cache this outside of the runner's memory, since this specific code is designed to be run in a serverless setting. I don't think the Viem transport offers anything like this (i.e. external caching), right?

nicholaspai:
@jxom thanks again for your review, I responded here: https://github.com/across-protocol/relayer/pull/1798#discussion_r1970634218

PR #1804, opened 2024-09-05
improve(providers): Enable RPC error passthrough

pxrl:
This should provide more descriptive error outputs when something goes wrong during an RPC query.


pxrl:
Reference from the ethers docs:
```
connection.errorPassThrough ⇒ boolean
If false, any server error is thrown as a SERVER_ERROR, otherwise a the processFunc is called with response including the error status intact. (default: false)
```

https://docs.ethers.org/v5/single-page/#/v5/api/utils/web/-%23-ConnectionInfo

james-a-morris (approved):
lgtm - this shouldn't affect the migration work @nicholaspai

PR #1812, opened 2024-09-10
fix(providers): pass logger to provider

james-a-morris:
The SDK provider requires a winston logger or it will default to console transports.


melisaguevara:
@james-a-morris do you think that [this](https://github.com/across-protocol/relayer/pull/1813) is also needed?

pxrl (approved):

james-a-morris:
> @james-a-morris do you think that [this](https://github.com/across-protocol/relayer/pull/1813) is also needed?

Both are complementary

PR #1814, opened 2024-09-10
refactor(relayer): Rate-limit maintenance in looping mode

pxrl:
There's no need to compulsively check ETH and WETH balances in looping mode. This clutters the logs a lot, increases the relayer's exposure to delays on RPC queries and generally unnecessarily increases RPC consumption. Instead, rate-limit these tasks to a default of 1/minute, with the option to override via the `RELAYER_MAINTENANCE_INTERVAL` configurable .

As part of this change, refactor some other calls from src/relayer/index.ts to methods within the relayer class. This is a small step towards conforming on a norm for relayer operation.


src/relayer/Relayer.ts
```diff
@@ -89,6 +94,76 @@ export class Relayer {
...
+  }
+
+  /**
+   * @description Perform inventory management as needed. This is capped to 1/minute in looping mode.
+   */
+  async manageInventory(): Promise<void> {
+    const { inventoryClient, tokenClient } = this.clients;
+
+    const currentTime = getCurrentTime();
+    if (currentTime < this.lastMaintenance + 60) {
```
nicholaspai, line 140: can we make this maintenance period configurable via env and have a reasonable default?

pxrl: Oh; it is actually configurable but I forgot to hook it up here.

pxrl: Fixed: https://github.com/across-protocol/relayer/pull/1814/commits/d354da8e75712de3da45f9716176585fb31744e0

PR #1817, opened 2024-09-11
fix(relayer): Don't prematurely mark deposits as filled

pxrl:
Per current implementation, the relayer can prematurely mark a deposit as filled if it _intends_ to fill it. This can however be derailed in the event that the destination chain has pending fills, in which case the fill transaction is discarded. This creates problems for fill status tracking, which is otherwise marked at the time the transaction is enqueued.

The simple fix is to filter candidate deposits by whether their destination chain has pending fills. A more optimised version, which might follow, would apply a pending status for these fills and would apply that pending status at the time of transaction submission. That's more complex for an (at best) marginal benefit, so proceed with the simple solution for now.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -214,7 +214,7 @@ export class Relayer {
     const fillAmountUsd = profitClient.getFillAmountInUsd(deposit);
     if (!isDefined(fillAmountUsd)) {
       this.logger.debug({
-        at: "Relayer::evaluateFill",
+        at: "Relayer::filterDeposit",
         message: `Skipping ${srcChain} deposit due to uncertain fill amount.`,
         destinationChainId,
         outputToken: deposit.outputToken,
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -589,20 +589,31 @@ export class Relayer {
...
+
+    if (isDefined(this.pendingTxnReceipts[destinationChainId])) {
+      this.logger.info({
+        at: "Relayer::evaluateFill",
+        message: `${destChain} transaction queue has pending fills; skipping ${originChain} deposit ${depositId}...`,
+        originChainId,
+        depositId,
+        transactionHash,
+      });
+      return;
```
nicholaspai, line 605: sorry, i don't fully understand why we can exit early if there are pending txns for this chain?

pxrl: If there are pending fills from a previous loop, we consider the transaction queue full. We can enqueue more fills once that pending transaction has been submitted. This allows the bot to continue to loop without waiting for transaction submission on a slow RPC, which can sometimes take 10s of seconds.

bmzig: Isn't this just for checking whether a relayer can make a fill or not? Why isn't [this](https://github.com/across-protocol/relayer/blob/8733abcb97fc71810fc572de80f3e8e1de09f573/src/relayer/Relayer.ts#L796) enough?

mrice32 (approved):

PR #1824, opened 2024-09-16
improve: swap all Ovm adapters (excl. Optimism/Base) w/ generic adapter

bmzig:
This is the first part of migrating the op-stack adapters. This moves Mode, Redstone, Zora, Blast, and Lisk to the generic adapter format, leaving only Optimism and Base as the only adapters still with the "legacy" structure.

### DIFF
--- a/src/clients/bridges/AdapterManager.ts
+++ b/src/clients/bridges/AdapterManager.ts
@@ -1,10 +1,16 @@
 import { utils } from "@across-protocol/sdk";
-import { spokesThatHoldEthAndWeth, SUPPORTED_TOKENS } from "../../common/Constants";
+import {
+  spokesThatHoldEthAndWeth,
+  SUPPORTED_TOKENS,
+  CUSTOM_BRIDGE,
+  CANONICAL_BRIDGE,
+  DEFAULT_GAS_MULTIPLIER,
+} from "../../common/Constants";
 import { InventoryConfig, OutstandingTransfers } from "../../interfaces";
 import { BigNumber, isDefined, winston, Signer, getL2TokenAddresses, TransactionResponse, assert } from "../../utils";
 import { SpokePoolClient, HubPoolClient } from "../";
 import { ArbitrumAdapter, PolygonAdapter, ZKSyncAdapter, LineaAdapter, OpStackAdapter, ScrollAdapter } from "./";
-import { CHAIN_IDs } from "@across-protocol/constants";
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
 
 import { BaseChainAdapter } from "../../adapter";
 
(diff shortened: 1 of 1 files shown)


src/clients/bridges/AdapterManager.ts
```diff
@@ -71,51 +90,71 @@ export class AdapterManager {
...
+        spokePoolClients,
         ZORA,
+        hubChainId,
+        filterMonitoredAddresses(ZORA),
         logger,
         SUPPORTED_TOKENS[ZORA],
-        spokePoolClients,
-        filterMonitoredAddresses(ZORA)
+        bridges,
+        DEFAULT_GAS_MULTIPLIER[BLAST] ?? 1
```
bmzig, line 157:
```suggestion
        DEFAULT_GAS_MULTIPLIER[ZORA] ?? 1
```

src/adapter/bridges/BlastBridge.ts
```diff
@@ -62,7 +62,7 @@ export class BlastBridge extends BaseBridgeAdapter {
     const l2Bridge = this.getL2Bridge();
     const events = await paginatedEventQuery(
       l2Bridge,
-      l2Bridge.filters.ERC20BridgeFinalized(l1Token, undefined, fromAddress),
+      l2Bridge.filters.ERC20BridgeFinalized(undefined, l1Token, fromAddress),
```
bmzig, line 65: This was actually an issue where I assumed the blast bridge followed the same interface as the OpStackDefaultERC20Bridge finalization events, but actually the token address positions are swapped. I.e. the first position is the address of USDB and the second position is DAI on mainnet, not the other way around.

bmzig:
This is ready to go in.

src/clients/bridges/AdapterManager.ts
```diff
@@ -71,51 +90,71 @@ export class AdapterManager {
...
+      this.adapters[MODE] = new BaseChainAdapter(
+        spokePoolClients,
         MODE,
+        hubChainId,
+        filterMonitoredAddresses(MODE),
         logger,
         SUPPORTED_TOKENS[MODE],
-        spokePoolClients,
-        filterMonitoredAddresses(MODE)
+        bridges,
```
pxrl, line 101: Is `bridges` used anywhere else? Could maybe squash it?

```suggestion
        constructBridges(MODE),
```

pxrl (approved):
LGTM! We should probably wait until Monday to get it in so we can react to any regressions.

pxrl:
@bmzig Should we get this one in today?

src/clients/bridges/AdapterManager.ts
bmzig: Nice. Yeah we can https://github.com/across-protocol/relayer/pull/1824/commits/0d9e84658ad4d8385522e9d91bfaf37549b20fb4

PR #1830, opened 2024-09-20
refactor: Replace ethers Event with Log

pxrl:
This follows a similar change in the SDK, with the same motivations. Reverting to Log instead of Event does not sacrifice any functionality but makes it easier to migrate away from ethers v5.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,7 +42,10 @@ module.exports = {
       "error",
       {
         patterns: [{ group: ["@ethersproject/bignumber"], message: "Use 'src/utils/BNUtils' instead" }],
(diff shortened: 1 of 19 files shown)


pxrl:
nb. This will currently fail until the upstream PR in the SDK is merged and a new release is produced (https://github.com/across-protocol/sdk/pull/738).

pxrl:
I've tested this a bit with the `validateRootBundle.ts` script - it's successfully validating old bundles within the past few days OK.

nicholaspai (approved):
Seems like a simple type swap, LGTM

bmzig (approved):
I think this makes sense, especially since the only fields of `Event` we dropped to create `Log` were the ones we didn't use in the first place.

PR #1831, opened 2024-09-20
fix(tests): handle edge case in probabilistic tests

bmzig:
In the tryMulticall tests, since we randomly determine if a transaction succeeds or fails, there is a chance that all the transactions in a bundle fail, meaning that we discard the bundle entirely and do not execute any transaction. Our checks assume that this does not happen, so this PR adds the edge case in our tests so we will no longer fail on these tests on occasion.

### DIFF
--- a/test/TryMulticallClient.ts
+++ b/test/TryMulticallClient.ts
@@ -386,11 +386,14 @@ describe("TryMulticallClient", async function () {
           } as AugmentedTransaction;
           multiCaller.enqueueTransaction(txn);
         }
+        const nBundles = nSucceed === 0 ? 0 : 1;
         expect(multiCaller.transactionCount()).to.equal(nTxns);
         const results = await multiCaller.executeTxnQueue(chainId, false);
-        expect(results.length).to.equal(1);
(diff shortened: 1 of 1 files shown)


test/TryMulticallClient.ts
```diff
@@ -417,10 +421,21 @@ describe("TryMulticallClient", async function () {
             mrkdwn: `This transaction is expected to ${succeed === 0 ? "fail" : "pass"} simulation.`,
           } as AugmentedTransaction;
           multiCaller.enqueueTransaction(txn);
+          queueCopy.push(succeed);
+        }
+        // Determine if all 5 transactions in a bundle failed. This assumes chunkSize | nTxns.
+        let emptyBundles = 0;
+        const nBundles = nTxns / chunkSize;
```
pxrl, line 428: Does it make sense to round this to an integer?

bmzig: It assumes that `nTxns` is a multiple of `chunkSize`, so there should theoretically be no rounding here.

pxrl (approved):

PR #1834, opened 2024-09-23
improve(TransactionUtils): Permit env-based nonce reset

pxrl:
Permit operators to trigger a nonce reset in case of a queue of pending transactions.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -28,6 +28,8 @@ export type TransactionSimulationResult = {
 
 const { isError, isEthersError } = typeguards;
 
+const nonceReset: { [chainId: number]: boolean } = {};
+
 const txnRetryErrors = new Set(["INSUFFICIENT_FUNDS", "NONCE_EXPIRED", "REPLACEMENT_UNDERPRICED"]);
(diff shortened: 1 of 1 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -61,7 +63,13 @@ export async function runTransaction(
   nonce: number | null = null,
   retriesRemaining = 2
 ): Promise<TransactionResponse> {
-  const chainId = (await contract.provider.getNetwork()).chainId;
+  const { provider } = contract;
+  const { chainId } = await provider.getNetwork();
+
+  if (process.env[`ACROSS_RESET_NONCE_${chainId}`] && !nonceReset[chainId]) {
+    nonce = await provider.getTransactionCount(await contract.signer.getAddress());
```
mrice32, line 70: What does this return when there are pending txns? The nonce after the pending txns? Or the nonce after the most recently executed txn?

pxrl: It returns confirmed by default, but blockTag `pending` can optionally be set. I _believe_ ethers is using `pending` because it has increased the nonce by > 100.

mrice32 (approved):

PR #1837, opened 2024-09-24
[WIP] improve(relayer): Improve resilience against non-updated SpokePoolClient

pxrl:
The BundleDataClient asserts that all SpokePoolClient instances have been updated. Ideally it could handle a non-updated SpokePoolClient, but in advance of that, filter the relayer's set of SpokePoolClients to exclude any that have not been updated after some defined time period.

### DIFF
--- a/src/relayer/index.ts
+++ b/src/relayer/index.ts
@@ -61,9 +61,9 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       const ready = await relayer.update();
       const activeRelayer = redis ? await redis.get(botIdentifier) : undefined;
 
-      // If there is another active relayer, allow up to 10 update cycles for this instance to be ready,
+      // If there is another active relayer, allow up to 60 seconds for this instance to be ready,
       // then proceed unconditionally to protect against any RPC outages blocking the relayer.
-      if (!ready && activeRelayer && run < 10) {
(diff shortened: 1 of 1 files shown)


src/relayer/index.ts
```diff
@@ -61,14 +61,32 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
...
+          const delta = pollingDelay - runTime;
+          logger.debug({ at: "Relayer#run", message: `Not ready to relay, waiting ${delta} seconds.` });
+          await delay(delta);
+          continue;
+        }
+
+        // Some SpokePoolClients have not updated. Filter out the ones that are not updated to avoid upsetting the
+        // BundleDataClient.
+        let droppedSpokePools: string[] = [];
+        spokePoolClients = relayerClients.spokePoolClients = Object.fromEntries(
```
pxrl, line 78: nb. There are a few scenarios to test with this and I haven't made it that far, but this may not be the only change required. We should also consider whether these chain IDs should be removed from configured origin/destination chains, and possibly other configurables.

nicholaspai (commented):
For the proposer/disputer to work, all spoke pool clients must be updated. Otherwise we'll always create in valid bundles and worst case send disputes. Why would this be a valid fix?

PR #1838, opened 2024-09-24
improve(relayer): Permit up to 120s for SpokePoolClient readiness

pxrl:
The existing limit of 10 loops was somewhat imprecise and was too short as observed in production. This would occasionally trigger asserts in the BundleDataClient because it requires all SpokePoolClient instances to be updated. Instead, permit up to 120 seconds before proceeding.

### DIFF
--- a/src/relayer/index.ts
+++ b/src/relayer/index.ts
@@ -61,9 +61,9 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       const ready = await relayer.update();
       const activeRelayer = redis ? await redis.get(botIdentifier) : undefined;
 
-      // If there is another active relayer, allow up to 10 update cycles for this instance to be ready,
+      // If there is another active relayer, allow up to 60 seconds for this instance to be ready,
       // then proceed unconditionally to protect against any RPC outages blocking the relayer.
-      if (!ready && activeRelayer && run < 10) {
+      if (!ready && activeRelayer && run * pollingDelay < 60) {
         const runTime = Math.round((performance.now() - tLoopStart) / 1000);
         const delta = pollingDelay - runTime;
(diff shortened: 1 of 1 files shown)


src/relayer/index.ts
```diff
@@ -61,9 +61,9 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       const ready = await relayer.update();
       const activeRelayer = redis ? await redis.get(botIdentifier) : undefined;
 
-      // If there is another active relayer, allow up to 10 update cycles for this instance to be ready,
+      // If there is another active relayer, allow up to 60 seconds for this instance to be ready,
       // then proceed unconditionally to protect against any RPC outages blocking the relayer.
-      if (!ready && activeRelayer && run < 10) {
+      if (!ready && activeRelayer && run * pollingDelay < 60) {
```
nicholaspai, line 66: If we pass through 60s and not all spoke pool clients are updated, the bundle data client will error. Is there a way we can force wait until they're all updated (or timeout with loud error) for the dataworker run?

nicholaspai (commented):
This seems potentially still problematic?

src/relayer/index.ts
pxrl: This is only on the relayer, so the dataworker is unaffected.

We can continue to wait in the relayer, but in the short term I think it's better to throw (and permit serverless to retry). I think we should follow this change up with 2 subsequent changes:
- Permit the relayer to proceed with the subset of SpokePoolClient instances that actually did update successfully, AND
- Permit the BundleDataClient to ignore chains that were not updated successfully (this one might be more tricky).

For (1), I have a draft PR here but it needs testing: https://github.com/across-protocol/relayer/pull/1837/files

nicholaspai (approved):

PR #1840, opened 2024-09-26
 improve(unwrapWeth): load WETH addresses from constants

nicholaspai:
Simple refactor to re-use constants


scripts/unwrapWeth.ts
```diff
@@ -23,6 +23,7 @@ const WETH_ADDRESSES = {
   288: "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
   324: "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",
   8453: "0x4200000000000000000000000000000000000006",
+  7777777: "0x4200000000000000000000000000000000000006",
```
pxrl, line 26: Should we just `import { TOKEN_SYMBOLS_MAP } from ../src/utils` and dereference `TOKEN_SYMBOLS_MAP.WETH.addresses[chainId]` ? Then we get these addresses for free.

pxrl (approved):

scripts/unwrapWeth.ts
```diff
@@ -36,10 +34,10 @@ export async function run(): Promise<void> {
   const signerAddr = await baseSigner.getAddress();
   const chainId = Number(args.chainId);
   const connectedSigner = baseSigner.connect(await getProvider(chainId));
-  if (!isKeyOf(chainId, WETH_ADDRESSES)) {
+  if (!TOKEN_SYMBOLS_MAP.WETH.addresses[chainId]) {
```
james-a-morris, line 37: Could we convert this to an `assert( )`

james-a-morris (approved):
one nit

PR #1842, opened 2024-09-26
feat(relayer): Add viem-based external SpokePool listener

pxrl:
viem supports keepalives and timeouts on websocket transports. In testing it has appeared to successfully websocket connections for longer than ethers v5.

### DIFF
--- a/src/libexec/SpokePoolListenerExperimental.ts
+++ b/src/libexec/SpokePoolListenerExperimental.ts
@@ -1,0 +1,269 @@
+import assert from "assert";
+import minimist from "minimist";
+import { Contract, providers as ethersProviders, utils as ethersUtils } from "ethers";
+import { BaseError, Block, createPublicClient, Log as viemLog, webSocket } from "viem";
+import * as chains from "viem/chains";
+import { utils as sdkUtils } from "@across-protocol/sdk";
+import * as utils from "../../scripts/utils";
+import {
+  CHAIN_IDs,
+  disconnectRedisClients,
+  EventManager,
+  exit,
+  isDefined,
+  getBlockForTimestamp,
+  getChainQuorum,
+  getDeploymentBlockNumber,
+  getNetworkName,
+  getNodeUrlList,
+  getOriginFromURL,
+  getProvider,
+  getRedisCache,
+  Logger,
+  winston,
+} from "../utils";
+import { ScraperOpts } from "./types";
+import { postEvents, removeEvent } from "./util/ipc";
+import { getEventFilterArgs, scrapeEvents as _scrapeEvents } from "./util/evm";
+
+const { NODE_SUCCESS, NODE_APP_ERR } = utils;
+
+const INDEXER_POLLING_PERIOD = 2_000; // ms; time to sleep between checking for exit request via SIGHUP.
+
+let logger: winston.Logger;
+let chainId: number;
+let chain: string;
+let stop = false;
+let oldestTime = 0;
+
+// This mapping is necessary because viem imposes extremely narrow type inference. @todo: Improve?
+const _chains = {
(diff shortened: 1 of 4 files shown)


src/relayer/RelayerClientHelper.ts
```diff
@@ -94,6 +94,7 @@ export async function constructRelayerClients(
     spokePoolClients = Object.fromEntries(
       await sdkUtils.mapAsync(enabledChains ?? configStoreClient.getEnabledChains(), async (chainId) => {
         const opts = {
+          path: config.indexerPath,
```
pxrl, line 97:
```suggestion
```

src/libexec/SpokePoolListenerExperimental.ts
```diff
@@ -0,0 +1,262 @@
...
+
+const INDEXER_POLLING_PERIOD = 2_000; // ms; time to sleep between checking for exit request via SIGHUP.
+
+let logger: winston.Logger;
+let chainId: number;
+let chain: string;
+let stop = false;
+let oldestTime = 0;
+
+// This mapping is necessary because viem imposes extremely narrow type inference. @todo: Improve?
```
nicholaspai, line 39: yeah this is a downside of viem, their chain objects are very opinionated

src/libexec/SpokePoolListenerExperimental.ts
```diff
@@ -0,0 +1,262 @@
...
+ */
+async function listen(eventMgr: EventManager, spokePool: Contract, eventNames: string[], quorum = 1): Promise<void> {
+  const urls = getNodeUrlList(chainId, quorum, "wss");
+  let nProviders = urls.length;
+  assert(nProviders >= quorum, `Insufficient providers for ${chain} (required ${quorum} by quorum)`);
+
+  const providers = urls.map((url) =>
+    createPublicClient({
+      chain: _chains[chainId],
+      transport: webSocket(url, { name: getOriginFromURL(url) }),
```
nicholaspai, line 90: Nice way to convert from URL's to viem transport. I haven't found a clean way to do this from an ethers provider, like here https://github.com/across-protocol/relayer/pull/1798/files#diff-0ca2c3f714bc267e8238e7f50ea56cb0b81cf25e80727c7426adc6c56d8189ddR122

nicholaspai (approved):

bmzig (approved):

src/libexec/SpokePoolListenerExperimental.ts
pxrl: Yeah - and not just that. I've hit a lot of cases where I can't seem to get it to align with their standard types. The PublicClient type is another example. Apparently there are some arcane incantations that will help ts to interpret the types correctly, but I don't understand them.

src/libexec/SpokePoolListenerExperimental.ts
```diff
@@ -0,0 +1,262 @@
...
+ * Main entry point.
+ */
+async function run(argv: string[]): Promise<void> {
+  const minimistOpts = {
+    string: ["lookback", "relayer"],
+  };
+  const args = minimist(argv, minimistOpts);
+
+  ({ chainId } = args);
+  const { lookback, relayer = null, maxBlockRange = 10_000 } = args;
```
james-a-morris, line 168: Should we use `superstruct` here to get this behavior? It allows us to specify ranges and other features that may be helpful than `assert( )`

james-a-morris (commented):
one question - otherwise looks good

src/libexec/SpokePoolListenerExperimental.ts
pxrl: Yeah; that's a good suggestion. I think we also possibly want to lean on SuperStruct more to do validation elsewhere. I'd like to get this in now but I will circle back to add superstruct validation.

PR #1847, opened 2024-10-02
improve(relayer): Defer deposit version computation

pxrl:
getUnfilledDeposits() currently unconditionally computes the deposit version (i.e. the ConfigStore VERSION value applicable at the deposit quoteTimestamp), and then filters out all the deposits that have been filled. Determining the relevant version implies a lot of Array.find() calls, all of which is wasted when the object is subsequently discarded.


nicholaspai (approved):
LGTM

PR #1856, opened 2024-10-09
feat: Support World Chain USDC bridges

pxrl:


src/adapter/bridges/OpStackUSDCBridge.ts
```diff
@@ -0,0 +1,76 @@
...
+    l2chainId: number,
+    hubChainId: number,
+    l1Signer: Signer,
+    l2SignerOrProvider: Signer | Provider,
+    _l1Token: string
+  ) {
+    // Lint Appeasement
+    _l1Token;
+    const { address: l1Address, abi: l1Abi } = CONTRACT_ADDRESSES[hubChainId].opStackUSDCBridge;
+    const { address: l2Address, abi: l2Abi } = CONTRACT_ADDRESSES[l2chainId].opStackUSDCBridge;
```
pxrl, line 19: I think the names of these are both off vs. the definitions in CONTRACT_ADDRESSES.

src/adapter/bridges/OpStackUSDCBridge.ts
```diff
@@ -0,0 +1,76 @@
...
+    const events = await paginatedEventQuery(
+      l1Bridge,
+      l1Bridge.filters.MessageSent(undefined, toAddress),
+      eventConfig
+    );
+
+    console.log(`Got events: ${JSON.stringify(events, null, 2)}.`);
+
+    return {
+      [this.resolveL2TokenAddress(l1Token)]: events.map((event) => processEvent(event, "from", "to", "from")),
```
pxrl, line 55: This is wrong; I'd added the console.log above to dump the events and transform them correctly but didn't get to testing it.

src/adapter/bridges/OpStackUSDCBridge.ts
```diff
@@ -0,0 +1,76 @@
...
+  ): Promise<BridgeEvents> {
+    const l2Bridge = this.getL2Bridge();
+    const events = await paginatedEventQuery(
+      l2Bridge,
+      l2Bridge.filters.MessageReceived(undefined, toAddress),
+      eventConfig
+    );
+    console.log(`Got events: ${JSON.stringify(events, null, 2)}.`);
+    return {
+      [this.resolveL2TokenAddress(l1Token)]: events.map((event) => processEvent(event, "amount", "to", "from")),
```
pxrl, line 73: This is wrong; I'd added the console.log above to dump the events and transform them correctly but didn't get to testing it.

bmzig: Should be good now: https://github.com/across-protocol/relayer/pull/1856/commits/5601e3d95f222a7962e4679aead74ddba699d8f1

src/adapter/bridges/OpStackUSDCBridge.ts
bmzig: Should be good now: https://github.com/across-protocol/relayer/pull/1856/commits/5601e3d95f222a7962e4679aead74ddba699d8f1

src/adapter/bridges/OpStackUSDCBridge.ts
bmzig: I kept the names in CONTRACT_ADDRESSES in case there are other OpStack chains which will use this same bridge format. https://github.com/across-protocol/relayer/pull/1856/commits/5601e3d95f222a7962e4679aead74ddba699d8f1

bmzig:
Token tracking should now work for USDC, but we probably shouldn't merge this in until USDC is enabled as a route, since otherwise rebalancing will fail since the adapter will not approve the new bridge address (because of [this](https://github.com/across-protocol/relayer/blob/d794b4b369edf8ea65fbc342946ee933d8ca8d8e/src/clients/bridges/AdapterManager.ts#L276) check).

bmzig (approved):

nicholaspai (approved):

PR #1865, opened 2024-10-16
refactor: Confine oldestTime update to Dataworker

pxrl:
The Dataworker is the only SpokePoolClient user that needs to know the time at a historical SpokePool block, so permit it to resolve this on the fly and free up other users from having to query it.

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -97,18 +97,19 @@ export async function blockRangesAreInvalidForSpokeClients(
       const conservativeBundleFrequencySeconds = Number(
         process.env.CONSERVATIVE_BUNDLE_FREQUENCY_SECONDS ?? CONSERVATIVE_BUNDLE_FREQUENCY_SECONDS
       );
-      if (
-        spokePoolClient.eventSearchConfig.fromBlock > spokePoolClient.deploymentBlock &&
+      if (spokePoolClient.eventSearchConfig.fromBlock > spokePoolClient.deploymentBlock) {
         // @dev The maximum lookback window we need to evaluate expired deposits is the max fill deadline buffer,
         // which captures all deposits that newly expired, plus the bundle time (e.g. 1 hour) to account for the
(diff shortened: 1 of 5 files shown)


pxrl:
nb. Build will fail until https://github.com/across-protocol/sdk/pull/752 is merged.

src/dataworker/DataworkerUtils.ts
```diff
@@ -97,18 +97,19 @@ export async function blockRangesAreInvalidForSpokeClients(
...
         // which captures all deposits that newly expired, plus the bundle time (e.g. 1 hour) to account for the
         // maximum time it takes for a newly expired deposit to be included in a bundle. A conservative value for
         // this bundle time is 3 hours. This `conservativeBundleFrequencySeconds` buffer also ensures that all deposits
         // that are technically "expired", but have fills in the bundle, are also included. This can happen if a fill
         // is sent pretty late into the deposit's expiry period.
-        endBlockTimestamps[chainId] - spokePoolClient.getOldestTime() <
-          maxFillDeadlineBufferInBlockRange + conservativeBundleFrequencySeconds
-      ) {
-        return true;
+        const oldestTime = await spokePoolClient.getTimeAt(spokePoolClient.eventSearchConfig.fromBlock);
```
nicholaspai, line 107: Do you think we should cache this oldest time in the spoke pool clients memory?

nicholaspai (commented):
I like this just one question to maybe improve performance

src/dataworker/DataworkerUtils.ts
pxrl: Added here: https://github.com/across-protocol/sdk/pull/752/commits/0c502ea9a939140c016d6daeda177503c8ae3593

nicholaspai (dismissed):

nicholaspai:
When we merge this, we can then remove the `oldestTime` prop and `getOldestTime` functions from the SpokePoolClient

bmzig (approved):

PR #1866, opened 2024-10-16
feat: Automate withdrawing ETH from OpStack chains

nicholaspai:
- Adds a new script to withdraw ETH via the OVM standard L2 bridge to the signer's address on L1
- Adds a feature to the OP stack finalizer that lets the user specify addresses that they want to finalize OVM withdrawals for, in addition to any existing SpokePool withdrawals. This pairs with the above script to allow us to one-step manually move ETH from OP stack chains back to the same EOA on Ethereum
- To use this new feature, we need to add new chains to the finalizer config (i.e. 7777777 and 1135) and also set `WITHDRAWAL_TO_ADDRESSES=[x]` where `x` are addresses that we plan to execute manual withdrawals from


src/common/Constants.ts
```diff
@@ -297,6 +297,7 @@ export const multicall3Addresses = {
   [CHAIN_IDs.WORLD_CHAIN]: "0xcA11bde05977b3631167028862bE2a173976CA11",
   [CHAIN_IDs.ZK_SYNC]: "0xF9cda624FBC7e059355ce98a31693d299FACd963",
   [CHAIN_IDs.ZORA]: "0xcA11bde05977b3631167028862bE2a173976CA11",
+  [CHAIN_IDs.REDSTONE]: "0xcA11bde05977b3631167028862bE2a173976CA11",
```
pxrl, line 300: minor nit: the keys in this object are sorted alphabetically.

scripts/withdrawFromOpStack.ts
```diff
@@ -0,0 +1,104 @@
...
+// \ --keys bot1
+
+export async function run(): Promise<void> {
+  assert(
+    cliArgs.every((cliArg) => Object.keys(args).includes(cliArg)),
+    `Missing cliArg, expected: ${cliArgs}`
+  );
+  const baseSigner = await retrieveSignerFromCLIArgs();
+  const signerAddr = await baseSigner.getAddress();
+  const chainId = Number(args.chainId);
```
pxrl, line 41:
```suggestion
  const chainId = parseInt(args.chainId);
```

scripts/withdrawFromOpStack.ts
```diff
@@ -0,0 +1,104 @@
...
+    { value: amount }, // msg.value
+  ];
+
+  console.log(
+    `Submitting bridgeETHTo on the OVM standard bridge @ ${ovmStandardBridge.address} with the following args: `,
+    ...bridgeETHToArgs
+  );
+  if (!(await askYesNoQuestion("\nDo you want to proceed?"))) {
+    return;
+  }
```
pxrl, line 86: wdyt about adding a few safeguards before this? i.e. there are some known functions on the L2StandardBridge we can call in order to sanity check that it's the right contract - i.e. we could confirm that `MESSENGER()` matches what we have recorded, and `l1TokenBridge()` maps to the L1 bridge that we expect from our constants.

It might be redundant, but this script will be moving some decent volumes, so it could be fairly cheap assurance.

src/finalizer/utils/opStack.ts
```diff
@@ -94,6 +95,53 @@ export async function opStackFinalizer(
...
+  // and because they are lite chains, our only way to withdraw them is to initiate a slow bridge of ETH from the
+  // the lite chain to Ethereum.
+  const withdrawalToAddresses: string[] = process.env.WITHDRAWAL_TO_ADDRESSES
+    ? JSON.parse(process.env.WITHDRAWAL_TO_ADDRESSES).map((address) => ethers.utils.getAddress(address))
+    : [];
+  if (!CONTRACT_ADDRESSES[chainId].ovmStandardBridge) {
+    logger.warn({
+      at: "opStackFinalizer",
+      message: `No OVM standard bridge contract found for chain ${networkName} in CONTRACT_ADDRESSES`,
+    });
```
pxrl, line 109: Should we return here? it should permit the subsequent code to drop a level of indentation.

pxrl (commented):
This will definitely be useful.

src/finalizer/utils/opStack.ts
nicholaspai: i don't think so since we'd still want to check that withdrawalAddresses.length > 0 before proceeding into the next bracket and also, i don't think we need ovmStandardBridge in the existing finalizer use case

scripts/withdrawFromOpStack.ts
nicholaspai: https://github.com/across-protocol/relayer/pull/1866/commits/087bc8154b3865d59e40e4760461298c660dbc28

src/finalizer/utils/opStack.ts
pxrl: My bad; I missed that.

pxrl (approved):

PR #1867, opened 2024-10-18
feat: add machine-readable JSON report to reporter

mrice32:
Note: this is an experimental change.

The idea is to have the reporter export a JSON that can be read by external charting software.


james-a-morris (commented):
OOC is this possible to do w/out lodash?

mrice32:
> OOC is this possible to do w/out lodash?

I think so, just more code. I'm sure I could also just rewrite that lodash function if needed. Figured it was fine to use lodash since it's already used.

Any reason to avoid it?

pxrl (approved):

PR #1868, opened 2024-10-19
feat: add spoke pool balance reporting to the monitor

bmzig:
This adds three new environment variables (currently only used by the new function, but could theoretically be reused for the relayer reporter) to the monitor: a boolean `REPORT_SPOKE_POOL_BALANCES`, a list of chain IDs `MONITORED_SPOKE_POOL_CHAINS`, and a list of token symbols `MONITORED_TOKEN_SYMBOLS`. They are used to activate a new function which lives in the monitor `checkSpokePoolRunningBalances` which outputs a configured chain's spoke pools balance based on its current balance of a token, the amount of refunds it must payout to relayers on that chain, the pool rebalances of the latest root bundle in the hub pool, and the "effective" balance given by `currentBalance-pendingRefunds+poolRebalances`. 

This was done so that manual rebalancing can be done if necessary to decrease utilization in the hub pool.


src/monitor/Monitor.ts
```diff
@@ -596,6 +598,158 @@ export class Monitor {
...
+              account: spokePool,
+            },
+          ])
+        )[0];
+      }
+    }
+
+    // Here are the refunds which come from the last validated root bundle. We expect most of these to be zero since the refunds
+    // have been executed by the dataworker.
+    const validatedBundleRefunds: CombinedRefunds[] =
```
nicholaspai, line 664: One edge case that might make numbers occasionally weird is if an older relayer refund leaf failed to execute or was delayed. Same with slow fill leaves. Same if someone sent tokens directly to the contract.

nicholaspai (commented):
Can you re use any of the logic from validateRunningBalances? It's a script that prints out pretty similar logic FYI. Otherwise this looks great

src/monitor/Monitor.ts
```diff
@@ -596,6 +598,158 @@ export class Monitor {
     }
   }
 
+  async checkSpokePoolRunningBalances(): Promise<void> {
+    // We define a custom format function since we do not want the same precision that `convertFromWei` gives us.
+    const formatWei = (weiVal: string, decimals: number) => {
```
james-a-morris, line 603: The main difference between this fn and `createFormatFunction` looks to be the 0 case. Is there a reason we're treating this differently?

src/monitor/Monitor.ts
```diff
@@ -596,6 +598,158 @@ export class Monitor {
...
+      }
+      const formatFunction = createFormatFunction(1, 4, false, decimals);
+      return formatFunction(weiVal);
+    };
+
+    const hubPoolClient = this.clients.hubPoolClient;
+    const monitoredTokenSymbols = this.monitorConfig.monitoredTokenSymbols;
+    const chainIds =
+      this.monitorConfig.monitoredSpokePoolChains.length !== 0
+        ? this.monitorConfig.monitoredSpokePoolChains
```
james-a-morris, line 615: Will the ordering of this array affect the ordering of the final reported value? If so do we want to respect the `this.monitorChains` ordering and filter the chains rather than replace the original ordering?

src/monitor/Monitor.ts
```diff
@@ -596,6 +598,158 @@ export class Monitor {
...
+          await this._getBalances([
+            {
+              token: l2Token,
+              chainId: chainId,
+              account: spokePool,
+            },
+          ])
+        )[0];
+      }
+    }
```
james-a-morris, line 660: We should parallelize these calls as much as possible here. Any chance we can leverage the already parallelized `this._getBalances` fn or use `Promise.all( )` over this nested for loop?

src/monitor/Monitor.ts
```diff
@@ -596,6 +598,158 @@ export class Monitor {
...
+    // Here are the refunds which come from the last validated root bundle. We expect most of these to be zero since the refunds
+    // have been executed by the dataworker.
+    const validatedBundleRefunds: CombinedRefunds[] =
+      await this.clients.bundleDataClient.getPendingRefundsFromValidBundles();
+    // Here are the current outstanding refunds.
+    const nextBundleRefunds = await this.clients.bundleDataClient.getNextBundleRefunds();
+    for (const chainId of chainIds) {
+      const l2TokenAddresses = monitoredTokenSymbols
+        .map((symbol) => l2TokenForChain(chainId, symbol))
+        .filter(isDefined);
```
james-a-morris, line 671: We reference something like this above. Can we hoist these higher in the fn to prevent needing to recompute?

james-a-morris (commented):
> This was done so that manual rebalancing can be done if necessary to decrease utilization in the hub pool.

This change looks

src/monitor/Monitor.ts
bmzig: Kinda did something similar to how the validateRunningBalances script addresses this by defining a `BUNDLES_COUNT` environment variable which is the number of bundles to include in this query. https://github.com/across-protocol/relayer/pull/1868/commits/8872c3d19b67d35189fbaaf146954b1005373c91

src/monitor/Monitor.ts
bmzig: Korpi wanted the format to be like this (once decimal place). No other reason.

src/monitor/Monitor.ts
bmzig: I don't think it matters, but I went ahead and changed it to what you suggested anyways (it doesn't hurt to preserve order). https://github.com/across-protocol/relayer/pull/1868/commits/8872c3d19b67d35189fbaaf146954b1005373c91

src/monitor/Monitor.ts
bmzig: Yeah, that's a good call. I made it into a `mapAsync` which does a `Promise.all`.

src/monitor/Monitor.ts
bmzig: Now we only do this once for each chain https://github.com/across-protocol/relayer/pull/1868/commits/8872c3d19b67d35189fbaaf146954b1005373c91

src/monitor/Monitor.ts
```diff
@@ -596,6 +599,157 @@ export class Monitor {
...
 
+  async checkSpokePoolRunningBalances(): Promise<void> {
+    // We define a custom format function since we do not want the same precision that `convertFromWei` gives us.
+    const formatWei = (weiVal: string, decimals: number) => {
+      if (weiVal === "0") {
+        return "0";
+      }
+      const formatFunction = createFormatFunction(1, 4, false, decimals);
+      return formatFunction(weiVal);
+    };
```
james-a-morris, line 610: RE: https://github.com/across-protocol/relayer/pull/1868/files#r1809105128

Given what we're trying to do maybe we can condense this?

```suggestion
    const formatWei = (weiVal: string, decimals: number) =>
        weiVal === "0" ? "0" :
        createFormatFunction(1, 4, false, decimals)(weiVal);
```

james-a-morris (approved):
lgtm - will defer to Nick's outstanding comments. Otherwise the logic makes sense. Good work

src/monitor/Monitor.ts
```diff
@@ -596,6 +599,152 @@ export class Monitor {
...
+      const l2TokenMap = this.getL2ToL1TokenMap(enabledTokens, leaf.chainId);
+      pendingRebalanceRoots[leaf.chainId] = {};
+      Object.entries(l2TokenMap).forEach(([l2Token, l1Token]) => {
+        const rebalanceAmount = leaf.netSendAmounts[leaf.l1Tokens.findIndex((token) => token === l1Token.address)];
+        pendingRebalanceRoots[leaf.chainId][l2Token] = rebalanceAmount ?? bnZero;
+      });
+    }
+
+    // Take the validated bundles from the hub pool client.
+    const validatedBundles = sortEventsDescending(hubPoolClient.getValidatedRootBundles()).slice(
```
nicholaspai, line 639: how many validated bundles does this fetch? I would limit this to maybe up to 2, otherwise I imagine this will take very long to compute to find all pending refunds in all these bundles.

src/monitor/Monitor.ts
```diff
@@ -596,6 +599,152 @@ export class Monitor {
...
+          .filter(isDefined)
+          .reduce(
+            (totalPendingRefunds, refunds) =>
+              Object.values(refunds).reduce(
+                (totalBundleRefunds, bundleRefund) => totalBundleRefunds.add(bundleRefund),
+                bnZero
+              ),
+            bnZero
+          );
+        const nextBundleDeductions = nextBundleRefunds
```
nicholaspai, line 671: So you add the next bundle refunds (i.e. refunds not pending) plus any leftover refunds that aren't executed from the last N validated bundles, but what about any pending bundles that are in the challenge period? `getNextBundleRefunds` won't include these

nicholaspai:
You've seen this script right? https://github.com/across-protocol/relayer/blob/3de3d54d37934e4cc217dd968c8d5980b8ec2c84/src/scripts/validateRunningBalances.ts#L77 It does a very similar computation

src/monitor/Monitor.ts
bmzig: It does 4 by default

bmzig:
> You've seen this script right?
> 
> https://github.com/across-protocol/relayer/blob/3de3d54d37934e4cc217dd968c8d5980b8ec2c84/src/scripts/validateRunningBalances.ts#L77
> It does a very similar computation

Yes I have. I think this commit _is_ similar to the script. I took what I can, as well as stuff from the `reportRelayerBalances` part of the monitor.

src/monitor/Monitor.ts
bmzig: Is this true? If so, the comments here are incorrect https://github.com/across-protocol/sdk/blob/c0695f60098ce66580cbfc6119fbe3f484def125/src/clients/BundleDataClient/BundleDataClient.ts#L404. But I also see here https://github.com/across-protocol/sdk/blob/c0695f60098ce66580cbfc6119fbe3f484def125/src/clients/BundleDataClient/BundleDataClient.ts#L451 what looks to be it fetching refunds from the outstanding bundle?

nicholaspai: Ah ok the comments there are correct

nicholaspai (approved):

PR #1872, opened 2024-10-22
fix: resolve proper zora usdc address

bmzig:
This is to prevent resolving to an `undefined` when calling `getTranslatedTokenAddress` for USDC on Zora.


nicholaspai (approved):
Dang these are annoying. Good find

pxrl (approved):
@bmzig Good catch - ooc what were you doing when you spotted this?

bmzig:
> @bmzig Good catch - ooc what were you doing when you spotted this?

I saw the undefined in the logs of `getOutstandingCrossChainTransfers` while working on the monitor.

PR #1873, opened 2024-10-23
feat: support custom gas token bridges in the rebalancer

bmzig:
This should allow the rebalancer to approve a `BaseBridgeAdapter` if it has a custom gas token that is below a threshold. We reuse the same approval logic as when bridging tokens, meaning that we approve the bridge of the custom gas token when it is below the allowance threshold.


src/adapter/BaseChainAdapter.ts
```diff
@@ -126,8 +128,28 @@ export class BaseChainAdapter {
           });
           return { token: erc20, bridges: bridgesToApprove };
         }
-      )
-    ).filter(({ bridges }) => bridges.length > 0);
+      ),
+      await mapAsync(
```
james-a-morris, line 132: I don't think we need the `await` here or above. I think this will actually make the code synchronous because TS does assignment through the right side.

src/adapter/BaseChainAdapter.ts
```diff
@@ -126,8 +128,28 @@ export class BaseChainAdapter {
...
+            }
+            return !aboveAllowanceThreshold(allowance);
+          });
+          return { token: erc20, bridges: bridgesToApprove };
+        }
+      ),
+    ]);
+    const tokensToApprove = bridgeTokensToApprove
+      .concat(gasTokensToApprove)
+      .filter(({ bridges }) => bridges.length > 0);
```
james-a-morris, line 152: Could this be one array? Does it need to be nested?

src/common/Constants.ts
```diff
@@ -307,6 +307,7 @@ export const multicall3Addresses = {
   [CHAIN_IDs.WORLD_CHAIN]: "0xcA11bde05977b3631167028862bE2a173976CA11",
   [CHAIN_IDs.ZK_SYNC]: "0xF9cda624FBC7e059355ce98a31693d299FACd963",
   [CHAIN_IDs.ZORA]: "0xcA11bde05977b3631167028862bE2a173976CA11",
+  [CHAIN_IDs.ALEPH_ZERO]: "0x3CA11702f7c0F28e0b4e03C31F7492969862C569",
```
james-a-morris, line 310: I thought we use Multicall3 in the SDK

nicholaspai: two things, maybe not neccessarily to be addressed in this PR:

1. an entry already exists for `CHAIN_IDs.ALEPH_ZERO` on line 294
2. we should remove `multicall3Addresses` usage in this repo and use the SDK's `getMulticall3()` function

src/adapter/BaseChainAdapter.ts
bmzig: I don't understand. It is one array? Right?

src/adapter/BaseChainAdapter.ts
bmzig: Done here https://github.com/across-protocol/relayer/pull/1873/commits/2c4d2d56264e500f065bcee6ef50ae4dec32c17b

src/common/Constants.ts
bmzig: Yeah the multicall here was a mistake. It was removed https://github.com/across-protocol/relayer/pull/1873/commits/e2c945c5adf4ab2af73d789e527aec8a357ce828

src/adapter/BaseChainAdapter.ts
```diff
@@ -126,8 +128,30 @@ export class BaseChainAdapter {
...
+            }
+            return !aboveAllowanceThreshold(allowance);
+          });
+          return { token: erc20, bridges: bridgesToApprove };
+        }
+      ),
+    ]);
+    // Dedup the `gasTokensToApprove` array so that we don't approve the same bridge to send the same token multiple times.
+    const tokensToApprove = gasTokensToApprove
+      .filter(({ token, bridges }, idx) => gasTokensToApprove.indexOf({ token, bridges }) === idx)
```
james-a-morris, line 152: OOC is this indicative of a duplication that starts earlier in the code?

bmzig: Yes

james-a-morris: Should we account for it earlier?

src/adapter/bridges/ArbitrumOrbitBridge.ts
```diff
@@ -55,12 +58,18 @@ export class ArbitrumOrbitBridge extends BaseBridgeAdapter {
     l2Token: string,
     amount: BigNumber
   ): Promise<BridgeTransactionDetails> {
-    const { l1GatewayRouter, l2GasLimit, l2GasPrice, transactionSubmissionData, l1SubmitValue } = this;
+    const { l1GatewayRouter, l2GasLimit, l2GasPrice, l1SubmitValue } = this;
+    const transactionSubmissionData = isDefined(this.gasToken)
```
nicholaspai, line 62: Have you tested this works?

nicholaspai (approved):
Code looks fine but I'd suggest you run an integration test to verify the contract calls

james-a-morris (approved):
lgtm - a bit weary about the duplication issue appearing higher than where we deal with it

src/adapter/BaseChainAdapter.ts
bmzig: We definitely can, but I'm not sure if we want to. We get duplicates since the code above is run in parallel (specifically in a `mapAsync`), so if we didn't want duplicates, we would basically need to check each bridge in order (and for `n` bridges, that is theoretically an `n` times slow down). In practice, it doesn't matter, though, but another good thing about just filtering afterwards is the code becomes easier to read IMO.

src/adapter/bridges/ArbitrumOrbitBridge.ts
bmzig: I have. I took generating the calldata from the custom gas token adapter [here](https://github.com/across-protocol/contracts/blob/a522da068343925b4f4204790fb2e7c605f5872d/contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol#L209). We definitely don't want to send the gateway router any ETH.

src/adapter/BaseChainAdapter.ts
```diff
@@ -106,8 +106,10 @@ export class BaseChainAdapter {
 
   async checkTokenApprovals(l1Tokens: string[]): Promise<void> {
     const unavailableTokens: string[] = [];
-    const tokensToApprove = (
-      await mapAsync(
+    // Approve tokens to bridges. This includes the tokens we want to send over a bridge as well as the custom gas tokens
+    // each bridge supports (if applicable).
+    const [bridgeTokensToApprove, gasTokensToApprove] = await Promise.all([
```
pxrl, line 111: This statement is now quite dense and has a lot of indentation. Not urgent, but I think it makes sense to refactor it a bit to prioritise readability.

bmzig: Yeah. I think this also aligns with James's comment. The calculation for `bridgeTokensToApprove` and `gasTokensToApprove` is essentially identical, so we may just want to factor it out into a utils function and then remove duplicates. It's a good follow up candidate.

PR #1875, opened 2024-10-24
improve(relayer): Make HubPoolClient update conditional

pxrl:
This skips an RPC call when the block number has not incremented beyond the last update cycle.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -119,7 +119,9 @@ export class Relayer {
       tokenClient.clearTokenData();
 
       await configStoreClient.update();
-      await hubPoolClient.update();
+      if (configStoreClient.latestBlockSearched > hubPoolClient.latestBlockSearched) {
+        await hubPoolClient.update();
+      }
     }
 
     await updateSpokePoolClients(spokePoolClients, [


src/relayer/Relayer.ts
```diff
@@ -119,7 +119,9 @@ export class Relayer {
       tokenClient.clearTokenData();
 
       await configStoreClient.update();
-      await hubPoolClient.update();
+      if (configStoreClient.latestBlockSearched > hubPoolClient.latestBlockSearched) {
```
nicholaspai, line 122: Does this happen often?

nicholaspai (approved):

src/relayer/Relayer.ts
pxrl: The fast relayer can currently complete maximum 4 loops per mainnet block, given a POLLING_DELAY of 3 seconds. This implies that we're performing up to 6 redundant eth_getBlockByNumber() calls per mainnet block given current implementation, and we can cut that down to 3 redundant calls by making HubPoolClient update conditional on the ConfigStoreClient having updated.

This can be further optimised by subscribing to mainnet blocks instead of polling for them. I'm currently proposing that in the restructuring tech spec. But this current PR is a very light touch to eliminate a few redundant calls, so I think it makes sense to get it in ahead of the subsequent work.

james-a-morris (approved):

PR #1877, opened 2024-10-27
Update spokepool.ts

sitch:
Null coalesce (??) doesn't work as one would expect with `NaN`. So this fixes an issue when not supplying the `--exclusivityDeadline` directly

## Example

```typescript
$> NaN ?? "works"
NaN
```

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -224,7 +224,7 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
     exclusiveRelayer ?? depositQuote.exclusiveRelayer,
     depositQuote.quoteTimestamp,
     Number(currentTime) + Number(fillDeadlineBuffer),
-    exclusivityDeadline ?? depositQuote.exclusivityDeadline,
+    isNaN(exclusivityDeadline) ? depositQuote.exclusivityDeadline : exclusivityDeadline,
     message
   );
   const { hash: transactionHash } = deposit;


pxrl:
@sitch Good observation. It looks like this would also revert to `depositQuote.exclusivityDeadline` in case the user intended to specify an `exclusivityDeadline` but messed it up on the command line. So I think we probably want to do this up at line 180 instead. What do you think?

sitch:
@pxrl I think generally what you want is a more robust arg parser function for number than `Number`. Looks like `minimalist` doesn't offer this and I don't see an easy one liner for how to do this on line 180 that makes the code clearer

pxrl:
> @pxrl I think generally what you want is a more robust arg parser function for number than `Number`. Looks like `minimalist` doesn't offer this and I don't see an easy one liner for how to do this on line 180 that makes the code clearer

Here's the simplest change I can produce without reverting to a third-party package. We might do that at some point, but it's not really a priority for now. I realised also that there's a second issue in the script - if `exclusivityDeadline` is unspecified or 0 then the deposit still fails because it needs to use `depositV3()` in that case.

```
diff --git a/scripts/spokepool.ts b/scripts/spokepool.ts
index 4d4d1b05..b395acf7 100644
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -2,6 +2,7 @@ import axios, { isAxiosError } from "axios";
 import minimist from "minimist";
 import { config } from "dotenv";
+import { string, integer } from "superstruct";
 import { Contract, ethers, Signer } from "ethers";
 import { LogDescription } from "@ethersproject/abi";
 import { constants as sdkConsts, utils as sdkUtils } from "@across-protocol/sdk";
@@ -177,7 +178,15 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
   const [fromChainId, toChainId, baseAmount] = [args.from, args.to, args.amount].map(Number);
   const [recipient, message] = [args.recipient ?? depositor, args.message ?? sdkConsts.EMPTY_MESSAGE].map(String);
   const exclusiveRelayer = args.exclusiveRelayer;
-  const exclusivityDeadline = Number(args.exclusivityDeadline);
+
+  let exclusivityDeadline: number | undefined = undefined;
+  if (args.exclusivityDeadline) {
+    exclusivityDeadline = parseInt(String(args.exclusivityDeadline));
+    if (isNaN(exclusivityDeadline)) {
+      console.log(`Invalid exclusivityDeadline (${args.exclusivityDeadline}).`);
+      return false;
+    }
+  }
 
   if (!utils.validateChainIds([fromChainId, toChainId])) {
     console.log(`Invalid set of chain IDs (${fromChainId}, ${toChainId}).`);
@@ -213,7 +222,11 @@ async function deposit(args: Record<string, number | string>, signer: Signer): P
     spokePool.fillDeadlineBuffer(),
   ]);
 
-  const deposit = await spokePool.depositExclusive(
+  exclusivityDeadline ??= depositQuote.exclusivityDeadline;
+  const depositFn = exclusivityDeadline > 0
+    ? spokePool.depositExclusive
+    : spokePool.depositV3;
+  const deposit = await depositFn(
     depositor,
     recipient,
     token.address,
```

james-a-morris (commented):
@sitch - while this makes sense I think the suggestion by @pxrl provides more downstream advantages.

PR #1880, opened 2024-10-28
fix: error log stringify

dohaki:
Closes ACX-3032

We've been seeing occasional error logs like `error: "[object Object]"`. AFAICT these can occur if the caught value is not an instance of the `Error` class. We _might_ be able to get more information if we use `JSON.stringify` instead if the thrown value is not an error. 

Also, if we use `JSON.stringify` on an error instance it returns `{}` if it doesn't implement a custom `toJSON` method. 

To cover both cases above, this PR therefore introduces a `stringifyThrownValue` that can be used for logging.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,15 @@
 import minimist from "minimist";
-import { config, delay, exit, retrieveSignerFromCLIArgs, help, Logger, usage, waitForLogger } from "./src/utils";
+import {
+  config,
+  delay,
+  exit,
+  retrieveSignerFromCLIArgs,
+  help,
+  Logger,
+  usage,
+  waitForLogger,
+  stringifyThrownValue,
+} from "./src/utils";
 import { runRelayer } from "./src/relayer";
 import { runDataworker } from "./src/dataworker";
 import { runMonitor } from "./src/monitor";
@@ -61,12 +71,13 @@ if (require.main === module) {
   run(args)
     .catch(async (error) => {
       exitCode = 1;
+      const stringifiedError = stringifyThrownValue(error);
       logger.error({
         at: cmd ?? "unknown process",
         message: "There was an execution error!",
-        reason: error,
-        e: error,
-        error,
+        reason: stringifiedError,
+        e: stringifiedError,
+        error: stringifiedError,
         args,
         notificationPath: "across-error",
       });
--- a/index.ts
+++ b/index.ts
@@ -75,8 +75,6 @@ if (require.main === module) {
       logger.error({
         at: cmd ?? "unknown process",
         message: "There was an execution error!",
-        reason: stringifiedError,
-        e: stringifiedError,
         error: stringifiedError,
         args,
         notificationPath: "across-error",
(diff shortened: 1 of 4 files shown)


nicholaspai (commented):
Great idea. Have you tested it out locally to make sure the parsing works?

src/utils/LogUtils.ts
```diff
@@ -1 +1,15 @@
 export type DefaultLogLevels = "debug" | "info" | "warn" | "error";
+
+export function stringifyThrownValue(value: unknown): string {
+  if (value instanceof Error) {
+    const errToString = value.toString();
+    return value.stack || value.message || errToString !== "[object Object]"
+      ? errToString
```
james-a-morris, line 7: OOC is there any case where this value is too long? And, if so, should we condense it or is this handled downstream?

dohaki:
> Great idea. Have you tested it out locally to make sure the parsing works?

Yep, I tested cases like
```ts
throw new Error("Some error");
throw { key: "value" }
throw "value"
throw new CustomError()
```

src/utils/LogUtils.ts
dohaki: I think that should be part of the logger itself. And more specifically the respective transport class. I think for example the SlackTransport handles that currently

index.ts
```diff
@@ -61,12 +71,13 @@ if (require.main === module) {
...
+      const stringifiedError = stringifyThrownValue(error);
       logger.error({
         at: cmd ?? "unknown process",
         message: "There was an execution error!",
-        reason: error,
-        e: error,
-        error,
+        reason: stringifiedError,
+        e: stringifiedError,
+        error: stringifiedError,
```
pxrl, line 80: Given that these are all the same I think we should be fine to consolidate onto a single object :+1:

src/clients/TransactionClient.ts
```diff
@@ -100,7 +101,7 @@ export class TransactionClient {
           mrkdwn,
           // @dev `error` _sometimes_ doesn't decode correctly (especially on Polygon), so fish for the reason.
           errorMessage: isError(error) ? (error as Error).message : undefined,
-          error,
+          error: stringifyThrownValue(error),
```
pxrl, line 104: ooc, have you seen this get printed incorrectly?

src/utils/LogUtils.ts
pxrl: There are some issues when the logger receives a message that's too long - it tries to truncate it and is mostly successful, but sometimes slices JSON the wrong way and corrupts it, and then something throws.

I think it should be OK here because I'd imagine we're not dumping any JSON in the error.

index.ts
dohaki: yep that makes sense. will do so

src/clients/TransactionClient.ts
dohaki: yep a few time like below
```
at: "TransactionClient#submit"
bot-identifier: "across-v2-relayer-sweeper"
error: "[object Object]"
mrkdwn: ""
notificationPath: "across-error"
```

pxrl (approved):

PR #1887, opened 2024-11-01
feat: Support AlephZero

melisaguevara:
This PR:
- Bumps Across constants, contracts and sdk versions. (Waiting for final versions of contracts and sdk though).
- Adds a new class `AlephZeroBridge`, mainly based on `ArbitrumOneBridge`.
- Declares constants and contract addresses needed to support AlephZero.


src/common/Constants.ts
```diff
@@ -40,6 +41,7 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK = 4 * 60 * 60;
 
 // Target ~4 days per chain. Should cover all events needed to construct pending bundle.
 export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
+  [CHAIN_IDs.ALEPH_ZERO]: 86400, // 1 block every ~4 seconds
```
melisaguevara, line 44: @bmzig @pxrl For all of the time based constants I used a block time of 4 seconds. The Linear project says they average block time is 250ms but it doesn't seem to be the case based on the explorer: https://evm-explorer.alephzero.org/blocks

src/adapter/bridges/AlephZeroBridge.ts
```diff
@@ -0,0 +1,89 @@
...
+  toBN,
+  toWei,
+} from "../../utils";
+import { CONTRACT_ADDRESSES } from "../../common";
+import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./BaseBridgeAdapter";
+import { processEvent } from "../utils";
+import { PRODUCTION_NETWORKS, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
+
+export class AlephZeroBridge extends BaseBridgeAdapter {
+  public gasToken: string;
```
bmzig, line 17: I think we should move this to `BaseBridgeAdapter.ts`. It will be undefined for all the other bridges, but it might be nice to put it there so in the future when we support more custom gas bridges, we can just get this variable for free.

src/common/Constants.ts
pxrl: @melisaguevara I think for this you're probably better under-estimating the block time than over-estimating. I expect their block time to trend towards 250ms and if it does then we risk having less than 4 days of data.

package.json
```diff
@@ -10,9 +10,9 @@
     "node": ">=20"
   },
   "dependencies": {
-    "@across-protocol/constants": "^3.1.16",
-    "@across-protocol/contracts": "^3.0.11",
-    "@across-protocol/sdk": "^3.2.7",
+    "@across-protocol/constants": "^3.1.18",
+    "@across-protocol/contracts": "^3.0.12-beta.3",
```
pxrl, line 14: @melisaguevara Just ooc, which branch was this build produced against? I checked the contracts repo but I don't see the commits.

src/common/Constants.ts
pxrl: I also noticed that the aleph zero block explorer indicates that it's still indexing, so I expect its update rate might not be representative of the actual chain performance.

package.json
melisaguevara: This one was created from this branch: https://github.com/across-protocol/contracts/tree/james/acx-2763-deploy-contracts
Not sure why the commits are not shown but it basically has the changes in this PR: https://github.com/across-protocol/contracts/pull/704

src/common/Constants.ts
melisaguevara: Oh I get your point, yes. I changed some of the values which are now the same as Arbitrum here [9ca807c](https://github.com/across-protocol/relayer/pull/1887/commits/9ca807c9cff19d31e68d42def5f7a71628351f30).
I am still not sure about the ones in `MIN_DEPOSIT_CONFIRMATIONS` as Arbitrum has 0 for all the thresholds.

src/adapter/bridges/AlephZeroBridge.ts
melisaguevara: Moved here [abe4703](https://github.com/across-protocol/relayer/pull/1887/commits/abe4703ee6bc303189d77b15611667c28102657e)

package.json
pxrl: We have fwiw produces recent packages from `audit-latest`. In general I think that's preferable than producing packages against another branch, because for subsequent deployments we'd then need to merge those into James' branch (or branch off from there). It becomes quite tricky to stay on top of.

james-a-morris: @pxrl we were instructed to branch off https://github.com/across-protocol/contracts/tree/bz/azeroChanges due to some of the adapter work at the time not being included in `audit-latest`. Is this ready to be rebased?

src/common/Constants.ts
pxrl: I think it's OK to just copy the values for Arbitrum for `MIN_DEPOSIT_CONFIRMATIONS` :+1:

melisaguevara: Changed here [18b57c1](https://github.com/across-protocol/relayer/pull/1887/commits/18b57c12d8f23c80aed7ffd8f09a3363e7f78c38)

src/common/ContractAddresses.ts
```diff
@@ -131,6 +132,17 @@ export const CONTRACT_ADDRESSES: {
       address: "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef",
       abi: ARBITRUM_ERC20_GATEWAY_ROUTER_L1_ABI,
     },
+    arbitrumErc20Gateway: {
+      abi: ARBITRUM_ERC20_GATEWAY_L1_ABI,
+    },
+    alephZeroErc20GatewayRouter: {
```
nicholaspai, line 138: I think you might want to copy the pattern we use for opStack where you use something like `arbOrbit_<chainId>`

src/finalizer/utils/arbStack.ts
```diff
@@ -81,10 +79,10 @@ async function multicallArbitrumFinalizations(
   };
 }
 
-async function finalizeArbitrum(message: L2ToL1MessageWriter): Promise<Multicall2Call> {
-  const l2Provider = getCachedProvider(CHAIN_ID, true);
+async function finalizeArbitrum(message: L2ToL1MessageWriter, chainId: number): Promise<Multicall2Call> {
+  const l2Provider = getCachedProvider(chainId, true);
   const proof = await message.getOutboxProof(l2Provider);
-  const { address, abi } = CONTRACT_ADDRESSES[CHAIN_ID].outbox;
+  const { address, abi } = CONTRACT_ADDRESSES[chainId].outbox;
```
nicholaspai, line 85: tiny note: this `outbox` actually belongs under the Ethereum chain Id `1` in `CONTRACT_ADDRESSES`. For example you can see the outbox entry [here](https://github.com/across-protocol/relayer/blob/30b7003bac89168dd0849a419d696f8ca92e8edb/src/common/ContractAddresses.ts#L312) actually exists on Mainnet. This should be as simple as moving it under the correct chain id

src/finalizer/utils/arbStack.ts
```diff
@@ -163,12 +165,13 @@ async function getMessageOutboxStatusAndProof(
...
+  logIndex: number,
+  chainId: number
 ): Promise<{
   message: L2ToL1MessageWriter;
   status: string;
 }> {
-  const l2Provider = getCachedProvider(CHAIN_ID, true);
+  const l2Provider = getCachedProvider(chainId, true);
   const receipt = await l2Provider.getTransactionReceipt(event.transactionHash);
   const l2Receipt = new L2TransactionReceipt(receipt);
```
nicholaspai, line 176: This object is part of the arbitrum SDK, and you can see there are several such SDK functions imported [here](https://github.com/across-protocol/relayer/blob/30b7003bac89168dd0849a419d696f8ca92e8edb/src/finalizer/utils/arbitrum.ts#L1) which may not work for Aleph zero

src/common/ContractAddresses.ts
melisaguevara: Changed here: [a372bff](https://github.com/across-protocol/relayer/pull/1887/commits/a372bffef6c9cea4b21181f2e871cb7747b3231b)

src/finalizer/utils/arbStack.ts
melisaguevara: Moved here: [4105a03](https://github.com/across-protocol/relayer/pull/1887/commits/4105a0309a540a66c440f9146bfa94856149fa92)

src/common/ContractAddresses.ts
```diff
@@ -127,10 +128,25 @@ export const CONTRACT_ADDRESSES: {
       address: "0x8484Ef722627bf18ca5Ae6BcF031c23E6e922B30",
       abi: POLYGON_BRIDGE_ABI,
     },
+    arbitrumOutbox: {
```
nicholaspai, line 131: I think we should add add a chain ID to end of this name (i.e. arbitrumOutbox_42161) since we'll need one of these for aleph zero

src/common/ContractAddresses.ts
```diff
@@ -127,10 +128,25 @@ export const CONTRACT_ADDRESSES: {
...
+      address: "0x0B9857ae2D4A3DBe74ffE1d7DF045bb7F96E4840",
+      abi: ARBITRUM_OUTBOX_ABI,
+    },
     arbitrumErc20GatewayRouter: {
```
nicholaspai, line 135: Should this be orbitErc20GatewayRouter_42161?

src/common/ContractAddresses.ts
```diff
@@ -127,10 +128,25 @@ export const CONTRACT_ADDRESSES: {
...
       address: "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef",
       abi: ARBITRUM_ERC20_GATEWAY_ROUTER_L1_ABI,
     },
+    arbitrumErc20Gateway: {
```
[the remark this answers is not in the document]
nicholaspai: Also, should these be instead renamed orbitErc20GatewayRouter_41455?

src/adapter/bridges/ArbitrumOneBridge.ts
```diff
@@ -33,15 +33,17 @@ export class ArbitrumOneBridge extends BaseBridgeAdapter {
     l2SignerOrProvider: Signer | Provider,
     l1Token: string
   ) {
-    const { address: gatewayAddress, abi: l1Abi } = CONTRACT_ADDRESSES[hubChainId].arbitrumErc20GatewayRouter;
+    const { address: gatewayAddress, abi: gatewayRouterAbi } =
+      CONTRACT_ADDRESSES[hubChainId].arbitrumErc20GatewayRouter;
```
nicholaspai, line 37: i think this shoul dbe calling `orbitErc20GatewayRouter_${chainId}` right?

src/finalizer/utils/arbStack.ts
```diff
@@ -81,10 +80,10 @@ async function multicallArbitrumFinalizations(
   };
 }
 
-async function finalizeArbitrum(message: L2ToL1MessageWriter): Promise<Multicall2Call> {
-  const l2Provider = getCachedProvider(CHAIN_ID, true);
+async function finalizeArbitrum(message: L2ToL1MessageWriter, chainId: number): Promise<Multicall2Call> {
+  const l2Provider = getCachedProvider(chainId, true);
   const proof = await message.getOutboxProof(l2Provider);
-  const { address, abi } = CONTRACT_ADDRESSES[CHAIN_ID].outbox;
+  const { address, abi } = CONTRACT_ADDRESSES[CHAIN_IDs.MAINNET].arbitrumOutbox;
```
nicholaspai, line 86: I think for this to work it should be orbitOutbox_${chainId}

melisaguevara: I handled all the renames in this commit [27a7b8c](https://github.com/across-protocol/relayer/pull/1887/commits/27a7b8ca176b70ac2e20788280852d0c2726e4b9)
I used orbit for all so that it was consistent.

package.json
melisaguevara: This was now updated to be the latest release we published from master 👍🏼

src/common/ContractAddresses.ts
```diff
@@ -127,10 +128,25 @@ export const CONTRACT_ADDRESSES: {
...
-    arbitrumErc20GatewayRouter: {
+    orbitOutbox_42161: {
+      address: "0x0B9857ae2D4A3DBe74ffE1d7DF045bb7F96E4840",
+      abi: ARBITRUM_OUTBOX_ABI,
+    },
+    orbitErc20GatewayRouter_42161: {
       address: "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef",
       abi: ARBITRUM_ERC20_GATEWAY_ROUTER_L1_ABI,
     },
+    orbitErc20Gateway_42161: {
```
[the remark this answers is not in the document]
melisaguevara: It is defined inside the `ArbitrumOneBridge` file, [here](https://github.com/across-protocol/relayer/blob/mguevara/acx-2767-update-relayer-repository/src/adapter/bridges/ArbitrumOneBridge.ts#L38) so we are not accessing it from here

src/adapter/bridges/AlephZeroBridge.ts
james-a-morris, line 1: Can we extend from the ArbitrumBridge.ts class to cut down on this diff?

src/adapter/bridges/BaseBridgeAdapter.ts
```diff
@@ -28,6 +28,7 @@ export type BridgeEvents = { [l2Token: string]: BridgeEvent[] };
 export abstract class BaseBridgeAdapter {
   protected l1Bridge: Contract;
   protected l2Bridge: Contract;
+  public gasToken: string | undefined;
```
james-a-morris, line 31: Is this in scope of AlephZero?

src/clients/InventoryClient.ts
```diff
@@ -981,7 +981,11 @@ export class InventoryClient {
             const { unwrapWethThreshold, unwrapWethTarget } = tokenConfig;
 
             // Ignore chains where ETH isn't the native gas token. Returning null will result in these being filtered.
-            if (chainId === CHAIN_IDs.POLYGON || unwrapWethThreshold === undefined || unwrapWethTarget === undefined) {
+            if (
+              [CHAIN_IDs.ALEPH_ZERO, CHAIN_IDs.POLYGON].includes(chainId) ||
```
james-a-morris, line 985: Should we add a `chainHasNonNativeEthGas` in the SDK? I suspect this will be a subtle error we encounter if we ever try to run the bot on a polygon/alephzero testnet. Centralizing will also make things more convenient as we add addtl chains with non-native eth gas.

src/adapter/bridges/AlephZeroBridge.ts
melisaguevara: I agree that would be better but I'm not sure I will be able to get this done before Tuesday. Could we keep this for a  follow up PR?

src/adapter/bridges/BaseBridgeAdapter.ts
melisaguevara: Yes. @bmzig needed this for some work that would be based on this changes.

src/clients/InventoryClient.ts
melisaguevara: Well we already have this function `[getNativeTokenSymbol](https://github.com/across-protocol/sdk/blob/master/src/utils/NetworkUtils.ts#L23)`
I think we can use that.

melisaguevara: wdyt about this [f67425e](https://github.com/across-protocol/relayer/pull/1887/commits/f67425e10050acaf7914785108a6fe18509d1b71)?

src/common/ContractAddresses.ts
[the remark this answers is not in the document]
nicholaspai: hm ok, we can always add later

src/clients/InventoryClient.ts
nicholaspai: I think that's a good solution for now. I can see in the future its possible that a chain has a native token named "ETH" but it doens't behave like ETH, but let's just assume that won't happen...

src/common/ContractAddresses.ts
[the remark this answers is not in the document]
bmzig: For reference, Arbitrum has different gateways to approve for different tokens. I.e. there are custom WETH/DAI/USDT/ERC20 gateways, all with different addresses, so bridging a specific asset means we need to approve a different gatway. Aleph Zero seems to only have a single gateway for all assets supported, so we don't need to select which gateway contract to use based on the token to bridge.

pxrl (commented):
This all looks OK to me. I'd recommend manual testing of the inventory configuration to ensure that the rebalancing and transfer tracking works as expected.

@bmzig wdyt?

bmzig:
> This all looks OK to me. I'd recommend manual testing of the inventory configuration to ensure that the rebalancing and transfer tracking works as expected.
> 
> @bmzig wdyt?

Definitely. I think the rebalancer won't recognize the Aleph Zero bridge until pool rebalance routes are set, but once they are, would definitely be good to check.

A good thing is that the bridge is similar to Arbitrum, so I think it should be fine, assuming we can approve the bridge fine.

src/common/Constants.ts
```diff
@@ -40,6 +41,7 @@ export const MAX_RELAYER_DEPOSIT_LOOK_BACK = 4 * 60 * 60;
 
 // Target ~4 days per chain. Should cover all events needed to construct pending bundle.
 export const DATAWORKER_FAST_LOOKBACK: { [chainId: number]: number } = {
+  [CHAIN_IDs.ALEPH_ZERO]: 1382400, // Same as Arbitrum
```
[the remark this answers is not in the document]
melisaguevara: Based on the explorer it is slower. But Paul suggested that we use the same value as Arbitrum since the block time should tend to be the same as it. We discussed that [here](https://github.com/across-protocol/relayer/pull/1887#discussion_r1826203915)

bmzig (approved):

james-a-morris (approved):

src/common/Constants.ts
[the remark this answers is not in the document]
pxrl: I just searched on this and their FAQ says 1 second: https://docs.alephzero.org/aleph-zero/explore/glossary#block-time

This differs from the onboarding form, which specified 250ms variable. I'll see if we can get confirmation.

PR #1888, opened 2024-11-01
improve: track slow fills in the monitor

bmzig:
This PR modifies the spoke pool reporter to load data at the widest possible block ranges and checks for `bundleSlowFillsV3` objects. If there is a slow fill object, then it is treated as a pending refund. There is also a small refactor where all the async tasks to fetch balances/load bundle data are put in a single `Promise.all`.


src/monitor/Monitor.ts
```diff
@@ -637,14 +639,77 @@ export class Monitor {
...
       return TOKEN_SYMBOLS_MAP[symbol]?.addresses[chainId];
     };
-    const currentSpokeBalances = {};
     const pendingRelayerRefunds = {};
     const pendingRebalanceRoots = {};
 
-    // Get the pool rebalance leaves of the currently outstanding proposed root bundle.
-    const poolRebalanceRoot = await this.clients.bundleDataClient.getLatestPoolRebalanceRoot();
+    // Take the validated bundles from the hub pool client.
+    const validatedBundles = sortEventsDescending(hubPoolClient.getValidatedRootBundles()).slice(
```
nicholaspai, line 646: why do you `slice(0,)`? Isn't that redundant and it just returns the same result as `sortEventsDescending(hubPoolClient.getValidatedRootBundles())`?

nicholaspai (commented):
This makes sense

bmzig:
The added refactor also changes the way refunds are calculated. Basically, we get the pending refunds from the previous `n` validated bundles, then get the pending refunds from the current bundle, and then finally call `getApproximateRefundsForBlockRange` for the block range beginning with the end block of the current proposed bundle (or recently validated bundle if there is no currently proposed bundle) and ending with the latest searched block for each spoke pool client.

nicholaspai (approved):

PR #1889, opened 2024-11-01
chore: Bump SDK

pxrl:
Relayer performance should improve with an optimisation in the SDK provider layer.


pxrl:
Here's the diff between SDK variants: https://github.com/across-protocol/sdk/compare/v3.2.7...v3.2.11

tldr no major changes and nothing especially risky - only minor tweaks.

PR #1881, opened 2024-10-29
refactor: Use sdk multicall3 implementation

pxrl:
Needs testing.

PR #1893, opened 2024-11-04
feat(alephzero): add multicall3

james-a-morris:
Deployment PR: https://github.com/across-protocol/contracts/pull/721


pxrl (approved):
@james-a-morris This is also needed in the SDK here: https://github.com/across-protocol/sdk/blob/master/src/utils/Multicall.ts

I have a PR to drop these defs from the relayer and use the SDK implementation instead: https://github.com/across-protocol/relayer/pull/1881

james-a-morris:
> @james-a-morris This is also needed in the SDK here: https://github.com/across-protocol/sdk/blob/master/src/utils/Multicall.ts
> 
> I have a PR to drop these defs from the relayer and use the SDK implementation instead: #1881

Will merge for convenience to keep changes in one PR, but I've put up a secondary PR to account for this - thanks 🙏🏻 https://github.com/across-protocol/sdk/pull/764

PR #1894, opened 2024-11-05
improve(finalizer): Don't globally suppress unused variables

pxrl:


james-a-morris (approved):
Fair call

pxrl:
> We can add this to our TypeScript config - any thoughts on enforcing this on a compiler level and requiring unused vars to be opt-in?

It seems to already be enforced via eslint so I think we're OK:

```
/home/paul/dev/across-protocol/relayer-v2/src/finalizer/utils/cctp/l1ToL2.ts             
  4:72  error  'chainIdsToCctpDomains' is defined but never used  @typescript-eslint/no-unused-vars
```

PR #1895, opened 2024-11-06
feat(alephzero): make arb finalizer generic

james-a-morris:
The Arbitrum finalizer was using `CHAIN_IDs.ARBITRUM` in all locations.


nicholaspai (approved):
FYI AlephZero is a lite chain so there won't be any finalizations, but this is still a good refactor

james-a-morris:
> FYI AlephZero is a lite chain so there won't be any finalizations, but this is still a good refactor

Agreed - this refactor is in now for future arb chains + the event that one of those arb chains is no longer a lite chain.

PR #1896, opened 2024-11-07
Feat/performance tracking

gsteenkamp89:
closes ACX-3198

Goal:
Provide a standardised way of logging performance metrics. 
The API is similar to using Node's native perf_hooks, but slightly easier to use.

The Profiler will ensure `{ datadog: true }` is appended to all logs so we can ingest this by searching for this key.

Since the Profiler class holds some state, each instance should be created within a function scope so it can be garbage collected.

```ts
async function functionToMeasure( logger: Logger ) {
   const profiler = new Profiler({
      logger,
      at: "Relayer::functionToMeasure"
   })

  // drop first mark
  profiler.mark("A");

  // do stuff
  await step1();

  profiler.mark("B", { depositId: 123 });

  // do stuff
  await step2();

  profiler.mark("C", { relayerAddress: "0xabcabc1231231" });

  // Measure durations between marks
  profiler.measure("task1", {
    from: "A",
    to: "B",
    message: "Time from task A to task B",
  });

  profiler.measure("task2", {
    from: "B",
    to: "C",
    message: "Time from task B to task C",
  });
  profiler.measure("Total", {
    from: "A",
    to: "C",
    message: "Time from task A to task C",
  });
}
```

or wrap an async function with `measureAsync`

```ts
async function calculateFoo(...args: any): Promise<Foo> {
   return Foo
}

const foo = await profiler.measureAsync(calculateFoo(...args), "calculateFoo", {
   message: "Time taken to calculate Foo",
   ...args // log additional data
});
```


src/utils/LogUtils.ts
```diff
@@ -13,3 +17,152 @@ export function stringifyThrownValue(value: unknown): string {
...
+    return result;
+  }
+
+  /**
+   * Marks the start of a performance measurement for a given task.
+   * @param taskName The name of the task.
+   * @param detail Optional additional data related to the task.
+   */
+  start(taskName: string, detail?: unknown): void {
+    this.tasks.set(taskName, {
```
pxrl, line 111: Is there a need to protect against `taskName` collisions?

I was looking at `performance.mark()` and `performance.measure()` because it seems to do something quite similar to `profiler.start()` and `profiler.stop()`, but I think its weakness is that it doesn't protect against collisions and allows names to be overwritten.

What if instead of `profiler` keeping state, it just returned an opaque task object to the caller, and the caller had responsibility for passing it into their `stop()` call? That saves the caller from having to specify the name in two places.

pxrl (commented):
Great to see this - I left a few comments; interested to see what you think.

src/utils/LogUtils.ts
gsteenkamp89: My first shot at building this used the `PerformanceObserver` class and using `performance.mark` and `performance.measure` inside start and stop. then the observer would listen for these events and log them, but I did away with this because it introduced unnecessary complexity (and the observer listens asynchronously so there's race conditions that I ran into).

I really like this idea of returning an object with some methods attached to it. 
I've reworkd this to mimic node's api, but instead of using the observer we will just capture and log directly form the class instance.

We can use marks and measure so we can log multiple measurements within a funciton scope.
for example
```ts
  // create task profiler with uuid for this function scope

  const taskProfiler = profiler.create();
  // drop first mark
  taskProfiler.mark("A");

  // do stuff
  await step1();

  taskProfiler.mark("B", { depositId: 123 });

  // do stuff
  await step2();

  taskProfiler.mark("C", { relayerAddress: "0xabcabc1231231" });

  // Measure durations between marks
  taskProfiler.measure("task1", "A", "B", {
    message: "Time from task A to task B",
  });

  taskProfiler.measure("task2", "B", "C", {
    message: "Time from task B to task C",
  });
  taskProfiler.measure("Total", "A", "C", {
    message: "Time from task A to task C",
  });

```

Example log output 

```json
{
  "data": {
    "depositId": 123,
    "relayerAddress": "0xabcabc1231231"
  },
  "datadog": true,
  "duration": 1151.977958,
  "id": "86811339-b729-4893-b6b7-b9364f310ab8",
  "level": "debug",
  "message": "Time from task A to task B",
  "taskName": "task1",
  "timestamp": "2024-11-08T12:00:01.437Z"
}
{
  "data": {
    "depositId": 123,
    "relayerAddress": "0xabcabc1231231"
  },
  "datadog": true,
  "duration": 1141.4170000000001,
  "id": "86811339-b729-4893-b6b7-b9364f310ab8",
  "level": "debug",
  "message": "Time from task B to task C",
  "taskName": "task2",
  "timestamp": "2024-11-08T12:00:01.443Z"
}
{
  "data": {
    "depositId": 123,
    "relayerAddress": "0xabcabc1231231"
  },
  "datadog": true,
  "duration": 2293.394958,
  "id": "86811339-b729-4893-b6b7-b9364f310ab8",
  "level": "debug",
  "message": "Time from task A to task C",
  "taskName": "Total",
  "timestamp": "2024-11-08T12:00:01.444Z"
}
```

src/utils/LogUtils.ts
```diff
@@ -13,3 +17,157 @@ export function stringifyThrownValue(value: unknown): string {
...
+  message: string;
+  data?: Record<string, unknown>; // Cumulative detail data
+};
+
+type ProfilerOptions = {
+  logger: Logger;
+  at?: string;
+} & Record<string, unknown>;
+
+export class Profiler {
```
nicholaspai, line 41: Should this have its own file? Am wary of this  LogUtils file getting too big as we add functionality

nicholaspai (commented):
where is the logic for this?

> The Profiler will ensure { datadog: true } is appended to all logs

This looks like a great approach, I like the class' interface a lot

src/utils/LogUtils.ts
gsteenkamp89: [here](https://github.com/across-protocol/relayer/blob/c16459b3b5667ccb21d3e5668f5b1d3eac1ad989/src/utils/LogUtils.ts#L51)

gsteenkamp89:
> where is the logic for this?
> 
> > The Profiler will ensure { datadog: true } is appended to all logs
> 
> This looks like a great approach, I like the class' interface a lot

the default metadata `{ datadog: true }` is being set [here](https://github.com/across-protocol/relayer/blob/c16459b3b5667ccb21d3e5668f5b1d3eac1ad989/src/utils/LogUtils.ts#L51)

src/utils/LogUtils.ts
gsteenkamp89: yes totally agree 👍 
since the purpose of this is to create some sort of standard for our performance logging, it might actually be useful to migrate this over to the sdk at some point. wdyt?

gsteenkamp89: for now, will move into its own file 👍

pxrl: @gsteenkamp89 @alexandrumatei36 also noted that they'd like to use this in the indexer, so it probably makes sense to relocate it to the SDK repository. wdyt?

src/utils/LogUtils.ts
```diff
@@ -13,3 +17,157 @@ export function stringifyThrownValue(value: unknown): string {
...
 }
+
+type Detail = {
+  message?: string;
+  [key: string]: unknown;
+};
+
+const defaultMeta = { datadog: true };
+
+export type PerformanceData = {
```
pxrl, line 28: fyi, it's a requirement for each log message to include the `at` field - otherwise the logger gets very unhappy and complains about malformed messages. The caller should be passing this in when they instantiate the Profiler, or when they call `measure()` / `measureAsync()`.

src/relayer/index.ts
```diff
@@ -56,8 +61,7 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       if (loop) {
         logger.debug({ at: "relayer#run", message: `Starting relayer execution loop ${run}.` });
       }
-
-      const tLoopStart = performance.now();
+      const tLoopStart = taskProfiler.mark("tLoopStart");
```
pxrl, line 64: Just toying around with some ideas, but since we typically only measure one thing from start to finish, I'm wondering if there's a simplified version that we could support here. Something like:

```
  constructor(...) {
    this.profiler = new Profiler({ name: <className>, logger });
  }

  async randomFn(...) {
    const x = profiler.start("randomFn");
    // Do something...
    profiler.stop(x, detail); // Standardised message gets emitted internally by the profiler.stop() call. 
  }
```

Internally within the `start()` call, it'd be important that the return object identified the start time and the name. The name could then be concatenated with the class name to be used in the `at` field (i.e. `{ at: `${this.name}::${handle.name}`.

Does that make sense?

src/utils/LogUtils.ts
gsteenkamp89: Oh really, okay I'll change the constructor args to require this then

src/utils/LogUtils.ts
```diff
@@ -13,3 +17,157 @@ export function stringifyThrownValue(value: unknown): string {
...
+export class Profiler {
+  id: string;
+  private marks: Map<string, { time: number; detail?: Record<string, unknown> }>;
+  private detail: Detail;
+  private logger: Logger;
+
+  constructor({ logger, ...detail }: ProfilerOptions) {
+    this.id = crypto.randomUUID();
+    this.marks = new Map();
+    this.detail = detail ?? {};
```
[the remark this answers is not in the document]
gsteenkamp89: Like you pointed our before, we need to require the caller to include the `at` field, so we'll have to keep it I think

src/utils/LogUtils.ts
```diff
@@ -13,3 +17,152 @@ export function stringifyThrownValue(value: unknown): string {
...
+   * Marks the end of a performance measurement for a given task and logs the measurement.
+   * @param taskName The name of the task.
+   * @param detail Optional additional data related to the task.
+   */
+  stop(taskName: string, detail?: unknown): void {
+    const task = this.tasks.get(taskName);
+    if (!task) {
+      this.logger.warn(`No start time found for task "${taskName}". Did you forget to call start()?`, detail);
+      return;
+    }
```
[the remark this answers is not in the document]
pxrl: I was thinking that we'd just want to bail and take the bot down, but given that this is only for performance tracking, I've reconsidered and think it does make sense with a warning :+1: 

The logger call does need to follow the standard format though:
```
{
  at: ...
  message: ...
  <other kv pairs>
}
```

src/utils/LogUtils.ts
amateima: ➕

src/utils/LogUtils.ts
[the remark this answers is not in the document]
gsteenkamp89: Okay perfect, so we'll keep the warning then, just format the log correctly 👍 
I think fwiw, your suggestion of the simplified start/stop api will completely remove this possibility from happening. And I expect we'll use that most of the time.

src/utils/Profiler.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Initializes a new instance of the Profiler class.
+   * @param logger The logger instance for logging performance data.
+   * @param at A string identifier for where the profiler is used.
+   * @param detail Optional additional detail data.
+   */
+  constructor({ logger, at, ...detail }: ProfilerOptions) {
+    this.marks = new Map();
+    this.at = at;
+    this.detail = detail;
+    this.logger = logger.child(defaultMeta);
```
pxrl, line 41: Do we need to instantiate a child instance here? The logger instance that we inherit is fwiw extended in the upstream UMA package and I noticed this warning in the docs:

```
.child is likely to be bugged if you're also extending the Logger class, due to some implementation details that make this keyword to point to unexpected things. Use with caution.
```

pxrl (commented):
@gsteenkamp89 LGTM. Pending that one question on the logger, are you OK to re-propose `Profiler` in the SDK? Then we can keep this PR open and use it to update/align all the uses of the profiler in the relayer.

src/utils/Profiler.ts
gsteenkamp89: Oh good catch on the possible .child being buggy.. Not a problem though, we can get around this pretty easily since we just need to ensure default metadata is always included in the log.
And yes happy to re-propose in the sdk, with changes 👍

gsteenkamp89:
> Is there any overlap here? [across-protocol/indexer#95](https://github.com/across-protocol/indexer/pull/95)

Oh jeez.. Yes there's definitely overlap. Although looking at David's PR, it seems the goal there as slightly different maybe? This profiler is just logging production performance data to datadog. 
The benchmarker is generating reports on demand or on a schedule which is maybe more useful in a CI job as a part of testing our code. 
🤔

yarn.lock
pxrl, line 1: @gsteenkamp89 Any idea why there are so many changes here? We didn't actually change any dependencies between SDK 3.2.14 and 3.2.15 so I'd normally expect that the only difference here should be the SDK itself. Is our yarn.lock on master out of sync?

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants": "^3.1.19",
     "@across-protocol/contracts": "^3.0.16",
-    "@across-protocol/sdk": "^3.2.13",
+    "@across-protocol/sdk": "^3.2.14",
```
pxrl, line 15: We were a version behind in the relayer; we need to go to 3.2.15 instead.

```suggestion
    "@across-protocol/sdk": "^3.2.15",
```

src/utils/index.ts
```diff
@@ -1,12 +1,13 @@
...
-import { constants as sdkConstants } from "@across-protocol/sdk";
+import { constants as sdkConstants, utils as sdkUtils } from "@across-protocol/sdk";
 import { constants as ethersConstants } from "ethers";
 
 import winston from "winston";
 import assert from "assert";
 export { winston, assert };
 
 export const { MAX_SAFE_ALLOWANCE } = sdkConstants;
+export const { Profiler } = sdkUtils;
```
pxrl, line 10: It's not 100% consistent, but we typically re-export SDK functions from `src/utils/SDKUtils.ts`.

src/clients/InventoryClient.ts
```diff
@@ -298,13 +299,17 @@ export class InventoryClient {
...
   async getBundleRefunds(l1Token: string): Promise<{ [chainId: string]: BigNumber }> {
     let refundsToConsider: CombinedRefunds[] = [];
-
+    const taskProfiler = new Profiler({
+      logger: this.logger,
+      at: "InventoryClient::getBundleRefunds",
+    });
+    taskProfiler.mark("A", {
+      l1Token,
+    });
```
pxrl, line 308: @gsteenkamp89 OK to try the start/stop pattern here? I'm hoping it should be pretty clean :crossed_fingers:

src/clients/InventoryClient.ts
```diff
@@ -298,13 +299,17 @@ export class InventoryClient {
...
   async getBundleRefunds(l1Token: string): Promise<{ [chainId: string]: BigNumber }> {
     let refundsToConsider: CombinedRefunds[] = [];
-
+    const taskProfiler = new Profiler({
+      logger: this.logger,
+      at: "InventoryClient::getBundleRefunds",
+    });
+    const A = taskProfiler.start("A", {
+      l1Token,
+    });
```
pxrl, line 308: Does it work if we instantiate the profiler as a class variable in the constructor, and then change the code here as follows?

```suggestion
    const profiler = this.taskProfiler.start("getBundleRefunds", {
      l1Token,
    });
```

src/clients/InventoryClient.ts
```diff
@@ -298,13 +299,17 @@ export class InventoryClient {
...
     // Increase virtual balance by pending relayer refunds from the latest valid bundle and the
     // upcoming bundle. We can assume that all refunds from the second latest valid bundle have already
     // been executed.
-    let startTimer: number;
     if (!isDefined(this.bundleRefundsPromise)) {
-      startTimer = performance.now();
```
pxrl, line 307: @gsteenkamp89 Seems like we should only be starting the profiler here btw.

gsteenkamp89: In that case we'll only be measuring the first time this function runs I think? 
According to the comment on line 311, we store the promise the first time to allow multiple parallel calls to skip this step, meaning we won't be marking those.

pxrl: Yeah; I think that's OK since this is the slow path that we want to monitor, and for now we just want to match the existing behaviour :+1:

src/relayer/index.ts
```diff
@@ -100,11 +101,11 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       if (!loop) {
         stop = true;
       } else {
-        const runTime = Math.round((performance.now() - tLoopStart) / 1000);
-        logger.debug({
-          at: "Relayer#run",
-          message: `Completed relayer execution loop ${run} in ${runTime} seconds.`,
+        const runTimeMilliseconds = tLoopStart.stop({
+          message: `Completed relayer execution loop ${run} times.`,
```
pxrl, line 105:
```suggestion
          message: `Completed relayer execution loop.`,
```

pxrl (approved):

PR #1897, opened 2024-11-07
feat: handle Polygon finalizer edge case when CCTP limit is breached

nicholaspai:
There is a bug in the Polygon finalizer where a single transaction contains multiple token withdrawals from Spoke to Hub in which more than 1 million USDC are withdrawn, resulting in a single `TokensBridged` event getting emitted but multiple USDC "burns" happening (due to the CCTP 1mil withdrawal limit).

This messes up the `logIndex` computation math for any withdrawals that take place after the USDC withdrawals. This logic isn't needed in other chains with CCTP like Arbitrum or Optimism because the `logIndex` is relative to all other non-CCTP withdrawals, whereas in Polygon it seems that CCTP withdrawals also count towards the `logIndex` of non CCTP withdrawals

### DIFF
--- a/src/finalizer/utils/polygon.ts
+++ b/src/finalizer/utils/polygon.ts
@@ -16,6 +16,8 @@ import {
   compareAddressesSimple,
   TOKEN_SYMBOLS_MAP,
   CHAIN_IDs,
+  sortEventsAscending,
+  toBNWei,
 } from "../../utils";
 import { EthersError, TokensBridged } from "../../interfaces";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
@@ -35,6 +37,8 @@ enum POLYGON_MESSAGE_STATUS {
 // Unique signature used to identify Polygon L2 transactions that were erc20 withdrawals from the Polygon
 // canonical bridge. Do not change.
 const BURN_SIG = "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
+
+const CCTP_WITHDRAWAL_LIMIT_WEI = toBNWei(1_000_000, 6);
 
 export interface PolygonTokensBridged extends TokensBridged {
   payload: string;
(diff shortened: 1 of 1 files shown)


src/finalizer/utils/polygon.ts
```diff
@@ -36,6 +38,8 @@ enum POLYGON_MESSAGE_STATUS {
 // canonical bridge. Do not change.
 const BURN_SIG = "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
 
+const CCTP_WITHDRAWAL_LIMIT_WEI = toBNWei(1_000_000, 6);
```
james-a-morris, line 41: I think this is a dynamic variable in CCTP's contracts. Would we want to dynamically resolve this in a follow up PR?

james-a-morris (commented):
One comment one not

src/finalizer/utils/polygon.ts
```diff
@@ -61,7 +65,26 @@ export async function polygonFinalizer(
...
+  recentTokensBridgedEvents.forEach((e) => {
+    if (
+      compareAddressesSimple(e.l2TokenAddress, TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_ID]) &&
+      e.amountToReturn.gt(CCTP_WITHDRAWAL_LIMIT_WEI)
+    ) {
+      // Inject one TokensBridged event for each CCTP withdrawal that needs to be processed.
+      const numberOfEventsToAdd = Math.ceil(e.amountToReturn.div(CCTP_WITHDRAWAL_LIMIT_WEI).toNumber());
+      for (let i = 0; i < numberOfEventsToAdd; i++) {
+        recentTokensBridgedEvents.push({
+          ...e,
```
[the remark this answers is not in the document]
nicholaspai: probably not but how else should we "duplicate" the event? I figured this was the least verbose way to do it

src/finalizer/utils/polygon.ts
nicholaspai: definitely, i'll add a comment

james-a-morris (approved):

PR #1902, opened 2024-11-11
improve(relayer): Pass SpokePool addresses to listener processes

pxrl:
This skips an RPC call per SpokePool listener in the fast relayer configuration, and permits each SpokePool listener instance to communicate with only a single chain, instead of having to always query mainnet to resolve the current SpokePool address. In case the caller does not supply the SpokePool address, the listener process will fall back to the address specified in the contracts repository.


nicholaspai (commented):
Just one question: https://github.com/across-protocol/relayer/pull/1902/files#r1836825620

src/libexec/SpokePoolListenerExperimental.ts
```diff
@@ -206,7 +207,7 @@ async function run(argv: string[]): Promise<void> {
   };
 
   logger.debug({ at: "RelayerSpokePoolListener::run", message: `Starting ${chain} SpokePool Indexer.`, opts });
-  const spokePool = await utils.getSpokePoolContract(chainId);
+  const spokePool = getDeployedContract("SpokePool", chainId);
```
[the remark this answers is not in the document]
pxrl: Yeah; it does remove that ability. My thinking was that it's only happened once, and with a huge amount of coordination and preparation, so it seems pretty unlikely that we'd sleepwalk into this issue.

There might be a better compromise; the relayer can alternatively supply the SpokePool address via the runtime arguments. I think this is doable (and we could subsequently multicall it from the relayer).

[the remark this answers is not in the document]
pxrl: Updated:
- https://github.com/across-protocol/relayer/pull/1902/commits/7e91ac099ac6276ca672c091bd06ca3bc6c8222b
- https://github.com/across-protocol/relayer/pull/1902/commits/ea073dfa46ad6b7f085a03cf142f0bb62d59017f

This permits the main relayer process to instruct the listener processes on what the respective SpokePool addresses are. Because the main relayer inherits this from event log scraping in the HubPoolClient, this permits the listener process to dynamically migrate to a new address without the need for any RPC query.

If the parent does not supply the address, the listener process will fall back on whatever is configured in the contracts repository.

wdyt?

nicholaspai (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -78,9 +78,10 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
    */
   protected startWorker(): void {
     const {
-      eventSearchConfig: { fromBlock, maxBlockLookBack: blockRange },
+      eventSearchConfig: { fromBlock, maxBlockLookBack: blockrange },
```
james-a-morris, line 81:
```suggestion
      eventSearchConfig: { fromBlock, maxBlockLookBack: blockRange },
```

src/utils/ContractUtils.ts
```diff
@@ -1,32 +1,43 @@
...
-  }
-  return `${networkName}_SpokePool`;
+  return `${networkName.replace(" ", "")}_SpokePool`;
+}
+
+// For a chain ID and optional SpokePool address, return a Contract instance with the corresponding ABI.
+export function getSpokePool(chainId: number, address?: string): Contract {
+  const factoryName = castSpokePoolName(chainId);
+  const artifact = typechain[`${factoryName}__factory`];
+  return new Contract(address ?? getDeployedAddress("SpokePool", chainId), artifact.abi);
```
james-a-morris, line 40: Would we not not want to use the latest spoke pool address for the fallback?

src/clients/SpokePoolClient.ts
pxrl: The intention is for the command-line arguments to be passed as all lower-case, so `blockrange` was actually intended here :+1:

src/utils/ContractUtils.ts
pxrl: Hm, I don't understand the question - could you clarify?

Also just to confirm the intention here, if the SpokePool address was supplied at runtime then that address is used. Per this change, this will always be latest SpokePool as resolved by the main relayer process via its SetCrossChainContracts event indexing. However, if it's not supplied at runtime then `getDeployedAddress()` will resolve to the latest SpokePool as inherited from the contracts repository. This should also match 1:1, since we don't change SpokePool addresses in practice.

The alternative is to query the HubPool contract directly, but the motivation for this change in the first place was to avoid querying the hub chain, so ideally we can avoid that.

james-a-morris (approved):
resolved my questions - lgtm

PR #1903, opened 2024-11-11
refactor: arbitrum orbit bridge

bmzig:
The Aleph Zero bridge is the same as the Arbitrum One bridge up to construction. This performs the refactor suggested [here](https://github.com/across-protocol/relayer/pull/1887#discussion_r1835099890)

Edit: The scope changed to instead just make the Arbitrum bridge work for all Arbitrum Orbit chains, similar to what we do for OpStack chains. Since there can be multiple gateways for tokens, we refactor the contract addresses for the gateways to `src/common/Constants`, and then fetch the relevant addresses based on the l2's chain ID.


src/adapter/bridges/AlephZeroBridge.ts
james-a-morris, line 1: Is this still needed? In the OpStack bridges we don't define a separate bridge class per chain.

bmzig: Huh that's a good point. I think we can get away with getting a single bridge if we do some refactoring. Will probably try to di this instead.

bmzig: Alright. Should be good now. https://github.com/across-protocol/relayer/pull/1903/commits/692290dd1b396fb477dbe6121622a21bf15a4101 https://github.com/across-protocol/relayer/pull/1903/commits/5a2e51509ee2b1c1b4d8ff2d376ec810041514ce

james-a-morris (approved):

melisaguevara (approved):
lgtm
thanks for working on this!

PR #1906, opened 2024-11-13
across: Update token balances unconditionally on token shortfalls

pxrl:
The relayer currently only updates its token balances every x seconds (nominally 60). This is a nice optimisation for the standard case, but results in unnecessary marginal delays in the case of proactive rebalancing or awaiting on repayments. Instead, modify the update logic to be conditional on token shortfalls OR at least x seconds (nominally 60).

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -148,8 +148,8 @@ export class Relayer {
     const { inventoryClient, tokenClient } = this.clients;
 
     const currentTime = getCurrentTime();
-    const update = tokenClient.anyCapturedShortFallFills() ||
-      currentTime < this.lastMaintenance + this.config.maintenanceInterval;
+    const update =
+      tokenClient.anyCapturedShortFallFills() || currentTime < this.lastMaintenance + this.config.maintenanceInterval;
 
     if (!update) {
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -148,7 +148,10 @@ export class Relayer {
     const { inventoryClient, tokenClient } = this.clients;
 
     const currentTime = getCurrentTime();
-    if (currentTime < this.lastMaintenance + this.config.maintenanceInterval) {
+    const update =
+      tokenClient.anyCapturedShortFallFills() || currentTime < this.lastMaintenance + this.config.maintenanceInterval;
```
nicholaspai, line 152: Why not just remove the LHS and always update if it's been long enough since an update? Would it result in too many rpc calls?

nicholaspai (approved):
One comment and then lgtm

src/relayer/Relayer.ts
pxrl: The LHS is added by this PR. I think it's advantageous to have it because it'll make the relayer more responsive in cases where it has a shortfall. Ideally I'd like to continue to avoid updating unconditionally on every loop, though.

PR #1898, opened 2024-11-07
WIP: support manual finalizations for AlephZero

nicholaspai:
Refactors arbitrum finalizer logic into being generalized for arb-orbit

Similar to https://github.com/across-protocol/relayer/pull/1866 but for AlephZero which is an Arbitrum Orbit chain
- Adds a manual withdrawal script using canonical Arbitrum Orbit bridge
- Adds automated finalization logic

PR #1909, opened 2024-11-14
feat(AlephZero): Add manual withdrawal script and automate finalizations

nicholaspai:
Similar to #1866 for OpStack but for Arbitrum Orbit.

Replaces original PR #1898 with smaller commit history. I essentially took the core changes and copied and pasted here, to avoid rebase hell

### DIFF
--- a/scripts/withdrawFromArbitrumOrbit.ts
+++ b/scripts/withdrawFromArbitrumOrbit.ts
@@ -60,10 +60,7 @@ export async function run(): Promise<void> {
     )} ${l2Token}`
   );
   console.log(
-    `Current native ${nativeTokenSymbol} token balance for account ${signerAddr}: ${fromWei(
-      currentNativeBalance,
-      l1TokenInfo.decimals
-    )}`
+    `Current native ${nativeTokenSymbol} token balance for account ${signerAddr}: ${fromWei(currentNativeBalance, 18)}`
   );
 
   // Now, submit a withdrawal:
--- a/scripts/withdrawFromArbitrumOrbit.ts
+++ b/scripts/withdrawFromArbitrumOrbit.ts
@@ -13,13 +13,14 @@ import {
   fromWei,
   blockExplorerLink,
   CHAIN_IDs,
+  getNativeTokenSymbol,
 } from "../src/utils";
 import { CONTRACT_ADDRESSES } from "../src/common";
 import { askYesNoQuestion } from "./utils";
 
 import minimist from "minimist";
 
-const cliArgs = ["amount", "chainId"];
+const cliArgs = ["amount", "chainId", "token"];
 const args = minimist(process.argv.slice(2), {
   string: cliArgs,
 });
@@ -28,6 +29,7 @@ const args = minimist(process.argv.slice(2), {
 // ts-node ./scripts/withdrawFromArbitrumOrbit.ts
 // \ --amount 3000000000000000000
 // \ --chainId 41455
+// \ --token WETH
 // \ --wallet gckms
 // \ --keys bot1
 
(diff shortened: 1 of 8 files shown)


james-a-morris (commented):
Did we not have this previously for arb?

nicholaspai:
> Did we not have this previously for arb?

We did not. Arbitrum orbit has never had a lite chain in Across, yet, so we've never had to manually withdraw from Arbitrum

scripts/withdrawFromArbitrumOrbit.ts
```diff
@@ -0,0 +1,117 @@
...
+  );
+  const baseSigner = await retrieveSignerFromCLIArgs();
+  const signerAddr = await baseSigner.getAddress();
+  const chainId = parseInt(args.chainId);
+  const connectedSigner = baseSigner.connect(await getProvider(chainId));
+  const l2Token = TOKEN_SYMBOLS_MAP[args.token]?.addresses[chainId];
+  assert(l2Token, `${args.token} not found on chain ${chainId} in TOKEN_SYMBOLS_MAP`);
+  const l1TokenInfo = getL1TokenInfo(l2Token, chainId);
+  console.log("Fetched L1 token info:", l1TokenInfo);
+  const amount = args.amount;
```
james-a-morris, line 48: Did we want to sanitize the input to ensure that `args.amount` is positive integer?

scripts/withdrawFromArbitrumOrbit.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // Now, submit a withdrawal:
+  let contract: Contract, functionName: string, functionArgs: any[];
+  if (l1TokenInfo.symbol !== nativeTokenSymbol) {
+    const arbErc20GatewayObj = CONTRACT_ADDRESSES[chainId].erc20Gateway;
+    contract = new Contract(arbErc20GatewayObj.address, arbErc20GatewayObj.abi, connectedSigner);
+    functionName = "outboundTransfer";
+    functionArgs = [
+      l1TokenInfo.address, // l1Token
+      signerAddr, // to
```
james-a-morris, line 74: Did we want to allow the user to specify the recipient or is this not needed?

scripts/withdrawFromArbitrumOrbit.ts
```diff
@@ -0,0 +1,117 @@
...
+    return;
+  }
+  const withdrawal = await contract[functionName](...functionArgs);
+  console.log(`Submitted withdrawal: ${blockExplorerLink(withdrawal.hash, chainId)}.`);
+  const receipt = await withdrawal.wait();
+  console.log("Receipt", receipt);
+}
+
+if (require.main === module) {
+  run()
```
james-a-morris, line 107: Is it easier to use this rather than `await`

james-a-morris (approved):
mostly nits + potential prods into how we'd like to use the script. Otherwise LGTM

melisaguevara (approved):

scripts/withdrawFromArbitrumOrbit.ts
nicholaspai: nah, you can't use a negative number at the contract level since its a `uint`

scripts/withdrawFromArbitrumOrbit.ts
nicholaspai: i don't think its neccessary, we could add it in the future

scripts/withdrawFromArbitrumOrbit.ts
nicholaspai: I'm just copying this outro from other scripts like `withdrawFromOpStack.ts`

PR #1910, opened 2024-11-14
feat: Manual withdrawals of ERC20 on OpStack lite chains

nicholaspai:
Updates script `withdrawFromOpStack` to support ERC20 withdrawals (e.g. USDT) and updates finalizer to automate these. Helps with withdrawing ERC20's from Lite chains

### DIFF
--- a/src/common/abi/OpStackStandardBridgeL2.json
+++ b/src/common/abi/OpStackStandardBridgeL2.json
@@ -15,27 +15,16 @@
   {
     "anonymous": false,
     "inputs": [
-      { "indexed": true, "internalType": "address", "name": "from", "type": "address" },
-      { "indexed": true, "internalType": "address", "name": "to", "type": "address" },
-      { "indexed": false, "internalType": "uint256", "name": "amount", "type": "uint256" },
-      { "indexed": false, "internalType": "bytes", "name": "extraData", "type": "bytes" }
-    ],
-    "name": "ETHBridgeInitiated",
-    "type": "event"
-  },
-  {
-    "anonymous": false,
-    "inputs": [
       {
         "indexed": true,
         "internalType": "address",
-        "name": "localToken",
+        "name": "l1Token",
         "type": "address"
       },
       {
         "indexed": true,
         "internalType": "address",
-        "name": "remoteToken",
+        "name": "l2Token",
         "type": "address"
       },
       {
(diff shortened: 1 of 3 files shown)


src/common/abi/OpStackStandardBridgeL2.json
```diff
@@ -15,12 +15,44 @@
...
+      },
+      {
+        "indexed": false,
+        "internalType": "bytes",
+        "name": "extraData",
+        "type": "bytes"
+      }
     ],
-    "name": "ETHBridgeInitiated",
+    "name": "WithdrawalInitiated",
```
pxrl, line 55: Do you know when `WithdrawalInitiated` was added to the Optimism bridge contracts?

iirc these contracts have some annoyingly-similar event and function names. We first noticed this on Blast and it popped up again throughout Bennet's work on the new bridge adapter structure. It seems like they had an initial set of names and then added a new set of names that were almost identical, but retained the old ones for backwards compat.

src/finalizer/utils/opStack.ts
```diff
@@ -114,17 +114,36 @@ export async function opStackFinalizer(
...
-      {
-        ...spokePoolClient.eventSearchConfig,
-        toBlock: spokePoolClient.latestBlockSearched,
+    const withdrawalEvents = (
+      await paginatedEventQuery(
+        ovmStandardBridge,
+        ovmStandardBridge.filters.WithdrawalInitiated(
+          null, // l1Token
+          null, // l2Token
+          withdrawalToAddresses // from
```
pxrl, line 123: Is it valid to assume that the EOA initiating the withdrawal is always the same as the mainnet receiver?

pxrl (approved):
Just a couple of minor points that might be actionable, but no showstoppers.

src/common/abi/OpStackStandardBridgeL2.json
nicholaspai: oof good call, i think WithdrawalInitiated is the legacy event, which is annoying because having two different events requires a branch here

src/finalizer/utils/opStack.ts
nicholaspai: yeah, if they're using the `withdrawFromOpStack` script it is true

src/common/abi/OpStackStandardBridgeL2.json
nicholaspai: I've reverted the commit https://github.com/across-protocol/relayer/pull/1910/commits/4070a62c5841ec27d34a2ec07e5742b9ae48f918 so we use the newer events going forwards

PR #1911, opened 2024-11-15
improve(relayer): Better support for custom Across API host

pxrl:
Permit the ACROSS_API_HOST env var to override the standard app.across.to and testnet.across.to API hostnames.

While here, sub out some hardcoded chain IDs.


melisaguevara (approved):
lgtm!

james-a-morris (approved):

PR #1912, opened 2024-11-15
chore: bump sdk

melisaguevara:


bmzig (approved):
Do you know why there's a large diff in `yarn.lock`? Weird that that would happen on just a sdk bump.

pxrl:
> Do you know why there's a large diff in `yarn.lock`? Weird that that would happen on just a sdk bump.

viem was added to the SDK inbetween. We're using a slightly different version in the relayer.

pxrl (approved):

PR #1916, opened 2024-11-19
improve(arweave): Change arweave data key from list of chain ID's to map

nicholaspai:
We've received feedback from a user of the arweave data that its easier if the arweave stored `bundleBlockRanges` is a mapping of chainId to bundle block ranges rather than a list. This PR also shortens the `tag` used to uniquely identify bundle data to just a mainnet block associated with the bundle, which should always be unique. This avoids the 2048 tag length limit which we'd eventually run into.

example txn: https://arweave.app/tx/jBk5WzugtYMSa1jnXxEH7XN3Q6n0R_jxYsO6WkDHMc4

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -442,6 +442,8 @@ export const RELAYER_DEFAULT_SPOKEPOOL_INDEXER = "./dist/src/libexec/RelayerSpok
 
 export const DEFAULT_ARWEAVE_GATEWAY = { url: "arweave.net", port: 443, protocol: "https" };
 
+export const ARWEAVE_TAG_BYTE_LIMIT = 2048;
+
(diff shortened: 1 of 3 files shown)


james-a-morris (approved):

src/common/Constants.ts
```diff
@@ -442,6 +442,8 @@ export const RELAYER_DEFAULT_SPOKEPOOL_INDEXER = "./dist/src/libexec/RelayerSpok
 
 export const DEFAULT_ARWEAVE_GATEWAY = { url: "arweave.net", port: 443, protocol: "https" };
 
+export const ARWEAVE_TAG_BYTE_LIMIT = 2048;
```
james-a-morris, line 445: Is this a protocol imposed limit? I.e. is it possible that this could be changed or is this truly a constant

PR #1928, opened 2024-11-25
improve: wrap cctp axios request in a try/catch

bmzig:
There is a chance for the CCTP attestation API to fall behind on ready attestations, causing the axios request to return a 404 error. This wraps the call in a try/catch block so that the finalizer will not exit early in this event.

### DIFF
--- a/src/utils/CCTPUtils.ts
+++ b/src/utils/CCTPUtils.ts
@@ -237,10 +237,15 @@ async function _resolveCCTPRelatedTxns(
           if (processed) {
             status = "finalized";
           } else {
-            // Generate the attestation proof for the message. This is required to finalize the message.
-            attestation = await generateCCTPAttestationProof(messageHash, utils.chainIsProd(destinationChainId));
-            // If the attestation proof is pending, we can't finalize the message yet.
-            status = attestation.status === "pending_confirmations" ? "pending" : "ready";
+            // Try to generate the attestation proof for the message. This is required to finalize the message.
+            // If the request fails, which may happen if Circle's attestation API is falling behind, ignore the finalization on this run.
+            try {
+              attestation = await generateCCTPAttestationProof(messageHash, utils.chainIsProd(destinationChainId));
+              // If the attestation proof is pending, we can't finalize the message yet.
+              status = attestation.status === "pending_confirmations" ? "pending" : "ready";
+            } catch (e) {
+              return undefined;
+            }
           }
 
           return {


james-a-morris (approved):
Will this then get resolved on the next iteration?

src/utils/CCTPUtils.ts
```diff
@@ -237,10 +237,15 @@ async function _resolveCCTPRelatedTxns(
...
-            // If the attestation proof is pending, we can't finalize the message yet.
-            status = attestation.status === "pending_confirmations" ? "pending" : "ready";
+            // Try to generate the attestation proof for the message. This is required to finalize the message.
+            // If the request fails, which may happen if Circle's attestation API is falling behind, ignore the finalization on this run.
+            try {
+              attestation = await generateCCTPAttestationProof(messageHash, utils.chainIsProd(destinationChainId));
+              // If the attestation proof is pending, we can't finalize the message yet.
+              status = attestation.status === "pending_confirmations" ? "pending" : "ready";
+            } catch (e) {
+              return undefined;
```
pxrl, line 247: I think we'd want to at least log something here because otherwise the failure is silently suppressed, so we might have a long period of issues trying to finalize before we actually notice it.

I note that we don't have a logger instance here and it's probably annoying to supply one, so I'm wondering if we should just catch this at the top level where we anyway need to catch and report on errors. wdyt?

bmzig: If we throw an error here, then wouldn't this interrupt `resolveCCTPRelatedTxns` and basically cause all CCTP finalizations for the chain to be ignored? So for example, if we have 3 ready to finalize transactions, one of which returns a 404 from an API request, then wouldn't propagating that error up a level effectively cut off this function from returning the other two ready to finalize transactions? 

To be fair, I think this is possible with some refactoring, so then the question is, if we want to do this approach, should we just close this PR and make a new one which does this type of refactoring.

bmzig:
> Will this then get resolved on the next iteration?

I theory, yes. These 404 errors generally only last one finalizer cycle since by the next run, the API is updated (which would mean that the `try` portion does not throw an error).

src/utils/CCTPUtils.ts
bmzig: I added a top level try/catch here https://github.com/across-protocol/relayer/pull/1928/commits/8ac78ddf3b29b084d289ae8b8fa524f97cf9ac91. This should mean that if the finalizers fail sometime in the future for whatever reason, we can catch it here and still continue. I didn't get rid of the try/catch in CCTP since I still think the point stands that if we just throw here, then we will not finalize any more cross-chain transfers. I.e. if we have 5 ready-to-finalize transfers, but one of them errors on the axios request, then all five would be ignored if we just threw an error.

src/finalizer/index.ts
```diff
@@ -227,21 +227,29 @@ export async function finalize(
...
-      totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+        totalWithdrawalsForChain += crossChainMessages.filter(({ type }) => type === "withdrawal").length;
+        totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
+        totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+      } catch (_e) {
+        logger.error({
+          at: "finalizer",
+          message: "Something errored in a chain-specific finalizer",
+          errorMsg: _e,
+        });
```
pxrl, line 251: Is it going to be possible to identify the chain that errored here? It's probably a good idea to dereference `client.chainId` so we're guaranteed to know which chain needs further investigation.

pxrl: We might also want to add a `notification-path` here. `logger.error()` should produce an alert on its own though. @nicholaspai Any thoughts?

src/utils/CCTPUtils.ts
bmzig: On second thought, it is best to fail loudly than fail silently. So this was removed https://github.com/across-protocol/relayer/pull/1928/commits/7b40c5173ebd2dec2f34b1011b4bd1b15d2a7b8e

src/finalizer/index.ts
bmzig: Good idea. We specify the chain id now: https://github.com/across-protocol/relayer/pull/1928/commits/7b40c5173ebd2dec2f34b1011b4bd1b15d2a7b8e

src/utils/CCTPUtils.ts
james-a-morris: We should avoid silently throwing errors. A simple debug log would help us detect failures over time. Failing silently risks missing CCTP finalization if the endpoint goes down without any GCP debug logs.

james-a-morris (commented):
One comment. I don't believe this is an error that requires us to be paged.

src/finalizer/index.ts
```diff
@@ -227,21 +227,29 @@ export async function finalize(
...
-      totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+        totalWithdrawalsForChain += crossChainMessages.filter(({ type }) => type === "withdrawal").length;
+        totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
+        totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+      } catch (_e) {
+        logger.error({
+          at: "finalizer",
+          message: `Something errored in a finalizer for chain ${client.chainId}`,
+          errorMsg: _e,
+        });
```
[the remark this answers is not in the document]
bmzig: Good point, especially since the error we made this PR for is a transient error. https://github.com/across-protocol/relayer/pull/1928/commits/55abaaf15b1f85008e6839ecf399d54ac6f91865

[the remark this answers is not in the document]
pxrl: My concern with not alerting on this is that there's a non-negligible chance of missing an actual issue. The current Linea finalizer issue is one example - if we weren't paging for that then we probably wouldn't know there was an issue.

Given that we currently page, my preference is to retain that behaviour (as annoying as it is; we need the pain until we address the actual issues).

pxrl:
> There is a chance for the CCTP attestation API to fall behind on ready attestations, causing the axios request to return a 404 error. This wraps the call in a try/catch block so that the finalizer will not exit early in this event.

@bmzig The description probably needs an update after the PR was re-worked a bit.

src/finalizer/index.ts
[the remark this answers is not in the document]
pxrl: @bmzig @james-a-morris Just flagging this explicitly because I think it's an important point.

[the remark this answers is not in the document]
bmzig: I see both sides. Theoretically, if we were to put this in a channel which isn't noisy, then we probably would be able to get away without paging. This PR was originally created because of a transient error for which there is no actionable thing to do (except to wait for Circle's API to update). But I also personally do not mind if it pages. https://github.com/across-protocol/relayer/pull/1928/commits/e957be7978b9b482847d63f993982c6d4c4e0c17

src/finalizer/index.ts
```diff
@@ -227,21 +227,29 @@ export async function finalize(
...
-      totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
-      totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+        totalWithdrawalsForChain += crossChainMessages.filter(({ type }) => type === "withdrawal").length;
+        totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
+        totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+      } catch (_e) {
+        logger.error({
+          at: "finalizer",
+          message: `Something errored in a finalizer for chain ${client.chainId}`,
+          errorMsg: _e,
```
pxrl, line 250: Not sure if `_e` will be able to be displayed here, since it should be an Error object. You might instead want to use this utility: https://github.com/across-protocol/relayer/blob/8f6617ed123a530f2827305d25a314b87912efc6/src/utils/LogUtils.ts#L3

bmzig: Good point. Fixed here https://github.com/across-protocol/relayer/pull/1928/commits/d481254cf94464d3fbf6fc07004ef45b60754f69

src/finalizer/index.ts
```diff
@@ -227,21 +228,29 @@ export async function finalize(
...
-      totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
-      totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+        totalWithdrawalsForChain += crossChainMessages.filter(({ type }) => type === "withdrawal").length;
+        totalDepositsForChain += crossChainMessages.filter(({ type }) => type === "deposit").length;
+        totalMiscTxnsForChain += crossChainMessages.filter(({ type }) => type === "misc").length;
+      } catch (_e) {
+        logger.error({
+          at: "finalizer",
+          message: `Something errored in a finalizer for chain ${client.chainId}`,
+          errorMsg: `${stringifyThrownValue(_e)}`,
```
pxrl, line 251:
```suggestion
          errorMsg: stringifyThrownValue(_e),
```

pxrl (approved):

PR #1931, opened 2024-11-26
improve(LineaFinalizer): Force RPC calls through custom provider

nicholaspai:
The functions that we use in the Linea SDK make event queries from block 0 to "latest" by default which means that some of our RPC's can't be used with the Finalizer. Additionally, we're missing out our custom provider retry and caching logic.

This PR removes all instances of making RPC calls via the Linea SDK's provider and replace with making the same call through our custom Provider.

The latest [Linea SDK](https://github.com/Consensys/linea-monorepo/blob/3fbe660683a318b6fa1b63ec518f948791536352/sdk/src/sdk/LineaSDK.ts#L56) code looks like it'll support injecting custom providers but its not released yet on NPM. So this code should be removed eventually once a later SDK version is released.

This PR also fixes a bug where the existing SDK's interface doesn't match what's on chain: 12e872e840bb1137daf30c4b126887a1c2c0aa97


src/finalizer/utils/linea/common.ts
```diff
@@ -96,6 +108,80 @@ export function makeGetMessagesWithStatusByTxHash(
...
+}
+export function getL2MessageServiceContractFromL1ClaimingService(
+  l1ClaimingService: L1ClaimingService,
+  l2Provider: ethers.providers.Provider
+): Contract {
+  return new Contract(
+    l1ClaimingService.l2Contract.contract.address,
+    l1ClaimingService.l2Contract.contract.interface,
+    l2Provider
+  );
```
pxrl, line 145: Could this alternatively work?

```suggestion
  return l1ClaimingService.l2Contract.contract.connect(l2Provider);
```

src/finalizer/utils/linea/common.ts
```diff
@@ -96,6 +108,80 @@ export function makeGetMessagesWithStatusByTxHash(
...
+export function getL1MessageServiceContractFromL1ClaimingService(
+  l1ClaimingService: L1ClaimingService,
+  l1Provider: ethers.providers.Provider
+): Contract {
+  return new Contract(
+    l1ClaimingService.l1Contract.contract.address,
+    l1ClaimingService.l1Contract.contract.interface,
+    l1Provider
+  );
+}
```
pxrl, line 156: Same as above.

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -4,17 +4,43 @@ import { Contract } from "ethers";
...
   findMessageSentEvents,
   getBlockRangeByHoursOffsets,
+  getL1MessageServiceContractFromL1ClaimingService,
   initLineaSdk,
 } from "./common";
 
+// Normally we avoid importing directly from a node_modules' /dist package but we need access to some
+// of the internal classes and functions in order to replicate SDK logic so that we can by pass hardcoded
+// ethers.Provider instances and use our own custom provider instead.
+import { L2MessageServiceContract } from "@consensys/linea-sdk/dist/lib/contracts";
```
pxrl, line 22: wdyt about exporting these from a new file, so we can contain it and only have to document it there? i.e. `src/finalizer/utils/linea/import.ts` or something.

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -4,17 +4,43 @@ import { Contract } from "ethers";
...
+
+// Temporary re-implementation of the SDK's `L2MessageServiceContract.getMessageStatus` functions that allow us to use
+// our custom provider, with retry and caching logic, to get around the SDK's hardcoded logic to query events
+// from 0 to "latest" which will not work on all RPC's.
+async function getL1ToL2MessageStatusUsingCustomProvider(
+  messageService: L2MessageServiceContract,
+  messageHash: string,
+  l2Provider: ethers.providers.Provider
+): Promise<OnChainMessageStatus> {
+  const l2Contract = new Contract(messageService.contract.address, messageService.contract.interface, l2Provider);
```
pxrl, line 32: Also wondering if you can just connect the provider to the existing contract, but not sure whether it might be some weird type.

```suggestion
  const l2Contract = messageService.contract.connect(l2Provider);
```

pxrl (approved):
No showstoppers for me here. I do worry a bit about the long-term maintainability of this though. Feels like we need to upstream some of this feedback.

bmzig (approved):
This looks good to me. Have you been able to finalize a L2 -> L1 message?

nicholaspai:
> This looks good to me. Have you been able to finalize a L2 -> L1 message?

I haven't been able to time it well. I think one way to do it is i can temporarily pause the production finalizer for linea and run this locally.

src/finalizer/utils/arbStack.ts
```diff
@@ -209,6 +210,7 @@ export async function arbStackFinalizer(
           logger.debug({
             at: `Finalizer#${networkName}Finalizer`,
             message: `Withdrawal event for ${amountFromWei} of ${l1TokenInfo.symbol} is too recent to finalize`,
+            timeUntilFinalization: `${((event.blockNumber - latestBlockToFinalize) * l2BlockTime) / 60 / 60}`,
```
james-a-morris, line 213: Did we want to floor this? Could be good for readability to at least `toFixed()` this output to keep the logs clean

src/finalizer/utils/linea/common.ts
```diff
@@ -16,11 +13,18 @@ import {
...
 import { HubPoolClient } from "../../../clients";
 import { CONTRACT_ADDRESSES } from "../../../common";
+import { Log } from "../../../interfaces";
+
+// Normally we avoid importing directly from a node_modules' /dist package but we need access to some
+// of the internal classes and functions in order to replicate SDK logic so that we can by pass hardcoded
+// ethers.Provider instances and use our own custom provider instead.
+import { L1MessageServiceContract, L2MessageServiceContract } from "@consensys/linea-sdk/dist/lib/contracts";
+import { L1ClaimingService } from "@consensys/linea-sdk/dist/lib/sdk/claiming/L1ClaimingService";
+import { MessageSentEvent } from "@consensys/linea-sdk/dist/typechain/L2MessageService";
```
james-a-morris, line 27: Is this the only location we need these imports? Could we make a patchfile to have these vars be exported?

james-a-morris (approved):
This makes sense but I think we should avoid accessing the Linea package in an unintended way. We've forked/patched packages before so that they play nicely with our imports - this could be a viable workaround.

Otherwise there's nothing that stands out - I'll approve pending my comments

src/finalizer/utils/arbStack.ts
nicholaspai: cf90eae01f2bc6c8a08f07e7e8f24a2d86db795e

src/finalizer/utils/linea/l1ToL2.ts
nicholaspai: 5a2ad1e0048dc00ba96daa520490ce407b977ef0

src/finalizer/utils/linea/common.ts
nicholaspai: 5a2ad1e0048dc00ba96daa520490ce407b977ef0

src/finalizer/utils/linea/common.ts
nicholaspai: 5a2ad1e0048dc00ba96daa520490ce407b977ef0

src/finalizer/utils/linea/l1ToL2.ts
nicholaspai: 953c2b4bad4d034c6457b531707ea094259cf121

src/finalizer/utils/linea/common.ts
nicholaspai: thoughts on 953c2b4bad4d034c6457b531707ea094259cf121

src/finalizer/utils/arbStack.ts
```diff
@@ -110,6 +110,7 @@ export async function arbStackFinalizer(
     undefined,
     redis
   );
+  const l2BlockTime = (await averageBlockTime(spokePoolClient.spokePool.provider)).average;
```
james-a-morris, line 113: If this fails are we ok rewinding the stack?

src/finalizer/utils/linea/l2ToL1.ts
```diff
@@ -3,15 +3,157 @@ import { Wallet } from "ethers";
...
+  if (!receipt || receipt.logs.length === 0) {
+    throw new Error(`Transaction does not exist or no logs found in this transaction: ${transactionHash}.`);
+  }
+  let treeDepth = 0;
+  const l2MerkleRoots = [];
+  const blocksNumber = [];
+  const filteredLogs = receipt.logs.filter((log) => log.address === l1Contract.contractAddress);
+  for (let i = 0; i < filteredLogs.length; i++) {
+    const log = filteredLogs[i];
+    // This part changes from the SDK: remove any logs with topic hashes that don't exist in the current SDK's ABI,
```
nicholaspai, line 112: Important bug fix here, without which we'd throw in `parseLog` if the event topic isn't one that we expect (i.e. the `l1Contract.contract.interface` doesn't contain the event)

nicholaspai: As an example, this event isn't contained in the current SDK's version: [etherscan link](https://etherscan.io/tx/0xc6b2c65a7b69276e191d11802d5cda28d8feb4736afd7d47695b9cd9ea3ceb81#eventlog)

nor is it in the latest version 0.3.0

src/finalizer/utils/linea/common.ts
```diff
@@ -96,6 +102,72 @@ export function makeGetMessagesWithStatusByTxHash(
   };
 }
 
+// Temporary re-implementation of the SDK's `L1ClaimingService.getMessageStatus` functions that allow us to use
```
james-a-morris, line 105: Is this a copy or did we change it outside of transposing to not use a class?

nicholaspai:
> This looks good to me. Have you been able to finalize a L2 -> L1 message?

Yes, now I have: https://etherscan.io/tx/0xdd512394eb0aed5b44ae4c9ed6619937555d8d67d7f8795e4c82f060cdbcce94

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -4,16 +4,38 @@ import { Contract } from "ethers";
...
   findMessageFromTokenBridge,
   findMessageFromUsdcBridge,
   findMessageSentEvents,
   getBlockRangeByHoursOffsets,
+  getL1MessageServiceContractFromL1ClaimingService,
   initLineaSdk,
 } from "./common";
+import { L2MessageServiceContract } from "./imports";
+
+// Temporary re-implementation of the SDK's `L2MessageServiceContract.getMessageStatus` functions that allow us to use
```
james-a-morris, line 20: To reduce the changes needed to eventually remove this should we co-locate these functions?

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -4,16 +4,38 @@ import { Contract } from "ethers";
...
+// our custom provider, with retry and caching logic, to get around the SDK's hardcoded logic to query events
+// from 0 to "latest" which will not work on all RPC's.
+async function getL1ToL2MessageStatusUsingCustomProvider(
+  messageService: L2MessageServiceContract,
+  messageHash: string,
+  l2Provider: ethers.providers.Provider
+): Promise<OnChainMessageStatus> {
+  const l2Contract = messageService.contract.connect(l2Provider);
+  const status: BigNumber = await l2Contract.inboxL1L2MessageStatus(messageHash);
+  switch (status.toString()) {
```
james-a-morris, line 30: It may be more intuitive to create a mapping and resolve the status by using mapping[status.toString()]

james-a-morris (commented):
Does this finalize the messages that are throwing errors? I've left a few nits as code makes sense.

As a general thought I think we should co-locate all of the code that will be removed when Linea releases 1.0.

nicholaspai:
> As a general thought I think we should co-locate all of the code that will be removed when Linea releases 1.0.

I think the issue here is that the entire linea/ folder will change, the code is very different

src/finalizer/utils/linea/common.ts
nicholaspai: this is a copy but I inject our custom providers

src/finalizer/utils/linea/l1ToL2.ts
nicholaspai: I'm not sure how much difference that would make https://github.com/across-protocol/relayer/pull/1931#issuecomment-2547261958

src/finalizer/utils/arbStack.ts
nicholaspai: hm I think its ok, if this fails we likely have other issues

src/finalizer/utils/linea/l1ToL2.ts
nicholaspai: f6bfe92e419955edd88f16cb6f384ff103fd87ca

PR #1933, opened 2024-11-28
fix(Dataworker): Account for total required netSendAmount when executing PoolRebalanceLeaves

nicholaspai:
## Context

The dataworker executor functionality is supposed to detect when to call `sync()` before executing L1 PoolRebalance and RelayerRefund leaves depending on the `liquidReserves` value of l1 tokens before executing those leaves.

There are two important areas where we might want to call `sync` before executing any pool rebalance leaves:

1. First, we check if we should call `sync` when executing the Ethereum pool rebalance leaf and the Ethereum relayer refund leaves. Pool rebalance leaves subtract `netSendAmount` from `liquidReserves` while refund leaves add `amountToReturn` to liquid reserves (because Ethereum refund leaves are on the same chain as the hub pool, this `amountToReturn` is transferred atomically). This step is done [here](https://github.com/across-protocol/relayer/blob/3e9c1e4108568b39fe007f8fcd71721db4bbe090/src/dataworker/Dataworker.ts#L1484).
2. Secondly, we [check](https://github.com/across-protocol/relayer/blob/3e9c1e4108568b39fe007f8fcd71721db4bbe090/src/dataworker/Dataworker.ts#L1529) if we need to call `sync` before executing the non-Ethereum pool rebalance leaves. These all will subtract their `netSendAmounts` from the HubPool. This step subtly [takes in any updated liquid reserves](https://github.com/across-protocol/relayer/blob/3e9c1e4108568b39fe007f8fcd71721db4bbe090/src/dataworker/Dataworker.ts#L1530) value from the first step if we executed a `sync` pre-execution of the Ethereum pool rebalance leaves. It also is aware of any used balance from `netSendAmounts` in the first step via the [BalanceAllocator](https://github.com/across-protocol/relayer/blob/3e9c1e4108568b39fe007f8fcd71721db4bbe090/src/dataworker/Dataworker.ts#L1531).

## Problem

In the second step where we call [`_updateExchangeRatesBeforeExecutingNonHubChainLeaves`](https://github.com/across-protocol/relayer/blob/3e9c1e4108568b39fe007f8fcd71721db4bbe090/src/dataworker/Dataworker.ts#L1765) we are not evaluating the total netSendAmount obligation of all non-Ethereum PoolRebalance leaves we want to execute. This means we can submit a transaction to execute a set of pool leaves whose `netSendAmounts` might sum greater than the `liquidReserves` (plus any BalanceAllocator adjustments), but individually the `netSendAmounts` are all less than `liquidReserves`. This batch transaction would then fail in simulation.

## Examples of problems

We saw a bundle of PoolRebalanceLeaves today fail to execute because three of the leaves, one Ethereum leaf and two non-Ethereum leaves, had a total `netSendAmount` greater than the HubPool's `liquidReserves`, but each individually had a `netSendAmount` < the `liquidReserves`. For example, the three leaves had `netSendAmounts` of:
- 40
- 90
- 70

While the hubPool's liquidReserves was 180:
- 40 + 90 + 70 = 200 > 180
- 40 < 180
- 90 < 180
- 70 < 180

If you take these numbers and run them through the `executePoolRebalanceLeaves` code above, you'll see how a PoolRebalanceLeaf execution was able to be submitted but then fail in simulation, without preceding the non-Ethereum pool rebalance leaf executions with a `sync` transaction.

## PR side effects

This PR also refines the logic within the various internal functions called by `executePoolRebalanceLeaves`:
- `_updateExchangeRatesBeforeExecutingHubChainLeaves`: enqueues a `sync` call if necessary to execute any hub chain pool leaves
- `_executePoolRebalanceLeaves`: submits `executeRootBundle` calls but also adds any `loadEthForL2Calls`/`transfer` calls required to pay for an Arbitrum orbit L1->L2 message
- `_updateExchangeRatesBeforeExecutingNonHubChainLeaves`: enqueues a `sync` call if necessary to execute any non-hub chain pool leaves. Useful after `_updateExchangeRatesBeforeExecutingHubChainLeaves` which might pull Ethereum_SpokePool funds into the HubPool that need to be synced right before executing non-Hubchain leaves
- Various Dataworker logs are refined to be more helpful. For example, the `blockRangesAreInvalid` warning will now indicate which chain specifically requires a longer lookback.

Unit tests are updated as well


src/dataworker/Dataworker.ts
```diff
@@ -1772,81 +1772,94 @@ export class Dataworker {
...
-        if (currHubPoolLiquidReserves.gte(leaf.netSendAmounts[idx])) {
-          this.logger.debug({
-            at: "Dataworker#_updateExchangeRatesBeforeExecutingNonHubChainLeaves",
-            message: `Skipping exchange rate update for ${tokenSymbol} because current liquid reserves > netSendAmount for chain ${leaf.chainId}`,
-            l2ChainId: leaf.chainId,
-            currHubPoolLiquidReserves,
-            netSendAmount: leaf.netSendAmounts[idx],
-            l1Token,
-          });
-          return;
```
nicholaspai, line 1811: The issue with this code is that each individual pool rebalance leaf is treated individually, and if the current liquid reserves are enough to execute one leaf, the liquid reserves value is not decremented when performing the same check on the next pool rebalance leaf. This can lead to a situation where all individual `netSendAmounts` are less than the liquid reserves, but their aggregate value is greater than the liquid reserves

src/dataworker/Dataworker.ts
```diff
@@ -1772,81 +1772,94 @@ export class Dataworker {
...
-            l1Token,
-            netSendAmount: leaf.netSendAmounts[idx],
-            updatedLiquidReserves,
-            virtualHubPoolBalance,
-          });
-          return;
-        }
+      // Subtract any used balance from previous pool rebalance leaf executions.
+      currHubPoolLiquidReserves = currHubPoolLiquidReserves.sub(
+        balanceAllocator.getUsed(hubPoolChainId, l1Token, hubPool.address)
```
nicholaspai, line 1808: TODO. I think I should remove all the balance allocator logic that reads the hub pools balance. For executing pool leaves we care about liquid reserves which isn't quite the same as the balance. This might be causing some weird issues where we think a leaf can be executed but it's not. 

We should instead pass the updated liquid reserves minus and net s nd amounts from the first _executeHubChainPoolLeaves method to the second _executeNonHubChainPoolLeaves. 

This will give us the same logic while removing balance allocator logic.

nicholaspai (commented):
Moving to draft stage as I think we can simplify this logic more https://github.com/across-protocol/relayer/pull/1933#discussion_r1863961291

I've been thinking for a couple days now that something feels weird about including the balance allocator logic when executing pool leaves. It works when executing spoke leaves because spokes don't have a separate accounting of "liquid reserves"--all ERC20 balances are used to execute leaves, whereas the hub pool has a special accounting method.

src/dataworker/Dataworker.ts
```diff
@@ -1551,111 +1552,74 @@ export class Dataworker {
...
-              requests.push({
-                tokens: [ZERO_ADDRESS],
-                amount: this._getRequiredEthForArbitrumPoolRebalanceLeaf(leaf),
-                holder: await this.clients.hubPoolClient.hubPool.signer.getAddress(),
-                chainId: hubPoolChainId,
-              });
-            }
-          }
-
-          const success = await balanceAllocator.requestBalanceAllocations(
```
nicholaspai, line 1598: We no longer use the BalanceAllocator here to decide whether a leaf can be executed using the hub pool's balance. This is because the BalanceAllocator is designed to read balances using ERC20.balanceOf queries, but this doesn't really apply for the HubPool which has its own "currentReserves" accounting system. Therefore, this PR changes this method `_executePoolRebalanceLeaves` to assume that all leaves passed into it can be executed using reserves. 

This method `_executePoolRebalanceLeaves` is therefore now responsible for:
- checking that there is enough ETH/AZERO to pay for an Arbitrum orbit leaf execution, and if not submit appropriate transactions to fund the hub pool
-  submit `executeRootBundle` transactions

src/dataworker/Dataworker.ts
```diff
@@ -1551,111 +1552,74 @@ export class Dataworker {
...
               args: [],
               message: `Loaded ETH for message to ${getNetworkName(leaf.chainId)} 📨!`,
-              mrkdwn,
-              value: this._getRequiredEthForArbitrumPoolRebalanceLeaf(leaf),
+              mrkdwn: `Root hash: ${tree.getHexRoot()}\nLeaf: ${leaf.leafId}\nChain: ${leaf.chainId}`,
+              value: requiredAmount,
+            });
+          } else {
+            this.clients.multiCallerClient.enqueueTransaction({
+              contract: new Contract(feeToken, ERC20.abi, signer),
```
nicholaspai, line 1609: New logic to handle messages relayed to custom gas token orbit chain

src/dataworker/Dataworker.ts
```diff
@@ -1724,29 +1686,31 @@ export class Dataworker {
...
-        this.logger.error({
-          at: "Dataworker#_updateExchangeRatesBeforeExecutingHubChainLeaves",
-          message: `Not enough funds to execute pool rebalance leaf on HubPool for token: ${tokenSymbol}`,
-          poolRebalanceLeaf,
-          netSendAmount: netSendAmounts[idx],
-          currentPooledTokens,
-          updatedPooledTokens,
-        });
-        return;
+        throw new Error(
```
nicholaspai, line 1700: New logic to throw early if we won't have enough funds to execute a pool rebalance leaf, rather than send a PD alert and then fail when the TransactionClient simulation fails

src/dataworker/Dataworker.ts
```diff
@@ -1759,94 +1723,90 @@ export class Dataworker {
...
-    return updatedL1Tokens;
+    return {
+      availableLiquidReserves,
+      syncedL1Tokens,
+    };
   }
 
   async _updateExchangeRatesBeforeExecutingNonHubChainLeaves(
     latestLiquidReserves: Record<string, BigNumber>,
-    balanceAllocator: BalanceAllocator,
```
nicholaspai, line 1767: Remove use of balanceAllocator in this function and instead assume that passed in `latestLiquidReserves` account for `liquidReserves` in the HubPool, including if it was `sync`-ed in a previous step, minus any netSendAmounts from prior pool rebalance leaf executions

src/dataworker/Dataworker.ts
```diff
@@ -1759,94 +1723,90 @@ export class Dataworker {
...
+        hubPool.interface.encodeFunctionData("sync", [l1Token]),
+        hubPool.interface.encodeFunctionData("pooledTokens", [l1Token]),
+      ];
+      const multicallOutput = await hubPool.callStatic.multicall(multicallInput);
+      const updatedPooledTokens = hubPool.interface.decodeFunctionResult("pooledTokens", multicallOutput[1]);
+      const updatedLiquidReserves = updatedPooledTokens.liquidReserves;
+
+      // If the virtual balance is still too low to execute all the pool leaves, then log an error
+      if (updatedLiquidReserves.lt(requiredNetSendAmountForL1Token)) {
+        throw new Error(
```
nicholaspai, line 1797: Similar to https://github.com/across-protocol/relayer/pull/1933/files#r1864476640, throw error rather than just log PD alert

test/Monitor.ts
```diff
@@ -237,7 +237,7 @@ describe("Monitor", async function () {
     await monitorInstance.updateCurrentRelayerBalances(reports);
 
     expect(reports[depositor.address]["L1Token1"][ALL_CHAINS_NAME][BalanceType.CURRENT].toString()).to.be.equal(
-      "60000000000000000000000"
+      "75000000000000000000000"
```
nicholaspai, line 240: Needed because we added a `setupTokensForWallet` call in the `Dataworker.fixture`

src/dataworker/Dataworker.ts
```diff
@@ -1724,29 +1723,31 @@ export class Dataworker {
...
+          `Not enough funds to execute Ethereum pool rebalance leaf on HubPool for token: ${tokenSymbol}`
+        );
       }
-
+      availableLiquidReserves[l1Token] = updatedLiquidReserves.sub(netSendAmounts[idx]);
+      syncedL1Tokens.add(l1Token);
       this.logger.debug({
         at: "Dataworker#_updateExchangeRatesBeforeExecutingHubChainLeaves",
         message: `Updating exchange rate update for ${tokenSymbol} because we need to update the liquid reserves of the contract to execute the hubChain poolRebalanceLeaf.`,
```
pxrl, line 1745:
```suggestion
        message: `Updating exchange rate for ${tokenSymbol} because we need to update the liquid reserves of the contract to execute the hubChain poolRebalanceLeaf.`,
```

src/dataworker/Dataworker.ts
```diff
@@ -1759,94 +1760,107 @@ export class Dataworker {
...
+
     await sdkUtils.forEachAsync(poolRebalanceLeaves, async (leaf) => {
       await sdkUtils.forEachAsync(leaf.l1Tokens, async (l1Token, idx) => {
-        const tokenSymbol = this.clients.hubPoolClient.getTokenInfo(hubPoolChainId, l1Token)?.symbol;
-
-        if (updatedL1Tokens.has(l1Token)) {
-          return;
+        if (aggregateNetSendAmounts[l1Token] === undefined) {
+          aggregateNetSendAmounts[l1Token] = bnZero;
         }
```
pxrl, line 1784:
```suggestion
        aggregateNetSendAmounts[l1Token] ??= bnZero;
```

src/dataworker/Dataworker.ts
```diff
@@ -1759,94 +1760,107 @@ export class Dataworker {
...
+      assert(requiredNetSendAmountForL1Token.gte(0), "Aggregate net send amount should be >= 0");
+      if (requiredNetSendAmountForL1Token.eq(0)) {
+        return;
+      }
+      // The "used" balance kept in the BalanceAllocator should have adjusted for the netSendAmounts and relayer refund leaf
+      // executions above. Therefore, check if the current liquidReserves is less than the pool rebalance leaf's netSendAmount
+      // and the virtual hubPoolBalance would be enough to execute it. If so, then add an update exchange rate call to make sure that
+      // the HubPool becomes "aware" of its inflow following the relayre refund leaf execution.
+      const currHubPoolLiquidReserves =
+        latestLiquidReserves[l1Token] ?? this.clients.hubPoolClient.getLpTokenInfoForL1Token(l1Token).liquidReserves;
```
pxrl, line 1808: `HubPoolClient.getLpTokenInfoForL1Token()` can return `undefined`. Dereferencing `liquidReserves`  will cause a crash in that case, whereas an assert will at least make it clear what happened.

```suggestion
        latestLiquidReserves[l1Token] ?? this.clients.hubPoolClient.getLpTokenInfoForL1Token(l1Token)?.liquidReserves;
```

src/dataworker/Dataworker.ts
nicholaspai: The nice property of throwing here is we get the invariant that all leavs are executed at once, or none at all

src/common/Constants.ts
```diff
@@ -611,3 +611,31 @@ export const DEFAULT_GAS_MULTIPLIER: { [chainId: number]: number } = {
 };
 
 export const CONSERVATIVE_BUNDLE_FREQUENCY_SECONDS = 3 * 60 * 60; // 3 hours is a safe assumption for the time
+
+export const ARBITRUM_ORBIT_L1L2_MESSAGE_FEE_DATA: {
+  [chainId: number]: {
+    // Amount of tokens required to send a single message to the L2
+    amountBNWei: number;
```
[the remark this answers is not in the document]
nicholaspai: Good point will change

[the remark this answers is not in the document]
nicholaspai: 1b6da03df2cbcca75116c2b20fa3b5f7695108ff

src/dataworker/Dataworker.ts
pxrl: RE https://github.com/across-protocol/relayer/pull/1933/files#r1872013595, should we try to execute as much as we can in order to minimise disruption to the protocol?

src/dataworker/Dataworker.ts
```diff
@@ -836,24 +833,20 @@ export class Dataworker {
...
         at: "Dataworke#validate",
         message: "Cannot validate bundle with insufficient event data. Set a larger DATAWORKER_FAST_LOOKBACK_COUNT",
-        rootBundleRanges: blockRangesImpliedByBundleEndBlocks,
-        availableSpokePoolClients: Object.keys(spokePoolClients),
-        earliestBlocksInSpokePoolClients,
-        spokeClientsEventSearchConfigs: Object.fromEntries(
-          Object.entries(spokePoolClients).map(([chainId, client]) => [chainId, client.eventSearchConfig])
+        failReason: blockRangesAreInvalid.reason,
+        bundleBlockRanges: Object.fromEntries(
+          chainIds.map((chainId, i) => [chainId, blockRangesImpliedByBundleEndBlocks[i]])
```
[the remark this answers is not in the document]
nicholaspai: yeah good point

src/dataworker/DataworkerUtils.ts
```diff
@@ -60,10 +60,14 @@ export async function blockRangesAreInvalidForSpokeClients(
     // There should be a spoke pool client instantiated for every bundle timestamp.
     assert(!Object.keys(endBlockTimestamps).some((chainId) => !isDefined(spokePoolClients[chainId])));
   }
-  return utils.someAsync(blockRanges, async ([start, end], index) => {
+  let reason: string | undefined;
+  const _result = await utils.someAsync(blockRanges, async ([start, end], index) => {
```
pxrl, line 64: ooc, why is this `_result` instead of `result` ?

```suggestion
  const result = await utils.someAsync(blockRanges, async ([start, end], index) => {
```

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: a638412cb31cc6cb054c1664d8dc4cb9bf3abb1e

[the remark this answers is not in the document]
pxrl: @nicholaspai Was that the complete intended change?

src/dataworker/Dataworker.ts
```diff
@@ -1569,93 +1555,106 @@ export class Dataworker {
...
+        });
+        if (submitExecution) {
+          const canFund = await balanceAllocator.requestBalanceAllocations([
+            {
+              tokens: [feeToken],
+              amount: requiredAmount,
+              holder: await signer.getAddress(),
+              chainId: hubPoolChainId,
+            },
+          ]);
```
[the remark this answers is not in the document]
nicholaspai: a22ddf7ec2d4a1b5d777ae7187c67ced149e98b1

src/dataworker/Dataworker.ts
```diff
@@ -1569,93 +1555,106 @@ export class Dataworker {
...
+              value: requiredAmount,
+            });
+          } else {
+            this.clients.multiCallerClient.enqueueTransaction({
+              contract: new Contract(feeToken, ERC20.abi, signer),
+              chainId: hubPoolChainId,
+              method: "transfer",
+              args: [holder, requiredAmount],
+              message: `Loaded orbit gas token for message to ${getNetworkName(leaf.chainId)} 📨!`,
+              mrkdwn: `Root hash: ${tree.getHexRoot()}\nLeaf: ${leaf.leafId}\nChain: ${leaf.chainId}`,
```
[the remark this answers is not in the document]
nicholaspai: I find it helpful and its only a `debug`. If we ever run out of funds and need to load more this will be useful

src/dataworker/DataworkerUtils.ts
nicholaspai: 770738fabd6c76bb2ff259a680a5450967d19674

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
pxrl: Ah, I mean specifically the `mrkdwn` key.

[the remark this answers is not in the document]
nicholaspai: oh probably not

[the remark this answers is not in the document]
nicholaspai: actually, we do it already in `loadEthForL2Calls` so I'm going to leave it

src/dataworker/Dataworker.ts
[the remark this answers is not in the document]
nicholaspai: were you also looking for this? 1115567fa1b1dae982f1b6395d2492482bc4487f sorry

src/dataworker/Dataworker.ts
nicholaspai: new logic here to not throw on failed mainnet leaf execution 11b7d95ec74e84b00bef04f1cc083842bc672408

src/dataworker/Dataworker.ts
nicholaspai: new logic to try to execute as many non hub chain leaves as possible: 41a2aae1a44d95ddc543534fa5f0888c6e3ac82a

src/dataworker/Dataworker.ts
```diff
@@ -1541,26 +1538,54 @@ export class Dataworker {
...
+          return true;
+        }
+        assert(
+          availableLiquidReserves[l1Token] !== undefined && availableLiquidReserves[l1Token].gte(0),
+          "availableLiquidReserves should be defined for all l1 tokens"
+        );
+        if (availableLiquidReserves[l1Token].lt(netSendAmountForLeaf)) {
+          return false;
+        }
+        availableLiquidReserves[l1Token] = availableLiquidReserves[l1Token].sub(netSendAmountForLeaf);
```
mrice32, line 1583: Maybe I'm missing something obvious, but isn't this logic essentially exactly what the balance allocator is designed to handle? You tell it the balance change you need and it tells you if it's possible, right?

If so, why not use it?

src/dataworker/Dataworker.ts
mrice32: Shouldn't current reserves always be equal to balanceOf (assuming you make the sync call as needed)? Or is there a case I'm missing?

mrice32 (commented):
Just two questions

src/dataworker/Dataworker.ts
bmzig: A little irrelevant to the changes made in this PR, but does the dataworker fund the hub pool? I thought it was the monitor.

src/dataworker/Dataworker.ts
```diff
@@ -1721,132 +1753,180 @@ export class Dataworker {
...
+          at: "Dataworker#_updateExchangeRatesBeforeExecutingNonHubChainLeaves",
+          message: `Skipping exchange rate update for ${tokenSymbol} because liquid reserves would not increase`,
+          currHubPoolLiquidReserves,
+          updatedLiquidReserves,
+          l1Token,
+        });
+      }
     });
 
     // Submit executions at the end since the above double loop runs in parallel and we don't want to submit
```
bmzig, line 1932: Re this comment: https://github.com/across-protocol/relayer/pull/1933#discussion_r1873487484 If we do not have enough liquid reserves to execute the mainnet rebalance leaf and then we also do not have enough to execute the l2 pool rebalances leaves, will we run this transaction twice (i.e. once from `_updateExchangeRatesBeforeExecutingHubChainLeaves` and once from `_updateExchangeRatesBeforeExecutingNonHubChainLeaves`)?

src/dataworker/Dataworker.ts
```diff
@@ -1721,132 +1753,180 @@ export class Dataworker {
...
+      const multicallInput = [
+        hubPool.interface.encodeFunctionData("sync", [l1Token]),
+        hubPool.interface.encodeFunctionData("pooledTokens", [l1Token]),
+      ];
+      const multicallOutput = await hubPool.callStatic.multicall(multicallInput);
+      const updatedPooledTokens = hubPool.interface.decodeFunctionResult("pooledTokens", multicallOutput[1]);
+      const updatedLiquidReserves = updatedPooledTokens.liquidReserves;
+
+      // If the post-sync balance is still too low to execute all the pool leaves, then log an error
+      if (updatedLiquidReserves.lt(requiredNetSendAmountForL1Token)) {
```
bmzig, line 1893: Re: These comments: https://github.com/across-protocol/relayer/pull/1933#discussion_r1873488122 We don't return anymore [here](https://github.com/across-protocol/relayer/blob/e393a92aad9273925160684eae6483b1aede228e/src/dataworker/Dataworker.ts#L1838), which means we might submit a duplicate transaction, since we continue here until the end. Is that right?

src/dataworker/Dataworker.ts
nicholaspai: re-added balance allocator here: 630cafc7dc9d2172fbb3c74b1b4e0225d5de0edb

it actually simplified logic

src/dataworker/Dataworker.ts
nicholaspai: In this case, we'll likely run a `sync` after the `_updateExchangeRatesBeforeExecutingNonHubChainLeaves` function. This will let us try to execute as many non-ethereum leaves as possible

src/dataworker/Dataworker.ts
nicholaspai: I don't think we'll send a duplicate `sync` because the `updatedL1Tokens` list is a set. So even if we hit the same l1 token on multiple leaves we'll only add to the set once

src/dataworker/Dataworker.ts
```diff
@@ -1721,132 +1753,180 @@ export class Dataworker {
...
+        syncedL1Tokens.add(l1Token);
+        this.logger.debug({
+          at: "Dataworker#_updateExchangeRatesBeforeExecutingHubChainLeaves",
+          message: `Updating exchange rate for ${tokenSymbol} because we need to update the liquid reserves of the contract to execute the hubChain poolRebalanceLeaf.`,
+          poolRebalanceLeaf,
+          netSendAmount: netSendAmounts[idx],
+          currentLiquidReserves,
+          updatedLiquidReserves,
         });
+        if (submitExecution) {
```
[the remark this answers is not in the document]
bmzig: Ah yeah OK this is why it works. Sounds good.

src/dataworker/Dataworker.ts
bmzig: Never mind. I see now.

src/dataworker/Dataworker.ts
nicholaspai: The dataworker funds the hub pool when trying to execute leaves

pxrl (approved):

src/dataworker/DataworkerUtils.ts
```diff
@@ -60,10 +60,14 @@ export async function blockRangesAreInvalidForSpokeClients(
...
   }
-  return utils.someAsync(blockRanges, async ([start, end], index) => {
+  let reason: string | undefined;
+  const result = await utils.someAsync(blockRanges, async ([start, end], index) => {
     const chainId = chainIdListForBundleEvaluationBlockNumbers[index];
     // If block range is 0 then chain is disabled, we don't need to query events for this chain.
     if (isNaN(end) || isNaN(start)) {
+      reason = `block range contains undefined block for chain ${chainId}: [isNaN(start): ${isNaN(
+        start
+      )}, isNaN(end): ${isNaN(end)}]`;
```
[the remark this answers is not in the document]
nicholaspai: Hmm this is a great call. I think we should implement this as an iterative loop then, since this is only called once on dataworker runs we don't have a very high need for speed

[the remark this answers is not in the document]
nicholaspai: Actually I think this function can be kept async if we change it to instead return all chain ids that are invalid. This might be more useful to us for debugging and it keeps the latency constant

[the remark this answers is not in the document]
nicholaspai: 2bf69736fa4b539ee70b1c66985812504af405ec

pxrl (approved):

PR #1934, opened 2024-11-30
fix(relayer): Respect chainId input argument in ethers listener

pxrl:


src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -122,7 +122,7 @@ async function run(argv: string[]): Promise<void> {
   };
   const args = minimist(argv, minimistOpts);
 
-  const { chainId: chainId, lookback, relayer = null, blockrange: maxBlockRange = 10_000 } = args;
+  const { chainid: chainId, lookback, relayer = null, blockrange: maxBlockRange = 10_000 } = args;
```
nicholaspai, line 125: Should we fix this typo ?

nicholaspai (approved):

PR #1936, opened 2024-12-03
chore: secp256k1

james-a-morris:


nicholaspai (approved):

package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@across-protocol/constants": "^3.1.20",
-    "@across-protocol/contracts": "^3.0.16",
+    "@across-protocol/contracts": "^3.0.17",
     "@across-protocol/sdk": "^3.3.21",
```
pxrl, line 15: @james-a-morris I think we also need to bump SDK to 3.3.24 here. Without that we probably end up in a subtle type mismatch relating to the MerkeTree type, because the SDK also imports this from the contracts repo.

james-a-morris: Agreed - I'm going to refactor against master when I land - I think there's an issue with this PR where circle CI isn't recognizing tsc so the CI is borked

pxrl (approved):
Nice!

PR #1940, opened 2024-12-08
refactor(config): Cleanup & permit selective override

pxrl:
Permit certain config objects (i.e. MAX_BLOCK_LOOK_BACK) to be selectively overridden, rather than requiring a complete definition. This will slightly improve new chain automation.

Also simplify a few instances of config parsing.


src/common/Config.ts
```diff
@@ -57,22 +58,25 @@ export class CommonConfig {
...
     // `maxRelayerLookBack` is how far we fetch events from, modifying the search config's 'fromBlock'
     this.maxRelayerLookBack = Number(MAX_RELAYER_DEPOSIT_LOOK_BACK ?? Constants.MAX_RELAYER_DEPOSIT_LOOK_BACK);
-    this.hubPoolChainId = Number(HUB_CHAIN_ID ?? CHAIN_IDs.MAINNET);
     this.pollingDelay = Number(POLLING_DELAY ?? 60);
-    this.spokePoolChainsOverride = SPOKE_POOL_CHAINS_OVERRIDE ? JSON.parse(SPOKE_POOL_CHAINS_OVERRIDE) : [];
-    this.maxBlockLookBack = MAX_BLOCK_LOOK_BACK ? JSON.parse(MAX_BLOCK_LOOK_BACK) : {};
-    if (Object.keys(this.maxBlockLookBack).length === 0) {
-      this.maxBlockLookBack = Constants.CHAIN_MAX_BLOCK_LOOKBACK;
-    }
-    this.maxTxWait = Number(MAX_TX_WAIT_DURATION ?? 180); // 3 minutes
```
pxrl, line 75: `maxTxWait` is deleted here because it's not used in the bots.

src/common/Config.ts
```diff
@@ -57,22 +58,25 @@ export class CommonConfig {
...
+    this.spokePoolChainsOverride = JSON.parse(SPOKE_POOL_CHAINS_OVERRIDE ?? "[]");
+
+    // Inherit the default eth_getLogs block range config, then sub in any env-based overrides.
+    this.maxBlockLookBack = { ...Constants.CHAIN_MAX_BLOCK_LOOKBACK };
+    const maxBlockLookBack = JSON.parse(MAX_BLOCK_LOOK_BACK ?? "{}");
+    Object.entries(JSON.parse(MAX_BLOCK_LOOK_BACK ?? "{}")).forEach((_chainId, lookback) =>
+      this.maxBlockLookBack[Number(_chainId)] = lookback
+    );
```
pxrl, line 78: This will cut about 10 - 12 lines of config for each bot in bot-configs.

src/common/Config.ts
```diff
@@ -57,22 +57,24 @@ export class CommonConfig {
...
-    if (Object.keys(this.maxBlockLookBack).length === 0) {
-      this.maxBlockLookBack = Constants.CHAIN_MAX_BLOCK_LOOKBACK;
-    }
-    this.maxTxWait = Number(MAX_TX_WAIT_DURATION ?? 180); // 3 minutes
+    this.spokePoolChainsOverride = JSON.parse(SPOKE_POOL_CHAINS_OVERRIDE ?? "[]");
+
+    // Inherit the default eth_getLogs block range config, then sub in any env-based overrides.
+    this.maxBlockLookBack = { ...Constants.CHAIN_MAX_BLOCK_LOOKBACK };
+    Object.entries(JSON.parse(MAX_BLOCK_LOOK_BACK ?? "{}")).forEach(
+      (_chainId, lookback) => (this.maxBlockLookBack[Number(_chainId)] = lookback)
```
nicholaspai, line 75: This looks pretty similar to the above added code block. Can you refactor into a single function?

nicholaspai (commented):
One refactoring suggestion!

src/common/Config.ts
```diff
@@ -45,6 +44,7 @@ export class CommonConfig {
     } = env;
 
     this.version = ACROSS_BOT_VERSION ?? "unknown";
+    this.hubPoolChainId = Number(HUB_CHAIN_ID ?? CHAIN_IDs.MAINNET);
```
james-a-morris, line 47: Should we enforce this ENV? this feels important to throw if the user doesn't specify

james-a-morris (commented):
one comment

src/common/Config.ts
pxrl: In 99% of cases the user wants mainnet, so I think it makes sense to keep that as the default.

src/common/Config.ts
pxrl: Good call - it could be refactored. I applied that here: https://github.com/across-protocol/relayer/pull/1940/commits/0c78797b8559b496f6d3d35164e0d08b7f4c9fac

james-a-morris (approved):

src/common/Config.ts
```diff
@@ -44,7 +43,11 @@ export class CommonConfig {
       ARWEAVE_GATEWAY,
     } = env;
 
+    const updateConfig = (envVar: string, config: Record<string, unknown>) =>
+      Object.entries(JSON.parse(envVar ?? "{}")).forEach(([k, v]) => (config[k] = v));
```
nicholaspai, line 47: this might be easier to reason about if you create an empty new config rather than update it in-place. I don't think there's anything wrong with the implementation but it seems like a potential nasty bug if there's an issue

nicholaspai (approved):

src/common/Config.ts
pxrl: Yeah, I see the point. The existing uses of it are perfectly safe because the initial objects start out as shallow copies. I experimented a bit with some alternative solutions but ended in type purgatory. I'll merge it as-is for now but I'll think about how it can be improved without making it annoying to maintain and use.

pxrl: OK I found a way: https://github.com/across-protocol/relayer/pull/1940/commits/75566e5c087f54ed85554463ebb7f3b966a1ba7e

This also adds some basic type validation, such that a type mismatch on the env-based JSON definition produces the following error:

`AssertionError [ERR_ASSERTION]: Invalid { "1": 5000, "10": "100" } configuration on key 10 (string != number)`

src/common/Config.ts
```diff
@@ -44,7 +43,20 @@ export class CommonConfig {
       ARWEAVE_GATEWAY,
     } = env;
 
+    const mergeConfig = <T>(config: T, envVar: string): T => {
```
nicholaspai, line 46: Would be cool to add this to utils/ObjectUtils and add a unit test. Can be a follow up and low prio

PR #1949, opened 2024-12-16
improve(Relayer): Add total gas price to `info` log when relayer sends fill

nicholaspai:
This will help debugging and give us broad oversight into how accurate our gas price estimates are. We probably should refactor the overall relay fee calculation flow but this PR uses the data we already have (native and token gas costs) to derive the gas price easily.

This PR also adds to the ProfitClient's `updateGasCosts` function a simple change to print out the gas price estimate per chain.

A more helpful fix might be to add the `gasPrice` field to the `TransactionCostEstimate` type in the SDK and force functions like `sdk.common.estimateTotalGasRequiredByUnsignedTransaction()` to return the priority fee and base fee broken down so the relayer can log it


src/clients/ProfitClient.ts
```diff
@@ -630,7 +630,13 @@ export class ProfitClient {
       assert(isDefined(outputToken), `Chain ${destinationChainId} SpokePool is not configured for ${symbol}`);
 
       const deposit = { ...sampleDeposit, destinationChainId, outputToken };
-      this.totalGasCosts[destinationChainId] = await this._getTotalGasCost(deposit, relayer);
+      const { nativeGasCost, tokenGasCost } = await this._getTotalGasCost(deposit, relayer);
+      const gasPrice = tokenGasCost.div(nativeGasCost);
+      this.totalGasCosts[destinationChainId] = {
```
nicholaspai, line 635: This adds the gas price estimate early in the relayer run to the ProfitClient's update function log

src/clients/ProfitClient.ts
```diff
@@ -33,7 +33,7 @@ import { Deposit, DepositWithBlock, L1Token, SpokePoolClientsByChain } from "../
 import { getAcrossHost } from "./AcrossAPIClient";
 import { HubPoolClient } from "./HubPoolClient";
 
-type TransactionCostEstimate = sdkUtils.TransactionCostEstimate;
+export type ProfitClientTransactionCostEstimate = sdkUtils.TransactionCostEstimate & { gasPrice: BigNumber };
```
pxrl, line 36: Should we just upstream this to the SDK? It's in `src/utils/common.ts`.

src/relayer/Relayer.ts
```diff
@@ -1375,12 +1376,14 @@ export class Relayer {
       .mul(fixedPointAdjustment)
       .div(deposit.inputAmount);
     const totalFeePct = formatFeePct(_totalFeePct);
+    const totalGasCosts = this.clients.profitClient.getGasCostsForChain(deposit.destinationChainId);
+    const gasPriceGwei = convertFromWei(totalGasCosts.gasPrice.toString(), 9);
```
pxrl, line 1380: This will fwiw almost always log the same result _per run_ because the ProfitClient will only read new gas prices when we it is updated, and that only occurs once per run at present. So simply logging the gas price that was read on ProfitClient update should be sufficient.

We should fwiw probably be updating the ProfitClient more frequently than we currently do. It should probably be at least once per minute so that we're up to date on the actual gas cost as measured in the gas token.

pxrl (approved):

src/clients/ProfitClient.ts
nicholaspai: good call: https://github.com/across-protocol/sdk/pull/795

src/relayer/Relayer.ts
nicholaspai: > So simply logging the gas price that was read on ProfitClient update should be sufficient.

This is true but I also think in the future if we make this more dynamic per fill, then this change will be backwards compatible

pxrl (approved):
LGTM

src/clients/ProfitClient.ts
```diff
@@ -465,13 +478,14 @@ export class ProfitClient {
...
-  ): Promise<Pick<FillProfit, "profitable" | "nativeGasCost" | "tokenGasCost" | "netRelayerFeePct">> {
+  ): Promise<Pick<FillProfit, "profitable" | "nativeGasCost" | "gasPrice" | "tokenGasCost" | "netRelayerFeePct">> {
     let profitable = false;
     let netRelayerFeePct = bnZero;
     let nativeGasCost = uint256Max;
     let tokenGasCost = uint256Max;
+    let gasPrice = uint256Max;
     try {
-      ({ profitable, netRelayerFeePct, nativeGasCost, tokenGasCost } = await this.getFillProfitability(
+      ({ profitable, netRelayerFeePct, nativeGasCost, tokenGasCost, gasPrice } = await this.getFillProfitability(
```
nicholaspai, line 488: You can see in this file how tedious it is to keep passing fill costs from function to function, its almost like we want a single function to get the profitability data for a single fill, and then also in this client we want a single function to get the client's stored profitability data

src/relayer/Relayer.ts
```diff
@@ -720,7 +728,14 @@ export class Relayer {
...
       // expensive fills on (for example) mainnet.
       const { lpFeePct } = lpFees.find((lpFee) => lpFee.paymentChainId === destinationChainId);
-      this.fillRelay(deposit, destinationChainId, lpFeePct);
+      // For self-relays, gas price is not a concern because we are bypassing profitability requirements so
+      // use profit client's gasprice.
+      this.fillRelay(
+        deposit,
+        destinationChainId,
+        lpFeePct,
+        this.clients.profitClient.getGasCostsForChain(destinationChainId).gasPrice
```
nicholaspai, line 737: I didn't want to call `resolveRepaymentChain` again here. Maybe we should compute it at the top level once outside all the if-elses but i didn't want to mess up any tests

nicholaspai: I think the issue here is we don't want to run the full resolveRepaymentChain logic if the deposit.recipient == relayer

pxrl: How about we just make `gasPrice` optional, so we only have to include it when we actually want to log it?

nicholaspai: i'd rather not have optional args and this seems easy enough to call/remove

PR #1956, opened 2024-12-18
chore(gskms): add file support for `GCKMS_CONFIG` env var

adamazad:
Adds support for file paths in the `GCKMS_CONFIG` environment variable, allowing configurations to be loaded from a file or directly from an inline JSON string. It improves error handling with clear messages for missing or empty configurations and refines the `.GckmsOverride.js` logic for better maintainability. These changes enhance flexibility and simplify integration with various deployment setups.

### DIFF
--- a/src/utils/GckmsUtils.ts
+++ b/src/utils/GckmsUtils.ts
@@ -3,6 +3,7 @@
 import kms from "@google-cloud/kms";
 import { Storage } from "@google-cloud/storage";
 import fs from "fs";
+import path from "node:path";
 export interface KeyConfig {
   projectId: string;
   locationId: string;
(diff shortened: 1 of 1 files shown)


james-a-morris (changes requested):
Hey! Thanks for the contribution 😎

I think it’s a good idea to refine the error logs. I wonder if we can refactor this change such that the file existence check (fs.existsSync) is only performed once. Currently, it’s being called in both branches of the if statement, which duplicates work.

adamazad:
Hi @james-a-morris, I made some changes, `fs.existsSync` is called once. It won't throw an error if you pass a JSON (when that's the case)

james-a-morris (commented):
Hey! Thanks for these Looks good - however this current build fails 😞

adamazad:
@james-a-morris I updated the branch, and now CI is passing. :)

PR #1957, opened 2024-12-19
improve(Dataworker): Remove 0-value refunds and 0-value empty-message slow fills

nicholaspai:
implements optimizations to [UMIP-179](https://github.com/UMAprotocol/UMIPs/pull/606)

The goal here is protecting dataworker from performing extra computations where no on-chain state changes can happen, in the case of 0-value refunds and 0-value empty-message slow fills.


pxrl:
+1. How do we synchronise this change on chain? I also have https://github.com/UMAprotocol/UMIPs/pull/602 open - I got stalled on it because it also requires synchronisation - i.e. we need to identify when the transition would occur. We could do this by mainnet block number, or some change to the config store (bumping the VERSION field?).

nicholaspai:
> +1. How do we synchronise this change on chain? I also have [UMAprotocol/UMIPs#602](https://github.com/UMAprotocol/UMIPs/pull/602) open - I got stalled on it because it also requires synchronisation - i.e. we need to identify when the transition would occur. We could do this by mainnet block number, or some change to the config store (bumping the VERSION field?).

I don't think we need any special logic to be honest, we likely don't need to run the dataworker code on any previous bundles anymore (I'll check no stuck leaves) and we can time this with a bundle with no 0-value refunds/0-value slow fills. 

This way we can avoid the version bump

src/dataworker/DataworkerUtils.ts
```diff
@@ -160,10 +161,12 @@ export function _buildSlowRelayRoot(bundleSlowFillsV3: BundleSlowFills): {
   // Append V3 slow fills to the V2 leaf list
   Object.values(bundleSlowFillsV3).forEach((depositsForChain) => {
     Object.values(depositsForChain).forEach((deposits) => {
-      deposits.forEach((deposit) => {
-        const v3SlowFillLeaf = buildV3SlowFillLeaf(deposit, deposit.lpFeePct);
-        slowRelayLeaves.push(v3SlowFillLeaf);
-      });
+      deposits
```
nicholaspai, line 164:
```suggestion
      // Do not create slow fill leaves where the amount to transfer would be 0 and the message is empty
      deposits
```

src/dataworker/DataworkerUtils.ts
```diff
@@ -160,10 +161,13 @@ export function _buildSlowRelayRoot(bundleSlowFillsV3: BundleSlowFills): {
   // Append V3 slow fills to the V2 leaf list
   Object.values(bundleSlowFillsV3).forEach((depositsForChain) => {
     Object.values(depositsForChain).forEach((deposits) => {
-      deposits.forEach((deposit) => {
-        const v3SlowFillLeaf = buildV3SlowFillLeaf(deposit, deposit.lpFeePct);
-        slowRelayLeaves.push(v3SlowFillLeaf);
-      });
+      // Do not create slow fill leaves where the amount to transfer would be 0 and the message is empty
+      deposits
+        .filter((deposit) => deposit.inputAmount.gt(0) || !utils.isMessageEmpty(deposit.message))
```
nicholaspai, line 166: can eventually use the `isZeroValueDeposit` function from the [sdk](https://github.com/across-protocol/sdk/pull/800/files) once released

mrice32 (dismissed):
LGTM!

pxrl (approved):

PR #1963, opened 2024-12-20
improve: split netValue from bridgeValue in the atomic depositor

bmzig:
Transfer tracking will not work with our current atomic depositor for any case where the bridge fee is included in the value to send to the L1-L2 bridge. The only case so far is ZkSync, but there may be more in the future. This lets us specify the amount we expect to receive on L2 to the function so that the event fired on L1 can be properly matched on L2.


bmzig:
Are there any opinions against making this change? The issue with the current atomic depositor is that the initiation event we observe is for `x` amount of eth, while the finalization is for some amount strictly less than `x`, so the bots can't match these events. 

If there are no objections, I can redeploy and reconfigure the depositor with the changes.

nicholaspai (approved):

PR #1966, opened 2024-12-23
feat: enable Ink in the relayer

james-a-morris:
Enables the Ink chain in the relayer.


src/common/Constants.ts
```diff
@@ -598,6 +615,21 @@ export const OPSTACK_CONTRACT_OVERRIDES = {
...
+  [CHAIN_IDs.INK]: {
+    l1: {
+      AddressManager: "0x9b7c9bbd6d540a8a4dedd935819fc4408ba71153",
+      L1CrossDomainMessenger: "0x69d3cf86b2bf1a9e99875b7e2d9b6a84426c171f",
+      L1StandardBridge: CONTRACT_ADDRESSES[CHAIN_IDs.MAINNET].ovmStandardBridge_57073.address,
+      StateCommitmentChain: ZERO_ADDRESS,
+      CanonicalTransactionChain: ZERO_ADDRESS,
+      BondManager: ZERO_ADDRESS,
+      OptimismPortal: "0x5d66c1782664115999c47c9fa5cd031f495d3e4f",
+      L2OutputOracle: ZERO_ADDRESS,
```
james-a-morris, line 627: I couldn't find this entry in the `OptimismPortal`. However, the `DisputeGameFactory` is available

bmzig: I don't think finalizations will work without this address, so you may need to ask Ink about it.

bmzig:
Since it doesn't look like we are supporting any USD stablecoin on Ink right away, you probably need to add `WETH` as an entry here: https://github.com/across-protocol/relayer/blob/e9ad355f90016218ec989c709ee7e62c94648a59/src/clients/ProfitClient.ts#L605

src/common/Constants.ts
bmzig, line 1: I think you also need to add entries to these since Ink is OpStack:
https://github.com/across-protocol/relayer/blob/197b995cb566fa61ff38a321b51bff3dcdf87859/src/common/Constants.ts#L261
https://github.com/across-protocol/relayer/blob/197b995cb566fa61ff38a321b51bff3dcdf87859/src/common/Constants.ts#L635

james-a-morris: https://github.com/across-protocol/relayer/pull/1966/commits/1ce7d09f55a48f294683850c5aaf29b2cb10d880

james-a-morris:
> Since it doesn't look like we are supporting any USD stablecoin on Ink right away, you probably need to add `WETH` as an entry here:
> 
> https://github.com/across-protocol/relayer/blob/e9ad355f90016218ec989c709ee7e62c94648a59/src/clients/ProfitClient.ts#L605

https://github.com/across-protocol/relayer/pull/1966/commits/1ce7d09f55a48f294683850c5aaf29b2cb10d880

src/common/Constants.ts
james-a-morris: I've reached out

bmzig (approved):
LGTM pending the address of the output oracle

src/common/Constants.ts
```diff
@@ -159,6 +164,7 @@ export const BUNDLE_END_BLOCK_BUFFERS = {
   [CHAIN_IDs.BOBA]: 0, // **UPDATE** 288 is disabled so there should be no buffer.
   [CHAIN_IDs.LINEA]: 40, // At 3s/block, 2 mins = 40 blocks.
   [CHAIN_IDs.LISK]: 60, // 2s/block gives 2 mins buffer time.
+  [CHAIN_IDs.INK]: 120, // 1s/block gives 2 mins buffer time
```
pxrl, line 167: @james-a-morris btw if Ink is 1s/block then we'll need to add a manual entry to the SDK BlockUtils:

https://github.com/across-protocol/sdk/pull/806

src/common/Constants.ts
james-a-morris: Per Ink the `L2PriceOracle` is deprecated on chains that have enabled fault proofs. ` DisputeGameFactory` is the replacement.

src/clients/ProfitClient.ts
```diff
@@ -608,6 +608,7 @@ export class ProfitClient {
       [CHAIN_IDs.LISK]: "USDT", // USDC is not yet supported on Lisk, so revert to USDT. @todo: Update.
       [CHAIN_IDs.REDSTONE]: "WETH", // Redstone only supports WETH.
       [CHAIN_IDs.WORLD_CHAIN]: "WETH", // USDC deferred on World Chain.
```
nicholaspai, line 610: Unrelated but is USDC still deferred on WorldChain and not supported on Lisk?

bmzig (approved):
LGTM. Sounds right that we don't need the output oracle for updated OP chains. I've tried running finalizations on Base/Optimism without the address and they work fine.

src/clients/bridges/AdapterManager.ts
```diff
@@ -185,6 +199,19 @@ export class AdapterManager {
       );
     }
 
+    if (this.spokePoolClients[INK] !== undefined) {
```
nicholaspai, line 202: Do you think we can consolidate this constructor? A lot of these BaseChainAdapter initializations, for example, seem similar

src/common/Constants.ts
```diff
@@ -505,6 +522,7 @@ export const EXPECTED_L1_TO_L2_MESSAGE_TIME = {
   [CHAIN_IDs.ARBITRUM]: 20 * 60,
   [CHAIN_IDs.BASE]: 20 * 60,
   [CHAIN_IDs.BLAST]: 20 * 60,
+  [CHAIN_IDs.INK]: 60 * 60,
```
nicholaspai, line 525: 60 minutes? damn that's slow, especially compared to the other OpStack chains. You sure this is right?

src/common/Constants.ts
nicholaspai: yeah this may or may not cause the finalizer to throw. I'd expect it to be ok since the SDK won't need deprecated contracts but I don't remember

nicholaspai (commented):
Left a few q's

src/common/Constants.ts
james-a-morris: Done

src/clients/ProfitClient.ts
james-a-morris: cc @pxrl

src/clients/bridges/AdapterManager.ts
james-a-morris: Could be worth a fast follow to not add this to scope

james-a-morris: https://github.com/across-protocol/relayer/pull/1973 - will close this current comment

src/common/Constants.ts
james-a-morris: Fair call - this should be 20.

james-a-morris: https://github.com/across-protocol/relayer/pull/1966/commits/b5bc42881210711aa376b9ac645614c66a26c783

nicholaspai (approved):

PR #1967, opened 2024-12-23
refactor(AdapterManager): define base chain adapters programmatically

bmzig:
This is the last change of code needed to fully switch to generic adapters. It was delayed so that these new base chain adapters would be out in the wild for long enough such that it is unlikely for them to contain any critical issues. This PR will effectively make all of the named adapters (e.g. ZKSyncAdapter, ArbitrumAdapter, CCTPAdapter, BaseAdapter, etc.) in `src/client/bridges` unused code (to be deleted in a follow-up), and also make the entire `op-stack` directory unused.

To be clear, there is still a substantial amount of code which needs to be removed, but all further follow-up removals will have no affect on the bots, since we would be removing code which is only touched in out-of-date unit tests.


src/clients/bridges/AdapterManager.ts
```diff
@@ -46,172 +43,37 @@ export class AdapterManager {
...
-        logger,
-        SUPPORTED_TOKENS[ALEPH_ZERO],
-        constructBridges(ALEPH_ZERO),
-        DEFAULT_GAS_MULTIPLIER[ALEPH_ZERO] ?? 1
-      );
-    }
-
-    if (this.spokePoolClients[INK] !== undefined) {
-      this.adapters[INK] = new BaseChainAdapter(
+    Object.keys(this.spokePoolClients).map((_chainId) => {
```
bmzig, line 62: A side-effect of defining the adapters this way is, since we have an empty Boba spoke pool client, we will define an empty boba adapter. This should have no affect on anything, though, because it is empty.

nicholaspai (approved):

james-a-morris (approved):

src/clients/bridges/AdapterManager.ts
```diff
@@ -46,172 +43,37 @@ export class AdapterManager {
...
+      const chainId = Number(_chainId);
+      assert(chainId.toString() === _chainId);
```
pxrl, line 64: If you iterate over the values and dereference the chainId directly this then you can drop the all-keys-are-strings dance:

```suggestion
    Object.values(this.spokePoolClients).map(({ chainId }) => {
```

pxrl (approved):

src/clients/bridges/AdapterManager.ts
pxrl: What's the actual content of this? An empty client, or `undefined`?

If it's `undefined´ then we could add a filter pass to skip it. This might actually be necessary if the below suggestion is used.

bmzig: It is an empty client.

src/clients/bridges/AdapterManager.ts
bmzig: The reason why I didn't do it this way was because our unit tests do not always define a chain id in the spoke pool client, so I assumed that we can't be positive that the chainId field is defined (at least, not in our unit tests).

PR #1968, opened 2024-12-24
improve(Constants): Update default deposit confirmations

pxrl:
The existing defaults are set so conservatively that third-party operators are pushed into overriding them with custom values. This introduces a greater risk of loss of funds. Instead, update the defaults with lower confirmation requirements that should permit new operators to make fills successfully without having to discard the defaults.

As a side benefit, this will marginally simplify the process of onboard OP or Orbit chains.


nicholaspai (dismissed):

james-a-morris (approved):
+1

nicholaspai (approved):

PR #1969, opened 2024-12-24
improve: Use base fee multiplier option in SDK to resolve gas fees

nicholaspai:
Builds on https://github.com/across-protocol/sdk/pull/801 and applies relayer's base fee multiplier to new SDK parameter to scale the base fee in one use case:
- Submitting transactions with an estimated gas fee via TransactionClient
- ~Estimating fill costs via the ProfitClient~
    - We decided against using the baseFeeMultiplier when estimating fill costs and instead using the default multiplier of 1.0. This ensures the fills are submitted using an estimated gas price based on current gas prices, which is fine for the filler because they know exactly when they are going to submit the fill.

We can test this by querying `/api/gas-prices` against the prod API deployment. The gas prices should be mostly in line otherwise the relayer will see too many deposits as unprofitable potentially

Fixes ACX3599


src/utils/TransactionUtils.ts
```diff
@@ -176,7 +184,7 @@ export async function getGasPrice(
 
   // Default to EIP-1559 (type 2) pricing.
   return {
-    maxFeePerGas: scaleByNumber(feeData.maxFeePerGas, Math.max(priorityScaler * maxFeePerGasScaler, 1)),
+    maxFeePerGas: feeData.maxFeePerGas,
     maxPriorityFeePerGas: scaleByNumber(feeData.maxPriorityFeePerGas, priorityScaler),
```
pxrl, line 188: If we are retaining the ability to scale the priority fee (which we probably need to), then we'd still need to adjust the base fee here. Otherwise we'd be eating into our base fee multiplier by increasing the priority fee.

nicholaspai: ok so do you suggest keeping the existing code here?

src/clients/ProfitClient.ts
```diff
@@ -243,6 +243,14 @@ export class ProfitClient {
...
     const gasToken = this.resolveGasToken(chainId);
     const gasTokenPriceUsd = this.getPriceOfToken(gasToken.symbol);
+    // TODO: This is where the relayer's computation of the tokenGasCost (gasUnits * gasFee) diverges from the
+    // quoting API's. The latter uses the SDK's GasPriceOracle.getGasPriceEstimate() function and passes in an optional
+    // baseFeeMultiplier. The relayer uses this.getTotalGasCost() to compute the gasUnits the same way that the
+    // API does but then computes the tokenGasCost by computing the gasPrice using a baseFeeMultiplier hardcoded
+    // to 1.0. The relayer applies its own padding on the full tokenGasCost (not just the base fee!) using
+    // this.gasPadding. To make the two implementations use the same computation, then the relayer should set
+    // baseFeeMultiplier = this.gasPadding when calling getTotalGasCost which itself will call the GasPriceOracle's
+    // getGasPriceEstimate function.
```
nicholaspai, line 253: This is challenging to reason about @pxrl . Should we just pass in the gasPadding as the baseFeeMultiplier when calling the SDK? Or should we keep the relayer padding the full token gas cost (gasPrice * gasUnits)?

src/clients/ProfitClient.ts
```diff
@@ -205,7 +205,9 @@ export class ProfitClient {
 
   private async _getTotalGasCost(deposit: Deposit, relayer: string): Promise<TransactionCostEstimate> {
     try {
-      return await this.relayerFeeQueries[deposit.destinationChainId].getGasCosts(deposit, relayer);
+      return await this.relayerFeeQueries[deposit.destinationChainId].getGasCosts(deposit, relayer, {
+        baseFeeMultiplier: this.gasPadding,
```
nicholaspai, line 209: Using gasPadding here should align the relayer with the API

src/utils/TransactionUtils.ts
```diff
@@ -82,7 +82,12 @@ export async function runTransaction(
       Number(process.env[`MAX_FEE_PER_GAS_SCALER_${chainId}`] || process.env.MAX_FEE_PER_GAS_SCALER) ||
       DEFAULT_GAS_FEE_SCALERS[chainId]?.maxFeePerGasScaler;
 
-    const gas = await getGasPrice(provider, priorityFeeScaler, maxFeePerGasScaler);
+    const gas = await getGasPrice(
+      provider,
+      priorityFeeScaler,
+      maxFeePerGasScaler,
+      await contract.populateTransaction[method](...(args as Array<unknown>), { value })
```
nicholaspai, line 89: getting gas price here should be same as the gas price used in estimating the fill cost in the ProfitClient

test/ProfitClient.ConsiderProfitability.ts
```diff
@@ -207,28 +208,49 @@ describe("ProfitClient: Consider relay profit", () => {
     }
   });
 
-  it("Verify gas padding", async () => {
+  it("Verify gas padding on SDK call is set correctly", async () => {
```
nicholaspai, line 211: In this test, I assume that the SDK works correctly as long as the relayer passes in the correct params into the SDK call. Namely-- does the relayer use `baseFeeMultiplier` correctly when calling relayerFeeQueries.getGasCosts

src/clients/ProfitClient.ts
nicholaspai: I decided to set baseFeeMultiplier = this.gasPadding to make it easier to reason about how the relayer computes gas prices. It should ideally be aligned with API

src/clients/ProfitClient.ts
pxrl: `gasPadding` has the valid range [0, 2] scaled to 18 decimals, so we need clamp this to a minimum of 1e18.

pxrl: I'm not 100% sure we want to do this though. The intention of the base fee multiplier is to account for time-based uncertainty around when the fill will be made. That's a problem in the API, but it's not a problem for the relayer because it's doing this calculation JIT, so it has a lot more certainty about when the fill would be submitted.

Additionally, the intention of `gasPadding` is to add some padding for the fill-time gas _consumption_ (i.e. how many units of gas are consumed). This change would instead pad the base fee, which may then be discounted by the gasMultiplier.

tldr I think the relayer should probably retain its existing behaviour here, such that:
- gasPadding is applied to the estimated gas consumption.
- baseFeeMultiplier is implicitly 1.0 because the relayer does not really need to consider uncertainty about the base fee when the transaction lands (the base fee is already returned from the pending block).

src/clients/ProfitClient.ts
```diff
@@ -636,24 +637,49 @@ export class ProfitClient {
...
+          destinationChainId === hubPoolClient.chainId
+            ? hubToken
+            : hubPoolClient.getL2TokenForL1TokenAtBlock(hubToken, destinationChainId);
+        assert(isDefined(outputToken), `Chain ${destinationChainId} SpokePool is not configured for ${symbol}`);
+
+        const deposit = { ...sampleDeposit, destinationChainId, outputToken };
+        const gasCosts = await this._getTotalGasCost(deposit, relayer);
+        // The scaledNativeGasCost is approximately what the relayer will set as the `gasLimit` when submitting
+        // fills on the destination chain.
+        const scaledNativeGasCost = gasCosts.nativeGasCost.mul(this.gasPadding).div(fixedPointAdjustment);
```
nicholaspai, line 654: This `scaledNativeGasCost` and `scaledTokenGasCost` are the only things changing in this function. It's only to produce a more detailed log in the relayer's update() run so we can compare the `scaledTokenGasCost` with the API's quoted `/gas-prices=>tokenGasCost` to make sure the relayer won't consider quotes as unprofitable

bmzig (approved):

src/clients/ProfitClient.ts
```diff
@@ -636,24 +637,49 @@ export class ProfitClient {
...
+          .div(fixedPointAdjustment)
+          .mul(this.gasMultiplier)
+          .div(fixedPointAdjustment);
+        this.totalGasCosts[destinationChainId] = gasCosts;
+        return [
+          destinationChainId,
+          {
+            ...gasCosts,
+            scaledNativeGasCost,
+            scaledTokenGasCost,
```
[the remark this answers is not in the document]
nicholaspai: yep

src/utils/TransactionUtils.ts
nicholaspai: I ended up adding a priority fee scaler parameter in the SDK that we'll use here instead

james-a-morris (approved):

PR #1970, opened 2024-12-26
chore: remove all unused adapter code

bmzig:
This removes all the unused code in `src/clients/bridges`. There are two tests which are removed with this change and justification for removing those tests in the comments of this PR.


test/AdapterManager.getOutstandingCrossChainTokenTransferAmount.ts
bmzig, line 1: We need to remove this test since the new base chain adapter does not store events (and therefore has no `l1DepositInitiatedEvents` nor `l2DepositFinalizedEvents` in state. It also has no `computeOutstandingCrossChainTransfers` function. This should be OK to delete since each adapter "type" e.g. OpStack, Arbitrum, zkSync, etc. has their own test in `test/generic-adapters` which tests cross chain transfer tracking.

test/cross-chain-adapters/OpStack.ts
bmzig, line 1: We need to remove this test since we no longer have the OpStackAdapter. This test has since been replaced by this test: https://github.com/across-protocol/relayer/blob/master/test/generic-adapters/OpStack.ts

nicholaspai (commented):
Very excited for this PR!

bmzig:
Moving all of the files from `src/adapter` into `src/clients/bridges` will increase the diff by thousands of lines, so I am going to hold off on doing that this PR.

james-a-morris (commented):
Have we tested this PR with simulated runs of the relayer/dataworker?

bmzig:
> Have we tested this PR with simulated runs of the relayer/dataworker?

I can run everything fine. This PR in general _should_ not break anything since it doesn't make any changes to code that we use. To be precise, we don't instantiate any `*Adapter` anymore (e.g. ArbitrumAdapter, LineaAdapter, etc.), everything in the `op-stack` directory has been unused since [this PR](https://github.com/across-protocol/relayer/pull/1839), and then everything else are tests.

nicholaspai (approved):

pxrl (approved):

PR #1973, opened 2024-12-30
refactor(AdapterManager): Simplify adapter initialization

james-a-morris:


nicholaspai (approved):

pxrl:
@james-a-morris This is also handled by Bennett's PR right? https://github.com/across-protocol/relayer/pull/1967

PR #1975, opened 2024-12-30
fix(Dataworker): Cannot fund DonationBox using MulticallerClient

nicholaspai:
The MulticallerClient has [logic](https://github.com/across-protocol/relayer/blob/40b0c7971a3548a0499e7fc5990d150437c527fb/src/clients/MultiCallerClient.ts#L361) that batches transactions together and attempts to `multicall()` them. This works normally with the `HubPool` and `SpokePool` which extend the `Multicaller()` interface.

However, we can't use this logic to make calls on ERC20's like we are doing to fund the `DonationBox`. Instead, we either need to update the `MulticallerClient` to handle cases where transactions are not multicall-able, or we can send these transactions to the `TransactionClient`.

In this PR, I added a new method `simulateAndSubmit` to the `TransactionClient` that the Dataworker uses to `transfer` custom gas tokens to the DonationBox.


src/dataworker/Dataworker.ts
```diff
@@ -1690,13 +1690,25 @@ export class Dataworker {
...
-            this.clients.multiCallerClient.enqueueTransaction({
-              contract: new Contract(feeToken, ERC20.abi, signer),
-              chainId: hubPoolChainId,
-              method: "transfer",
-              args: [holder, requiredAmount],
-              message: `Loaded orbit gas token for message to ${getNetworkName(leaf.chainId)} 📨!`,
-              mrkdwn: `Root hash: ${tree.getHexRoot()}\nLeaf: ${leaf.leafId}\nChain: ${leaf.chainId}`,
+            // We can't use the multicaller client here because the feeToken is not guaranteed to be a Multicaller
+            // contract and this is a permissioned function where the msg.sender needs to be the
+            // feeToken balance owner.
```
nicholaspai, line 1695:
```suggestion
            // feeToken balance owner, so we can't simply set `unpermissioned: true` to send it through the Multisender
```

test/Dataworker.executePoolRebalanceUtils.ts
```diff
@@ -706,15 +706,20 @@ describe("Dataworker: Utilities to execute pool rebalance leaves", async functio
...
-      expect(queuedTransactions[0].method).to.equal("transfer");
-      expect(queuedTransactions[0].args).to.deep.equal([customGasTokenFunder, expectedFeeLeaf1]);
-      expect(queuedTransactions[1].method).to.equal("transfer");
-      expect(queuedTransactions[1].args).to.deep.equal([customGasTokenFunder, expectedFeeLeaf2]);
-      expect(queuedTransactions[2].method).to.equal("executeRootBundle");
-      expect(queuedTransactions[3].method).to.equal("executeRootBundle");
+      // expect(queuedTransactions[0].method).to.equal("transfer");
+      // expect(queuedTransactions[0].args).to.deep.equal([customGasTokenFunder, expectedFeeLeaf1]);
+      // expect(queuedTransactions[1].method).to.equal("transfer");
+      // expect(queuedTransactions[1].args).to.deep.equal([customGasTokenFunder, expectedFeeLeaf2]);
```
melisaguevara, line 715: nit: can we delete these?

melisaguevara (approved):
lgtm!

melisaguevara (dismissed):

src/clients/TransactionClient.ts
```diff
@@ -121,4 +121,29 @@ export class TransactionClient {
...
   }
+
+  async simulateAndSubmit(chainId: number, txns: AugmentedTransaction[]): Promise<TransactionResponse[]> {
+    const simulationResults = await this.simulate(txns);
+    const txnsToSubmit: AugmentedTransaction[] = [];
+    this.logger.debug({
+      at: "TransactionClient",
+      message: `Simulating ${txns.length} transactions before submitting the ones that pass simulation`,
+    });
+    for (let i = 0; i < simulationResults.length; i++) {
```
bmzig, line 132: Is there any benefit to doing the simulations in parallel? Something like
```
forEachAsync(
simulationResults,
async (simulationResult) => {
  const { reason, succeed, transaction } = simulationResult;
  ...
  } else {
    txnsToSubmit.push(transaction);
  }
});
```

bmzig (approved):
One minor question.

src/clients/TransactionClient.ts
nicholaspai: good call!

nicholaspai: wait actually, the `this.simulate(txns)` call already processes the simulations for all transactions in parallel so we can't really improve this for loop

bmzig: Ah of course. My mistake.

bmzig (dismissed):

src/clients/TransactionClient.ts
james-a-morris, line 132: Any chance we could use `.reduce( )` here? It's not required but we usually defer to array function calls

james-a-morris (commented):
one comment otherwise lgtm

src/clients/TransactionClient.ts
nicholaspai: sure 0c42fc3fa9e144f89e47844b50aaf73b7c7fa548

src/clients/TransactionClient.ts
```diff
@@ -121,4 +121,31 @@ export class TransactionClient {
...
+            txnRequestData,
+          });
+        } else {
+          _txnsToSubmit.push(txns[i]);
+        }
+        return _txnsToSubmit;
+      },
+      []
+    );
+    return await this.submit(chainId, txnsToSubmit);
```
james-a-morris, line 149: nit
```suggestion
    return this.submit(chainId, txnsToSubmit);
```

src/dataworker/Dataworker.ts
```diff
@@ -1690,13 +1690,25 @@ export class Dataworker {
               value: requiredAmount,
             });
           } else {
-            this.clients.multiCallerClient.enqueueTransaction({
```
james-a-morris, line 1693: Potentially a very simple question but shouldn't MATIC have a non-empty `feeToken` or do we handle this differently because it's a much older integration?

nicholaspai: exactly ^ we don't need MATIC to execute a pool leaf because executions take place on L1 and the polygon canonical bridge doesn't charge any special fees

bmzig (dismissed):

src/dataworker/Dataworker.ts
james-a-morris: Very fair - tysm

james-a-morris (approved):

src/clients/MultiCallerClient.ts
```diff
@@ -80,6 +80,7 @@ export class MultiCallerClient {
   protected txnClient: TransactionClient;
   protected txns: { [chainId: number]: AugmentedTransaction[] } = {};
   protected valueTxns: { [chainId: number]: AugmentedTransaction[] } = {};
+  protected nonMulticallTxns: { [chainId: number]: AugmentedTransaction[] } = {};
```
nicholaspai, line 83: We probably could get away with renaming `valueTxn` to be `nonMulticallTxns` but I didn't want a large diff on this PR.

src/clients/MultiCallerClient.ts
```diff
@@ -101,8 +101,8 @@ export class MultiCallerClient {
 
   // Adds all information associated with a transaction to the transaction queue.
   enqueueTransaction(txn: AugmentedTransaction): void {
-    // Value transactions are sorted immediately because the UMA multicall implementation rejects them.
-    const txnQueue = txn.value && txn.value.gt(0) ? this.valueTxns : this.txns;
+    // We do not attempt to batch together transactions that have value or are explicitly nonMulticall.
+    const txnQueue = (txn.value && txn.value.gt(0)) || txn.nonMulticall ? this.nonMulticallTxns : this.txns;
```
nicholaspai, line 105: Closely check this boolean syntax

bmzig (approved):
+1 Very clever to use the logic of (what was formerly) value transactions to get the same outcome.

src/clients/MultiCallerClient.ts
bmzig: Looks right to me. Seems to just generalize the value transactions

PR #1978, opened 2024-12-31
chore: Bump SDK

pxrl:
For a provider-related quorum change, such that
eth_getBlockByNumber("pending") will not be subjected to the normal quorum config.

See here for the diff between sdk versions:
https://github.com/across-protocol/sdk/compare/v3.3.30...v3.3.31


pxrl:
nb will follow up with the subsequent yarn.lock sync.

dohaki (approved):

PR #1983, opened 2025-01-06
fix: sepolia bridge definitions

bmzig:
We can't run the relayer on Sepolia anymore since we are trying to define adapters without configuration. In the past, the testnet relayer would not instantiate any adapters, since it did not support rebalancing. While we still do not have support for rebalancing on testnet, since we define adapters programmatically now, we need to have a canonical bridge defined for all supported testnet chains.


nicholaspai (dismissed):

bmzig:
I also needed to add some bridge contract addresses so that the bridges themselves could construct.

nicholaspai (approved):

PR #1989, opened 2025-01-08
improve(Relayer): Log clearer about overallocated lite chain deposits 

nicholaspai:
We currently log deposits that are ignored because they are from over-allocated lite-chains as unprofitable, but this isn't super clear.

These deposits will have no preferred repayment chain [chosen](https://github.com/across-protocol/relayer/blob/72279954ddb71e008b1e7a6c43c8b901b55acc53/src/relayer/Relayer.ts#L1064) by the inventory client and the lpFeePct will get set to [infinity](https://github.com/across-protocol/relayer/blob/72279954ddb71e008b1e7a6c43c8b901b55acc53/src/relayer/Relayer.ts#L1082). This produces a very noisy log that claims the deposit is unproftable but in reality, it was ignored because it came from an overallocated lite chain.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1352,8 +1352,9 @@ export class Relayer {
         const formattedLpFeePct = formatFeePct(lpFeePct);
 
         // @dev If the origin chain is a lite chain and the LP fee percentage is infinity, then we can assume that the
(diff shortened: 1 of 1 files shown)


bmzig (approved):

mrice32 (approved):
LGTM!

PR #1990, opened 2025-01-09
improve(Relayer): Add information about message size to unprofitability log

nicholaspai:
When determining why a deposit is unprofitable its definitely useful to know upfront if there is a non empty message in the deposit, which often indicates we failed to simulate the fill.

This PR also sends the warning about failure to estimate gas costs for deposit to a different slack channel than "unprofitable fills". 

We should isolate across-unprofitable-fills channel to only include messages for deposits that are unprofitable. The developer should then go to across-warn to figure

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1324,8 +1324,8 @@ export class Relayer {
   }
 
   // TODO: This should really be renamed to "handleUnfillableDeposit" since it not only logs about unprofitable relayer
-  // fees but also about fills with messages that fail to simulate and deposits from lite chains that are 
-  // over-allocated. 
+  // fees but also about fills with messages that fail to simulate and deposits from lite chains that are
+  // over-allocated.
   private handleUnprofitableFill() {
     const { profitClient } = this.clients;
     const unprofitableDeposits = profitClient.getUnprofitableFills();
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -1356,7 +1356,7 @@ export class Relayer {
         // this lite chain edge case.
         const fromOverallocatedLiteChain = deposit.fromLiteChain && lpFeePct.isEqualTo(bnUint256Max);
         depositMrkdwn +=
-          `- Deposit ${
+          `- Deposit${
             isMessageEmpty(deposit.message)
               ? ""
               : ` with message of size ${ethersUtils.hexDataLength(deposit.message)} bytes`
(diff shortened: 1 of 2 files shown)


src/relayer/Relayer.ts
```diff
@@ -1071,7 +1071,6 @@ export class Relayer {
           ? `Deposit ${depositId} originated from over-allocated lite chain ${originChain}`
           : `Unable to identify a preferred repayment chain for ${originChain} deposit ${depositId}.`,
         txn: blockExplorerLink(transactionHash, originChainId),
-        notificationPath: "across-unprofitable-fills",
```
nicholaspai, line 1074: This is a debug log now and we don't need this notificationPath variable

src/relayer/Relayer.ts
```diff
@@ -1357,7 +1356,12 @@ export class Relayer {
         // this lite chain edge case.
         const fromOverallocatedLiteChain = deposit.fromLiteChain && lpFeePct.isEqualTo(bnUint256Max);
         depositMrkdwn +=
-          `- DepositId ${deposit.depositId} (tx: ${depositblockExplorerLink})` +
+          `- Deposit${
+            isMessageEmpty(deposit.message)
+              ? ""
+              : ` with message of size ${ethersUtils.hexDataLength(deposit.message)} bytes`
```
nicholaspai, line 1362: Maybe the hexDataLength info is overkill. It could be useful...or not. WDYT?

bmzig: Have you noticed that unprofitable fills tend to be ones with messages? If so, I'd say we should log it. Would be interesting to see if estimation gets worse (or better) the longer the message.

nicholaspai: No I've just seen that anytime a deposit has a message that causes the fill to revert that it gets logged as an unprofitable fill. This becomes confusing when you realize it's not actually unprofitable, it's just unfillable. So changing this log to make it more clear what's happening. 

I think there are very few instances where the message content itself affects the profitability but I can see this possibly changing the gas costs a lot for chains that use the Viem estimateGas() method in the sdk's gas price oracle. This is because this method takes in the fill txn calldata to simulate the call to return a different priority fee. Specifically Linea will return a priority fee dependent on the size of the data.

james-a-morris (dismissed):

src/relayer/Relayer.ts
```diff
@@ -1356,16 +1358,21 @@ export class Relayer {
...
           ` of input amount ${formattedInputAmount} ${inputSymbol}` +
           ` and output amount ${formattedOutputAmount} ${outputSymbol}` +
           ` from ${getNetworkName(originChainId)} to ${getNetworkName(destinationChainId)}` +
+          `${fromOverallocatedLiteChain ? " is from an over-allocated lite chain" : ""}` +
           `${
-            fromOverallocatedLiteChain
-              ? " is from an over-allocated lite chain.\n"
-              : ` with relayerFeePct ${formattedRelayerFeePct}%, lpFeePct ${formattedLpFeePct}%, and gas cost ${formattedGasCost} ${gasTokenSymbol} is unprofitable!\n`
-          }`;
+            depositFailedToSimulateWithMessage
```
james-a-morris, line 1369: General mention: not in love with the entire `+=` for the markdown but I understand we do it elsewhere.

james-a-morris (approved):

PR #2002, opened 2025-01-16
improve(Constants): Auto-populate OP stack deposit confirmations

pxrl:
Minor QoL improvement for new-chain additions.


james-a-morris (dismissed):
Could be worth also doing this for Arb Orbit chains

nicholaspai (approved):

bmzig (approved):

PR #2003, opened 2025-01-17
Feat: add initial SVM signer utils

chrismaree:
This PR adds the initial utilities to set us up to use SVM wallets within the repository. It's been tested against all wallet types we care about:
- Private keys
- GCKMS (by converting an EVM wallet to SVM wallet)
- Solana wallet file

Note that Solana and Ethereum private keys are structured and stored differently. `getSvmSignerFromEvmSigner` method is only useful if you cant access the associated private key in any other context, such as taking a GCKMs key from GCP and wanting to turn said key into a Solana wallet.


src/utils/SvmSignerUtils.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  createKeyPairSignerFromPrivateKeyBytes,
+  KeyPairSigner,
+  createKeyPairFromBytes,
+  createSignerFromKeyPair,
+  Wallet,
+} from "./";
+import fs from "fs";
+
+export async function getSvmSignerFromEvmSigner(evmSigner: Wallet): Promise<KeyPairSigner> {
```
chrismaree, line 10: ^

james-a-morris (commented):
This fails CI - is it ready to be reviewed?

chrismaree:
> This fails CI - is it ready to be reviewed?

fixing the CI issue. it is indeed ready to be reviewed!

src/utils/SvmSignerUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+export async function getSvmSignerFromFile(filePath: string): Promise<KeyPairSigner> {
+  const keypairFile = fs.readFileSync(filePath);
+  const keypairBytes = new Uint8Array(JSON.parse(keypairFile.toString()));
+
+  // Create a KeyPairSigner from the bytes.
+  const keys = await createKeyPairFromBytes(keypairBytes);
+  return await createSignerFromKeyPair(keys);
+}
+
+export async function getSvmSignerFromPrivateKey(privateKey: string): Promise<KeyPairSigner> {
```
james-a-morris, line 25: Could be worthwhile to add a comment here that this is a SVM specific private key and not the 32byte EVM private key given that we have a conversion function above.

src/utils/SvmSignerUtils.ts
james-a-morris: Do we have an idea if this is the way we plan on instantiating our SVM signers or if we will always get this from a file.

Makes sense if we don't know at this point, just curious

bmzig (approved):

PR #2006, opened 2025-01-20
fix: regenerated yarn.lock to fix bigint warning issue

md0x:
**Motivation**
When running the relayer, the following warning log appears:
`bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)`
This warning is also present in the tests run on CircleCI, as shown here: https://app.circleci.com/pipelines/github/across-protocol/relayer/6446/workflows/91e772d1-f4bf-4a2b-b27f-6b368c902b3c/jobs/17967?invite=true#step-104-3811_76

During my tests, I only identified a way to fix it by regenerating a new `yarn.lock` file.

**Steps taken**
To regenerate the `yarn.lock` file, I followed these steps:
1. Deleted the old `yarn.lock` file and the `node_modules` folder.
2. Ran `yarn install` to generate a fresh dependency tree and lock file.
3. Verified the integrity of the `node_modules` directory with:
`yarn check --integrity`

**Observations**
The regeneration resulted in a significant diff in the yarn.lock file.
This could impact the dependency tree in ways that are not fully understood, as some dependencies may have been updated to newer versions.
This could impact the dependency tree in ways that are not fully understood, as some dependencies may have been updated to newer versions.
While this resolves the warning issue, it might introduce unintended side effects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
             - cache-node-v1-{{ checksum "yarn.lock" }}
       - run:
           name: Install Dependencies
-          command: yarn install --immutable
+          command: yarn install --frozen-lockfile
       - save_cache:
           name: Save Yarn Package Cache
           key: cache-node-v1-{{ checksum "yarn.lock" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   install:
     docker:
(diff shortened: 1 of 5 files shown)


pxrl:
@james-a-morris This should ideally address the `bigint` issue. The problem is that we can't really comprehend the impact of it, given the huge diff. I think you've run dedup cleanup on `yarn.lock` previously, with a smaller diff. Could that be worth a try here?

PR #2015, opened 2025-01-25
fix(svm): bigint-buffer warning

md0x:
Motivation
When running the relayer, the following warning log appears:
`bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)`
This warning is also present in the tests run on CircleCI, as shown here: https://app.circleci.com/pipelines/github/across-protocol/relayer/6446/workflows/91e772d1-f4bf-4a2b-b27f-6b368c902b3c/jobs/17967?invite=true#step-104-3811_76

Steps taken
- Add node-gyp to dev dependencies
- Add postinstall script to handle native build requirements for the bigint-buffer package.

Reason for the Change:
The `bigint-buffer` package requires native bindings to be built using node-gyp. For some reason the default `yarn install` installation process does not automatically trigger this build step, leading to the warnings. Adding the postinstall script ensures that these bindings are always compiled correctly after installation.

Results After the Change:
After these changes, the tests no longer show the warning, as verified in this CircleCI pipeline: https://app.circleci.com/pipelines/github/across-protocol/relayer/6519/workflows/26952b16-715f-4677-af13-ed6668c4981f/jobs/18149?invite=true#step-104-3811_59


nicholaspai (dismissed):

package.json
```diff
@@ -49,7 +49,8 @@
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "scripts": {
-    "postinstall": "cd node_modules/bigint-buffer && node-gyp configure && node-gyp build",
+    "build-bigint-buffer": "[ -d node_modules/bigint-buffer ] && command -v node-gyp > /dev/null && cd node_modules/bigint-buffer && node-gyp configure && node-gyp build || echo 'Skipping bigint-buffer build: folder or node-gyp not found'",
```
md0x, line 52: This script checks if the `node_modules/bigint-buffer` folder exists and `node-gyp` is installed. If both conditions are met, it configures and builds the native bindings; otherwise, it skips the process with a message.

We will be able to remove this script from postinstall once we fully migrate to `@solana/web3.js` v2, drop v1 from the dependency structure, and eliminate the use of `bigint-buffer`.

pxrl (approved):

james-a-morris (approved):

nicholaspai (approved):

PR #2017, opened 2025-01-26
improve(Relayer): Don't fill exclusive deposits

nicholaspai:
SpokePool's today prevent filling deposits if `exclusivityDeadline >=  currentTime and msg.sender !== exclusiveRelayer`

The exclusive deadline is still respected even if the exclusiveRelayer = zero address so the relayer has a bad check in it currently that is causing it to try to fill deposits with an exclusivity deadline > current time but the exclusiveRelayer == 0

This PR should greatly reduce the simulation errors in the relayer

This is the existing contract code
image

## Soneium and Ink SpokePool

These two contracts actually allow ignore the exclusivityDeadline upon filling if the exclusiveRelayer == 0 but rather than adding special cases for these chains I propose using this fix unilaterally as it will be forwards compatible with the changes to exclusivity soon to be deployed


src/relayer/Relayer.ts
```diff
@@ -292,11 +291,7 @@ export class Relayer {
       return false;
     }
 
-    if (
-      deposit.exclusiveRelayer !== ZERO_ADDRESS &&
-      deposit.exclusivityDeadline > currentTime &&
```
nicholaspai, line 297: I also think this should be a >= instead of a > because the deposit is exclusive up until the exclusive timestamp exactly

mrice32 (approved):

nicholaspai:
Let's wait on merging this until all spoke pools are upgraded such that the exclusive relayer being 0 is impossible on the deposit side if an exclusivity deadline is sent and the exclusive relayer being 0 on the fill side is not a special case. 

We can live with the extra noise for now

james-a-morris (approved):
lgtm

nicholaspai:
Can merge this after spoke pool upgrade, otherwise we'll ignore filling deposits that have `exclusivityDeadline > currentTime` even if the exclusiveRelayer is 0x

mrice32 (approved):

PR #2021, opened 2025-01-27
fix(IndexedSpokePoolClient): Cast depositId correctly

nicholaspai:
SpokePoolClient doesn't set depositId correctly in a rare case.

Also, to be safe, log a known BN as a string where we currently don't do so

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -285,8 +285,8 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
     const _firstDepositId = events[fundsDeposited].at(0)?.args?.depositId;
     const _latestDepositId = events[fundsDeposited].at(-1)?.args?.depositId;
     const [firstDepositId, latestDepositId] = [
-      _firstDepositId ? BigNumber.from(_firstDepositId) : this.getDeposits().at(0)?.depositId ?? bnZero,
-      _latestDepositId ? BigNumber.from(_latestDepositId) : this.getDeposits().at(-1)?.depositId ?? bnZero,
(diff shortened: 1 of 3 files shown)


bmzig (dismissed):

pxrl (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -282,9 +282,11 @@ export class IndexedSpokePoolClient extends clients.SpokePoolClient {
 
     // Find the latest deposit Ids, and if there are no new events, fall back to already stored values.
     const fundsDeposited = eventsToQuery.indexOf("V3FundsDeposited");
+    const _firstDepositId = events[fundsDeposited].at(0)?.args?.depositId;
+    const _latestDepositId = events[fundsDeposited].at(-1)?.args?.depositId;
     const [firstDepositId, latestDepositId] = [
-      events[fundsDeposited].at(0)?.args?.depositId ?? this.getDeposits().at(0) ?? bnZero,
-      events[fundsDeposited].at(-1)?.args?.depositId ?? this.getDeposits().at(-1) ?? bnZero,
+      _firstDepositId ? BigNumber.from(_firstDepositId) : this.getDeposits().at(0)?.depositId ?? bnZero,
```
bmzig, line 288: Might want to do `isDefined(_firstDepositId)` since if the first deposit Id is 0, then this will evaluate as false. (Maybe this isn't an issue?)

bmzig: It's actually probably fine to keep, since `_firstDepositId` will become a BigNumber type anyways post contract upgrade.

bmzig (approved):

PR #2023, opened 2025-01-27
improve(dataworker): Fund more fee tokens into fee payer address

nicholaspai:
Reduces amortized gas cost at expense of locking up some capital. I think this is worth it given how often we need to top up the fee payer for executing L1 leaves


james-a-morris (approved):
Fair call. Is the change from 10 -> 50  based on data or is this a good estimate

nicholaspai:
> Fair call. Is the change from 10 -> 50 based on data or is this a good estimate

estimate

PR #2025, opened 2025-01-29
improve: prefer slow fills over self relays

bmzig:
We should prefer sending a slow fill over self relays.


nicholaspai (approved):

james-a-morris (commented):
looks good from a code standpoint, but what is the reason for this?

nicholaspai (approved):

PR #2027, opened 2025-01-29
improve(relayer): Try simulating transactions against pending blocks

pxrl:
Just throwing this up for discussion; it's unclear whether this is actually supported by RPC providers or whether it will work to address the exclusivity deadline issues that we occasionally see.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -191,6 +191,8 @@ export async function willSucceed(transaction: AugmentedTransaction): Promise<Tr
     return { transaction, succeed: true };
   }
 
+  const overrides = { blockTag: "pending" };
+
   const { contract, method } = transaction;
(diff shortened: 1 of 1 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -200,7 +202,7 @@ export async function willSucceed(transaction: AugmentedTransaction): Promise<Tr
   // relay custom errors well: https://github.com/ethers-io/ethers.js/discussions/3291#discussion-4314795
   let data;
   try {
-    data = await contract.callStatic[method](...args);
+    data = await contract.callStatic[method](...args, { overrides });
```
pxrl, line 205:
```suggestion
    data = await contract.callStatic[method](...args, overrides);
```

pxrl:
Doesn't work

PR #2032, opened 2025-01-31
feat: bump sdk to use new utilites

bmzig:
The main utility of this PR is to now coerce specific address fields ingested by the spoke pool client to bytes20 addresses. Functionally, this should have no effect on any of the bots, since all spoke pool events emit `address`es and not `bytes32` values, but this will be necessary to have active when the contracts upgrade goes live.

### DIFF
--- a/test/Dataworker.loadData.fill.ts
+++ b/test/Dataworker.loadData.fill.ts
@@ -818,7 +818,9 @@ describe("Dataworker: Load data used in all functions", async function () {
         const relayer2 = randomAddress();
         fillV3Events.push(generateV3FillFromDeposit(deposits[0]));
         fillV3Events.push(generateV3FillFromDeposit(deposits[1]));
-        fillV3Events.push(generateV3FillFromDeposit(deposits[2], {}, ethers.utils.randomBytes(32)));
+        fillV3Events.push(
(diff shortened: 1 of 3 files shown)


nicholaspai (dismissed):

test/Dataworker.loadData.fill.ts
```diff
@@ -818,7 +818,9 @@ describe("Dataworker: Load data used in all functions", async function () {
         const relayer2 = randomAddress();
         fillV3Events.push(generateV3FillFromDeposit(deposits[0]));
         fillV3Events.push(generateV3FillFromDeposit(deposits[1]));
-        fillV3Events.push(generateV3FillFromDeposit(deposits[2], {}, ethers.utils.randomBytes(32)));
+        fillV3Events.push(
+          generateV3FillFromDeposit(deposits[2], {}, ethers.utils.hexlify(ethers.utils.randomBytes(32)))
```
nicholaspai, line 822: Why is hexlify needed?

bmzig: It's not needed since ethers takes care of if it were a byte array under the hood, but when this test fails, the error logs look nicer. I.e. it might fail and say invalid address: [54, 53, 88, 134, ..., etc.] vs. invalid address: 0xf48eg....3f4.

PR #2042, opened 2025-02-04
chore: Prepare for deprecated message field

pxrl:


nicholaspai (approved):

pxrl:
nb we must bump the SDK before this can be merged; otherwise it hits a runtime error.

james-a-morris (dismissed):

nicholaspai (approved):

src/dataworker/Dataworker.ts
```diff
@@ -1190,21 +1192,17 @@ export class Dataworker {
 
     const sortedFills = client.getFills();
```
nicholaspai, line 1193: Would be nice if `client.getFills` auto populated the message hash

nicholaspai (approved):

src/dataworker/Dataworker.ts
bmzig: It should already have a message hash since it returns fill objects. Not sure if I understand the question.

mrice32 (approved):

PR #2043, opened 2025-02-04
feat: support updated spoke pool functions

bmzig:
The functions `requestSlowFill`, `executeSlowFillLeaf`, and `fillRelayWithUpdatedDeposit` are functions which do not have a backwards compatible counterpart in the new spoke pools. This means we must call these functions instead of the older ones when interacting with the updated spoke pools. This should be merged as soon as all spoke pools are migrated over (or even better, 10-20 minutes before that).


nicholaspai (dismissed):
What do you think is the best way to deploy this and make it backwards compatible so we have minimal downtime? These are rare functions to be called so maybe we don't need any switching.

Also, should we eventually call the new `fillRelay` function or are we good to keep calling the existing?

bmzig:
> What do you think is the best way to deploy this and make it backwards compatible so we have minimal downtime? These are rare functions to be called so maybe we don't need any switching.
> 
> Also, should we eventually call the new `fillRelay` function or are we good to keep calling the existing?

Calling `fillRelay` is strictly an improvement I think, so it is worthwhile swapping there. I just wanted to make this PR an non-invasive as possible.

pxrl (approved):

src/relayer/Relayer.ts
```diff
@@ -1001,7 +1001,7 @@ export class Relayer {
     const [method, messageModifier, args] = !isDepositSpedUp(deposit)
       ? ["fillV3Relay", "", [deposit, repaymentChainId]]
       : [
-          "fillV3RelayWithUpdatedDeposit",
+          "fillRelayWithUpdatedDeposit",
```
pxrl, line 1004: @bmzig I think we're missing to supply the relayer's repayment address as the 3rd argument to this function.

bmzig: True. https://github.com/across-protocol/relayer/pull/2043/commits/b18eec1094ecc3613b41f5a58905419c9f466ed9

src/relayer/Relayer.ts
```diff
@@ -1013,13 +1015,14 @@ export class Relayer {
...
       ? ["fillV3Relay", "", [deposit, repaymentChainId]]
       : [
-          "fillV3RelayWithUpdatedDeposit",
+          process.env.ENABLE_V6 ? "fillRelayWithUpdatedDeposit" : "fillV3RelayWithUpdatedDeposit",
           " with updated parameters ",
           [
-            deposit,
+            process.env.ENABLE_V6 ? convertRelayDataParamsToBytes32(deposit) : deposit,
             repaymentChainId,
+            process.env.ENABLE_V6 ? toBytes32(this.relayerAddress) : this.relayerAddress,
```
bmzig, line 1023: The repayment address argument doesn't exist in `fillV3RelayWithUpdatedDeposit`, so I think you will need to do `ENABLE_V6 ? [new args] : [old args]` instead.

src/relayer/Relayer.ts
```diff
@@ -1013,16 +1015,26 @@ export class Relayer {
     const [method, messageModifier, args] = !isDepositSpedUp(deposit)
       ? ["fillV3Relay", "", [deposit, repaymentChainId]]
       : [
-          "fillV3RelayWithUpdatedDeposit",
+          process.env.ENABLE_V6 ? "fillRelayWithUpdatedDeposit" : "fillV3RelayWithUpdatedDeposit",
```
mrice32, line 1018: Isn't fillRelayWithUpdatedDeposit sticking around?

nicholaspai: yes, we use it if the flag is set to true

PR #2044, opened 2025-02-05
Feat(cctpFinalizer) use same lookback as parent finalizer

nicholaspai:
I've tested this out on prod and it doesn't slow the finalizer down materially since there aren't that many L2 to L1 withdrawals in general, even when expanding the lookback from 24 hours to whatever the production l2 to l1 finalizer uses (20 days). On the other hand, the L1 to L2 finalizer in prod only uses an 8 hour lookback so this PR won't affect this finalizer in production


james-a-morris (dismissed):

james-a-morris (approved):
RE: the PR description

CCTP doesn't have a concept of L2 <> L1 finalizations

The code makes sense LGTM

PR #2046, opened 2025-02-05
feat: support doctor who

dohaki:


src/common/Constants.ts
```diff
@@ -654,11 +670,27 @@ export const OPSTACK_CONTRACT_OVERRIDES = {
     },
     l2: DEFAULT_L2_CONTRACT_ADDRESSES,
   },
+  [CHAIN_IDs.DOCTOR_WHO]: {
+    l1: {
+      AddressManager: "0x8098f676033a377b9defe302e9fe6877cd63d575",
```
dohaki, line 675: Taken from L1CrossDomainMessenger creation tx https://dashboard.tenderly.co/tx/1/0x3fc109d839b7e7962e45e580d85d554bee7f3284f88dc79c4a0ed4d6e0e8d6ee/state-diff

src/common/Constants.ts
```diff
@@ -415,6 +426,10 @@ export const CUSTOM_BRIDGE: {
     [TOKEN_SYMBOLS_MAP.DAI.addresses[CHAIN_IDs.MAINNET]]: BlastBridge,
     [TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET]]: OpStackWethBridge,
   },
+  [CHAIN_IDs.DOCTOR_WHO]: {
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: UsdcCCTPBridge,
```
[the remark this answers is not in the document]
bmzig: I think this is OK. We've never done it before so definitely make sure that it tests correctly, but it looks to be alright. If by chance something doesn't work as expected, you should be able to just use the token splitter bridge since it will auto use this bridge if the l2 token is native USDC, but I do think this is fine to keep as is.

src/finalizer/index.ts
```diff
@@ -128,6 +128,10 @@ const chainFinalizers: { [chainId: number]: { finalizeOnL2: ChainFinalizer[]; fi
     finalizeOnL1: [opStackFinalizer],
     finalizeOnL2: [],
   },
+  [CHAIN_IDs.DOCTOR_WHO]: {
+    finalizeOnL1: [opStackFinalizer],
+    finalizeOnL2: [],
```
bmzig, line 133: Should you add cctp info here too?

dohaki: good catch! yea you are right

src/common/Constants.ts
```diff
@@ -654,11 +670,27 @@ export const OPSTACK_CONTRACT_OVERRIDES = {
...
+  [CHAIN_IDs.DOCTOR_WHO]: {
+    l1: {
+      AddressManager: "0x8098f676033a377b9defe302e9fe6877cd63d575",
+      L1CrossDomainMessenger: "0x9A3D64E386C18Cb1d6d5179a9596A4B5736e98A6",
+      L1StandardBridge: CONTRACT_ADDRESSES[CHAIN_IDs.MAINNET].ovmStandardBridge_130.address,
+      StateCommitmentChain: ZERO_ADDRESS,
+      CanonicalTransactionChain: ZERO_ADDRESS,
+      BondManager: ZERO_ADDRESS,
+      OptimismPortal: "0x0bd48f6B86a26D3a217d0Fa6FfE2B491B956A7a2",
+      L2OutputOracle: ZERO_ADDRESS,
```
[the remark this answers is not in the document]
dohaki: Do you know where I can find this address? Is it set somewhere in their contracts?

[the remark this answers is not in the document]
dohaki: ah ok got it

src/common/Constants.ts
```diff
@@ -71,6 +72,7 @@ export const MIN_DEPOSIT_CONFIRMATIONS: { [threshold: number | string]: { [chain
     [CHAIN_IDs.ARBITRUM]: ORBIT_MIN_DEPOSIT_CONFIRMATIONS,
     [CHAIN_IDs.BASE]: OP_STACK_MIN_DEPOSIT_CONFIRMATIONS,
     [CHAIN_IDs.BLAST]: OP_STACK_MIN_DEPOSIT_CONFIRMATIONS,
+    [CHAIN_IDs.DOCTOR_WHO]: OP_STACK_MIN_DEPOSIT_CONFIRMATIONS,
```
bmzig, line 75: I thought I remembered hearing that this chain does not have standard op-stack block times (i.e. not 2 secs/block). Am I wrong?

bmzig (approved):
Addresses check out

bmzig (approved):

pxrl (approved):
LGTM

PR #2051, opened 2025-02-06
fix(Relayer): Identify correct fills when computing commitments

nicholaspai:
The relayer calls SpokePoolClient.getFillForDeposit to find a fill to determine if its committed capital to filling deposits from an origin chain. This function however can match with an invalid fill.


nicholaspai:
This will require https://github.com/across-protocol/sdk/pull/869 to work

PR #2061, opened 2025-02-07
fix(relayer): Ensure argument keys are sorted before hashing

pxrl:
Unclear whether this has actually caused an issue, but implementation details at the upstream RPC provider may lead to undetected duplicate events being passed into the relayer.


nicholaspai (approved):

pxrl:
Marked as `do not merge` for now; this current iteration doesn't work well with fill events where the event contains nested objects.

bmzig:
To be fair, we probably don't need this anymore, right? Do we still want it only if there are weird RPC quirks in the future?

pxrl:
> To be fair, we probably don't need this anymore, right? Do we still want it only if there are weird RPC quirks in the future?

Yeah..I don't think we've been impacted by this, but in general it's probably a good idea to be consistent in how these events are handled, so I'd like to apply sorting.

pxrl:
@bmzig wdyt about this change?

PR #2063, opened 2025-02-07
feat(relayer): Rate-limit invalid fill logs

pxrl:


src/relayer/Relayer.ts
```diff
@@ -331,7 +339,11 @@ export class Relayer {
 
     // Skip deposits that contain invalid fills from the same relayer. This prevents potential corrupted data from
     // making the same relayer fill a deposit multiple times.
-    if (!acceptInvalidFills && invalidFills.some((fill) => fill.relayer === this.relayerAddress)) {
+    if (
+      !acceptInvalidFills &&
+      this.maintenanceIntervalOpen() &&
```
nicholaspai, line 344: +1 I think you should add this to every single log in this function `filterDeposit`. Thoughts? This way GCP won't get clogged up with deposits that fail any of these checks

nicholaspai (commented):
one suggestion, my intuition could be wrong though

src/relayer/Relayer.ts
pxrl: I think we've effectively rate-limited these by adding them to the set of ignored deposits, so we only log about them once per minute (see https://github.com/across-protocol/relayer/commit/a99b25a33822171a2754787fbcffd8bf1c5a5e91). I've re-worked this PR to do the same - it's a much simpler approach.

nicholaspai (approved):

PR #2059, opened 2025-02-07
Revert "hotfix: Temporarily check legacy executed refund root event"

nicholaspai:
Reverts across-protocol/relayer#2058

PR #2068, opened 2025-02-07
chore(relayer): Skip queries for V3 event types

pxrl:
The fast relayer doesn't need to care about these anymore. Most other bots still do though.

### DIFF
--- a/src/libexec/RelayerSpokePoolIndexer.ts
+++ b/src/libexec/RelayerSpokePoolIndexer.ts
@@ -204,7 +204,7 @@ async function run(argv: string[]): Promise<void> {
   oldestTime ??= latestBlock.timestamp;
 
   // Events to listen for.
-  const events = ["FundsDeposited", "RequestedSpeedUpV3Deposit", "FilledRelay"];
+  const events = ["FundsDeposited", "RequestedSpeedUpDeposit", "FilledRelay"];
   const eventMgr = new EventManager(logger, chainId, quorum);
   const providers = getWSProviders(chainId, quorum);
   let nProviders = providers.length;
--- a/src/libexec/RelayerSpokePoolIndexer.ts
+++ b/src/libexec/RelayerSpokePoolIndexer.ts
@@ -195,12 +195,7 @@ async function run(argv: string[]): Promise<void> {
   };
 
   if (latestBlock.number > startBlock) {
-    const events = [
-      "FundsDeposited",
-      "FilledRelay",
-      "RelayedRootBundle",
-      "ExecutedRelayerRefundRoot",
-    ];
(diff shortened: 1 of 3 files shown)


nicholaspai (dismissed):

nicholaspai (approved):

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -211,7 +209,7 @@ async function run(argv: string[]): Promise<void> {
   oldestTime ??= latestBlock.timestamp;
 
   // Events to listen for.
-  const events = ["FundsDeposited", "V3FundsDeposited", "RequestedSpeedUpV3Deposit", "FilledRelay", "FilledV3Relay"];
+  const events = ["FundsDeposited", "RequestedSpeedUpV3Deposit", "FilledRelay"];
```
bmzig, line 212: I think we should add `RequestedSpeedUpDeposit` here now. I put it in this PR but we might as well move it here https://github.com/across-protocol/relayer/pull/2059

nicholaspai: good call can you cherry pick that commit?

bmzig: https://github.com/across-protocol/relayer/pull/2068/commits/7e23062234d115651c4878b89bdac9785238b580

src/libexec/RelayerSpokePoolIndexer.ts
```diff
@@ -195,14 +195,7 @@ async function run(argv: string[]): Promise<void> {
...
   if (latestBlock.number > startBlock) {
-    const events = [
-      "FundsDeposited",
-      "V3FundsDeposited",
-      "FilledRelay",
-      "FilledV3Relay",
-      "RelayedRootBundle",
-      "ExecutedRelayerRefundRoot",
-    ];
+    const events = ["FundsDeposited", "FilledRelay", "RelayedRootBundle", "ExecutedRelayerRefundRoot"];
```
nicholaspai, line 198: @bmzig why do we not also need to search for RequestedSpeedUp here?

nicholaspai (dismissed):

bmzig (dismissed):

src/libexec/RelayerSpokePoolIndexer.ts
pxrl: iirc the intention was to scrape for historical speed ups on startup, but not to subscribe for them. My thinking is that speed-ups are so rare that it doesn't make sense to listen for them, and they shouldn't be time sensitive in the same way as deposits and fills are.

nicholaspai: reading them once a loop seems fine

nicholaspai (dismissed):

bmzig (approved):

PR #2070, opened 2025-02-08
fix(ink): add missing `OptimismPortal2` address

dohaki:


dohaki:
closing in favor of https://github.com/across-protocol/relayer/pull/2071

PR #2083, opened 2025-02-10
fix(relayer): Suppress duplicate event data from external listeners

pxrl:


src/utils/EventUtils.ts
```diff
@@ -173,6 +180,11 @@ export class EventManager {
         }
 
         quorumEvents.push(event);
+
+        // Protect against re-sending this event if it later arrives from another provider.
+        const eventKey = this.hashEvent(event);
```
nicholaspai, line 185: Do you think it'd be safer to wrap any pushes to quorumEvents in an internal function? There you can also check the duplicate events mapping

nicholaspai (dismissed):

src/utils/EventUtils.ts
```diff
@@ -109,6 +110,12 @@ export class EventManager {
     const events = (this.events[event.blockNumber] ??= []);
     const storedEvent = this.findEvent(event);
 
+    const eventKey = this.hashEvent(event);
```
bmzig, line 113: Is there any benefit to checking this right away? i.e. right after the assert? This way we don't allocate an empty event and don't need to find the event if we already sent it to the listener.

src/utils/EventUtils.ts
```diff
@@ -52,6 +52,7 @@ type QuorumEvent = Log & { providers: string[] };
 export class EventManager {
   public readonly chain: string;
   public readonly events: { [blockNumber: number]: QuorumEvent[] } = {};
+  public readonly finalisedEvents: { [eventKey: string]: boolean } = {};
```
mrice32, line 55: nit: prefer [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) (lower mem overhead better perf).

pxrl: https://github.com/across-protocol/relayer/pull/2083/commits/2ea308b17e88a9af01acc62ff94450bac6bd7e67

src/utils/EventUtils.ts
pxrl: I refactored it a bit but it felt awkward passing in the `quorumEvents` array, so I left it where it was. wdyt?

src/utils/EventUtils.ts
```diff
@@ -82,6 +83,28 @@ export class EventManager {
     );
   }
 
+  /**
+   * For a given Log, verify whether it has already been processed.
+   * @param event An Log instance to check.
+   * @returns void
```
pxrl, line 89:
```suggestion
   * @returns True if the event has been processed, else false.
```

bmzig (approved):

dohaki (approved):

PR #2085, opened 2025-02-11
improve: Optimize getRunningBalances computation

nicholaspai:
Improves relayer and monitor performance by loading more data from arweave optimistically, or in the relayer's case, not at all.

While working on https://github.com/across-protocol/sdk/pull/884, I realized that the only usage of the `BundleDataClient'` getLatestPoolRebalanceRoot function is the InventoryClient's getLatestRunningBalances function.

`getLatestPoolRebalanceRoot` is unneccessarily complicated and I actually think we can bypass it entirely in the InventoryClient.

The goal of `getLatestRunningBalances` is to estimate the latest running balances most accurately. This can be done entirely with reconstructing a bundle:
- get the last validated running balance for a chain and token
- get all deposits for that chain and token sent after the last validated bundle end block corresponding to the running balance we computed in the first step
- approximate all refunds taken on the chain and token since the last validated running balance we computed in the first step


src/clients/InventoryClient.ts
```diff
@@ -620,71 +620,91 @@ export class InventoryClient {
...
+        // which chains have too many running balances and therefore should be selected for repayment, so returning
+        // 0 here means this chain will never be selected for repayment as a "slow withdrawal" chain.
+        let lastValidatedBundleEndBlock = 0;
+        let proposedRootBundle: ProposedRootBundle | undefined;
+        if (latestValidatedBundle) {
+          proposedRootBundle = this.hubPoolClient.getLatestFullyExecutedRootBundle(
+            latestValidatedBundle.blockNumber // The ProposeRootBundle event must precede the ExecutedRootBundle
+            // event we grabbed above.
+          );
+          lastValidatedBundleEndBlock = proposedRootBundle.bundleEvaluationBlockNumbers[chainIdIndex].toNumber();
```
pxrl, line 667: Do we need to handle the scenario where a new chain is added and doesn't exist in the `bundleEvaluationBlockNumbers` from the previous bundle?

pxrl (approved):

src/clients/InventoryClient.ts
pxrl: Ah yeah - 100%. Thanks.

PR #2092, opened 2025-02-13
feat: support sepolia testnet

bmzig:
The goal of this PR is to be able to be able to run the dataworker on testnet to refund ourselves when necessary. To do this, we need to clean up the misconfigurations in the hub pool and add support for the dataworker/finalizer for testnet. I thought cleaning up the hub pool was a lost cause: there were a number of deposits which preceeded chain the token activation block, there were multiple different token mappings for weth, and a handful of minor issues (e.g. some pool rebalance routes were set for L2s only, some routes were activated and then deactivated later, etc.). Therefore, I proposed an invalid root bundle to the hub pool which should have "reset" its state (https://sepolia.etherscan.io/tx/0xfc6b72be45215f96e318e62511284c64e7675b7ed55fb74be1a4e36dcac4d680). This essentially means we will not honor any testnet refunds (as we have been doing) up until the time of the proposal. After this reset, we are able to propose valid bundles (e.g. https://sepolia.etherscan.io/tx/0x3cd375ea435197a8b3799a0b914d661c6578218af216aa12f5f259585a1c3c9c) and execute them (e.g. https://sepolia.etherscan.io/tx/0xfcb9e0c3e6ec2b7ae5aa3c66020b407bb105245c7378cffd459460f3d7f95dc2).

We need to support finalizations since otherwise we run the risk of having unexecutable root bundles on L1 due to insufficient funds. (e.g. https://sepolia.etherscan.io/tx/0x018a37ffc55b22ff871bfab794d37f709d0130fe703b2a10a21524f01add103a)

Obviously, since there have been deposits on chains which are now unrefundable without sending an admin bundle, we may need to manually pull those funds back, but I don't think this is immediately necessary since we can use this excess in the spoke pools to simply refund relayers from now on.


bmzig:
The rebalancer should work for all chains except for lens (since there is no zkstack adapter made yet). Unichain is not set up since we are pending a contract bump. The arbitrum finalizer will not work until the sdk is bumped to v4.1.12.

This should unblock some work for supporting lens, i.e. we can now reasonably make the bridge adapters on testnet as well as finalizers for spoke pool withdrawals (which we can also now initiate with a working dataworker).

src/finalizer/utils/opStack.ts
```diff
@@ -78,7 +78,14 @@ export async function opStackFinalizer(
       (e) =>
         // USDC withdrawals for Base and Optimism should be finalized via the CCTP Finalizer.
         !compareAddressesSimple(e.l2TokenAddress, TOKEN_SYMBOLS_MAP["USDC"].addresses[chainId]) ||
-        !(chainId === CHAIN_IDs.BASE || chainId === CHAIN_IDs.OPTIMISM || chainId === CHAIN_IDs.UNICHAIN)
+        !(
```
nicholaspai, line 81: Can you use this function instead? https://github.com/across-protocol/relayer/blob/master/src/utils/CCTPUtils.ts#L65

nicholaspai (approved):
LGTM! Just One suggestion

src/finalizer/utils/opStack.ts
bmzig: https://github.com/across-protocol/relayer/pull/2092/commits/949133e9f20d6a8ff81b9f21b7a23a12c3569d02

nicholaspai (approved):

PR #2100, opened 2025-02-14
feat: add zkStack adapters

bmzig:
Needed to interface with ZkStack chains. The weth bridges will utilize the new functionality of the atomic depositor. The adapters should be functionally equivalent to the contract versions.

### DIFF
--- a/src/adapter/bridges/ZKStackBridge.ts
+++ b/src/adapter/bridges/ZKStackBridge.ts
@@ -1,0 +1,130 @@
+import { Contract, BigNumber, Signer, EventSearchConfig, Provider, ethers, TOKEN_SYMBOLS_MAP } from "../../utils";
+import { CONTRACT_ADDRESSES } from "../../common";
+import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./BaseBridgeAdapter";
+import * as zksync from "zksync-ethers";
+import { gasPriceOracle } from "@across-protocol/sdk";
+import { PUBLIC_NETWORKS } from "@across-protocol/constants";
+
+/* For both the canonical bridge (this bridge) and the ZkSync Weth
+ * bridge, we need to assume that the l1 and l2 signers contain
+ * associated providers, since we need to get information about
+ * addresses and gas prices (this is also why `constructL1toL2Txn`
+ * is an async fn).
+ */
+export class ZKStackBridge extends BaseBridgeAdapter {
+  readonly gasPerPubdataLimit = zksync.utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT;
+  readonly l2GasLimit = BigNumber.from(2_000_000);
+  readonly sharedBridgeAddress;
+
+  constructor(
+    l2chainId: number,
+    hubChainId: number,
+    l1Signer: Signer,
+    l2SignerOrProvider: Signer | Provider,
+    _l1Token: string
+  ) {
+    // Lint Appeasement
+    _l1Token;
+    const sharedBridgeAddress = CONTRACT_ADDRESSES[hubChainId][`zkStackSharedBridge_${l2chainId}`].address;
+    super(l2chainId, hubChainId, l1Signer, l2SignerOrProvider, [sharedBridgeAddress]);
+
+    const nativeToken = PUBLIC_NETWORKS[l2chainId].nativeToken;
+    // Only set nonstandard gas tokens.
+    if (nativeToken !== "ETH") {
+      this.gasToken = TOKEN_SYMBOLS_MAP[nativeToken].addresses[hubChainId];
+    }
+
+    const { address: l1Address, abi: l1Abi } = CONTRACT_ADDRESSES[hubChainId].zkStackBridgeHub;
+    this.l1Bridge = new Contract(l1Address, l1Abi, l1Signer);
+
+    this.sharedBridgeAddress = sharedBridgeAddress;
+
+    const { address: l2Address, abi: l2Abi } = CONTRACT_ADDRESSES[l2chainId].zkSyncDefaultErc20Bridge;
+    this.l2Bridge = new Contract(l2Address, l2Abi, l2SignerOrProvider);
+  }
+
+  async constructL1ToL2Txn(
+    toAddress: string,
+    l1Token: string,
+    l2Token: string,
+    amount: BigNumber
+  ): Promise<BridgeTransactionDetails> {
+    // The zkStack bridges need to know information relating to the l2 gas price bid.
+    const txBaseCost = await this._txBaseCost();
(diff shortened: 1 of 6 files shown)


src/common/abi/ZkStackBridgeHub.json
bmzig, line 1: I will shorten this a lot since most of these are functions/events we don't need. I want to be able to have a full e2e test first though so that I know which events/functions we _do_ need.

bmzig:
We can currently only test with weth.
Weth: [init](https://sepolia.etherscan.io/tx/0x887dc0111d834fdff11d33e04da14e572fa407a40065ea819053751e4370af71) [recv](https://block-explorer.testnet.lens.dev/tx/0x37f54f742c185b801846647c9df7b25067437f3b06c1dc92df9e039d959468c5)
Transfer tracking works as expected.

We can also rebalance with grass
Grass: [init](https://sepolia.etherscan.io/tx/0xdd739d3b6104bb7e3d37b1be20c3d5df74210120efcf1934813fe3e7b3fa13c2) [recv](https://block-explorer.testnet.lens.dev/tx/0xc22c93130742fb5701db963a738d1bc9ff5eb15136ca19b6d00d0fb03b1d8c54)
but transfer tracking is slightly off. If we wanted to support it, We would have to either query more/different events in the case of bridging the custom gas token (which would just add complexity/RPC queries). I am holding off on doing this since, like aleph zero, I don't think this is a requirement, but can support it if necessary.

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,236 @@
...
+  readonly tokenVault;
+
+  constructor(
+    l2chainId: number,
+    hubChainId: number,
+    l1Signer: Signer,
+    l2SignerOrProvider: Signer | Provider,
+    _l1Token: string
+  ) {
+    // Lint Appeasement
```
nicholaspai, line 44: Is this necessary? I find prefixing the param with an underscore does the trick.

nicholaspai (dismissed):
Seems fine to me. Hard to verify this works but I assume you've been integration testing

src/adapter/bridges/ZKStackBridge.ts
bmzig: Yeah probably good to get into the habit of not doing this. Fixed here. https://github.com/across-protocol/relayer/pull/2100/commits/046ee3164b22a1ea0359f9ce9899dff48ac4b898

dohaki (dismissed):
This looks good 👍 Left only a few nits

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,235 @@
...
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // If the fromAddress is the hub pool then ignore the query. This is because for calculating cross-chain
+    // transfers, we query both the hub pool outstanding transfers *and* the spoke pool outstanding transfers,
+    // meaning that querying this function for the hub pool as well would effectively double count the outstanding transfer amount.
+    if (compareAddressesSimple(fromAddress, this.hubPool.address)) {
+      return Promise.resolve({});
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/2100/commits/430c189c04ab50578c0c30e3e9bf44fb900e9a69

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  async queryL2BridgeFinalizationEvents(
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // Ignore hub pool queries for the same reason as above.
+    if (compareAddressesSimple(fromAddress, this.hubPool.address)) {
+      return Promise.resolve({});
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/2100/commits/430c189c04ab50578c0c30e3e9bf44fb900e9a69

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,235 @@
...
+ * is an async fn).
+ */
+export class ZKStackBridge extends BaseBridgeAdapter {
+  readonly gasPerPubdataLimit = zksync.utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT;
+  readonly l2GasLimit = BigNumber.from(2_000_000);
+  readonly sharedBridgeAddress;
+  readonly l2GasToken;
+  readonly l2WrappedGasToken;
+  readonly hubPool;
+  readonly tokenVault;
```
pxrl, line 35: ooc, what types are each of these? Is tsc able to derive that or are they all treated as `any` ?

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,235 @@
...
+    // If the fromAddress is the hub pool then ignore the query. This is because for calculating cross-chain
+    // transfers, we query both the hub pool outstanding transfers *and* the spoke pool outstanding transfers,
+    // meaning that querying this function for the hub pool as well would effectively double count the outstanding transfer amount.
+    if (compareAddressesSimple(fromAddress, this.hubPool.address)) {
+      return {};
+    }
+
+    // Logic changes based on whether we are sending tokens to the spoke pool or to an EOA.
+    const isL2Contract = await this._isContract(toAddress, this.getL2Bridge().provider!);
+    let processedEvents;
```
pxrl, line 143: Does tsc derive the type of this OK? Ideally it might be able to do that based on evaluating both of the possible subsequent codepaths and concludes that it can only be one type.

pxrl (commented):
Only two minor queries - otherwise this looks good. Has it all been tested on mainnet?

src/adapter/bridges/ZKStackBridge.ts
bmzig: They are all contract types for the most part. I made them verbose here https://github.com/across-protocol/relayer/pull/2100/commits/8a666e12fde06a62417f025bfce90c44a9bea9d5. I assumed tsc would be able to infer the type.

src/adapter/bridges/ZKStackBridge.ts
bmzig: Yeah, I'm pretty sure for these types, it is coerced into the proper type due to the return value of the function. I do the same thing in the other adapters.

bmzig:
> Only two minor queries - otherwise this looks good. Has it all been tested on mainnet?

This is only testnet changes, so nothing should happen to the mainnet bots.

dohaki (approved):
LGTM

PR #2103, opened 2025-02-17
improve(relayer): Ensure relayer yields execution between loops

pxrl:
If the relayer's execution loop is taking long enough that it does not sleep between loops, a transaction that is pending submission may appear to take a long time because its post-submission task cannot be scheduled.

This change enforces a minimum delay of 0.1 seconds when there are transactions pending submission in order to ensure the submission tasks are not CPU-starved.


dohaki (approved):
LGTM! def worth trying

PR #2108, opened 2025-02-18
improve(TransactionClient): Track nonces on submitted transactions

pxrl:
Track nonces internally within the TransactionClient as an attempt to mitigate the high number of nonce collisions that are observed when a chain is subject to a lot of fills.


james-a-morris (commented):
Fair call but failing tests

dohaki (approved):
LGTM

PR #2109, opened 2025-02-19
feat(InventoryClient): Support withdrawing from L2 to L1

nicholaspai:
Adds a feature that we've executed manually via the [withdrawFromArbOrbit](https://github.com/across-protocol/relayer/blob/master/scripts/withdrawFromArbitrumOrbit.ts) and [withdrawFromOpStack](https://github.com/across-protocol/relayer/blob/master/scripts/withdrawFromOpStack.ts) scripts to the InventoryClient.

Re-uses existing `targetOverageBuffer` and `targetPct` variables in the `TokenConfig` and adds a new `withdrawPeriodSeconds` variable.

## The Algorithm:

If `currentAllocationPct >= (targetPct * targetOverageBuffer)` withdraw `currentAllocationPct - targetPct` of the `cumulativeBalance` from L2 to L1.

However, don't do the above if the total withdrawal volume over the past `withdrawPeriodSeconds` exceeds `(targetOverageBuffer - targetPct) * cumulativeBalance`


src/relayer/Relayer.ts
```diff
@@ -169,9 +169,12 @@ export class Relayer {
       tokenClient.clearTokenData();
       await tokenClient.update();
       await inventoryClient.rebalanceInventoryIfNeeded();
+      // After computing allocation %'s and rebalancing from L1 to L2, send any excess balances back to L1. We could
+      // re-update token balances here if we think the rebalance from L1 to L2 function took a long time to run.
+      await inventoryClient.withdrawExcessBalances();
     }
 
-    // Unwrap WETH after filling deposits, but before rebalancing.
```
nicholaspai, line 174: This is an old comment, I think this `unwrapWeth` is still ok to be called here. @pxrl wdyt?

src/relayer/RelayerConfig.ts
```diff
@@ -80,6 +81,7 @@ export class RelayerConfig extends CommonConfig {
       RELAYER_INVENTORY_CONFIG,
       RELAYER_TOKENS,
       SEND_RELAYS,
+      SIM_MODE,
```
nicholaspai, line 84: We should start adding this `simModeEnabled` flag throughout all InventoryClient functions. Currently this is used to simulate wrapping L2 eth, sending L1 to L2 and L2 to L1 rebalances. Its not currently used when setting spoke pool approvals and unwrapping L2 weth

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1109,185 @@ export class InventoryClient {
...
+
+        if (chainIsOPStack(chainId)) {
+          const ovmStandardBridgeObj = CONTRACT_ADDRESSES[chainId].ovmStandardBridge;
+          assert(ovmStandardBridgeObj, `ovmStandardBridge for ${chainId} not found in CONTRACT_ADDRESSES`);
+          const ovmStandardBridge = new Contract(
+            ovmStandardBridgeObj.address,
+            ovmStandardBridgeObj.abi,
+            this._getL2Signer(chainId)
+          );
+          if (l1TokenInfo.symbol === "ETH") {
```
bmzig, line 1176: Do you also need to account for "WETH" here? Or does the hub pool client name all weth tokens as "ETH".

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1109,185 @@ export class InventoryClient {
...
+          }
+        } else if (chainIsOrbit(chainId)) {
+          const nativeTokenSymbol = getNativeTokenSymbol(chainId);
+          if (l1TokenInfo.symbol === nativeTokenSymbol) {
+            this.logger.warn({
+              at: "InventoryClient",
+              message: "Native token withdrawals for orbit chain is not supported",
+            });
+          } else {
+            const arbErc20GatewayObj = CONTRACT_ADDRESSES[chainId].erc20Gateway;
```
bmzig, line 1239: I'm fairly confident you will need to use the [L2 gateway router](https://arbiscan.io/address/0x5288c571Fd7aD117beA99bF60FE0846C4E84F933) and not the gateway here, since, for example, the [erc20 gateway](https://arbiscan.io/address/0x09e9222E96E7B4AE2a407B98d48e330053351EEe) is different from the [weth gateway](https://arbiscan.io/address/0x6c411aD3E74De3E7Bd422b94A27770f5B86C623B), which is different from other, custom gateways.

src/clients/InventoryClient.ts
nicholaspai: It's not ideal but ETH is correct because of how [this function[(https://github.com/across-protocol/sdk/blob/56badd383ec9bd53b1438b20aa324761589dbc9a/src/utils/TokenUtils.ts#L115) is implemented. I matches on the first token object that matches, so ETH before WETH

src/clients/InventoryClient.ts
nicholaspai: I think that's right. I was copying code from the `withdrawFromArbitrumOrbit` script but I think we should change all erc20Gateway mentions to the router instead. We get away with this currently because (1) we only use this address for AlephZero in the script and (2) the erc20 gateway for Aleph zero is used for most tokens on that chain specifically

https://github.com/across-protocol/relayer/pull/2109/commits/2c2225f4d102d5cb7cad7679e3c58456ea9b4f7a

nicholaspai:
Since we're changing the erc20gateway that we call to withdraw tokens from AlephZero to Mainnet, we should not merge this in until all pending l2 to l1 withdrawals from AlephZero to mainnet are flushed

src/clients/InventoryClient.ts
nicholaspai: Here's the latest fix to use the gateway router throughout the repo: https://github.com/across-protocol/relayer/pull/2109/commits/d468e91a9ca9207e26251dfcc51cabdb176342dd

src/finalizer/utils/arbStack.ts
```diff
@@ -135,24 +137,20 @@ export async function arbStackFinalizer(
     ? JSON.parse(process.env.FINALIZER_WITHDRAWAL_TO_ADDRESSES).map((address) => ethers.utils.getAddress(address))
     : [];
   if (getOrbitNetwork(chainId) !== undefined && withdrawalToAddresses.length > 0) {
-    // ERC20 withdrawals emit events in the erc20Gateway.
+    // ERC20 withdrawals emit events in the erc20GatewayRouter.
     // Native token withdrawals emit events in the ArbSys contract.
     const l2ArbSys = CONTRACT_ADDRESSES[chainId].arbSys;
     const arbSys = new Contract(l2ArbSys.address, l2ArbSys.abi, spokePoolClient.spokePool.provider);
-    const l2Erc20Gateway = CONTRACT_ADDRESSES[chainId].erc20Gateway;
```
bmzig, line 142: Just to be sure, this was only working before since this was only used for aleph zero?

src/interfaces/InventoryManagement.ts
```diff
@@ -7,6 +7,12 @@ export type TokenBalanceConfig = {
   thresholdPct: BigNumber; // Threshold, below which, we will execute a rebalance.
   unwrapWethThreshold?: BigNumber; // Threshold for ETH to trigger WETH unwrapping to maintain ETH balance.
   unwrapWethTarget?: BigNumber; // Amount of WETH to unwrap to refill ETH. Unused if unwrapWethThreshold is undefined.
+  excessThresholdBalance?: BigNumber; // Threshold for excess token balance to trigger a slow transfer from Spoke to Hub.
+  // Typically set on Lite chains where the relayer filling deposits from the chain has no choice but to take
+  // repayment on that chain and subsequently builds up a large balance over time that it otherwise has no ability
+  // to rebalance.
+  excessTargetBalance?: BigNumber; // When the balance on a chain exceeds the excessThresholdBalance, withdraw
```
pxrl, line 14: ooc, how do you forsee this being set? Somewhere below `targetPct` ?

src/relayer/Relayer.ts
pxrl: I guess the thinking of the original comment was that if we need to unwrap, we should do that _before_ rebalancing so as to avoid rebalancing WETH and then having insufficient remaining to top up our ETH balance. In the worst-case scenario that might cause us to run out of ETH mainnet and deadlock the bot.

In practice I think it's pretty unlikely because we'd have to run with astronomically low unwrap thresholds. But I am wondering if we should relocate the `unwrapWeth()` call to be above `rebalanceInventoryIfNeeded()`.

src/relayer/RelayerConfig.ts
pxrl: What's the motivation for adding a new flag vs. using the existing `sendingRebalancesEnabled`, which is currently what toggles "inventory client sim mode" ? We do already have a lot of flags that are used to toggle various things on and off - it's starting to feel a bit like option soup and it's difficult to know exactly what each toggle does.

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1109,185 @@ export class InventoryClient {
     }
   }
 
+  async withdrawExcessBalances(): Promise<void> {
```
pxrl, line 1112: I think we need an additional rate-limit on this functionality, such that we don't just endlessly rebalance tokens via slow bridges. Some of the lite chains would otherwise tend to suck a lot of capital into those bridges if they're left unchecked.

src/interfaces/InventoryManagement.ts
nicholaspai: I foresee this being set as a fixed number like we do today. Do you think this should be specified as a percentage?

src/relayer/RelayerConfig.ts
nicholaspai: The issue is that we use `sendingRebalances` [here](https://github.com/across-protocol/relayer/blob/410871547c8e8f9257834a02df6f3438689a350f/src/relayer/Relayer.ts#L167) to decide whether to enter into the rebalance loop and then also [here](https://github.com/across-protocol/relayer/blob/410871547c8e8f9257834a02df6f3438689a350f/src/relayer/RelayerClientHelper.ts#L176) to activate "sim" mode in the inventory client. I think the best way to consoliate is to remove the sendingRebalances check in the first link above and then treat SEND_REBALANCES as a sim-mode toggle

src/relayer/Relayer.ts
nicholaspai: WDYT of this: 84ea4043d9fb0ba167405c40ae109929d1ecd1c6

src/relayer/RelayerConfig.ts
nicholaspai: 72b0730b727ce5e93072447cf713dbbdaec54599

.gitignore
```diff
@@ -7,6 +7,7 @@ coverage
 coverage.json
 typechain*
 dump.rdb*
+inventoryConfig.json
```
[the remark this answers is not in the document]
nicholaspai: Good point, will revert this change and use an external file

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1109,185 @@ export class InventoryClient {
...
+    // pull tokens from the msg.sender.
+    const multicallerClient = new MultiCallerClient(this.logger);
+    Object.keys(withdrawalsRequired).forEach((_chainId) => {
+      const chainId = Number(_chainId);
+      withdrawalsRequired[chainId].forEach((withdrawal) => {
+        const { l2Token, amountToWithdraw } = withdrawal;
+        const l1TokenInfo = getL1TokenInfo(l2Token, chainId);
+        const formatter = createFormatFunction(2, 4, false, l1TokenInfo.decimals);
+
+        if (chainIsOPStack(chainId)) {
```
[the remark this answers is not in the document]
bmzig: I think that's a good idea since we already define these bridge adapters for each chain, but it would definitely increase the diff, since we would need part of this logic here to be in a new function in the adapter manager, we would need a new function in the `BaseChainAdapter` that the adapter manager calls to actually execute the rebalances, and then we would need to have, for each bridge, the logic to construct the withdrawal transaction (so that the `BaseChainAdapter` knows what to call). Maybe we should do this in a follow-up instead? The benefit of doing this is we can also probably easily support transfer tracking for withdrawals (and maybe this would help with the finalizers in the future).

src/clients/InventoryClient.ts
nicholaspai: but in this scenario, would we not want to keep withdrawing that 150k usdt into the bridge? Otherwise its stuck on aleph zero..?

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
nicholaspai: good call: 9f8cffff63595afa15fbbbf3ce1d5418ee335716

src/clients/InventoryClient.ts
nicholaspai: here's a rate limit implementation: https://github.com/across-protocol/relayer/pull/2109/commits/cc476aebd9086983c1b48d0317ca7db34b2efc2f

it does not attempt to figure out which finaliazations are pending, mostly for complexity reasons, but I think this is still very useful if we limit our total volume over some period time

src/clients/bridges/AdapterManager.ts
```diff
@@ -59,6 +61,26 @@ export class AdapterManager {
         }) ?? []
       );
     };
+    const constructL2Bridges = (chainId: number) => {
+      if (chainId === hubChainId) {
+        return {};
+      }
+      const l2Signer = spokePoolClients[chainId].spokePool.signer;
+      return Object.fromEntries(
+        SUPPORTED_TOKENS[chainId]
```
bmzig, line 70: I think you will need another constant besides `SUPPORTED_TOKENS` here, since they contain L1 token symbols. So e.g. if we want to withdraw USDC from Zora (in our `TOKEN_SYMBOLS_MAP`, it is USDzC), then this won't work since we won't resolve an address for USDC.

src/adapter/l2Bridges/ArbitrumOrbitNativeTokenBridge.ts
bmzig, line 1: Is this needed in general? Why under your current setup, why wouldn't we just withdraw weth using the arbitrum orbit bridge?

src/clients/bridges/AdapterManager.ts
bmzig, line 70: Another thing (which isn't immediately necessary since we are only doing lite chains) is that this won't work with CCTP chains since we will try to withdraw native USDC over the canonical bridge. Likely not actionable for this PR though.

src/adapter/l2Bridges/ArbitrumOrbitBridge.ts
```diff
@@ -0,0 +1,75 @@
...
+        fromAddress // from
+      ),
+      eventConfig
+    );
+    const withdrawalAmount = withdrawalEvents.reduce((totalAmount, event) => {
+      if (event.args.l1Token === l1TokenInfo.address) {
+        return totalAmount.add(event.args._amount);
+      }
+      return totalAmount;
+    }, BigNumber.from(0));
```
bmzig, line 72: nit: might be better to use bnZero for these and all other instances (in the opStack bridge/weth bridge)

bmzig (dismissed):
Would be good to try it out with a small amount, but in general it looks fine to me.

src/adapter/l2Bridges/OpStackWethBridge.ts
pxrl, line 1: I think we probably need a mapping for some chains to opt out of bridging USDC with this adapter. I don't think we risk deleting funds since approvals should be a showstopper, but it'd create noise in the bots if we activated it by accident.

src/clients/InventoryClient.ts
pxrl: >but in this scenario, would we not want to keep withdrawing that 150k usdt into the bridge? Otherwise its stuck on aleph zero..?

That's basically the intention; otherwise the relayer could in theory quickly end up with a large percentage of its tokens in the slow bridge.

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1107,125 @@ export class InventoryClient {
...
+                  }
+                );
+                return;
+              }
+
+              // For now, we do not support partial desired withdrawals where the desired amount would push the
+              // total withdrawal volume over the limit. This is because in practice this could lead to withdrawing
+              // very small amounts over the bridge, unless we set a minimum withdrawal amount. This means that we'll
+              // delay occasionally some withdrawals until the total withdrawal volume goes down, but this should be
+              // relatively infrequent and we can change this logic in the future.
```
pxrl, line 1175: wdyt about a imposing a minimum of something like 20k USD, and then permitting partial withdrawals? Then we could at least get a partial withdrawal in and open up some headroom.

src/adapter/l2Bridges/ArbitrumOrbitNativeTokenBridge.ts
nicholaspai: We don't need this but I included it for future use cases since I already had the code for it in the `withdrawArbitrumOrbitStack.ts` script. It's not mapped to any tokens yet in `Constants`

src/clients/bridges/AdapterManager.ts
nicholaspai: hm good point. I think we'd want to leave this up to the `CUSTOM_L2_BRIDGE` dictionary in Constants to sort out. We can control that dictionary and only add a CCTP bridge for chains with native L2 USDC

src/adapter/l2Bridges/OpStackWethBridge.ts
nicholaspai: I can see maybe doing this in the canonical OpStack and ArbitrumOrbit bridge, but this class is named WethBridge and I doubt we'll set it incorrectly in the CUSTOM_L2_BRIDGES mapping

nicholaspai: in fact, I don't really see how to support this unless we make each `adapter/l2Bridge` specify the token symbols it supports. I thought we got away from that by making use of the CUSTOM_BRIDGES mapping instead

src/clients/InventoryClient.ts
nicholaspai: That's just more configuration and I wanted to remove that, I guess I'll add that since it seems useful

src/adapter/l2Bridges/ArbitrumOrbitBridge.ts
```diff
@@ -0,0 +1,76 @@
...
+    return [withdrawTxn];
+  }
+
+  async getL2WithdrawalAmount(
+    eventConfig: EventSearchConfig,
+    fromAddress: string,
+    l2Token: string
+  ): Promise<BigNumber> {
+    const l1TokenInfo = getL1TokenInfo(l2Token, this.l2chainId);
+    const erc20GatewayForToken = await this.l2Bridge.getGateway(l1TokenInfo.address);
```
bmzig, line 58: I believe this is the same gateway that we have stored in constants. It might be nicer if we just get it from there by doing something like `CUSTOM_ARBITRUM_GATEWAYS[this.l2chainId][l1TokenInfo.address] ?? DEFAULT ARBITRUM_GATEWAY[this.l2chainId] instead of taking the extra RPC call.

bmzig (approved):

src/adapter/l2Bridges/ArbitrumOrbitBridge.ts
nicholaspai: The only issue with this is not being able to dynamically respond to the chain switching gateways. Do we have any reason to assume these won't change ?

nicholaspai: Another option I thought of was querying the TransferRouted event to get the gateways but that's more eth_getLogs calls

bmzig: Hmm good points. If this is only used for event tracking though, right? So worst case we withdraw more tokens than the we should since we are above threshold but there are untracked tokens in the bridge.

If this extra RPC call is only incurred in the rebalancer and not the relayer, then yeah, we probably don't have a need to change it.

src/adapter/l2Bridges/OpStackWethBridge.ts
pxrl: > I can see maybe doing this in the canonical OpStack and ArbitrumOrbit bridge, but this class is named WethBridge and I doubt we'll set it incorrectly in the CUSTOM_L2_BRIDGES mapping

Yeah; I tagged this comment on the wrong file - it was intended for the OP L2 ERC20 bridge.

`CUSTOM_L2_BRIDGES` is OK with me :+1:

src/common/Constants.ts
```diff
@@ -489,6 +506,24 @@ export const CUSTOM_BRIDGE: {
...
+export const CUSTOM_L2_BRIDGE: {
+  [chainId: number]: {
+    [tokenAddress: string]: {
+      new (l2chainId: number, hubChainId: number, l2Signer: Signer): BaseL2BridgeAdapter;
+    };
+  };
+} = {
+  [CHAIN_IDs.LISK]: {
+    [TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET]]: L2OpStackWethBridge,
+  },
```
pxrl, line 518: How about USDT on Lisk? That one regularly taps out.

src/interfaces/InventoryManagement.ts
pxrl: By this I mean, what would the value be in relation to `targetPct` ? Do we expect to rebalance somewhere below `targetPct`, or above? It seems like there's an idea here, because otherwise we'd rebalance exactly to `targetPct`, right?

src/relayer/Relayer.ts
pxrl: I'd probably prefer to avoid 3x tokenClient updates as part of the maintenance loop. These update calls involve an RPC call on each chain, so if one of them hangs or fails we'll delay the bot. If we can't avoid tokenClient updates then I'm in favour of leaving it as it was, and just correcting the comment.

src/relayer/RelayerConfig.ts
pxrl: SGTM!

src/clients/InventoryClient.ts
pxrl: The comment identified the desire to avoid very low-value transfers, which is definitely a good thing to protect against. I still like the idea of supporting a partial withdrawal because it could allow us to rebalance a lesser amount more regularly, instead of having to do a large rebalance every week. The latter gives the bridge a hard on/off property and makes it harder for us to support smaller transfers. I suspect it would become more noticeable for users when it's off for days on end.

nicholaspai: Partial withdrawals would only give one us one more rebalance for that week so I don't think it would be that noticeable in this situation where the chain is very overbalanced

src/adapter/l2Bridges/ArbitrumOrbitBridge.ts
nicholaspai: yeah I only see the extra RPC call being incurred in the rebalancer so I don't think its worth the optimization

src/common/Constants.ts
nicholaspai: that will use the default bridge

src/relayer/Relayer.ts
nicholaspai: Reverted the change. Added the excess withdraw run right after teh rebalance inventory. No need to update token client since it shouldn't be affected by the L1 to L2 rebalance direction. If an L2 balance will be higher post-l1-l2-rebalance, then it wouldn't be an excess that we'd want to withdraw. If the L2 balance is already at an excess, then it won't change post-l1-l2-rebalance

bmzig (approved):

src/finalizer/utils/arbStack.ts
```diff
@@ -135,24 +137,20 @@ export async function arbStackFinalizer(
...
-    const arbitrumGateway = new Contract(
-      l2Erc20Gateway.address,
-      l2Erc20Gateway.abi,
+    const l2Erc20GatewayRouter = CONTRACT_ADDRESSES[chainId].erc20GatewayRouter;
+    const gatewayRouter = new Contract(
+      l2Erc20GatewayRouter.address,
+      l2Erc20GatewayRouter.abi,
       spokePoolClient.spokePool.provider
     );
-    // TODO: For this to work for ArbitrumOrbit, we need to first query ERC20GatewayRouter.getGateway(l2Token) to
```
nicholaspai, line 148: This code is changing because currently we assume that all L2 tokens use the same `l2Erc20Gateway` address which is not always true--we've just gotten away with it for AlephZero so far (the only chain where we could enter into this code section, because we never send manual withdrawals from Arbitrum)

src/adapter/BaseChainAdapter.ts
```diff
@@ -167,6 +177,52 @@ export class BaseChainAdapter {
     this.log("Approved whitelisted tokens! 💰", { mrkdwn }, "info");
   }
 
+  async withdrawTokenFromL2(address: string, l2Token: string, amount: BigNumber, simMode: boolean): Promise<string[]> {
```
nicholaspai, line 180: These two functions are the new ones that let us (1) withdraw from L2 to L1 and (2) track pending withdrawal amounts from L2 to L1

src/common/ContractAddresses.ts
```diff
@@ -395,9 +392,9 @@ export const CONTRACT_ADDRESSES: {
     },
   },
   [CHAIN_IDs.ALEPH_ZERO]: {
-    erc20Gateway: {
-      address: "0x2A5a79061b723BBF453ef7E07c583C750AFb9BD6",
-      abi: ARBITRUM_ERC20_GATEWAY_L2_ABI,
+    erc20GatewayRouter: {
```
nicholaspai, line 395: Currently, we implicitly assume that all L2 tokens on AlephZero use the same ERC20Gateway, which is practically true but does not have to be true. I think its safer to query the GatewayRouter to find the correct Gateway

src/interfaces/InventoryManagement.ts
```diff
@@ -7,6 +7,8 @@ export type TokenBalanceConfig = {
   thresholdPct: BigNumber; // Threshold, below which, we will execute a rebalance.
   unwrapWethThreshold?: BigNumber; // Threshold for ETH to trigger WETH unwrapping to maintain ETH balance.
   unwrapWethTarget?: BigNumber; // Amount of WETH to unwrap to refill ETH. Unused if unwrapWethThreshold is undefined.
+  withdrawExcessPeriod?: number; // Period in seconds over which to withdraw any excess balance over the
```
nicholaspai, line 10: This is the only variable that will trigger evaluating L2 to L1 withdrawals. For example, we'd set this value only for Lite chains where we would sometimes want to withdraw from L2 to L1.

Alternatively, I considered using the `targetOverageBuffer` as a trigger to evaluate these L2 to L1 withdrawals since we only set that for Lite chains currently. However, I think its possible that we'd re-use this `targetOverageBuffer` on non Lite chains.

src/clients/InventoryClient.ts
nicholaspai: I changed the code to now always send the desired withdraw amount as long as the pendingWithdrawalAmount < limit. This means we'll go over the limit by one transfer but avoids the case where we hold too many funds on a chain. I think this is a better decision to make to avoid holding L2 funds while limiting the amount stuck in the Slow L2 to L1 bridge. WDYT?

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1106,183 @@ export class InventoryClient {
...
+          // by the target overage buffer.
+          const discountToTargetOverageBuffer = toBNWei("0.95");
+          const targetPctMultiplier = targetOverageBuffer.mul(discountToTargetOverageBuffer).div(this.scalar);
+          assert(
+            targetPctMultiplier.gte(toBNWei("1")),
+            `Target overage buffer multiplied by discount must be >= 1, got ${targetPctMultiplier.toString()}`
+          );
+          const excessWithdrawThresholdPct = targetPct.mul(targetPctMultiplier).div(this.scalar);
+
+          const l1TokenInfo = getL1TokenInfo(l2Token, Number(chainId));
```
pxrl, line 1170: Minor: `chainId` is already a number.

```suggestion
          const l1TokenInfo = getL1TokenInfo(l2Token, chainId);
```

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1106,183 @@ export class InventoryClient {
...
+          const formatter = createFormatFunction(2, 4, false, l1TokenInfo.decimals);
```
pxrl, line 1171: Very minor: I'd probably consider moving this up to line 1119 so it doesn't get repeated for every L2 for a given token.

src/clients/InventoryClient.ts
```diff
@@ -1105,6 +1106,183 @@ export class InventoryClient {
...
+              withdrawalParams: shouldWithdrawExcess
+                ? {
+                    withdrawPct: formatUnits(withdrawPct, 18),
+                    desiredWithdrawalAmount: formatter(desiredWithdrawalAmount),
+                  }
+                : undefined,
+            }
+          );
+
+          if (shouldWithdrawExcess) {
```
pxrl, line 1198: Minor: We can drop a level of indentation if the logic is inverted (i.e. return early if `!shouldWithdrawExcess`).

src/relayer/RelayerConfig.ts
```diff
@@ -185,10 +197,15 @@ export class RelayerConfig extends CommonConfig {
         tokenConfig.targetPct = toBNWei(targetPct).div(100);
         tokenConfig.thresholdPct = toBNWei(thresholdPct).div(100);
 
+        if (isDefined(withdrawExcessPeriod)) {
+          tokenConfig.withdrawExcessPeriod = withdrawExcessPeriod;
+        }
```
pxrl, line 202: Should be OK to just assign this directly.

```suggestion
        tokenConfig.withdrawExcessPeriod = withdrawExcessPeriod;
```

pxrl (approved):
A couple of minor comments, but otherwise LGTM!

src/interfaces/InventoryManagement.ts
pxrl: `targetOverageBuffer´ defaults to 1.5x, so I think it's good to make this opt-in by setting `withdrawalExcessPeriod`.

nicholaspai: ah yes that too, forgot to mention that in my comment but I noticed that

pxrl (approved):

PR #2113, opened 2025-02-20
improve(spokepool): Updates for new events

pxrl:
- Display both messageHash and relayDataHash on deposit.
- Skip relayDataHash on fill (not available due to missing message).

### DIFF
--- a/scripts/spokepool.ts
+++ b/scripts/spokepool.ts
@@ -13,6 +13,7 @@ import {
   BigNumber,
   formatFeePct,
   getDeploymentBlockNumber,
+  getMessageHash,
   getNetworkName,
   getProvider,
   getSigner,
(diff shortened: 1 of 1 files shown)


scripts/spokepool.ts
```diff
@@ -45,14 +46,19 @@ const DEPOSIT_EVENT = "FundsDeposited";
...
+  const { destinationChainId, inputToken, message } = log.args;
   const eventArgs = Object.keys(log.args).filter((key) => isNaN(Number(key)));
-  const padLeft = eventArgs.reduce((acc, cur) => (cur.length > acc ? cur.length : acc), 0);
+  const relayData = Object.fromEntries(eventArgs.map((arg) => [arg, log.args[arg]])) as RelayData;
+  const relayDataHash = sdkUtils.getRelayDataHash({ ...relayData, originChainId }, destinationChainId);
 
   const fields = {
     tokenSymbol: resolveTokenSymbols([inputToken], originChainId)[0],
     ...Object.fromEntries(eventArgs.map((key) => [key, log.args[key]])),
+    messageHash: getMessageHash(message),
```
pxrl, line 57: This makes it easier to identify an invalid fill, because we can now compare messageHash directly from the fill event.

scripts/spokepool.ts
```diff
@@ -45,14 +46,19 @@ const DEPOSIT_EVENT = "FundsDeposited";
 const FILL_EVENT = "FilledRelay";
 
 function printDeposit(originChainId: number, log: LogDescription): void {
-  const { inputToken } = log.args;
+  const { destinationChainId, inputToken, message } = log.args;
```
bmzig, line 49: Just to be sure, `log` is guaranteed to be a deposit? I think this would error on the destructuring if `log` is a fill since the `message` field would be undefined?

bmzig (approved):

james-a-morris (approved):

PR #2116, opened 2025-02-21
chore: Bump SDK for fast "replacement underpriced" resolution

pxrl:
Currently impacted by this on mainnet, and the relayer is retrying all providers before ultimately failing.


nicholaspai (approved):

pxrl:
Not required - @bmzig's PR will bump to sdk patch level 18.

PR #2121, opened 2025-02-24
fix(relayer): Safe default for unknown MDC

pxrl:


mrice32 (dismissed):

src/relayer/Relayer.ts
```diff
@@ -337,7 +337,7 @@ export class Relayer {
       return ignoreDeposit();
     }
 
-    const { minConfirmations } = minDepositConfirmations[originChainId].find(({ usdThreshold }) =>
+    const { minConfirmations = 100_000 } = minDepositConfirmations[originChainId]?.find(({ usdThreshold }) =>
```
pxrl, line 340:
```suggestion
    const { minConfirmations = 100_000 } = minDepositConfirmations[originChainId].find(({ usdThreshold }) =>
```

PR #2126, opened 2025-02-25
[WIP] improve(TransactionUtils): Simple priority scaler for replacement transactions

pxrl:
If transaction submission fails and the word "underpriced" is found in the error message, request that the current estimated priority fee be increased on the subsequent retry.

This isn't foolproof but is cheap to try and makes sense as an initial step towards successfully submitting replacement transactions during gas price spikes.

This can be improved at least by aligning with the provider layer to supply an objective code that indicates an underpriced transaction, rather than having to rely on string matching. It would additionally be beneficial to be able to source the actual gas price of for other transactions with the same nonce. This however involves additional RPC queries

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -9,6 +9,7 @@ import {
   bnZero,
   Contract,
   isDefined,
+  fixedPointAdjustment,
   TransactionResponse,
   ethers,
   getContractInfoFromAddress,
@@ -63,7 +64,8 @@ export async function runTransaction(
   value = bnZero,
   gasLimit: BigNumber | null = null,
   nonce: number | null = null,
-  retriesRemaining = 2
+  retriesRemaining = 2,
+  bumpGas = false
 ): Promise<TransactionResponse> {
   const { provider } = contract;
   const { chainId } = await provider.getNetwork();
@@ -87,6 +89,15 @@ export async function runTransaction(
       maxFeePerGasScaler,
       await contract.populateTransaction[method](...(args as Array<unknown>), { value })
     );
+
+    // Bump the priority fee by 20% to try to successfully replace a pending transaction.
+    // Success is not guaranteed since the bot does not know the gas price of the transaction it is trying to replace.
+    if (bumpGas && gas.maxPriorityFeePerGas) {
+      const oldPriorityFee = gas.maxPriorityFeePerGas;
+      const newPriorityFee = oldPriorityFee.mul(toBNWei("1.2").div(fixedPointAdjustment));
+      gas.maxFeePerGas = gas.maxFeePerGas.add(newPriorityFee).sub(oldPriorityFee);
+      gas.maxPriorityFeePerGas = newPriorityFee;
+    }
 
     logger.debug({
       at: "TxUtil",
@@ -118,7 +129,8 @@ export async function runTransaction(
         retriesRemaining,
       });
 
-      return await runTransaction(logger, contract, method, args, value, gasLimit, null, retriesRemaining);
+      const bumpGas = isEthersError(error) && error.message.toLowerCase().includes("underpriced");
+      return await runTransaction(logger, contract, method, args, value, gasLimit, null, retriesRemaining, bumpGas);
     } else {
       // Empirically we have observed that Ethers can produce nested errors, so we try to recurse down them
       // and log them as clearly as possible. For example:


src/utils/TransactionUtils.ts
```diff
@@ -88,6 +90,15 @@ export async function runTransaction(
       await contract.populateTransaction[method](...(args as Array<unknown>), { value })
     );
 
+    // Bump the priority fee by 20% to try to successfully replace a pending transaction.
+    // Success is not guaranteed since the bot does not know the gas price of the transaction it is trying to replace.
+    if (bumpGas && gas.maxPriorityFeePerGas) {
```
pxrl, line 95: Legacy transactions are currently ignored since issues have only been seen on eip-1559 chains. It would make sense to scale legacy transactions though.

src/utils/TransactionUtils.ts
```diff
@@ -88,6 +92,16 @@ export async function runTransaction(
       await contract.populateTransaction[method](...(args as Array<unknown>), { value })
     );
 
+    // Bump the priority fee incrementally on each retry to try to successfully replace a pending transaction.
+    // Success is not guaranteed since the bot does not know the gas price of the transaction it is trying to replace.
+    if (bumpGas && gas.maxPriorityFeePerGas) {
+      const priorityScaler = 1.1 + (1 + DEFAULT_RETRIES - Math.min(retriesRemaining, DEFAULT_RETRIES)) / 10;
```
pxrl, line 98: At least some aspect of this scaler should probably be configurable.

src/utils/TransactionUtils.ts
```diff
@@ -118,7 +132,8 @@ export async function runTransaction(
         retriesRemaining,
       });
 
-      return await runTransaction(logger, contract, method, args, value, gasLimit, null, retriesRemaining);
+      const bumpGas = isEthersError(error) && error.message.toLowerCase().includes("underpriced");
```
pxrl, line 135: todo: link up more cleanly with the provider layer. The provider layer should be able to explicitly identify these cases and tag them with a code we can evaluate.

src/utils/TransactionUtils.ts
```diff
@@ -141,6 +145,16 @@ export async function runTransaction(
       };
     }
 
+    // Bump the priority fee incrementally on each retry to try to successfully replace a pending transaction.
+    // Success is not guaranteed since the bot does not know the gas price of the transaction it is trying to replace.
+    if (bumpGas && gas.maxPriorityFeePerGas) {
+      const priorityScaler = 1.1 + (1 + DEFAULT_RETRIES - Math.min(retriesRemaining, DEFAULT_RETRIES)) / 10;
```
pxrl, line 151:
```suggestion
      const priorityScaler = 1.1 + (1 + DEFAULT_RETRIES - retriesRemaining) / 10;
```

src/utils/TransactionUtils.ts
```diff
@@ -178,7 +192,8 @@ export async function runTransaction(
       });
 
       nonceReset[chainId] = false;
-      return await runTransaction(logger, contract, method, args, value, gasLimit, null, retriesRemaining);
+      const bumpGas = isEthersError(error) && error.message.toLowerCase().includes("underpriced");
```
fusmanii, line 195: I think this might not work linea

pxrl: Yeah; that's true. We'll need separate handling for that unfortunately.

bmzig (dismissed):

src/utils/TransactionUtils.ts
```diff
@@ -132,85 +104,112 @@ export async function runTransaction(
...
-    const baseFeeDelta = maxPriorityFeePerGas.sub(gas.maxPriorityFeePerGas);
-    gas = {
-      maxFeePerGas: gas.maxFeePerGas.add(baseFeeDelta),
+    const maxPriorityFeePerGas = sdkUtils.bnMax(scaledGas.maxPriorityFeePerGas, flooredPriorityFeePerGas);
+    const baseFeeDelta = maxPriorityFeePerGas.sub(scaledGas.maxPriorityFeePerGas);
+    scaledGas = {
+      maxFeePerGas: scaledGas.maxFeePerGas.add(baseFeeDelta),
       maxPriorityFeePerGas,
     };
   }
```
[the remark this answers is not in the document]
pxrl: Done.

src/utils/TransactionUtils.ts
```diff
@@ -97,17 +70,13 @@ export async function runTransaction(
...
+  gasScaler = 1.0
 ): Promise<TransactionResponse> {
   const at = "TxUtil#runTransaction";
   const { provider, signer } = contract;
   const { chainId } = await provider.getNetwork();
-
-  if (!nonceReset[chainId]) {
-    nonce = await provider.getTransactionCount(await signer.getAddress());
-    nonceReset[chainId] = true;
-  }
```
pxrl, line 109: If the caller didn't set (or re-set) the nonce, query it unconditionally. Relocate this to the try/catch statement below, where the gas price is also queried. It's not expected that `getTransactionCount()` should fail much/ever.

bmzig (approved):

src/utils/TransactionUtils.ts
```diff
@@ -152,53 +122,82 @@ export async function runTransaction(
...
+        break;
+
+      case errors.REPLACEMENT_UNDERPRICED:
+        message = `Transaction replacement on ${chain} failed at nonce ${nonce} (${cause}).`;
+        scaleGas = true;
+        break;
+
+      // Undiagnosed issue. Can be a nonce issue, so try to re-sync, and otherwise
+      // pull out the underlying error to support manual diagnosis if necessary.
+      case errors.SERVER_ERROR:
```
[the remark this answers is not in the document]
pxrl: Good question; I wasn't conscious of `CALL_EXCEPTION`. I checked the docs though, and it should only be emitted on static calls or when waiting on transaction confirmations, so we shouldn't be able to hit it here. Anecdotally, I also don't think I've ever seen it be thrown here either.

We will need to handle it as part of the follow-up PR that actually waits on transaction confirmation, but that will be in a separate function.

https://docs.ethers.org/v5/single-page/#/v5/api/utils/logger/-%23-errors--call-exception

fusmanii (dismissed):

bmzig (approved):

PR #2133, opened 2025-03-06
feat: support rebalancing via OFT

grasphoper:
This PR introduces `OFTBridge` object, which supports sending any token from Ethereum to OFT-supported chain.

I've tested:
 

- [x] sending tokens L1 -> L2 in via running `yarn relay` with custom inventory config. [tx](https://etherscan.io/tx/0xed49a76e27d8292e288a19b49549c8869f68d1164ad63871c1a25689898a654a)
- [X] initiation deposit was found successfully and was part of vitrual reserves 
- [X] finalization event was found and matched with initiation

P.S. This PR depends on another [PR in contracts repo](https://github.com/across-protocol/contracts/pull/938) for some IOFT interfaces from typechain.

Closes ACX-3763, ACX-3764, ACX-3765


scripts/oftTest.ts
```diff
@@ -0,0 +1,277 @@
...
+    });
+
+    // Create an IOFT contract instance with the correct interface
+    const oftContract = new ethers.Contract(txDetails.contract.address, IOFT__factory.abi, signer);
+
+    // Extract the arguments from txDetails
+    if (txDetails.method === "send" && txDetails.args.length >= 3) {
+      const sendParam = txDetails.args[0] as SendParamStruct;
+      const fee = txDetails.args[1] as MessagingFeeStruct;
+      const refundAddress = txDetails.args[2] as string;
```
pxrl, line 210: It's curious that these type assertions are needed; are the types not available natively in the response from `constructL1ToL2Txn()` ?

```suggestion
      const sendParam = txDetails.args[0] as SendParamStruct;
      const fee = txDetails.args[1] as MessagingFeeStruct;
      const refundAddress = txDetails.args[2] as string;
```

scripts/oftTest.ts
```diff
@@ -0,0 +1,277 @@
...
+          const directReceipt = await directTx.wait();
+          console.log(`Direct transaction confirmed in block ${directReceipt.blockNumber}`);
+        } catch (directError) {
+          console.error("Error sending direct transaction:", directError);
+        }
+      }
+    } else {
+      console.error(`Unexpected method or arguments: ${txDetails.method}, args length: ${txDetails.args.length}`);
+    }
+  } catch (error) {
```
pxrl, line 270: I'd probably recommend narrowing the scope of this a lot once you're through debugging.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,340 @@
...
+   */
+  private addOftContract(
+    tokenAddress: string,
+    chainId: number,
+    contractAddress: string,
+    signerOrProvider: Signer | Provider
+  ): void {
+    if (!this.oftContracts[tokenAddress]) {
+      this.oftContracts[tokenAddress] = {};
+    }
```
pxrl, line 74: Minor space-saver:

```suggestion
    this.oftContracts[tokenAddress] ??= {};
```

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,340 @@
...
+
+    // ! note : if (msg.value != _nativeFee) revert NotEnoughNative(msg.value);
+    // ! nativeFee value has to be EXACTLY the msg.value
+
+    console.log("ethFee: ", ethFee.toString());
+
+    const cap = ethers.utils.parseEther("0.01");
+
+    if (ethFee.gt(cap)) {
+      throw "can't go over fee cap";
```
pxrl, line 199:
```suggestion
      throw `can't go over fee cap (${ethFee} > ${cap})`;
```

Would be nice to include the amounts in the error here.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,340 @@
...
+      eventConfig
+    );
+
+    // Filter events by destination EID since it's not an indexed parameter
+    // and can't be filtered directly in the query
+    const destinationEId = this.l2DestinationEId;
+    const events = allEvents.filter((event) => {
+      // The dstEid is the first parameter in the OFTSent event
+      return event.args.dstEid === destinationEId;
+    });
```
pxrl, line 255:
```suggestion
    // The dstEid is the first parameter in the OFTSent event
    const events = allEvents.filter((event) => event.args.dstEid === destinationEId);
```

scripts/oftTest.ts
grasphoper: yea, ofc. This is just a 'pile everything on top' test script for now :D

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,317 @@
...
+    if (l1Token !== this.hubTokenAddress) {
+      throw new Error(`This bridge instance only supports token ${this.hubTokenAddress}, not ${l1Token}`);
+    }
+
+    const isSpokePool = await isContractDeployedToAddress(toAddress, this.l2Bridge.provider);
+    if (isSpokePool && fromAddress != this.hubPoolAddress) {
+      return Promise.resolve({});
+    } else if (fromAddress != toAddress) {
+      return Promise.resolve({});
+    }
```
grasphoper, line 223: Appreciate opinions on this resolving of queries. `OFTSent` has no `toAddress` field, so have to do some deduction.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,317 @@
...
```
grasphoper, line 271: Appreciate opinions on this resolving of queries. `OFTReceived` has no `fromAddress` field, so have to do some deduction.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,317 @@
...
+  // Structure: { tokenAddress: { dstChainId: OFTRouteInfo } }
+  private static readonly SUPPORTED_ROUTES: OFTRoutes = {
+    // USDT supports transfers from Ethereum to Arbitrum
+    [TOKEN_SYMBOLS_MAP.USDT.addresses[CHAIN_IDs.MAINNET]]: {
+      [CHAIN_IDs.ARBITRUM]: {
+        hubChainIOFTAddress: "0x6C96dE32CEa08842dcc4058c14d3aaAD7Fa41dee",
+        dstIOFTAddress: "0x14E4A1B13bf7F943c8ff7C51fb60FA964A298D92",
+      },
+    },
+  };
```
grasphoper, line 45: I figured it's nice to have this live 'close to home' for OFT as no one else would want to use this config, but open to moving it also

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+      throw new Error(`OFT bridge only supports Ethereum as hub chain, got chain ID: ${hubChainId}`);
+    }
+
+    // Check if the route exists for this token and chain
+    const route = OFTBridge.SUPPORTED_ROUTES[hubChainToken]?.[dstChainId];
+    if (!isDefined(route)) {
+      throw new Error(`No route found for token ${hubChainToken} from chain ${hubChainId} to ${dstChainId}`);
+    }
+
+    // todo: do we really want to approve every supported IOFT for `export const MAX_SAFE_ALLOWANCE = "79228162514264337593543950335";`?
```
bmzig, line 87: I think this is a good point. We do this now since we rebalance quite a bit, and so having to approve all bridges each time before rebalancing would get expensive fast, especially since this is all on L1.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+   * @returns Transaction details for execution
+   */
+  async constructL1ToL2Txn(
+    toAddress: string,
+    l1Token: string,
+    _l2Token: string,
+    amount: BigNumber
+  ): Promise<BridgeTransactionDetails> {
+    // Verify the token matches the one this bridge was constructed for
+    if (l1Token !== this.hubTokenAddress) {
```
bmzig, line 127: I also don't think this check is necessary since each adapter (in this case, `ArbitrumAdapter`), has a map from `token => bridge`, so it shouldn't be possible for us to use this bridge for anything else (also, this bridge shouldn't get approvals to spend other tokens for that same reason).

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+      throw new Error(`This bridge instance only supports token ${this.hubTokenAddress}, not ${l1Token}`);
+    }
+
+    // SAFETY LIMIT: Cap transfer amount to prevent large transfers
+    // This is retained as a safety measure to protect against large accidental transfers
+    if (amount.gt(OFTBridge.MAX_AMOUNT)) {
```
bmzig, line 133: We will almost certainly trigger this warning a lot (i.e. attempt to bridge > 100k USDT), so you may want to bump it up to something like 500k.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+    // Get the messaging fee for this transfer
+    const l1Bridge = this.l1Bridge;
+    const feeStruct: MessagingFeeStruct = await (l1Bridge as IOFT).quoteSend(sendParamStruct, false);
+
+    if (BigNumber.from(feeStruct.nativeFee).gt(OFTBridge.FEE_CAP)) {
+      throw new Error(`Fee exceeds maximum allowed (${feeStruct.nativeFee} > ${OFTBridge.FEE_CAP})`);
+    }
+
+    // Set refund address to signer's address (used for dust refunds)
+    const refundAddress = await l1Bridge.signer.getAddress();
```
bmzig, line 162: I think it's better to use `toAddress` (which will always be the signer in the context of the relayer) and save an async call.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+   * Validates that the token amount has the correct precision for OFT transfer.
+   * The last (tokenDecimals - sharedDecimals) digits must be zero to prevent rounding.
+   * @param amount - Amount to validate
+   * @throws Error if the amount has invalid precision
+   */
+  private async validateAmountPrecision(amount: BigNumber): Promise<void> {
+    // Lazy-load shared decimals if not loaded yet
+    if (this.sharedDecimals === undefined) {
+      this.sharedDecimals = await this.l1Bridge.sharedDecimals();
+    }
```
bmzig, line 183: nit
```suggestion
    this.sharedDecimals ??= await this.l1Bridge.sharedDecimals();
```

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    // Verify the amount has the correct precision
+    // The last (tokenDecimals - sharedDecimals) digits must be zero
+    const decimalDifference = this.tokenDecimals - this.sharedDecimals;
```
bmzig, line 187: Have you experienced issues here when running the rebalancer? I believe we rebalance by calculating some percent allocation multiplied by a token amount, so I'd think that, unless we get _very_ lucky, that this will always throw in practice. Is it OK if we just round down the amount to bridge instead ourselves so we don't need to throw if the input isn't perfect?

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+   * @returns Events grouped by token address
+   */
+  async queryL1BridgeInitiationEvents(
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // Verify the token matches the one this bridge was constructed for
+    if (l1Token !== this.hubTokenAddress) {
```
bmzig, line 215: Same as https://github.com/across-protocol/relayer/pull/2133/files#r2040240319

src/adapter/bridges/OFTBridge.ts
bmzig: Yeah this inconsistency only arises in our testing suite. `fromAddress` will never not be `toAddress` when we call this function normally, so I think you can safely remove it.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+      ),
+      eventConfig
+    );
+
+    // Filter events by destination eid
+    const events = allEvents.filter(({ args }) => args.dstEid === this.dstChainEid);
+
+    return {
+      [this.dstTokenAddress]: events.map((event) => {
+        event.args.toAddress = toAddress;
```
bmzig, line 242: We no longer look for the `toAddress` in the event anyways. 
```suggestion
```

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+      eventConfig
+    );
+
+    // Filter events by source eid
+    const sourceEid = getOFTEidForChainId(this.hubChainId);
+    const events = allEvents.filter((event) => event.args.srcEid === sourceEid);
+
+    return {
+      [this.dstTokenAddress]: events.map((event) => {
+        event.args.fromAddress = fromAddress;
```
bmzig, line 291: Same as https://github.com/across-protocol/relayer/pull/2133/files#r2040240791

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+    dstSignerOrProvider: Signer | Provider,
+    hubChainToken: string
+  ) {
+    // OFT bridge currently only supports Ethereum as hub chain
+    if (hubChainId !== CHAIN_IDs.MAINNET) {
+      throw new Error(`OFT bridge only supports Ethereum as hub chain, got chain ID: ${hubChainId}`);
+    }
+
+    // Check if the route exists for this token and chain
+    const route = OFTBridge.SUPPORTED_ROUTES[hubChainToken]?.[dstChainId];
```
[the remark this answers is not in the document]
grasphoper: Well, it's strictly not necessary, but ensures consistency between the map stored here in this file and the `CUSTOM_BRIDGES` map. It's also a constructor check, so no runtime overhead.

Overall I feel like it's nice, but not married to this check :D

src/adapter/bridges/OFTBridge.ts
grasphoper: Right. It's a tradeoff for sure. Doing it like this makes our system rely on OFT security a lot more. E.g. if we approve it for USDT, than all USDT on mainnet is in danger if OFT contract gets hacked. Well although for OFT something like infinite mint is more plausible than an attack involving `transferFrom` for all approvers (I think?)

Would love to hear thoughts of others as well here

src/adapter/bridges/OFTBridge.ts
grasphoper: I feel like having checks like this is nice because it puts some implicit assumptions into "code rules". E.g. if you try to run it with different assumptions, code would crash -- it's useful for debugging

src/adapter/bridges/OFTBridge.ts
grasphoper: Yeah, I just put a somewhat random number in there. Used this for testing. Also, right now, the call just throws which is not ideal. I'd change it to something else perhaps. I'd tread it as a sort of hard cap on how much we want to transfer in one tx via OFT.

src/adapter/bridges/OFTBridge.ts
grasphoper: I couldn't find the `.toAddress()` fn you were referring to. Could you elaborate?

src/adapter/bridges/OFTBridge.ts
grasphoper: I feel like rounding down on the caller level is better? because it creates a correct expectation for the caller as to what amount will actually be sent. If it doesn't cause any side-effects, we could round down here as well.

Do you think it's better? We for sure need to add this rounding somewhere, it's a good point

src/adapter/bridges/OFTBridge.ts
grasphoper: Ah I get it, `toAddress` the argument. Hmm. That requires us to send to `toAddress` that is the same that our `fromAddress` in this function. It becomes an implicit call requirement. I feel like it's better to not have implicit requirements. I believe this async is calculating the address from PrivateKey, shouldn't take too long?

src/adapter/bridges/OFTBridge.ts
grasphoper: I've just pulled a commit from master that removes fromAddress / toAddress fields completely from `processEvent` call. Do I even need this `fromAddress` deduction logic anymore?

src/adapter/bridges/OFTBridge.ts
grasphoper: Right, so the deduction logic can be safely removed also?

src/common/ContractAddresses.ts
```diff
@@ -34,6 +34,8 @@ import BLAST_OPTIMISM_PORTAL_ABI from "./abi/BlastOptimismPortal.json";
 import SCROLL_GATEWAY_ROUTER_L1_ABI from "./abi/ScrollGatewayRouterL1.json";
 import SCROLL_GATEWAY_ROUTER_L2_ABI from "./abi/ScrollGatewayRouterL2.json";
 import SCROLL_GAS_PRICE_ORACLE_ABI from "./abi/ScrollGasPriceOracle.json";
+import IOFT_ABI_FULL from "./abi/IOFT.json";
+export { IOFT_ABI_FULL };
```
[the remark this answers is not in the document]
grasphoper: The semantics of `CONTRACT_ADDRESSES` says something like: We store contract address per chain here. But I need multiple contract addresses per chain with OFTs, so it didn't feel right adding smth like `CONTRACT_ADDRESSES[CHAIN_IDs.MAINNET][DEFAULT_OFT_ABI] = {address: "0x", abi: OFT_ABI_FULL}`

grasphoper:
@bmzig About the safety checks and argument consistency checks in `OFTBridge`. Honeslty, I think if we're going to care about consistency of call arguments at the Adapter level (who's calling these bridge objects) -- it's fine

I just don't know that part of the system yet and wanted to "make my object sort of like a robust API". But now thinking about this, it may not make sense for such a primitive object and can be enforced 1 level higher at the level of the caller

grasphoper:
> @bmzig About the safety checks and argument consistency checks in `OFTBridge`. Honeslty, I think if we're going to care about consistency of call arguments at the Adapter level (who's calling these bridge objects) -- it's fine
> 
> I just don't know that part of the system yet and wanted to "make my object sort of like a robust API". But now thinking about this, it may not make sense for such a primitive object and can be enforced 1 level higher at the level of the caller

Overall, we need *some* object do do these checks, so it doesn't really matter which one does it

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+      oftCmd: "0x", // Empty bytes
+    };
+
+    // Get the messaging fee for this transfer
+    const l1Bridge = this.l1Bridge;
+    const feeStruct: MessagingFeeStruct = await (l1Bridge as IOFT).quoteSend(sendParamStruct, false);
+
+    if (BigNumber.from(feeStruct.nativeFee).gt(OFTBridge.FEE_CAP)) {
+      throw new Error(`Fee exceeds maximum allowed (${feeStruct.nativeFee} > ${OFTBridge.FEE_CAP})`);
+    }
```
pxrl, line 159: If we ever hit this in production then it'll take the rebalancer down and will prevent bridging to any other chains. That would require a code change, or at least a config change. I think it makes sense to instead alert about it via a log messsage and abort the deposit.

@bmzig We don't have a logger instance in this class; do you think there could be a case for adding it to the `BaseBridgeAdapter` ?

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    // Set refund address to signer's address (used for dust refunds)
+    const refundAddress = await l1Bridge.signer.getAddress();
+
+    // Return transaction details
+    return Promise.resolve({
```
pxrl, line 165: Is `Promise.resolve()` required here?

src/adapter/bridges/OFTBridge.ts
pxrl: Is there any reason we should limit querying events for a token even if it's not the token that we support bridging via this adapter? That seems like an artificial constraint that reduces the usability of this method.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+  async queryL2BridgeFinalizationEvents(
+    l1Token: string,
+    fromAddress: string,
+    toAddress: string,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // Verify the token matches the one this bridge was constructed for
+    if (l1Token !== this.hubTokenAddress) {
+      throw new Error(`This bridge instance only supports token ${this.hubTokenAddress}, not ${l1Token}`);
+    }
```
pxrl, line 265: Same as above - these query methods don't have any side-effects, so I think we can probably drop this check.

src/adapter/bridges/OFTBridge.ts
bmzig: Is this not at the caller level? If you mean doing this in the inventory client, I'd disagree that we should round there, since that means we have some conditional checking whether we are using a bridge that requires rounding. That conditional would obviously be hit more than just rounding here.

src/adapter/bridges/OFTBridge.ts
bmzig: You still need this. That commit you pulled only changed what you return from this function.

bmzig:
I am still generally against doing all these checks in the bridge since I think they aren't that helpful and just add diff to the PR (not doing the checks would also still cause the bot to throw further down in the code anyways, which would be effectively an equivalent outcome), but if you are dead set on keeping them, then I agree w/ Paul in that instead of doing
```
if (something) {
   throw new Error(...);
}
```
it would be cleaner to just do an `assert(something, "Something didn't happen")`.

src/common/ContractAddresses.ts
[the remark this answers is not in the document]
bmzig: Fair enough :+1:

src/adapter/bridges/OFTBridge.ts
grasphoper: It's not artificial. We're querying events on a specific `IOFT`, which is per token.

grasphoper: If we query with a different token, we'll just return events for the exact same amounts as if we queried for a correct token. So this check is important I think. But instead of throwing, I can just return an empty event list

grasphoper:
> I am still generally against doing all these checks in the bridge since I think they aren't that helpful and just add diff to the PR (not doing the checks would also still cause the bot to throw further down in the code anyways, which would be effectively an equivalent outcome), but if you are dead set on keeping them, then I agree w/ Paul in that instead of doing
> 
> ```
> if (something) {
>    throw new Error(...);
> }
> ```
> 
> it would be cleaner to just do an `assert(something, "Something didn't happen")`.

I think I'll remove most of the checks, or move towards returning empty responses instead of throwing.

Might keep checks on construction?

src/adapter/bridges/OFTBridge.ts
bmzig: The inventory client would try to bridge an amount, say 123456 USDT (in wei), it would get to here when constructing the transaction, be rounded to 123400 USDT, and then sent off. On the next run, when calling `queryL1BridgeInitiationEvents`, it would observe that there is an outstanding transfer of 123400 USDT, which would be above threshold, so we would not bridge again. Basically, the inventory client is OK with not bridging the exact amount it wants to, since it is just trying to get near its targets.

grasphoper: Ok, great! Then I'll impl rounding here.

src/adapter/bridges/OFTBridge.ts
grasphoper: > So even without granular on-demand approvals, the amount of trust placed in this bridge is already very significant.

Agreed. Will just leave the approve in place

src/adapter/bridges/OFTBridge.ts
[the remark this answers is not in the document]
grasphoper: > but ensures consistency between the map stored here in this file and the CUSTOM_BRIDGES map

Will leave it because this

src/adapter/bridges/OFTBridge.ts
grasphoper: > also, this bridge shouldn't get approvals to spend other tokens for that same reason

So, this check is interesting. The `calldata` that we're constructing has no notion of "token". The token is implicit and is decided on `OFTBridge` creation when we create `this.l1Bridge`.

If we called this fn with a wrong token, the tx will still get created correctly, and it will work and send the token associated with this `OFTBridge` object instead. It'd be bad? to have this happen.

Wdythink @bmzig ?

src/adapter/bridges/OFTBridge.ts
grasphoper: > I'm not sure what any cap provides in terms of safety, because if there's something wrong with the adapter or the bridge interface then we'll just nuke the same amount of funds in n transactions, instead of 1.

Agree. Will remove this check

src/adapter/bridges/OFTBridge.ts
grasphoper: Yeah, I'll remove the `throw` here. As for aborting: in general, what's our policy on returning errors? If I return a `null`, will the caller handle it fine? Or how do you suggest we cancel the tx here?

src/adapter/bridges/OFTBridge.ts
grasphoper: Nope, will remove all such occurrences

src/adapter/bridges/OFTBridge.ts
grasphoper: See [this comment](https://github.com/across-protocol/relayer/pull/2133/files#r2054949458)

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+ * Converts an Ethereum address to bytes32 format for OFT bridge. Zero-pads from the left.
+ * @param address - The Ethereum address to convert
+ * @returns The bytes32 representation of the address
+ */
+export function oftAddressToBytes32(address: string): string {
+  return ethers.utils.hexZeroPad(address, 32);
+}
```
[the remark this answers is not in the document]
grasphoper: Will do. The sdk method is unlikely to change I guess. If it does though, the method might break

src/adapter/bridges/OFTBridge.ts
grasphoper: > That seems like an artificial constraint that reduces the usability of this method.

Also, do you think it's good to allow this object to be used in other contexts? E.g. with other tokens as args? I feel like we should be restrictive instead, using this object only for what it was created to do

src/adapter/bridges/OFTBridge.ts
pxrl: At the moment we're forced to return a `BridgeTransactionDetails` type, which doesn't really allow for failure.

In some places we return a union type of (<success>|<failure>) (i.e. in the SpokePoolClient [here](https://github.com/across-protocol/sdk/blob/9bce8dd8ca80587b2f7a65a6472d0fe763f3ee19/src/clients/SpokePoolClient/SpokePoolClient.ts#L58)). In practice this works well if there are various types of sometimes-expected failures that can happen.

This seems more like a "should never ever happen" type thing, so the simplest is probably to keep this exception, and then add a try/catch clause [here](https://github.com/across-protocol/relayer/blob/02a666f39dd426b3e4422b95b97a316fd6430d58/src/adapter/BaseChainAdapter.ts#L236).

src/adapter/bridges/OFTBridge.ts
pxrl: I retract - I think it makes sense to keep these (probably as asserts).

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,318 @@
...
+import { CHAIN_IDs, PUBLIC_NETWORKS } from "@across-protocol/constants";
+import { CONTRACT_ADDRESSES, IOFT_ABI_FULL } from "../../common";
+
+import {
+  IOFT,
+  MessagingFeeStruct,
+  SendParamStruct,
+} from "@across-protocol/contracts/dist/typechain/contracts/interfaces/IOFT";
+
+import { resolveToken } from "../../../scripts/utils";
```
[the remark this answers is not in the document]
grasphoper: You mean re-export from there?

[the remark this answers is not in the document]
grasphoper: Ok I moved it

src/adapter/bridges/OFTBridge.ts
grasphoper: Looking at this once more, in `sendTokenToTargetChain` that you link to (for adding a try/catch, we actually throw on [e.g. simulation failure](https://github.com/across-protocol/relayer/blob/02a666f39dd426b3e4422b95b97a316fd6430d58/src/adapter/BaseChainAdapter.ts#L252). So if we throw from `constructL1ToL2Txn` if the params are incorrect it should be fine too.

The only extra thing that `try / catch` would allow us to do is to log, as like you mentioned there's no logger in `BaseBridgeAdapter`.

So we could either:

1. Leave throw as is, let some caller catch it (which should happen if `sendTokenToTargetChain` throws tell us anything)
2. Try - catch + log + re-throw

Which one do you prefer we do? @pxrl @bmzig

grasphoper:
@pxrl @bmzig 

Addressed all outstanding comments, I think we have consensus on how to handle most errors. The only unresolved issue I left is [this one](https://github.com/across-protocol/relayer/pull/2133#discussion_r2082537796).

I also simplified code somewhat, removed redundant comments, should be easier to review

src/adapter/bridges/OFTBridge.ts
pxrl: > If we called this fn with a wrong token, the tx will still get created correctly, and it will work and send the token associated with this OFTBridge object instead. It'd be bad? to have this happen.

Good insight - that would be scary. I'm totally in favour of leaving this check in.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,266 @@
...
+import { CONTRACT_ADDRESSES, IOFT_ABI_FULL } from "../../common";
+
+import {
+  IOFT,
+  MessagingFeeStruct,
+  SendParamStruct,
+} from "@across-protocol/contracts/dist/typechain/contracts/interfaces/IOFT";
+
+import { utils } from "@across-protocol/sdk";
+const { toBytes32 } = utils;
```
pxrl, line 27: `toBytes32()` is re-exported out of `src/utils/SDKUtils.ts` :+1: 

```suggestion
import {
  BigNumber,
  EventSearchConfig,
  Provider,
  TOKEN_SYMBOLS_MAP,
  isDefined,
  paginatedEventQuery,
  isContractDeployedToAddress,
  assert,
  EvmAddress,
  Address,
  resolveToken,
  toBytes32,
} from "../../utils";
import { processEvent } from "../utils";
import { CHAIN_IDs, PUBLIC_NETWORKS } from "@across-protocol/constants";
import { CONTRACT_ADDRESSES, IOFT_ABI_FULL } from "../../common";

import {
  IOFT,
  MessagingFeeStruct,
  SendParamStruct,
} from "@across-protocol/contracts/dist/typechain/contracts/interfaces/IOFT";
```

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,266 @@
...
+    assert(
+      isDefined(route),
+      new Error(`No route found for token ${hubTokenAddress.toAddress()} from chain ${hubChainId} to ${dstChainId}`)
+    );
+
+    super(dstChainId, hubChainId, hubSigner, dstSignerOrProvider, [route.hubChainIOFTAddress]);
+
+    this.dstTokenAddress = this.resolveL2TokenAddress(hubTokenAddress);
+    this.dstChainEid = getOFTEidForChainId(dstChainId);
+    this.hubPoolAddress = CONTRACT_ADDRESSES[hubChainId]?.hubPool?.address;
```
pxrl, line 94: For test it might be better to support passing in the `hubPoolAddress`, otherwise you'd need to do some post-instantiation modification of `this.hubPoolAddress` to whatever the deployed variant has. This could be done by supplying a `HubPoolClient` instance instead of `hubChainId`, because from `HubPoolClient` you get all of chainId, address and signer. It does reduce the modularity of the code a little bit because instantiating the `HubPoolClient` is a bit annoying, but we do typically have `HubPoolClient` instances deployed in test.

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,266 @@
...
+    l1Token: EvmAddress,
+    _l2Token: Address,
+    amount: BigNumber
+  ): Promise<BridgeTransactionDetails> {
+    // Verify the token matches the one this bridge was constructed for
+    assert(
+      l1Token.eq(this.hubTokenAddress),
+      new Error(
+        `This bridge instance only supports token ${this.hubTokenAddress.toAddress()}, not ${l1Token.toAddress()}`
+      )
```
pxrl, line 112: Is there any need to supply an `Error` instance to `assert` ? It should handle text just fine.

```suggestion
      `This bridge instance only supports token ${this.hubTokenAddress.toAddress()}, not ${l1Token.toAddress()}`
```

src/adapter/bridges/OFTBridge.ts
pxrl: Good point about the fact that we can already throw errors from within `sendTokenToTargetChain()`. I think in practice this almost never throws, because it looks like it would take down the bot (I don't see any error suppression upstream in the `AdapterManager` or `InventoryClient` classes). So the fact that we don't suppress the error doesn't necessarily mean we would handle such an error gracefully.

I do still think it's important to ensure that the failure to deposit a single token doesn't result in contagion preventing any other tokens from being rebalanced. The `InventoryClient` class expects to receive a `TransactionResponse`  promise but it only retains the `hash`, so that might be an OK place to log and suppress the failure without having to re-jig types all over the place.

src/utils/TokenUtils.ts
```diff
@@ -50,3 +52,33 @@ export function formatUnitsForToken(symbol: string, amount: BigNumberish): strin
...
+  if (_token === undefined) {
+    throw new Error(`Token ${token} on chain ID ${chainId} unrecognised`);
+  }
+
+  return {
+    address: _token.addresses[chainId],
+    decimals: _token.decimals,
+    symbol: _token.symbol,
+  };
+}
```
pxrl, line 84: I might actually propose an update to the spokepool script to remove this code as part of [this ](https://github.com/across-protocol/relayer/pull/2236) PR; this code is all a bit snowflake-y and not really intended for use in production, so I probably made the wrong call when I suggested to relocate it from the scripts directory.

pxrl (commented):
Looking good! I have only minor comments, though it'd also be good to check on tests for this code. There are some pre-established tests for these bridge adapters under `test/generic-adapters`.

src/utils/TokenUtils.ts
grasphoper: So should I recreate the logic from existing TokenUtils.ts fns?

src/adapter/bridges/OFTBridge.ts
grasphoper: fixed [41dbeb8](https://github.com/across-protocol/relayer/pull/2133/commits/41dbeb8bd086e8947107fa3c23301f1d32eb1a59)

src/adapter/bridges/OFTBridge.ts
grasphoper: fixed [41dbeb8](https://github.com/across-protocol/relayer/pull/2133/commits/41dbeb8bd086e8947107fa3c23301f1d32eb1a59)

src/adapter/bridges/OFTBridge.ts
grasphoper: @pxrl [581865f](https://github.com/across-protocol/relayer/pull/2133/commits/581865fc03e352e64bab19a3e61027c900d00984) is this what you had in mind?

src/utils/TokenUtils.ts
grasphoper: Ok did so here [dbf73b7](https://github.com/across-protocol/relayer/pull/2133/commits/dbf73b7c84ffbec760940fb869c174b2e3e77b81)

pxrl (dismissed):

pxrl:
> Looking good! I have only minor comments, though it'd also be good to check on tests for this code. There are some pre-established tests for these bridge adapters under `test/generic-adapters`.

Discussed separately - the experience with the existing tests is that they require a lot of boilerplate, such that the ratio of test:implementation starts to become overweight test. The adapter implementation itself is not super complicated; we should instead aim for a future improvement to narrow the scope of our tests for high leverage and low-maintenance testing.

src/adapter/bridges/OFTBridge.ts
pxrl: Yeah; this is OK. It would blow up if the caller calls `TransactionResponse.wait()` on a failed response, but we don't do that in practice, and if we do it'd imply a larger refactoring anyway.

grasphoper:
@pxrl @bmzig I've updated this PR and corrected the logic for tracking L1 -> L2 events. Previous if conditions made Hub -> Spoke event tracking incorrect. Please take a look. This is the change [b3d65ac](https://github.com/across-protocol/relayer/pull/2133/commits/b3d65ac813e8231a69127f54b4b2132e1f186505)

✅ L1 -> L2 rebalance test: [layerzero scan link](https://layerzeroscan.com/tx/0x5fbca6dd72c1d5214001a9d1e1acd417d93fb7940a619a7243f31b7d52184e1c)

src/adapter/bridges/OFTBridge.ts
```diff
@@ -140,7 +128,7 @@ export class OFTBridge extends BaseBridgeAdapter {
 
   /**
    * Rounds the token amount down to the correct precision for OFT transfer.
-   * The last (tokenDecimals - sharedDecimals) digits must be zero to prevent rounding.
+   * The last (tokenDecimals - sharedDecimals) digits must be zero to prevent constract-side rounding.
```
pxrl, line 131:
```suggestion
   * The last (tokenDecimals - sharedDecimals) digits must be zero to prevent contract-side rounding.
```

src/adapter/bridges/OFTBridge.ts
```diff
@@ -241,11 +215,7 @@ export class OFTBridge extends BaseBridgeAdapter {
   }
 }
 
-/**
- * Retrieves the OFT EID for a given chainId.
- * @param chainId - The chainId to get the OFT EID for
- * @returns The OFT EID for the given chainId
- */
+// Retrieves the OFT EID for a given chainId.
```
pxrl, line 218: Just ooc, why back this out?

grasphoper: Felt like it was to verbose for what that function was

grasphoper: More comments reduced readability rather than improved in that particular case IMO

grasphoper: Do you agree?

grasphoper: Maybe smth like `// Retrieves the OFT Endpoint ID for a given chainId.` could be even clearer

src/adapter/bridges/OFTBridge.ts
```diff
@@ -0,0 +1,230 @@
...
+      amountLD: roundedAmount,
+      // @dev Setting `minAmountLD` equal to `amountLD` ensures we won't hit contract-side rounding
+      minAmountLD: roundedAmount,
+      extraOptions: "0x",
+      composeMsg: "0x",
+      oftCmd: "0x",
+    };
+
+    // Get the messaging fee for this transfer
+    const feeStruct: MessagingFeeStruct = await (this.l1Bridge as IOFT).quoteSend(sendParamStruct, false);
```
bmzig, line 113: Why can't you just do `this.getL1Bridge().quoteSend(...)` here? I don't love the import from `contracts/dist`.

grasphoper: Hm, maybe I can do without it altogether, let me see

grasphoper: [3a27d27](https://github.com/across-protocol/relayer/pull/2133/commits/3a27d272100aa683ed1b7a90c2835aea3eb6cc65)

bmzig (approved):

src/common/abi/IOFT.json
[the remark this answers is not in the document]
grasphoper: @james-a-morris will wait for your response there before I merge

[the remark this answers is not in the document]
grasphoper: Actually going to merge this. I think all the interfaces are there, can clean this up in a follow-up PR too.

PR #2136, opened 2025-03-07
improve(relayer): Warn on invalid repaymentChainId selection

pxrl:
This asserted in production today, taking the relayer down. Warn about this condition instead.


pxrl:
I debugged this with @melisaguevara and we identified why the repaymentChainId selection was incorrect. @melisaguevara will follow up with a fix for the inventory client.

bmzig (approved):

PR #2137, opened 2025-03-07
fix(InventoryClient): avoid hubChain as eligible refund chain when origin is a lite chain

melisaguevara:
We were including mainnet as a possible repayment chain even when deposits were originated from a lite chain.
That was raising the error `Lite chain deposits must be filled on its origin chain`.


nicholaspai (approved):
Nice find!

bmzig (approved):

PR #2139, opened 2025-03-12
Fix typo in log message

Pricstas:
- Fixed typo: `"Enqueing new root bundle proposal txn"` → `"Enqueuing new root bundle proposal txn"`
- Updated test assertions to match the corrected log message.


james-a-morris (commented):
lgtm - thanks for the contrib!

pxrl (approved):

PR #2140, opened 2025-03-12
improve(finalizer): Finalize ZkStack withdrawals via SharedBridge 

nicholaspai:
this isn't required but its probably more future proof. We can still use the old mailbox to finalize


pxrl (approved):
Also +1 nice simplification.

nicholaspai:
> Also +1 nice simplification.

Thanks for the quick review, added one more optimization because I noticed that zksync finalizations are happening now after ~6 hours instead of 24 so we can finalize more frequently

PR #2143, opened 2025-03-14
chore: Bump SDK

pxrl:


pxrl:
nb Marking this as a draft PR for now because the corresponding SDK has not yet been built. The change in this PR is needed to align with SDK PR https://github.com/across-protocol/sdk/pull/910.

bmzig:
Since the sdk package isn't bumped yet, we might want to merge this in: https://github.com/across-protocol/sdk/pull/920. [Adding the hre networks](https://github.com/across-protocol/sdk/blob/d804fbe2e207047309a567823a84e27f7d76c032/src/utils/Multicall.ts#L38) to the list of multicall addresses will break some of the tests where we use multicall, presumably since the contract is not deployed. We can either copy this file (or really just the [txn](https://github.com/across-protocol/sdk/blob/d804fbe2e207047309a567823a84e27f7d76c032/test/utils/multicall.ts#L10)) to the relayer or just merge in the other PR so we expose the transaction to the relayer.

Otherwise, if you want to go the route of adding the multicall deployment utility to the tests here, I can close that PR.

pxrl:
> Otherwise, if you want to go the route of adding the multicall deployment utility to the tests here, I can close that PR.

I'm fine w/ using the sdk variant w/ your PR - definitely happy to avoid duplicating this!

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -114,7 +114,7 @@ export class ZKStackBridge extends BaseBridgeAdapter {
       (event) => compareAddressesSimple(event.args.l1Token, l1Token) && compareAddressesSimple(event.args.to, toAddress)
     );
     return {
-      [this.resolveL2TokenAddress(l1Token)]: bridgeEvents.map((event) => processEvent(event, "amount", "from", "to")),
```
bmzig, line 117: Also just including this driveby fix in this bump. The ordering here was backwards (the l2 recipient has the l1 address and vice versa). This would have just caused the monitor to think there are outstanding transactions for hub -> spoke transfers when there are none.

bmzig (dismissed):

pxrl:
Due to an unexpected issue on Arbitrum chains, we should wait until this SDK PR can be included: https://github.com/across-protocol/sdk/pull/922

tldr, for block.number, Arbitrum chains return the block number of the underlying chain, not the block number on the chain itself :\

bmzig (approved):

bmzig (approved):

PR #2145, opened 2025-03-18
feat: support Lens mainnet

bmzig:
This should generalize the zkstack adapter to work for custom gas token chains.


src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -73,26 +85,38 @@ export class ZKStackBridge extends BaseBridgeAdapter {
...
+            secondBridgeCalldata,
+          ],
+        ];
 
     return {
       contract: this.getL1Bridge(),
       method,
       args,
-      value: txBaseCost,
+      value: isDefined(this.gasToken) ? bnZero : txBaseCost,
```
pxrl, line 119: Very minor nit, but I wonder if this would be safer if we had only a single if statement that sets each field correctly. I don't see anything wrong with this code, but I wonder if having repeated ternary statements might increase the chance of the code being incorrect with future updates/refactoring.

src/adapter/bridges/ZKStackWethBridge.ts
```diff
@@ -128,27 +142,38 @@ export class ZKStackWethBridge extends ZKStackBridge {
...
     let processedEvents;
-    if (isL2Contract) {
-      // Assume the transfer came from the hub pool. If the chain has a custom gas token, then query weth. Otherwise,
-      // query ETH.
+    // ZkSync uses different logic for ETH L2 finalization. Most notably, the transfer events on L2 mark the aliased L1 sender as the sender, while
+    // for custom gas token L2s, the L1 sender is the zero address.
+    if (!usingCustomGasToken) {
+      if (isL2Contract) {
+        // Assume the transfer came from the hub pool. If the chain has a custom gas token, then query weth. Otherwise,
+        // query ETH.
```
pxrl, line 153: This comment is now guarded by `!usingCustomGasToken`; is the comment still correct to reference custom gas tokens?

src/common/ContractAddresses.ts
```diff
@@ -407,6 +403,24 @@ export const CONTRACT_ADDRESSES: {
...
     },
   },
+  [CHAIN_IDs.LENS]: {
+    zkStackBridge: {
+      address: "0xfBEC23c5BB0E076F2ef4d0AaD7fe331aE5A01143",
+      abi: ZK_SYNC_DEFAULT_ERC20_BRIDGE_L2_ABI,
+    },
+    nativeToken: {
+      address: "0x000000000000000000000000000000000000800A",
+      abi: WETH_ABI,
```
pxrl, line 413: This is an interesting quirk; might be worth a comment to clarify that it's actually GHO and not WETH.

src/common/ContractAddresses.ts
```diff
@@ -407,6 +403,24 @@ export const CONTRACT_ADDRESSES: {
...
+    },
+    wrappedNativeToken: {
+      address: "0x6bDc36E20D267Ff0dd6097799f82e78907105e2F",
+      abi: WETH_ABI,
```
pxrl, line 417: Same as above - worth a comment?

src/adapter/BaseChainAdapter.ts
```diff
@@ -345,17 +345,16 @@ export class BaseChainAdapter {
...
           assign(outstandingTransfers, [monitoredAddress, l1Token, l2Token], {
-            totalAmount: outstandingInitiatedEvents.reduce((acc, event) => acc.add(event.amount), BigNumber.from(0)),
+            totalAmount,
             depositTxHashes: outstandingInitiatedEvents.map((event) => event.transactionHash),
           });
         });
       });
     });
 
-    this.baseL1SearchConfig.fromBlock = l1SearchConfig.toBlock + 1;
```
bmzig, line 356: I think updating the search config here was wrong. Some unit tests (the zkSync and Polygon `getOutstandingCrossChainTransfers` ones) were failing after modifying how we report outstanding transfers (taking a sum, as seen above) because:
- In block 1 there would be a deposit event, we would update the spoke pool clients, and run `getOutstandingCrossChainTransfers`, observing one outstanding transfer.
- Then, in block 2 we would finalize the events, update the spoke pool client, and then run `getOutstandingCrossChainTransfers`, but since the fromBlock is now beyond the initiation block, we would falsely think that there are extra finalizations.
The tests didn't fail there since we expected there to be no events after a deposit/finalization pair happens, but in this old case, there would just be no initiation pairs observed in the block range, giving us a false positive.

src/adapter/bridges/ZKStackBridge.ts
bmzig: I think you're right. Updated here https://github.com/across-protocol/relayer/pull/2145/commits/3d8d28dbd86b02b118530d69016673ebe3021ef2

src/adapter/bridges/ZKStackWethBridge.ts
bmzig: Yeah, it's updated here https://github.com/across-protocol/relayer/pull/2145/commits/3d8d28dbd86b02b118530d69016673ebe3021ef2

src/common/ContractAddresses.ts
bmzig: Added comments. https://github.com/across-protocol/relayer/pull/2145/commits/3d8d28dbd86b02b118530d69016673ebe3021ef2

src/common/ContractAddresses.ts
bmzig: https://github.com/across-protocol/relayer/pull/2145/commits/3d8d28dbd86b02b118530d69016673ebe3021ef2

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -9,16 +9,19 @@ import {
...
 } from "../../utils";
-import { processEvent } from "../utils";
+import { processEvent, matchL2EthDepositAndWrapEvents } from "../utils";
 import { CONTRACT_ADDRESSES } from "../../common";
 import { BridgeTransactionDetails, BaseBridgeAdapter, BridgeEvents } from "./BaseBridgeAdapter";
 import { gasPriceOracle } from "@across-protocol/sdk";
 import { PUBLIC_NETWORKS } from "@across-protocol/constants";
 import * as zksync from "zksync-ethers";
 
 // Scale gas price estimates by 20%.
```
pxrl, line 23: @bmzig Should this be 50% after your change?

src/adapter/BaseChainAdapter.ts
pxrl: Just to round out some discussion in DMs:
- The only bots that are currently using this only ever update once, so the issue you've found here is probably only present in test.
- Because of this tracking constraint, the fast relayer skips querying the bridges for transfer status. I battled a lot with this when getting the fast relayer running; I saw that it would often ingest the same deposit events multiple times and incrementally assumed the balance stuck in bridges was a lot greater than in reality. It'd be really nice to eventually refactor the cross chain transfer tracking to handle successive updates so the fast relayer could use it.

src/adapter/bridges/ZKStackBridge.ts
bmzig: Sorry, that was an artifact of testing. I changed it back to 20% https://github.com/across-protocol/relayer/pull/2145/commits/b7f808f32635714a276436cbf23dca077991ba39

pxrl (approved):

PR #2149, opened 2025-03-20
improve: generalize native token queries

bmzig:
This will be needed for Lens since we will actually be supporting the native token as a rebalance route. Ideally this can be merged a while before Lens support so we can be confident of its equivalence to what we have currently.


src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,61 @@ export class BaseChainAdapter {
...
     threshold: BigNumber,
     target: BigNumber,
     simMode: boolean
   ): Promise<TransactionResponse | null> {
-    const wethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[this.chainId];
-    const ethBalance = await this.getSigner(this.chainId).getBalance();
-    if (ethBalance.lte(threshold)) {
-      this.log("ETH balance below threshold", { threshold, ethBalance });
+    const nativeTokenSymbol = getNativeTokenSymbol(this.chainId);
+    const nativeTokenAddress = TOKEN_SYMBOLS_MAP[`W${nativeTokenSymbol}`].addresses[this.chainId];
```
pxrl, line 277: We might want to flag this line with a comment, because it's not guaranteed that the wrapped symbol always starts with `W` - it could also be `w`. I don't have an elegant way of handling both yet - we do the same hack in the contracts repo - but it's worth noting, at least.

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,61 @@ export class BaseChainAdapter {
...
+        );
+      }
     }
 
-    const value = ethBalance.sub(target);
+    const value = nativeTokenBalance.sub(target);
     this.log(
-      `Wrapping ETH on chain ${getNetworkName(this.chainId)}`,
-      { threshold, target, value, ethBalance },
+      `Wrapping native token on chain ${getNetworkName(this.chainId)}`,
```
pxrl, line 306:
```suggestion
      `Wrapping ${nativeTokenSymbol} on chain ${getNetworkName(this.chainId)}`,
```

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,61 @@ export class BaseChainAdapter {
...
+      { threshold, target, value, nativeTokenBalance },
       "debug",
-      "wrapEthIfAboveThreshold"
+      "wrapNativeTokenIfAboveThreshold"
     );
     const method = "deposit";
     const formatFunc = createFormatFunction(2, 4, false, 18);
     const mrkdwn =
-      `${formatFunc(toBN(value).toString())} Ether on chain ${
+      `${formatFunc(toBN(value).toString())} native tokens on chain ${
```
pxrl, line 314:
```suggestion
      `${formatFunc(toBN(value).toString())} ${nativeTokenSymbol} on chain ${
```

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,61 @@ export class BaseChainAdapter {
...
         this.chainId
-      } was wrapped due to being over the threshold of ` + `${formatFunc(toBN(threshold).toString())} ETH.`;
-    const message = `${formatFunc(toBN(value).toString())} Eth wrapped on target chain ${this.chainId}🎁`;
+      } was wrapped due to being over the threshold of ` + `${formatFunc(toBN(threshold).toString())}.`;
+    const message = `${formatFunc(toBN(value).toString())} native tokens wrapped on target chain ${this.chainId}🎁`;
```
pxrl, line 317:
```suggestion
    const message = `${formatFunc(toBN(value).toString())} ${nativetokenSymbol} wrapped on target chain ${this.chainId}🎁`;
```

src/dataworker/DataworkerUtils.ts
```diff
@@ -354,13 +355,17 @@ export function _getRefundLeaves(
...
  */
-function getWethAndEth(chainId: number): string[] {
+function getNativeTokens(chainId: number): string[] {
+  const nativeTokenSymbol = getNativeTokenSymbol(chainId);
   // Can't use TOKEN_SYMBOLS_MAP for ETH because it duplicates the WETH addresses, which is not correct for this use case.
-  const wethAndEth = [TOKEN_SYMBOLS_MAP.WETH.addresses[chainId], CONTRACT_ADDRESSES[chainId].eth.address];
-  if (wethAndEth.some((tokenAddress) => !isDefined(tokenAddress))) {
-    throw new Error(`WETH or ETH address not defined for chain ${chainId}`);
+  const nativeTokens = [
+    TOKEN_SYMBOLS_MAP[`W${nativeTokenSymbol}`].addresses[chainId],
```
pxrl, line 362: Re the above comment about the leading `W` being incidental, we can live with the wart for now but we should probably centralise the implementation so we can later fix it in one go.

pxrl, line 362: We might not make it to the asswert below if the token symbol doesn't exist at all.

```suggestion
    TOKEN_SYMBOLS_MAP[`W${nativeTokenSymbol}`]?.addresses[chainId],
```

bmzig: Factored out here https://github.com/across-protocol/relayer/pull/2149/commits/4e8f22564117bbbbe0c997cd50e705d19ed032b2

src/adapter/BaseChainAdapter.ts
bmzig: Yeah this is at least true for GHO. (Temporarily) fixed here https://github.com/across-protocol/relayer/pull/2149/commits/4e8f22564117bbbbe0c997cd50e705d19ed032b2

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,63 @@ export class BaseChainAdapter {
...
   ): Promise<TransactionResponse | null> {
-    const wethAddress = TOKEN_SYMBOLS_MAP.WETH.addresses[this.chainId];
-    const ethBalance = await this.getSigner(this.chainId).getBalance();
-    if (ethBalance.lte(threshold)) {
-      this.log("ETH balance below threshold", { threshold, ethBalance });
+    const nativeTokenSymbol = getNativeTokenSymbol(this.chainId);
+    const nativeTokenAddress = getWrappedNativeTokenAddress(this.chainId);
+    const nativeTokenBalance = await this.getSigner(this.chainId).getBalance();
+    if (nativeTokenBalance.lte(threshold)) {
+      this.log("Native token balance below threshold", { threshold, nativeTokenBalance });
```
pxrl, line 280: Minor:
```suggestion
      this.log(`${nativeTokenSymbol} balance below threshold`, { threshold, nativeTokenBalance });
```

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,63 @@ export class BaseChainAdapter {
...
+        assert(
+          symbol === "WGHO",
+          `Critical (may delete $$$): Unable to verify ${this.adapterName} native token address (${contract.address})`
+        );
+      } else {
+        assert(
+          symbol === "WETH",
+          `Critical (may delete ETH): Unable to verify ${this.adapterName} WETH address (${contract.address})`
+        );
+      }
```
pxrl, line 301: Could this work?

```suggestion
      assert(
        symbol === nativeTokenSymbol,
        `Critical (may delete ${nativeTokenSymbol}): Unable to verify ${this.adapterName} ${nativeTokenSymbol} address (${contract.address})`
      );
```

pxrl (approved):
LGTM; just to double check, has this been tested locally?

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +267,56 @@ export class BaseChainAdapter {
...
+    const nativeTokenSymbol = getNativeTokenSymbol(this.chainId);
+    const nativeTokenAddress = getWrappedNativeTokenAddress(this.chainId);
+    const nativeTokenBalance = await this.getSigner(this.chainId).getBalance();
+    if (nativeTokenBalance.lte(threshold)) {
+      this.log(`${nativeTokenSymbol} balance below threshold`, { threshold, nativeTokenBalance });
       return null;
     }
     const l2Signer = this.getSigner(this.chainId);
-    const contract = new Contract(wethAddress, WETH_ABI, l2Signer);
+    const contract = new Contract(nativeTokenAddress, WETH_ABI, l2Signer);
```
james-a-morris, line 283: Wouldn't this fail if the wrapped native token doesn't follow the `WETH_ABI` interface?

src/clients/bridges/AdapterManager.ts
```diff
@@ -184,13 +184,13 @@ export class AdapterManager {
     return await this.adapters[chainId].getL2PendingWithdrawalAmount(lookbackPeriodSeconds, fromAddress, l2Token);
   }
 
-  // Check how much ETH is on the target chain and if it is above the threshold the wrap it to WETH. Note that this only
-  // needs to be done on chains where rebalancing WETH from L1 to L2 results in the relayer receiving ETH
-  // (not the ERC20), or if the relayer expects to be sent ETH perhaps as a gas refund from an original L1 to L2
+  // Check how much native tokens are on the target chain and if it is above the threshold the wrap, execute a wrap. Note that this only
```
james-a-morris, line 187:
```suggestion
  // Check how many native tokens are on the target chain and if the number of tokens is above the wrap threshold, execute a wrap. Note that this only
```

james-a-morris (commented):
Are we positive that our existing wrapped tokens follow the `WETH9` ABI?

src/adapter/BaseChainAdapter.ts
bmzig: Yeah, but we only assume a select few functions from the weth interface, most of which come from ExtendedERC20 anyways, so I think this is a safe assumption to make. See https://github.com/across-protocol/relayer/blob/master/src/common/abi/Weth.json

bmzig:
Here's some example wraps we can still make with this change: 
- https://arbiscan.io/tx/0x76ff4e2e4714b20e31ec71f7f8fc97fe722887fe9d1b674d676616e152497669
- https://lineascan.build/tx/0x8e6e0a7da061eaff16145ee5a210d1775895175f68febd30929b7aa98c63e970
Wraps in general will be safeguarded by checking for the symbol. The disputer also runs fine with this change.

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +267,56 @@ export class BaseChainAdapter {
...
 
     // First verify that the target contract looks like WETH. This protects against
     // accidentally sending ETH to the wrong address, which would be a critical error.
     // Permit bypass if simMode is set in order to permit tests to pass.
     if (simMode === false) {
       const symbol = await contract.symbol();
       assert(
-        symbol === "WETH",
-        `Critical (may delete ETH): Unable to verify ${this.adapterName} WETH address (${contract.address})`
+        symbol === `W${nativeTokenSymbol}`,
```
pxrl, line 291: @bmzig We're just about to change the gas token to be WGHO - will this break as a result?

bmzig: Good catch. Simple fix here https://github.com/across-protocol/relayer/pull/2149/commits/fe611051bf41ca8f396b113f68408af960fed69b. Will probably have to follow up with a more general solution in another PR.

src/adapter/BaseChainAdapter.ts
```diff
@@ -266,53 +268,57 @@ export class BaseChainAdapter {
...
     const l2Signer = this.getSigner(this.chainId);
-    const contract = new Contract(wethAddress, WETH_ABI, l2Signer);
+    const contract = new Contract(nativeTokenAddress, WETH_ABI, l2Signer);
 
     // First verify that the target contract looks like WETH. This protects against
     // accidentally sending ETH to the wrong address, which would be a critical error.
     // Permit bypass if simMode is set in order to permit tests to pass.
     if (simMode === false) {
       const symbol = await contract.symbol();
+      const expectedTokenSymbol = this.chainId === CHAIN_IDs.LENS ? nativeTokenSymbol : `W${nativeTokenSymbol}`;
```
nicholaspai, line 291: Should this be `nativeTokenSymbol === "ETH" ? "WETH" : nativeTokenSymbol;` to be more future proof?

src/common/ContractAddresses.ts
```diff
@@ -443,7 +444,7 @@ export const CONTRACT_ADDRESSES: {
       address: "0x4200000000000000000000000000000000000010",
       abi: OVM_L2_STANDARD_BRIDGE_ABI,
     },
-    eth: {
+    nativeToken: {
```
nicholaspai, line 447: I did a search naively for `.eth` in the repo and I found instances in `ZKStackWETHBridge.ts` as well, did you miss that?

nicholaspai (approved):
One question

src/adapter/BaseChainAdapter.ts
bmzig: Yeah, that's a much better way to do this https://github.com/across-protocol/relayer/pull/2149/commits/abccbcb2bb29133720bc58529d62149e8595baa1

src/common/ContractAddresses.ts
bmzig: We need it there since we need a separate eth address and ABI for querying L2 events.

nicholaspai: good call, saw you didn't delete `eth` from that entry

nicholaspai (approved):

src/common/ContractAddresses.ts
bmzig: Never mind, we don't. I removed it https://github.com/across-protocol/relayer/pull/2149/commits/3350df1579a8c55e1e96fd72e631074cc3febadd

pxrl (approved):

PR #2152, opened 2025-03-26
feat(providers): Permit env-based header override

pxrl:


bmzig (approved):

james-a-morris (approved):

src/utils/ProviderUtils.ts
```diff
@@ -167,18 +167,34 @@ export async function getProvider(
 
   // See ethers ConnectionInfo for field descriptions.
   // https://docs.ethers.org/v5/api/utils/web/#ConnectionInfo
+
+  // Permit env-based HTTP headers to be specified.
+  // RPC_PROVIDER_<chainId>_HEADERS=auth
+  // RPC_PROVIDER_<chainId>_HEADER_AUTH=xxx-auth-header
+  let headers: { [k: string]: string };
+  const _headers = process.env[`RPC_PROVIDER_${chainId}_HEADERS`];
+  _headers.split(",").forEach((header) => {
```
bmzig, line 176: We need to have a check here since `_headers` can be undefined. So maybe something like 
```
if (isDefined(_headers)) {
...
}

File: src/utils/ProviderUtils.ts [hunk 2]
```diff
@@ -167,18 +167,34 @@ export async function getProvider(
 
   // See ethers ConnectionInfo for field descriptions.
   // https://docs.ethers.org/v5/api/utils/web/#ConnectionInfo
+
+  // Permit env-based HTTP headers to be specified.
+  // RPC_PROVIDER_<chainId>_HEADERS=auth
+  // RPC_PROVIDER_<chainId>_HEADER_AUTH=xxx-auth-header
+  let headers: { [k: string]: string };
+  const _headers = process.env[`RPC_PROVIDER_${chainId}_HEADERS`];
```
pxrl, line 175: One thing to note: this doesn't work super well with quorum > 1, since it specifies one header for all providers for a given chain. We probably need to proceed with this as-is, but we might want to revisit it to enable it to isolate per provider.

bmzig (dismissed):

bmzig (approved):

PR #2155, opened 2025-03-27
fix(ProviderUtils): Prevent crash on undefined RPC provider

pxrl:


james-a-morris (approved):

pxrl:
Not required after all; undefined providers is already handled by the implementation.

PR #2157, opened 2025-03-28
fix(ClientHelper): Always resolve latest block on Lens

pxrl:
Lens block production is sporadic and bursty, leading to problems determining the right block range for SpokePoolClient search config.


pxrl:
This should be unnecessary now.

PR #2165, opened 2025-04-01
feat: Support CCTP V2 in rebalancer, finalizer, and monitor

nicholaspai:
~Goal is to build this in such a way that its easy in the future to deprecate V1 support. I think there are overall refactors i'd like to make but they would make the PR harder to review. So I'm aiming to make this PR be purely additive to the existing V1 stack so we can delete the V1 stack eventually.~

Yeah I ended up not doing this and rebuilding the CCTP utilities file from scratch so that it ended up much cleaner.

The `CCTPUtils` file now contains exported functions used by the USDCctpAdapter, L1->L2, and L2->L1 finalizers. The `l1ToL2.ts` and `l2ToL1.ts` finalizers are now almost identical to each other and have much less code.

## Blockers

Depends on the following PR's being merged upstream first:
1. [Constants bump to move USDC.e to USDC entry for Linea](https://github.com/across-protocol/constants/pull/122)
2. [Minor SDK bump to make logging show USDC instead of USDC.e](https://github.com/across-protocol/sdk/pull/958/files)

## Deriving V2 `nonceHash`
I want to remove some logic that reads the `nonceHash` from the V2 attestation API response, because I can't figure out how to derive the deterministic hash from scratch. The logic in [the docs](https://developers.circle.com/stablecoins/message-format) is mostly clear but not precise enough and I can't recreate the `nonceHash`, therefore the flow for V1 and V2 is slightly different.
    - For V1: query DepositForBurn events --> query `MessageTransmitter.usedNonces` using nonce from DepositForBurn event --> for any un-used nonces, query attestation API --> submit attestation to destination chain
    - The ideal V2 flow would be identical to V1
    - However, the flow is currently: query DepositForBurn events --> query attestation API, get nonceHash --> query `MessageTransmitter.usedNonces` using nonce from API response --> for any un-used nonces, submit attestation to destination chain

However, we need to rely on Circle to help us implement an algorithm to derive the `nonceHash`, so until then I'll query the API with just the `transactionHash` and then map the `messages` in the response to the logIndex of the Message.

## Testing
- [x] Initiate L1 to L2 rebalance with `UsdcCCTP` adapter: [l1 txn](https://etherscan.io/tx/0xb0942874904b9bddd097064a4c28b43fb2023a774e73709505bd15328cb2eefa)
- [x] Finalize L1 to L2 rebalance with `cctpL1ToL2Finalizer`: [l2 txn](https://lineascan.build/tx/0x7ea22fbea5015401dbbf0bba07c8e760e0a28e19de65967238992912986a067a)
- [x] Finalize batch L1 to L2 rebalance: [l2 txn](https://lineascan.build/tx/0xbb4b3a2d0ab99badc1c77935c19d2156aeeefab4852280aab3eaa827e183ba68). I had to use a special `LineaAdapter` deployed [here](https://etherscan.io/address/0x5977f0C729b2582AdCc551780Aab312756359646) that chunked up bridges into multiple 1-token atomic bridges. [batch relayTokens l1 txn](https://etherscan.io/tx/0x51682eab9de71468f6997d45064d4212f2b8a54acec62bb8c252b48f7fe609d4).
- [x] Finalize batch L2 to L1 rebalance: [l1 txn](https://etherscan.io/tx/0x59b469b9667903ef03cd6b4b1c20850ea8a54d00024dd9a717464040a98dd76d). I had to use a special `LineaSpokePool` deployed [here](https://lineascan.build/address/0x038e789392f8B3F5dF8e5974Ea25575BA9d63730) that chunked up bridges into multiple 1-token atomic bridges. [batch testBridgeTokens l2 txn](https://lineascan.build/tx/0xe65bfecefa68b3e25efa209174975bf659d55db36ba59c659a4715bd62e2d60b)


src/finalizer/index.ts
```diff
@@ -230,18 +230,20 @@ export async function finalize(
...
-    if (sdkUtils.chainRequiresL1ToL2Finalization(chainId)) {
-      const addressesToEnsure = [
-        hubPoolClient.hubPool.address,
-        CONTRACT_ADDRESSES[hubChainId]?.atomicDepositor?.address,
-      ];
-      // Add the spoke pool address to the list of addresses to ensure.
-      addressesToEnsure.push(spokePoolClients[chainId].spokePool.address);
-      l1ToL2AddressesToFinalize = enrichL1ToL2AddressesToFinalize(l1ToL2AddressesToFinalize, addressesToEnsure);
-    }
+    // Not all finalizer adapters query TokensBridged events on the L2 spoke pools to discover withdrawals that
```
nicholaspai, line 233: Removing this `if (sdkUtils.chainRequiresL1ToL2Finalization(chainId)) {` line allows us to take a step towards generalizing the finalizer adapters so that they all can finalize withdrawals from a user specified list of EOA's. We can't currently do this easily because the L2 to L1 finalizers all "discover" L2 withdrawals by querying `TokensBridged` events emitted by the SpokePool. This doesn't include any relayer rebalances, which is why we have some custom logic in the `opStackFinalizer` and `orbitStackFinalizer` to handle this.  

Here's a follow up [PR](https://github.com/across-protocol/relayer/pull/2166) that can add context for how we can take advantage of removing this if statement.

src/adapter/bridges/UsdcCCTPBridge.ts
```diff
@@ -64,34 +85,31 @@ export class UsdcCCTPBridge extends BaseBridgeAdapter {
...
-    l1Token;
-    fromAddress;
-    eventConfig;
     assert(compareAddressesSimple(l1Token, TOKEN_SYMBOLS_MAP.USDC.addresses[this.hubChainId]));
-
-    // The function queryL1BridgeInitiationEvents already comuptes outstanding CCTP Bridge transfers,
-    // so we can return nothing here.
-    return Promise.resolve({});
+    const eventFilterArgs = [toAddress, undefined, this.resolveL2TokenAddress(this.l1UsdcTokenAddress)];
+    const eventFilter = this.getL2Bridge().filters.MintAndWithdraw(...eventFilterArgs);
```
bmzig, line 108: Should this also be behind `this.IS_CCTP_V2`? I don't see this event being emitted for e.g. Optimism. So maybe just check for a `Mint` event to the `toAddress`?

nicholaspai: I see it [here](https://arbiscan.io/tx/0x8ec06b7882d31f5dcf27206849b83a36b33a93bf9058e6e1838efd6d296d4624#eventlog) on arbitrum but on optimism its not decoded properly [here](https://optimistic.etherscan.io/tx/0x590ffbdd8a73542dfe988553dfc5cdaee4b3129077767bee67a86b1fff1ebefa#eventlog) for example.

When looking at Tenderly I can see its emitted.

bmzig: Ah I see. You're right.

src/utils/CCTPUtils.ts
```diff
@@ -284,10 +398,27 @@ async function _resolveCCTPRelatedTxns(
...
   return attestationResponse;
 }
+
+// @todo: We can pass in a nonceHash here once we know how to recreate the nonceHash.
+async function _generateCCTPV2AttestationProof(
+  sourceDomainId: number,
+  transactionHash: string,
+  isMainnet: boolean
+): Promise<CCTPV2APIGetAttestationResponse> {
+  const httpResponse = await axios.get<CCTPV2APIGetAttestationResponse>(
```
bmzig, line 418: Just to be sure, this works even when there are multiple CCTP deposits in the same transaction?

nicholaspai: yep have verified it with a contract I deployed to send batch txns (see pr desc)

bmzig (approved):
I couldn't find anything wrong. It also helps that you've done integration tests.

PR #2167, opened 2025-04-02
improve: configure Lens finalizer and update zkstack adapter w/ new ABI

bmzig:
This adds the Lens finalizer, which uses the same logic as the ZkSync finalizer. As is, **this requires a defined multicall3 address before it can be activated**.

There is also a quick fix to the rebalancer included. Basically, since in the context of L1->L2 rebalances, `totalAmount = initiatedAmount - finalizedAmount`, if the lookback of the rebalancer is between the initiation and finalization events, then we will observe a finalization event with no initiation event and mistakenly assign negative outstanding transfer amount. To address this, we just floor the finalization amount to 0.


src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -141,16 +145,16 @@ export class ZKStackBridge extends BaseBridgeAdapter {
...
-        this.sharedBridge.filters.BridgehubDepositInitiated(this.l2chainId, undefined, annotatedFromAddress),
-        eventConfig
-      );
+      const [rawEvents, assetId] = await Promise.all([
+        paginatedEventQuery(
+          this.sharedBridge,
+          this.sharedBridge.filters.BridgehubDepositInitiated(this.l2chainId, undefined, annotatedFromAddress),
+          eventConfig
+        ),
+        this.nativeTokenVault.assetId(l1Token),
```
bmzig, line 154: The ABI for the shared bridge changed. Most notably we need to query a token's `assetId` since that is what is emitted from the event in lieu of the `l1Token` address.

bmzig:
The three event queries that needed to change were for bridging a non-native token to a non-custom gas token chain (e.g. bridging USDT to ZkSync), bridging a custom gas token to a custom gas token chain (e.g. bridging LGHO to Lens), and then finalizing a non-custom gas token on a non-custom gas token chain. The actual submission of these transactions does not change, so we just have to verify event tracking works. I've tested with USDT ([L1](https://etherscan.io/tx/0x55cea3bf5b63f486e6fbd0554536ce8c04dbf4f76cd64a644c3c38672af44525) [L2](https://era.zksync.network/tx/0x7d3789f242f3c553456f43944ea4f98787bb0a131c7f918b84aa3b23eb12e879)) and WGHO ([L1](https://etherscan.io/tx/0xa9e6b35fdff7b2efb0f400f475977971b79b78477c29917d6be79fe8aab58460) [L2](https://explorer.lens.xyz/tx/0x2bb50675f9f9307581746dcc4562d1b69795e43fbaa9e961bb0b21e3b3cae6b2)) and saw no issues with event tracking. 

For clarity, we don't need to check other token/chain pairs since we track events which are emitted from contracts which were not changed, e.g. the hub pool or atomic depositor.

nicholaspai (approved):

PR #2173, opened 2025-04-06
improve(finalizer): Bump Viem version to handle OpStack error

nicholaspai:
The finalizer has been throwing because of a new OpStack revert reason unhandled by the current `viem` version, which is patched [here](https://github.com/wevm/viem/commit/58db8b90d72983814664b12bd27bcd2bdf0f15f3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "redis4": "npm:redis@^4.1.0",
     "superstruct": "^1.0.3",
     "ts-node": "^10.9.1",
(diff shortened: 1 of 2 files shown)


james-a-morris (approved):

bmzig:
I've seen this too, but I'm not convinced it finalizes. I.e. there's been many times when running with the new version of viem says that the withdrawal is `waiting-to-prove` when it should be `ready-to-prove`.

PR #2174, opened 2025-04-06
improve(BaseBridgeAdapter): Remove unused event parsing logic

nicholaspai:
The motivation behind this PR is to reduce the amount of work it costs to support a new canonical bridge in the BaseBridgeAdapter. Currently this involves adding event querying logic for any L1-->L2 transfers through a bridge contract which involves adding logic to handle two events:

1. Initiating the deposit on L1
2. Receiving the deposit on L2

Despite each bridge event looking different, we attempt to unify the logic (so that we can easily track the status of these deposits across different bridges) by implementing an interface for each bridge:
- `queryL1BridgeInitiationEvents()`
- `queryL2BridgeFinalizationEvents()`

These functions ideally should return the same object type, `BridgeEvent` that we can use to then implement `getOutstandingCrossChainTransfers`, which is used by the rebalancer and monitor. To make this work, we modify all `BridgeEvents` to have the same properties: `from`, `to` and `amount`. The latter property, `amount` is super important to make `getOutstandingCrossChainTransfers` work but the former two, `from` and `to` are not actually used anywhere in the code.

This PR removes `from` and `to` from `BridgeEvent` which marginally reduces the work to add a new event.

### DIFF
--- a/src/adapter/bridges/ZKStackBridge.ts
+++ b/src/adapter/bridges/ZKStackBridge.ts
@@ -157,7 +157,7 @@ export class ZKStackBridge extends BaseBridgeAdapter {
       );
       processedEvents = rawEvents
         .filter((event) => compareAddressesSimple(event.args.receiver, toAddress))
-        .map((e) => processEvent(e, "amount", "receiver", "sender"));
+        .map((e) => processEvent(e, "amount"));
     } else {
       if (isL2Contract) {
         const rawEvents = await paginatedEventQuery(this.hubPool, this.hubPool.filters.TokensRelayed(), eventConfig);
@@ -167,7 +167,7 @@ export class ZKStackBridge extends BaseBridgeAdapter {
           )
           .map((e) => {
             return {
-              ...processEvent(e, "amount", "to", "to"),
+              ...processEvent(e, "amount"),
               from: this.hubPool.address,
             };
           });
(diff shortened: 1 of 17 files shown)


src/adapter/bridges/ZKStackWethBridge.ts
```diff
@@ -118,7 +118,7 @@ export class ZKStackWethBridge extends ZKStackBridge {
         .filter((e) => compareAddressesSimple(e.args.to, toAddress) && compareAddressesSimple(e.args.l1Token, l1Token))
         .map((e) => {
           return {
-            ...processEvent(e, "amount", "to", "to"),
+            ...processEvent(e, "amount"),
             from: this.hubPool.address,
```
bmzig, line 122:
```suggestion
```

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -222,7 +222,7 @@ export class ZKStackBridge extends BaseBridgeAdapter {
         .filter((event) => compareAddressesSimple(event.args.receiver, toAddress))
         .map((event) => {
           return {
-            ...processEvent(event, "amount", "receiver", "receiver"),
+            ...processEvent(event, "amount"),
             from: isSpokePool ? this.hubPool.address : fromAddress,
```
bmzig, line 226:
```suggestion
```

src/adapter/bridges/ZKStackBridge.ts
```diff
@@ -179,7 +179,7 @@ export class ZKStackBridge extends BaseBridgeAdapter {
         );
         processedEvents = rawEvents.map((e) => {
           return {
-            ...processEvent(e, "amount", "from", "from"),
+            ...processEvent(e, "amount"),
             to: toAddress, // Overwrite the from field with toAddress since if we hit this branch we know an EOA initiated the bridge.
```
bmzig, line 183:
```suggestion
```

bmzig (approved):

PR #2177, opened 2025-04-08
feat: Handle cases where a deposit's inputToken and originChain cannot be mapped to a PoolRebalanceRoute

nicholaspai:
Relayer:
- Refuses to fill such deposits

Dataworker:
- Does not let such deposits affect bundle LP fees and running balance calculations

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -237,26 +237,37 @@ export function _buildRelayerRefundRoot(
   Object.entries(combinedRefunds).forEach(([_repaymentChainId, refundsForChain]) => {
     const repaymentChainId = Number(_repaymentChainId);
     Object.entries(refundsForChain).forEach(([l2TokenAddress, refunds]) => {
-      const l1TokenCounterpart = clients.hubPoolClient.getL1TokenForL2TokenAtBlock(
-        l2TokenAddress,
-        repaymentChainId,
-        endBlockForMainnet
-      );
+      // If the token cannot be mapped to any PoolRebalanceRoute, then the amount to return must be 0 since there
+      // is no way to send the token back to the HubPool.
+      if (!clients.hubPoolClient.l2TokenHasPoolRebalanceRoute(l2TokenAddress, repaymentChainId)) {
+        relayerRefundLeaves.push(..._getRefundLeaves(
+          refunds,
+          bnZero,
+          repaymentChainId,
+          l2TokenAddress,
+          maxRefundCount
+        ));
+      } else {
+        const l1TokenCounterpart = clients.hubPoolClient.getL1TokenForL2TokenAtBlock(
+          l2TokenAddress,
+          repaymentChainId,
+          endBlockForMainnet
+        );
+  
+        const spokePoolTargetBalance = clients.configStoreClient.getSpokeTargetBalancesForBlock(
+          l1TokenCounterpart,
+          repaymentChainId,
+          endBlockForMainnet
+        );
+  
+        // The `amountToReturn` for a { repaymentChainId, L2TokenAddress} should be set to max(-netSendAmount, 0).
+        const amountToReturn = getAmountToReturnForRelayerRefundLeaf(
+          spokePoolTargetBalance,
+          runningBalances[repaymentChainId][l1TokenCounterpart]
+        );
 
-      const spokePoolTargetBalance = clients.configStoreClient.getSpokeTargetBalancesForBlock(
-        l1TokenCounterpart,
-        repaymentChainId,
-        endBlockForMainnet
-      );
-
-      // The `amountToReturn` for a { repaymentChainId, L2TokenAddress} should be set to max(-netSendAmount, 0).
-      const amountToReturn = getAmountToReturnForRelayerRefundLeaf(
-        spokePoolTargetBalance,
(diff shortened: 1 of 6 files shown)


src/dataworker/DataworkerUtils.ts
```diff
@@ -237,26 +237,35 @@ export function _buildRelayerRefundRoot(
...
-        l2TokenAddress,
-        repaymentChainId,
-        endBlockForMainnet
-      );
+      // If the token cannot be mapped to any PoolRebalanceRoute, then the amount to return must be 0 since there
+      // is no way to send the token back to the HubPool.
+      if (!clients.hubPoolClient.l2TokenHasPoolRebalanceRoute(l2TokenAddress, repaymentChainId)) {
+        relayerRefundLeaves.push(
+          ..._getRefundLeaves(refunds, bnZero, repaymentChainId, l2TokenAddress, maxRefundCount)
+        );
```
pxrl, line 245: Should we just return after this so we can avoid the `else` clause? It saves having to indent all the code below.

src/relayer/Relayer.ts
```diff
@@ -292,6 +292,15 @@ export class Relayer {
...
     }
 
+    if (!hubPoolClient.l2TokenHasPoolRebalanceRoute(inputToken, originChainId)) {
+      this.logger.debug({
+        at: "Relayer::filterDeposit",
+        message: `Skipping ${srcChain} deposit for input token ${inputToken} due to missing pool rebalance route.`,
+        transactionHash: deposit.transactionHash,
+      });
+      return ignoreDeposit();
+    }
```
pxrl, line 302: I think this fits better up at line 236 (i.e. above `if (!this.routeEnabled(...))`).

src/relayer/Relayer.ts
```diff
@@ -715,6 +724,16 @@ export class Relayer {
...
+    if (!hubPoolClient.l2TokenHasPoolRebalanceRoute(inputToken, originChainId)) {
+      this.logger.debug({
+        at: "Relayer::evaluateFill",
+        message: `Skipping ${originChain} deposit ${depositId.toString()} for input token ${
+          deposit.inputToken
+        } due to missing pool rebalance route.`,
+        transactionHash,
+      });
+      return;
+    }
```
pxrl, line 736: Is this change needed? All deposits should be passed through `Relayer.filterDeposit()` before they are seen by `Relayer.evlauateFill()`, so we should already be guaranteed to have a route in place.

src/relayer/Relayer.ts
```diff
@@ -976,7 +995,16 @@ export class Relayer {
...
-    if (deposit.fromLiteChain || deposit.toLiteChain) {
+    if (
+      deposit.fromLiteChain ||
+      deposit.toLiteChain ||
+      !this.clients.hubPoolClient.areTokensEquivalent(
+        deposit.inputToken,
+        deposit.originChainId,
+        deposit.outputToken,
+        deposit.destinationChainId
+      )
```
pxrl, line 1006: I think this should already be handled at line 317 in `Relayer.filterDeposit()`, where we validate `outputToken`. I'm not 100% sure, but I _think_ that should mean we never trigger on this evaluation.

nicholaspai: you're right!

test/Dataworker.loadData.deposit.ts
```diff
@@ -250,6 +250,26 @@ describe("Dataworker: Load bundle data", async function () {
       expect(data1.expiredDepositsToRefundV3).to.deep.equal({});
     });
 
+    it("Can refund deposits with output token as expired deposit", async function () {
```
nicholaspai, line 253:
```suggestion
    it("Can refund expired deposits with unresolvable output token", async function () {
```

test/Dataworker.loadData.deposit.ts
```diff
@@ -250,6 +250,26 @@ describe("Dataworker: Load bundle data", async function () {
...
 
+    it("Can refund expired deposits with unresolvable output token", async function () {
+      const bundleBlockTimestamps = await dataworkerInstance.clients.bundleDataClient.getBundleBlockTimestamps(
+        [originChainId, destinationChainId],
+        getDefaultBlockRange(5),
+        spokePoolClients
+      );
+
+      generateV3Deposit({
+        inputToken: ZERO_ADDRESS,
```
nicholaspai, line 261:
```suggestion
        outputToken: ZERO_ADDRESS,
```

test/InventoryClient.RefundChain.ts
```diff
@@ -580,90 +577,4 @@ describe("InventoryClient: Refund chain selection", async function () {
       expect(possibleRepaymentChains.length).to.equal(4);
     });
   });
-
-  describe("In-protocol swap", async function () {
```
nicholaspai, line 584: I think this test isn't doing what we think its doing currently. Currently the `determineRefundChain` should throw if input token and output token are not "equivalent" [here](https://github.com/across-protocol/relayer/blob/96b61320cce08525a051d995d264a9625f9ac725/src/clients/InventoryClient.ts#L438) and in this test, the input token is native USDC which has no pool rebalance route mappings to bridged USDC, so it should not allow `determineRefundChain` to return successfully.

bmzig (approved):

PR #2179, opened 2025-04-10
chore: Bump SDK

pxrl:
This permits runtime updating of the Risk Labs address list.

### DIFF
--- a/scripts/fetch-addresses.ts
+++ b/scripts/fetch-addresses.ts
@@ -7,7 +7,7 @@ const OUTPUT_PATH = "addresses.json";
 let logger: typeof Logger;
 
 async function run(): Promise<number> {
-  const addressList = new AddressAggregator([new addressAdapters.risklabs.AddressList()], logger);
+  const addressList = new AddressAggregator([new addressAdapters.risklabs.AddressList({ throwOnError: false })], logger);
   const addresses = await addressList.update();
   await writeFile(OUTPUT_PATH, JSON.stringify(addresses, null, 4));
   console.log(`Stored ${addresses.size} addresses at ${OUTPUT_PATH}.`);
--- a/src/common/Config.ts
+++ b/src/common/Config.ts
@@ -146,7 +146,13 @@ export class CommonConfig {
       return Promise.resolve();
     }
 
-    const addressAggregator = new AddressAggregator([new addressAdapters.fs.AddressList(path)], logger);
+    const addressAggregator = new AddressAggregator(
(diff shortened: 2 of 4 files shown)


pxrl:
nb. there's no rush on getting this PR in; I'm only submitting it because this changes are needed as part of the next time we bump.

src/common/Config.ts
```diff
@@ -146,7 +146,13 @@ export class CommonConfig {
       return Promise.resolve();
     }
 
-    const addressAggregator = new AddressAggregator([new addressAdapters.fs.AddressList(path)], logger);
+    const addressAggregator = new AddressAggregator(
+      [
+        new addressAdapters.fs.AddressList({ path }),
+        new addressAdapters.risklabs.AddressList({ path, throwOnError: false }),
```
pxrl, line 152:
```suggestion
        new addressAdapters.risklabs.AddressList({ throwOnError: false }),
```

src/common/Config.ts
```diff
@@ -146,7 +146,13 @@ export class CommonConfig {
       return Promise.resolve();
     }
 
-    const addressAggregator = new AddressAggregator([new addressAdapters.fs.AddressList(path)], logger);
+    const addressAggregator = new AddressAggregator(
+      [
+        new addressAdapters.fs.AddressList({ path }),
+        new addressAdapters.risklabs.AddressList({ throwOnError: false }),
```
pxrl, line 152: tldr try to pull down the latest address list (permitting up to 1 retry by default), but suppress errors if it can't be retrieved. In this case, the build-time `addresses.json` should still include a reasonably-recent list of addresses that can be used as a fallback.

pxrl, line 152:
```suggestion
        new addressAdapters.fs.AddressList({ path, logger}),
        new addressAdapters.risklabs.AddressList({ logger, throwOnError: false }),
```

scripts/fetch-addresses.ts
```diff
@@ -7,7 +7,10 @@ const OUTPUT_PATH = "addresses.json";
 let logger: typeof Logger;
 
 async function run(): Promise<number> {
-  const addressList = new AddressAggregator([new addressAdapters.risklabs.AddressList()], logger);
+  const addressList = new AddressAggregator(
+    [new addressAdapters.risklabs.AddressList({ throwOnError: false })],
```
pxrl, line 11: Don't permit a failure to retrieve the address list to foul the build.

PR #2189, opened 2025-04-18
improve(InventoryClient): Normalize l2 chain amounts to correct decimals if l2 token decimals differs from l1 token equivalent's

nicholaspai:
Currently the InventoryClient would not be able to compute allocation %'s correctly if an L2 token had a different decimal than the L1 token.

This PR normalizes the following L2 values to the same decimals as L1 so that the internal inventory calculations work:
- chain virtual balance
- cumulative virtual balance
- short falls when a deposit is outstanding for a chain
- repayment amounts (relevant for `determineRefundChain` for a Deposit)
- upcoming bundle refunds

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -747,13 +747,15 @@ export class InventoryClient {
     l1Token: string,
     refundAmount: BigNumber
   ): { [chainId: number]: BigNumber } {
+    const l1TokenDecimals = this.hubPoolClient.getTokenInfoForL1Token(l1Token).decimals;
     const pcts = Object.fromEntries(
-      Object.entries(excessRunningBalances).map(([chainId, excess]) => {
-        const target = this.hubPoolClient.configStoreClient.getSpokeTargetBalancesForBlock(
-          l1Token,
-          Number(chainId)
-        ).target;
-        const excessPostRelay = excess.sub(refundAmount);
+      Object.entries(excessRunningBalances).map(([_chainId, excess]) => {
+        const chainId = Number(_chainId);
+        const target = this.hubPoolClient.configStoreClient.getSpokeTargetBalancesForBlock(l1Token, chainId).target;
+        const l2Token = this.hubPoolClient.getL2TokenForL1TokenAtBlock(l1Token, chainId);
+        const l2TokenDecimals = this.hubPoolClient.getTokenInfoForAddress(l2Token, chainId).decimals;
+        const refundAmountInL1TokenDecimals = sdkUtils.ConvertDecimals(l2TokenDecimals, l1TokenDecimals)(refundAmount);
+        const excessPostRelay = excess.sub(refundAmountInL1TokenDecimals);
         const returnObj = {
           pct: toBN(0),
           target,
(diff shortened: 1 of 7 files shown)


src/clients/InventoryClient.ts
bmzig, line 1: I think we also need to normalize all calls which take balance info from the token client, right? Specifically:
- [This shortfall](https://github.com/across-protocol/relayer/blob/9e8f7174d7db818c477b82b8cd561724194779ea/src/clients/InventoryClient.ts#L206) would be in l2 token decimals, so when we sub from `this.getBalanceOnChain`, it will probably be a negative number (if L2 decimals > L1 decimals)
- I believe the same is true for [this shortfall](https://github.com/across-protocol/relayer/blob/9e8f7174d7db818c477b82b8cd561724194779ea/src/clients/InventoryClient.ts#L521)
- [This call](https://github.com/across-protocol/relayer/blob/9e8f7174d7db818c477b82b8cd561724194779ea/src/clients/InventoryClient.ts#L1169) looks like it could cause a formatting error in the log only.
- Same log issue [with this](https://github.com/across-protocol/relayer/blob/9e8f7174d7db818c477b82b8cd561724194779ea/src/clients/InventoryClient.ts#L1338).

nicholaspai: Ah these are great catches. I think we also need to normalize upcoming bundle refunds and the repayment amount which I've added in this commit: d39c22dfff13f7439a86d10aa1f1c6ace25bee2d

bmzig: Ah yeah I didn't even check those the first time around. I did another pass through the inventory client file and didn't find anything major. I'm just saying (mainly for myself) that when we do support rebalancing, the crossChainTransferClient's `getOutstandingCrossChainTransferAmount` will need to return an amount in terms of the l1 decimals (which makes sense anyways). 
- I think [this](https://github.com/across-protocol/relayer/blob/d39c22dfff13f7439a86d10aa1f1c6ace25bee2d/src/clients/InventoryClient.ts#L944) and [this](https://github.com/across-protocol/relayer/blob/d39c22dfff13f7439a86d10aa1f1c6ace25bee2d/src/clients/InventoryClient.ts#L973) would now log weird since the formatter would be interpreting a smaller, normalized amount as having more decimals than in reality (since the formatter is using L2 decimals).

src/clients/InventoryClient.ts
```diff
@@ -520,7 +540,7 @@ export class InventoryClient {
         chainId === destinationChainId &&
         this.hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)
           ? chainVirtualBalanceWithShortfall
-          : chainVirtualBalanceWithShortfall.add(inputAmount);
+          : chainVirtualBalanceWithShortfall.add(l2BalanceToL1Decimals(inputAmount));
```
bmzig, line 543: I believe this needs to be something like "otherL2BalanceToL1Decimals" where the closure does `sdkUtils.convertDecimals(originChainTokenDecimals, l1TokenDecimals)` since inputAmount will be in terms of the origin chain's token decimals, while l2BalanceToL1Decimals is converting the repayment chain's token decimals to l1 decimals.

nicholaspai: You're totally right https://github.com/across-protocol/relayer/pull/2189/commits/30bf069c7e59a3798631ed4efcb6976ee86c8171

src/clients/InventoryClient.ts
nicholaspai: > I think [this](https://github.com/across-protocol/relayer/blob/d39c22dfff13f7439a86d10aa1f1c6ace25bee2d/src/clients/InventoryClient.ts#L944) and [this](https://github.com/across-protocol/relayer/blob/d39c22dfff13f7439a86d10aa1f1c6ace25bee2d/src/clients/InventoryClient.ts#L973) would now log weird since the formatter would be interpreting a smaller, normalized amount as having more decimals than in reality (since the formatter is using L2 decimals).

More good catches, fixed here: 11999704e6c3c7685a378b1261b4c1f379548e65

bmzig: Alright one more run through:
- I believe [this](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L956) should be `l1Formatter` since `amount` is the l1 send amount. I believe this same switch to `l1Formatter` also applies to [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L986), [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L987), and [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L993).

Some nits, but I think it's worth mentioning [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L1246) that we expect an amount in L2 decimals, since it's different from a "normalized" balance.
It may also be nice to have a comment [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L795) which says that it is assumed that `chainsToEvaluate` contains chains where, for the `l1Token`, the `l2Token` has the same number of decimals, since the function will break without that assumption.

I suspect this is everything but I'll give it one more run through. I still didn't see any logic issues.

nicholaspai: > It may also be nice to have a comment [here](https://github.com/across-protocol/relayer/blob/fa70f6380ac6b2bd932b5e9a6b2a80e98726703f/src/clients/InventoryClient.ts#L795) which says that it is assumed that chainsToEvaluate contains chains where, for the l1Token, the l2Token has the same number of decimals, since the function will break without that assumption.

This is a good catch. I think we can probably fix this by converting the `inputAmount` to the L1 token decimals that we pass into `getExcessRunningBalancePcts()`, right? Something like 4b5df732123609cb0ce9176c1470202ef6292bd9

bmzig (approved):

PR #2190, opened 2025-04-24
improve: Follow up fixes to use correct precision on output token amounts

nicholaspai:
- ProfitClient assumes in a few places that output amount is denominated in same precision as equivalent L1 token
- Monitor will not currently report balances of L2 tokens that are not mapped to L1 tokens, which I've left a TODO comment for
- PoolRebalance logs might not render correctly without adjusting for output token decimals

This PR also replaces use of functions in the SDK that might not resolve the correct `L1TokenInfo` object if different entries in TOKEN_SYMBOLS_MAP have the same mainnet token. These functions are removed in this SDK [pr](https://github.com/across-protocol/sdk/pull/981/files) and this PR uses the safer SDK functions that grabs the correct L2 chain's decimals

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -403,23 +403,12 @@ export class ProfitClient {
     deposit: Pick<Deposit, "destinationChainId" | "outputToken" | "outputAmount">
   ): BigNumber | undefined {
     const { destinationChainId, outputToken, outputAmount } = deposit;
-    let l1Token: L1Token;
 
-    try {
-      l1Token = this.hubPoolClient.getL1TokenInfoForL2Token(outputToken, destinationChainId);
-    } catch {
-      this.logger.debug({
-        at: "ProfitClient#getFillAmountInUsd",
-        message: `Cannot resolve output token ${outputToken} on ${getNetworkName(destinationChainId)}.`,
-      });
-      return undefined;
-    }
-
-    const tokenPriceInUsd = this.getPriceOfToken(l1Token.symbol);
+    const { symbol, decimals } = this.hubPoolClient.getTokenInfoForAddress(outputToken, destinationChainId);
+    const tokenPriceInUsd = this.getPriceOfToken(symbol);
 
     // The USD amount of a fill must be normalised to 18 decimals, so factor out the token's own decimal promotion.
-    const outputTokenDecimals = this.hubPoolClient.getTokenInfoForAddress(outputToken, destinationChainId).decimals;
-    return outputAmount.mul(tokenPriceInUsd).div(bn10.pow(outputTokenDecimals));
+    return outputAmount.mul(tokenPriceInUsd).div(bn10.pow(decimals));
   }
 
   async getFillProfitability(
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -331,7 +331,7 @@ export class ProfitClient {
   ): Promise<FillProfit> {
     const { hubPoolClient } = this;
 
-    const inputTokenInfo = hubPoolClient.getL1TokenInfoForL2Token(deposit.inputToken, deposit.originChainId);
+    const inputTokenInfo = hubPoolClient.getTokenInfoForAddress(deposit.inputToken, deposit.originChainId);
     const inputTokenPriceUsd = this.getPriceOfToken(inputTokenInfo.symbol);
     const inputTokenScalar = toBNWei(1, 18 - inputTokenInfo.decimals);
     const scaledInputAmount = deposit.inputAmount.mul(inputTokenScalar);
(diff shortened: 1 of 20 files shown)


src/monitor/Monitor.ts
```diff
@@ -253,6 +253,7 @@ export class Monitor {
       address: TOKEN_SYMBOLS_MAP["USDC.e"].addresses[this.clients.hubPoolClient.chainId],
       decimals: 6,
     });
+    // @dev TODO: Handle special case for tokens that do not have an L1 token mapped to them via PoolRebalanceRoutes
```
nicholaspai, line 256: we need to either hardcode additional tokens to the `allL1Tokens` array here or change how this function works and read from some hard coded list. Going forwards we won't always be able to assume that all L2 balances have an "equivalent" L1 token

src/monitor/Monitor.ts
```diff
@@ -253,6 +253,7 @@ export class Monitor {
       address: TOKEN_SYMBOLS_MAP["USDC.e"].addresses[this.clients.hubPoolClient.chainId],
       decimals: 6,
     });
```
pxrl, line 255: Driveby on the above - are we potentially missing Base USDbC and Zora USDzC in this reporting?

src/monitor/Monitor.ts
pxrl: I considered whether we could scrape previous RelayerRefundRoot executions to use the token addresses from there, but then we have to query every single event since genesis, so that seems like a non-starter. I can't think of any other way to infer the complete set of relevant tokens dynamically. So for now it seems like we'll need to make this configurable via the environment, and will have to update that every time we support each additional unmapped token. wdyt?

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -144,14 +144,13 @@ export function generateMarkdownForRootBundle(
...
-      hubPoolClient.getTokenInfo(leaf.chainId, leaf.l2TokenAddress).decimals
+      hubPoolClient.getTokenInfoForAddress(leaf.l2TokenAddress, leaf.chainId).decimals
     );
     leaf.refundAmounts = convertTokenListFromWei(
-      leaf.chainId,
       Array(leaf.refundAmounts.length).fill(leaf.l2TokenAddress),
       leaf.refundAmounts
     );
-    leaf.l2Token = convertTokenAddressToSymbol(leaf.chainId, leaf.l2TokenAddress);
+    leaf.l2Token = convertTokenAddressToSymbol(leaf.l2TokenAddress);
```
pxrl, line 153: `convertTokenAddressToSymbol()` seems to assume that it receives an L1 token, but we could supply an address from any chain here. Should we use `HubPoolClient.getTokenInforForAddress()` instead? We already rely on it for sourcing decimals correctly.

```suggestion
    // At line 145
    const { symbol, decimals } = hubPoolClient.getTokenInfoForAddress(leaf.l2TokenAddress, leaf.chainId);
    leaf.amountToReturn = convertFromWei(..., decimals);
 
    // ...
    
    // At line 153
    leaf.l2Token = symbol;
```

src/monitor/Monitor.ts
nicholaspai: we might be yeah

src/monitor/Monitor.ts
nicholaspai: I do think reading from the inventory config makes sense--the only drawback is having to maintain an inventory config in yet another place.

src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -144,10 +144,9 @@ export function generateMarkdownForRootBundle(
     delete leaf.leafId;
     leaf.amountToReturn = convertFromWei(
       leaf.amountToReturn,
-      hubPoolClient.getTokenInfo(leaf.chainId, leaf.l2TokenAddress).decimals
+      hubPoolClient.getTokenInfoForAddress(leaf.l2TokenAddress, leaf.chainId).decimals
     );
     leaf.refundAmounts = convertTokenListFromWei(
```
bmzig, line 149: Does this work now? Seems like it takes in an L2 token amount and outputs it in L1 decimals.

src/clients/ProfitClient.ts
```diff
@@ -331,32 +330,18 @@ export class ProfitClient {
...
-        deposit.destinationChainId
-      ));
-    } else {
-      // This function will throw if the token is not found in the TOKEN_SYMBOLS_MAP for the destination chain.
-      ({ symbol: outputTokenSymbol, decimals: outputTokenDecimals } = hubPoolClient.getTokenInfoForAddress(
-        deposit.outputToken,
-        deposit.destinationChainId
-      ));
-    }
+    const { symbol: outputTokenSymbol, decimals: outputTokenDecimals } = hubPoolClient.getTokenInfoForAddress(
```
bmzig, line 341: Just want to confirm. The reason why this is OK to do is since any deposit we have here has an `outputToken` in `TOKEN_SYMBOLS_MAP`?

src/dataworker/PoolRebalanceUtils.ts
nicholaspai: Does it? These are all L2 decimals/amounts right?

src/clients/ProfitClient.ts
nicholaspai: yea that's my thinking. We shouldn't ever get here with a deposit that has a token that isn't in TOKEN_SYMBOLS_MAP

src/dataworker/PoolRebalanceUtils.ts
bmzig: Yeah I think so, so when we call `convertTokenListFromWei` w/ l2 tokens and l2 amounts, then [wouldn't we be calling](https://github.com/across-protocol/relayer/blob/f205c780a927eaef0bb35818e0c54d91914753eb/src/dataworker/PoolRebalanceUtils.ts#L115) `getTokenInfoForL1Token` on l2 token addresses?

nicholaspai: Ah shoot good call  7a5f4a129ece5a5f8979da106d1ceefd5645aff3

PR #2195, opened 2025-04-28
feat: Binance rebalancing

bmzig:
This should let us send rebalances from mainnet to BNB via Binance. We need to query the API to get deposit addresses/find in-transit transfers and finalize L2 withdrawals.


src/adapter/bridges/BinanceCEXBridge.ts
```diff
@@ -0,0 +1,155 @@
...
+  async constructL1ToL2Txn(
+    _toAddress: EvmAddress,
+    l1Token: EvmAddress,
+    _l2Token: EvmAddress,
+    amount: BigNumber
+  ): Promise<BridgeTransactionDetails> {
+    assert(l1Token.toAddress() === this.getL1Bridge().address);
+    // Fetch the deposit address from the binance API.
+
+    const depositAddress = await this.binanceApiClient.depositAddress({
```
nicholaspai, line 58: Does this address change often or is it fixed until we do something in the binance account?

src/clients/TokenClient.ts
```diff
@@ -36,7 +37,8 @@ export class TokenClient {
     readonly logger: winston.Logger,
     readonly relayerAddress: string,
     readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
-    readonly hubPoolClient: HubPoolClient
+    readonly hubPoolClient: HubPoolClient,
+    readonly additionalTokens: string[] = []
```
nicholaspai, line 41: nit: should maybe name this additionalL1Tokens to make it clear these should be l1 tokens

src/common/Constants.ts
```diff
@@ -327,6 +329,7 @@ export const SUPPORTED_TOKENS: { [chainId: number]: string[] } = {
   [CHAIN_IDs.ARBITRUM]: ["USDC", "USDT", "WETH", "DAI", "WBTC", "UMA", "BAL", "ACX", "POOL"],
   [CHAIN_IDs.BASE]: ["BAL", "DAI", "ETH", "WETH", "USDC", "POOL"],
   [CHAIN_IDs.BLAST]: ["DAI", "WBTC", "WETH"],
+  [CHAIN_IDs.BSC]: ["CAKE", "BNB", "USDC", "USDT", "ETH"],
```
nicholaspai, line 332: not WETH? or is it named ETH on BSC?

src/clients/TokenClient.ts
bmzig: I agree. Renamed https://github.com/across-protocol/relayer/pull/2195/commits/eb4cb544c272062fd20d1db9e8fc8060e7e1e3b5

src/adapter/bridges/BinanceCEXBridge.ts
bmzig: I don't think it changes often, but I'm pretty sure it can change, since the [response](https://developers.binance.com/docs/wallet/capital/deposite-address) gives an `isDefault` flag and it's set to `false` from these requests. There might be something to configure in a binance account, though. I just think it's safe to always request a deposit address.

src/common/Constants.ts
bmzig: It is named ETH on BSC. ETH rebalancing will probably have to be done via the atomic depositor, which I'd argue should be added in a follow up so that this PR doesn't get an even bigger diff.

nicholaspai (approved):

nicholaspai (approved):

nicholaspai (approved):

PR #2196, opened 2025-04-28
feat: Treat lite chain deposits and forced origin chain repayments similarly

nicholaspai:
## Relayer and InventoryClient changes

If a deposit's inputToken is not mapped to a PoolRebalanceRoute mapping then the fill must take repayment on origin chain, which is similar to a lite chain deposit.

There is a lot of logic in the Relayer and InventoryClient that deals with lite chain deposits that should therefore also apply to these "forced origin chain repayment" deposits.

## Replacing token equivalency mapping lookups

Additionally, this PR replaces more token matching lookups via PoolRebalanceRoute events with looking up in the TOKEN_SYMBOLS_MAP constants file. This is required to make all bots be able to support deposited tokens that do not map to any L1 token via PoolRebalanceRoutes

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -30,7 +30,6 @@ import {
   getL1TokenInfo,
   depositForcesOriginChainRepayment,
   getRemoteTokenForL1Token,
-  depositCanTakeDestinationChainRepayment,
 } from "../utils";
 import { HubPoolClient, TokenClient, BundleDataClient } from ".";
 import { Deposit, L1Token, ProposedRootBundle } from "../interfaces";
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1,4 +1,4 @@
-import { constants, utils as sdkUtils } from "@across-protocol/sdk";
+import { utils as sdkUtils } from "@across-protocol/sdk";
 import WETH_ABI from "../common/abi/Weth.json";
 import {
   bnZero,
@@ -23,8 +23,6 @@ import {
   MAX_UINT_VAL,
   toBNWei,
   assert,
-  compareAddressesSimple,
-  getUsdcSymbol,
   Profiler,
   getNativeTokenSymbol,
   getL1TokenInfo,
@@ -54,7 +52,6 @@ export type Rebalance = {
   amount: BigNumber;
 };
 
-const { CHAIN_IDs } = constants;
 const DEFAULT_TOKEN_OVERAGE = toBNWei("1.5");
 
 export class InventoryClient {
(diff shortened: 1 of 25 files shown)


src/relayer/Relayer.ts
```diff
@@ -1253,7 +1245,11 @@ export class Relayer {
     //
     // Additionally we don't want to take this code path if the origin chain is a lite chain because we can't
     // reason about destination chain repayments on lite chains.
-    if (!isDefined(preferredChain) && !preferredChainIds.includes(destinationChainId) && !fromLiteChain) {
+    if (
+      !isDefined(preferredChain) &&
+      !preferredChainIds.includes(destinationChainId) &&
```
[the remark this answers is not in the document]
nicholaspai: are we?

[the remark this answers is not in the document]
bmzig: I missed the `!`. Never mind.

bmzig (dismissed):

src/clients/InventoryClient.ts
```diff
@@ -367,20 +366,39 @@ export class InventoryClient {
...
+    if (
+      this.canTakeDestinationChainRepayment(deposit) &&
+      !depositForcesOriginChainRepayment(deposit, this.hubPoolClient)
+    ) {
+      chainIds.push(destinationChainId);
+    }
 
-    if (this.isInventoryManagementEnabled()) {
-      chainIds.push(...this.getSlowWithdrawalRepaymentChains(l1Token));
+    if (this.isInventoryManagementEnabled() && originChainId !== this.hubPoolClient.chainId) {
```
bmzig, line 379: Why the second check for `originChainId` now? If we have a deposit with `originChainId = 1` and there is an overallocated slow withdrawal chain then it will be included [here](https://github.com/across-protocol/relayer/blob/ff9834d273435b5cf1d5774eb68f9499be381e28/src/clients/InventoryClient.ts#L542). It the looks like [we'd throw](https://github.com/across-protocol/relayer/blob/ff9834d273435b5cf1d5774eb68f9499be381e28/src/clients/InventoryClient.ts#L581)? Is that right?

nicholaspai: that's a good point. I actually think we want a `!depositForcesOriginChainRepayment(deposit, this.hubPoolClient)` which was my original intention. Basically we shouldn't be evaluating these other chains if we need to take repayment on origin chain. WDYT?

nicholaspai: this is a bit cleaner IMO wdyt? db0cd3d3751ba6f7f4bca7508c29e0d786f1f2dc

bmzig: Yeah I think it's much easier to read now.

bmzig (dismissed):

src/clients/InventoryClient.ts
```diff
@@ -401,6 +417,20 @@ export class InventoryClient {
       return true;
     }
 
+    // Return true if the input and output token are defined as equivalent according to a hardcoded mapping
+    // of equivalent tokens. This should allow relayer to define which tokens it should be able to fill despite them
+    // not being linked via a PoolRebalanceRoute.
+    try {
+      const l1TokenMappedToInputToken = this.getL1TokenInfo(inputToken, originChainId).address;
```
bmzig, line 424: This is minor, but I think doing `validateOutputToken` this way lets us get rid of the second part about USDC/USDC.e since in this new case, `l1TokenMappedToInputToken === l1TokenMappedToOutputToken` even if input is USDC or USDC.e and output is one of USDC/USDC.e. I think we can just `return false` if `l1TokenMappedToInputToken !== l1TokenMappedToOutputToken` now (or if it throws, as you do).

src/clients/InventoryClient.ts
```diff
@@ -247,7 +246,7 @@ export class InventoryClient {
       return Object.keys(tokenConfig).filter((k) => isDefined(tokenConfig[k][chainId]));
     }
 
-    const destinationToken = this.getRepaymentTokenForL1Token(l1Token, chainId);
```
bmzig, line 250: The comments above this function about the return value are confusing me. It says it "may also include other 'equivalent' tokens (i.e. as with Bridged & Native USDC)," but it seems this was never true?

bmzig: It says it returns an array of tokens, though. It seems like no matter what we are returning a single token address?

bmzig: Ah OK yeah I see. I think you're correct.

PR #2198, opened 2025-04-29
fix(Monitor): Handle LGHO <> WGHO conversion in reporter

nicholaspai:


bmzig (dismissed):

src/monitor/Monitor.ts
```diff
@@ -1101,6 +1101,20 @@ export class Monitor {
     relayer: string,
     balanceType: BalanceType
   ) {
+    const getL1TokenInfoForRelayerReport = (l2Token: string, chainId: number): L1Token => {
```
nicholaspai, line 1104: Refactored this a bit to make it clear we should only use this logic in this function, since it has special TOKEN_EQUIVALENCE_REMAPPING remapping logic

PR #2200, opened 2025-04-30
fix(EventUtils): Fix typo

nicholaspai:


pxrl (approved):

src/utils/EventUtils.ts
```diff
@@ -37,7 +37,7 @@ export function getUniqueLogIndex(events: { txnRef: string }[]): number[] {
   for (const event of events) {
     const logIndex = uniqueTokenhashes[event.txnRef] ?? 0;
     logIndexesForMessage.push(logIndex);
```
pxrl, line 39: fwiw, this was the alternate fix that my change would have needed.

```suggestion
    const logIndex = (uniqueTokenhashes[event.txnRef] ??= 0);
```

PR #2203, opened 2025-04-30
feat: add finalizer support for messages from `HubPool` to `UniversalSpoke`

grasphoper:
This PR adds an "adapter" for a finalizer to be able to finalize messages from `HubPool` to "Universal chains" using ZK proofs: `helios.ts`. This finalization only works in L1 -> L2 direction.

Like for other adapters, there's a main function: `heliosL1toL2Finalizer` , which does roughly the following:

### Step 1: identify messages that require finalizing

1. get events on Source Chain `HubPoolStore` that were recently stored (up to lookback)
2. get events on Dest. Chain `SP1Helios` that signify recently verified (via ZK proofs) Source Chain storage slots
3. get events on Dest. Chain `Universal_SpokePool` that signify execution of messages that were sent.

Using this information, we can identify which messages have been executed (3.) -- skip those. For other messages, if we received an event on (2.) but not (3.), the message only needs execution, and not proof verification -- so we do only that.

For all the other messages we do the full flow: ZK proof generation + `Sp1Helios.update` + `Universal_SpokePool.executeMessage`

### Step 2: request ZK proofs for the messages that have yet to be verified

1. **get_proof** from ZK_API  
   1. **If not found**  
      - Call `request_proof()`  
      - Exit (will retry next run)  
   2. **If found**, check `proofState.status`  
      1. **pending**  
         - Exit  
      2. **success**  
         - Take the proof  
         - Save it for Step 3  
      3. **errored**  
         - Produce a warning  
         - Call `request_proof()`  
         - Exit  

### Step 3: generate transactions that need to be executed.

For all messages that only require `executeMessage` -> generate a transaction.
For all messages that require both proof and `executeMessage`, only if there's proof data available already, generate 2 transactions.


src/common/abi/Universal_SpokePool.json
```diff
@@ -0,0 +1,1135 @@
...
+      { "internalType": "address", "name": "_wrappedNativeTokenAddress", "type": "address" },
+      { "internalType": "uint32", "name": "_depositQuoteTimeBuffer", "type": "uint32" },
+      { "internalType": "uint32", "name": "_fillDeadlineBuffer", "type": "uint32" },
+      { "internalType": "contract IERC20", "name": "_l2Usdc", "type": "address" },
+      { "internalType": "contract ITokenMessenger", "name": "_cctpTokenMessenger", "type": "address" }
+    ],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  { "inputs": [], "name": "AdminCallAlreadySet", "type": "error" },
```
nicholaspai, line 16: Do you need this entire ABI..? Why not just keep the functions you actually need.

src/common/abi/HubPoolStore.json
```diff
@@ -0,0 +1,43 @@
+[
+  {
+    "inputs": [{ "internalType": "address", "name": "_hubPool", "type": "address" }],
+    "stateMutability": "nonpayable",
```
nicholaspai, line 4: same comment https://github.com/across-protocol/relayer/pull/2203/files#r2070908019

src/common/abi/SP1Helios.json
```diff
@@ -0,0 +1,376 @@
+[
+  {
+    "inputs": [
+      {
+        "components": [
+          { "internalType": "bytes32", "name": "executionStateRoot", "type": "bytes32" },
```
nicholaspai, line 6: same comment https://github.com/across-protocol/relayer/pull/2203/files#r2070908019

src/finalizer/index.ts
```diff
@@ -307,25 +313,39 @@ export async function finalize(
...
-
-    if (succeed) {
-      // Increase running counter of estimated gas cost for batch finalization.
-      // gasLimit should be defined if succeed is True.
-      const updatedGasEstimation = gasEstimation.add(transaction.gasLimit);
-      if (updatedGasEstimation.lt(batchGasLimit)) {
-        gasEstimation = updatedGasEstimation;
+  const finalizations = await sdkUtils.filterAsync(finalizerResponseTxns, async ({ txn: _txn, crossChainMessage }) => {
+    let simErrorReason: string;
+    if ("callData" in _txn && "target" in _txn) {
```
nicholaspai, line 318: It's easier to maintain if you use the same condition as below to identify a Multicall vs an AugmentedTransaction so I'd prefer this `if` statement to be `if ("contract" in txn) {` like you have it below

nicholaspai: Then you can refactor into a helper function `if(isAugmentedTransaction(_txn))`

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+  if (!pendingMessages || pendingMessages.length === 0) {
+    logger.debug({
+      at: `Finalizer#heliosL1toL2Finalizer:${l2ChainId}`,
+      message: "No pending Helios messages found requiring action.",
+    });
+    return { callData: [], crossChainMessages: [] };
+  }
+
+  // Separate messages based on required action
+  const needsProofAndExecution = pendingMessages.filter((m) => m.status === "NeedsProofAndExecution");
```
nicholaspai, line 134: rather than iterate over `pendingMessages` twice you might want to use `pendingMessages.reduce` and maybe even [this function](https://github.com/across-protocol/sdk/blob/4b856bdd2cf4020ea1dde286048afe83d2d1648a/src/utils/ObjectUtils.ts#L77) can help. We use it [throughout the finalizer](https://github.com/across-protocol/relayer/blob/982600e34f8a86f06632227a86fb4b0ab477c029/src/finalizer/utils/arbStack.ts#L288)

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+  const pendingMessages = await identifyPendingHeliosMessages(
+    logger,
+    hubPoolClient,
+    l1SpokePoolClient,
+    l2SpokePoolClient,
+    l1ChainId,
+    l2ChainId
+  );
+
+  if (!pendingMessages || pendingMessages.length === 0) {
```
nicholaspai, line 125: you should probably just return an empty array when there are no pending messages, so you don't have to check for both undefined and length == 0

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+  const relevantStoredCallDataEvents = await getRelevantL1Events(
+    logger,
+    hubPoolClient,
+    l1SpokePoolClient,
+    l1ChainId,
+    l2ChainId,
+    l2SpokePoolAddress
+  );
+
+  if (!relevantStoredCallDataEvents || relevantStoredCallDataEvents.length === 0) {
```
nicholaspai, line 207: same comment above about returning empty arrays vs undefined

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+        compareAddressesSimple(event.args.target, ethers.constants.AddressZero)
+    );
+
+    logger.debug({
+      at: `Finalizer#heliosL1toL2Finalizer:getAndFilterL1Events:${l2ChainId}`,
+      message: `Found ${relevantStoredCallDataEvents.length} StoredCallData events on L1 (${l1ChainId})`,
+    });
+
+    return relevantStoredCallDataEvents;
+  } catch (error) {
```
nicholaspai, line 343: why would this fail? I think large try-catch statements are not good to keep around because they can catch unexpected errors also. We should probably just let the error crash the process as I don't see why this should ever fail

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+  const sp1HeliosContract = new ethers.Contract(sp1HeliosAddress, sp1HeliosAbi as any, l2Provider);
+
+  const l2SearchConfig: EventSearchConfig = {
+    fromBlock: l2SpokePoolClient.eventSearchConfig.fromBlock,
+    toBlock: l2SpokePoolClient.latestBlockSearched,
+    maxBlockLookBack: l2SpokePoolClient.eventSearchConfig.maxBlockLookBack,
+  };
+  const storageVerifiedFilter = sp1HeliosContract.filters.StorageSlotVerified();
+
+  try {
```
nicholaspai, line 379: same comment about large try-catches

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+): Promise<StoredCallDataEvent[] | null> {
+  const l1Provider = hubPoolClient.hubPool.provider;
+
+  const hubPoolStoreInfo = CONTRACT_ADDRESSES[l1ChainId]?.hubPoolStore;
+  if (!hubPoolStoreInfo?.address || !hubPoolStoreInfo.abi) {
+    logger.error({
+      at: `Finalizer#heliosL1toL2Finalizer:getAndFilterL1Events:${l2ChainId}`,
+      message: `HubPoolStore contract address or ABI not found for L1 chain ${l1ChainId}.`,
+    });
+    return null;
```
nicholaspai, line 301: you should throw here instead of returning null. We should always have the ABI/address available to us otherwise its the coder's error

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+      message: `Querying StoredCallData events on L1 (${l1ChainId})`,
+      hubPoolStoreAddress: hubPoolStoreInfo.address,
+      fromBlock: l1SearchConfig.fromBlock,
+      toBlock: l1SearchConfig.toBlock,
+    });
+
+    const rawLogs = await paginatedEventQuery(hubPoolStoreContract, storedCallDataFilter, l1SearchConfig);
+
+    // Explicitly cast logs to the correct type
+    const events: StoredCallDataEvent[] = rawLogs.map((log) => ({
```
nicholaspai, line 326: Try using [spreadEventWithBlockNumber](https://github.com/across-protocol/sdk/blob/4b856bdd2cf4020ea1dde286048afe83d2d1648a/src/utils/EventUtils.ts#L216) here to cast the logs to the correct type

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+): Promise<Map<string, StorageSlotVerifiedEvent> | null> {
+  const l2Provider = l2SpokePoolClient.spokePool.provider;
+
+  const { address: sp1HeliosAddress, abi: sp1HeliosAbi } = getSp1Helios(l2ChainId);
+  if (!sp1HeliosAddress || !sp1HeliosAbi) {
+    logger.error({
+      at: `Finalizer#heliosL1toL2Finalizer:getL2VerifiedKeys:${l2ChainId}`,
+      message: `SP1Helios contract not found for destination chain ${l2ChainId}. Cannot verify Helios messages.`,
+    });
+    return null;
```
nicholaspai, line 368: same comment above-- this should never be missing in production and we should crash here if so

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+    toBlock: l2SpokePoolClient.latestBlockSearched,
+    maxBlockLookBack: l2SpokePoolClient.eventSearchConfig.maxBlockLookBack,
+  };
+  const storageVerifiedFilter = sp1HeliosContract.filters.StorageSlotVerified();
+
+  try {
+    const rawLogs = await paginatedEventQuery(sp1HeliosContract, storageVerifiedFilter, l2SearchConfig);
+
+    // Explicitly cast logs to the correct type
+    const events: StorageSlotVerifiedEvent[] = rawLogs.map((log) => ({
```
nicholaspai, line 383: same comment on `spreadEventWithBlockNumber`

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+    logger.debug({
+      at: `Finalizer#heliosL1toL2Finalizer:getL2VerifiedSlotsMap:${l2ChainId}`,
+      message: `Found ${events.length} StorageSlotVerified events on L2 (${l2ChainId})`,
+      sp1HeliosAddress,
+      fromBlock: l2SearchConfig.fromBlock,
+      toBlock: l2SearchConfig.toBlock,
+    });
+
+    // Store events in a map keyed by the storage slot (key)
+    const verifiedSlotsMap = new Map<string, StorageSlotVerifiedEvent>();
```
nicholaspai, line 397: Why use a `Map` over a `Set`?

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+    logger.debug({
+      at: `Finalizer#heliosL1toL2Finalizer:getL2RelayedNonces:${l2ChainId}`,
+      message: `Found ${events.length} RelayedCallData events on L2 (${l2ChainId})`,
+      spokePoolAddress: l2SpokePoolAddress,
+      fromBlock: l2SearchConfig.fromBlock,
+      toBlock: l2SearchConfig.toBlock,
+    });
+
+    // Return a Set of nonces (as strings for easy comparison)
+    return new Set<string>(events.map((event) => event.args.nonce.toString()));
```
nicholaspai, line 456: why make this a `Set`? Would there ever be a duplicate event? We can assume the contracts work as they work and that the RPC's won't give us duplicate events. I've never seen a duplicate event--only missing events

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,927 @@
...
+
+  return { callData: transactions, crossChainMessages: crossChainMessages };
+}
+
+/**
+ * Calculates the storage slot in the HubPoolStore contract for a given nonce.
+ * This assumes the data is stored in a mapping at slot 0, keyed by nonce.
+ * storage_slot = keccak256(h(k) . h(p)) where k = nonce, p = mapping slot position (0) ✅ tested
+ */
+function calculateHubPoolStoreStorageSlot(eventArgs: StoredCallDataEventArgs): string {
```
nicholaspai, line 892: You might want to move these stateless utility functions to a `utils` file, this file is already quite large

src/utils/TransactionUtils.ts
```diff
@@ -16,9 +16,13 @@ import {
   toBNWei,
   winston,
   stringifyThrownValue,
+  CHAIN_IDs,
 } from "../utils";
 dotenv.config();
 
+// Define chains that require legacy (type 0) transactions
+const LEGACY_TRANSACTION_CHAINS = new Set<number>([CHAIN_IDs.BSC]);
```
nicholaspai, line 24: no need for a `Set` here

src/utils/TransactionUtils.ts
```diff
@@ -81,13 +85,24 @@ export async function runTransaction(
...
       priorityFeeScaler,
       maxFeePerGasScaler,
       await contract.populateTransaction[method](...(args as Array<unknown>), { value })
     );
 
+    // Check if the chain requires legacy transactions
+    if (LEGACY_TRANSACTION_CHAINS.has(chainId)) {
+      gas = { gasPrice: gas.maxFeePerGas };
+      logger.debug({
+        at: "TxUtil#runTransaction",
```
nicholaspai, line 99: I don't think we need this `debug` log in production --it'll get really noisy

nicholaspai (commented):
This is a great start. The code is very organized. I think there are some broad comments I want to make before I submit more nitty-gritty comments:
- We should crash/throw errors as early as possible. Especially when they are programmer errors versus run-time errors. A lot of the errors you are trying to catch in your code are programmer errors that should never exist in a working production system.
- We should avoid try-catch statements if at all possible. They are dangerous and can hide bugs added in the future.

grasphoper:
> We should crash/throw errors as early as possible

I agree. I usually tend to prefer throwing as well.

We had a similar discussion recently with @pxrl and @bmzig [here](https://github.com/across-protocol/relayer/pull/2133) which led me in the direction of logging errors more instead of throwing. This strategy has a nice property of having the code still complete a run even if it had an error in one of the parts, while reporting an error to the logging channels and alerting someone on PD.

I wonder about our general philosophy of error handling. E.g. in go, rust errors are explicit. I.e. you can return an error type from a function. In TS, it feels like returning `null` is the "error return". @pxrl mentioned that sometimes returning `T | Error` makes sense, but I didn't see it yet in the code.

src/finalizer/utils/helios.ts
grasphoper: `null` is returned when there's an error in some of the functions. I could return an empty array there. Do you think it's better?

src/finalizer/utils/helios.ts
grasphoper: Agree, this shouldn't be here.

grasphoper: I think the try-catch was meant to catch RPC call problems, but not sure `paginatedEventQuery` can ever throw.

src/finalizer/utils/helios.ts
grasphoper: I take that back, the try-catch in the internal fn is not needed

src/finalizer/utils/helios.ts
grasphoper: Actually, now that I look at `paginatedEventQuery` it can throw. This would mean that if out BSC RPCs are down, we'd crash the whole finalizer if we remove this try-catch. Do we want that?

src/finalizer/utils/helios.ts
grasphoper: I need a value from a map later: `verifiedHead: verifiedEvent.args.head,`: that is used to generate a proof

src/finalizer/utils/helios.ts
grasphoper: Has an easy `.has` method: `const isExecuted = relayedNonces.has(nonce.toString()); // Use nonce string as key`

src/utils/TransactionUtils.ts
grasphoper: What would you suggest instead? I like a set here for it's `.has` check

src/finalizer/utils/helios.ts
grasphoper: Hm, I actually started to try to move these and I'd need to move e.g. types like `StoredCallDataEventArgs` and `ApiProofRequest` , which are not strictly `utils`. You think worth moving those over too?

grasphoper:
@nicholaspai I changed all the address / abi lookup errors for throws [4340345](https://github.com/across-protocol/relayer/pull/2203/commits/43403453269f31b32a4e8e04148643572ecdbac0). Do you think I should introduce throws in other places too?

src/finalizer/utils/helios.ts
grasphoper: Actually, yeah it can throw. See https://github.com/across-protocol/relayer/pull/2203#discussion_r2071089377

src/finalizer/utils/helios.ts
nicholaspai: I think in this specific case you should not catch the thrown errors and just let the whole program crash. It will make your code a lot simpler by not having to handle the `null` case

src/finalizer/utils/helios.ts
nicholaspai: you should just let `paginatedEventQuery` throw like we do in the rest of the codebase. We have specific channels to monitor these alerts and the providers have fallback/quorum logic that should prevent most normal RPC failures.

In general, you should not assume that RPC queries fail and if they do-- your code is too high level to handle it. We should let the provider class do its job

src/finalizer/utils/helios.ts
nicholaspai: That's fine, as we have a big catch statement in the `index.ts` file [here](https://github.com/across-protocol/relayer/blob/6a4ed8e4aabeb82d7035cf17333e6c521e0511ff/src/finalizer/index.ts#L267) to prevent a single finalizer adapter from breaking the other unrelated adapters

src/finalizer/utils/helios.ts
nicholaspai: why not just use an `array` which has an `includes()` function?

src/finalizer/utils/helios.ts
nicholaspai: I actually think both of those belong in a `utils` file, as I can see other clients potentially using the API or the HubPoolStore that would want to access those files.

They are not strictly "finalizer" objects

src/utils/TransactionUtils.ts
nicholaspai: just an array. It has a `.includes()` native function

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+    logger.error({
+      at: `Finalizer#heliosL1toL2Finalizer:processUnfinalizedHeliosMessages:${l2ChainId}`,
+      message: "HELIOS_PROOF_API_URL environment variable not set. Cannot process Helios messages.",
+    });
+    return [];
+  }
+
+  const hubPoolStoreInfo = CONTRACT_ADDRESSES[l1ChainId]?.hubPoolStore;
+  if (!hubPoolStoreInfo?.address) {
+    logger.error({
```
nicholaspai, line 485: we should throw here instead of returning []

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+    currentHeader = await sp1HeliosContract.headers(headBn);
+    if (!currentHeader || currentHeader === ethers.constants.HashZero) {
+      throw new Error(`Invalid header found for head ${currentHead}`);
+    }
+    logger.debug({
+      at: `Finalizer#heliosL1toL2Finalizer:processUnfinalizedHeliosMessages:${l2ChainId}`,
+      message: `Using SP1Helios head ${currentHead} and header ${currentHeader} for proof requests.`,
+      sp1HeliosAddress: sp1HeliosContract.address,
+    });
+  } catch (error) {
```
nicholaspai, line 510: we shouldn't have a try-catch here as the rpc eth_calls should succeed. If they don't, then we want this entire function to crash

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+
+      const proofId = calculateProofId(apiRequest);
+      const getProofUrl = `${apiBaseUrl}/api/proofs/${proofId}`;
+
+      logger.debug({ ...logContext, message: "Attempting to get proof", proofId, getProofUrl, storageSlot });
+
+      let proofState: ProofStateResponse | null = null;
+      let getError: any = null;
+
+      try {
```
nicholaspai, line 551: double nested try-catch is nasty let's remove it

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+        const response = await axios.get<ProofStateResponse>(getProofUrl);
```
nicholaspai, line 552: See this try-catch is actually useful for us. Wrapping an axios call in  a try-catch is informative for us.

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+  for (const pendingMessage of unfinalizedMessages) {
+    const l1Event = pendingMessage.l1Event; // Extract the L1 event
+    const logContext = {
+      at: `Finalizer#heliosL1toL2Finalizer:processUnfinalizedHeliosMessages:${l2ChainId}`,
+      l1TxHash: l1Event.transactionHash,
+      nonce: l1Event.args.nonce.toString(),
+      target: l1Event.args.target,
+    };
+
+    try {
```
nicholaspai, line 532: we should remove this outer try-catch

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+            proofId,
+            postUrl: `${apiBaseUrl}/api/proofs`,
+            postError: postError.message,
+            postResponseData: postError.response?.data,
+          });
+          // Exit flow for this message
+        }
+      } else if (getError) {
+        // Other error during GET
+        logger.warn({
```
nicholaspai, line 582: we should throw here, no? Seems like something is wrong and we should be aware of it and let the catch in finalizer/index.ts handle it

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+        } else if (proofState.status === "errored") {
+          // 1c. SUCCESS ("errored") -> Log high severity, request again, exit flow
+          logger.error({
+            // Use error level log
+            ...logContext,
+            message: "Proof generation errored on ZK API side. Requesting again.",
+            proofId,
+            errorMessage: proofState.error_message,
+          });
+          try {
```
nicholaspai, line 606: Why re-try here? Should we just log an error and then wait for the next finalizer run to re-request?

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+  // --- Process messages needing only execution ---
+  for (const message of needsExecutionOnlyMessages) {
+    const { l1Event, verifiedHead } = message;
+    const nonce = l1Event.args.nonce;
+    const l1Target = l1Event.args.target; // Get target from L1 event
+    const l1Data = l1Event.args.data; // Get data from L1 event
+
+    if (!verifiedHead) {
+      // @dev This shouldn't happen. If it does, there's a bug that needs fixing
+      logger.error({
```
nicholaspai, line 692: throw here if you don't expect this to happen, don't just log an error

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,887 @@
...
+        message: `Finalize Helios msg (HubPoolStore nonce ${nonce.toString()}) - Step 2 ONLY: Execute on SpokePool`,
+      };
+      transactions.push(executeTx);
+      crossChainMessages.push({
+        type: "misc",
+        miscReason: "ZK bridge finalization (Execute Message Only)",
+        originationChainId: l1ChainId,
+        destinationChainId: l2ChainId,
+      });
+    } catch (error) {
```
nicholaspai, line 731: i don't see why anything in this try-block should fail

src/finalizer/utils/helios.ts
grasphoper: Here, ZK API tried to generate a proof but failed. We re-request it here and log an error. In practice, we should never have proofs for which generation errors. The next finalizer run will not re-request it unless we explicitly do so here

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+  }
+  const hubPoolStoreAddress = hubPoolStoreInfo.address;
+  const sp1HeliosContract = getSp1HeliosContract(l2ChainId, l2Provider);
+
+  const headBn: ethers.BigNumber = await sp1HeliosContract.head();
+  // todo: well, currently we're taking currentHead to use as prevHead in our ZK proof. There's a particular scenario where we could speed up proofs
+  // todo: (by not making them to wait for finality longer than needed) if our blockNumber that we need a proved slot for is older than this head.
+  const currentHead = headBn.toNumber();
+  const currentHeader = await sp1HeliosContract.headers(headBn);
+  if (!currentHeader || currentHeader === ethers.constants.HashZero) {
```
nicholaspai, line 347: i'm not sure calling a contract can ever return `undefined`-- it should either error or return you a type that you expect--in this case a BN

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+
+  // --- Process messages needing only execution ---
+  for (const message of needsExecutionOnlyMessages) {
+    const { l1Event, verifiedHead } = message;
+    const nonce = l1Event.args.nonce;
+    const l1Target = l1Event.args.target; // Get target from L1 event
+    const l1Data = l1Event.args.data; // Get data from L1 event
+
+    if (!verifiedHead) {
+      // @dev This shouldn't happen. If it does, there's a bug that needs fixing.
```
nicholaspai, line 479: is `verified` head an optional property in the type of `message`? If its expected to be here then you can let TS do its job and not have to have this `if` statement

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+          nonce: nonce.toString(),
+          storageSlot: expectedStorageSlot,
+          verifiedOnL2TxHash: verifiedEvent.transactionHash,
+          verifiedHead: verifiedEvent.args.head.toString(),
+        });
+      } else {
+        // Not verified and not executed -> Needs Proof and Execution
+        pendingMessages.push({
+          l1Event: l1Event,
+          status: "NeedsProofAndExecution",
```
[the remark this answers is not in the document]
grasphoper: No typos is enforced by 

```
type CrossChainMessageStatus = "NeedsProofAndExecution" | "NeedsExecutionOnly";

interface PendingCrosschainMessage {
  l1Event: StoredCallDataEvent; // The original HubPoolStore event triggering the flow
  status: CrossChainMessageStatus;
  verifiedHead?: ethers.BigNumber; // Head from the StorageSlotVerified event, only present if status is NeedsExecutionOnly
}
```

`CrossChainMessageStatus` is a sort of enum?

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+  });
+
+  return { callData: transactions, crossChainMessages: crossChainMessages };
+}
+
+/**
+ * Retrieves an ethers.Contract instance for the SP1Helios contract on the specified chain.
+ * @throws {Error} If the SP1Helios contract address or ABI is not found for the given chainId in CONTRACT_ADDRESSES.
+ */
+function getSp1HeliosContract(chainId: number, signerOrProvider: Signer | Provider): ethers.Contract {
```
nicholaspai, line 588: nit: should we move `getSp1HeliosContract` and  `getHubPoolStoreContract` into `UniversalUtils`?

src/finalizer/utils/helios.ts
grasphoper: Not sure how to check this. IMO this is a nice check that prevents us from sending a request with `dst_chain_contract_from_header: undefined` down below

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+          logger.debug({ ...logContext, message: "Proof successfully retrieved.", proofId });
+          successfulProofs.push({
+            proofData: proofState.update_calldata,
+            sourceNonce: l1Event.args.nonce, // Use nonce from L1 event
+            target: l1Event.args.target, // Use target from L1 event
+            sourceMessageData: l1Event.args.data, // Use data from L1 event
+          });
+        } else {
+          // todo? Might want to log an error here
+          logger.warn({
```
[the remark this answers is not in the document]
grasphoper: It can happen if API returns an incorrectly formatted response. The thinking is, "it must exist" yeah. But networking stuff is weird

src/finalizer/utils/helios.ts
grasphoper: But will it crash on doing e.g. this? `const executeArgs = [nonce, encodedMessage, verifiedHead]; // Use encodedMessage` I feel like it won't and we'll just try to execute with undefined. The only thing that would save us is this log below:

```
    // @dev Warn about messages that require only half of finalization. Means that either a tx from prev. run got stuck or failed or something else weird happened
    logger.warn({
      at: `Finalizer#heliosL1toL2Finalizer:generateTxnItem:${l2ChainId}`,
      message: "Generating SpokePool.executeMessage ONLY for partially finalized message.",
      nonce: nonce.toString(),
      l1TxHash: l1Event.txnRef,
      verifiedHead: verifiedHead.toString(),
    });
```
    
Where `verifiedHead.toString()` will throw. But if we decide to remove that log, the logic will break

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,607 @@
...
+
+    if (!verifiedHead) {
+      // @dev This shouldn't happen. If it does, there's a bug that needs fixing.
+      throw new Error(`Logic error: Message ${nonce.toString()} needs execution only but verifiedHead is missing.`);
+    }
+
+    // @dev Warn about messages that require only half of finalization. Means that either a tx from prev. run got stuck or failed or something else weird happened
+    logger.warn({
+      at: `Finalizer#heliosL1toL2Finalizer:generateTxnItem:${l2ChainId}`,
+      message: "Generating SpokePool.executeMessage ONLY for partially finalized message.",
```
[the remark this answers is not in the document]
grasphoper: This one shouldn't really happen. It only will if we failed to submit and execute `.update` followed by `.execute` correclty in the prev. run and only `.update` went through. I feel like this warrants a warn

src/finalizer/utils/helios.ts
grasphoper: Well. It's nuanced. These fns are opinionated in the sense that they throw if the contract is not there. This throwing opinion is really of `helios.ts`, not of utils package. But I agree that we should probably always throw on missing addr / abi so we can move?

src/finalizer/utils/helios.ts
nicholaspai: It would crash because `execute()` would revert with an undefiend `verifiedHead` and then we'd have this log showing `verifiedHead` is missing.

However in general I think we should be trusting that the `needsExecutionOnlyMessages: PendingCrosschainMessage[],` type is what it says it is: `verifiedHead` is an expected prop so it should be here, right?

grasphoper: > However in general I think we should be trusting that the needsExecutionOnlyMessages: PendingCrosschainMessage[], type is what it says it is: verifiedHead is an expected prop so it should be here, right?

We're trusting that this will never happen, that's why we're adding this explicit check: expecting it to never throw here. It's putting our implicit understanding about the logic of the whole system as an explicit code constraint.

grasphoper: > It would crash because execute() would revert with an undefiend verifiedHead and then we'd have this log showing verifiedHead is missing.

It will crash yeah, but then it's up to us to figure out later why this happened in the first place. This check saves us some mental effort it that case

src/finalizer/utils/helios.ts
grasphoper: Also, I don't feel like these fns are super-useful to reuse. I don't mind either way

grasphoper:
@nicholaspai I addressed most of your comments -- resolved the addressed ones. What else should I 100% address before we can merge?

src/finalizer/utils/helios.ts
nicholaspai: ok i think its unneccessary but your call

src/finalizer/utils/helios.ts
nicholaspai: i don't think the opinionated throwing is a bad thing to have in a utility function named as such. We should move it to a utility folder.

src/finalizer/utils/helios.ts
```diff
@@ -0,0 +1,582 @@
...
+    const encodedMessage = ethers.utils.defaultAbiCoder.encode(["address", "bytes"], [l1Target, l1Data]);
+    // ------------------------------------
+
+    const executeArgs = [proof.sourceNonce, encodedMessage, decodedOutputs.newHead]; // Use encodedMessage
+    const executeTx: AugmentedTransaction = {
+      contract: universalSpokePoolContract,
+      chainId: l2ChainId,
+      method: "executeMessage",
+      args: executeArgs,
+      unpermissioned: true,
```
nicholaspai, line 557: This should be false right? Only the EOA with the updater role can call this right?

PR #2209, opened 2025-05-01
improve: use token symbols map over hub pool client for rebalancing

bmzig:
We will need to rebalance tokens which are not enabled for liquidity provision, so the inventory client and adapter manager should not use the hub pool client to resolve L1 token information. They should instead use `TOKEN_SYMBOLS_MAP.` This is safe since all calls use an L1 token as input, and `getTokenInfo` has a specific block of code [here](https://github.com/across-protocol/sdk/blob/4b856bdd2cf4020ea1dde286048afe83d2d1648a/src/utils/TokenUtils.ts#L114) which ensures that if there are many L1 definitions for a token, we will use the remapped symbol when resolving token information.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -28,6 +28,7 @@ import {
   getL1TokenInfo,
   depositForcesOriginChainRepayment,
   getRemoteTokenForL1Token,
+  getTokenInfo,
 } from "../utils";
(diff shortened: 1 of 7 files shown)


nicholaspai (dismissed):
I think we need to replace this in other parts of the code as well

nicholaspai (approved):

PR #2218, opened 2025-05-03
feat(finalizer): Remove individual txn client simulation

nicholaspai:
We should rely on the multicaller client's transaction simulation logic rather than directly call the transaction client to simulate finalization txns invidually. This code is leftover from legacy code before we used to submit finalization txns to the multicaller client.

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -16,7 +16,6 @@ import {
 import { DataworkerConfig } from "../dataworker/DataworkerConfig";
 import { SpokePoolClientsByChain } from "../interfaces";
 import {
-  bnZero,
   Signer,
   blockExplorerLink,
   config,
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -304,8 +304,6 @@ export async function finalize(
       .map(([k]) => k)}`
   );
 
-  const gasEstimation = bnZero;
-
   // @dev use multicaller client to execute batched txn to take advantage of its native txn simulation
   // safety features. This only works because we assume all finalizer transactions are
   // unpermissioned (i.e. msg.sender can be anyone). If this is not true for any chain then we'd need to use
@@ -324,7 +322,6 @@ export async function finalize(
         chainId: Number(chainId),
         method: "aggregate",
         args: [finalizerTxns],
-        gasLimit: gasEstimation,
         gasLimitMultiplier: 2,
         unpermissioned: true,
         message: `Batch finalized ${finalizerTxns.length} txns`,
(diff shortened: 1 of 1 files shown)


src/finalizer/index.ts
```diff
@@ -305,128 +304,83 @@ export async function finalize(
...
 
-  const txnClient = new TransactionClient(logger);
-
-  let gasEstimation = bnZero;
-  const batchGasLimit = BigNumber.from(10_000_000);
-  // @dev To avoid running into block gas limit in case the # of finalizations gets too high, keep a running
-  // counter of the approximate gas estimation and cut off the list of finalizations if it gets too high.
-
-  // Ensure each transaction would succeed in isolation.
-  const finalizations = await sdkUtils.filterAsync(finalizationsToBatch, async ({ txn: _txn, crossChainMessage }) => {
```
nicholaspai, line 316: The diff looks large but all I'm doing is deleting this filter

src/finalizer/index.ts
```diff
@@ -305,128 +304,83 @@ export async function finalize(
...
+      ({ crossChainMessage }) => crossChainMessage.destinationChainId
+    );
+    for (const [chainId, finalizations] of Object.entries(finalizationsByChain)) {
+      const finalizerTxns = finalizations.map(({ txn }) => txn);
+      const txnToSubmit: AugmentedTransaction = {
+        contract: multicall2Lookup[Number(chainId)],
+        chainId: Number(chainId),
+        method: "aggregate",
+        args: [finalizerTxns],
+        gasLimit: gasEstimation,
```
bmzig, line 327: Did we simulate finalizations twice then? Once in the filter you deleted and once when calling `executeTxnQueue`? 

Also, `gasEstimation` is zero here, so do these finalizations work or do they just immediately run out of gas?

bmzig (dismissed):
One (clarifying) comment.

src/finalizer/index.ts
nicholaspai: > Did we simulate finalizations twice then? Once in the filter you deleted and once when calling executeTxnQueue?

Yeah i think that's what we're doing now.

> Also, gasEstimation is zero here, so do these finalizations work or do they just immediately run out of gas?

Ah this is a good point. We were before adding the gas estimate for each individual sim to this counter to get an aggregate gas estimation value.

I guess actually one feature of this simulation step is that we were filtering out finalizations that fail and no letting them block the whole thing. Do you think we can still remove this code and maintain this logic by using `tryAggregate` instead of `aggregate`?

nicholaspai:
I'm actually going to close this because I think the simulation step is useful for removing finalization txns that fail to prevent the full batch from being blocked.

src/finalizer/index.ts
nicholaspai: I removed the gasEstimation line https://github.com/across-protocol/relayer/pull/2218/commits/915e7d5c9931d892fee38e46048a2a1cc7150e51

bmzig (approved):

nicholaspai:
Replaced by https://github.com/across-protocol/relayer/pull/2229 due to merge conflicts being too complex and it was just easier to reopen

PR #2219, opened 2025-05-04
feat(MulticallerClient): Allow caller to use PermissionedMulticaller

nicholaspai:
Can be used to batch txns to contracts that whitelist the msg.sender

Depends on https://github.com/across-protocol/sdk/pull/1017

### DIFF
--- a/src/clients/MultiCallerClient.ts
+++ b/src/clients/MultiCallerClient.ts
@@ -16,6 +16,7 @@ import {
   getProvider,
   Multicall2Call,
   assert,
+  getPermissionedMultisender,
 } from "../utils";
 import { AugmentedTransaction, TransactionClient } from "./TransactionClient";
 import lodash from "lodash";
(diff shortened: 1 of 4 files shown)


src/clients/MultiCallerClient.ts
```diff
@@ -272,12 +284,17 @@ export class MultiCallerClient {
     const callData: Multicall2Call[] = [];
     let gasLimit: BigNumber | undefined = bnZero;
     transactions.forEach((txn, idx) => {
-      if (!txn.unpermissioned || txn.chainId !== chainId) {
+      if (!txn.unpermissioned || !txn.sendThroughPermissionedMulticall || txn.chainId !== chainId) {
```
grasphoper, line 287: Is this correct?

This check seems to be saying: "if any of the txns in `transactions` required permissions => `throw`". But this is not what we're going for with the PermissionedMulticall?

grasphoper: Before we fix it here, let's first discuss our strat for using this in finalizer in Slack

nicholaspai: right,  should be:
```suggestion
      if ((!txn.unpermissioned || !txn.sendThroughPermissionedMulticall) || txn.chainId !== chainId) {
```

PR #2221, opened 2025-05-05
improve(Relayer): Make inventoryClient.canTakeDestinationChainRepayment more robust

nicholaspai:
Currently our check for whether a destination chain has a valid pool rebalance route for a deposit is flawed because it doesn't check if the output token and input token are mapped to the same L1 token.

This isn't a problem in production because the relayer filters out any deposits that involve cross chain swaps before it gets to choosing a repayment chain, but this PR will add more future protection in case this filter gets removed.


src/clients/InventoryClient.ts
```diff
@@ -460,9 +474,7 @@ export class InventoryClient {
 
     const forceOriginRepayment = depositForcesOriginChainRepayment(deposit, this.hubPoolClient);
     if (!this.isInventoryManagementEnabled()) {
-      return [
-        forceOriginRepayment || !this.canTakeDestinationChainRepayment(deposit) ? originChainId : destinationChainId,
```
nicholaspai, line 464: `forceOriginRepayment` and `canTakeDestinationChainRepayment` should be mutually exclusive now

bmzig (approved):

PR #2222, opened 2025-05-05
fix(ProfitClient): Query additional token prices

pxrl:
Update the ProfitClient and TokenClient to support additional tokens,
beyond what is known by the HubPoolClient.


bmzig (dismissed):

nicholaspai (approved):

bmzig (approved):

src/clients/ProfitClient.ts
```diff
@@ -576,6 +579,15 @@ export class ProfitClient {
       tokens[symbol] ??= nativeTokenAddress;
     });
 
+    // Manually add additional token prices as required. This can be
+    // needed for tokens without a pool rebalance route. @todo: fix.
+    if (this.enabledChainIds.includes(CHAIN_IDs.BSC)) {
```
bmzig, line 584: You can get rid of this now, right?

pxrl: Nope, we still need it (or some variation of).

There's also at least one other issue - we rely on a pool rebalance route somewhere else, so I'll flush that one out as part of this PR.

pxrl:
This requires inventory config to be enabled in the relayer. I'll have a tweak to work around that shortly.

nicholaspai (approved):

PR #2229, opened 2025-05-06
feat(finalizer): Remove individual txn client simulation

nicholaspai:
We should rely on the multicaller client's transaction simulation logic rather than directly call the transaction client to simulate finalization txns invidually. This code is leftover from legacy code before we used to submit finalization txns to the multicaller client.

Replaces https://github.com/across-protocol/relayer/pull/2218

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -3,7 +3,7 @@ import assert from "assert";
 import { Contract, ethers } from "ethers";
 import { getAddress } from "ethers/lib/utils";
 import { groupBy, uniq } from "lodash";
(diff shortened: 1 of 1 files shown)


bmzig (approved):

grasphoper (approved):

nicholaspai:
> I think part of the reasoning we were doing this if was so we could determine if the individual txns were failing rather than the full error of the multicall. However, this does pose the risk that individual transactions succeed but the entire multicall could fail (i.e. overlapping demand for the same resources)

Yeah but the multicaller client does this now for us

PR #2232, opened 2025-05-06
feat(relayer): Support Circle Bridged USDC on Lens

pxrl:


src/adapter/bridges/ZKStackUSDCBridge.ts
```diff
@@ -0,0 +1,110 @@
...
+    return { contract, method: "requestL2TransactionTwoBridges", args, value };
+  }
+
+  async queryL1BridgeInitiationEvents(
+    l1Token: EvmAddress,
+    _from: EvmAddress,
+    to: EvmAddress,
+    eventConfig: EventSearchConfig
+  ): Promise<BridgeEvents> {
+    // Logic changes based on whether we are sending tokens to the spoke pool or to an EOA.
```
nicholaspai, line 79: How is it different when sending token sto spoke vs EOA?

pxrl:
Rebalancing tested OK in the relayer:
- Deposit https://etherscan.io/tx/0x3bd468ccd7a5f90588e7fd165318a2017680afb52c5ccade03611c7713d4f98b
- Receipt https://explorer.lens.xyz/tx/0x5d3f3a1d6d50cda10b7a20293924aaf33a2163ce2b4aec0ad62a0546b380d4c7

src/adapter/bridges/ZKStackUSDCBridge.ts
pxrl: Good catch - that's a hangover from the ERC20 bridge, which was used as the scaffolding for this work. The comment isn't valid in this case because there's no distinction - we only check for the recipient address.

```suggestion
```

nicholaspai (approved):
Seems like it should work

bmzig (approved):

PR #2238, opened 2025-05-08
refactor: unify Binance finalizers

bmzig:
Generally the Binance l1 -> l2 finalizer and l2 -> l1 finalizers run in parallel. This leads to two issues:
- It's easy for the bot to be rate limited by the API since we request data from `4 + 2n` endpoints (where `n` is the number of `SUPPORTED_TOKENS`). 
- There are race conditions between the two finalizers for withdrawals. This normally isn't an issue, but since it's possible for lookbacks vary, we may want to withdraw dust on L2 and withdraw a sizable amount (say `x`) on L1. What could happen is the L2 -> L1 finalizer withdraws all `x` of the token, and the L1 -> L2 finalizer attempts to withdraw dust (which it now cannot do since everything was withdrawn to L1 already). Ultimately, this comes down to the timing of calling `accountCoins`. Unifying the finalizers fixes this since now for each token, we start by finalizing on L2 and then move to finalizing on L1.


src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+      // For the l1 to l2 finalizer, we need to re-map WBNB -> BNB and re-map WETH -> ETH.
+      const symbol = _symbol[0] === "W" ? _symbol.slice(1) : _symbol;
+
+      const coin = accountCoins.find((coin) => coin.symbol === symbol);
+      const l1Token = TOKEN_SYMBOLS_MAP[symbol].addresses[hubChainId];
+      const { decimals: l1Decimals } = getTokenInfo(l1Token, hubChainId);
+      const withdrawals = await getBinanceWithdrawals(binanceApi, symbol, fromTimestamp);
+
+      // Start by finalizing L1 -> L2, then go to L2 -> L1.
+      await mapAsync(["ETH", "BSC"], async (depositNetwork) => {
```
bmzig, line 103: This is a copy of the l1ToL1Finalizer/l2ToL1Finalizer up to here, where we iterate though the deposit networks instead of hardcoding the deposit/withdrawal network. Essentially, the only things that's changed here is making the network selection variable (and generalizing some of the utility functions, which is commented below).

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+  l2SpokePoolClient: SpokePoolClient,
+  l1SpokePoolClient: SpokePoolClient,
+  senderAddresses: string[]
+): Promise<FinalizerPromise> {
+  const chainId = l2SpokePoolClient.chainId;
+  const hubChainId = l1SpokePoolClient.chainId;
+  const l1EventSearchConfig = l1SpokePoolClient.eventSearchConfig;
+
+  const binanceApi = await getBinanceApiClient(process.env["BINANCE_API_BASE"]);
+  const fromTimestamp = (await getTimestampForBlock(hubSigner.provider, l1EventSearchConfig.fromBlock)) * 1_000;
```
nicholaspai, line 65: should we promise all these?

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+    });
+  });
+  return {
+    callData: [],
+    crossChainMessages: [],
+  };
+}
+
+// Gets all binance deposits for the Binance account starting from `startTime`-present.
+async function getBinanceDeposits(
```
bmzig, line 182: These should be near copies of the functions that were formerly in `common.ts`. The difference is we no longer accept a `network` as an argument and instead return _all_ the deposits/withdrawals for the network. This lets us query the API a lot less since we can reuse the deposit/withdrawal data in the actual l1 -> l2 and l2 -> l1 logic.

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  const [binanceDeposits, accountCoins] = await Promise.all([
+    getBinanceDeposits(binanceApi, hubSigner.provider, fromTimestamp),
+    await getAccountCoins(binanceApi),
+  ]);
+
+  logger.debug({
+    at: "BinanceFinalizer",
+    message: `Found ${binanceDeposits.length} historical deposits with status === 1.`,
+    fromTimestamp: fromTimestamp / 1_000,
```
nicholaspai, line 75: math.round here could be useful

also what does status === 1 mean? I think its better to have more human-translatatble status logs here like "X deposits can be finalized" or whatever, so someone else reading the logs who didn't write this code could understand a bit more.

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+  });
+
+  await mapAsync(senderAddresses, async (address) => {
+    // Filter our list of deposits by the withdrawal address. We will only finalize deposits when the depositor EOA is in `senderAddresses`.
+    const depositsInScope = binanceDeposits.filter((deposit) =>
+      compareAddressesSimple(deposit.externalAddress, address)
```
nicholaspai, line 81: how are you guaranteeting that the deposit.externalAddress is an EOA? Couldn't it have been sent from an AtomicDepositor, for example?

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+    );
+    if (depositsInScope.length === 0) {
+      logger.debug({
+        at: "BinanceFinalizer",
+        message: `No finalizable deposits found for ${address}`,
+        fromTimestamp: fromTimestamp / 1_000,
```
nicholaspai, line 87: same comment above on timestamps, might be best to use a utility

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+      });
+      return;
+    }
+
+    // The inner loop finalizes all deposits for all supported tokens for the address.
+    await mapAsync(SUPPORTED_TOKENS[chainId], async (_symbol) => {
+      // For the l1 to l2 finalizer, we need to re-map WBNB -> BNB and re-map WETH -> ETH.
+      const symbol = _symbol[0] === "W" ? _symbol.slice(1) : _symbol;
```
nicholaspai, line 95: Might be worth using `getWrappedNativeTokenAddress` and updating its logic to support BNB/WBNB

src/finalizer/utils/binance.ts
```diff
@@ -0,0 +1,241 @@
...
+  };
+}
+
+// Gets all binance deposits for the Binance account starting from `startTime`-present.
+async function getBinanceDeposits(
+  binanceApi: Binance,
+  provider: ethers.providers.Provider,
+  startTime: number
+): Promise<BinanceInteraction[]> {
+  const _depositHistory = await binanceApi.depositHistory({ startTime });
```
nicholaspai, line 187: nit: its your call here but I like printing out all deposits with pending as well as fully processed statuses here and printing them in a debug log. See [here](https://github.com/across-protocol/relayer/blob/8f2637303a679f45d003dce780e5b3eef7ca74fd/src/finalizer/utils/arbStack.ts#L293) for example

nicholaspai (commented):
a few comments

src/finalizer/utils/binance.ts
bmzig: Yeah that's probably a good thing to do. I'll return the deposit with the status included, log about it, and then filter. https://github.com/across-protocol/relayer/pull/2238/commits/da3f7a2c0dbc31584a022b499c572ddf8c2eb968

src/finalizer/utils/binance.ts
bmzig: I ended up removing one of the `fromTimestamps` since both logs emit the same value. I also just log off the ms value since that's probably want we would realistically want since that is what the API expects (in case we ever need to reproduce a return value from the API). https://github.com/across-protocol/relayer/pull/2238/commits/da3f7a2c0dbc31584a022b499c572ddf8c2eb968

src/finalizer/utils/binance.ts
bmzig: I added a comment about it here https://github.com/across-protocol/relayer/pull/2238/commits/da3f7a2c0dbc31584a022b499c572ddf8c2eb968. Basically, this will always be an EOA since we fetch the transaction receipt of the deposit transaction, meaning that we will get the `tx.origin` of the deposit.

src/finalizer/utils/binance.ts
bmzig: I made an enum to make this more readable. I also removed the division so that we will log the exact same values we sent to Binance. I think it's more useful that way. https://github.com/across-protocol/relayer/pull/2238/commits/da3f7a2c0dbc31584a022b499c572ddf8c2eb968

src/finalizer/utils/binance.ts
bmzig: We should https://github.com/across-protocol/relayer/pull/2238/commits/da3f7a2c0dbc31584a022b499c572ddf8c2eb968

src/finalizer/utils/binance.ts
bmzig: I opted to hardcode the logic since using any of our utility functions would be slower, since we would first need to get the native token symbol for both BSC and mainnet, compare both of them to the symbol, and then if one of them matched, change the symbol to the respective network's native token symbol. https://github.com/across-protocol/relayer/pull/2238/commits/82ed4404e677d3058331b8c30181e231ad867a36

nicholaspai (approved):

PR #2239, opened 2025-05-08
fix: filter CCTP initiation events by destination domain

bmzig:
Without this, we think we have a much larger amount of outstanding transfers for a given chain than we do in reality, since each CCTP bridge thinks that _any_ rebalance is a rebalance to their l2 network.


nicholaspai (approved):
great catch

PR #2241, opened 2025-05-08
improve: add periodic `KeepAlive` messages for `SP1Helios`

grasphoper:
If `SP1Helios.head` is left unupdated for > 1 WEEK, the contract gets bricked. We let 24hrs go by with no updates, but after then we call `.update()` even if there are no new messages to finalize (by requesting a new ZK proof with empty storage slots to prove).

This turned out to be a large refactor. IMO, the current flow is nice though: 
- we identify required `HeliosAction`s
- try to enrich them with ZK proof data
- for all the actions that are "ready", generate on-chain txs

Closes ACX-4075


src/finalizer/utils/helios.ts
```diff
@@ -60,77 +75,76 @@ export async function heliosL1toL2Finalizer(
...
+  sp1HeliosHead: number;
+  sp1HeliosHeader: string;
+}
+
+async function getSp1HeliosHeadData(l2SpokePoolClient: SpokePoolClient): Promise<Sp1HeliosHeadData> {
+  const sp1HeliosL2 = getSp1HeliosContract(l2SpokePoolClient.chainId, l2SpokePoolClient.spokePool.provider);
+  const sp1HeliosHeadBn: ethers.BigNumber = await sp1HeliosL2.head();
+  const sp1HeliosHead = sp1HeliosHeadBn.toNumber();
+  const sp1HeliosHeader = await sp1HeliosL2.headers(sp1HeliosHeadBn);
+  if (!sp1HeliosHeader || sp1HeliosHeader === ethers.constants.HashZero) {
```
nicholaspai, line 127: I don't think `sp1HeliosHeader` will ever be undefined unless there's an RPC error in which case an error will be thrown by the Provider.

src/finalizer/utils/helios.ts
```diff
@@ -140,75 +154,250 @@ async function identifyPendingHeliosMessages(
...
+      const verifiedEvent = verifiedSlotsMap.get(expectedStorageSlot);
+      actions.push({
+        type: "ExecuteOnly",
+        l1Event: l1Event,
+        verifiedSlotHead: verifiedEvent.head,
+      });
+      // Warn about a half-finished finalization
+      logger.warn({
+        at: `Finalizer#identifyRequiredActions:${l2ChainId}`,
+        message:
```
nicholaspai, line 186: Should this be `warn` and not `debug`? It seems like something that could happen quite regularly and it would add noise to the higher channels (slack, PD)

nicholaspai (commented):
Good feature, left some style comments mostly

src/finalizer/utils/helios.ts
grasphoper: I added as a `warn` to have visibility into this (I thought slack only). Esp. cause we don't simulate the executeTx. So we can be in a loop of trying to submit a tx and it reverting and we wouldn't know about it I think?

src/finalizer/utils/helios.ts
grasphoper: Yeah I guess I can remove this

nicholaspai (approved):
LGTM

src/finalizer/utils/helios.ts
```diff
@@ -122,10 +122,8 @@ async function getSp1HeliosHeadData(l2SpokePoolClient: SpokePoolClient): Promise
   const sp1HeliosHeadBn: ethers.BigNumber = await sp1HeliosL2.head();
   const sp1HeliosHead = sp1HeliosHeadBn.toNumber();
   const sp1HeliosHeader = await sp1HeliosL2.headers(sp1HeliosHeadBn);
-  if (!sp1HeliosHeader || sp1HeliosHeader === ethers.constants.HashZero) {
-    throw new Error(
-      `Invalid or zero header found for SP1Helios head ${sp1HeliosHead}. Cannot proceed with proof generation.`
-    );
+  if (sp1HeliosHeader === ethers.constants.HashZero) {
```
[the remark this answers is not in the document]
grasphoper: Yeah. Wrote a quick test script, it's a "string", the checks works as intended!

PR #2242, opened 2025-05-09
chore: bump sdk

bmzig:
This is to sync the relayer with the epic branch in the sdk, which should be merged into master sometime soon.


bmzig:
The package just needs to change but there are no differences between the beta package and 4.2.0

pxrl (approved):

PR #2245, opened 2025-05-12
fix: Revert "improve: add periodic `KeepAlive` messages for `SP1Helios` (#…

grasphoper:
…2241)"

This reverts commit 75eff00444556f50b21098a2a0008eb4f8e675f0.

PR #2246, opened 2025-05-13
improve: add periodic KeepAlive messages for SP1Helios

grasphoper:
If SP1Helios.head is left unupdated for > 1 WEEK, the contract gets bricked. We let 24hrs go by with no updates, but after then we call .update() even if there are no new messages to finalize (by requesting a new ZK proof with empty storage slots to prove).

This turned out to be a large refactor. IMO, the current flow is nice though:

we identify required HeliosActions
try to enrich them with ZK proof data
for all the actions that are "ready", generate on-chain txs
Closes ACX-4075


grasphoper:
@nicholaspai This is a reintroduction of #2241 after I reverted it in #2245 . I added a fix for the broken import here: [dda8b76](https://github.com/across-protocol/relayer/pull/2246/commits/dda8b7664179ff0245ae536d264b9419bdc8e94f)

nicholaspai (approved):

PR #2247, opened 2025-05-13
fix(finalizer): Correctly dereference Blast requestId

pxrl:


amateima (approved):

src/finalizer/utils/opStack.ts
```diff
@@ -792,17 +792,18 @@ async function multicallOptimismFinalizations(
...
     Promise.all(
-      claimableWithdrawalRequests.map((requestId) =>
+      claimableWithdrawalRequests.map(({ requestId }) =>
         usdYieldManager.queryFilter(usdYieldManager.filters.WithdrawalClaimed(requestId), fromBlock)
       )
     ),
   ]);
-  const withdrawalRequestIsClaimed = withdrawalClaims.map((_id, i) => withdrawalClaims[i].length > 0);
+
+  const withdrawalRequestIsClaimed = withdrawalClaims.map((claims) => claims.length > 0);
```
pxrl, line 806: Driveby change, but unless I'm missing something the `i` variable was unnecessary since it was only performing a lookup inside the array that was being iterated over.

bmzig (approved):

PR #2249, opened 2025-05-13
refactor: allow arbitrary arguments to bridge adapters

bmzig:
There are two changes to the bridge adapters here:
1. We no longer define `l2SignerOrProvider` on the interface level of the bridge adapters. We didn't really even use this anywhere, so it's good to cut off for the sake of contravariance, but we need this relaxation anyways since the Solana bridge will require a different l2 provider type entirely, so it will be incompatible with a BaseBridgeAdapter given its current state.
2. The `CUSTOM_BRIDGE`/`CANONICAL_BRIDGE` maps now let us take an arbitrary set of arguments. Our "contract" is that we will supply some extension of the BaseBridgeAdapter, but the inputs we need to get there are no longer explicit. This means we no longer require each extension of BaseBridgeAdapter to have the same constructor arguments, meaning that we can pass in an `SVMProvider` in the future to these values. This comes at the cost of not getting type checking at compilation time. I think this is OK, since these bridge adapters implement a known interface, so we really only need to be more careful [here](https://github.com/across-protocol/relayer/blob/302b2ea9ef8cff08a9357494528f025fb11c2156/src/clients/bridges/AdapterManager.ts#L73) when defining the arguments to pass into each bridge.

### DIFF
--- a/src/adapter/bridges/BinanceCEXBridge.ts
+++ b/src/adapter/bridges/BinanceCEXBridge.ts
@@ -10,7 +10,6 @@ import {
   isDefined,
   getTimestampForBlock,
   mapAsync,
-  ethers,
   getBinanceApiClient,
   floatToBN,
   CHAIN_IDs,
@@ -25,6 +24,7 @@ export class BinanceCEXBridge extends BaseBridgeAdapter {
   protected readonly binanceApiClientPromise;
   protected binanceApiClient;
   protected tokenSymbol: string;
+  protected l2Provider: Provider;
 
   constructor(
     l2chainId: number,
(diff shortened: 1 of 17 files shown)


src/adapter/bridges/BinanceCEXBridge.ts
```diff
@@ -48,6 +48,13 @@ export class BinanceCEXBridge extends BaseBridgeAdapter {
     const _tokenSymbol = getTokenInfo(l1Token.toAddress(), this.hubChainId).symbol;
     // Handle the special case for when we are bridging WBNB to BNB on L2.
     this.tokenSymbol = _tokenSymbol === "WBNB" ? "BNB" : _tokenSymbol;
+
+    // Cast the input Signer | Provider to a Provider.
+    if (l2SignerOrProvider instanceof Signer) {
+      this.l2Provider = l2SignerOrProvider.provider;
```
nicholaspai, line 54: nit probably could. make this one line with a `? : `

nicholaspai (approved):

src/adapter/bridges/BinanceCEXBridge.ts
bmzig: Very true https://github.com/across-protocol/relayer/pull/2249/commits/a6040a69f5adbd675d6af41104e4037afa0be953

nicholaspai (approved):

PR #2257, opened 2025-05-16
Revert "Update index.ts (#2229)"

pxrl:
This reverts commit 54ff57e91d167af51485bf83ed6809c23a323ddc.

The key revert here is to avoid the bundling of transactions before they're being supplied to the MultiCallerClient, because this prevents the MultiCallerClient from filtering out failing transactions and allows a single bad transaction to spoil the bundle.

The rationale for reverting this change is that there's currently an impossible-to-execute message awaiting finalization on Linea. Since the message can't be executed on Linea (due to SpokePool v4 upgrade) it can never be finalized. The only (apparent) way to be rid of it is to wait for it to drop out of the finalizer lookback. Without this, the sweeper finalizer would continue to error for the next ~15 days (due to its long lookback configuration).

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -3,7 +3,7 @@ import assert from "assert";
 import { Contract, ethers } from "ethers";
 import { getAddress } from "ethers/lib/utils";
 import { groupBy, uniq } from "lodash";
-import { AugmentedTransaction, HubPoolClient, MultiCallerClient } from "../clients";
+import { AugmentedTransaction, HubPoolClient, MultiCallerClient, TransactionClient } from "../clients";
 import {
   CONTRACT_ADDRESSES,
   Clients,
@@ -16,6 +16,8 @@ import {
 import { DataworkerConfig } from "../dataworker/DataworkerConfig";
 import { SpokePoolClientsByChain } from "../interfaces";
 import {
+  BigNumber,
+  bnZero,
   Signer,
   blockExplorerLink,
   config,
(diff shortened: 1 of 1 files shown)


bmzig:
Doesn't the multicaller client filter out failed simulations? https://github.com/across-protocol/relayer/blob/fc8bef91040b4dfb2a3fd0b859988c17e42ca600/src/clients/MultiCallerClient.ts#L228

pxrl:
> Doesn't the multicaller client filter out failed simulations?
> 
> https://github.com/across-protocol/relayer/blob/fc8bef91040b4dfb2a3fd0b859988c17e42ca600/src/clients/MultiCallerClient.ts#L228

Not when they're bundled like this:

```
  396         if (multicallTxns.length > 0) {                                                                                                                                                                                                                  
  397           const txnToSubmit: AugmentedTransaction = {                                                                                                                                                                                                    
  398             contract: multicall2Lookup[Number(chainId)],                                                                                                                                                                                                 
  399             chainId: Number(chainId),                                                                                                                                                                                                                    
  400             method: "aggregate",                                                                                                                                                                                                                         
  401             args: [multicallTxns],                                                                                                                                                                                                                       
  402             gasLimit: gasEstimation,                                                                                                                                                                                                                     
  403             gasLimitMultiplier: 2,                                                                                                                                                                                                                       
  404             unpermissioned: true,                                                                                                                                                                                                                        
  405             message: `Batch finalized ${multicallTxns.length} txns`,                                                                                                                                                                                     
  406             mrkdwn: `Batch finalized ${multicallTxns.length} txns`,                                                                                                                                                                                      
  407           };                                                                                                                                                                                                                                             
  408           multicallerClient.enqueueTransaction(txnToSubmit);                                                                                                                                                                                             
  409         }
```

bmzig (approved):

PR #2258, opened 2025-05-16
improve(finalizer): Respect finalizer multicall chunk size

pxrl:
Multicall chunk sizes are only set in Config.validate(), so make sure to call that after client initialisation.

### DIFF
--- a/src/finalizer/index.ts
+++ b/src/finalizer/index.ts
@@ -481,9 +481,8 @@ export async function constructFinalizerClients(
   await updateFinalizerClients(commonClients);
 
   if (config.chainsToFinalize.length === 0) {
-    config.chainsToFinalize = commonClients.configStoreClient
-      .getChainIdIndicesForBlock()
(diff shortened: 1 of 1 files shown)


src/finalizer/index.ts
```diff
@@ -476,6 +480,14 @@ export async function constructFinalizerClients(
   const commonClients = await constructClients(_logger, config, baseSigner, hubPoolLookBack);
   await updateFinalizerClients(commonClients);
 
+  if (config.chainsToFinalize.length === 0) {
+    config.chainsToFinalize = commonClients.configStoreClient
+      .getChainIdIndicesForBlock()
+      .filter((chainId) => isDefined(spokePoolClients[chainId]));
```
pxrl, line 486: There's something messed up here; `spokePoolClients` isn't defined at this point, but tsc isn't complaining about it.

In any case, I think I'll just drop the filter statement; if the operator hasn't supplied an array of chainIds then they just inherit the complete set.

bmzig (approved):

PR #2262, opened 2025-05-20
improve: do not assume all spoke pool clients to be evm clients

bmzig:
There are a variety of assumptions that we make throughout the relayer on the format of the spoke pool client which cannot be done post solana.


bmzig:
A lot of times we just assert that the input client is an EVM client since we heavily rely on the network being EVM for so much of our secondary client types.

pxrl (dismissed):

PR #2265, opened 2025-05-21
chore: bump sdk version

grasphoper:


pxrl (approved):
We're bumping sdk 4.2.1 -> 4.2.5. I checked the changes and I don't see anything of concern (perhaps subject to tests all passing).

PR #2266, opened 2025-05-21
fix: Permit ezETH on BSC & World Chain

pxrl:


grasphoper (dismissed):

pxrl:
Marked `do not merge` because there's conjecture that this change might break CEX rebalancing on BSC.

PR #2270, opened 2025-05-24
feat: finalize *tokenless* HubPool CCTP messages

grasphoper:
Current implementation of CCTP finalizer is only tracking `DepositForBurn` events, and thus is not considering other types of CCTP messages. CCTP can send *tokenless* messages via `MessageTransmitter` contract, which emits `MessageSent(bytes message)` event.

We use `HubPool` events `TokensRelayed` and `MessageRelayed` to find relevant tx hashes to search for such `MessageTransmitter` `MessageSent` events in addition to all `DepositForBurn` events.


src/utils/CCTPUtils.ts
```diff
@@ -31,15 +37,27 @@ type CCTPV2APIAttestation = {
   cctpVersion: number;
 };
 type CCTPV2APIGetAttestationResponse = { messages: CCTPV2APIAttestation[] };
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function isCctpV2ApiResponse(
   obj: CCTPAPIGetAttestationResponse | CCTPV2APIGetAttestationResponse
 ): obj is CCTPV2APIGetAttestationResponse {
   return (obj as CCTPV2APIGetAttestationResponse).messages !== undefined;
 }
```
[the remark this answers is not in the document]
grasphoper: Thanks, removed!

src/utils/CCTPUtils.ts
```diff
@@ -245,98 +465,188 @@ async function getCCTPDepositEventsWithStatus(
...
       }
     })
   );
-  return attestedDeposits;
+  return attestedMessages;
 }
 
-function _getPendingV2AttestationStatus(attestation: string): CCTPMessageStatus {
-  return attestation === "pending_confirmation" ? "pending" : "ready";
+function _getPendingV2AttestationStatus(attestation: CCTPV2APIAttestation): CCTPMessageStatus {
```
pxrl, line 640: This looks identical to `_getPendingAttestationStatus()` - should we make attestation a union of `CCTPV2APIAttestation | CCTPAPIGetAttestatuibResponse` and drop the 2nd implementation? Could alternatively do something like `Pick<CCTPV2APIAttestation, "attestation" | "status">`.

src/utils/CCTPUtils.ts
```diff
@@ -114,19 +132,102 @@ export function getCctpDomainForChainId(chainId: number): number {
...
+  srcProvider: RetryProvider,
   sourceChainId: number,
   destinationChainId: number,
   l2ChainId: number,
-  sourceEventSearchConfig: EventSearchConfig
-): Promise<CCTPDepositEvent[]> {
-  const isCctpV2 = isCctpV2L2ChainId(l2ChainId);
+  senderAddresses: string[],
+  sourceEventSearchConfig: EventSearchConfig,
+  includeTokenlessHubPoolMessages: boolean
```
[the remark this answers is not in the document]
grasphoper: The naming might be better here probably. This flag:
- enables scanning for any CCTP events in the HubPool-related txs.
- in practice, this means querying MessageRelayed, TokenRelayed events on HubPool
- if HubPool address is not defined for a sourceChain, this funnction will crash. So I use `includeTokenlessHubPoolMessages = false` for L2 -> L1 CCTP for example

src/utils/CCTPUtils.ts
```diff
@@ -135,91 +236,210 @@ async function getCCTPDepositEvents(
...
+          // Skip non-`DepositForBurn` events
+          return;
+        }
+        if (lastMessageSentEventIdx == -1) {
+          throw new Error(
+            "DepositForBurn event found without corresponding MessageSent event. " +
+              "Each DepositForBurn event must have a preceding MessageSent event in the same transaction. " +
+              `Transaction: ${receipt.transactionHash}, DepositForBurn log index: ${i}`
+          );
+        }
```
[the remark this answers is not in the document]
grasphoper: Sure. Here, we go one by one over a receipts that we "suspect might have CCTP messages for finalization".

It's important to understand what we're looking for in the logs. It's either of 2 things:
- `MessageSent` + `DepositForBurn` == a single USDC transfer from one chain to another (1)
- solo `MessageSent` with no `DepositForBurn` -> a single crosschain message (2)

Then for each receipt we traverse all logs in order, and keep the latest encountered `MessageSent` index in `lastMessageSentEventIdx` . Then if we encounter next `MessageSent` -> the one stored in `lastMessageSentEventIdx` was (2). If we encounter `DepositForBurn`, then we bundle stored `lastMessageSentEventIdx` with it to receive (1)

src/utils/CCTPUtils.ts
```diff
@@ -114,19 +132,102 @@ export function getCctpDomainForChainId(chainId: number): number {
...
+
+    if (isHubPoolAmongSenders) {
+      const hubPool = new Contract(hubPoolAddress, abi, srcProvider);
+
+      const messageRelayedFilter = hubPool.filters.MessageRelayed();
+      const messageRelayedEvents = await paginatedEventQuery(hubPool, messageRelayedFilter, sourceEventSearchConfig);
+      messageRelayedEvents.forEach((e) => txHashesFromHubPool.push(e.transactionHash));
+
+      const tokensRelayedFilter = hubPool.filters.TokensRelayed();
+      const tokensRelayedEvents = await paginatedEventQuery(hubPool, tokensRelayedFilter, sourceEventSearchConfig);
```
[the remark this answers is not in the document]
grasphoper: Indeed. Interestingly, I tried to just create a correct combined topic filter but failed. Can tell if I did it wrong or `paginatedEventQuery` doesn't support that

src/utils/CCTPUtils.ts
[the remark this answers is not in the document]
grasphoper: Renamed to `sourceIsHubChain`

src/utils/CCTPUtils.ts
```diff
@@ -114,19 +132,102 @@ export function getCctpDomainForChainId(chainId: number): number {
...
+ * @param sourceChainId - Chain ID where the messages/deposits originated.
+ * @param destinationChainId - Chain ID where the messages/deposits are targeted.
+ * @param l2ChainId - Chain ID of the L2 chain involved (can be same as `sourceChainId`), used for CCTP versioning.
+ * @param senderAddresses - Addresses that initiated the `DepositForBurn` events. If `includeTokenlessHubPoolMessages` is true, the HubPool address itself is implicitly a sender for its messages.
+ * @param sourceEventSearchConfig - Configuration for event searching on the source chain.
+ * @param includeTokenlessHubPoolMessages - If true, includes messages relayed by the HubPool. **WARNING:** If true, this function assumes a HubPool contract is configured for `sourceChainId`; otherwise, it will throw an error.
+ * @returns A Set of unique transaction hashes.
+ */
+async function getRelevantCCTPTxHashes(
+  srcProvider: RetryProvider,
```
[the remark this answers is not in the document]
grasphoper: `getCachedProvider` is convenient to use and returns a `RetryProvider`. Do you have suggestions on how to egt the `Provider` instead?

src/utils/CCTPUtils.ts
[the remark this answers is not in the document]
grasphoper: [5b8bda4](https://github.com/across-protocol/relayer/pull/2270/commits/5b8bda4d3353dcd573dbe5e4711e967197586e39)

src/utils/CCTPUtils.ts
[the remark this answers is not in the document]
pxrl: Is it fair to expect that a `DepositForBurn` event always follows a `MessageSent` event? If so, I'm wondering if we could first identify all `DepositForBurn` events and then use their `logIndex` fields for a subsequent pass over `MessageSent` events. If a `MessageSent` event has a `logIndex` that immediately precedes a `DepositForBurn` event then it'd be excluded. Could that work? (Motivation for this would be a bit less nesting, at the expense of additional loops - but I think it should be possible to do that in quite compact and concise code).

Also curious about some lines like 351 & 354 - it seems `lastMessageSentEventIdx = i` is performed unconditionally because it sits on either side of the `if` statement. Is that intended?

src/utils/CCTPUtils.ts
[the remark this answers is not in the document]
grasphoper: [9cc365f](https://github.com/across-protocol/relayer/pull/2270/commits/9cc365f550b50b7fe29c032a64a5bb62a035295e)

grasphoper:
@pxrl Did my best to simplify the loop we were talking about. It still is a big chunk of code, but I created one extra fn for readability `getRelevantCCTPEventsFromReceipt` (and for less indentations) and tried to structure the code a bit better. Underlying logic is the same though

src/utils/CCTPUtils.ts
```diff
@@ -154,93 +251,253 @@ async function getCCTPDepositEvents(
...
+        ...depositForBurnLog,
+        event: logDescription.name,
+        args: spreadArgs,
+      },
     };
+    if (_isRelevantEvent(event)) {
+      relevantEvents.push(event);
+    }
+  };
+  // --- END OF Define helper functions --- //
```
nicholaspai, line 416: 100 lines of helper functions within another function seems like a lot and might be hard to maintain... might want to move them outside this function

src/utils/CCTPUtils.ts
```diff
@@ -154,93 +251,253 @@ async function getCCTPDepositEvents(
...
+    1. Pairs: [MessageSent + DepositForBurn] == *cctp token transfers* we need to finalze
+    2. Single MessageSent events(either not followed by any other cctp event we're tracking, or followed by another MessageSent event in the logs array, not necessarily consecutively)
+    == *cctp crosschain message* we need to finalize
+  */
+  // Index in `receipt.logs` of a `MessageSent` log with matching CCTP version
+  let lastMessageSentIdx = -1;
+  receipt.logs.forEach((log, i) => {
+    // --- Try to parse log as `MessageSent` --- //
+    const messageSentVersion = _getMessageSentVersion(log);
+    const isMessageSentEvent = messageSentVersion != -1;
```
nicholaspai, line 430:
```suggestion
    const isMessageSentEvent = messageSentVersion !== -1;
```

src/utils/CCTPUtils.ts
grasphoper: addressed here [e630265](https://github.com/across-protocol/relayer/pull/2270/commits/e630265097023d5f48de6491cede887e94d1502d)

src/utils/CCTPUtils.ts
```diff
@@ -154,141 +251,373 @@ async function getCCTPDepositEvents(
...
+  let lastMessageSentIdx = -1;
+
+  receipt.logs.forEach((log, i) => {
+    // Attempt to parse as `MessageSent`
+    const messageSentVersion = _getMessageSentVersion(log);
+    const isMessageSentEvent = messageSentVersion !== -1;
+
+    if (isMessageSentEvent) {
+      if (_isMatchingCCTPVersion(messageSentVersion, isCctpV2)) {
+        // If we already have a pending MessageSent, process it before overwriting
```
nicholaspai, line 442: Why don't you process the event after you enter `if (isMessageSentEvent) {`?

This way you could remove the `if (lastMessageSentIdx !== -1) ` check and maybe also the `lastMessageSentIdx` variable, which makes this function more difficult to understand

src/utils/CCTPUtils.ts
```diff
@@ -154,141 +251,373 @@ async function getCCTPDepositEvents(
...
+        correspondingMessageSentLog,
+        log,
+        isCctpV2,
+        tokenMessengerInterface
       );
+      if (_isRelevantCCTPEvent(event, sourceDomainId, destinationDomainId, senderAddresses, usdcAddress)) {
+        relevantEvents.push(event);
+      }
+
+      lastMessageSentIdx = -1; // Reset pointer after processing pair
```
nicholaspai, line 486: I think this algorithm makes sense and its efficient because you only need to go through the function once, but its kind of hard to understand with the pointer.

An alternative implementation might be to have two arrays: one with all of the `logs` and one as an  array with only MessageSent type logs. Then, as you go through the `logs` array and you find a `DepositForBurn` event, you can pop off the last `MessageSent` event if and only if the `MessageSent` event's logIndex is less than the deposit event.

Something like that.

I'm not sure that's more readable though, just thinking out loud

src/utils/CCTPUtils.ts
```diff
@@ -297,81 +602,129 @@ export async function getAttestationsForCCTPDepositEvents(
...
+  const destinationMessageTransmitter = new ethers.Contract(address, abi, dstProvider);
+  const attestationResponses: Map<string, CCTPV2APIGetAttestationResponse> = new Map();
+
+  if (isCctpV2) {
+    // For v2, we fetch an API response for every txn hash we have. API returns an array of both v1 and v2 attestations
+    const sourceDomainId = getCctpDomainForChainId(sourceChainId);
+    const uniqueTxHashes = Array.from(new Set([...messagesWithStatus.map((message) => message.log.transactionHash)]));
+
+    // Circle rate limit is 35 requests / second. To avoid getting banned, batch calls into chunks with 1 second delay between chunks
+    // For v2, this is actually required because we don't know if message is finalized or not before hitting the API. Therefore as our
```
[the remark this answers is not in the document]
nicholaspai: right I ran into this issue as well.

src/utils/CCTPUtils.ts
```diff
@@ -297,81 +626,134 @@ export async function getAttestationsForCCTPDepositEvents(
...
+    destinationMessageTransmitter = new ethers.Contract(address, abi, dstProvider);
+
+    // For v2, we fetch an API response for every txn hash we have. API returns an array of both v1 and v2 attestations
+    const sourceDomainId = getCctpDomainForChainId(sourceChainId);
+    const uniqueTxHashes = Array.from(new Set([...messagesWithStatus.map((message) => message.log.transactionHash)]));
+
+    // Circle rate limit is 35 requests / second. To avoid getting banned, batch calls into chunks with 1 second delay between chunks
+    // For v2, this is actually required because we don't know if message is finalized or not before hitting the API. Therefore as our
+    // CCTP v2 list of chains grows, we might require more than 35 calls here to fetch all attestations
+    const chunkSize = 8;
```
nicholaspai, line 657: How is 8 related to 35..? Not seeing the logic on how you picked 8 here.

nicholaspai: Also, how many chunks would we send today? For exmaple, running the finalizer with realistic configs, I'm curious what this chunk # looks like. It'd help me better understand the usage of this function

grasphoper: 8 is just a conservative limit in light of a 35 real limit. Which also doesn't even fully protect us if we're trying to finalize multiple CCTP chains in parallel.

This code is being called for a single src -> dst chain. So the maximum number of unique tx hashes based on a week lookback time is 24*7=168, so ~20 seconds to query this. However realistically we're never sending a CCTP every bundle, so it's more like 2*7=14, so 2s to query. However, if we're querying multiple chains, we still may get banned on the API side.

If we find that we're getting banned, one solution would be to just put these responses from CCTP API in redis

grasphoper:
> Have you tried running this code using existing finalizer configs? I think this is a prime candidate for integration testing with live inputs

Good point. Haven't tested with full configs

src/utils/CCTPUtils.ts
grasphoper: Just remembered why I didn't implement this solution initially. Because when you're popping events from the messageSentArray, imagine you have:

[messageSent, messageSent, deposit, deposit]

Then when processing deposit at index 3 (0-indexed), you'll match it with messageSent from index 0.

That said, the solution ends up looking cleaner. And we do have this protection in play:
```
  // Ensure data integrity between MessageSent and DepositForBurn events
  if (
    !compareAddressesSimple(eventData.sender, cctpAddressToBytes32(spreadArgs.depositor)) ||
    !compareAddressesSimple(eventData.recipient, cctpAddressToBytes32(spreadArgs.mintRecipient)) ||
    !BigNumber.from(eventData.amount).eq(spreadArgs.amount)
  ) {
```

And what is more, the correct log sequence should never look like the one in my example, it's only possible if the RPC output is incorrect.

Taking all this into account, I chose to impl. the solution here [3df04aa](https://github.com/across-protocol/relayer/pull/2270/commits/3df04aa516cd6170cc93375bff564508a3783b49). LMK if you agree that it looks cleaner

grasphoper:
@pxrl @nicholaspai @bmzig 

Completed the integration testing for this PR: 

- Solana -> Ethereum token transfer: [solana send](https://solscan.io/tx/5p68rg1uWEYmD4N3bvsKogtLE3rF8Cyd2LSQvHanPpe9pCkNtCQUfj3DCSf155MhfwS9ow1by4Kaf4NEcEuB87hn), [ethereum finalization](https://etherscan.io/tx/0x8476ff27e1b87ab0d01efbb558aa61383c668c4a8321c7045e1f6a99d159f8f4)
- [Execute bundle on ethereum](https://etherscan.io/tx/0x2a91d0d1b345c53d423d6e7c6e812d9d38dc9038820a52530fa736d0c9c7af26) (send message to Solana to bridge some funds back) -> [finalize message on Solana side](https://solscan.io/tx/2Y4rN68FgzpUSsb5vdE4zdASmaVwPcPM4wE8ChWJvogSBEvswXFDG5FjbjddrBqRcF9g6hQuVubR3UZk8vcoRU5M) -> [spoke sends money back](https://solscan.io/tx/4m3cketRxfoyT9GqosGn5d1v3L6ug5Vng9xa35zztt2LumG4sna6oyUuBdoruzWNb27qjx1YP2XNJAwZDDUzBgCC) -> [finalize on ethereum side](https://etherscan.io/tx/0xfc6d77fc82d614067f090131085e95bf6c41dd1d7d2cb2743dabf050448397f5)
- confirmed that regular cctp finalizations are working as expected between evm chains
- ran with fast finalizer config (most of the relevant config: chains / addrs). Same with sweeper config, lead to a small bugfix

A nuance about l2 -> l1 finalizations from solana side. There's an issue with addresses tracking. In `src/finalizer/index.ts`, we define `senderAddresses` that we pass on into finalizers. We use those to understand which CCTP events we want to try to finalize (e.g. when depositor is in `senderAddresses`). We also enforce evm-style format for addresses in `senderAddresses` [here](https://github.com/across-protocol/relayer/blob/2d0740b9d36c05c8ce941e3932b83508d65e2b8c/src/finalizer/index.ts#L256).

Now, for EVM events this works fine as we can do smth like `senderAddresses.includes(depositForBurnEvent.depositor)` to check if we need to finalize that message. 

For Solana we can't do the same, because events on Solana side emit base58 32byte long addresses for CCTP events.

It leads to 2 issues(not bugs, but just fragile setups):
- for EOAs, we need to: take Solana address we actually want to finalize depositForBurns from -> convert to hex -> trim -> insert into env `FINALIZER_WITHDRAWAL_TO_ADDRESSES`
- for SpokePool, see `l2ToL1.ts`, function `augmentSendersListForSolana` for how it's included in `senderAddresses`. Then compared against events in `CCTPUtils.ts` like so:

```
const trimmedDepositor = cctpBytes32ToAddress(SvmAddress.from(eventData.depositor, "base58").toBytes32());
...
senderAddresses.some((addr) => compareAddressesSimple(addr, trimmedDepositor))
```

There's another solution possible for EOAs: use a separate env var instead of `FINALIZER_WITHDRAWAL_TO_ADDRESSES`: `FINALIZER_CCTP_SOLANA_EOAS`. Have base58 EOAs there and have `l2ToL1.ts` and `CCTPUtils.ts` instead work with base58 `senderAddresses` directly. I think I like that solution better, it's implemented in a separate branch [here](https://github.com/across-protocol/relayer/compare/if/finalize-cctp-raw-messages2...if/finalizer-sender-addrs-solana)

src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -34,8 +39,14 @@ export async function cctpL2toL1Finalizer(
     to: spokePoolClient.latestHeightSearched,
     maxLookBack: spokePoolClient.eventSearchConfig.maxLookBack,
   };
-  const outstandingDeposits = await getAttestationsForCCTPDepositEvents(
-    senderAddresses,
+
+  const finalizingFromSolana = [CHAIN_IDs.SOLANA, CHAIN_IDs.SOLANA_DEVNET].includes(spokePoolClient.chainId);
```
nicholaspai, line 43: nit: can you use `chainIsSvm` here

src/utils/CCTPUtils.ts
```diff
@@ -297,81 +629,134 @@ export async function getAttestationsForCCTPDepositEvents(
...
-      if (isCctpV2ApiResponse(attestation)) {
-        const attestationForDeposit = attestation.messages[count];
-        const destinationMessageTransmitter = new ethers.Contract(address, abi, dstProvider);
-        const processed = await _hasCCTPMessageBeenProcessed(
-          attestationForDeposit.eventNonce,
+      if (isCctpV2) {
+        const attestations = attestationResponses.get(message.log.transactionHash);
+
+        const matchingAttestation = attestations.messages.find((apiAttestation) => {
+          return cmpAPIToEventMessageBytesV2(apiAttestation.message, message.messageBytes);
```
nicholaspai, line 689: +1, this should solve the problem today where we break if the API returns messages in the wrong `logIndex` order

PR #2273, opened 2025-05-26
feat(relayer): Barebones SVM listener implementation

pxrl:
This change lays the groundwork for an SVM implementation. It's extremely barebones and only listens for new slots. Follow-up work will add event subscriptions.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,14 +1,13 @@
 import assert from "assert";
 import { ChildProcess, spawn } from "child_process";
-import { Contract } from "ethers";
 import { clients, utils as sdkUtils } from "@across-protocol/sdk";
 import { Log, DepositWithBlock } from "../interfaces";
-import { CHAIN_MAX_BLOCK_LOOKBACK, RELAYER_DEFAULT_SPOKEPOOL_LISTENER } from "../common/Constants";
+import { RELAYER_SPOKEPOOL_LISTENER_EVM, RELAYER_SPOKEPOOL_LISTENER_SVM } from "../common/Constants";
 import {
-  EventSearchConfig,
+  Address,
+  chainIsSvm,
   getNetworkName,
   isDefined,
-  MakeOptional,
   winston,
   getRelayEventKey,
   getMessageHash,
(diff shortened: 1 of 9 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -43,256 +42,271 @@ export function isSpokePoolEventRemoved(message: unknown): message is SpokePoolE
...
+  chainId: number;
+  spokePoolAddress: Address;
+  deploymentBlock: number;
+  _queryableEventNames: () => string[];
+  eventSearchConfig: { from: number; to?: number; maxLookBack?: number };
+  depositHashes: { [depositHash: string]: DepositWithBlock };
+  logger: winston.Logger;
+};
+
+export function SpokeListener<T extends Constructor<MinGenericSpokePoolClient>>(SpokePoolClient: T) {
```
pxrl, line 69: Just calling this out for discussion. The existing SpokePool listener construct relies on extending a single class (EvmSpokePoolClient) with messaging capabilities. Implementing a generic class that can extend either an EVM or SVM SpokePoolClient turned out to be trickier than expected, but I found a way with Typescript mixins. This permits the SpokeListener to extend some minimal description of a SpokePoolClient. Taking this approach avoids duplicating or otherwise refactoring all of the listener/messaging implementation out, which would be a maintenance headache.

src/clients/SpokePoolClient.ts
```diff
@@ -43,256 +42,271 @@ export function isSpokePoolEventRemoved(message: unknown): message is SpokePoolE
...
+      if (pendingEventIdx !== -1) {
+        removed = true;
+
+        // Drop the relevant event.
+        pendingEvents.splice(pendingEventIdx, 1);
+
+        this.logger.debug({
           at: "SpokePoolClient#removeEvent",
-          message: `Removed 1 pre-ingested ${this.chain} ${eventName} event.`,
+          message: `Removed ${this.#chain} ${eventName} event for block ${blockNumber}.`,
```
pxrl, line 240: Did not intend to drop this.

```suggestion
          message: `Removed 1 pre-ingested ${this.#chain} ${eventName} event for block ${blockNumber}.`,
```

src/libexec/RelayerSpokePoolListenerSVM.ts
```diff
@@ -0,0 +1,249 @@
...
+    logger.debug({ at: "Relayer::run", message: `${chain} parent disconnected, stopping...` });
+    abortController.abort();
+  });
+
+  const eventsClient = await arch.svm.SvmCpiEventsClient.create(getSvmProvider());
+  if (latestSlot > startSlot) {
+    const events = [
+      "FundsDeposited",
+      "FilledRelay",
+      // "RequestedSpeedUpDeposit", @todo supported?
```
pxrl:
```suggestion
```

src/libexec/RelayerSpokePoolListenerSVM.ts
```diff
@@ -0,0 +1,243 @@
...
+  } else if (isDefined(lookback)) {
+    // Resolve `lookback` seconds from head to a specific block.
+    assert(Number.isInteger(Number(lookback)), `Invalid lookback (${lookback})`);
+
+    const now = await provider.getBlockTime(latestSlot).send();
+    assert(typeof now === "bigint"); // Should be unnecessary; tsc still complains.
+    startSlot = BigInt(
+      Math.max(
+        deploymentBlock,
+        await getBlockForTimestamp(chainId, Number(now - lookback), blockFinder, await getRedisCache())
```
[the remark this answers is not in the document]
pxrl: Good catch - tsc doesn't actually complain about it, but if I fire node up then it doesn't like it. Updated: https://github.com/across-protocol/relayer/pull/2273/commits/f4fe83da4f938b58397015cb38385ebacdbee931

src/libexec/RelayerSpokePoolListenerSVM.ts
```diff
@@ -0,0 +1,246 @@
...
+  eventsClient: arch.svm.SvmCpiEventsClient,
+  eventNames: string[],
+  quorum = 1
+): Promise<void> {
+  const urls = Object.values(getNodeUrlList(chainId, quorum, "wss"));
+  const nProviders = urls.length;
+  assert(nProviders >= quorum, `Insufficient providers for ${chain} (required ${quorum} by quorum)`);
+
+  const eventAuthority = await arch.svm.getEventAuthority();
+  const config = { commitment: "confirmed" } as const;
```
md0x, line 116: We can make the commitment level a parameter so the relayer can choose. “Confirmed” isn’t the highest certainty, it just means a super-majority (≥66 %) has voted for the block. Depending on the relayer’s strategy, we might prefer “[finalized](https://www.helius.dev/blog/solana-commitment-levels#finalized)” instead.

pxrl: @md0x For the EVM side we always take "latest", and the relayer separately implements deposit confirmation requirements downstream of this.

One question though - any idea how Solana notifies of a re-orged/dropped event? On the EVM side each event has a `removed` boolean; I didn't see anything similar on the SVM side...

md0x: Ok, if the relayer implements its own deposit confirmation logic, that works. I was just thinking that listening to events with a finalized commitment is an equivalent form of validation since in that case a reorg is (practically) impossible.

Otherwise, I believe Solana doesn’t tag logs with a removed boolean like Ethereum. However, you can subscribe via websockt using `slotsUpdatesSubscribe` and watch for events with type "dead". See: https://solana.com/docs/rpc/websocket/slotsupdatessubscribe

An alternative approach: if you store the blockhash, you can later call `isBlockhashValid` with commitment: "finalized" to check whether the block still exists on the chain.

Reinis-FRP (approved):

PR #2275, opened 2025-05-27
improve(monitor): Permit blanket opt-out of follow distance

pxrl:
The follow distance can cause under-reporting on pending repayments for any fills that were made within the default follow distance. In general this is undesirable for the monitor. The operator could override this manually for each chain, but that's a maintenance overhead when new chains are added. Instead, permit a global opt-out of the follow distance configurable.


pxrl:
> Does it make sense to have a follow distance for the monitor?
> 
> I.e. could we make the follow distance opt-in instead of opt-out like we have with `MONITOR_IGNORE_FOLLOW_DISTANCE`?

Fair call - I really don't think it's useful for the monitor and I'm on board with making it opt-in rather than opt-out :+1: 

https://github.com/across-protocol/relayer/pull/2275/commits/39f354b120195c9eaadf5b99442a5fa842ff28a6

PR #2290, opened 2025-06-09
feat: Added yalc usage instructions

fusmanii:


nicholaspai (approved):

pxrl (approved):
@fusmanii I read a bit about `yalc` - it seems to be aimed at addressing shortcomings in other packages (i.e. yarn, eslint, jest, ...) ([ref](https://github.com/yarnpkg/yarn/issues/1761#issuecomment-259706202).). However, it's unchanged for the past 2 years, and I'm wondering whether some or all of the issues that we're seeing would be addressed by some combination of:
- Moving off yarn 1.
- Upgrading dependencies.
- Restructuring repository imports/dependencies for simplicity (i.e. especially RE importing contracts directly and via the SDK).

It'd be interesting to see whether some or all of these issues disappear just by cleaning house.

PR #2292, opened 2025-06-10
improve: Relocate invalid fill logging to monitor

pxrl:
This logging is currently emitted by SDK implementation, irrespective of which bot personality is running. The relayer is focused on speed and the large number of perceived invalid fills is causing slowdowns and rate-limiting on messages to Slack. Relocate this to the monitor to ensure that the log messages are emitted at a more sensible rate.


src/monitor/Monitor.ts
pxrl, line 1: nb. it probably makes sense for this change to be conditional on this SDK PR (https://github.com/across-protocol/sdk/pull/1080), such that we don't end up emitting duplicate logs.

pxrl:
@dijanin-brat We have some additional improvements that could be made to invalid fill logging; we can sync on that later.

src/monitor/Monitor.ts
```diff
@@ -193,8 +193,28 @@ export class Monitor {
     const unfilledDeposits: Record<number, DepositWithBlock[]> = Object.fromEntries(
       await mapAsync(Object.values(spokePoolClients), async ({ chainId: destinationChainId }) => {
         const deposits = getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient).map(
-          ({ deposit }) => deposit
+          ({ deposit, invalidFills: invalid }) => {
+            if (invalid.length > 0) {
+              const origin = getNetworkName(deposit.originChainId);
+              const invalidFills = Object.fromEntries(
+                invalid.map(({ relayer, destinationChainId, depositId, txnRef }) => {
```
pxrl, line 200: tbd whether it makes sense to filter these for depositId < UINT32_MAX, because deterministic deposit IDs will often cause erroneous logging.

nicholaspai: I think should ignore those deterministic deposit ID's for now, thoughts?

nicholaspai (dismissed):

nicholaspai (approved):

nicholaspai (dismissed):

nicholaspai (dismissed):

PR #2294, opened 2025-06-10
feat: report Binance limits in the monitor

bmzig:
Small addition to report daily withdrawal quotas in the monitor and warn if the used quota is below a configured threshold.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -519,7 +519,7 @@ export class Monitor {
     const binanceApi = await getBinanceApiClient(process.env["BINANCE_API_BASE"]);
     const wdQuota = await getBinanceWithdrawalLimits(binanceApi);
     const aboveThreshold = wdQuota.usedWdQuota / wdQuota.wdQuota > this.monitorConfig.binanceWithdrawWarnThreshold;
-    this.logger[aboveThreshold ? "error" : "debug"]({
+    this.logger[aboveThreshold ? "warn" : "debug"]({
       at: "Monitor#checkBinanceWithdrawalLimits",
       message: "Binance withdrawal quota",
       wdQuota,


src/monitor/MonitorConfig.ts
```diff
@@ -82,6 +83,7 @@ export class MonitorConfig extends CommonConfig {
       unknownRootBundleCallersEnabled: UNKNOWN_ROOT_BUNDLE_CALLERS_ENABLED === "true",
       stuckRebalancesEnabled: STUCK_REBALANCES_ENABLED === "true",
       spokePoolBalanceReportEnabled: REPORT_SPOKE_POOL_BALANCES === "true",
+      binanceWithdrawalLimitsEnabled: CHECK_BINANCE_WITHDRAWAL_LIMITS === "true",
```
pxrl, line 86: @bmzig Could it make sense to infer this based on whether `BINANACE_WITHDRAW_WARN_THRESHOLD` is 0 or not?

bmzig: Yeah https://github.com/across-protocol/relayer/pull/2294/commits/c612b9ca174959cf35aa6148b9082d1cd6b6f541

src/monitor/Monitor.ts
```diff
@@ -513,6 +515,17 @@ export class Monitor {
     }
   }
 
+  async checkBinanceWithdrawalLimits() {
+    const binanceApi = await getBinanceApiClient(process.env["BINANCE_API_BASE"]);
+    const wdQuota = await getBinanceWithdrawalLimits(binanceApi);
+    const aboveThreshold = wdQuota.usedWdQuota / wdQuota.wdQuota > this.monitorConfig.binanceWithdrawWarnThreshold;
```
nicholaspai, line 521: Do we need parantheses around `wdQuota.usedWdQuota / wdQuota.wdQuota`? Also can the usedWdQuota ever be 0?

nicholaspai (commented):
Couple q's

src/monitor/Monitor.ts
bmzig: Lint removes the parentheses, so I'm assuming it's OK. Also, `usedWdQuota` can be 0 (but `wdQuota` can't). If the env var isn't set, then we warn by default.

src/monitor/Monitor.ts
```diff
@@ -513,6 +515,17 @@ export class Monitor {
...
+    this.logger[aboveThreshold ? "error" : "debug"]({
```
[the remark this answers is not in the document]
bmzig: True, we probably don't want an error here. I changed to warn. https://github.com/across-protocol/relayer/pull/2294/commits/085bbd4c0e89a98fa586c91e46a63d08449fe3b9

PR #2295, opened 2025-06-10
improve(InventoryClient): Don't skip BSC deposits

nicholaspai:
The current IM logic will return zero possible repayment chains if the deposit forces origin chain repayment and the origin chain is over allocated. This lets us avoid accruing too much balance on these chains, typically lite chains, that force origin chain repayment.

However, BSC Is now included as one of these chains since it does force origin chain repayment. However, BSC is relatively easy to rebalance tokens out of--so we shouldn't impose the same constraints.

This PR therefore allows the IM to return the origin chain as a possible repayment chain if the origin chain is BSC.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -30,6 +30,7 @@ import {
   getRemoteTokenForL1Token,
   getTokenInfo,
   isEVMSpokePoolClient,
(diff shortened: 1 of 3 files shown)


pxrl (approved):
LGTM! I really like that this is such a light touch in the inventory client.

PR #2300, opened 2025-06-12
improve(relayer): Catch IPC failures

pxrl:
I'd previously (mis)understood that process.send() should be actively set undefined if the IPC has been closed, but this is not the documented behaviour, nor what is observed in practice.

I think this was previously OK when ethers websocket transports were used, but viem seems to suppress the exception so that the process continues to run despite being an orphan.

### DIFF
--- a/src/libexec/util/ipc.ts
+++ b/src/libexec/util/ipc.ts
@@ -22,8 +22,10 @@ export function postEvents(blockNumber: number, currentTime: number, events: Log
     nEvents: events.length,
     data: JSON.stringify(events, sdkUtils.jsonReplacerWithBigNumbers),
   };
+
+  const msg = JSON.stringify(message);
   try {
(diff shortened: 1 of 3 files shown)


src/libexec/util/ipc.ts
```diff
@@ -22,7 +22,13 @@ export function postEvents(blockNumber: number, currentTime: number, events: Log
     nEvents: events.length,
     data: JSON.stringify(events, sdkUtils.jsonReplacerWithBigNumbers),
   };
-  process.send(JSON.stringify(message));
+  try {
+    process.send(JSON.stringify(message));
```
mrice32, line 26: nit: I would suggest pulling the json stringify out of the try catch to avoid swallowing unrelated json issues.

mrice32 (approved):

PR #2306, opened 2025-06-16
refactor: Drop loop support in Dataworker

pxrl:
This isn't used in practice and it's hard to see a concrete use case for it. Drop it to remove a level of indentation.

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -1,12 +1,4 @@
-import {
-  winston,
-  config,
-  startupLogLevel,
-  Signer,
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):
lgtm

src/dataworker/index.ts
pxrl, line 1: The diff here is pretty messy; the only functional change is to remove the for loop and the loop termination logic, so ideally the diff would mostly show a reduction in indentation, but it's unfortunately not handling that very well.

nicholaspai (approved):

PR #2308, opened 2025-06-17
improve(CCTPFinalizer): Add robustness for case where CCTP API returns V2 attestations in unexpected order

nicholaspai:
We've seen a real situation where the batched `DepositForBurn` events in a [transaction](https://etherscan.io/tx/0xa64e3d831c3a0c94a25863d5fa3627f9626ccded551edab147fddae6b28f1d82#eventlog) are in a different order than the messages returned by the [CCTP API Response](https://iris-api.circle.com/v2/messages/0?transactionHash=0xa64e3d831c3a0c94a25863d5fa3627f9626ccded551edab147fddae6b28f1d82):
- logIndex 125 corresponds to API message index 3
- logIndex 173 corresponds to API message 2

This PR adds a simple hacky solution that we can use to specifically add robustness in the L1 to L2 case where we batch sending CCTP deposits from L1 to L2. It takes advantage of the fact that we're almost never sending multiple L1 deposits to the same L2. This could happen if we are sending more than 1mil USDC from L1 to L2

### DIFF
--- a/src/utils/CCTPUtils.ts
+++ b/src/utils/CCTPUtils.ts
@@ -328,7 +328,28 @@ export async function getAttestationsForCCTPDepositEvents(
       // We can remove this V2 custom logic once we can derive nonceHashes locally and filter out already "finalized"
       // deposits in getCCTPDepositEventsWithStatus().
       if (isCctpV2ApiResponse(attestation)) {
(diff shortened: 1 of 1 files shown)


nicholaspai:
This might not be necessary since we have been notified that Circle API will start ordering by logIndex

PR #2309, opened 2025-06-17
fix(Dataworker): Make sure that each spoke pool client start-search timestamp is > maxFillDeadlineHours earlier than lowest bundle start block timestamp

nicholaspai:
This will ensure the Dataworker throws whenever a single chain halts for a long time, causing the subsequent bundle block ranges to show a very early start timestamp for this newly unhalted chain, relative to other chains.

For now, the user will have to manually reset DATAWORKER_FAST_LOOKBACK_COUNT to a higher value to overcome this hurdle but at least an invalid proposal won't be submitted

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -22,13 +22,14 @@ import {
   buildSlowRelayTree,
   fixedPointAdjustment,
   getRefundsFromBundle,
-  getTimestampsForBundleEndBlocks,
+  getTimestampsForBundleStartBlocks,
   isDefined,
   MerkleTree,
   Profiler,
   winston,
   getNativeTokenSymbol,
   getWrappedNativeTokenAddress,
+  MAX_UINT_VAL,
 } from "../utils";
 import { DataworkerClients } from "./DataworkerClientHelper";
 import {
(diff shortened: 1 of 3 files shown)


src/dataworker/DataworkerUtils.ts
```diff
@@ -53,15 +53,30 @@ export async function blockRangesAreInvalidForSpokeClients(
   isV3 = false
 ): Promise<InvalidBlockRange[]> {
   assert(blockRanges.length === chainIdListForBundleEvaluationBlockNumbers.length);
-  let endBlockTimestamps: { [chainId: number]: number } | undefined;
+  let earliestStartBlockTimestamp: { timestamp: number; chainId: number } | undefined = undefined;
```
nicholaspai, line 56: If we use the earliest start block, instead of the end block, then we should be resistant to re-orgs, since the start block of the next bundle block range post-reorg will always be the same, regardless of how long the chain was halted for

src/dataworker/DataworkerUtils.ts
```diff
@@ -53,15 +53,30 @@ export async function blockRangesAreInvalidForSpokeClients(
...
+        if (timestamp < currMax.timestamp) {
+          currMax = {
+            chainId: Number(chainId),
+            timestamp,
+          };
+        }
+        return currMax;
+      },
+      {
+        timestamp: Object.values(startBlockTimestamps)[0],
```
bmzig, line 76: I don't think this is that big a deal since we would only log something wrong, but if Ethereum has the earliest block timestamp, it looks like we would return `{ timestamp: [eth_timestamp], chainId: -1 }`. I think we should change `chainId` to `chainId` and `timestamp` to `timestamp`.

nicholaspai: w.b. d068afec5958c39ffd4484b3ee58492c46fb712a

PR #2319, opened 2025-06-20
fix: Relayer fill log for tokens with different decimals

fusmanii:
e.g. USDC on BSC is 18 decimal and 6 everywhere else

Updating the input amount to match the output amount decimals

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -50,6 +50,18 @@ const txnRetryable = (error?: unknown): boolean => {
   return expectedRpcErrorMessages.has((error as Error)?.message);
 };
 
+const isFillRelayError = (error: unknown): boolean => {
+  const fillRelaySelector = "0xdeff4b24"; // keccak256("fillRelay()")[:4]
+  const multicallSelector = "0xac9650d8"; // keccak256("multicall()")[:4]
+
+  const errorStack = (error as Error).stack;
+  const isFillRelayError = errorStack?.includes(fillRelaySelector);
+  const isMulticallError = errorStack?.includes(multicallSelector);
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -1509,15 +1509,18 @@ export class Relayer {
     const depositor = blockExplorerLink(deposit.depositor, deposit.originChainId);
     const inputAmount = createFormatFunction(2, 4, false, decimals)(deposit.inputAmount.toString());
 
+    const { symbol: outputTokenSymbol, decimals: outputTokenDecimals } =
+      this.clients.hubPoolClient.getTokenInfoForAddress(deposit.outputToken, deposit.destinationChainId);
+
     let msg = `Relayed depositId ${deposit.depositId.toString()} from ${srcChain} to ${dstChain} of ${inputAmount} ${symbol}`;
     const realizedLpFeePct = formatFeePct(_realizedLpFeePct);
-    const _totalFeePct = deposit.inputAmount
+    const adjustedInputAmount = deposit.inputAmount.mul(outputTokenDecimals).div(decimals);
```
nicholaspai, line 1517: I think you should use `sdkUtils.ConvertDecimals(decimals, outputTokenDecimals)` here

src/relayer/Relayer.ts
```diff
@@ -1509,15 +1509,18 @@ export class Relayer {
...
+    const _totalFeePct = adjustedInputAmount
       .sub(deposit.outputAmount)
       .mul(fixedPointAdjustment)
```
nicholaspai, line 1520: Maybe can also use ConvertDecimals here, this time going from outputTokenDecimals to fixedPointAdjustment

PR #2345, opened 2025-07-02
fix: update monitor indexing and finalizer address casting

bmzig:


nicholaspai (dismissed):

src/monitor/Monitor.ts
```diff
@@ -734,7 +737,7 @@ export class Monitor {
         : this.monitorChains;
 
     const l2TokenForChain = (chainId: number, symbol: string) => {
-      return TOKEN_SYMBOLS_MAP[symbol]?.addresses[chainId];
+      return toAddressType(TOKEN_SYMBOLS_MAP[symbol]?.addresses[chainId], chainId);
```
nicholaspai, line 740: Unrelated to this PR but this function feels a bit inefficient for us to keep maintaining because its replicating a lot of the bundle data logic and its a bit of a beast. I wonder if we can remove it soon and either call the indexer or replace it otherwise

nicholaspai (dismissed):

src/monitor/Monitor.ts
pxrl, line 740: If `TOKEN_SYMBOLS_MAP[symbol]` is undefined then this will throw.

```suggestion
      const token = TOKEN_SYMBOLS_MAP[symbol];
      return isDefined(token) ? toAddressType(token.addresses[chainId], chainId) : undefined;
```

src/monitor/Monitor.ts
```diff
@@ -881,7 +884,7 @@ export class Monitor {
       pendingRelayerRefunds[chainId] = {};
       l2TokenAddresses.forEach((l2Token) => {
         const pendingValidatedDeductions = accountedBundleRefunds
-          .map((refund) => refund[chainId]?.[l2Token])
+          .map((refund) => refund[chainId]?.[l2Token.toBytes32()])
```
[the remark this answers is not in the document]
bmzig: It's imposed by the bundle data client.

PR #2346, opened 2025-07-02
improve: use lookup tables, fix account metas in solana executor

bmzig:
Ideally, we should use LUTs for relayer refund leaves to compress tx accounts. Also, the remaining accounts for `executeRelayerRefundLeafDeferred` were incorrectly set to the relayer ATAs instead of the relayer claim accounts.


src/dataworker/Dataworker.ts
```diff
@@ -2968,15 +3018,57 @@ export class Dataworker {
...
+        payer: kitKeypair,
+        addresses: lutAddresses,
+      });
+
       const executeRelayerRefundLeafDeferredTx = pipe(
         createTransactionMessage({ version: 0 }),
         (tx) => setTransactionMessageFeePayer(kitKeypair.address, tx),
         (tx) => setTransactionMessageLifetimeUsingBlockhash(recentBlockhash.value, tx),
-        (tx) => appendTransactionMessageInstructions([executeRelayerRefundLeafDeferredIx], tx)
+        (tx) => appendTransactionMessageInstructions([extendLookupTableIx, executeRelayerRefundLeafDeferredIx], tx),
```
bmzig, line 3048: I suspect this cannot be done atomically but it's probably worth testing since if it can, it would save a whole lot of complexity (since the data we populate the LUT with differs based on the type of relayer refund leaf execution).

src/dataworker/Dataworker.ts
```diff
@@ -2987,15 +3058,57 @@ export class Dataworker {
...
+        (tx) => compressTransactionMessageUsingAddressLookupTables(tx, addressByLookupTableAddresses)
+      );
+      const refundLeafSignature = await runTransactionSvm(executeRelayerRefundLeafDeferredTx, kitKeypair, provider);
+
+      // We should clean up the LUT now before returning.
+      const deleteLutIx = getCloseLookupTableInstruction({
```
Reinis-FRP, line 3094: one strategy on dealing with ALT lifecycle would be to treat them as more permanent and extend them only as needed without closing. Assuming the relayers would not be rotating their repayment addresses on each fill we could reuse their ALT and extend it as needed.

we could leave this also for a follow-up improvement, but if there is still deactivation and cooldown requirement this would anyway demand a separate garbage collector process to deal with.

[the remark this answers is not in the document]
bmzig: You're right. Also this is related to the comment below, but testing this, it seems like the cooldown requirement is a thing, so the best I can do now is deactivate it and let it be handled in the future. https://github.com/across-protocol/relayer/pull/2346/commits/da719c5cbd72e4168b3fe606c6aaf8880f724d51

src/dataworker/Dataworker.ts
```diff
@@ -2964,17 +2998,54 @@ export class Dataworker {
...
+        addresses: lutAddresses,
+      });
+
+      const extendLutTx = pipe(
+        createTransactionMessage({ version: 0 }),
+        (tx) => setTransactionMessageFeePayer(kitKeypair.address, tx),
+        (tx) => setTransactionMessageLifetimeUsingBlockhash(recentBlockhash.value, tx),
+        (tx) => appendTransactionMessageInstructions([extendLookupTableIx], tx)
+      );
+      await runTransactionSvm(extendLutTx, kitKeypair, provider);
```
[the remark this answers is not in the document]
bmzig: Good point. Fixed here. https://github.com/across-protocol/relayer/pull/2346/commits/da719c5cbd72e4168b3fe606c6aaf8880f724d51

src/dataworker/Dataworker.ts
```diff
@@ -2964,17 +2998,54 @@ export class Dataworker {
...
+      });
+      executeRelayerRefundLeafIx.accounts.push(...refundATAs);
+      const lutAddresses = Object.values(executeRelayerRefundLeafIx.accounts).map((account) =>
+        address(account.address)
+      );
+      const addressByLookupTableAddresses: AddressesByLookupTableAddress = {
+        [lookupTablePda]: lutAddresses,
+      };
+
+      const extendLookupTableIx = getExtendLookupTableInstruction({
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/relayer/pull/2346/commits/280e97614f5f2e7bb57b434fc2ff3e7957ffefd6 I updated it to extend the ALT by 32 addresses at a time. I  haven't been able to verify if we can do that much/if we can do more, but the contracts script uses 30 at a time (and 32 divides 256, so we can have an even amount of transactions before the LUT gets maxed out). I do not have logic here which accomodates a refund which would require > 256 accounts (and therefore two or more lookup tables, though). I think that can be a follow-up.

src/dataworker/Dataworker.ts
Reinis-FRP: No, I think the limit is 256 addresses per lookup table as they are referenced as `u8` indices in V0 transactions. But its also possible to use multiple lookup tables.

src/dataworker/Dataworker.ts
```diff
@@ -2863,77 +2897,243 @@ export class Dataworker {
...
       });
     }
 
-    // The remaining accounts for a relayer refund leaf are the refund addresses
-    const _l2TokenAddress = leaf.l2TokenAddress;
-    assert(
-      _l2TokenAddress.isSVM(),
-      `Dataworker#executeRelayerRefundLeafSvm: Attempting to execute a relayer refund leaf with token address ${leaf.l2TokenAddress}`
+    // Get the lookup table Pda.
+    const recentSlot = (await provider.getSlot({ commitment: "finalized" }).send()) as bigint;
```
pxrl, line 2911: Is it important for a block to exist at this slot? If so, it's probably better to use `const { slot: recentSlot} = await getNearestSlotTime({ commitment: "finalized" });`

(The name is a bit counter-intuitive; it returns both a slot and a timestamp).

src/utils/TransactionUtils.ts
```diff
@@ -203,15 +202,35 @@ export async function runTransaction(
...
+  };
+  const signedTx = await signTransactionMessageWithSigners(unsignedTransaction);
   const serializedTx = getBase64EncodedWireTransaction(signedTx);
-  return provider.sendTransaction(serializedTx, { encoding: "base64" }).send();
+  const txSignature = await provider
+    .sendTransaction(serializedTx, { preflightCommitment: "confirmed", skipPreflight: false, encoding: "base64" })
+    .send();
+  let confirmed = false;
+  let _cycles = 0;
+  while (!confirmed && ++_cycles < cycles) {
```
pxrl, line 222: Does pre-increment make sense for this? A caller could supply `cycles = 1` and it'd bail without ever doing anything.

pxrl (approved):

src/utils/TransactionUtils.ts
bmzig: Good point: https://github.com/across-protocol/relayer/pull/2346/commits/b65e05b2852e686b292245d7f20423d7302eb91e

src/dataworker/Dataworker.ts
bmzig: I don't believe we need a block since the ALT only uses this as a seed when deriving its address. And since it's _slightly_ more efficient to just keep the provider call as is, I didn't change it.

PR #2347, opened 2025-07-03
feat: execute fills on SVM chains

grasphoper:
This PR implements filling of Deposits on Solana.

We introduce `SvmFillerClient` to support the current flow in `Relayer.ts` with minimal effort:
1. queue txs one by one
2. execute a tx queue

`SvmFillerClient` stores an SVM signer and a transaction queue


src/relayer/Relayer.ts
```diff
@@ -1096,15 +1130,15 @@ export class Relayer {
         ? [
             "fillRelay",
             "",
-            [convertRelayDataParamsToBytes32(deposit), repaymentChainId, this.relayerAddress.toBytes32()],
+            [convertRelayDataParamsToBytes32(deposit), repaymentChainId, this.relayerEvmAddress.toBytes32()],
```
bmzig, line 1133: You'll probably need to change this and line 1141 to `this.getRelayerAddrOn(repaymentChainId)` too since it's possible that origin is solana, which means repayment could be solana.

src/clients/SvmFillerClient.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.queuedFills = [];
+
+    if (simulate) {
+      await this.simulateQueue(queue);
+      return [];
+    }
+
+    // @dev Execute transactions sequentially, returning signatures of successful ones.
+    const signatures: string[] = [];
+    for (const { txPromise, message, mrkdwn } of queue) {
```
bmzig, line 98: Not sure if I am mistaken here, but it doesn't look like we check whether the transaction will succeed before executing it onchain? I'd understand if that's a follow up, though, since I don't imagine Solana will be that competitive off the bat. I'm pretty sure the main utility for doing that is to prevent us from sending transactions which are going to fail since the relay has already been filled.

grasphoper: I saw some `preflightCheck` or something somewhere in solana client configs. Will investigate

src/relayer/Relayer.ts
grasphoper: Related https://github.com/across-protocol/relayer/pull/2347#discussion_r2185860260

src/relayer/Relayer.ts
```diff
@@ -881,22 +881,24 @@ export class Relayer {
...
       return [];
     }
-    pendingTxnReceipts[chainId] = multiCallerClient.executeTxnQueue(chainId, simulate);
-    const txnReceipts = await pendingTxnReceipts[chainId];
-    delete pendingTxnReceipts[chainId];
+    pendingTxnHashes[chainId] = (async () => {
+      return (await fillExecutorClient.executeTxnQueue(chainId, simulate)).map((response) => response.hash);
+    })();
+    const txnReceipts = await pendingTxnHashes[chainId];
+    delete pendingTxnHashes[chainId];
```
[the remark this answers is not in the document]
pxrl: `Relayer.executeFills()` is scheduled but is not awaited on, so it executes completely decoupled from the main relayer flow. `pendingTxnHashes` is therefore a form of synchronisation; the relayer can check it to know whether it has any pending transaction submissions for a given chain, in which case it won't attempt to enqueue any more.

This means the relayer can effectively schedule transaction submission without having to wait for it to complete, so it can proceed to make fills on other chains in the mean time. This is most useful in cases where a transaction submission timeout would otherwise cause the bot to wait longer than desired.

src/relayer/Relayer.ts
```diff
@@ -349,7 +349,7 @@ export class Relayer {
 
     // Skip deposits that contain invalid fills from the same relayer. This prevents potential corrupted data from
     // making the same relayer fill a deposit multiple times.
-    if (!acceptInvalidFills && invalidFills.some((fill) => fill.relayer.eq(this.relayerAddress))) {
+    if (!acceptInvalidFills && invalidFills.some((fill) => fill.relayer.eq(this.relayerEvmAddress))) {
```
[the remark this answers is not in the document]
pxrl: Yeah; we will need to generalise the way we resolve an address given the relevant chainId.

src/clients/SvmFillerClient.ts
```diff
@@ -0,0 +1,208 @@
...
+      });
+    }
+  }
+
+  getTxnQueueLen(): number {
+    return this.queuedFills.length;
+  }
+
+  getRelayerAddr(): SvmAddress {
+    return SvmAddress.from(this.signer.address, "base58");
```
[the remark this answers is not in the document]
grasphoper: https://github.com/across-protocol/relayer/pull/2347/commits/bf9854831c2c7cd61651606d48f853e487161181

src/clients/SvmFillerClient.ts
md0x: Yep, setting preflightCheck to true tells the node to run the simulation on its side. And if you’re calling [getGasCosts](https://github.com/across-protocol/sdk/blob/960a1997dae7c49b022409f45e1a411060472070/src/relayFeeCalculator/chain-queries/svmQuery.ts#L55), it already runs a simulation internally through Solana’s `computeUnitEstimator`, which is just a simulation wrapper.

bmzig (approved):

PR #2359, opened 2025-07-10
feat:Proposer handover logic

fusmanii:


src/dataworker/Dataworker.ts
```diff
@@ -1590,16 +1590,6 @@ export class Dataworker {
       return leafCount;
     }
 
-    // Exit early if challenge period timestamp has not passed:
-    if (this.clients.hubPoolClient.currentTime <= pendingRootBundle.challengePeriodEndTimestamp) {
```
bmzig, line 1594: I think you'll also need to revert this change and change the conditional to also require `this.config.awaitChallengePeriod`. i.e. `if (this.clients.hubPoolClient.currentTime <= pendingRootBundle.challengePeriodEndTimestamp && !this.config.awaitChallengePeriod)`

bmzig (approved):

PR #2360, opened 2025-07-10
improve: Remove Amount for Unfilled Deposits

dijanin-brat:
Remove Amount for Unfilled Depostis.


src/monitor/Monitor.ts
```diff
@@ -203,29 +203,21 @@ export class Monitor {
...
-            return deposit;
+        const deposits = getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient).map(({ deposit }) => {
+          // Ignore depositId >= bnUInt32Max; these tend to be pre-fills that are eventually valid and
+          // tend to confuse this reporting because there are multiple deposits with the same depositId.
+          if (deposit.depositId < bnUint32Max) {
+            this.logger.warn({
+              at: "SpokePoolClient",
+              chainId: destinationChainId,
+              message: `Unfilled deposit found matching ${getNetworkName(deposit.originChainId)} deposit.`,
+              deposit,
```
pxrl, line 214: Minor annoyance here - because `deposit` contains `Address` types, this will produce a big dump of JSON objects in the log. @bmzig has a utility function that converts all the objects to strings - see `convertRelayDataParamsToBytes32` here in this commit: 91f642e9aeaebf9fcd05a853c2df5ec8530644ad

src/monitor/Monitor.ts
```diff
@@ -203,29 +203,21 @@ export class Monitor {
...
-                invalidFills,
-                notificationPath: "across-invalid-fills",
-              });
-            }
-
-            return deposit;
+        const deposits = getUnfilledDeposits(destinationChainId, spokePoolClients, hubPoolClient).map(({ deposit }) => {
+          // Ignore depositId >= bnUInt32Max; these tend to be pre-fills that are eventually valid and
+          // tend to confuse this reporting because there are multiple deposits with the same depositId.
+          if (deposit.depositId < bnUint32Max) {
```
pxrl, line 209: I might be reading this wrong, but I'm a bit confused on how this is working. `getUnfilledDeposits` returns an array of objects of `{ deposit, version, invalidFills }`. So isn't `depsit.depositId` non-existent?

This change also drops invalid fills logging, and logs unfilled deposits at loglevel warn. In the monitor we'd like to retain invalid fill logging (at loglevel warn).

src/utils/FillUtils.ts
```diff
@@ -31,14 +31,13 @@ export function getUnfilledDeposits(
...
-    .map((deposit) => {
-      const { unfilledAmount, invalidFills } = destinationClient.getValidUnfilledAmountForDeposit(deposit);
-      return { deposit, unfilledAmount, invalidFills };
+    .filter((deposit) => {
+      return !destinationClient.isDepositFilled(deposit);
     })
-    .filter(({ unfilledAmount }) => unfilledAmount.gt(bnZero))
-    .map(({ deposit, ...rest }) => {
+    .map((deposit) => {
+      const invalidFills = destinationClient.getFillsForDeposit(deposit) ?? [];
```
pxrl, line 38: Does `getFillsForDeposit()` not already return an empty array if there are no invalid fillls? If not, we should probably add that.

dijanin-brat: it can return undefined

`public getFillsForDeposit(deposit: Deposit): FillWithBlock[] {
    return this.depositHashesToFills[this.getDepositHash(deposit)];
  }`
  
I will create a PR in SDK for that.

src/monitor/Monitor.ts
dijanin-brat: [14a690c](https://github.com/across-protocol/relayer/pull/2360/commits/14a690c61840010978478811a289d4bb6ba07b1f)

PR #2362, opened 2025-07-10
refactor Permit early exit in HubPool proposer/executor

pxrl:
There are cases where the proposer & executor are unable to do anthing. This can be detected well in advance of doing a HubPoolClient update, which is otherwise a quite heavy procedure.

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -1,9 +1,12 @@
+import { Contract } from "ethers";
 import { TokenClient } from "../clients";
 import {
   EvmAddress,
   SvmAddress,
   winston,
   config,
+  getDeployedContract,
+  getProvider,
   startupLogLevel,
   Signer,
   disconnectRedisClients,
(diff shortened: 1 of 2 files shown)


src/dataworker/index.ts
```diff
@@ -57,15 +61,55 @@ export async function createDataworker(
...
   });
   logger = _logger;
 
-  const { clients, config, dataworker } = await profiler.measureAsync(
-    createDataworker(logger, baseSigner),
+  const config = new DataworkerConfig(process.env);
+  const challengeRemaining = await getChallengeRemaining(config.hubPoolChainId);
+
+  const personality = resolvePersonality(config);
+  if (challengeRemaining > 600 && (config.proposerEnabled || config.l1ExecutorEnabled)) {
```
pxrl, line 106: todo: challengeRemaining threshold to be configurable or inferred based on something else.

bmzig (approved):

PR #2366, opened 2025-07-14
improve(proposer): Guard early exit in proposer/executor

pxrl:
tldr this restores the existing proposer & executor behaviour (when run 
locally) from before the recent change to bail early.

RUN_IDENTIFIER is typically only defined in production and it has 
previously been used in the relayer to infer whether the bot is running
in a production environment.

Also, irrespective of environment, the proposer should run if both the 
propser is enabled and FORCE_PROPOSE is set.

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -87,6 +87,8 @@ async function getChallengeRemaining(chainId: number): Promise<number> {
 }
 
 export async function runDataworker(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
+  const { RUN_IDENTIFIER: runIdentifier } = process.env;
+
   const profiler = new Profiler({
     at: "Dataworker#index",
     logger: _logger,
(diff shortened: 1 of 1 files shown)


src/dataworker/index.ts
```diff
@@ -87,6 +87,8 @@ async function getChallengeRemaining(chainId: number): Promise<number> {
 }
 
 export async function runDataworker(_logger: winston.Logger, baseSigner: Signer): Promise<void> {
+  const { RUN_IDENTIFIER: runIdentifier } = process.env;
```
nicholaspai, line 90: > RUN_IDENTIFIER is typically only defined in production and it has
previously been used in the relayer to infer whether the bot is running
in a production environment.

Is this true? Where else is it?

I wonder if we should set an IS_PRODUCTION flag in the Config.ts file so we can be more clear about this usage, this feels like future tech debt

nicholaspai (commented):
one q

src/dataworker/index.ts
pxrl: `RUN_IDENTIFIER` is set by the hub, so it's only defined in the bot if you run the serverless infra, or if you manually define it. We've used it in the relayer for the past ~12 months without any issues so I'm not really worried about it being tech debt here.

src/dataworker/index.ts
```diff
@@ -97,7 +99,11 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
   const personality = resolvePersonality(config);
   const challengeRemaining = await getChallengeRemaining(config.hubPoolChainId);
 
-  if (challengeRemaining > config.minChallengeLeadTime && (config.proposerEnabled || config.l1ExecutorEnabled)) {
+  if (
+    !isDefined(runIdentifier) &&
```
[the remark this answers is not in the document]
pxrl: Yeah; you are right. Good catch. Fixed: https://github.com/across-protocol/relayer/pull/2366/commits/c9d0579b93f58266bbd626d24974c154df27b1f1

src/dataworker/index.ts
pxrl: Updated after merge: https://github.com/across-protocol/relayer/pull/2366/commits/c9d0579b93f58266bbd626d24974c154df27b1f1 :+1:

PR #2367, opened 2025-07-14
feat: optimistically enqueue root bundle proposals

bmzig:
This should let us ready the next root bundle's proposal during an outstanding proposal's liveness period, which should then let us propose immediately after the pending proposal clears liveness.


src/dataworker/Dataworker.ts
```diff
@@ -354,11 +354,12 @@ export class Dataworker {
     // Construct a list of ending block ranges for each chain that we want to include
     // relay events for. The ending block numbers for these ranges will be added to a "bundleEvaluationBlockNumbers"
     // list, and the order of chain ID's is hardcoded in the ConfigStore client.
-    const nextBundleMainnetStartBlock = this.getNextHubChainBundleStartBlock();
+    const nextBundleMainnetStartBlock = this.getOptimisticChainBundleStartBlock();
```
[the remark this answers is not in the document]
nicholaspai: the other two should only be called when a bundle is already pending so they're slightly different but I think they could be changed as well.

We might want to change the `getNextHubChainBundleStartBlock` function to just call `getOptimisticChainBundleStartBlock` instead of changing this line here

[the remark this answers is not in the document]
nicholaspai: Yeah on second thought you're right. I was thinking when there was a pending bundle that these start blocks would be the same

nicholaspai (dismissed):

src/dataworker/Dataworker.ts
```diff
@@ -334,7 +334,7 @@ export class Dataworker {
     if (!hubPoolClient.isUpdated) {
       throw new Error("HubPoolClient not updated");
     }
-    if (!this.forceProposal && hubPoolClient.hasPendingProposal()) {
+    if (!this.forceProposal && hubPoolClient.hasPendingProposal() && !this.config.awaitChallengePeriod) {
```
nicholaspai, line 337: you might want to move the config check before the hubPoolClient call but it doesn't really matter

src/dataworker/index.ts
```diff
@@ -31,6 +31,7 @@ config();
 let logger: winston.Logger;
 
 const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier = "across-dataworker" } = process.env;
+const MAINNET_BLOCK_TIME = 12;
```
nicholaspai, line 34: Any reason not to call `averageBlockTimeSeconds(ChainIds.MAINNET)`?

src/dataworker/index.ts
```diff
@@ -101,7 +111,11 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
   const personality = resolvePersonality(config);
   const challengeRemaining = await getChallengeRemaining(config.hubPoolChainId);
 
-  if (challengeRemaining > config.minChallengeLeadTime && (config.proposerEnabled || config.l1ExecutorEnabled)) {
+  if (
+    isDefined(runIdentifier) &&
```
nicholaspai, line 115: Can you add a comment why we check runIdentifier? Its basically an isProd check but its not clear and we might get confused in the future

src/dataworker/index.ts
```diff
@@ -261,7 +275,7 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
       poolRebalanceLeafExecutionCount > 0 &&
       (pendingProposal.unclaimedPoolRebalanceLeafCount !== poolRebalanceLeafExecutionCount ||
         pendingProposal.challengePeriodEndTimestamp > clients.hubPoolClient.currentTime);
-    if (proposalCollision || (executorCollision && !config.awaitChallengePeriod)) {
+    if ((proposalCollision || executorCollision) && !config.awaitChallengePeriod) {
```
nicholaspai, line 278: Are executor and proposer collisions still possible?

src/dataworker/index.ts
```diff
@@ -311,6 +335,33 @@ export async function runDataworker(_logger: winston.Logger, baseSigner: Signer)
 
           updatedChallengeRemaining = await getChallengeRemaining(config.hubPoolChainId);
         }
+        if (config.proposerEnabled) {
+          let canPropose = await canProposeRootBundle(config.hubPoolChainId);
+          while (!canPropose && ++counter < 10) {
```
nicholaspai, line 340: How'd you choose 10? Just curious about considerations for waiting longer or shorter

src/dataworker/Dataworker.ts
bmzig: Done https://github.com/across-protocol/relayer/pull/2367/commits/eb0e581b428059cf77bebcb22db468abde8f5fc5

src/dataworker/index.ts
bmzig: Added a comment here https://github.com/across-protocol/relayer/pull/2367/commits/eb0e581b428059cf77bebcb22db468abde8f5fc5. I agree we might not need runIdentifier.

nicholaspai (approved):

PR #2368, opened 2025-07-14
improve: report cause when fill simulation fails

grasphoper:


src/clients/ProfitClient.ts
```diff
@@ -237,10 +238,28 @@ export class ProfitClient {
       return await this.relayerFeeQueries[deposit.destinationChainId].getGasCosts(deposit, relayer);
     } catch (err) {
       const reason = isEthersError(err) ? err.reason : isError(err) ? err.message : "unknown error";
+      // Attempt to extract an underlying cause, if the error object exposes one. This is
+      // helpful for libraries that populate the standard `cause` property on Error
```
nicholaspai, line 242: I just want to understand better how cause and reason are related. Are they expected to both be populated? Or mutually exclusive? I know it's implementation specific so I'm just wondering

nicholaspai (approved):

src/clients/ProfitClient.ts
grasphoper: @nicholaspai  This PR is helpful for Solana errors as those do put relevant info into `cause`.

`cause` exists on all TS errors that extend `Error`. This is how TS defines it:

```
interface Error {
    cause?: unknown;
}
```

Whereas `EthersError` is this:

```
export type EthersError = Error & {
  code: string;
  reason: string;
  error?: Error;
};
```

`EthersError` seems to put relevant info into `.reason` , which is what we're reporting as `reason` in logger.

Solana-side errors seem to report reason like this:

```
  "reason": "Transaction failed when it was simulated in order to estimate the compute unit consumption. The compute unit estimate provided is for a transaction that failed when simulated and may not be representative of the compute units this transaction would consume if successful. Inspect the `cause` property of this error to learn more"
  ```
  
So both cause and reason are somewhat relevant there

bmzig (approved):

nicholaspai (approved):

PR #2369, opened 2025-07-14
refactor(dataworker): Drop bundle volume computation

pxrl:
This isn't required in the dataworker anymore.


bmzig:
Can you also drop the price client from `src/dataworker/DataworkerClients.ts`?

pxrl:
> Can you also drop the price client from `src/dataworker/DataworkerClients.ts`?

Done, plus a few other fixups: https://github.com/across-protocol/relayer/pull/2369/commits/799b475188f6786df36d4365b4851905882f4203

bmzig (approved):

PR #2370, opened 2025-07-14
refactor(dataworker): Drop redundant txn submission arguments

pxrl:
The Dataworker already has the DataworkerConfig instance, so each method can independently know whether how to set arguments like `submitX`. Removing these arguments helps to clean up the interfaces.


nicholaspai (approved):
Wow very nice!

bmzig (approved):

PR #2372, opened 2025-07-15
fix: Converted binance withdrawal to run in series

fusmanii:
Also update the order of the loops from relayer -> token -> network to token -> relayer -> network which allows us to run the outer loop in parallel

Easier to see the diff without whitespaces: https://github.com/across-protocol/relayer/pull/2372/files?w=1

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -300,6 +300,7 @@ export class InventoryClient {
   }
 
   async getAllBundleRefunds(): Promise<CombinedRefunds[]> {
+    const mark = this.profiler.start("bundleRefunds");
     const refunds: CombinedRefunds[] = [];
     const [pendingRefunds, nextBundleRefunds] = await Promise.all([
       this.bundleDataClient.getPendingRefundsFromValidBundles(),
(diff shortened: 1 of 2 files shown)


src/finalizer/utils/binance.ts
```diff
@@ -192,17 +195,19 @@ export async function binanceFinalizer(
...
         // (since a prior withdrawal's fees were paid for in part from the current withdrawal's balance). In this case, set `amountToFinalize` as `min(amountToFinalize, accountBalance)`.
-        if (amountToFinalize > Number(coin.balance)) {
+        if (amountToFinalize > Number(coinBalance)) {
           logger.debug({
             at: "BinanceFinalizer",
             message: `(${depositNetwork} -> ${withdrawNetwork}) Need to reduce the amount to finalize since hot wallet balance is less than desired withdrawal amount.`,
             amountToFinalize,
-            balance: coin.balance,
+            balance: coinBalance,
           });
```
pxrl, line 204: @fusmanii @bmzig Is this still relevant if the loop is run sequentially?

Is there additionally some finality mismatch, where we might be counting deposits on the origin chain in advance of them being finalised by binance?

bmzig (dismissed):

src/finalizer/utils/binance.ts
pxrl: Yep, makes sense - I had thought we were scraping deposits from origin chain transfer events, but I see they are sourced via the binanace API.

src/finalizer/utils/binance.ts
```diff
@@ -125,6 +125,8 @@ export async function binanceFinalizer(
   });
   const binanceDeposits = _binanceDeposits.filter((deposit) => deposit.status === Status.Confirmed);
 
+  // We can run this in parallel since in practice each relayer has different balance of tokens.
+  // If in the future we want to have relayer relay the same tokens then this logic needs to change to run in series.
```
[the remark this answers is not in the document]
pxrl: Ah, understood.

The situation with the two relayers having non-overlapping tokens is an implementation choice on our side but I'm not sure we should bake that into the implementation of the relayer itself. Is there much of a performance gain to be had by running in parallel?

bmzig (approved):

PR #2377, opened 2025-07-16
feat(Monitor): Add function to report on invalid fills related to SVM

nicholaspai:
Temporary additional monitoring while we roll out SVM support


bmzig:
Do you need to add `reportInvalidFillsRelatedToSvm()` to `src/monitor/Monitor.ts` now?

PR #2383, opened 2025-07-18
improve: allow l1-executor submission time to be delayed

bmzig:
This is to avoid simulation false negatives in the l1 executor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants": "^3.1.68",
     "@across-protocol/contracts": "^4.1.0",
-    "@across-protocol/sdk": "4.3.15",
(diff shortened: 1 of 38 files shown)


nicholaspai (dismissed):

pxrl (commented):
@bmzig In a previous variant of this change we waited for 1 block duration (i.e. 12 seconds) after the challenge period.

My concern with suppressing simulation failures on execution is that it seems much more likely to suppress alerts about actual execution issues (i.e. bridge/adapter issues, or liquidity issues). IMO it's more important to be reactive to those than to squeeze out 1 few seconds between the challenge period expiry and the mainnet execution.

wdyt?

pxrl (approved):

PR #2390, opened 2025-07-22
feat: Dynamically fetch SVM token programId

dijanin-brat:
Dynamically fetch SVM token programId

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -94,7 +94,7 @@ import {
   fetchEncodedAccount,
   type KeyPairSigner,
 } from "@solana/kit";
-import { TOKEN_PROGRAM_ADDRESS, getCreateAssociatedTokenIdempotentInstruction } from "@solana-program/token";
+import { fetchMint, getCreateAssociatedTokenIdempotentInstruction } from "@solana-program/token";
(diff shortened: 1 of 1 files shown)


md0x (approved):
LGTM!

src/dataworker/Dataworker.ts
```diff
@@ -3029,13 +3029,16 @@ export class Dataworker {
     let recipientCreateTokenAccountInstruction;
     const associatedTokenAccountExists = (await fetchEncodedAccount(provider, recipientTokenAccount)).exists;
     if (!associatedTokenAccountExists) {
+      const mint = arch.svm.toAddress(leaf.relayData.outputToken);
+      const mintInfo = await fetchMint(spokePoolClient.svmEventsClient.getRpc(), mint);
+      const programAddress = mintInfo.programAddress;
```
pxrl, line 3034: ooc, where else in the code do we do this? I wonder if an SDK utility under src/arch/svm might make sense.

dijanin-brat: we are doing similar things in SDK `src/arch/svm/SpokeUtils.ts`
But, i don't see any major benefit of having utility function that will just call `fetchMint(..)`

PR #2392, opened 2025-07-22
chore: bump sdk

bmzig:
This is still pending an sdk bump which includes this PR: https://github.com/across-protocol/sdk/pull/1142


pxrl:
@bmzig Should be OK to close this now right?

bmzig:
> @bmzig Should be OK to close this now right?

This bump was originally meant for this PR https://github.com/across-protocol/sdk/pull/1142, but that might be delayed, so it's fine to close and re-open in the future.

PR #2407, opened 2025-07-24
fix(OpStackFinalizer): Use viem fix to support new OptimismPortal2 wi…

nicholaspai:
…thdrawals

Patch file only exists until https://github.com/wevm/viem/pull/3824 is merged and released as NPM package

### DIFF
--- a/src/finalizer/utils/opStack.ts
+++ b/src/finalizer/utils/opStack.ts
@@ -391,30 +391,12 @@ async function viem_multicallOptimismFinalizations(
       hash: event.txnRef as `0x${string}`,
     });
     const withdrawal = getWithdrawals(receipt)[logIndexesForMessage[i]];
-    let withdrawalStatus: GetWithdrawalStatusReturnType;
-    try {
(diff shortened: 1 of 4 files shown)


src/finalizer/utils/opStack.ts
```diff
@@ -391,30 +391,12 @@ async function viem_multicallOptimismFinalizations(
...
-        // 0xd9bc01be = bytes4(keccak256("OptimismPortal_ProofNotOldEnough()"))
-        withdrawalStatus = "waiting-to-finalize";
-      } else if ((error as BaseError).shortMessage.includes("0xcca6afda")) {
-        // 0xcca6afda = bytes4(keccak256("OptimismPortal_Unproven()"))
-        withdrawalStatus = "waiting-to-prove";
-      } else {
-        throw error;
-      }
-    }
+    const withdrawalStatus = await getWithdrawalStatus(publicClientL1 as viem.Client, {
```
nicholaspai, line 394: This should revert this code back to before 47bf9553ac25cbfed63dc65efed2d124299d62ac

pxrl (approved):

PR #2416, opened 2025-07-27
feat(TransactionUtils): Do not send `error` level log in `runTransaction()`

nicholaspai:
## Motivation
The motivation for this PR is seeing `error` level logs emitted by `runTransaction` caused by seemingly colliding transactions due to interleaving bot runs. For example, if the executor attempts to execute a leaf at the same time across two different runs, they will both pass simulation in the `MulticallerClient` but due to the latency between simulation transactions and ultimately submitting them, only one of the transactions will succeed while the other one will revert.

## Implementation
This function currently sends an error level log when a transaction submission fails. However, in practice this is troublesme for a few reasons specific to how the repository typically uses this function:
- When the MulticallerClient calls runTransaction, it first simulates the transaction calldata before submitting. If the simulation fails, then the MulticallerClient will send an error level log. So if the transaction calldata passes simulation but then fails on submission, it indicates more than likely that the transaction has failed for a non-persistent reason, such as colliding with a duplicate calldata submitted by a parallel bot run. Therefore in these cases we do not want `runTransaction` to send an error level log about a colliding transaction.
- In other cases where `runTransaction` is called directly without first simulating the calldata, then the code will crash if the transaction submission fails, and an error level log will be sent regardless. Therefore in these cases, sending an error level log from within `runTransaction` before `throw`-ing just creates extra noise.

## Additional QoL improvements

Adds a warning to the `TransactionClient`'s `submit()` function if the user didn't call `simulate()` first before `submitting` because the `TransactionClient` catches any errors thrown by `runTransaction`. This is currently not an issue with all uses of the `submit()` function in this repo.

### DIFF
--- a/src/clients/TransactionClient.ts
+++ b/src/clients/TransactionClient.ts
@@ -42,8 +42,13 @@ const DEFAULT_GASLIMIT_MULTIPLIER = 1.0;
 export class TransactionClient {
   readonly nonces: { [chainId: number]: number } = {};
 
+  /**
+   * @param logger - The logger to use for logging.
+   * @param logErrorOnFailure - Whether to log errors on failure as `error` level logs. Set this to false if the
+   * caller plans to simulate all transactions and handle the errors themselves before submitting.
+   */
   // eslint-disable-next-line no-useless-constructor
-  constructor(readonly logger: winston.Logger) {}
+  constructor(readonly logger: winston.Logger, readonly logErrorOnFailure = true) {}
 
   protected _simulate(txn: AugmentedTransaction): Promise<TransactionSimulationResult> {
     return willSucceed(txn);
@@ -57,7 +62,7 @@ export class TransactionClient {
 
   protected _submit(txn: AugmentedTransaction, nonce: number | null = null): Promise<TransactionResponse> {
     const { contract, method, args, value, gasLimit } = txn;
-    return runTransaction(this.logger, contract, method, args, value, gasLimit, nonce);
+    return runTransaction(this.logger, contract, method, args, value, gasLimit, nonce, this.logErrorOnFailure);
   }
 
   async submit(chainId: number, txns: AugmentedTransaction[]): Promise<TransactionResponse[]> {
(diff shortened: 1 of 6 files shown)


src/clients/TransactionClient.ts
```diff
@@ -60,7 +68,27 @@ export class TransactionClient {
...
+   * unexpectedly.
+   * @param chainId - The chain ID to submit the transactions to.
+   * @param txns - The transactions to submit.
+   * @returns The transaction responses.
+   */
   async submit(chainId: number, txns: AugmentedTransaction[]): Promise<TransactionResponse[]> {
+    if (!this._simulatedBeforeSubmit) {
+      // Force user to call simulateFirst() first and handle the error, otherwise this function might crash silently
+      // because any errors thrown by _submit() are caught.
+      this.logger.warn({
```
bmzig, line 85: Why do we not get a race here when using the multi caller client. https://github.com/across-protocol/relayer/blob/fc1c174fc2cb6cf5ee25ed0d57d03e3a42897047/src/clients/MultiCallerClient.ts#L136. E.g. we need to submit txns on 1, 10. Both simulate before submitting, but as soon as one of them hits `submit` (say the txn for 1), then `_simulateBeforeSubmit` is now false, so the `submit` for 10 now emits a warn.

bmzig (dismissed):

pxrl (approved):
@nick Just to be sure, will we continue to alert in cases where (for example) the mainnet executor fails to execute, or the disputer fails to submit a dispute? I think we just want to be extra confident about that before merging this.

> When the MulticallerClient calls runTransaction, it first simulates the transaction calldata before submitting. If the simulation fails, then the MulticallerClient will send an error level log. So if the transaction calldata passes simulation but then fails on submission, it indicates more than likely that the transaction has failed for a non-persistent reason, such as colliding with a duplicate calldata submitted by a parallel bot run. Therefore in these cases we do not want runTransaction to send an error level log about a colliding transaction.

One comment to this - there are places where `canFailInSimulation` causes the `TransactionClient` to skip simulation, in which case a failure on submission is the only place we'd actually report on the failure.

This change will definitely help to reduce a lot of noise from the bots. I am mindful though, that a lot of this noise is symptomatic of things that should be fixed elsewhere, so if we turn the volume down we might reduce our awareness of some of the underlying issues.

PR #2436, opened 2025-07-30
fix: properly initialize headers object in getProviderHeaders

reallesee:
Squashed a sneaky bug in getProviderHeaders where headers could be undefined when accessed. Now properly initializing as empty object and returning undefined when no headers exist. This prevents potential "Cannot set property of undefined" runtime errors that would be hard to debug in production. Small change, big impact for reliability


pxrl:
Thanks for this report.

> This prevents potential "Cannot set property of undefined" runtime errors that would be hard to debug in production.

Can you clarify exactly how this change is materially different from the existing implementation?

PR #2450, opened 2025-08-06
improve: Filter deposits by both origin and destination SpokePoolClients

dijanin-brat:
Filter deposits by both origin and destination SpokePoolClients

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -432,18 +432,28 @@ export class Relayer {
    */
   private _getUnfilledDeposits(): Record<number, RelayerUnfilledDeposit[]> {
     const { hubPoolClient, spokePoolClients } = this.clients;
-    const { relayerDestinationChains } = this.config;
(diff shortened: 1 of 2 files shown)


src/monitor/Monitor.ts
```diff
@@ -195,29 +195,32 @@ export class Monitor {
...
-                notificationPath: "across-invalid-fills",
-              });
-            }
-
-            return deposit;
+        const deposits = getUnfilledDeposits(
+          spokePoolClients,
+          hubPoolClient,
+          {},
+          spokePoolClients[destinationChainId]
```
pxrl, line 202: If we flip the order of `FillStatus` and `destionationChainId` then we don't need to supply the empty `FillStatus` object, because it defaults to `{}` :+1:

dijanin-brat: [f20d394](https://github.com/across-protocol/relayer/pull/2450/commits/f20d39485a2be1f4b6cf74f07a89cdc87ab5e2aa)

pxrl (approved):

PR #2454, opened 2025-08-07
feat: Prepare bundleRefundsPromise on init of InventoryClient

dijanin-brat:
Prepare bundleRefundsPromise on init of InventoryClient. This will speed up the relayers bcs, during the relayer handover, it will load bundleRefunds into a relayer so the relayer doesn't need to wait for it when it wants to fill the deposit.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -332,6 +332,12 @@ export class InventoryClient {
     return refunds;
   }
 
+  public prepareBundleRefundsPromise(): void {
+    if (!isDefined(this.bundleRefundsPromise)) {
+      this.bundleRefundsPromise = this.getAllBundleRefunds();
+    }
+  }
+
   // Return the upcoming refunds (in pending and next bundles) on each chain.
   // @notice Returns refunds using decimals of the l1Token.
   private async getBundleRefunds(l1Token: EvmAddress): Promise<{ [chainId: string]: BigNumber }> {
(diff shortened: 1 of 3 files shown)


src/clients/InventoryClient.ts
```diff
@@ -342,11 +353,7 @@ export class InventoryClient {
     // Increase virtual balance by pending relayer refunds from the latest valid bundle and the
     // upcoming bundle. We can assume that all refunds from the second latest valid bundle have already
     // been executed.
-    if (!isDefined(this.bundleRefundsPromise)) {
-      // @dev Save this as a promise so that other parallel calls to this function don't make the same call.
-      mark = this.profiler.start(`bundleRefunds for ${l1Token}`);
-      this.bundleRefundsPromise = this.getAllBundleRefunds();
-    }
+    this.prepareBundleRefundsPromise();
     refundsToConsider = lodash.cloneDeep(await this.bundleRefundsPromise);
```
pxrl, line 357:
```suggestion
    refundsToConsider = lodash.cloneDeep(await this.executeBundleRefundsPromise());
```

src/relayer/index.ts
```diff
@@ -84,6 +84,8 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
           .map(({ chainId }) => getNetworkName(chainId));
         throw new Error(`Unable to start relayer due to chains ${badChains.join(", ")}`);
       }
+      // Execute bundleRefundsPromise only after all spokePoolClients are updated.
+      await relayer.clients.inventoryClient.executeBundleRefundsPromise();
```
pxrl, line 88: We probably want to guard this call so we only do it under the following circumstances:
- inventoryClient.isInventoryManagementEnabled()
- One-time init (i.e. a boolean flag we can set on init, so we don't keep calling it on subsequent loops.

So something like:

```
// Out of main relayer loop:
const { inventoryClient } = relayer.clients;
const inventoryManagement = inventoryClient.isInventoryManagementEnabled();
let inventoryInit = !inventoryManagement;

// In loop:
if (!inventoryInit && inventoryManagement) {
  await inventoryClient.executeBundleRefundsPromise();
  inventoryInit = true;
}
```

wdyt?

src/clients/InventoryClient.ts
```diff
@@ -342,12 +353,7 @@ export class InventoryClient {
     // Increase virtual balance by pending relayer refunds from the latest valid bundle and the
     // upcoming bundle. We can assume that all refunds from the second latest valid bundle have already
     // been executed.
-    if (!isDefined(this.bundleRefundsPromise)) {
-      // @dev Save this as a promise so that other parallel calls to this function don't make the same call.
-      mark = this.profiler.start(`bundleRefunds for ${l1Token}`);
```
pxrl, line 347: We probably want to move this to `executeBundleRefundsPromise()` and make it `async` so we can await on it in order to capture the execution time.

dijanin-brat: [b91632b](https://github.com/across-protocol/relayer/pull/2454/commits/b91632b612b6d3d8bb87b7b19ec2fcd005c9c467)

pxrl (approved):

PR #2463, opened 2025-08-08
feat: add support for bob sepolia

abg4:
- Updated @across-protocol/constants to version 3.1.71
- Updated @across-protocol/contracts to version 4.1.1
- Added support for BOB_SEPOLIA in various constants and configurations

### DIFF
--- a/src/clients/ProfitClient.ts
+++ b/src/clients/ProfitClient.ts
@@ -668,6 +668,7 @@ export class ProfitClient {
       [CHAIN_IDs.WORLD_CHAIN]: "WETH", // USDC deferred on World Chain.
       [CHAIN_IDs.LENS_SEPOLIA]: "WETH", // No USD token on Lens Sepolia
       [CHAIN_IDs.TATARA]: "WETH",
+      [CHAIN_IDs.BOB_SEPOLIA]: "WETH",
     };
     const prodRelayer = process.env.RELAYER_FILL_SIMULATION_ADDRESS ?? PROD_RELAYER;
     const [defaultTestSymbol, _relayer] =
--- a/src/common/ContractAddresses.ts
+++ b/src/common/ContractAddresses.ts
@@ -711,4 +711,10 @@ export const CONTRACT_ADDRESSES: {
       abi: OVM_L2_STANDARD_BRIDGE_ABI,
     },
   },
+  [CHAIN_IDs.BOB_SEPOLIA]: {
+    ovmStandardBridge: {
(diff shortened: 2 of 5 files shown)


src/clients/ProfitClient.ts
```diff
@@ -668,6 +668,7 @@ export class ProfitClient {
       [CHAIN_IDs.WORLD_CHAIN]: "WETH", // USDC deferred on World Chain.
       [CHAIN_IDs.LENS_SEPOLIA]: "WETH", // No USD token on Lens Sepolia
       [CHAIN_IDs.TATARA]: "WETH",
+      [CHAIN_IDs.BOB_SEPOLIA]: "WETH",
```
pxrl, line 671: It should be fine to drop this because we handle testnet down at line 675:

```
const [defaultTestSymbol, _relayer] =
      this.hubPoolClient.chainId === CHAIN_IDs.MAINNET ? ["USDC", prodRelayer] : ["WETH", TEST_RELAYER];
```

pxrl (dismissed):

src/common/ContractAddresses.ts
bmzig, line 1: Somewhere in the `[CHAIN_IDs.SEPOLIA]` (line 527-602) you need to also add the ovmStandardBridge_808813: { address: 0x..., abi: OVM_L1_STANDARD_BRIDGE_ABI, }` since although the relayer may not rebalance, something needs to be defined there for the relayer to start. So to be clear, the address can be some garbage address, but it needs to be defined.

src/clients/ProfitClient.ts
```diff
@@ -674,6 +674,7 @@ export class ProfitClient {
...
```
pxrl, line 677:
```suggestion
```

src/clients/ProfitClient.ts
```diff
@@ -672,8 +672,6 @@ export class ProfitClient {
       [CHAIN_IDs.REDSTONE]: "WETH", // Redstone only supports WETH.
       [CHAIN_IDs.SONEIUM]: "WETH", // USDC deferred on Soneium.
       [CHAIN_IDs.WORLD_CHAIN]: "WETH", // USDC deferred on World Chain.
-      [CHAIN_IDs.LENS_SEPOLIA]: "WETH", // No USD token on Lens Sepolia
-      [CHAIN_IDs.TATARA]: "WETH",
```
pxrl, line 676: Dropping these because they're not required.

bmzig (approved):

PR #2476, opened 2025-08-12
improve: Bump SDK

nicholaspai:


bmzig (approved):

src/libexec/RelayerSpokePoolListenerSVM.ts
```diff
@@ -83,7 +83,7 @@ async function scrapeEvents(
 ): Promise<void> {
   const provider = eventsClient.getRpc();
   const [{ timestamp: currentTime }, ...events] = await Promise.all([
-    arch.svm.getNearestSlotTime(provider, logger),
+    arch.svm.getNearestSlotTime(provider, { commitment: "confirmed" }, logger),
```
pxrl, line 86: @nicholaspai Do we need to be explicit about the commitment here? Up in the SDK I deliberately initialised `opts` to `undefined`, so we inherit the defaults. Otherwise we have a higher chance of some calls tagging `confirmed` and others `finalized`, `processed`, etc.

PR #2485, opened 2025-08-14
improve: add `STYLE.md`

grasphoper:


STYLE.md
```diff
@@ -0,0 +1,4 @@
+This repo adpots [Google TypeScript Style Guide](https://google.github.io/styleguide/tsguide.html) for the sake of consistency and clarity of code.
+
+In addition to the style guide above, the following principles should be followed:
+- write good code
```
grasphoper, line 4: I had the following ideas for the "principles to be followed" part:

- functions should be short, < 100 (50) LOC
- functions should rarely have more than a couple of indentations within them, e.g. max 3. If there's more, it's usually a sign that functionality can be extracted to a separate function with readable name
- within the same file / class, there should be minimal redundancy

Please share your ideas so that we can compose our initial "principles" version

pxrl: One additional input for consideration: comments should not exceed the maximum configured line length (120 chars). This unfortunately seems to require manual wrapping; prettier apparently doesn't want to reflow comments.

grasphoper: It seems we mostly agree on the 3 points [above](https://github.com/across-protocol/relayer/pull/2485#discussion_r2275009817).

@bmzig taking into account what Paul mention about should / shall, are you OK with including 1. as well?

> Also, I'd think it would be helpful if we start commenting a lot more on the code. Maybe we mandate that all functions have some documentation at the very least and then any function with length > 100 LOC should have inner comments describing each step of the function.

+1 on commenting complex functions. IMO the point about keeping the functions short aligns with this too. If a complex function is split into functions with readable names (even if the impl. of each of those is complex), it effectively breaks a function down into comprehensible chunks.

It also seems that we _don't_ have an agreement on the indentation / line width. It seems like 2 spaces is the norm in TS world, with 100+ max line length generally.

I'm planning on updating this with the points above pending Bennett's response.

@fusmanii what do you think of all this?

fusmanii: > functions should be short, < 100 (50) LOC

100p on board for this, I think we can do a similar guideline for files as well. IMO around 500 is ideal, def no more than 1000 LOC/file

The other two points I am ok with as well.

I think we should also include rule around not having functions defined inside other functions, it makes code hard to follow and can easily break rule 2.

>A somewhat more opinionated idea: 4 space indents are more readable than 2, esp. with bigger functions / codebases

I am kind of used to 2

grasphoper: > > A somewhat more opinionated idea: 4 space indents are more readable than 2, esp. with bigger functions / codebases
>
> I am kind of used to 2

It seems like it makes sense for us to just stick with 2

> I think we can do a similar guideline for files as well. IMO around 500 is ideal, def no more than 1000 LOC/file

Agree, files should not be long

pxrl: I'd prefer if we put in a guideline that comments should be limited to the prevailing line length (120 chars), though. We have a couple of instances where comments run off the page and it looks bad in git and locally.

pxrl (approved):
One remark to comment line length - I'd like it if they were limited by the prevailing line length (120 chars in current config).

PR #2488, opened 2025-08-14
improve: Bump Linea SDK version

dijanin-brat:
Bump Linea SDK version


src/finalizer/utils/linea/common.ts
```diff
@@ -28,14 +31,24 @@ export type MessageWithStatus = Message & {
...
+export function initLineaSdk(l1ChainId: number, l2ChainId: number, signer?: Signer): LineaSDK {
+  assert(isSignerWallet(signer), "Signer is not a Wallet");
   return new LineaSDK({
     l1RpcUrl: Object.values(getNodeUrlList(l1ChainId))[0],
     l2RpcUrl: Object.values(getNodeUrlList(l2ChainId))[0],
     network: l1ChainId === CHAIN_IDs.MAINNET ? "linea-mainnet" : "linea-goerli",
-    mode: "read-only",
+    mode: "read-write",
+    l1SignerPrivateKey: signer._signingKey().privateKey,
+    l2SignerPrivateKey: signer._signingKey().privateKey,
```
dijanin-brat, line 51: FYI, for new version of linea, we need to use `read-write` mode to claim message. And in that mode, we need to provide `l1SignerPrivateKey` and `l2SignerPrivateKey`.

src/finalizer/utils/linea/common.ts
```diff
@@ -28,14 +31,24 @@ export type MessageWithStatus = Message & {
...
+  logIndex: number;
+  transactionHash: string;
+  blockNumber: number;
+}
+
 export const lineaAdapterIface = Linea_Adapter__factory.createInterface() as ethers.utils.Interface;
 
-export function initLineaSdk(l1ChainId: number, l2ChainId: number): LineaSDK {
+export function initLineaSdk(l1ChainId: number, l2ChainId: number, signer?: Signer): LineaSDK {
+  assert(isSignerWallet(signer), "Signer is not a Wallet");
```
nicholaspai, line 44: what if `signer` is undefined? What will happen here

nicholaspai (approved):
Looks good

src/finalizer/utils/linea/l1ToL2.ts
```diff
@@ -132,17 +130,19 @@ export async function lineaL1ToL2Finalizer(
...
-      return l2MessageServiceContract.contract.populateTransaction.claimMessage(
-        message.messageSender,
-        message.destination,
-        message.fee,
-        message.value,
-        await signer.getAddress(),
-        message.calldata,
-        message.messageNonce
-      );
+      return l2MessageServiceContract.claim({
```
bmzig, line 133: What does this return now? I'm guessing we can't use `populateTransaction` because `l2MessageServiceContract` isn't an ethers contract, but we also expected to return calldata, so does this return calldata?

bmzig (approved):
One non-actionable question.

pxrl (approved):

PR #2492, opened 2025-08-14
fix(relayer): Enforce ignoring unprofitable deposits

pxrl:
A recent change to permit the relayer to rebalance for tokens where it was forced to take origin chain repayment unintentionally changed behaviour such that the relayer no longer ignored unprofitable deposits. Reinstate that behaviour and simplify the code.

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -763,13 +763,22 @@ export class Relayer {
       lpFeePct: realizedLpFeePct,
       gasPrice,
     } = repaymentChainProfitability;
-    if (tokenClient.hasBalanceForFill(deposit)) {
-      if (!isDefined(repaymentChainId)) {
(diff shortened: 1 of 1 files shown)


nicholaspai (dismissed):

src/relayer/Relayer.ts
```diff
@@ -802,20 +811,7 @@ export class Relayer {
...
-      // Exit early if we want to request a slow fill for a lite chain.
-      if (depositForcesOriginChainRepayment(deposit, this.clients.hubPoolClient)) {
-        this.logger.debug({
-          at: "Relayer::evaluateFill",
-          message: "Skipping requesting slow fill for deposit that forces origin chain repayment",
-          originChainId,
-          depositId: depositId.toString(),
-        });
-        return;
-      }
```
pxrl, line 818: This can be dropped here because the same check exists inside the `Relayer.requestSlowFill()` method.

bmzig (dismissed):

PR #2495, opened 2025-08-15
refactor: remove unnecessary Anchor calls

bmzig:
This removes unused functions and swaps out the use of anchor in a specific function in the finalizer.


pxrl (approved):
Big :+1:

PR #2500, opened 2025-08-19
feat(InventoryClient): Take repayment on origin chain if we can quickly rebalance

nicholaspai:
This is an opinionated PR so i'm looking for feedback. It changes the inventory client's behavior to take repayment on the origin chain if the origin chain is CCTP and the input token is USDC. It will only do so if `prioritizeLpUtilization` is set to true.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -516,7 +516,12 @@ export class InventoryClient {
     // If the deposit forces origin chain repayment but the origin chain is one we can easily rebalance inventory from,
     // then don't ignore this deposit based on perceived over-allocation. For example, the hub chain and chains connected
     // to the user's Binance API are easy to move inventory from so we should never skip filling these deposits.
-    if (forceOriginRepayment && repaymentChainCanBeQuicklyRebalanced(deposit, this.hubPoolClient)) {
+    // If the relayer wants to prioritize LP utilization, then we should always take repayment on the origin chain
+    // if it is a quick rebalance source.
+    if (
+      (this.prioritizeLpUtilization || forceOriginRepayment) &&
+      repaymentChainCanBeQuicklyRebalanced(deposit, this.hubPoolClient)
+    ) {
       return [deposit.originChainId];
     }
 
@@ -703,7 +708,7 @@ export class InventoryClient {
 
     // Always add hubChain as a fallback option if inventory management is enabled and origin chain is not a lite chain.
     // If none of the chainsToEvaluate were selected, then this function will return just the hub chain as a fallback option.
-    if (!depositForcesOriginChainRepayment(deposit, this.hubPoolClient) && !eligibleRefundChains.includes(hubChainId)) {
+    if (!forceOriginRepayment && !eligibleRefundChains.includes(hubChainId)) {
       eligibleRefundChains.push(hubChainId);
     }
     return eligibleRefundChains;
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -1,6 +1,7 @@
 import { HubPoolClient, SpokePoolClient } from "../clients";
 import { FillStatus, FillWithBlock, SpokePoolClientsByChain, DepositWithBlock, RelayData } from "../interfaces";
-import { CHAIN_IDs, EMPTY_MESSAGE } from "../utils";
+import { CHAIN_IDs, compareAddressesSimple, EMPTY_MESSAGE, TOKEN_SYMBOLS_MAP } from "../utils";
+import { utils as sdkUtils } from "@across-protocol/sdk";
 
 export type RelayerUnfilledDeposit = {
   deposit: DepositWithBlock;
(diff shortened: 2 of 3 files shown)


src/clients/InventoryClient.ts
```diff
@@ -703,7 +708,7 @@ export class InventoryClient {
 
     // Always add hubChain as a fallback option if inventory management is enabled and origin chain is not a lite chain.
     // If none of the chainsToEvaluate were selected, then this function will return just the hub chain as a fallback option.
-    if (!depositForcesOriginChainRepayment(deposit, this.hubPoolClient) && !eligibleRefundChains.includes(hubChainId)) {
+    if (!forceOriginRepayment && !eligibleRefundChains.includes(hubChainId)) {
```
nicholaspai, line 711: Opportunistic refactor

bmzig (dismissed):

src/clients/InventoryClient.ts
```diff
@@ -469,14 +469,6 @@ export class InventoryClient {
       );
     }
 
-    // If the deposit forces origin chain repayment but the origin chain is one we can easily rebalance inventory from,
-    // then don't ignore this deposit based on perceived over-allocation. For example, the hub chain and chains connected
-    // to the user's Binance API are easy to move inventory from so we should never skip filling these deposits.
-    const forceOriginRepayment = depositForcesOriginChainRepayment(deposit, this.hubPoolClient);
```
nicholaspai, line 475: I modified this logic and movied it down to line 524, after evaluating all slow repayment chains. Based on discussions with @pxrl about how best to reduce LP utilization

pxrl: @nicholaspai I think fwiw we could _also_ leave this here, since if we know in advance that we're forced to take origin repayment then there's no point in doing any further evaluation. But the additional conditional down at line ~530 makes good sense.

src/utils/FillUtils.ts
```diff
@@ -81,10 +82,23 @@ export function depositForcesOriginChainRepayment(
...
 export function repaymentChainCanBeQuicklyRebalanced(
-  deposit: Pick<DepositWithBlock, "originChainId">,
+  repaymentChainId: number,
+  repaymentToken: Address,
   hubPoolClient: HubPoolClient
 ): boolean {
-  return [hubPoolClient.chainId, CHAIN_IDs.BSC].includes(deposit.originChainId);
+  const originChainIsCctpEnabled =
+    compareAddressesSimple(TOKEN_SYMBOLS_MAP.USDC.addresses[repaymentChainId], repaymentToken.toNative()) &&
+    sdkUtils.chainIsCCTPEnabled(repaymentChainId);
```
pxrl, line 91: There's a minor optimisation to flip these two, since it'll be faster to check `chainIsXEnabled()` than to compare all the addresses. Same for the OFT variant.

src/clients/InventoryClient.ts
nicholaspai: Yeah good point, wdyt of the overall algo now? Essentially the PR adds a new condition after the excess running balances check

src/clients/InventoryClient.ts
```diff
@@ -528,6 +531,16 @@ export class InventoryClient {
...
+    // If the relayer wants to prioritize LP utilization, then we should always take repayment on the origin chain
+    // if it is a quick rebalance source.
+    if (
+      this.prioritizeLpUtilization &&
+      // @todo: Consider requiring also that the deposit amount is large enough that its worth taking repayment on
+      // origin chain instead of waiting the ~30 mins for the "fast rebalancing" CCTP/OFT bridge to complete.
+      repaymentChainCanBeQuicklyRebalanced(deposit.originChainId, inputToken, this.hubPoolClient)
+    ) {
+      return [deposit.originChainId];
+    }
```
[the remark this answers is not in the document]
nicholaspai: https://github.com/across-protocol/relayer/pull/2500/commits/4451e93ce405f16bc8ae1f9b162a2690a7007fa2#diff-2329b23d0437737611d7b358678c3459a0b33e19b2fd68dd32b95b72bd2aa776

test/InventoryClient.RefundChain.ts
```diff
@@ -266,171 +232,6 @@ describe("InventoryClient: Refund chain selection", async function () {
...
-      tokenClient.setTokenData(
-        ARBITRUM,
-        toAddressType(l2TokensForWeth[ARBITRUM], ARBITRUM),
-        initialAllocation[ARBITRUM][mainnetWeth].sub(toWei(5))
-      );
-      expect(await inventoryClient.determineRefundChainId(sampleDepositData)).to.deep.equal([ARBITRUM, MAINNET]);
-      expect(lastSpyLogIncludes(spy, 'expectedPostRelayAllocation":"35555555555555555')).to.be.true; // 4.8/(140-5)
-    });
-
-    it("Normalizes shortfalls to correct precision", async function () {
```
nicholaspai, line 338: This test is testing internal logic and I think its not really useful but also very flaky

src/clients/InventoryClient.ts
```diff
@@ -532,8 +532,9 @@ export class InventoryClient {
     // is a lite chain, which should allow the relayer to take more repayments away from the lite chain. Because
     // lite chain deposits force repayment on origin, we end up taking lots of repayment on the lite chain so
     // we should take repayment away from the lite chain where possible.
+    // We can also want to take repayment on the origin chain if it is a quick rebalance source.
```
nicholaspai, line 535:
```suggestion
    // We also want to take repayment on the origin chain if it is a quick rebalance source.
```

PR #2501, opened 2025-08-19
feat: Added RedisPubSub class

fusmanii:
The same redis client can't be used for both commands and pub/sub so need to create a new class so that the two instances don't conflict


dijanin-brat (approved):
LGTM

bmzig (approved):

dijanin-brat (approved):

PR #2508, opened 2025-08-20
fix: Don't re-throw on Linea gas estimation error

fusmanii:
Linea has a special method `linea_estimateGas` that doesn't return a revert reason. In those cases we shouldn't be re-throwing the error because the deposit has mostly likely been already filled


nicholaspai (approved):
I think this is reasonable, curious what @pxrl thinks

pxrl:
> I think this is reasonable, curious what @pxrl thinks

I think we need to look at why the `isFillRelayError()` call isn't firing. I see these are logging as `error`, but that shouldn't be the case:

```
[error] across-v2-fast-relayer (TxUtil#runTransaction)⭢Error executing tx
• target:
- targetAddress: 0x7E63A5f1a8F0B4d0934B2f2327DAED3F6bb2ee75
• method: fillRelay
```

If we have to have Linea-specific handling, then I'd probably prefer to do it based on the combination of chainId and method, rather than hardcoding a path and inspecting the stack trace.

PR #2511, opened 2025-08-21
chore: Remove .toNative() from logger calls

fusmanii:
With https://github.com/across-protocol/sdk/pull/1169 `.toNative()` for addresses in logger is redundant


grasphoper:
@fusmanii have you confirmed that this is the case? I remember hitting some instances of full JSON object properties logging for `Address` sometimes

pxrl:
> @fusmanii have you confirmed that this is the case? (That the .toNative() is not required) I remember hitting some instances of full JSON object properties logging for `Address` sometimes

I've also noticed this - we currently have an issue with an error message dumping the full Address class. I'm not 100% sure, but I _think_ it might be an issue when an Address class is supplied directly as a logging argument, vs. when it's bundled inside another object. Needs further investigation to confirm.

PR #2513, opened 2025-08-22
fix: ignore refunds which do not relate to the monitored relayer

bmzig:
This should make the reporter stop throwing when we need to refund an unknown token.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -1373,10 +1373,10 @@ export class Monitor {
 
       for (const _tokenAddress of Object.keys(fillsToRefund)) {
         const tokenAddress = toAddressType(_tokenAddress, chainId);
-	// If there are no refunds for the monitored relayer for this token, then ignore this token.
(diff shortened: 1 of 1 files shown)


mrice32 (approved):

src/monitor/Monitor.ts
```diff
@@ -1373,6 +1373,10 @@ export class Monitor {
 
       for (const _tokenAddress of Object.keys(fillsToRefund)) {
         const tokenAddress = toAddressType(_tokenAddress, chainId);
+        // If there are no refunds for the monitored relayer for this token, then ignore this token.
+        if (!isDefined(fillsToRefund[tokenAddress.toBytes32()][relayer.toBytes32()])) {
```
nicholaspai, line 1377: This is good. We could arguably ignore any token we can't find in our token symbols map but we can fast follow with that.

bmzig: Yeah that's true. I assumed that if it is a monitored relayer then they will only receive refunds from tokens in TOKEN_SYMBOLS_MAP.

PR #2526, opened 2025-08-26
fix: Added txn method to `submit()` log

fusmanii:


src/utils/TransactionUtils.ts
```diff
@@ -214,6 +215,11 @@ export async function runTransaction(
           notificationPath: isWarning ? "across-warn" : "across-error",
           error: stringifyThrownValue(error),
         });
+
+        // If the error is due to a relay collision and the chain is Linea, then we can ignore it.
+        if (isWarning && chainIsLinea(contract.chainId)) {
```
nicholaspai, line 220: If we get here it doesn't always mean its an `isFillRelayError`, we might need to replace the `isWarning` with `isFillRelayError`

fusmanii: good point, fixed https://github.com/across-protocol/relayer/pull/2526/commits/f6c496aef3e2d6bd2d5c0425c68e3bdee0480fca

src/utils/TransactionUtils.ts
pxrl, line 1: I'm a bit puzzled on this logic; the non-Linea EVM doesn't cause any noise here - do we know why it's taking another path, and why special handling is required for Linea? I thought that we should now be detecting fillRelay errors equally for all EVM chains?

fusmanii:
I dont think this is relevant anymore

PR #2535, opened 2025-08-28
improve(Monitor): Refactor getPendingWithdrawalAmounts and skip BSC

nicholaspai:
Adds a function `getTotalPendingWithdrawalAmount` to the `AdapterManager` that will be easier to use in the InventoryClient in the future, also skips BSC since `getL2PendingWithdrawalAmount` makes async API calls which tend to get rate limited by Binance API

### DIFF
--- a/src/clients/bridges/AdapterManager.ts
+++ b/src/clients/bridges/AdapterManager.ts
@@ -14,17 +14,16 @@ import {
   isDefined,
   winston,
   Signer,
-  getL2TokenAddresses,
   TransactionResponse,
   assert,
   EvmAddress,
-  toAddressType,
   TOKEN_EQUIVALENCE_REMAPPING,
   getRemoteTokenForL1Token,
   getTokenInfo,
   isEVMSpokePoolClient,
   Address,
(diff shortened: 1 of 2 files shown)


src/clients/bridges/AdapterManager.ts
```diff
@@ -248,11 +281,6 @@ export class AdapterManager {
       if (!l2TokenForL1Token) {
         throw new Error(`No L2 token found for L1 token ${l1Token} on chain ${chainId}`);
       }
-      if (!l2TokenForL1Token.eq(toAddressType(getL2TokenAddresses(l1Token.toEvmAddress())[chainId], chainId))) {
```
bmzig, line 251: Why do we need to get rid of this?

bmzig (dismissed):

src/clients/bridges/AdapterManager.ts
nicholaspai: So basically if we ever call `getL2TokenAddresses` with something like "USDC-BNB" it will return the USDC mapping which won't have a chain 56 entry

PR #2538, opened 2025-08-28
improve: retry on Binance deposit/withdraw event queries

bmzig:
Should help with throwing when we are get rate-limited. Also addresses the "... outside of recvWindow" error, which seems to happen sporadically (not sure why it happens, but it could be because API time drifts from the host time?).

### DIFF
--- a/src/adapter/bridges/BinanceCEXBridge.ts
+++ b/src/adapter/bridges/BinanceCEXBridge.ts
@@ -9,13 +9,14 @@ import {
   assert,
   isDefined,
   getTimestampForBlock,
-  mapAsync,
   getBinanceApiClient,
   floatToBN,
   CHAIN_IDs,
   compareAddressesSimple,
   isContractDeployedToAddress,
   winston,
+  getBinanceDeposits,
+  getBinanceWithdrawals,
 } from "../../utils";
 import { BaseBridgeAdapter, BridgeTransactionDetails, BridgeEvents } from "./BaseBridgeAdapter";
 import ERC20_ABI from "../../common/abi/MinimalERC20.json";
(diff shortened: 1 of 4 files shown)


src/utils/BinanceUtils.ts
```diff
@@ -67,3 +115,115 @@ export async function getBinanceWithdrawalLimits(binanceApi: BinanceApi): Promis
...
+        // Alternatively, if this was a native token transfer, then there were no logs, so just assign 0. This should not
+        // affect `sortEvents*` since the transaction index should be able to discriminate any two rebalances.
+        logIndex: depositTxnReceipt.logs[0]?.logIndex ?? 0,
+        txnIndex: depositTxnReceipt.transactionIndex,
+      };
+    });
+  } catch (_err) {
+    const err = _err.toString();
+    if (KNOWN_BINANCE_ERROR_REASONS.some((errorReason) => err.includes(errorReason)) && nRetries != maxRetries) {
+      const delaySeconds = 2 ** nRetries + Math.random();
```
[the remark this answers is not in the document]
dijanin-brat: Ofc, we will need to change elemnts in `KNOWN_BINANCE_ERROR_REASONS` to lower case as well.

src/adapter/bridges/BinanceCEXBridge.ts
```diff
@@ -96,32 +97,29 @@ export class BinanceCEXBridge extends BaseBridgeAdapter {
...
-      coin: this.tokenSymbol,
-      startTime: fromTimestamp,
-    });
+    const _depositHistory = await getBinanceDeposits(
+      binanceApiClient,
+      this.getL1Bridge().provider,
+      this.l2Provider,
+      fromTimestamp
+    );
+
```
dijanin-brat, line 106: Ooc, why aren't we passing `this.tokenSymbol` like we did in deleted part of the code?

src/utils/BinanceUtils.ts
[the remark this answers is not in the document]
bmzig: I personally don't think this is much of an issue. I think of it as equivalent to matching signatures, and if the signature changes because the API changes, then it would be pretty clear due to the newly thrown errors (and changing to lower case wouldn't do anything unless if they literally just made some upper case character lower case).

fusmanii (commented):
Looks good, just two changes around retry logic

src/utils/BinanceUtils.ts
```diff
@@ -116,22 +124,40 @@ export async function getBinanceDeposits(
...
-    return {
-      amount: Number(deposit.amount),
-      externalAddress: depositTxnReceipt.from,
-      coin: deposit.coin,
-      network: deposit.network,
-      status: deposit.status,
-    };
-  });
+  try {
+    const _depositHistory = await binanceApi.depositHistory({ startTime });
```
nicholaspai, line 132: I'm wary of large try-catch blocks, don't we only need to wrap this `binanceApi` call in the try-catch? For example, we woul dnever want the `provider` call below to trigger a retry here, right

src/utils/BinanceUtils.ts
```diff
@@ -116,22 +124,40 @@ export async function getBinanceDeposits(
   binanceApi: BinanceApi,
   l1Provider: ethers.providers.Provider,
   l2Provider: ethers.providers.Provider,
-  startTime: number
+  startTime: number,
+  nRetries = 0,
+  maxRetries = 10
```
nicholaspai, line 129: 10 retries max seems like a high default, typically we use 2/3: https://github.com/across-protocol/sdk/blob/80ae10a647eec01a32f2c3a8e388c818686f58f6/src/arch/svm/SpokeUtils.ts#L142

src/utils/BinanceUtils.ts
```diff
@@ -141,19 +167,40 @@ export async function getBinanceDeposits(
...
-    return {
-      amount: Number(withdrawal.amount),
-      externalAddress: withdrawal.address,
-      coin,
-      network: withdrawal.network,
-      status: withdrawal.status,
-    };
-  });
+  try {
+    const _withdrawHistory = await binanceApi.withdrawHistory({ coin, startTime });
```
nicholaspai, line 177: Same point above about only wrapping the binanceApi calls in try-catch

bmzig: Yeah that's a good point. Fixed here: https://github.com/across-protocol/relayer/pull/2538/commits/2219bd52dad442a0e8125f9779f1c263db5371c9

PR #2540, opened 2025-08-29
improve: log unfinalized amounts in more finalizers

bmzig:
Will give more info on unfinalized amounts.

### DIFF
--- a/src/finalizer/utils/cctp/l2ToL1.ts
+++ b/src/finalizer/utils/cctp/l2ToL1.ts
@@ -17,6 +17,8 @@ import {
   Address,
   toKitAddress,
   getStatePda,
+  toAddressType,
+  createFormatFunction,
 } from "../../../utils";
 import {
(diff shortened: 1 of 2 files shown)


src/finalizer/utils/cctp/l2ToL1.ts
```diff
@@ -60,10 +62,21 @@ export async function cctpL2toL1Finalizer(
     outstandingDeposits,
     (message: { status: CCTPMessageStatus }) => message.status
   );
+  const pendingDeposits = outstandingDeposits.filter(({ status }) => status === "pending").map((deposit) => {
```
pxrl: Do we want to just call this `pending`. The terminology is a bit weird as to whether it's a deposit or a withdrawal - I think from Circle's perspective it's a deposit, but from our perspective you could argue it's a withdrawal.

pxrl (approved):

PR #2553, opened 2025-09-04
improve(monitor): Use BundleDataApproxClient to estimate upcoming refunds

nicholaspai:
The motivation for this is to remove dependence on the BundleDataClient in the Monitor. With this PR we can delete most of the helper functions in the BundleDataClient that are only used by the Monitor

This PR is being merged into https://github.com/across-protocol/relayer/pull/2506

This also removes BundleDataClient dependency in the `validateRunningBalances` script

### DIFF
--- a/test/Monitor.ts
+++ b/test/Monitor.ts
@@ -263,9 +263,9 @@ describe("Monitor", async function () {
 
     // setupDataworker seeds relayer with 10 * 1500 erc20_2, erc20_1, and l1Token_1 tokens on two different
     // spoke pools, adding to a total of 6 * 10 * 1500 = 90,000 tokens.
-    expect(
-      reports[relayerAddress.toBytes32()]["L1Token1"][ALL_CHAINS_NAME][BalanceType.CURRENT].toString()
-    ).to.be.equal("90000000000000000000000");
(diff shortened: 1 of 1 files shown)


src/monitor/Monitor.ts
```diff
@@ -976,34 +963,15 @@ export class Monitor {
 
     // Calculate the pending refunds.
     for (const chainId of chainIds) {
+      const l2TokenMap = this.getL2ToL1TokenMap(enabledTokens, chainId);
       const l2TokenAddresses = monitoredTokenSymbols
         .map((symbol) => l2TokenForChain(chainId, symbol))
         .filter(isDefined);
       pendingRelayerRefunds[chainId] = {};
       l2TokenAddresses.forEach((l2Token) => {
-        const pendingValidatedDeductions = accountedBundleRefunds
```
nicholaspai, line 984: When checking relayer balances there is really no need to split up upcoming refunds between those in the pending bundle and those in following bundles for our accounting purposes. Therefore we can simplify this logic a bit by asking the BundleApproxClient to give us all upcoming refunds

pxrl (approved):
Nice simplification!

pxrl (approved):

PR #2565, opened 2025-09-10
improve(relayer): Proceed to active state despite degraded chains

pxrl:
Observed this morning on Polygon - the relayer refused to start due to Polygon node issues. Try to proceed to active state even though one or more chains did not come up successfully.

### DIFF
--- a/src/relayer/index.ts
+++ b/src/relayer/index.ts
@@ -71,7 +71,6 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
       const activeRelayer = redis ? await redis.get(botIdentifier) : undefined;
 
       // If there is another active relayer, allow up to 120 seconds for this instance to be ready.
-      // If this instance can't update, throw an error (for now).
       if (!ready && activeRelayer) {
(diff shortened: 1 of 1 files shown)


dohaki (dismissed):
nice

src/relayer/index.ts
```diff
@@ -81,10 +80,10 @@ export async function runRelayer(_logger: winston.Logger, baseSigner: Signer): P
           continue;
         }
 
-        const badChains = Object.values(spokePoolClients)
+        const degraded = Object.values(spokePoolClients)
           .filter(({ isUpdated }) => !isUpdated)
           .map(({ chainId }) => getNetworkName(chainId));
-        throw new Error(`Unable to start relayer due to chains ${badChains.join(", ")}`);
+        logger.debug({ at: "Relayer#run", message: "Assuming active relayer role in degraded state", degraded });
```
bmzig, line 86: This would only ever log at most once per relayer run, right? Do you think it's worth making it a warn log so it's more visible?

bmzig (dismissed):

bmzig (approved):

PR #2575, opened 2025-09-13
improve: add a `vkey` check to helios finalizer

grasphoper:
There's no functionality to erase a proof from the ZK Api if it was requested and didn't error. To prevent incidental manual cleanup required on a wrong proof request, we compare vkeys between an API and contract to only generate matching proofs


src/finalizer/utils/helios.ts
```diff
@@ -757,3 +770,31 @@ function addUpdateOnlyTxn(
...
+ */
+async function ensureVkeysMatch(apiBaseUrl: string, sp1Helios: ethers.Contract): Promise<void> {
+  const [apiResp, contractVkeyRaw] = await Promise.all([
+    axios.get<VkeyResponse>(`${apiBaseUrl}/v1/api/vkey`),
+    sp1Helios.heliosProgramVkey(),
+  ]);
+
+  const apiVkeyRaw = apiResp?.data?.vkey;
+  const apiVkey = apiVkeyRaw !== undefined ? apiVkeyRaw.toLowerCase() : undefined;
+  const contractVkey = contractVkeyRaw !== undefined ? contractVkeyRaw.toLowerCase() : undefined;
```
pxrl, line 793: Minor driveby comment - you could consider this to avoid having to repeat the variable names, which can be error-prone:

```suggestion
  const apiVkey = apiVkeyRaw?.toLowerCase();
  const contractVkey = contractVkeyRaw?.toLowerCase();
```

src/utils/HeliosUtils.ts
```diff
@@ -21,5 +21,5 @@ export async function getSp1HeliosContractEVM(
     evmSpokePool.provider
   );
   const heliosAddress = await universalSpokePoolContract.helios();
-  return new ethers.Contract(heliosAddress, SP1_HELIOS_ABI as any, signerOrProvider);
+  return new ethers.Contract(heliosAddress, SP1_HELIOS_ABI, signerOrProvider);
```
pxrl, line 24: :+1:

src/finalizer/utils/helios.ts
```diff
@@ -757,3 +770,31 @@ function addUpdateOnlyTxn(
...
+ *
+ * When upgrading the ZK Helios setup, ZK API will always have to be upgraded _after_ we send an upgrade message to all
+ * the V4 chains, because ZK API has to _first_ generate all the proofs for "upgrade message" to get confirmed on the
+ * destination chain. This means that ELF vkey might become stale in the ZK API (say some contract already upgraded but
+ * we didn't yet re-deploy the ZK API with the new ELF). This check eases the operational overhead to the upgrade by not
+ * allowing to request a proof with an incorrect(stale) ELF vkey.
+ */
+async function ensureVkeysMatch(apiBaseUrl: string, sp1Helios: ethers.Contract): Promise<void> {
+  const [apiResp, contractVkeyRaw] = await Promise.all([
+    axios.get<VkeyResponse>(`${apiBaseUrl}/v1/api/vkey`),
```
pxrl, line 787: For future, we probably want 1 - 2 retries on this to avoid sporadic issues.

pxrl (dismissed):

src/finalizer/utils/helios.ts
grasphoper: Added with `axios-retry` here [daf7cea](https://github.com/across-protocol/relayer/pull/2575/commits/daf7cea6e4026949555bf6204c1dc56710bd2cb1)

grasphoper:
I actually changed my mind on having this `axios-retry` dependency. Will reimpl. dependency-free on Monday

src/finalizer/utils/helios.ts
```diff
@@ -762,3 +760,27 @@ function addUpdateOnlyTxn(
...
+ * match, the generated proof cannot be used to confirm messages on the destination chain.
+ *
+ * When upgrading the ZK Helios setup, ZK API will always have to be upgraded _after_ we send an upgrade message to all
+ * the V4 chains, because ZK API has to _first_ generate all the proofs for "upgrade message" to get confirmed on the
+ * destination chain. This means that ELF vkey might become stale in the ZK API (say some contract already upgraded but
+ * we didn't yet re-deploy the ZK API with the new ELF). This check eases the operational overhead to the upgrade by not
+ * allowing to request a proof with an incorrect(stale) ELF vkey.
+ */
+async function ensureVkeysMatch(apiBaseUrl: string, sp1Helios: ethers.Contract): Promise<void> {
+  const [apiResp, contractVkeyRaw] = await Promise.all([getVkeyWithRetries(apiBaseUrl), sp1Helios.heliosProgramVkey()]);
```
nicholaspai, line 776: Are these likely to change frequently? Should we cache this in redis to avoid extra queries on every finalizer run? We can follow a pattern like [this](https://github.com/across-protocol/relayer/blob/b98267a466cc46ae75f842566229b472d5ae936c/src/adapter/BaseChainAdapter.ts#L161) where we load the `redis` cache within this function so we don't need to pass it by input

PR #2577, opened 2025-09-15
feat(finalizer): Add HyperEVM support for helios finalizer + CCTP utils support

nicholaspai:


bmzig (approved):

src/utils/CCTPUtils.ts
```diff
@@ -81,15 +81,17 @@ const CCTP_DEPOSIT_FOR_BURN_TOPIC_HASH_V2 = ethers.utils.id(
...
  * @param chainId
  * @returns True if the chainId is a CCTP V2 chain
  */
 export function isCctpV2L2ChainId(chainId: number): boolean {
-  return CCTP_V2_L2_CHAINS.includes(chainId);
+  return (
+    chainId !== CHAIN_IDs.MAINNET &&
+    CONTRACT_ADDRESSES[chainId].cctpV2TokenMessenger !== undefined &&
+    CONTRACT_ADDRESSES[chainId].cctpV2MessageTransmitter !== undefined
+  );
```
grasphoper, line 94: We seem to be using `isCctpV2L2ChainId` mostly in bridge methods. This change means that we prefer V2 there. Seems fine

Are there any other side effects to this change?
It means that we auto prefer V2 if both V1 and V2 are defined in CONTRACT_ADDRESSES

nicholaspai: Both V1 and V2 are not defined for any chains, and I think we should keep it that way. I think this is fine as implemented

PR #2579, opened 2025-09-16
chore: Trim down data that is logged for invalid fills

dijanin-brat:
Current logs of the invalid fills is hard to read. Trim down data that is logged for invalid fills.

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -218,12 +218,22 @@ export class Monitor {
 
     invalidFills.forEach((invalidFill) => {
       // Log the fill data
+      const deposit = invalidFill.deposit
+        ? {
+            txnRef: invalidFill.deposit.txnRef,
+            inputToken: invalidFill.deposit.inputToken,
+            depositor: invalidFill.deposit.depositor,
+          }
+        : undefined;
       this.logger.warn({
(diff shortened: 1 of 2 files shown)


src/monitor/Monitor.ts
```diff
@@ -217,13 +218,40 @@ export class Monitor {
...
-      // Log the fill data
+      const destinationChainId = invalidFill.fill.destinationChainId;
+      const outputToken = invalidFill.fill.outputToken;
+      let tokenInfo: TokenInfo;
+
+      try {
+        tokenInfo = this.clients.hubPoolClient.getTokenInfoForAddress(outputToken, destinationChainId);
+      } catch {
+        tokenInfo = { symbol: "UNKNOWN TOKEN", decimals: 18, address: outputToken };
+      }
```
pxrl, line 229: @dijanin-brat You might want to use this function directly: https://github.com/across-protocol/sdk/blob/1e09b0bf430405ee58fc215e3a749652baa71ff4/src/utils/TokenUtils.ts#L113

This is what the HubPoolClient uses anyway, fwiw.

dijanin-brat: Yea, but i saw that there is some USDC handling in the HubPoolClient so i decided to go with this one because of that

PR #2580, opened 2025-09-16
refactor(TransactionUtils): Narrow try/catch scope

pxrl:
The existing try/catch scope spans both gas price retrieval and transaction submission. This creates difficulty in handling errors, especially due to the use of viem for transaction simulation via linea_estimateGas on Linea.

Narrow the scope of transaction submission to improve error handling, and drop gas price error handling from the try/catch scope entirely. While here, consolidate some neighbouring code to reduce duplication and simplify the implementation.

More changes to follow, including an update to error handling and retry logic on transaction submission.


pxrl:
nb. most implementation pulled out of this PR because it was getting tricky to read and is held up by some additional testing on various chains: https://github.com/across-protocol/relayer/pull/2126

nicholaspai (approved):

PR #2589, opened 2025-09-17
improve: Log number of invalid fills per chain

dijanin-brat:
Log number of invalid fills per chain

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -217,7 +217,10 @@ export class Monitor {
   async reportInvalidFills(): Promise<void> {
     const invalidFills = await sdkUtils.findInvalidFills(this.clients.spokePoolClients);
 
+    const invalidFillsByChainId: Record<string, number> = {};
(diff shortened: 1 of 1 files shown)


src/monitor/Monitor.ts
```diff
@@ -255,6 +258,12 @@ export class Monitor {
         notificationPath: "across-invalid-fills",
       });
     });
+
+    this.logger.info({
+      at: "Monitor::invalidFillsByChain",
+      message: "Invalid fills by chain",
+      invalidFillsByChainId,
```
bmzig, line 265: I think the notification path needs to change, since it's good to keep spam away from the monitor channel as there are so many potentially important things logged there.
```suggestion
      invalidFillsByChainId,
      notificationPath: "across-invalid-fills",
```

dijanin-brat: [28cb41b](https://github.com/across-protocol/relayer/pull/2589/commits/28cb41bff5606ef86da2368cffcc0e12d0788794)

PR #2593, opened 2025-09-17
feat: Create Refiller bot

nicholaspai:
This removes all the `refillBalances` logic from the Monitor and moves it into its own bot. This should be much easier to maintain

### DIFF
--- a/src/refiller/Refiller.ts
+++ b/src/refiller/Refiller.ts
@@ -1,0 +1,419 @@
+import winston from "winston";
+import { RefillerConfig } from "./RefillerConfig";
+import {
+  Address,
+  assert,
+  BigNumber,
+  blockExplorerLink,
+  CHAIN_IDs,
+  chainIsEvm,
+  chainIsSvm,
+  Contract,
+  ERC20,
+  EvmAddress,
+  formatUnits,
+  getNativeTokenAddressForChain,
+  getNativeTokenSymbol,
+  getNetworkName,
+  getProvider,
+  getRedisCache,
+  getSolanaTokenBalance,
+  getSvmProvider,
+  getTokenInfo,
+  mapAsync,
+  parseUnits,
+  runTransaction,
+  sendRawTransaction,
+  Signer,
+  toAddressType,
+  toBN,
+  TOKEN_SYMBOLS_MAP,
+  TransactionReceipt,
+  WETH9,
+  ZERO_ADDRESS,
+} from "../utils";
+import { arch } from "@across-protocol/sdk";
+import { AcrossSwapApiClient, BalanceAllocator, HubPoolClient, MultiCallerClient } from "../clients";
+import { RedisCache } from "../caching/RedisCache";
+
+export interface RefillerClients {
+  balanceAllocator: BalanceAllocator;
+  hubPoolClient: HubPoolClient;
+  multiCallerClient: MultiCallerClient;
+}
+
+type SwapRoute = {
+  inputToken: EvmAddress;
+  outputToken: EvmAddress;
+  originChainId: number;
+  destinationChainId: number;
+};
+/**
+ * @notice This class is in charge of refilling native token balances for accounts running other bots, like the relayer and
+ * dataworker. It is run with an account whose funds are used to refill balances for other accounts by either swapping
+ * and transferring on the same chain or sending a cross chain swap.
+ */
+export class Refiller {
+  private balanceCache: { [chainId: number]: { [token: string]: { [account: string]: BigNumber } } } = {};
+  private decimals: { [chainId: number]: { [token: string]: number } } = {};
+  private acrossSwapApiClient: AcrossSwapApiClient;
+  private redisCache: RedisCache;
+  private baseSigner: Signer;
+  private baseSignerAddress: EvmAddress;
+  private enabledChainIds: number[];
+
+  public constructor(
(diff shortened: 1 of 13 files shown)


src/monitor/index.ts
```diff
@@ -45,12 +45,6 @@ export async function runMonitor(_logger: winston.Logger, baseSigner: Signer): P
         logger.debug({ at: "AcrossMonitor", message: "Report disabled" });
       }
 
-      if (config.botModes.refillBalancesEnabled) {
-        await acrossMonitor.refillBalances();
-      } else {
-        logger.debug({ at: "Monitor#index", message: "Refiller disabled" });
-      }
```
pxrl, line 52: Do we maybe want to leave this in for the time being, so we can toggle it on or off on demand?

src/refiller/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    logger[startupLogLevel(config)]({ at: "Refiller#index", message: "Refiller started ⛽️", config });
+    for (;;) {
+      const loopStart = Date.now();
+      await refiller.refillNativeTokenBalances();
+
+      await clients.multiCallerClient.executeTxnQueues();
+
+      logger.debug({ at: "Monitor#index", message: `Time to loop: ${(Date.now() - loopStart) / 1000}s` });
+
+      if (await processEndPollingLoop(logger, "Monitor#index", config.pollingDelay)) {
```
pxrl, line 25: Do we want to retain support for looping behaviour? We don't use it in practice, some clients don't handle it very well, and it kinda pollutes the code.

src/refiller/Refiller.ts
```diff
@@ -0,0 +1,432 @@
...
+import { arch } from "@across-protocol/sdk";
+import { AcrossSwapApiClient, BalanceAllocator, MultiCallerClient } from "../clients";
+import { RedisCache } from "../caching/RedisCache";
+
+export interface RefillerClients {
+  balanceAllocator: BalanceAllocator;
+  // We can further constrain the HubPoolClient type since we don't call any functions on it.
+  hubPoolClient: { hubPool: Contract; chainId: number };
+  multiCallerClient: MultiCallerClient;
+}
```
pxrl, line 43: I'm curious how it'd look if we handled HubPool refills separately. It looks like handling native transfers + HubPool refills complicates the refill logic a bit and produces some fairly long functions.

src/refiller/Refiller.ts
pxrl, line 1: ooc on this, how much is pure copy/paste vs. opportunistic refactoring? It's a bit hard to gauge because git perceives it as new code.

src/refiller/Refiller.ts
```diff
@@ -0,0 +1,432 @@
...
+        this.logger.debug({
+          at: "Refiller#refillBalances",
+          message: "Balance is above trigger",
+          account,
+          balanceTrigger,
+          currentBalance: currentBalance.toString(),
+          token,
+          chainId,
+        });
+      }
```
pxrl, line 276: For subsequent refactoring, I think we should flip the order of the if statement that guards this code. We can drop a level of indentation on a lot of code if we do:

```
if (!isBelowTrigger) {
  // log message
  return;
}
```

It should really be done separately though, because otherwise it gets hard to isolate why something broke in all this re-arranging.

src/refiller/Refiller.ts
```diff
@@ -0,0 +1,432 @@
...
+      this.logger.warn({
+        at: "Refiller#refillNativeTokenBalances",
+        message: `Trying to unwrap WETH balance from ${this.baseSignerAddress} to use for refilling ETH but not enough WETH to unwrap`,
+        chainId,
+        requiredUnwrapAmount: amount.toString(),
+        wethAddress: weth.address,
+        account: this.baseSignerAddress,
+      });
+      return;
+    }
```
pxrl, line 304: Can drop this indentation since the alt scenario is to return.

src/refiller/Refiller.ts
```diff
@@ -0,0 +1,432 @@
...
+      // miscellaneous reasons.
+      this.logger.warn({
+        at: "Monitor#refillBalances",
+        message: `Failed to swap ETH on ${getNetworkName(swapRoute.originChainId)} to MATIC on Polygon`,
+        swapRoute,
+        amount,
+        swapper: this.baseSignerAddress,
+        recipient: recipient.toEvmAddress(),
+      });
+    }
```
pxrl, line 368: This is another one where it probably makes sense to do:

```
if (!swapData) {
  // log message
  return:
}
```

src/monitor/index.ts
nicholaspai: I think we should remove it, its fine if the refiller misses a few cycles, and I can prepare the config file for this in advanced before merging

src/refiller/index.ts
nicholaspai: 3dfb8ef8a65f2820e8b9d9640585af807c4ddcb2

src/refiller/Refiller.ts
nicholaspai: Its not that bad, its just one small if section. I'd rather not change the code that much in this PR

src/refiller/Refiller.ts
nicholaspai: good call I'll add a commment

PR #2598, opened 2025-09-18
improve(InventoryClient): Initialize BundleDataApproxClient with all possible L1 tokens

nicholaspai:
Downside:
- initialize()-ing the BundleDataApproxClient will be slower because it is now fetching refunds over all L1 tokens, regardless if the user sets a smaller subset in the inventory config

Upside:
- Relayer will not complain if a token is enabled in RELAYED_SUPPORTED_TOKENS but not in the INVENTORY_CONFIG because the refund data will now be available in the BundleDataApproxClient

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -92,7 +92,7 @@ export class InventoryClient {
       this.tokenClient?.spokePoolClients ?? {},
       this.hubPoolClient,
       this.chainIdList,
-      this.getL1Tokens(),
+      this.getAllL1Tokens(),
       this.logger
     );
   }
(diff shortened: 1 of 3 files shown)


src/clients/InventoryClient.ts
```diff
@@ -310,11 +310,21 @@ export class InventoryClient {
     return this.getEnabledChains().filter((chainId) => chainId !== hubPoolChainId);
   }
 
-  getL1Tokens(): EvmAddress[] {
-    return (
-      Object.keys(this.inventoryConfig?.tokenConfig ?? {}).map((token) => EvmAddress.from(token)) ||
```
nicholaspai, line 315: I think this code is currently buggy because the RHS of the `||` seems impossible to hit? wdyt

pxrl: Agree; `Object.keys()` will always return an array so the `||` conditional should never be hit.

```
> [] || [1,2,3]
[]
```

src/clients/InventoryClient.ts
```diff
@@ -92,7 +92,7 @@ export class InventoryClient {
       this.tokenClient?.spokePoolClients ?? {},
       this.hubPoolClient,
       this.chainIdList,
-      this.getL1Tokens(),
+      this.getL1TokensFromInventoryConfig().concat(this.getL1TokensEnabledInHubPool()),
```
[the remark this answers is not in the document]
nicholaspai: https://github.com/across-protocol/relayer/pull/2598/commits/7e347f783a8151879192045b082df3048b35171c

bmzig (approved):

PR #2599, opened 2025-09-18
refactor(TransactionUtils): Factor out gas price scaling

pxrl:
More diff reduction from the overarching effort to improve transaction submission robustness. Proceed with a gas scaler of 1.0 for now; this will be made dynamic in a follow-up change.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -9,6 +9,7 @@ import {
   bnZero,
   Contract,
   isDefined,
+  fixedPointAdjustment,
   TransactionResponse,
   ethers,
   getContractInfoFromAddress,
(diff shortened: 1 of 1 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -118,12 +119,13 @@ export async function runTransaction(
...
   let gas: Partial<FeeData>;
   try {
-    gas = await getGasPrice(
+    const preGas = await getGasPrice(
       provider,
       priorityFeeScaler,
       maxFeePerGasScaler,
       sendRawTxn ? undefined : await contract.populateTransaction[method](...(args as Array<unknown>), { value })
     );
+    gas = scaleGasPrice(chainId, preGas);
```
pxrl, line 128: There's a slightly weird overlap here, whereby `getGasPrice()` gets a gas price with base + priority fee scalers that are env-configured, and then there's a subsequent adjustment to set a floor and apply an additional scaler. I'm not really happy with this and it needs to be consolidated; some of this functionality has anyway emerged from trying to combat stuck transactions.

I'm pretty confident we can circle back here and consolidate the scaling once we have addressed the underlying issue of transactions getting stuck on occasion.

nicholaspai (approved):

PR #2602, opened 2025-09-18
fix: Suppress logs if there are no Invalid Fills

dijanin-brat:


src/monitor/Monitor.ts
```diff
@@ -266,12 +266,14 @@ export class Monitor {
       });
     });
 
-    this.logger.info({
-      at: "Monitor::invalidFillsByChain",
-      message: "Invalid fills by chain",
-      invalidFillsByChainId,
-      notificationPath: "across-invalid-fills",
-    });
+    if (invalidFillsByChainId.length > 0) {
```
pxrl, line 269: invalidFillsByChainId is a Record type; Should this be `Object.keys(invalidFillsByChainId).length > 0` ?

pxrl (approved):

PR #2606, opened 2025-09-19
fix(rebalancer): Don't flush TokenClient state prematurely

pxrl:
Flushing TokenClient state before considering withdrawals means that the rebalancer assumes that there are no deposits awaiting to be filled. This can cause the rebalancer to pull back tokens that were intended to be used by the relayer to fill shortfalls. The rebalancer should instead factor in any simulated fills that were made as part of identifying shortfalls.


bmzig (dismissed):

src/relayer/index.ts
```diff
@@ -188,9 +188,6 @@ export async function runRebalancer(_logger: winston.Logger, baseSigner: Signer)
     await rebalancer.checkForUnfilledDepositsAndFill(false, true);
     await rebalancer.runMaintenance();
```
pxrl, line 189:
```suggestion
```

bmzig (approved):

PR #2607, opened 2025-09-20
improve(proposer): Update proposal log format

pxrl:
This log message can be pretty difficult to read in Slack. Prioritise logging fields that are sought the most often.

### DIFF
--- a/src/dataworker/PoolRebalanceUtils.ts
+++ b/src/dataworker/PoolRebalanceUtils.ts
@@ -106,23 +106,32 @@ export function generateMarkdownForRootBundle(
   };
   let poolRebalanceLeavesPretty = "";
   poolRebalanceLeaves.forEach((leaf, index) => {
-    // Shorten keys for ease of reading from Slack.
-    delete leaf.leafId;
-    leaf.groupId = leaf.groupIndex;
-    delete leaf.groupIndex;
-    leaf.bundleLpFees = convertTokenListFromWei(hubPoolChainId, leaf.l1Tokens, leaf.bundleLpFees);
-    leaf.runningBalances = convertTokenListFromWei(hubPoolChainId, leaf.l1Tokens, leaf.runningBalances);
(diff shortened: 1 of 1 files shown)


src/dataworker/PoolRebalanceUtils.ts
```diff
@@ -106,23 +106,32 @@ export function generateMarkdownForRootBundle(
...
+      chainId,
+      l1Tokens,
+      netSendAmounts,
+      runningBalances,
+      refundAddresses,
+      refundAmounts,
+      bundleLpFees,
+      groupId,
+    };
+    poolRebalanceLeavesPretty += `\n\t\t\t${index}: ${JSON.stringify(orderedLeaf)}`;
```
pxrl, line 126: I'm mulling over some ideas about how this might be improved even further. I'd like to get this in as-is, but we might circle back with another change too.

pxrl:
nb currently looks like the screenshots below. There are some issues:
- Monospace font (for alignment) seems to require a code block, which is kinda ugly.
- Large logs can be split over multiple messages in Slack; this causes formatting issues for the leaf that spans two messages.

It seems like we'll at least need to split the current single log message into:
- Bundle block numbers
- Pool rebalance leaves
- Relayer refund leaves

We will probably also need to split the logs for relayer refund leaves into smaller chunks to avoid splitting them over multiple Slack messages.

nicholaspai (approved):

bmzig:
> nb currently looks like the screenshots below. There are some issues:
Do you plan on addressing the issue about a table being split over multiple messages in this PR?

pxrl:
> > nb currently looks like the screenshots below. There are some issues:
> 
> Do you plan on addressing the issue about a table being split over multiple messages in this PR?

Yeah; so far I am splitting the existing single message into bundle blocks, pool rebalance leaves, relayer refund leaves and slow fill leaves. I think the relayer refund leaves will need to be chunked into smaller messages again, though.

PR #2609, opened 2025-09-22
chore(relayer): Cleanup fill simulation overrides

pxrl:
The chains updated here now support USDC.


pxrl:
nb. to be merged following Plasma support.

PR #2619, opened 2025-09-26
Guard stringifyThrownValue against null/undefined

MarkFizz77:
## Description

Add a null/undefined guard to stringifyThrownValue in `src/utils/LogUtils.ts` to prevent `null.toString()`/`undefined.toString()` TypeErrors during error logging. This path is hit when non-Error values (e.g., rejected promises, library throws) bubble up to logging.


pxrl:
No response to the feedback, so submitting separately here: https://github.com/across-protocol/relayer/pull/2643

PR #2620, opened 2025-09-26
feat: Support CCTP V2 rebalancing for EVM chains

nicholaspai:
This should significantly improve capital efficiency by rebalancing via the CCTP V2 Fast Transfer protocol wherever possible. The fee is taken out from the destination chain mint transaction.

A follow up PR will add this to the Solana adapters but I'm not sure we support CCTP V2 on Solana yet. Once we confirm that we do, then we can easily switch over to using fast transfers, but if we're assuming in the code that we're interacting with V1 contracts and interfaces, then we can't easily switch. Regardless, I want to first demo this with EVM chains.

### DIFF
--- a/src/adapter/bridges/UsdcCCTPBridge.ts
+++ b/src/adapter/bridges/UsdcCCTPBridge.ts
@@ -66,6 +66,8 @@ export class UsdcCCTPBridge extends BaseBridgeAdapter {
     amount: BigNumber
   ): Promise<BridgeTransactionDetails> {
     assert(l1Token.eq(this.l1UsdcTokenAddress));
+    // Check for fast-transfer allowance and also min fee, and if they are reasonable, then
+    // construct a fast transfer, otherwise default to a standard transfer.
     amount = amount.gt(this.CCTP_MAX_SEND_AMOUNT) ? this.CCTP_MAX_SEND_AMOUNT : amount;
     return Promise.resolve({
       contract: this.getL1Bridge(),
--- a/src/adapter/bridges/UsdcCCTPBridge.ts
+++ b/src/adapter/bridges/UsdcCCTPBridge.ts
@@ -13,9 +13,11 @@ import {
   paginatedEventQuery,
   ethers,
   winston,
+  getTokenInfo,
+  toBNWei,
 } from "../../utils";
 import { processEvent } from "../utils";
-import { getCctpTokenMessenger, isCctpV2L2ChainId } from "../../utils/CCTPUtils";
+import { getCctpTokenMessenger, getV2DepositForBurnMaxFee, isCctpV2L2ChainId } from "../../utils/CCTPUtils";
 import { CCTP_NO_DOMAIN } from "@across-protocol/constants";
 import { CCTP_MAX_SEND_AMOUNT } from "../../common";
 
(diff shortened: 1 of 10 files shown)


src/adapter/bridges/UsdcCCTPBridge.ts
```diff
@@ -116,7 +128,22 @@ export class UsdcCCTPBridge extends BaseBridgeAdapter {
     const events = await paginatedEventQuery(this.getL2Bridge(), eventFilter, eventConfig);
     // There is no "from" field in this event, so we set it to the L2 token received.
     return {
-      [this.resolveL2TokenAddress(this.l1UsdcTokenAddress)]: events.map((event) => processEvent(event, "amount")),
+      [this.resolveL2TokenAddress(this.l1UsdcTokenAddress)]: events.map((event) => {
+        const eventSpread = spreadEventWithBlockNumber(event) as unknown as SortableEvent & {
+          amount: string;
+          feeCollected: string;
+        };
+        const amount = toBN(eventSpread.amount);
```
bmzig, line 136: Do you need to have an
```
if (isV2) {
 do this
 } else {
 do what we had before
 }
 ```
[Looking at a V1 contract,](https://etherscan.io/address/0xbd3fa81b58ba92a82136038b25adec7066af3155#events) `MintAndWithdraw` doesn't have a `feeCollected` field, so I think this would break since `toBN(undefined)` likely throws.

Comment (nicholaspai) [reply src/adapter/bridges/UsdcCCTPBridge.ts]: ah yes true: https://github.com/across-protocol/relayer/pull/2620/commits/79e6a4054c436ac07700681c9d60ba394f13f9e8

**Review by bmzig — DISMISSED** [review]
Looks correct

**Comment by nicholaspai** [comment]
This PR has been changed now to only use fast transfers when there is a shortfall

File: src/clients/InventoryClient.ts [hunk 2]
```diff
@@ -1027,7 +1028,19 @@ export class InventoryClient {
       // is already complex logic and most of the time we'll not be sending batches of rebalance transactions.
       for (const rebalance of possibleRebalances) {
         const { chainId, l1Token, l2Token, amount } = rebalance;
-        const { hash } = await this.sendTokenCrossChain(chainId, l1Token, amount, this.simMode, l2Token);
+        // Send a faster transfer if there is an active shortfall on the chain, otherwise use the slower, cheaper,
+        // default transfer.
+        const optionalParams: TransferTokenParams = {
+          fastMode: this.tokenClient.getShortfallTotalRequirement(chainId, l2Token).gt(bnZero),
```
nicholaspai, line 1034: I have a follow up branch where this logic becomes a lot more accurate: we check each rebalance individually to see if it specifically covers a shortfall, and then use that to toggle this `fastMode` flag. For now, I think this is a good heuristic: if there is a shortfall on this chain, fast rebalance to it

src/adapter/bridges/UsdcCCTPBridge.ts
```diff
@@ -62,25 +71,39 @@ export class UsdcCCTPBridge extends BaseBridgeAdapter {
...
+          amount,
+          this.l2DestinationDomain,
+          toAddress.toBytes32(),
+          this.l1UsdcTokenAddress.toNative(),
+          ethers.constants.HashZero, // Anyone can finalize the message on domain when this is set to bytes32(0)
+          maxFee,
+          finalityThreshold,
+        ],
+      });
+    } else {
```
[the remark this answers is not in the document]
nicholaspai: 2efb1ffd5f0a51f7f9d392598846ea7f7f9c6624

src/clients/InventoryClient.ts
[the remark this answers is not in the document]
nicholaspai: https://github.com/across-protocol/relayer/pull/2620/commits/2fa92840c048075ba510bd46af8e3b861bea0ebb

fusmanii (approved):

pxrl (approved):
No comments from me 👍

PR #2622, opened 2025-09-26
feat: ignore large SVM messages

grasphoper:
Estimated how many bytes would a fill for a deposit take. Drop deposit processing if > 1232 bytes as we don't support multi-part SVM-side filling

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -349,6 +349,7 @@ export class Relayer {
         recipient: recipient.toString(),
         outputToken: deposit.outputToken.toString(),
         exclusiveRelayer: deposit.exclusiveRelayer.toString(),
+        txnRef: deposit.txnRef,
       });
 
       return ignoreDeposit();
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -28,10 +28,12 @@ import {
   CHAIN_IDs,
   convertRelayDataParamsToBytes32,
   chainIsSvm,
+  stringifyThrownValue,
+  addrsMatchDstChain,
 } from "../utils";
 import { RelayerClients } from "./RelayerClientHelper";
 import { RelayerConfig } from "./RelayerConfig";
-import { MultiCallerClient } from "../clients";
+import { MultiCallerClient, SOLANA_TX_SIZE_LIMIT } from "../clients";
 
 const { getAddress } = ethersUtils;
 const { isDepositSpedUp, isMessageEmpty, resolveDepositMessage } = sdkUtils;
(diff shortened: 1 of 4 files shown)


src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+import { chainIsEvm } from "./SDKUtils";
+
+// Checks if deposit's .recipient, .outputToken and .exclusiveRelayer are correct for its destinationChainId
+export function addrsMatchDstChain(deposit: Deposit): boolean {
+  const destinationChainId = deposit.destinationChainId;
+  const isEvmChain = chainIsEvm(destinationChainId);
+
+  const addrValidityChecks = [
+    isEvmChain ? deposit.recipient.isEVM() : deposit.recipient.isSVM(),
+    isEvmChain ? deposit.outputToken.isEVM() : deposit.outputToken.isSVM(),
```
grasphoper, line 11: For deposits with `outputToken == zeroaddr`, should I add that exemption here?

src/relayer/Relayer.ts
```diff
@@ -336,6 +338,23 @@ export class Relayer {
       return ignoreDeposit();
     }
 
+    // Skip deposits which have ill-formatted addresses for the destination chain
+    if (!addrsMatchDstChain(deposit)) {
```
pxrl, line 342: nb. I think we should already be covered by [this check](https://github.com/across-protocol/relayer/blob/1ea167193ce498ca439ae5463c8ca3590d9c5e70/src/relayer/Relayer.ts#L265).

src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+    deposit.exclusiveRelayer.isZeroAddress()
+      ? isEvmChain
+        ? deposit.exclusiveRelayer.isEVM()
+        : deposit.exclusiveRelayer.isSVM()
+      : true,
+  ];
```
pxrl, line 17: This check is fwiw not a good idea, since the deposit can specify any 32-byte address on deposit, but the deposit is still valid and can still be filled after `exclusivityDeadline`. So we don't actually want to filter on this field.

src/relayer/Relayer.ts
grasphoper: Oh, you're totally right. Then I can just remove the `addrsMatchDstChain` alltogether and keep the asserts in `evaluateFill`

grasphoper: [3472959](https://github.com/across-protocol/relayer/pull/2622/commits/347295935bfc2588ba6ed9532615fca0ce1d34c8)

src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { Deposit } from "../interfaces";
+import { chainIsEvm } from "./SDKUtils";
+
+// Checks if deposit's .recipient, .outputToken and .exclusiveRelayer are correct for its destinationChainId
+export function addrsMatchDstChain(deposit: Deposit): boolean {
+  const destinationChainId = deposit.destinationChainId;
+  const isEvmChain = chainIsEvm(destinationChainId);
+
+  const addrValidityChecks = [
```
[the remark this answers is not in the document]
grasphoper: Removed here [3472959](https://github.com/across-protocol/relayer/pull/2622/commits/347295935bfc2588ba6ed9532615fca0ce1d34c8)

pxrl (dismissed):

nicholaspai (approved):

fusmanii (approved):

bmzig (approved):

PR #2627, opened 2025-09-30
feat: Add CCTP V2 any<>any finalizer, separate V1 and V2 code

nicholaspai:
This PR is large and has several parts. The diff is large but the files are much cleaner as a result with better separation between V1 and V2 logic. This will make it easier in the future to deprecate V1 entirely.

- Adds a V2 any<>any chain finalizer for DepositForBurn events. Can replace the V1 finalizer entirely for chains like HyperEVM, WorldChain and Linea which have always used CCTP V2 and where the HubPool doesn't send messages via CCTP.
- Forces the inventory client USDC CCTP adapters to use V2. There are no chains where we should be using V1 over V2 to rebalance inventory. This cleans up the adapter code quite a bit.
- Forces Solana USDC CCTP adapter to use V1. We only use V1 and we can hard switch over to V2 once appropriate
- Refactors `utils/CCTPUtils` entirely. This file is much cleaner and also removes some unneeded functionality like checking for HubPool as a sender and querying HubPool events (in this case, we can just query DepositForBurn events on V1 contracts)


src/clients/InventoryClient.ts
```diff
@@ -1387,6 +1387,19 @@ export class InventoryClient {
           if (pendingWithdrawalAmount.gte(maxL2WithdrawalVolume)) {
             return;
           }
+
+          // Before default withdrawing to L1, check if there are other chains for this L1 token that have
+          // underallocations, and then withdraw to those chains first.
+          const otherChainsToEvaluate = chainIds.filter(
```
nicholaspai, line 1393: Example of how we can use CCTP L2 <>L2 rebalancing in the inventory client. Unrelated to this PR but I accidentally committed this

src/finalizer/utils/cctp/anyToAny.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @notice This finalizes only CCTP V2 token relays that emit DepositForBurn events and where the sender or
+ * recipient addresses are in the list of addressesToFinalize passed in.
+ * @param logger
+ * @param spokePoolClient Origin chain spoke pool client
+ * @param addressesToFinalize Addresses sending or receiving Deposits that we want to finalize for.
+ * @returns
+ */
+export async function cctpV2Finalizer(
```
nicholaspai, line 29: I started this PR by implementing this file. Its a very simple file which I like, so I modeled the refactors of the `l1ToL2.ts` and `l2ToL1.ts` files after it

src/finalizer/index.ts
```diff
@@ -67,124 +68,160 @@ type FinalizationType = "l1->l2" | "l2->l1" | "l1<->l2";
...
   },
   [CHAIN_IDs.LENS]: {
     finalizeOnL1: [zkSyncFinalizer],
     finalizeOnL2: [],
+    finalizeOnAny: [],
   },
   [CHAIN_IDs.LINEA]: {
-    finalizeOnL1: [lineaL2ToL1Finalizer, cctpL2toL1Finalizer],
-    finalizeOnL2: [lineaL1ToL2Finalizer, cctpL1toL2Finalizer],
+    finalizeOnL1: [lineaL2ToL1Finalizer],
```
nicholaspai, line 111: For chains like Linea, WorldChain and HyperEVM which have never supported CCTP V1, we no longer need to add the V1 finalizers here

src/finalizer/index.ts
```diff
@@ -225,18 +262,27 @@ export async function finalize(
 
     // We should only finalize the direction that has been specified in
     // the finalization strategy.
-    const chainSpecificFinalizers: ChainFinalizer[] = [];
+    const chainSpecificFinalizers: { genericFinalizer: boolean; finalizer: ChainFinalizer | Finalizer }[] = [];
```
nicholaspai, line 265: I couldn't think of a more clever way to distinguish between the generic finalizer and the chain finalizer, unless I made the generic finalizer's function interface larger and take in params like `hubPoolClient` which would be unused in the function

src/finalizer/index.ts
```diff
@@ -263,16 +309,27 @@ export async function finalize(
...
-          hubPoolClient,
-          client,
-          spokePoolClients[hubChainId],
-          addressesToFinalize
-        );
+        let callData: (Multicall2Call | AugmentedTransaction)[], crossChainMessages: CrossChainMessage[];
+        if (isChainSpecificFinalizer(finalizer, genericFinalizer)) {
+          ({ callData, crossChainMessages } = await finalizer(
+            logger,
+            hubSigner,
```
nicholaspai, line 324: I guess we could change the chain-specific finalizer interface to take in `hubPoolClient`, `hubSigner` and `spokePoolClient` last so that we don't need this switch

src/finalizer/index.ts
```diff
@@ -297,17 +354,12 @@ export async function finalize(
...
   const multicall2Lookup = Object.fromEntries(
     await Promise.all(
-      uniq([
-        // We always want to include the hub chain in the finalization.
-        // since any L2 -> L1 transfers will be finalized on the hub chain.
-        hubChainId,
-        ...configuredChainIds,
-      ])
+      finalizerResponseTxns
+        .map(({ crossChainMessage }) => crossChainMessage.destinationChainId)
```
nicholaspai, line 358: This was a drive by optimization: we only need to create MultiSender contracts for destination chains where we have `crossChainMessage`'s for

src/finalizer/index.ts
```diff
@@ -263,16 +309,27 @@ export async function finalize(
...
+            hubPoolClient,
+            client,
+            spokePoolClients[hubChainId],
+            addressesToFinalize
+          ));
+        } else {
+          ({ callData, crossChainMessages } = await finalizer(logger, client, addressesToFinalize));
+        }
 
         callData.forEach((txn, idx) => {
```
bmzig, line 334: It might be nice to try and consolidate the data in `finalizerResponseTxns` after this code block. We don't have it now, but I think if we ever did any <-> any rebalancing, we would have multiple txns in here with the same destination chain ID. e.g. a mainnet -> Optimism and Arb -> Optimism finalization would produce two different entries in `finalizerResponseTxns`, which would in turn provide two separate multicall `aggregate` calls, when I believe this can be consolidated into 1. (I don't think we hit this because we still only do l1 -> l2 or l2 -> l1).

src/utils/CCTPUtils.ts
```diff
@@ -581,201 +778,40 @@ async function getCCTPMessagesWithStatus(
...
+): Promise<{ [sourceTxnHash: string]: CCTPV2APIGetAttestationResponse }> {
+  // For v2, we fetch an API response for every txn hash we have. API returns an array of both v1 and v2 attestations
+  const sourceDomainId = getCctpDomainForChainId(sourceChainId);
+  const isMainnet = utils.chainIsProd(sourceChainId);
+
+  // Circle rate limit is 35 requests / second. To avoid getting banned, batch calls into chunks with 1 second delay between chunks
+  // For v2, this is actually required because we don't know if message is finalized or not before hitting the API. Therefore as our
+  // CCTP v2 list of chains grows, we might require more than 35 calls here to fetch all attestations
+  const attestationResponses = {};
+  const chunkSize = 8;
```
bmzig, line 793: It's a nit, but this probably should be an env var, since it's probably going to be annoying to change this in case of rate-limiting in the future.

nicholaspai: fair, this is copied fromexisting code but i can add an env var override

PR #2638, opened 2025-10-02
improve(Relayer): Remove inventoryClient.update() from relayer.update()

nicholaspai:
The inventory client's `update()` function is very slow, so we should only call it once per run


src/relayer/index.ts
```diff
@@ -183,6 +183,7 @@ export async function runRebalancer(_logger: winston.Logger, baseSigner: Signer)
   const rebalancer = new Relayer(await baseSigner.getAddress(), logger, clients, config);
 
   try {
+    await inventoryClient.update(rebalancer.inventoryChainIds);
```
nicholaspai, line 186: Is there a better place to put this for the rebalancer?

PR #2641, opened 2025-10-02
fix: cap max CCTP transfer amount at 1M

bmzig:
I think we can go over the max transfer size when we want to rebalance > 1M using fast mode.


src/adapter/l2Bridges/UsdcCCTPBridge.ts
```diff
@@ -74,6 +74,7 @@ export class UsdcCCTPBridge extends BaseL2BridgeAdapter {
       if (optionalParams?.fastMode) {
         ({ maxFee, finalityThreshold } = await this._getCctpV2DepositForBurnMaxFee(amount));
       }
+      let adjustedAmount = amount.add(maxFee);
```
nicholaspai, line 77: Can be a const

nicholaspai (approved):

nicholaspai (approved):

nicholaspai (approved):

fusmanii (approved):

PR #2642, opened 2025-10-03
feat: add tests for aux native fees

grasphoper:


src/clients/ProfitClient.ts
```diff
@@ -454,15 +438,30 @@ export class ProfitClient {
...
+
+    // Normalise token amounts to USD terms.
+    const scaledLpFeeAmount = scaledInputAmount.mul(lpFeePct).div(fixedPoint);
+    const lpFeeUsd = scaledLpFeeAmount.mul(inputTokenPriceUsd).div(fixedPoint);
+
+    // Calculate gross relayer fee (excluding gas cost of fill).
+    const grossRelayerFeeUsd = inputAmountUsd.sub(outputAmountUsd).sub(lpFeeUsd).sub(auxiliaryNativeTokenCostUsd);
+    const grossRelayerFeePct = grossRelayerFeeUsd.gt(bnZero)
+      ? grossRelayerFeeUsd.mul(fixedPoint).div(inputAmountUsd)
+      : bnZero;
```
grasphoper, line 457: We cap `grossRelayerFeePct` at 0 min, and not reporting negative values, whereas we do report negative values `netRelayerFeePct` (which is also crucial for our min profitability caps because if we drop the `minRelayerProfitability` to e.g. 0, then we don't want to be filling relays with profitability -100%, which is giving away all of our money).

So should we report negatives for `grossRelayerFeePct`? We don't really use `grossRelayerFeePct` anywhere other than in logs, so it's not that mission critical _for now_.

Also, `grossRelayerFeePct` is from user POV, whereas `netRelayerFeePct` is from relayer POV. A bit confusing

PR #2648, opened 2025-10-05
fix: convert shortfall amount to l1 decimals in rebalancer

bmzig:


src/clients/InventoryClient.ts
```diff
@@ -960,10 +960,13 @@ export class InventoryClient {
...
 
   _getPossibleShortfallRebalances(l1Token: EvmAddress, chainId: number, l2Token: Address): Rebalance[] {
+    const { decimals: l1TokenDecimals } = getTokenInfo(l1Token, this.hubPoolClient.chainId);
+    const { decimals: l2TokenDecimals } = getTokenInfo(l2Token, chainId);
     // Order unfilled amounts from largest to smallest to prioritize larger shortfalls.
     const unfilledDepositAmounts = this.tokenClient
       .getUnfilledDepositAmounts(chainId, l2Token)
-      .sort(bnComparatorDescending);
+      .sort(bnComparatorDescending)
+      .map((unfilledAmount) => sdkUtils.ConvertDecimals(l2TokenDecimals, l1TokenDecimals)(unfilledAmount));
```
mrice32, line 969: Maybe a dumb question, but why convert the decimals _after_ the sort rather than before? For example, couldn't this cause mis-sorting on USDT/USDC on BSC?

bmzig: I think `getUnfilledDepositAmounts` is a bit of a misnomer, since the [amounts it returns](https://github.com/across-protocol/relayer/blob/c5d42ef1ed2e2985df2b9beda2a60c2d4cf040b1/src/clients/TokenClient.ts#L112) corresponds to the deposit's `ouptuAmount`, not `inputAmount`. I think this means having before/after the sort should do nothing, since all the amounts here are denominated in the same decimals (e.g. a deposit from OP to BSC or a deposit from Arb to BSC will both have unfilled amounts in BSC decimals. On the other end, a deposit from BSC to OP or Arb to OP will have unfilled amounts in OP decimals).

PR #2653, opened 2025-10-07
chore: Reorder finalizer chains

pxrl:
There are enough to justify sorting them alphabetically.


pxrl:
Dropped in favour of https://github.com/across-protocol/relayer/pull/2655.

PR #3120, opened 2026-03-26
improve: add l2 token splitter and fix Bridge API transfer tracking

bmzig:
This PR makes two changes:
- Adds a L2 token splitter bridge, which lets us withdraw two different tokens on L2 which both map to the same L1 token.
- Fixes the timing issue where we double rebalance via the Bridge API due to the rebalancer run frequency being faster than the time it takes for the Bridge API to update. This is done by checking for a transfer of matching amounts to the deposit address of every transfer which is still in the `awaiting_funds` state. If a transfer is found, then we can safely assume the transfer is outstanding, so we do not need to send another one. If no transfer is found, then we can assume that we still need to initiate another rebalance.


src/adapter/l2Bridges/TokenSplitterBridge.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  async getL2PendingWithdrawalAmount(
+    l2EventConfig: EventSearchConfig,
+    l1EventConfig: EventSearchConfig,
+    fromAddress: EvmAddress,
+    l2Token: EvmAddress
+  ): Promise<BigNumber> {
+    const [bridge1Pending, bridge2Pending] = await Promise.all([
+      this.bridge1.getL2PendingWithdrawalAmount(l2EventConfig, l1EventConfig, fromAddress, l2Token),
+      this.bridge2.getL2PendingWithdrawalAmount(l2EventConfig, l1EventConfig, fromAddress, l2Token),
```
nicholaspai: Do we need to convert decimals here to the same units?

nicholaspai (commented):
Nice pr

nicholaspai (commented):
The filterInitiatedTransfers function is pretty hard to understand can we simplify it?

bmzig:
@codex

src/utils/BridgeUtils.ts
```diff
@@ -127,6 +143,70 @@ export class BridgeApiClient {
...
+      const originToken = Object.entries(BRIDGE_API_DESTINATION_TOKEN_SYMBOLS).find(
+        ([, bridgeSymbol]) => bridgeSymbol === deposit.source_deposit_instructions.currency
+      );
+      assert(isDefined(originToken));
+      return getTokenInfo(toAddressType(originToken[0], this.srcNetworkId), this.srcNetworkId);
+    };
+
+    // Deduplicate queries by toAddress to avoid redundant event fetches.
+    const seen = new Set<string>();
+    const uniqueDeposits = deposits.filter((deposit) => {
```
nicholaspai, line 165: You can probably make this cleaner by doing a .map straight into a new Set<string>(...mapResult)

bmzig: Can I? The `seen` set is to help with filtering out all the unique deposits with different address/token pairs, but I need the unique deposits themselves and not the seen set.

src/utils/BridgeUtils.ts
```diff
@@ -127,6 +143,70 @@ export class BridgeApiClient {
...
+    deposits: BridgeResponse[],
+    fromAddress: Address,
+    eventConfig: EventSearchConfig,
+    originProvider: Provider
+  ): Promise<BridgeResponse[]> {
+    const getOriginTokenInfo = (deposit: BridgeResponse) => {
+      const originToken = Object.entries(BRIDGE_API_DESTINATION_TOKEN_SYMBOLS).find(
+        ([, bridgeSymbol]) => bridgeSymbol === deposit.source_deposit_instructions.currency
+      );
+      assert(isDefined(originToken));
```
pxrl: Might be nice to reference `bridgeSymbol` in an eventual assert message, to aid debugging if it's ever needed.

src/utils/BridgeUtils.ts
```diff
@@ -127,6 +143,70 @@ export class BridgeApiClient {
...
+      }
+      seen.add(seenKey);
+      return true;
+    });
+    // Map containing all origin chain transfers for a specific deposit address/token address combo.
+    const cachedTransfers: Record<string, Record<string, Log[]>> = {};
+    await mapAsync(uniqueDeposits, async (deposit) => {
+      const { address: originTokenAddress } = getOriginTokenInfo(deposit);
+      const toAddress = deposit.source_deposit_instructions.to_address;
+      const tokenContract = new Contract(originTokenAddress.toNative(), ERC20_ABI, originProvider);
```
pxrl: Possible to instantiate the contract outside of the loop and then just `connect` on the fly within the loop?

Instantiating contracts from JSON ABIs can be fairly expensive and wasteful; should be faster to make a copy from an existing Contract instance.

src/adapter/bridges/BridgeApi.ts
```diff
@@ -47,7 +51,7 @@ export class BridgeApi extends BaseBridgeAdapter {
     this.l2Bridge = new Contract(BRIDGE_API_DESTINATION_TOKENS[this.l2chainId], ERC20_ABI, l2SignerOrProvider);
 
     // We need to fetch some API configuration details from environment.
-    const { BRIDGE_API_BASE, BRIDGE_API_KEY, BRIDGE_CUSTOMER_ID } = process.env;
+    const { BRIDGE_API_BASE = "https://api.bridge.xyz", BRIDGE_API_KEY, BRIDGE_CUSTOMER_ID } = process.env;
```
pxrl, line 54: ooc, are these env-sourced because the overriding interface doesn't permit custom opts/args on instantiation?

pxrl (dismissed):

src/adapter/bridges/BridgeApi.ts
bmzig: Yes

src/utils/BridgeUtils.ts
bmzig: https://github.com/across-protocol/relayer/pull/3120/commits/8686080778722d1f1dffd5f4993041a1dc17585a

src/utils/BridgeUtils.ts
bmzig: https://github.com/across-protocol/relayer/pull/3120/commits/8686080778722d1f1dffd5f4993041a1dc17585a

pxrl (approved):

PR #3160, opened 2026-04-06
improve: Extend Binance rebalancer routes for WETH

nicholaspai:
## What changed
- extended the built-in rebalancer route generation in `src/rebalancer/buildRebalanceRoutes.ts` to add Binance `WETH <-> USDC` and `WETH <-> USDT` routes, but only when the `WETH` leg is on `MAINNET` and `WETH` is configured in `cumulativeTargetBalances`
- updated the Binance adapter to treat on-chain `WETH` as Binance `ETH`, enforce that intermediate bridge legs stay limited to bridgeable stablecoins, and skip the Binance spot swap leg when a route resolves to the same Binance coin
- added the mainnet-specific execution pieces needed for those routes: Atomic Depositor approval/deposit handling for `WETH -> Binance`, wrap-after-withdraw handling for `Binance -> WETH`, pending-balance accounting for finalized ETH withdrawals awaiting wrap, and the quote/fee/min-size precision fixes needed for mixed-asset routes
- refreshed rebalancer docs and expanded route-builder/Binance helper/quote/conversion coverage around the new WETH path and mixed-asset cumulative rebalancing behavior

## Supported WETH Routes For Now
- Only Binance `WETH <-> USDC` and `WETH <-> USDT` routes with the `WETH` leg on `MAINNET` are enabled right now. However this must be enabled at the configuration level in REBALANCER_CONFIG for this route to be executed in production.
- Reverse directions are supported, so examples include `MAINNET/WETH -> BASE/USDC`, `BASE/USDC -> MAINNET/WETH`, `MAINNET/WETH -> HYPEREVM/USDT`, and `HYPEREVM/USDT -> MAINNET/WETH`.

### DIFF
--- a/src/rebalancer/adapters/binance.ts
+++ b/src/rebalancer/adapters/binance.ts
@@ -779,6 +779,17 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
         truncate(withdrawalDetails.amount, binanceWithdrawalNetworkTokenInfo.decimals),
         binanceWithdrawalNetworkTokenInfo.decimals
       );
+      if (destinationToken === "WETH") {
+        this.logger.debug({
+          at: "BinanceStablecoinSwapAdapter.getPendingRebalances",
+          message:
+            `Withdrawal for order ${cloid} has finalized to ETH on ${binanceWithdrawalNetwork}, but the order remains pending until the ETH is wrapped into WETH. Keeping the destination-chain WETH credit until then`,
+          cloid: cloid,
+          orderDetails: orderDetails,
+          withdrawalDetails,
+        });
+        continue;
+      }
       this.logger.debug({
         at: "BinanceStablecoinSwapAdapter.getPendingRebalances",
         message: `Withdrawal for order ${cloid} has finalized, subtracting the order's virtual balance of ${withdrawAmountWei.toString()} from binance withdrawal network ${binanceWithdrawalNetwork}`,
@@ -1200,11 +1211,12 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
       atomicDepositorContracts.transferProxyAbi
     );
     const bridgeCalldata = transferProxy.interface.encodeFunctionData("transfer", [depositAddress]);
-    const binanceDepositNetwork = await this._getEntrypointNetwork(sourceChain, "WETH");
+    // @dev The AtomicWethDepositor today is only deployed to Ethereum and the only way to use it to deposit ETH
+    // into Binance is to use the bridgeCalldata as the whitelisted function selector mapped to chain ID 56.
     return this._submitTransaction({
       contract: atomicDepositor,
       method: "bridgeWeth",
-      args: [binanceDepositNetwork, amountToDeposit, amountToDeposit, bnZero, bridgeCalldata],
+      args: [CHAIN_IDs.BSC, amountToDeposit, amountToDeposit, bnZero, bridgeCalldata],
       chainId: sourceChain,
       nonMulticall: true,
       unpermissioned: false,
--- a/src/rebalancer/adapters/binance.ts
+++ b/src/rebalancer/adapters/binance.ts
@@ -1420,8 +1420,9 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
     }
     const priceData = await this._getLatestPrice(sourceToken, destinationToken, sourceChain, sourceAmount);
     const spotMarketMeta = await this._getSpotMarketMetaForRoute(sourceToken, destinationToken);
+    const destinationAmount = toBNWei(truncate(value, destinationTokenInfo.decimals), destinationTokenInfo.decimals);
     return convertBinanceRouteAmount({
-      amount: sourceAmount,
+      amount: destinationAmount,
       sourceTokenDecimals: sourceTokenInfo.decimals,
       destinationTokenDecimals: destinationTokenInfo.decimals,
       isBuy: spotMarketMeta.isBuy,
--- a/src/rebalancer/adapters/binance.ts
+++ b/src/rebalancer/adapters/binance.ts
@@ -746,7 +746,6 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
       const { unfinalizedWithdrawals, finalizedWithdrawals } = await this._getBinanceWithdrawals(
         destinationToken,
         binanceWithdrawalNetwork,
-
         isDefined(matchingFill) ? Math.floor(matchingFill.time / 1000) - 5 * 60 : getCurrentTime() - 6 * 60 * 60,
         // If there is a matching fill, then look up withdrawals after the fill time. If there is no fill because
         // its not a swap route, then use a conservative lookback period. If the withdrawal is older than this lookback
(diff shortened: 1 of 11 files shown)


src/rebalancer/adapters/binance.ts
```diff
@@ -43,32 +45,91 @@ interface SPOT_MARKET_META {
...
+      ? readableAmount / params.price
+      : readableAmount * params.price
+    : params.isBuy
+      ? readableAmount * params.price
+      : readableAmount / params.price;
+
+  return toBNWei(truncate(convertedAmount, outputDecimals), outputDecimals);
+}
+
+function resolveStepPrecision(stepSize: string): number {
```
nicholaspai: this function should be better commented its unclear what's going on here and what its intended purpose is

src/rebalancer/adapters/binance.ts
```diff
@@ -126,13 +173,14 @@ function resolveStepPrecision(stepSize: string): number {
 
 export class BinanceStablecoinSwapAdapter extends BaseAdapter {
   private binanceApiClient: Binance;
+  private tradeFeesPromise?: ReturnType<Binance["tradeFee"]>;
```
nicholaspai:
```suggestion
```

src/rebalancer/adapters/binance.ts
```diff
@@ -126,13 +173,14 @@ function resolveStepPrecision(stepSize: string): number {
...
   private exchangeInfoPromise?: ReturnType<Binance["exchangeInfo"]>;
   private orderBookPromiseBySymbol = new Map<string, Promise<Awaited<ReturnType<Binance["book"]>>>>();
   private orderBookSnapshotBySymbol = new Map<
     string,
     { fetchedAtMs: number; book: Awaited<ReturnType<Binance["book"]>> }
   >();
-  private tradeFeesPromise?: ReturnType<Binance["tradeFee"]>;
+  private spotMarketMetaPromiseByRoute = new Map<string, Promise<SPOT_MARKET_META>>();
```
nicholaspai, line 183:
```suggestion
  private tradeFeesPromise?: ReturnType<Binance["tradeFee"]>;
  private spotMarketMetaPromiseByRoute = new Map<string, Promise<SPOT_MARKET_META>>();
```

src/rebalancer/adapters/binance.ts
```diff
@@ -577,8 +746,11 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
       const { unfinalizedWithdrawals, finalizedWithdrawals } = await this._getBinanceWithdrawals(
         destinationToken,
         binanceWithdrawalNetwork,
-        Math.floor(matchingFill.time / 1000) - 5 * 60, // Floor this so we can grab the initiated withdrawal data whose
-        // ID we've already saved into Redis
+
```
nicholaspai:
```suggestion
```

src/rebalancer/adapters/binance.ts
```diff
@@ -1153,33 +1401,23 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
     const destinationTokenInfo = this._getTokenInfo(destinationToken, destinationChain);
-    const destinationAmountInSourcePrecision = this._getAmountConverter(
-      destinationChain,
-      destinationTokenInfo.address,
-      sourceChain,
-      sourceTokenInfo.address
-    )(destinationAmount);
-    if (sourceToken === destinationToken) {
-      return destinationAmountInSourcePrecision;
+    const sourceAmount = toBNWei(value, sourceTokenInfo.decimals);
```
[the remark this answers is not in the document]
nicholaspai: The `value` is assumed to just be a number that we want to convert into source precision, its not in any decimals--its just a number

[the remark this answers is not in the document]
nicholaspai: Current branch behavior here is intentional: `_convertNumberToSource` takes a human-readable destination-side number, not a raw destination-unit integer. The code normalizes that scalar into source precision before using it for the price lookup, and I added focused coverage for the mixed-decimal path in `test/BinanceAdapter.conversions.ts` so `WETH/USDC` sizing stays pinned to the current implementation. I don't think there is a remaining behavior bug on the current diff.

src/rebalancer/adapters/binance.ts
```diff
@@ -981,13 +1170,54 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
     return Number(coin.balance);
   }
 
   private async _getSymbol(sourceToken: string, destinationToken: string) {
-    this.exchangeInfoPromise ??= this.binanceApiClient.exchangeInfo();
+    const sourceAsset = resolveBinanceCoinSymbol(sourceToken);
+    const destinationAsset = resolveBinanceCoinSymbol(destinationToken);
     let exchangeInfo;
     try {
       exchangeInfo = await this.exchangeInfoPromise;
```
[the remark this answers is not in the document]
nicholaspai: This one is still real on the current diff. I haven't addressed it inside `#3160`; the exchange-info promise initialization/reset work is being handled separately rather than folded into the already-large WETH route PR. Leaving the thread open for that follow-up instead of claiming it is fixed here.

nicholaspai:
@codex

nicholaspai:
@codex

src/rebalancer/adapters/binance.ts
```diff
@@ -817,44 +975,37 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
-
-    let spreadPct = 0;
-    let spreadFee = bnZero;
-    if (routeRequiresSwap) {
-      const isBuy = spotMarketMeta.isBuy;
-      if (isBuy) {
-        // if is buy, the fee is positive if the price is over 1
-        spreadPct = latestPrice - 1;
-      } else {
-        spreadPct = 1 - latestPrice;
```
nicholaspai, line 849: This deleted logic no longer works with WETH where you can't simply subtract 1 from the price to figure out the "spread"

src/rebalancer/adapters/binance.ts
```diff
@@ -817,44 +975,37 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
-      }
-      spreadFee = toBNWei(spreadPct.toFixed(18), 18).mul(amountToTransfer).div(toBNWei(1, 18));
-    }
+    const spreadPct = latestPrice.slippagePct; // slippage is a percentage so we need to divide it by an additional
```
nicholaspai, line 992: Probably should rename this to slippagePct rather than spreadPct

src/rebalancer/adapters/binance.ts
```diff
@@ -462,6 +566,27 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
       }
 
+      const binanceWithdrawalNetworkTokenInfo = this._getTokenInfo(destinationToken, binanceWithdrawalNetwork);
+      const withdrawAmountWei = toBNWei(
+        truncate(withdrawalDetails.amount, binanceWithdrawalNetworkTokenInfo.decimals),
+        binanceWithdrawalNetworkTokenInfo.decimals
+      );
+
+      // Check if we need to wrap the withdrawal to WETH:
+      if (destinationToken === "WETH") {
```
bmzig, line 576: To me it seems pretty unscalable to have specific logic for WETH. I'd prefer it if you had "similar asset" logic and then dissimilar asset logic, so e.g. this and other cases could be `!isStablecoin(destinationToken)`

bmzig (approved):

src/rebalancer/adapters/binance.ts
nicholaspai: Hm I think that's right but WETH is a special case because of the wrapping/unwrapping so we should generalize it to be "native asset" support

PR #3183, opened 2026-04-08
improve: Final noImplicitAny fixes (4/4)

pxrl:
Introduce a new `resolveAcrossToken()` method to centralise all 
the checking that's required when resolving tokens via 
TOKEN_SYMBOLS_MAP.


src/utils/TokenUtils.ts
```diff
@@ -3,14 +3,36 @@ import { constants, utils, arch } from "@across-protocol/sdk";
...
 
+type TokenSymbolInfo = (typeof TOKEN_SYMBOLS_MAP)[keyof typeof TOKEN_SYMBOLS_MAP];
+
+// Safely index TOKEN_SYMBOLS_MAP with a string key, returning undefined for unknown symbols.
+// When chainId is supplied, returns the token address on that chain (or undefined).
+// When assertExists is true, throws if the token (or address) is not found.
+export function resolveTokenBySymbol(symbol: string): TokenSymbolInfo | undefined;
+export function resolveTokenBySymbol(symbol: string, chainId: number): string | undefined;
+export function resolveTokenBySymbol(symbol: string, chainId: number, assertExists: true): string;
+export function resolveTokenBySymbol(
```
nicholaspai: It seems redundant that we export this function AND getTokenInfoFromSymbol, and potentially confusing to users. I'm hoping we can have fewer of these getTokenFromAddress/Symbol/getRemoteToken, etc. functions eventually

pxrl: Yeah, agree. I think most of this belongs upstream, rather than directly in the relayer repo. For now I've renamed to something slightly less confusing. Hopefully that's enough to get this in, because it's otherwise tricky to keep up with all the changes that are going in. With this PR in then we can ratchet away the noImplicitAny escape hatch.

pxrl:
@codex review

nicholaspai:
~Isn't this redundant with getTokenInfoFromSymbol~

Never mind I see the new token utils function is upstream of getTokenInfoFromSymbol

nicholaspai (approved):

PR #3203, opened 2026-04-09
improve(Refiller): use ensureConfirmations: true with submitTransaction

nicholaspai:
No need to .wait() when using submitTransactions with ensureConfirmations: true


nicholaspai:
@codex

PR #3204, opened 2026-04-09
improve(TransactionClient): Avoid using txnClient.submit() directly

nicholaspai:
The TransactionClient.submit() is not designed for general use because it can return undefined rather than throw if a txn succeeds which is usually not the caller's intuition. `submitTransaction` should be used instead.

### DIFF
--- a/src/utils/TransactionUtils.ts
+++ b/src/utils/TransactionUtils.ts
@@ -214,7 +214,7 @@ export async function submitTransaction(
   return response[0];
 }
 
-export async function dispatchTransaction(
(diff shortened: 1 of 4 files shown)


src/utils/TransactionUtils.ts
```diff
@@ -214,7 +214,7 @@ export async function submitTransaction(
   return response[0];
 }
 
-export async function dispatchTransaction(
+async function dispatchTransaction(
```
nicholaspai, line 217: driveby, this function doesn't need to be exported

PR #3209, opened 2026-04-10
refactor(InventoryClient): [codex] Refactor refund chain ranking pipeline

nicholaspai:
## Summary

Refactor `InventoryClient.determineRefundChainId()` to simplify the refund-chain ranking pipeline without changing repayment behavior.

- derive slow-withdrawal candidates directly from `possibleChains`
- make origin and destination standard-candidate eligibility explicit with local booleans
- use a final ordered `Set` dedupe to preserve slow-withdrawal priority while avoiding duplicate origin/destination candidates
- collapse repayment-token selection into a single expression inside the evaluation loop

## Why

The function was still carrying duplicate filtering and ranking logic in the candidate-building block. This change trims that control flow down further while preserving the existing ordering and fallback semantics.

## Impact

This keeps the repayment-chain selection logic easier to reason about and slightly smaller, with no interface or config changes.

## Validation

- `npm exec eslint -- src/clients/InventoryClient.ts`
- `npm test -- test/InventoryClient.RefundChain.ts`

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -734,8 +734,7 @@ export class InventoryClient {
     const standardCandidateChains = [...new Set(standardChainOrder)].filter((chainId) =>
       chainId === originChainId ? canEvaluateOrigin : canEvaluateDestination
     );
-    const candidateChains = [...new Set([...slowWithdrawalCandidateChains, ...standardCandidateChains])];
-    const rankedChains = candidateChains;
+    const rankedChains = [...new Set([...slowWithdrawalCandidateChains, ...standardCandidateChains])];
 
     const eligibleChains: number[] = [];
     // At this point, all ranked chains have defined token configs or are destination chains that can be accepted
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -38,6 +38,7 @@ import {
   max,
   getLatestRunningBalances,
   getInventoryBalanceContributorTokens,
+  dedupArray,
 } from "../utils";
 import { BundleDataApproxClient, BundleDataState } from "./BundleDataApproxClient";
 import { HubPoolClient, TokenClient, TransactionClient } from ".";
@@ -693,7 +694,7 @@ export class InventoryClient {
     const inputAmountInL1TokenDecimals = sdkUtils.ConvertDecimals(inputTokenDecimals, l1TokenDecimals)(inputAmount);
     const totalRefundsPerChain: { [chainId: number]: BigNumber } = Object.fromEntries(
       this.chainIdList.map((chainId) => [chainId, this.getUpcomingRefunds(chainId, l1Token, this.relayer)])
-    ) as { [chainId: number]: BigNumber };
+    );
     const destinationTokensAreEquivalent = this.hubPoolClient.areTokensEquivalent(
       inputToken,
       originChainId,
@@ -718,7 +719,7 @@ export class InventoryClient {
 
     const slowWithdrawalCandidateChains = possibleChains
       .filter(
-        (chainId) => slowWithdrawalRepaymentChainSet.has(chainId) && (excessRunningBalancePcts[chainId] ?? bnZero).gt(0)
+        (chainId) => slowWithdrawalRepaymentChainSet.has(chainId) && (excessRunningBalancePcts[chainId] ?? bnZero).gt(bnZero)
       )
       .sort((chainIdx, chainIdy) =>
         bnComparatorDescending(excessRunningBalancePcts[chainIdx], excessRunningBalancePcts[chainIdy])
@@ -730,10 +731,10 @@ export class InventoryClient {
       this._l1TokenEnabledForChain(l1Token, originChainId) && (prioritizeOrigin || originChainId !== hubChainId);
     const canEvaluateDestination =
       this.canTakeDestinationChainRepayment(deposit) && this._l1TokenEnabledForChain(l1Token, destinationChainId);
-    const standardCandidateChains = [...new Set(standardChainOrder)].filter((chainId) =>
+    const standardCandidateChains = [...standardChainOrder].filter((chainId) =>
       chainId === originChainId ? canEvaluateOrigin : canEvaluateDestination
     );
-    const rankedChains = [...new Set([...slowWithdrawalCandidateChains, ...standardCandidateChains])];
+    const rankedChains = dedupArray([...slowWithdrawalCandidateChains, ...standardCandidateChains]);
 
     const eligibleChains: number[] = [];
     // At this point, all ranked chains have defined token configs or are destination chains that can be accepted
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -528,13 +528,8 @@ export class InventoryClient {
     }
 
     if (this.isInventoryManagementEnabled()) {
-      // @dev Because the `deposit` is a for an input token that have already filtered then we shouldn't expect
-      // to see any undefined return values from `getL1TokenAddress()`.
-      const l1Token = this.getL1TokenAddress(inputToken, originChainId);
-      assert(
-        isDefined(l1Token),
-        `InventoryClient#getPossibleRepaymentChainIds: No L1 token found for input token ${inputToken.toNative()} on origin chain ${originChainId}`
-      );
+      // @dev Because the `deposit` is for an input token that has already been filtered, this mapping should exist.
+      const l1Token = this.getRequiredL1TokenAddress(inputToken, originChainId);
       this.getSlowWithdrawalRepaymentChains(l1Token).forEach((chainId) => {
         if (this.hubPoolClient.l2TokenEnabledForL1Token(l1Token, chainId)) {
           chainIds.add(chainId);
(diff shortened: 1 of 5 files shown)


src/clients/InventoryClient.ts
```diff
@@ -700,93 +692,64 @@ export class InventoryClient {
...
-      chainsToEvaluate.push(originChainId);
-    }
+    const possibleChainSet = new Set(possibleChains);
+    const standardChainOrder = prioritizeOrigin
+      ? [originChainId, destinationChainId]
+      : [destinationChainId, originChainId];
+    const canEvaluateOrigin =
+      this._l1TokenEnabledForChain(l1Token, originChainId) && (prioritizeOrigin || originChainId !== hubChainId);
+    const canEvaluateDestination =
+      possibleChainSet.has(destinationChainId) && this._l1TokenEnabledForChain(l1Token, destinationChainId);
```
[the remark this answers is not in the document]
nicholaspai: Fixed in 0aedd20d. I restored the destination standard-candidate gate to use `canTakeDestinationChainRepayment(deposit)` instead of deriving it from `possibleChains`, so destination being present only via slow-withdrawal or hub-fallback paths no longer makes it eligible as a destination candidate. I also added a regression test that forces the destination into `getPossibleRepaymentChainIds()` through the slow-withdrawal path while its direct destination route is unsupported, and `determineRefundChainId()` now still returns `[POLYGON, MAINNET]` rather than ranking destination.

nicholaspai:
@codex

src/clients/InventoryClient.ts
```diff
@@ -566,6 +561,15 @@ export class InventoryClient {
     }
   }
 
+  private getRequiredL1TokenAddress(originToken: Address, originChainId: number): EvmAddress {
```
bmzig, line 565: Why make this new function and not just use `this.getL1TokenAddress`?

src/clients/InventoryClient.ts
```diff
@@ -700,93 +692,63 @@ export class InventoryClient {
...
+      : [destinationChainId, originChainId];
+    const canEvaluateOrigin =
+      this._l1TokenEnabledForChain(l1Token, originChainId) && (prioritizeOrigin || originChainId !== hubChainId);
+    const canEvaluateDestination =
+      this.canTakeDestinationChainRepayment(deposit) && this._l1TokenEnabledForChain(l1Token, destinationChainId);
+    const standardCandidateChains = [...new Set(standardChainOrder)].filter((chainId) =>
+      chainId === originChainId ? canEvaluateOrigin : canEvaluateDestination
+    );
+    const candidateChains = [...new Set([...slowWithdrawalCandidateChains, ...standardCandidateChains])];
+    const rankedChains = candidateChains;
```
bmzig: Why the reinstantiation?

src/clients/InventoryClient.ts
nicholaspai: The idea is that we repeat the `this.getL1TokenAddress(); assert(isDefined(l1Token)` pattern in a couple places so I wanted to save some lines

src/clients/InventoryClient.ts
```diff
@@ -699,93 +691,62 @@ export class InventoryClient {
...
-    // We need to calculate the latest running balance for each optimistic rollup chain.
-    // We'll add the last proposed running balance plus new deposits and refunds.
-    if (!forceOriginRepayment && this.prioritizeLpUtilization) {
-      const excessRunningBalancePcts = await this.getExcessRunningBalancePcts(
-        l1Token,
-        inputAmountInL1TokenDecimals,
-        this.getSlowWithdrawalRepaymentChains(l1Token)
+    const totalRefundsPerChain: { [chainId: number]: BigNumber } = Object.fromEntries(
+      this.chainIdList.map((chainId) => [chainId, this.getUpcomingRefunds(chainId, l1Token, this.relayer)])
+    ) as { [chainId: number]: BigNumber };
```
pxrl: Is this type assertion needed / can it be avoided?

src/clients/InventoryClient.ts
```diff
@@ -699,93 +691,62 @@ export class InventoryClient {
...
+    // @dev The async call to `getExcessRunningBalancePcts` should be very fast compared to upcoming refund lookups,
+    // so we choose not to compute them in parallel.
+    const excessRunningBalancePcts =
+      !forceOriginRepayment && this.prioritizeLpUtilization
+        ? await this.getExcessRunningBalancePcts(l1Token, inputAmountInL1TokenDecimals, slowWithdrawalRepaymentChains)
+        : {};
+
+    const slowWithdrawalCandidateChains = possibleChains
+      .filter(
+        (chainId) => slowWithdrawalRepaymentChainSet.has(chainId) && (excessRunningBalancePcts[chainId] ?? bnZero).gt(0)
```
pxrl:
```suggestion
        (chainId) => slowWithdrawalRepaymentChainSet.has(chainId) && (excessRunningBalancePcts[chainId] ?? bnZero).gt(bnZero)
```

src/clients/InventoryClient.ts
```diff
@@ -699,93 +691,62 @@ export class InventoryClient {
...
-      chainsToEvaluate.push(originChainId);
-    }
+    const standardChainOrder = prioritizeOrigin
+      ? [originChainId, destinationChainId]
+      : [destinationChainId, originChainId];
+    const canEvaluateOrigin =
+      this._l1TokenEnabledForChain(l1Token, originChainId) && (prioritizeOrigin || originChainId !== hubChainId);
+    const canEvaluateDestination =
+      this.canTakeDestinationChainRepayment(deposit) && this._l1TokenEnabledForChain(l1Token, destinationChainId);
+    const standardCandidateChains = [...new Set(standardChainOrder)].filter((chainId) =>
```
pxrl: I am a bit surprised we dedup this though, given it only contains the pair of [origin, destination].

```suggestion
    const standardCandidateChains = dedupArray(standardChainOrder).filter((chainId) =>
```

src/clients/InventoryClient.ts
```diff
@@ -699,93 +691,62 @@ export class InventoryClient {
...
+      chainId === originChainId ? canEvaluateOrigin : canEvaluateDestination
+    );
+    const rankedChains = [...new Set([...slowWithdrawalCandidateChains, ...standardCandidateChains])];
```
pxrl:
```suggestion
    const rankedChains = dedupArray([...slowWithdrawalCandidateChains, ...standardCandidateChains]);
```

src/clients/InventoryClient.ts
```diff
@@ -794,25 +755,11 @@ export class InventoryClient {
...
       const chainVirtualBalance = this.getBalanceOnChain(chainId, l1Token, repaymentToken);
       const chainVirtualBalanceWithShortfall = chainVirtualBalance.sub(chainShortfall);
-      // @dev Do not subtract outputAmount from virtual balance if output token and input token are not equivalent.
-      // This is possible when the output token is USDC.e and the input token is USDC which would still cause
-      // validateOutputToken() to return true above.
-      let chainVirtualBalanceWithShortfallPostRelay =
-        chainId === destinationChainId &&
-        this.hubPoolClient.areTokensEquivalent(inputToken, originChainId, outputToken, destinationChainId)
-          ? chainVirtualBalanceWithShortfall
-          : chainVirtualBalanceWithShortfall.add(inputAmountInL1TokenDecimals);
```
pxrl, line 805: This feels like it would have been an issue with USDC and USDT on BSC.

src/clients/InventoryClient.ts
```diff
@@ -794,25 +755,11 @@ export class InventoryClient {
...
-        totalRefundsPerChain[chainId] ?? bnZero
-      );
-      // To correctly compute the allocation % for this destination chain, we need to add all upcoming refunds for the
-      // equivalents of l1Token on all chains.
-      const cumulativeVirtualBalancePostRefunds = this.getCumulativeBalanceWithApproximateUpcomingRefunds(l1Token);
-
-      // Compute what the balance will be on the target chain, considering this relay and the finalization of the
-      // transfers that are currently flowing through the canonical bridge.
+      const inputAmountAddedPostRelay =
+        chainId === destinationChainId && destinationTokensAreEquivalent ? bnZero : inputAmountInL1TokenDecimals;
```
pxrl, line 763: Do we need to normalise for cases where token decimals differ? ATM it's only for BSC and we indirectly avoid it because we're forced to take origin repayment there, but we should still be robust to it (nb. this seems like a pre-existing issue).

src/clients/InventoryClient.ts
nicholaspai: thats fair ill remove the dedup

test/constants.ts
pxrl, line 1: @nicholaspai Is this + other test-related changes (separate from InventoryClient.RefundChain) related to this PR?

src/clients/InventoryClient.ts
pxrl: Resolved - not an issue.

test/constants.ts
pxrl: Those changes address test fragility; no problem bundling them here.

pxrl (approved):

PR #3213, opened 2026-04-13
fix: remove commas from request amounts to Bridge API

bmzig:
The formatter is adding commas to amounts >= 1000, which is causing 400 errors when posting to the Bridge API. We need to send amounts which do not have the commas.


pxrl (approved):
Feels like we could do this in a different way, but this is fine for now.

PR #3220, opened 2026-04-14
refactor(svm): Drop direct anchor dependency

pxrl:
Claude identifies that the direct dependency on Anchor can be dropped by instead using Kit funcitonalities. Anchor is still required as a transitive dependency by upstream functions, but this helps to simplify the codebase.


pxrl:
nb. this is currently untested.

PR #3223, opened 2026-04-14
improve(rebalancer): improve rebalance info logging

nicholaspai:
## What changed
This improves cumulative rebalancer logging so actual rebalance submissions emit a richer `info` log with the key transfer and balance context operators need when a rebalance is initialized.

## Why
Rebalance initialization is an operator-relevant event, but the existing log output was sparse and harder to interpret quickly. This change keeps the detailed route-evaluation logs at `debug` and adds a more readable, higher-signal `info` log only when a rebalance is actually being sent.

## Impact
When `sendingTransactionsEnabled` is true, operators now get an `info` log that includes the adapter, transfer amount, expected fees, cumulative balances, current on-chain balances, and the remaining deficit after the rebalance is initialized. Dry runs still stay at `debug`.

## Validation
- `yarn build`

### DIFF
--- a/src/rebalancer/clients/CumulativeBalanceRebalancerClient.ts
+++ b/src/rebalancer/clients/CumulativeBalanceRebalancerClient.ts
@@ -3,6 +3,7 @@ import {
   BigNumber,
   bnZero,
   ConvertDecimals,
+  fromWei,
   getNetworkName,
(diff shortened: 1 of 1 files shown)


nicholaspai:
@codex

PR #3231, opened 2026-04-14
improve: Clear relayer pending txn promises after failures

nicholaspai:
## What changed
- wraps `pendingTxnHashes[chainId]` cleanup in `finally`
- ensures a rejected fill execution does not leave a permanently cached rejected promise for that chain

## Why
`Relayer.executeFills()` memoizes an in-flight per-chain transaction queue. If queue execution rejects, the code never deletes the cached promise, so later runs keep thinking the chain still has pending fills and skip execution forever.

## Impact
A transient fill-execution failure no longer wedges relaying on that chain until process restart.

## Validation
- `yarn build`

### DIFF
--- a/src/relayer/Relayer.ts
+++ b/src/relayer/Relayer.ts
@@ -893,10 +893,11 @@ export class Relayer {
     pendingTxnHashes[chainId] = (async () => {
       return (await fillExecutorClient.executeTxnQueue(chainId, simulate)).map((response) => response.hash);
     })();
-    const txnReceipts = await pendingTxnHashes[chainId];
-    delete pendingTxnHashes[chainId];
-
-    return txnReceipts;
+    try {
+      return await pendingTxnHashes[chainId];
+    } finally {
+      delete pendingTxnHashes[chainId];
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/relayer/Relayer.ts
```diff
@@ -893,10 +893,11 @@ export class Relayer {
     pendingTxnHashes[chainId] = (async () => {
       return (await fillExecutorClient.executeTxnQueue(chainId, simulate)).map((response) => response.hash);
     })();
-    const txnReceipts = await pendingTxnHashes[chainId];
-    delete pendingTxnHashes[chainId];
-
-    return txnReceipts;
+    try {
+      return await pendingTxnHashes[chainId];
```
pxrl, line 897: @nicholaspai I think the agent is misfiring on this one. If we make this change, the relayer can only actively fill on one chain at a time. It'd totally kill our performance.

pxrl: Reading this one again with fresh eyes - I misunderstood the change. It makes sense to me now - happy for this to go in :+1:

pxrl (approved):

PR #3232, opened 2026-04-14
improve: Reset Binance bootstrap promises after failures

nicholaspai:
## What changed
- clears the global Binance secret-key lookup promise when key retrieval rejects
- clears cached Binance bridge client bootstrap promises on failure in both EVM and L2 Binance bridge adapters
- preserves the existing success-path caching behavior

## Why
These call sites memoize one bootstrap promise per process or bridge instance. If the first attempt fails, every later caller reuses the same rejected promise and cannot recover without restarting.

## Impact
Transient Binance bootstrap failures no longer permanently poison secret-key retrieval or Binance bridge client initialization.

## Validation
- `yarn build`


bmzig:
I don't understand this one. If the key retrieval fails, then we throw an error and the program crashes, which also happens here? So there's no utility to adding retries?

PR #3238, opened 2026-04-15
chore: Update GaslessRelayer logging with processing time

dijanin-brat:


src/gasless/GaslessRelayer.ts
```diff
@@ -487,9 +491,13 @@ export class GaslessRelayer {
           authorizer,
           nonce,
           requestId: depositMessage.requestId,
+          submittedAt: depositMessage.submittedAt,
+          secondsSinceSubmitted,
+          depositProcessingSlaStatus: secondsSinceSubmitted > sla ? "WARRNING_SLA_VIOLATION" : "NORMAL",
```
bora-gjura:
```suggestion
          depositProcessingSlaStatus: secondsSinceSubmitted > sla ? "WARNING_SLA_VIOLATION" : "NORMAL",
```

src/gasless/GaslessRelayerConfig.ts
```diff
@@ -69,5 +74,8 @@ export class GaslessRelayerConfig extends CommonConfig {
         new Set(outputSymbols),
       ])
     );
+
+    this.depositProcessingSlaSeconds = Number(GASLESS_DEPOSIT_PROCESSING_SLA_SECONDS ?? 60);
```
bora-gjura: (will confirm w/ Ryan, lg for now!)

src/gasless/GaslessRelayer.ts
```diff
@@ -697,7 +699,8 @@ export class GaslessRelayer {
       delete this.fillLock[fillKey];
       const tEnd = performance.now();
       const delta = (tEnd - tStart) / 1000;
-      log("info", `Processed ${origin} depositId ${depositId} in ${delta} seconds.`);
+      const submitToProcessedTime = (tEnd - new Date(depositMessage.submittedAt).getTime()) / 1000;
+      log("info", `Processed ${origin} depositId ${depositId} in ${delta} seconds.`, { submitToProcessedTime });
```
pxrl: Naming is hard, but I am wondering about `tSubmitDelta` or something?

Note also that we will emit this log when we are re-processing an existing deposit (i.e. if it was already deposited and/or filled), in which case the timestamp won't be reliable. That might reduce the utility of this.

src/gasless/GaslessRelayer.ts
```diff
@@ -475,6 +475,7 @@ export class GaslessRelayer {
       const fillKey = `${authorizer}:${originChainId}`;
 
       const at = "GaslessRelayer#evaluateApiSignatures";
+
```
pxrl:
```suggestion
```

dijanin-brat: [8c4501c](https://github.com/across-protocol/relayer/pull/3238/commits/8c4501ce937420107af7f3779dde4fc34c0a7f1e)

src/gasless/GaslessRelayer.ts
dijanin-brat: [8c4501c](https://github.com/across-protocol/relayer/pull/3238/commits/8c4501ce937420107af7f3779dde4fc34c0a7f1e)

Yes, it will log even if the deposit is processed already, tho this can be the final trigger for us to refactor this and discard all deposits that have already been processed when new bot instance start?

PR #3241, opened 2026-04-15
feat: add direct Binance same-asset stablecoin rebalance routes

nicholaspai:
## What changed
- extracted the built-in production route generation into `src/rebalancer/buildRebalanceRoutes.ts`
- added direct Binance same-asset routes for `USDC -> USDC` and `USDT -> USDT` when both sides are direct Binance networks for that asset
- kept the existing bridge-backed same-asset routes (`USDC` via CCTP and `USDT` via OFT) and the existing `USDC <-> USDT` swap routes on Binance and Hyperliquid
- updated the Binance adapter to skip the spot swap leg for same-asset routes and move directly from deposit to withdrawal
- taught pending-withdrawal accounting to handle same-asset routes that never produce a Binance fill record
- added focused coverage for route generation and same-asset stablecoin withdrawal fee accounting
- documented the built-in stablecoin route families in the rebalancer README

## Why
The route set previously only let Binance compete on `USDC <-> USDT` swap routes. Same-asset stablecoin rebalances still had to use CCTP or OFT even when both chains were direct Binance networks and a deposit-then-withdraw path was available.

## Impact
- the rebalancer can now compare direct Binance same-asset stablecoin paths against existing bridge-backed routes
- same-asset Binance routes avoid placing an unnecessary spot order and instead finalize as deposit-plus-withdraw flows
- route generation lives in a dedicated module with direct unit coverage

## Validation
- `yarn test --bail test/BinanceAdapter.conversions.ts test/RebalancerClient.buildRebalanceRoutes.ts`

### DIFF
--- a/src/rebalancer/adapters/binance.ts
+++ b/src/rebalancer/adapters/binance.ts
@@ -19,6 +19,7 @@ import {
   getBinanceApiClient,
   getBinanceTransactionTypeKey,
   getBinanceWithdrawals,
+  getCurrentTime,
   getNetworkName,
   getProvider,
   isDefined,
@@ -69,6 +70,10 @@ export function isTerminalBinanceWithdrawal(status?: number): boolean {
     default:
       return false;
   }
+}
+
+function supportsBinanceIntermediateBridgeToken(token: string): boolean {
+  return token === "USDC" || token === "USDT";
 }
 
 export function deriveBinanceSpotMarketMeta(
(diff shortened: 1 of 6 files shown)


src/rebalancer/adapters/binance.ts
```diff
@@ -794,19 +830,22 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
     let spreadPct = 0;
-    if (isBuy) {
-      // if is buy, the fee is positive if the price is over 1
-      spreadPct = latestPrice - 1;
-    } else {
-      spreadPct = 1 - latestPrice;
+    let spreadFee = bnZero;
+    if (routeRequiresSwap) {
+      const isBuy = spotMarketMeta.isBuy;
+      if (isBuy) {
```
bmzig, line 845: Might be cleaner to just make this a ternary?

bmzig (dismissed):

src/rebalancer/adapters/binance.ts
nicholaspai: i'm gonna replace this in a follow up PR so ill leave it for now

bmzig (approved):

PR #3246, opened 2026-04-16
fix: Gasless Initialize function

dijanin-brat:
Updating GaslessRelayer initialize function so it starts using messageState that was introduced in refactored handler.
This update will only discard (set the state of the message to FILLED) deposits that were fully processed. Eventually, we want to refactor the gaslessRelayer in a way where we can set the correct state for every deposit (i.e. if deposit message already have on-chain deposit but no fill, we should be able to set deposit state so it doesn't need to pass through all the checks and states from scratch, but go directly to the filling, etc.)

### DIFF
--- a/src/gasless/GaslessRelayer.ts
+++ b/src/gasless/GaslessRelayer.ts
@@ -47,7 +47,6 @@ import {
   APIGaslessDepositResponse,
   DepositWithBlock,
   FillStatus,
-  FillWithBlock,
   GaslessDepositMessage,
   RelayData,
 } from "../interfaces";
@@ -273,7 +272,7 @@ export class GaslessRelayer {
         );
         for (const filledRelay of destinationFilledRelayEvents) {
           const fill = unpackFillEvent(spreadEventWithBlockNumber(filledRelay), destinationChainId);
-          observedFills.add(this._getFilledRelayKey(fill));
+          observedFills.add(this._getFilledRelayKey(fill.originChainId, fill.depositId.toString()));
         }
       }),
     ]);
@@ -406,7 +405,7 @@ export class GaslessRelayer {
         continue;
       }
 
-      const fillKey = this._getFilledRelayKey({ originChainId, depositId: toBN(depositId) });
+      const fillKey = this._getFilledRelayKey(originChainId, depositId);
       const fillSeen = this.observedFills[destinationChainId]?.has(fillKey) ?? false;
       if (fillSeen) {
         this._setState(depositKey, MessageState.FILLED);
@@ -1007,8 +1006,7 @@ export class GaslessRelayer {
    * @dev We key on the origin chain and depositId only since this is what uniquely identifies a deposit on an origin chain for a specific user (the only way to have a collision here with
    * a valid, unfilled deposit is by finding a collision in keccak).
    */
-  private _getFilledRelayKey(filledRelay: Pick<FillWithBlock, "originChainId" | "depositId">): string {
-    const { originChainId, depositId } = filledRelay;
+  private _getFilledRelayKey(originChainId: number, depositId: string): string {
     return `${originChainId}:${depositId}`;
   }
 }
--- a/src/gasless/GaslessRelayer.ts
+++ b/src/gasless/GaslessRelayer.ts
@@ -68,12 +68,6 @@ import {
   validateDeposit,
 } from "../utils/GaslessUtils";
 
-type GaslessRelayerUpdate = {
-  observedFundsDeposited: {
-    [chainId: number]: Omit<DepositWithBlock, "fromLiteChain" | "toLiteChain" | "quoteBlockNumber">[];
-  };
-  observedFilledRelay: { [chainId: number]: FillWithBlock[] };
-};
 const DEPOSIT_EVENT = "FundsDeposited";
 
 export enum MessageState {
@@ -187,9 +181,6 @@ export class GaslessRelayer {
       this.abortController.abort();
     });
 
-    // For the first runthrough, we want to specifically check the API for nonces which have corresponding deposits but no corresponding fills, and then
-    // submit fills for those deposits. This is because this scenario may happen as an edge case when a prior relayer process is killed in the middle of
-    // a deposit/fill flow.
     this.logger.debug({
       at: "GaslessRelayer#initialize",
       message: "Querying gasless API for initial messages",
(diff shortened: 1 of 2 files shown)


dijanin-brat:
@codex please review

src/gasless/GaslessRelayer.ts
```diff
@@ -449,6 +412,43 @@ export class GaslessRelayer {
...
+      const isCctp = this._isCctpDeposit(originChainId, spokePool);
+      if (isSwap || isCctp) {
+        this._setState(depositKey, MessageState.FILLED);
+        markedFilledCount++;
+        continue;
+      }
+
+      const fillKey = this._getFilledRelayKey({ originChainId, depositId: toBN(depositId) });
+      const fillSeen = this.observedFills[destinationChainId]?.has(fillKey) ?? false;
+      if (fillSeen) {
```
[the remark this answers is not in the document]
dijanin-brat: This is intentional for now, we will need to refactor the GaslessRelayer to cover this as well eventually.

dijanin-brat:
@codex re-review

dijanin-brat:
@codex review

src/gasless/GaslessRelayer.ts
```diff
@@ -449,6 +378,43 @@ export class GaslessRelayer {
...
+      if (isSwap || isCctp) {
+        this._setState(depositKey, MessageState.FILLED);
+        markedFilledCount++;
+        continue;
+      }
+
+      const fillKey = this._getFilledRelayKey(originChainId, depositId);
+      const fillSeen = this.observedFills[destinationChainId]?.has(fillKey) ?? false;
+      if (fillSeen) {
+        this._setState(depositKey, MessageState.FILLED);
```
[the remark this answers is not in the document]
dijanin-brat: This is intentional for now, we will need to refactor the GaslessRelayer to cover this as well eventually.

src/gasless/GaslessRelayer.ts
```diff
@@ -449,6 +378,43 @@ export class GaslessRelayer {
...
+      const { originChainId, depositId, spokePool, depositFlowType } = depositMessage;
+      const { destinationChainId, inputToken } = extractGaslessDepositFields(depositMessage);
+      const depositKey = this._getDepositKey(inputToken.toNative(), originChainId, depositId);
+
+      const depositSeen = this.observedDeposits[originChainId]?.has(depositKey) ?? false;
+      if (!depositSeen) {
+        continue;
+      }
+
+      const isSwap = depositFlowType === "swapAndBridge";
```
bmzig, line 398: Just to be sure, we can mark a `swapAndBridge` as filled since all swapAndBridge transactions are mint/burn transactions?

dijanin-brat: for now, yes. We added `swapAndBridge` just with CCTP at the moment. I think there will be feature with intents as well, but currently we support only CCTP.
Our whole code in GaslessRelayer assumes that as well.

src/gasless/GaslessRelayer.ts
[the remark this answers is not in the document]
bmzig: Can we actually do 
```
if (fillSeen) { ... } 
else { this._setState(depositKey, MessageState.DEPOSIT_CONFIRMED }`
```
or does that break the handler?

bmzig (dismissed):
Looks good

src/gasless/GaslessRelayer.ts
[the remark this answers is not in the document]
dijanin-brat: unfortunately, not right now because of the instantFill logic...
My next idea is to set correct state for all initial deposits so we don't need to process them from the INITIAL state but from real state where that deposit ended.

src/gasless/GaslessRelayer.ts
```diff
@@ -280,78 +228,54 @@ export class GaslessRelayer {
...
+        for (const event of originFundsDepositedEvents) {
+          const deposit = unpackDepositEvent(spreadEventWithBlockNumber(event), originChainId);
+          const apiMessage = apiMessages.find(({ depositId }) => deposit.depositId.toString() === depositId);
+          if (!isDefined(apiMessage)) {
+            continue;
+          }
+          observedDeposits.add(
+            this._getDepositKey(deposit.inputToken.toNative(), originChainId, deposit.depositId.toString())
           );
+        }
```
pxrl: Minor style/safety - mixing a for loop and `Array.forEach()` could be a footgun. The `continue` statement on line 253 relates to the outer `for` loop, but it could be read as relating to the `Array.forEach()` loop.

There's a judgement call, but it might be safer to avoid mixing the two - it could be done with Array-native methods like this (approximately):

```
  originFundsDepositedEvents                                                                                                                                                                                                                                   
    .map((event) => unpackDepositEvent(spreadEventWithBlockNumber(event), originChainId))
    .filter((deposit) => apiMessages.some(({ depositId }) => deposit.depositId.eq(depositId)))                                                                                                                                                                 
    .forEach((deposit) => observedDeposits.add(this._getDepositKey(deposit.inputToken.toNative(), originChainId, deposit.depositId.toString())));
```

src/gasless/GaslessRelayer.ts
```diff
@@ -371,6 +295,11 @@ export class GaslessRelayer {
       return false;
     }
 
+    // If the deposit is a refund test deposit, we do not want to fill it at all.
+    if (this.config.refundFlowTestEnabled && deposit.outputAmount.eq(MAX_UINT_VAL)) {
+      return false;
+    }
```
pxrl, line 299: This is duplicated now - probably worth factoring it out to a common method.

pxrl (dismissed):
LGTM - only two minor comments 👍

src/gasless/GaslessRelayer.ts
dijanin-brat: [3ed491b](https://github.com/across-protocol/relayer/pull/3246/commits/3ed491bbea10f11767796d1c180838836bb50352)

src/gasless/GaslessRelayer.ts
dijanin-brat: [3ed491b](https://github.com/across-protocol/relayer/pull/3246/commits/3ed491bbea10f11767796d1c180838836bb50352)

dijanin-brat:
@codex review

pxrl (approved):
LGTM!

PR #3247, opened 2026-04-16
feat: Emulate Tron event subscriptions

pxrl:
This change implements Tron event subscriptions-ish behaviour by polling
for events on new block emissions.

### DIFF
--- a/src/clients/EventListener.ts
+++ b/src/clients/EventListener.ts
@@ -3,6 +3,7 @@ import { EventEmitter } from "node:events";
 import { AbiEvent, BaseError, Block, createPublicClient, http, Log as viemLog, parseAbiItem, webSocket } from "viem";
 import { Log } from "../interfaces";
 import {
+  chainIsTvm,
   EventManager,
   getNetworkName,
   getNodeUrlList,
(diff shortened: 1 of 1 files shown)


src/clients/EventListener.ts
```diff
@@ -131,8 +163,38 @@ export class EventListener extends EventEmitter {
           });
         };
 
+        if (tvm) {
+          this.tvmBlocks[provider.name] ??= {};
+          this.on("block", async (blockNumber: number) => {
+            const toBlock = BigInt(blockNumber);
+            const previousBlockNumber = (this.tvmBlocks[provider.name][eventDescriptor] ??= toBlock);
+            if (previousBlockNumber >= toBlock) {
```
bmzig, line 171: Is this reliable to catch reorgs? I thought you need to compare the block's parent hash with the previous block hash to catch a reorg (and then unwind).

pxrl: > Is this reliable to catch reorgs?

Nope; that's a known limitation of this. It's flagged on line 172 :+1: Currently working on that.

pxrl:
Closing in favour of https://github.com/across-protocol/relayer/pull/3250

PR #3267, opened 2026-04-20
feat(arweave): set Arweave data in Redis in addition to trying to set to Arweave

nicholaspai:
## Problem
The dataworker write hot path already computes the exact bundle and merkle-tree payloads that downstream readers need, but after persisting to Arweave those payloads still have to be rediscovered through external Arweave gateway reads. That keeps later read paths dependent on external gateway availability and propagation even when the writer has already materialized the payload locally.

There was also no relayer-side wiring to share a Redis-backed topic cache between the writer path and the `BundleDataClient` read path.

## Solution
- initialize a shared Redis-backed Arweave topic cache during dataworker client construction
- pass that cache into `BundleDataClient` so Arweave-assisted bundle reads can consult Redis first
- pass the same cache into both `bundles-*` and `merkletree-*` calls to `persistDataToArweave()`
- regardless of whetehr `persistDataToArweave()` successfully writes a topic, seed Redis with the payload under `arweave-topic:{tag}`
- when `persistDataToArweave()` detects that the topic already exists on Arweave, still seed Redis with the current payload so later readers do not need to hit Arweave first
- Only set data in Redis if its not already set

## Integration
This PR is designed to pair with the SDK-side Arweave topic-cache reader, so `BundleDataClient` can read these `arweave-topic:{tag}` entries from Redis before falling back to external Arweave gateways.

## Verification
- `eslint src/dataworker/DataworkerUtils.ts src/dataworker/DataworkerClientHelper.ts src/dataworker/Dataworker.ts test/DataworkerUtils.Arweave.ts`
- `prettier --check src/dataworker/DataworkerUtils.ts src/dataworker/DataworkerClientHelper.ts src/dataworker/Dataworker.ts test/DataworkerUtils.Arweave.ts`
- `hardhat test test/DataworkerUtils.Arweave.ts`

### DIFF
--- a/src/dataworker/DataworkerUtils.ts
+++ b/src/dataworker/DataworkerUtils.ts
@@ -456,7 +456,12 @@ export function l2TokensToCountTowardsSpokePoolLeafExecutionCapital(
 
 /**
  * Persists data to Arweave with a given tag, given that the data doesn't
- * already exist on Arweave with the tag.
+ * already exist on Arweave with the tag. Also persists the data to the cache (e.g. a Redis cache) if provided.
+ * Contract:
+ * - `tag` is a deterministic identifier for the bundle payload.
+ * - Repeated calls with the same `tag` must provide semantically identical `data`.
+ * - Because of that invariant, it is safe to seed Redis from `data` even if Arweave is unavailable.
+ * - When Arweave already has data for `tag`, that persisted payload is assumed to must match `data`.
  * @param client The Arweave client to use for persistence.
  * @param data The data to persist to Arweave.
  * @param logger A winston logger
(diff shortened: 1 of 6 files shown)


nicholaspai:
Leaving this PR in draft is intentional.

This change depends on a new `@across-protocol/sdk` release that adds the Arweave topic-cache reader / updated `BundleDataClient` constructor. Until that SDK version is published and this repo bumps to it, the PR will not build cleanly in CI.

In other words: this relayer PR should not be merged on its own. It should be merged only after the paired SDK release is cut and the relayer dependency is updated to that version.

src/dataworker/DataworkerUtils.ts
```diff
@@ -466,10 +466,12 @@ export async function persistDataToArweave(
   client: caching.ArweaveClient,
   data: Record<string, unknown>,
   logger: winston.Logger,
-  tag?: string
+  tag?: string,
+  topicCache?: interfaces.CachingMechanismInterface
 ): Promise<void> {
+  const normalizedTag = tag ?? "";
```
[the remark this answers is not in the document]
nicholaspai: Fixed in `58bec2ed`.

`persistDataToArweave()` now fails fast with `Arweave tag is required` before any Arweave lookup when the tag is missing/empty, so we no longer query `getByTopic("")` or risk writing under an empty topic.

I also added a regression test covering the omitted-tag case and verified:
- `yarn eslint src/dataworker/DataworkerUtils.ts test/DataworkerUtils.Arweave.ts`
- `yarn prettier --check src/dataworker/DataworkerUtils.ts test/DataworkerUtils.Arweave.ts`
- `yarn test --bail test/DataworkerUtils.Arweave.ts`

src/dataworker/DataworkerUtils.ts
```diff
@@ -466,8 +466,10 @@ export async function persistDataToArweave(
   client: caching.ArweaveClient,
   data: Record<string, unknown>,
   logger: winston.Logger,
-  tag?: string
+  tag: string,
+  topicCache?: interfaces.CachingMechanismInterface
 ): Promise<void> {
+  assert(tag.length > 0, "Arweave tag is required");
```
pxrl, line 477: Do we have to make this mandatory? Or is there just no use case where we don't supply it?

nicholaspai: Yeah the latter. Codex suggested we add this and then when I suggested making the param required I realized we never leave it undefined anyways. We can probably delete this line though

nicholaspai: Here's codex's comment it claims that no tag matches on unexpected things https://github.com/across-protocol/relayer/pull/3267#discussion_r3113128320

nicholaspai: I think we should force the caller to set a tag otherwise it'll be very hard to find whateever is setl

nicholaspai: `tag` is already mandatory at the type level here since the signature is `tag: string`; the assertion is only guarding the runtime empty-string case.

All current callsites pass a non-empty tag, and I want to keep the fail-fast behavior because `""` would turn into an effectively unscoped `getByTopic()` lookup. So there isn't an intended optional-tag mode here, just a runtime invariant check on top of the type contract.

src/dataworker/DataworkerUtils.ts
```diff
@@ -466,67 +471,119 @@ export async function persistDataToArweave(
...
-      message: "Arweave balance is above minimum target balance",
-      address,
-      balance: formatWinston(balance),
-      minimumBalance: formatWinston(MINIMUM_AR_BALANCE),
     });
+    const mark = profiler.start("persistDataToArweave");
+
+    // Check if data already exists on Arweave with the given tag.
+    // If so, we don't need to persist it again.
+    const [matchingTxns, address, balance] = await Promise.all([
```
nicholaspai: All of the code here is just copied and moved into try-catch, unfortunately the diff is large.

src/dataworker/DataworkerUtils.ts
```diff
@@ -466,67 +471,119 @@ export async function persistDataToArweave(
...
+        message: "Time to persist to Arweave",
+      });
+    }
+  } finally {
+    await persistArweaveTopicToCache(topicCache, tag, data, logger);
   }
+}
 
-  if (matchingTxns.length > 0) {
+async function persistArweaveTopicToCache(
```
nicholaspai, line 568: This function is new and called in the `finally{}` block in the try-catch above

src/dataworker/DataworkerUtils.ts
pxrl: The try/finally statement is really large here - can we squash it down? We should only need try/finally when we want to unconditionally do something, even on error, but most of the code in here shouldn't actually be throwing any errors.

nicholaspai:
@codex

src/dataworker/DataworkerUtils.ts
```diff
@@ -514,19 +536,72 @@ export async function persistDataToArweave(
...
+      mark.stop({
+        message: "Time to persist to Arweave",
+      });
+    } catch (error) {
+      logger.debug({
+        at: "DataworkerUtils#persistDataToArweave",
+        message: "Failed to persist data to Arweave",
+        tag,
+        error: String(error),
+      });
```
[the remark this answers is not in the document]
nicholaspai: this is by design

bmzig (approved):

src/dataworker/DataworkerUtils.ts
nicholaspai: https://github.com/across-protocol/relayer/pull/3267/changes/c3ff24ec246a7aee9910dbc0b60e13163fa0bd19

PR #3306, opened 2026-04-27
improve(test): Build concurrently with lint & test

pxrl:
Build is only used to detect typing issues - it's not actually used by test or lint. Identify it specifically as a type-checking mechanism (i.e. drop dist output) and run it concurrently with lint and test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,9 +27,6 @@ jobs:
           paths:
             - node_modules
       - run:
-          name: Build
-          command: yarn build
-      - run:
           name: Extract Solana version
           command: |
             set -euo pipefail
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -61,10 +58,18 @@ jobs:
           root: .
           paths:
             - node_modules
-            - dist
```
dijanin-brat, line 64: hmm, why remove dist from the workspace?

pxrl: There are no downstream consumers in the test or lint workflows, so it's not actually required. This is also why `tsc --noEmit` is called - it just checks the build, but doesn't retain it anywhere.

dijanin-brat (approved):

PR #3307, opened 2026-04-27
fix(test): Use correct Solana version

pxrl:
If the package used is an alpha/beta release, or something other than whatever the latest is on contracts master, then the version will mis-resolve. Anchor to the version that's actually installed via package.json instead.

While here, OVERRIDE_SOLANA_VERISON -> SOLANA_VERSION.


dijanin-brat (approved):
make sense

PR #3309, opened 2026-04-27
feat: Add support for swapAndBridgeWithPermit flow in GaslessRelayer

dijanin-brat:
This PR will enable `swapAndBridgeWithPermit` flow in the GaslessRelayer.


src/gasless/GaslessRelayer.ts
```diff
@@ -548,15 +560,25 @@ export class GaslessRelayer {
...
 
               if (!found) {
-                if (depositMessage.permitType === "receiveWithAuthorization") {
-                  const authToken = isSwap ? toAddressType(depositMessage.swapToken, originChainId) : inputToken;
+                const authToken = isSwap ? toAddressType(depositMessage.swapToken, originChainId) : inputToken;
+                if (depositMessage.permitType === "erc3009") {
                   found = await this._findAuthorizationUsed(originChainId, authToken, authorizer, nonce);
+                } else if (depositMessage.permitType === "permit") {
+                  const nonceConsumed = await isErc2612PermitNonceConsumed({
+                    tokenAddress: authToken.toNative(),
```
bmzig: Should you use `.toNative()` here? Would that work with TRON? Maybe we need to use `toEvmAddress()`. Same with line 348

dijanin-brat: we are not using permit flow on Tron, so i think we are good there, but i somehow agree that maybe `toEvmAddress()` is more suitable

dijanin-brat: [abfee97](https://github.com/across-protocol/relayer/pull/3309/commits/abfee978ecb31d73b4abb4da6fd6f49ebeb01947)

src/utils/GaslessUtils.ts
```diff
@@ -147,11 +150,12 @@ export function restructureGaslessDeposits(depositMessages: APIGaslessDepositRes
   return depositMessages.map((msg): AnyGaslessDepositMessage => {
     const { swapTx, requestId, signature } = msg;
     const { chainId: originChainId, data } = swapTx;
-    const { depositId, witness, integratorId, metadata } = data;
-    const permitType = data.type === "permit2" ? "permit2" : "receiveWithAuthorization";
+    const { depositId, witness, integratorId, metadata, type } = data;
+    const permitType = type as GaslessPermitType;
```
pxrl: Are we able to validate this, rather than just asserting type?

src/utils/GaslessUtils.ts
```diff
@@ -147,11 +150,12 @@ export function restructureGaslessDeposits(depositMessages: APIGaslessDepositRes
...
 
     if ("BridgeAndSwapWitness" in witness) {
       const raw = witness.BridgeAndSwapWitness.data;
+      const swapMsg = msg as APIGaslessSwapAndBridgeDepositResponse;
```
pxrl, line 175: Same as above. Feels like this could be fragile if the API changes.

src/utils/GaslessUtils.ts
```diff
@@ -353,6 +359,34 @@ export async function isPermit2NonceUsed(permit2: Contract, owner: string, permi
...
+  tokenAddress: string;
+  owner: string;
+  signedNonce: string;
+  provider: Provider;
+}): Promise<boolean> {
+  const onChainNonce = await getTokenPermitNonce({
+    tokenAddress: params.tokenAddress,
+    owner: params.owner,
+    provider: params.provider,
+  });
```
pxrl: ooc, why not just supply `params` here? i.e. `await gettokenPermitNonce(params)`.

src/utils/GaslessUtils.ts
```diff
@@ -353,6 +359,34 @@ export async function isPermit2NonceUsed(permit2: Contract, owner: string, permi
...
+  return onChainNonce.gt(BigNumber.from(params.signedNonce));
```
pxrl: Might be OK to drop the `BigNumber.from()` here.

```suggestion
  return onChainNonce.gt(params.signedNonce);
```

pxrl (commented):
Only minor - but lgtm

src/utils/GaslessUtils.ts
dijanin-brat: [dadb120](https://github.com/across-protocol/relayer/pull/3309/commits/dadb1204d0618e2f8e307bbdf7c489148a8ac0f7)

src/utils/GaslessUtils.ts
dijanin-brat: yup, good catch

[dadb120](https://github.com/across-protocol/relayer/pull/3309/commits/dadb1204d0618e2f8e307bbdf7c489148a8ac0f7)

src/utils/GaslessUtils.ts
dijanin-brat: [dc8ce13](https://github.com/across-protocol/relayer/pull/3309/commits/dc8ce13b89787056f11f026b28803ff9513a7145)

src/utils/GaslessUtils.ts
dijanin-brat: We don't need to check this because if the witness has `BridgeAndSwapWitness`, then the msg is surely `APIGaslessSwapAndBridgeDepositResponse`

If API changes the response it will be problematic even with the type check

src/gasless/GaslessRelayer.ts
```diff
@@ -323,20 +326,30 @@ export class GaslessRelayer {
...
           const owner = getGaslessAuthorizerAddress(depositMessage);
           const permitNonce = getGaslessPermitNonce(depositMessage);
-          const nonceUsed = await isPermit2NonceUsed(this.permit2Contracts[originChainId], owner, permitNonce);
+          const nonceUsed =
+            depositMessage.permitType === "permit2"
+              ? await isPermit2NonceUsed(this.permit2Contracts[originChainId], owner, permitNonce)
+              : await isErc2612PermitNonceConsumed({
+                  tokenAddress: swapTokenAddr.toEvmAddress(),
+                  owner,
+                  signedNonce: permitNonce,
```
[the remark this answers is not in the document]
dijanin-brat: that is intended

dijanin-brat:
@codex review

src/gasless/GaslessRelayer.ts
```diff
@@ -548,15 +560,25 @@ export class GaslessRelayer {
...
-                  const authToken = isSwap ? toAddressType(depositMessage.swapToken, originChainId) : inputToken;
+                const authToken = isSwap ? toAddressType(depositMessage.swapToken, originChainId) : inputToken;
+                if (depositMessage.permitType === "erc3009") {
                   found = await this._findAuthorizationUsed(originChainId, authToken, authorizer, nonce);
+                } else if (depositMessage.permitType === "permit") {
+                  const nonceConsumed = await isErc2612PermitNonceConsumed({
+                    tokenAddress: authToken.toEvmAddress(),
+                    owner: authorizer,
+                    signedNonce: nonce,
+                    provider: this.providersByChain[originChainId],
```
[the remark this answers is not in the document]
dijanin-brat: Both of them should be the same value

pxrl (approved):

PR #3328, opened 2026-04-28
improve(test): Optimise Dataworker tests for one-time deployments

pxrl:
This saves a huge amount of time in test execution when run locally. The 
benefits in circleci are less pronounced because it runs with
concurrency 20. This change may however permit that to be reudced to
save costs. Additionally, enable the Solana tests to run locally by
ensuring that the Solana validator is started.


package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants": "^3.1.109",
     "@across-protocol/contracts": "5.0.7",
-    "@across-protocol/sdk": "4.3.150",
+    "@across-protocol/sdk": "4.3.150-alpha.2",
```
pxrl: nb. Needs package release before merge.

pxrl: Done: https://github.com/across-protocol/relayer/pull/3328/commits/f1d618447d525c2ead6b9370b62d7373aed49fb2

dijanin-brat (approved):

PR #3337, opened 2026-04-30
chore: Drop redundant exports

pxrl:
- Remove export attribute when a function or variable is only used internally.
- Remove functions or variables where they are not used at all.

Redundancy surfaced by knip.


pxrl:
@codex review

nicholaspai (dismissed):

PR #3343, opened 2026-04-30
improve: strictNullChecks (A/n)

pxrl:


src/adapter/bridges/OFTBridge.ts
```diff
@@ -95,9 +95,8 @@ export class OFTBridge extends BaseBridgeAdapter {
    */
   async roundAmountToSend(amount: BigNumber): Promise<BigNumber> {
     // Fetch `sharedDecimals` if not already fetched
-    this.sharedDecimals ??= await this.getL1Bridge().sharedDecimals();
-
-    return OFT.roundAmountToSend(amount, this.l1TokenInfo.decimals, this.sharedDecimals);
+    const sharedDecimals = (this.sharedDecimals ??= await this.getL1Bridge().sharedDecimals());
+    return OFT.roundAmountToSend(amount, this.l1TokenInfo.decimals, sharedDecimals);
```
dijanin-brat, line 99: hmm, do we really need this change? it was much cleaner as it was IMO

src/adapter/l2Bridges/OFTL2Bridge.ts
```diff
@@ -206,9 +206,8 @@ export class OFTL2Bridge extends BaseL2BridgeAdapter {
    */
   private async roundAmountToSend(amount: BigNumber, decimals: number): Promise<BigNumber> {
     // Fetch `sharedDecimals` if not already fetched
-    this.sharedDecimals ??= await this.getL2Bridge().sharedDecimals();
-
-    return OFT.roundAmountToSend(amount, decimals, this.sharedDecimals);
+    const sharedDecimals = (this.sharedDecimals ??= await this.getL2Bridge().sharedDecimals());
+    return OFT.roundAmountToSend(amount, decimals, sharedDecimals);
```
dijanin-brat, line 210: same as above, i don't really see how this is helping us with strictNullChecks

src/adapter/bridges/OFTBridge.ts
pxrl: Agree it was cleaner before; the reason it's like this currently is that `this.sharedDecimals` can be `undefined`, so the agent has chosen to dereference it and ensure it's not `undefined` prior to using it. In this case it's specifically because of the runtime call to the contract to query `sharedDecimals`. We could probably avoid this by using an async static constructor that handled this init, but that's a larger refactor.

dijanin-brat (approved):

nicholaspai (approved):

PR #3344, opened 2026-04-30
improve: strictNullChecks (B/n)

pxrl:


src/gasless/GaslessRelayer.ts
```diff
@@ -117,7 +123,13 @@ export class GaslessRelayer {
...
 
   private api: AcrossSwapApiClient;
-  protected signerAddress: EvmAddress;
+  private _signerAddress?: EvmAddress;
+
+  // Populated by initialize(); pre-init access throws.
+  protected get signerAddress(): EvmAddress {
+    assert(isDefined(this._signerAddress), "GaslessRelayer: signerAddress accessed before initialize()");
+    return this._signerAddress;
+  }
```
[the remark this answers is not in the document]
pxrl: Updated: https://github.com/across-protocol/relayer/pull/3344/commits/5a72a93f1b98290f431607ecd56bdc3a8b637b16

@codex review

src/deposit-address/DepositAddressHandler.ts
```diff
@@ -70,11 +91,13 @@ export class DepositAddressHandler {
       message: "Initializing DepositAddressHandler",
     });
 
-    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier } = process.env;
+    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier = "deposit-address-handler" } = process.env;
+    assert(isDefined(runIdentifier), "DepositAddressHandler: RUN_IDENTIFIER env var is required");
```
dijanin-brat: hmm, i would like it more to throw if there is no `BOT_IDENTIFIER` provided then default to something like this. There can be the case where we want to spin up a test instance and it kills production instance because of the same `BOT_IDENTIFIER` for example.

src/gasless/GaslessRelayer.ts
```diff
@@ -141,11 +159,13 @@ export class GaslessRelayer {
       message: "Initializing GaslessRelayer",
     });
 
-    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier } = process.env;
+    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier = "across-relayer-gasless" } = process.env;
+    assert(isDefined(runIdentifier), "GaslessRelayer: RUN_IDENTIFIER env var is required");
```
dijanin-brat: same as the above

src/hyperliquid/HyperliquidExecutor.ts
```diff
@@ -184,17 +205,19 @@ export class HyperliquidExecutor {
       });
       this.abortController.abort();
     });
-    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier } = process.env;
+    const { RUN_IDENTIFIER: runIdentifier, BOT_IDENTIFIER: botIdentifier = "across-relayer-hyperliquid" } = process.env;
+    assert(isDefined(runIdentifier), "HyperliquidExecutor: RUN_IDENTIFIER env var is required");
```
dijanin-brat: same here

src/deposit-address/DepositAddressHandler.ts
pxrl: IMO `BOT_IDENTIFIER` defaults are OK - we've had this in the relayer since this feature was introduced. I'm actually also thinking that `RUN_IDENTIFIER` should be defaulting to something random, because it's actually annoying when running this locally and `RUN_IDENTIFIER` isn't set.

In general we're safe to assume the env is providing these, so I don't worry about them being undefined in production.

dijanin-brat (dismissed):

nicholaspai (dismissed):

src/deposit-address/DepositAddressHandler.ts
pxrl: Ended up resolving this via a broader refactor here: https://github.com/across-protocol/relayer/pull/3363

tldr we now have defaults for botIdentifier and runIdentifier.

src/scripts/validateRunningBalances.ts
```diff
@@ -255,13 +253,14 @@ export async function runScript(baseSigner: Signer): Promise<void> {
...
-                const spokePoolAddress = spokePoolClients[leaf.chainId].spokePoolAddress.toEvmAddress();
+                const spokePoolAddressStr = spokePoolAddress.toEvmAddress();
                 let depositsToSpokePool: Log[];
                 // Handle the case that L1-->L2 deposits for some chains for ETH do not emit Transfer events, but
                 // emit other events instead. This is the case for OpStack chains which emit DepositFinalized events
                 // including the L1 and L2 ETH (native gas token) addresses.
                 if (chainIsOPStack(leaf.chainId) && tokenInfo.symbol === "WETH") {
-                  const ovmL2BridgeContractInfo = CONTRACT_ADDRESSES[leaf.chainId].ovmStandardBridge;
+                  const ovmL2BridgeContractInfo = getContractEntry(leaf.chainId, "ovmStandardBridge");
+                  const nativeTokenInfo = getContractEntry(leaf.chainId, "nativeToken");
```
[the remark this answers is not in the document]
pxrl: Updated: https://github.com/across-protocol/relayer/pull/3344/commits/9ddc3c41b9092df4444856914f1cc97aac251163

dijanin-brat (approved):

PR #3356, opened 2026-05-04
 feat(deposit-address): skip non-correct classifications

amateima:
Indexer PR [across-protocol/indexer#929](https://github.com/across-protocol/indexer/pull/929) adds a `transferClassification` field on every deposit-address ERC20 transfer (`correct_transfer | mis_route | intent_refund`). 

This PR makes `DepositAddressHandler` execute only `correct_transfer` rows; everything else is dropped with a structured info log so it's visible in pre-prod without paging on-call.

The withdraw-refund path that should consume `mis_route` / `intent_refund` rows is not implemented yet — see OPS-353. Without this filter, the relayer would attempt to bridge them; in particular, `intent_refund` would re-loop the same intent every bundle as the SpokePool keeps refunding it.

Closes OPS-407


src/deposit-address/DepositAddressHandler.ts
```diff
@@ -168,7 +168,29 @@ export class DepositAddressHandler {
...
+    // Only execute correct_transfer rows. mis_route and intent_refund transfers are routed to the
+    // refund-withdraw path (OPS-353), which is not implemented yet — drop them here so the bot
+    // doesn't try to bridge them (intent_refund in particular would re-loop the same intent).
+    // Unknown classifications are also dropped (forward-compat) until explicitly supported.
+    const correctTransferMessages = depositMessages.filter((m) => {
+      const classification = m.erc20Transfer.transferClassification;
+      if (classification === "correct_transfer") {
+        return true;
+      }
+      this.logger.info({
```
amateima: SHould this be a debug log? 🤔 I don't want to polute Slack channel

dijanin-brat (approved):

PR #3370, opened 2026-05-07
feat(deposit-address): gate withdraw flow behind WITHDRAW_ENABLED

amateima:
Partition indexer messages by classification: 
- `correct_transfer` keeps the existing forward path
- `mis_route / intent_refund` are routed to a new `initiateWithdraw` stub only when `WITHDRAW_ENABLED=true` (strict equality with "true"), and dropped otherwise.


src/deposit-address/DepositAddressHandler.ts
```diff
@@ -230,6 +256,27 @@ export class DepositAddressHandler {
     await forEachAsync(filtered, async (depositMessage) => {
       await this.initiateDeposit(depositMessage);
     });
+
+    await forEachAsync(filteredWithdraws, async (depositMessage) => {
+      await this.initiateWithdraw(depositMessage);
+    });
```
pxrl, line 261: Is this robust to initiating concurrent withdrawals on the same chain? Specifically RE things like nonce collisions. Ultimately there needs to be some serialisation, and ideally it's internal to the bot, not at the RPC/chain level.

pxrl: (nb. I see for now `initiateWithdrawal()` only emits a log, but presumably it will subsequently submit a transaction)

amateima: Good point. I just followed the approach we use for deposits execution ([code here](https://github.com/across-protocol/relayer/blob/master/src/deposit-address/DepositAddressHandler.ts#L229-L231)) so it's more of a systematic issue we need to fix in all execution flows, not only withdrawals

amateima: > only emits a log, but presumably it will subsequently submit a transaction

Am I good to merge the implementation that only produces logs and improve concurrent txn submissions in a dedicated PR?

dijanin-brat: I suggest 2 things:
1. To execute both deposits and withdrawals in parallel
2. To use dispatcher as we do in the gasless-relayer so we don't have that much collisions

I am okey with doing both of this things in the follow-up PR since `initiateWithdrawal` is just a placeholder for now 👍

dijanin-brat (approved):

PR #3371, opened 2026-05-07
fix: use legacy bridge interface for Lens

bmzig:
The Lens USDC bridge does not support `finalizeDeposit` so we need to still use `finalizeWithdrawal` there.

### DIFF
--- a/src/finalizer/utils/zkSync.ts
+++ b/src/finalizer/utils/zkSync.ts
@@ -21,6 +21,7 @@ import {
   EvmAddress,
   Address,
   Provider,
+  CHAIN_IDs,
 } from "../../utils";
 import { FinalizerPromise, CrossChainMessage } from "../types";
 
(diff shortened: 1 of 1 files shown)


pxrl (dismissed):

src/finalizer/utils/zkSync.ts
```diff
@@ -225,18 +226,22 @@ async function prepareFinalization(
...
     withdrawal.l2MessageIndex,
-    withdrawal.sender,
+    isLegacyBridge ? undefined : withdrawal.sender,
     withdrawal.l2TxNumberInBlock,
     withdrawal.message,
     withdrawal.proof,
   ];
+  const calldata = await (isLegacyBridge
+    ? l1SharedBridge.populateTransaction.finalizeWithdrawal(...args.filter(isDefined))
+    : l1SharedBridge.populateTransaction.finalizeDeposit(args));
```
pxrl: Might be a little bit cleaner to avoid the "global" await:

```suggestion
  const calldata = isLegacyBridge
    ? await l1SharedBridge.populateTransaction.finalizeWithdrawal(...args.filter(isDefined))
    : await l1SharedBridge.populateTransaction.finalizeDeposit(args);
```

bmzig: I actually think it's cleaner to have the global await as opposed to two separate ones. Is there an advantage to having two?

pxrl: It's just style preference for me; the parentheses that have to span the `isLegacyBridge` ternary statement ends up making the code IMO harder to read because it's harder to see where the bracket closes. The behaviour should be the same either way; not a hard requirement from me either way though :+1:

pxrl (approved):

PR #3373, opened 2026-05-08
refactor: Split Redis cache from underlying client

pxrl:
Redis can be used independently for non-cache services. This gives a better structure and cleans up some tech debt.


src/utils/index.ts
```diff
@@ -72,7 +72,6 @@ export * from "./TypeGuards";
 export * from "./Help";
 export * from "./LogUtils";
 export * from "./TypeUtils";
-export * from "./RedisUtils";
```
dijanin-brat, line 75: why not include `./redis.ts` here as well? In this case we need to import `redis.ts` separately from everything else we have in utils dir.

Also, might be good to change the name to `Redis.ts` so its consistent.

nicholaspai (dismissed):

src/utils/index.ts
pxrl: Fair call :+1: https://github.com/across-protocol/relayer/pull/3373/commits/6d41b778d5d27d6d2f2efa8c83e4c601564bed3e

dijanin-brat (approved):

PR #3374, opened 2026-05-08
chore: Bump redis 4.7 -> 5.12

pxrl:
- v4 is now in maintenance; v5 is the future.
- Reconnect cause now logged per attempt.
- SET options use typed { expiration }, not deprecated { EX }.
- Enables future RESP3 opt-in (client-side caching).


src/cache/Redis.ts
```diff
@@ -60,8 +60,10 @@ export class RedisCache implements RedisCacheInterface {
       // No TTL
       return (await this.client.set(key, String(val))) ?? undefined;
     } else if (expirySeconds > 0) {
-      // EX: Expire key after expirySeconds.
-      return (await this.client.set(key, String(val), { EX: expirySeconds })) ?? undefined;
+      // Expire key after expirySeconds.
+      return (
+        (await this.client.set(key, String(val), { expiration: { type: "EX", value: expirySeconds } })) ?? undefined
+      );
```
dijanin-brat, line 65: do we need  this outer bracket? It should work with just

`return (await this.client.set(key, String(val), { expiration: { type: "EX", value: expirySeconds } })) ?? undefined;`

pxrl: It seems like we need something following `return`, otherwise tsc complains about unreachable code.

screenshot-1778489228

dijanin-brat (dismissed):

dijanin-brat (approved):

PR #3389, opened 2026-05-13
feat: add a usdh sponsored route

bmzig:
This PR is meant to enable sponsored flows where USDH is the base token.

### DIFF
--- a/src/hyperliquid/HyperliquidExecutor.ts
+++ b/src/hyperliquid/HyperliquidExecutor.ts
@@ -326,12 +326,12 @@ export class HyperliquidExecutor {
         message: "Observed new order event",
         swapFlowInitiated,
       });
-      const baseToken =
-        this.dstOftMessenger.address === log.address
-          ? "USDT0"
-          : this.dstCctpMessenger.address === log.address
-            ? "USDC"
-            : "USDH";
+      const baseToken = {
+        [this.dstOftMessenger.address]: "USDT0",
+        [this.dstCctpMessenger.address]: "USDC",
+        [this.dstUsdhMessenger.address]: "USDH",
+      }[log.address];
+      assert(isDefined(baseToken), `Could not find the base token for handler ${log.address}`);
       const finalToken = getTokenInfo(EvmAddress.from(swapFlowInitiated.finalToken), this.chainId).symbol;
       const pairId = `${baseToken}-${finalToken}`;
       const pair = this.pairs[pairId];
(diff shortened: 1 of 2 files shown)


src/hyperliquid/HyperliquidExecutor.ts
```diff
@@ -125,6 +127,7 @@ export class HyperliquidExecutor {
     const signer = clients.hubPoolClient.hubPool.signer;
     this.dstOftMessenger = getDstOftHandler().connect(signer.connect(this.clients.dstProvider));
     this.dstCctpMessenger = getDstCctpHandler().connect(signer.connect(this.clients.dstProvider));
+    this.dstUsdhMessenger = getDstUsdhHandler().connect(signer.connect(this.clients.dstProvider));
```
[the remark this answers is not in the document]
bmzig: @codex You should assume this isn't an issue and `getDstUsdhHandler()` will always return a valid contract. Can you review again with that in mind?

src/hyperliquid/HyperliquidExecutor.ts
```diff
@@ -323,7 +326,12 @@ export class HyperliquidExecutor {
         message: "Observed new order event",
         swapFlowInitiated,
       });
-      const baseToken = this.dstOftMessenger.address === log.address ? "USDT0" : "USDC";
+      const baseToken =
+        this.dstOftMessenger.address === log.address
+          ? "USDT0"
+          : this.dstCctpMessenger.address === log.address
+            ? "USDC"
+            : "USDH";
```
dijanin-brat: I see that we are doing this just here, but maybe for the sake of cleaner code we can isolate this into separate function? I am not strongly against this, would like to hear your thought on this.

Also, i really think we should try to avoid nested ternary operators also.

src/hyperliquid/HyperliquidExecutor.ts
```diff
@@ -603,6 +611,14 @@ export class HyperliquidExecutor {
...
   }
 
+  private dstHandler(tokenSymbol: string): Contract {
+    return tokenSymbol === "USDT0"
+      ? this.dstOftMessenger
+      : tokenSymbol === "USDH"
+        ? this.dstUsdhMessenger
+        : this.dstCctpMessenger;
+  }
+
```
dijanin-brat, line 623: The same as above RE nested ternary operators

src/hyperliquid/HyperliquidExecutor.ts
bmzig: What do you suggest? We do `if x, else if y, else`? I feel like the nested ternary is cleaner than that, but if there's another idea, we can try it.

dijanin-brat: i would maybe prefer switch for this

src/hyperliquid/HyperliquidExecutor.ts
bmzig: Addressed in other comment.

src/hyperliquid/HyperliquidExecutor.ts
bmzig: I updated here: https://github.com/across-protocol/relayer/pull/3389/commits/023a72a162dc0aa3b0078a59a6d3d63a07220a38

dijanin-brat (approved):

src/hyperliquid/HyperliquidExecutor.ts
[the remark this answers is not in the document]
bmzig: @codex. This is not an issue. The USDH handler _does_ exist. Keep that in mind and review again.

PR #3399, opened 2026-05-21
fix: Use Base58 Tron address format when quering SwapAPI

dijanin-brat:


src/utils/DepositAddressUtils.ts
```diff
@@ -53,6 +54,17 @@ export function normalizeDepositAddressMessage(message: DepositAddressMessage):
...
 
+/**
+ * Returns the chain-native address string. TVM (Tron) uses Base58Check (`T...`); EVM chains keep `0x`.
+ * Accepts either form as input (e.g. after {@link normalizeDepositAddressMessage}).
+ */
+export function toChainNativeAddress(chainId: number, address: string): string {
+  if (!chainIsTvm(chainId)) {
+    return address;
+  }
+  return TvmAddress.from(address).toNative();
```
pxrl: Hm, are we unable to use `toAddressType()` here? I _think_ it should work, but could there be some directionality issues with it?

dijanin-brat: yea, 100% right

[a44b60f](https://github.com/across-protocol/relayer/pull/3399/commits/a44b60f60e6ac239584049000be84d6b5f04532b)

pxrl (approved):

PR #3406, opened 2026-05-23
fix(dataworker): reuse SVM instruction params PDA when sized to fit

droplet-rl:
Skip the close+init pair on the instruction_params PDA when an existing PDA from a previously crashed run is already large enough to hold the current refund leaf payload. Only close+reinit when the existing buffer is undersized.

executeRelayerRefundLeaf auto-closes the PDA via Anchor's close = signer constraint, so a stale PDA at the start of _executeRelayerRefundLeafSvm only persists when a prior run crashed before reaching the execute step. The unconditional close+init in that recovery path races stale-RPC preflight simulation: the close TX confirms before some providers update their view of the slot, so init's preflight sees the still-allocated PDA and account(init) rejects deterministically. Before SDK v4.3.159 this was masked by quorum retries; after fast-fail on SVM_TRANSACTION_PREFLIGHT_FAILURE it surfaces on every recovery (PD-325142 and prior dataworker pages).

Reuse is safe because the codama-encoded payload puts the 8-byte Anchor discriminator at offset 0, the fragment-write loop overwrites starting at offset 0, and Anchor's Account<T>::try_deserialize reads only the schema-required prefix and ignores trailing bytes. WriteInstructionParamsFragment bounds-checks against the allocated size, so reuse is only attempted after verifying the existing buffer is at least as large as the new payload.

Scope: _executeRelayerRefundLeafSvm only. _executeSlowFillLeafSvm and SvmFillerClient's fill-relay-via-instruction-params path can take the same fix in follow-ups. Does not introduce padded totalSize on init.

Slack thread: PD-325142 (https://risk-labs.pagerduty.com/incidents/Q2WQ66XKTQ8YIU). Upstream SDK change that exposed this: across-protocol/sdk@e8ab986.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -2925,14 +2925,26 @@ export class Dataworker {
       vault,
     });
 
-    // Optionally close the existing instruction params account and add an instruction which loads new data into the instruction params PDA.
+    const relayerRefundLeafParamsEncoder = SvmSpokeClient.getExecuteRelayerRefundLeafParamsEncoder();
+    const relayerRefundLeafBytes = relayerRefundLeafParamsEncoder.encode({
+      rootBundleId,
+      relayerRefundLeaf: toSvmRelayerRefundLeaf(leaf),
+      proof,
+    });
+
+    // PDA auto-closes via `close = signer`; existence here implies a prior crashed run. Reuse if sized to fit.
     const instructionParamsAccount = await fetchEncodedAccount(provider, instructionParamsPda);
-    // If the account exists, define the instruction needed to close the instruction account.
-    if (instructionParamsAccount.exists) {
+    const needsClose =
+      instructionParamsAccount.exists && instructionParamsAccount.data.length < relayerRefundLeafBytes.length;
+    const needsInit = !instructionParamsAccount.exists || needsClose;
+    let txSignature;
+    if (needsClose) {
       this.logger.debug({
         at: "Dataworker#executeRelayerRefundLeafSvm",
-        message: "Need to close existing instruction params account",
+        message: "Existing instruction params account is undersized; closing and reinitializing",
         instructionParamsAccount: instructionParamsAccount.address,
+        existingSize: instructionParamsAccount.data.length,
+        requiredSize: relayerRefundLeafBytes.length,
       });
       const closeInstructionParamsIx = SvmSpokeClient.getCloseInstructionParamsInstruction({
         signer: kitKeypair,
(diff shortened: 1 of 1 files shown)


droplet-rl:
### Review

**Logic** — the three branches are exhaustive and correct:
- no PDA → `needsClose=false, needsInit=true` (init only)
- PDA exists & ≥ required → both false (reuse)
- PDA exists & undersized → both true (close + init)

`txSignature` is left `undefined` on the reuse path but is reassigned inside the fragment-write loop before any later use, so no UB drift.

**Safety of reuse** — the claim holds under two assumptions worth calling out explicitly:

1. **Encoder output starts at account offset 0.** The fragment-write loop writes `relayerRefundLeafBytes` from offset 0, so whatever the codama encoder produces (discriminator + struct, or just struct) overwrites the prior payload's leading bytes byte-for-byte. After the loop, bytes `[0, relayerRefundLeafBytes.length)` are identical to what a fresh `init + write` would have produced. Trailing bytes (when the prior allocation was larger) are stale, but that's fine **iff** the on-chain consumer is lenient about size.
2. **`Account<T>::try_deserialize` is lenient.** Anchor's borsh path reads only as many bytes as `T` needs and ignores tail bytes, which is the standard behavior. Worth a quick scan of the SVM program to confirm there's no explicit `data.len() == expected` check or strict deserializer wrapper around `instruction_params` — if one exists, reuse with a larger prior allocation would regress. The reuse path here only triggers when `prior_size >= required_size`, so equal-size reuse is always safe even under a strict check; the only risk is when the prior allocation was strictly larger.

**Remaining race** — the close+init path is unchanged and still vulnerable to the stale-RPC preflight race when the new payload exceeds the prior allocation. In practice that's a narrow window (size growth across consecutive recovery cycles), but the same SDK-4.3.159 short-circuit will hard-fail it. The PR description correctly flags padding-on-init as the durable fix; suggest filing a follow-up so it doesn't get lost.

**Out-of-scope correctness** — verified the two cousins mentioned in the PR description:
- `_executeSlowFillLeafSvm` submits the leaf inline via `getExecuteSlowRelayLeafInstruction` and doesn't use the `instructionParams` PDA at all — no follow-up needed there.
- `SvmFillerClient` doesn't reference `instructionParams` / close+init either.

So the "same shape, take the same fix in follow-up" framing in the PR body is slightly stale — the actual instructionParams close+init pattern lives only in `_executeRelayerRefundLeafSvm`. Worth tightening the description so future readers don't go hunting.

**Nits (non-blocking)**
- `needsClose` reads as "we want to close" but it's really "existing is undersized." Renaming to `existingUndersized` (or similar) and deriving `needsClose = existingUndersized; needsInit = !exists || existingUndersized;` makes the intent obvious. Optional.
- `let txSignature;` is pre-existing untyped — leave it.

**Bottom line:** ship it. The fix is narrow, the reasoning is sound, and the reuse case is the common one post-crash. The two follow-ups (padding-on-init for the size-growth race; tightening the PR description re: slow-fill/filler scope) are paper cuts, not blockers.

pxrl (approved):

PR #3407, opened 2026-05-23
chore: bump @across-protocol/sdk to 4.3.160

droplet-rl:
## Summary

Bump `@across-protocol/sdk` from `4.3.159` to `4.3.160`.

## Why

SDK `4.3.160` ships `across-protocol/sdk@f33ce4d` — `fix(providers): preserve Error prototype and cause across quorum failures`. That fixes `createSendErrorWithMessage`, which previously did `{ ...sendError, ...new Error(message) }` and stripped the Error prototype (and the message itself, since V8's `Error.message` is non-enumerable). The result was that every quorum/fallback failure across both SVM and EVM paths surfaced as a plain object — most visibly as `error: "could not extract error from 'Object' instance {}"` in the dataworker page that motivated #3406.

With this SDK bump the relayer will see real `Error` instances with the wrapper message plus the underlying RPC error in `.cause` next time a quorum failure surfaces.

## Test plan

- [x] `yarn install` regenerates `yarn.lock` cleanly.
- [x] `yarn lint` passes.
- [ ] CI green on PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@across-protocol/constants": "^3.1.109",
     "@across-protocol/contracts": "5.0.10",
-    "@across-protocol/sdk": "4.3.159",
+    "@across-protocol/sdk": "4.3.160",
     "@arbitrum/sdk": "^4.0.5",
     "@consensys/linea-sdk": "^0.3.0",
     "@coral-xyz/anchor": "^0.31.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -39,10 +39,10 @@
     bs58 "^6.0.0"
     ethers "5.7.2"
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/relayer",
-  "version": "3.0.0",
+  "version": "3.0.1",
```
pxrl: @droplet-rl Revert this - we don't bump package version in this repository.

droplet-rl: Done — dropped the version-bump commit. Rebased the branch onto master so only the SDK bump remains (head `68f0cc7`, force-pushed with `--force-with-lease`). PR title/body updated to match.

pxrl (approved):

PR #3416, opened 2026-05-26
fix(svm): silence FillRelay txn failures (debug, not error)

droplet-rl:
## Summary

Recent SDK changes — [across-protocol/sdk#1440](https://github.com/across-protocol/sdk/pull/1440) and [#1443](https://github.com/across-protocol/sdk/pull/1443) — preserve `SolanaError` type information through `QuorumFallbackSolanaRpcFactory`. As a side effect, `SvmFillerClient`'s `Failed to send fill transaction (Solana error code: -32002)` log line now consistently fires at **error level with `mrkdwn` set** on every Solana preflight failure (most commonly: a competing relayer landed the fill first, so our preflight rejects). Previously the wrapped `SolanaError` failed the `isSolanaError(e)` typeguard and fell through into a quieter branch, so the alert is effectively a regression.

### Operational context

Last 24h on `zion-across-relayer-primary` saw 8 distinct `-32002` alerts to Slack. Cross-referenced each against the indexer's final `fillTxHash`:

- **6 / 8** were filled by a **competing relayer** while our preflight kept failing — i.e. we lost the race and the Slack alert announced it.
- **2 / 8** were filled by zion itself (one on the next outer-loop iteration — a genuine retry).

In all 8 cases the deposit was filled within ~6 minutes. There is no user-visible incident — these alerts are spamming `#alerts` for what is, by design, a recoverable outcome.

## Change

Downgrade `this.logger.error` → `this.logger.debug` at the two FillRelay-failure catch sites in `src/clients/SvmFillerClient.ts`:

- `executeFillImmediately` (~L174)
- `executeTxnQueue` (~L256)

`simulateQueue`'s `logger.error` with `notificationPath: "across-error"` is unchanged — it's a distinct batch-level signal about simulation health.

`mrkdwn` is retained on the (now debug-level) line so deposit context is preserved for log-trawling diagnosis. Per ops directive, FillRelay-based errors should be silent (a failed fill is recoverable: another relayer fills, or our outer loop re-picks).

## Symmetry follow-up (not in this PR)

The EVM filler (`MultiCallerClient`) already encodes "this revert is fine, don't alert" via `knownRevertReasons` (`"RelayFilled"`, `"relay filled"`, ...) + `canIgnoreRevertReason()`: matching reverts log at `debug`; unexpected ones log at `error` with `notificationPath`. SVM has no equivalent — `SvmFillerClient` treats every fill failure as the same class today.

To bring EVM and SVM to parity, a follow-up should:

1. Extract a `knownSolanaFillErrorCodes` constant (start: `[SVM_TRANSACTION_PREFLIGHT_FAILURE]`) and a `canIgnoreSvmFillError(e)` helper that mirrors `canIgnoreRevertReason`.
2. Use it to gate log level: `debug` for known-benign failures, `warn`/`error` for unexpected ones (wallet / RPC outage / unrecognised codes).
3. Plumb the inner Solana error (`error.data.err`, `error.data.logs`) from the `@solana/rpc-transformers` wrapper into the catch block so fine-grained gating becomes possible — e.g. distinguishing `FillStatus already filled` (silent) from `compute budget exceeded` (worth alerting on).

(2) and (3) are deliberately deferred to keep this PR minimal and immediately deployable.

## Test plan

- [ ] Solana fill that hits `-32002` (preflight failure): no Slack alert; appears at debug level with `mrkdwn` context preserved.
- [ ] Successful Solana fill: unchanged — `"Filled v3 deposit on SVM 🚀"` still logs at info.
- [ ] `simulateQueue` failure path: unchanged — still alerts via `notificationPath: "across-error"`.

---

Originated from a Slack thread investigating a `-32002` alert for deposit `4363585`. Full triage transcript including the regression diagnosis is in-thread.

### DIFF
--- a/src/clients/SvmFillerClient.ts
+++ b/src/clients/SvmFillerClient.ts
@@ -171,7 +171,7 @@ export class SvmFillerClient {
         errorMessage = `Solana error code: ${e.context.__code}`;
       }
 
-      this.logger.error({
+      this.logger.debug({
         at: "SvmFillerClient#executeFillImmediately",
         message: `Failed to send fill transaction (${errorMessage})`,
         mrkdwn,
@@ -253,7 +253,7 @@ export class SvmFillerClient {
           errorMessage = `Solana error code: ${e.context.__code}`;
         }
 
-        this.logger.error({
+        this.logger.debug({
           at: "SvmFillerClient#executeTxnQueue",
           message: `Failed to send fill transaction (${errorMessage})`,
           mrkdwn,
--- a/src/clients/SvmFillerClient.ts
+++ b/src/clients/SvmFillerClient.ts
@@ -49,15 +49,15 @@ type QueuedSvmFill = {
   mrkdwn: string;
 };
 
-// Known-benign FillRelay error codes (recoverable: another relayer won the race, or our outer loop retries).
-// EVM parity: `knownRevertReasons` / `canIgnoreRevertReason` in MultiCallerClient.ts.
+// Known-benign Solana error codes for fillRelay / slowFillRequest submission (recoverable: another
+// actor won the race, or our outer loop retries). EVM parity: `knownRevertReasons` in MultiCallerClient.ts.
 const knownSolanaFillErrorCodes = new Set<number>([arch.svm.SVM_TRANSACTION_PREFLIGHT_FAILURE]);
 const retryDelaySeconds = 1;
 
-// Suppression is gated on both the tx kind (only fillRelay — slow-fill request failures stay loud)
-// and the error code (only the known-benign codes). Either dimension alone would be too broad.
-function canIgnoreSvmFillError(kind: SvmTxKind, e: unknown): boolean {
-  return kind === "fillRelay" && isSolanaError(e) && knownSolanaFillErrorCodes.has(e.context.__code);
+// Both fillRelay and slowFillRequest preflight failures can be recoverable races (another actor won
+// the race, or our outer loop retries), so suppression is gated only on the benign error code.
+function canIgnoreSvmFillError(e: unknown): boolean {
+  return isSolanaError(e) && knownSolanaFillErrorCodes.has(e.context.__code);
 }
 
 export class SvmFillerClient {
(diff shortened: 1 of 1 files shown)


droplet-rl:
Codex re-reviewed the gating fix and found no issues. PR is clean from automated review — waiting on human review.

src/clients/SvmFillerClient.ts
```diff
@@ -39,15 +39,27 @@ type ReadyTransactionsPromise = Promise<SolanaTransaction[]>;
...
-const retryableErrorCodes = [arch.svm.SVM_TRANSACTION_PREFLIGHT_FAILURE];
+// Known-benign FillRelay error codes (recoverable: another relayer won the race, or our outer loop retries).
+// EVM parity: `knownRevertReasons` / `canIgnoreRevertReason` in MultiCallerClient.ts.
+const knownSolanaFillErrorCodes = new Set<number>([arch.svm.SVM_TRANSACTION_PREFLIGHT_FAILURE]);
 const retryDelaySeconds = 1;
 
+// Suppression is gated on both the tx kind (only fillRelay — slow-fill request failures stay loud)
+// and the error code (only the known-benign codes). Either dimension alone would be too broad.
+function canIgnoreSvmFillError(kind: SvmTxKind, e: unknown): boolean {
+  return kind === "fillRelay" && isSolanaError(e) && knownSolanaFillErrorCodes.has(e.context.__code);
```
pxrl: `slowFillRequest` should also be ignored here.

droplet-rl: Good catch — fixed in `bc2d66e` (will land on push).

Dropped the `kind === "fillRelay"` gate from `canIgnoreSvmFillError`. Both `fillRelay` and `slowFillRequest` preflight failures hit the same `-32002` for the same reason (someone else landed it first / outer loop will retry), so neither should alert.

Kept the `SvmTxKind` discriminator + the `kind` log field — still useful when triaging a failed submission post-hoc.

pxrl (approved):

PR #3417, opened 2026-05-26
fix(svm): reconnect on Solana WS close instead of aborting listener

droplet-rl:
## Summary

- Solana hosted WS providers (notably Chainstack) routinely drop subscriptions without a close frame (code 1006). Chainstack does this on a deterministic ~10-minute wallclock schedule — measured at 95.9% lock-step over a 7-day prod sample. Alchemy does it less predictably. The Anza keepalive issue [#7022](https://github.com/anza-xyz/agave/issues/7022) referenced in the existing comment is unrelated to these close events.
- Previous behaviour: any single subscription error aborted the entire SVM listener. The parent \`SpokePoolClient#childExit\` handler treats clean exits as no-op (see comment at \`src/clients/SpokePoolClient.ts:162\` — *"Future: Optionally restart it based on the exit code"*), so Solana event ingestion silently stopped for the rest of the relayer process lifetime.
- New behaviour: wrap \`readSlot\` and \`readEvent\` in a reconnect loop with bounded exponential backoff (1s → 30s, reset on any successful update). Exit only via the shared \`abortSignal\` (SIGHUP / parent disconnect still tear down cleanly).
- Behaviour change worth flagging: the \`--nProviders < quorum\` abort in \`readEvent\` is removed — with reconnect, there's no permanent provider loss to count. The startup-time \`nProviders >= quorum\` assert is kept. Previously a 3-provider, \`quorum=2\` config would tear the listener down once two of three providers had dropped; it now reconnects all three indefinitely. With Chainstack dropping every 10 min that is the desired behaviour.

## Out of scope (potential follow-ups)

- \`readSlot\` is still single-provider (\`providers[0]\`). With reconnect that SPOF is mostly defanged, but a fundamentally broken \`providers[0]\` URL would still keep us blind on slot updates. Failover would be a larger structural change.
- Log severity is still \`warn\` on every reconnect to preserve the existing alert signal. With Chainstack's 10-min cadence that's ~144 warns/day per bot — downgrading to debug after the first occurrence within an outage is a sensible follow-up.

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] \`eslint\` clean
- [x] \`prettier --check\` clean
- [ ] Deploy to one bot (e.g. \`usdh\`); confirm warns continue to fire on Chainstack ~every 10 min but the listener keeps delivering slot updates and events.
- [ ] Confirm SIGHUP / parent disconnect still tears the listener down cleanly (no orphaned child).
- [ ] Confirm no regression in EVM / TVM listeners (this PR only touches the SVM file).

### DIFF
--- a/src/libexec/RelayerSpokePoolListenerSVM.ts
+++ b/src/libexec/RelayerSpokePoolListenerSVM.ts
@@ -21,7 +21,7 @@ import {
 } from "../utils";
 import { getRedisCache } from "../cache/Redis";
 import { ScraperOpts } from "./types";
-import { bootstrap } from "./util/bootstrap";
+import { bootstrap, waitForAbort } from "./util/bootstrap";
 import { postBlock, postEvents } from "./util/ipc";
 import { scrapeEvents as _scrapeEvents } from "./util/svm";
 
@@ -124,6 +124,11 @@ async function listen(
   const RECONNECT_BACKOFF_MIN_S = 1;
   const RECONNECT_BACKOFF_MAX_S = 30;
 
+  // Abortable sleep: returns once the backoff elapses or the abort signal fires, whichever
+  // comes first. Without this, a SIGHUP arriving mid-backoff would block child shutdown for
+  // up to RECONNECT_BACKOFF_MAX_S, conflicting with the bootstrap teardown contract.
+  const abortableDelay = (seconds: number) => Promise.race([delay(seconds), waitForAbort(abortSignal)]);
+
   const logProviderError = (providerName: string, err: unknown, backoffS: number) => {
     const message = "Caught error on Solana provider; reconnecting.";
     const ctx = { at, message, provider: providerName, backoffS };
@@ -155,7 +160,7 @@ async function listen(
           return;
         }
         logProviderError(providerName, err, backoffS);
-        await delay(backoffS);
+        await abortableDelay(backoffS);
         backoffS = Math.min(backoffS * 2, RECONNECT_BACKOFF_MAX_S);
       }
     }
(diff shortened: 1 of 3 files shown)


src/libexec/RelayerSpokePoolListenerSVM.ts
```diff
@@ -117,62 +118,79 @@ async function listen(
...
-        if (!postBlock(Number(slot), currentTime)) {
-          abortController.abort();
+  const readSlot = async (provider: WSProvider, providerName: string) => {
+    let backoffS = RECONNECT_BACKOFF_MIN_S;
+    while (!abortSignal.aborted) {
+      try {
+        const subscription = await provider.slotNotifications().subscribe({ abortSignal });
+        for await (const update of subscription) {
+          backoffS = RECONNECT_BACKOFF_MIN_S; // reset on any successful update
+          const { slot } = update as { slot: bigint }; // Bodge: pretend slots are blocks.
```
pxrl: @droplet-rl do we still need this type assertion? WE might be able to get rid of it.

src/utils/Tasks.ts
```diff
@@ -7,6 +7,29 @@ export function fireAndForget(task: () => Promise<unknown>): () => void {
   return () => void task().catch(() => undefined);
 }
 
+/**
+ * Sleep for `seconds`, returning early if `signal` aborts. Clears the pending timer on abort
+ * (so it doesn't keep the event loop alive past shutdown) and detaches the abort listener on
+ * normal completion (so listeners don't accumulate when called in a retry loop).
+ */
+export function abortableDelay(seconds: number, signal: AbortSignal): Promise<void> {
```
pxrl, line 15: Can you add a test for this?

pxrl (approved):

pxrl (approved):

PR #3420, opened 2026-05-27
fix(finalizer): resolve Binance pending rebalance symbols via equivalence remapping

droplet-rl:
## Summary

The Binance finalizer added in #3347 throws `AssertionError: Token USDC not found on chain 56` whenever the pending-rebalance map contains a USDC entry for BSC, taking down the whole finalizer run for chain 56.

The `BinanceStablecoinSwapAdapter.getPendingRebalances` keys entries by the order's logical token symbol (`sourceToken` / `destinationToken`, e.g. `"USDC"`) against the Binance entrypoint chain. When the entrypoint resolves to BSC, you get `{ 56: { "USDC": <amount> } }`. The finalizer's new `getPositivePendingRebalanceAmountsByBinanceCoin` then calls `resolveAcrossToken("USDC", 56, true)`, but `TOKEN_SYMBOLS_MAP["USDC"]` has no address on chain 56 — BSC's USDC is registered as the separate symbol `USDC-BNB`. The assert fires and rejects the chain-56 finalizer promise.

This switches the decimals lookup to `getTokenInfoFromSymbol`, which routes through `TOKEN_EQUIVALENCE_REMAPPING` and matches how the rebalancer adapter itself resolves token info (`_getTokenInfo` → `getTokenInfoFromSymbol`). The values stored in the pendingRebalances map were already produced via the same equivalence-aware path, so decimals stay consistent.

## Test plan

- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test --grep "Binance finalizer helpers"` (added a BSC USDC case that reproduces the original crash without the fix)

### DIFF
--- a/test/BinanceFinalizer.ts
+++ b/test/BinanceFinalizer.ts
@@ -37,6 +37,16 @@ describe("Binance finalizer helpers", function () {
     expect(deductions).to.deep.equal({ USDC: 60 });
   });
 
+  it("resolves logical USDC symbol on BSC where the on-chain symbol is USDC-BNB", function () {
+    const deductions = getPositivePendingRebalanceAmountsByBinanceCoin({
+      [CHAIN_IDs.BSC]: {
+        USDC: toBNWei("50", 18),
+      },
+    });
+
+    expect(deductions).to.deep.equal({ USDC: 50 });
+  });
+
   it("subtracts credited, swap, and pending rebalance amounts from orphan sweep candidates", function () {
     const sweepableBalance = getSweepableOrphanBinanceBalance(250_000, 10_000, 20_000, 30_000);
 
--- a/test/BinanceFinalizer.ts
+++ b/test/BinanceFinalizer.ts
@@ -47,6 +47,16 @@ describe("Binance finalizer helpers", function () {
     expect(deductions).to.deep.equal({ USDC: 50 });
   });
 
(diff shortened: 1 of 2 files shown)


test/BinanceFinalizer.ts
```diff
@@ -37,6 +37,16 @@ describe("Binance finalizer helpers", function () {
     expect(deductions).to.deep.equal({ USDC: 60 });
   });
 
+  it("resolves logical USDC symbol on BSC where the on-chain symbol is USDC-BNB", function () {
+    const deductions = getPositivePendingRebalanceAmountsByBinanceCoin({
+      [CHAIN_IDs.BSC]: {
+        USDC: toBNWei("50", 18),
```
pxrl, line 43: @droplet-rl Add a test for USDT here.

droplet-rl: Added in 779bf4e (`test/BinanceFinalizer.ts:50-58`) — mirrors the USDC-BNB case with `USDT: toBNWei("75", 18)` on BSC, which exercises the `USDT-BNB → USDT` entry in `TOKEN_EQUIVALENCE_REMAPPING` (BSC USDT-BNB is 18 decimals vs. L1 USDT at 6). Let me know if you wanted broader USDT coverage and I'll add more cases.

pxrl (approved):

PR #3424, opened 2026-05-27
feat: Add Arc support

dijanin-brat:


src/clients/ProfitClient.ts
```diff
@@ -402,7 +402,7 @@ export class ProfitClient {
     const nativeTokenPriceUsd = this.getPriceOfToken(nativeToken.symbol);
     const auxiliaryNativeTokenCostUsd = auxiliaryNativeTokenCost
       .mul(nativeTokenPriceUsd)
-      .div(bn10.pow(nativeToken.decimals));
+      .div(bn10.pow(chainId === CHAIN_IDs.ARC ? 18 : nativeToken.decimals));
```
pxrl: Is this safe? I wonder if we should be resolving decimals more deliberately - maybe in a separate function. Then we would put some unit tests around it to ensure that we're also working OK with other weird chains (i.e. I'm also thinking about the _next_ weird snowflake chain that we add).

src/common/Constants.ts
```diff
@@ -315,7 +319,7 @@ Object.values(CHAIN_IDs)
 // These are the spokes that can hold the native token for that network, so they should be added together when calculating whether
 // a bundle execution is possible with the funds in the pool.
 const resolveNativeTokenSpokes = () => {
-  const chains = [CHAIN_IDs.LINEA, CHAIN_IDs.ZK_SYNC];
+  const chains = [CHAIN_IDs.LINEA, CHAIN_IDs.ZK_SYNC]; // @TODO: Should we add ARC here?
```
pxrl: I think we can probably skip it for Arc because we should still read the correct balance via ERC20.balanceOf(spoke).

src/common/Constants.ts
```diff
@@ -422,6 +428,7 @@ type L2BridgeConstructor<T extends BaseL2BridgeAdapter> = new (
 const resolveCanonicalBridges = (): Record<number, L1BridgeConstructor<BaseBridgeAdapter>> => {
   const bridges = {
     [CHAIN_IDs.ARBITRUM]: ArbitrumOrbitBridge,
+    // @TODO: No canonical bridge?
```
pxrl: Yeah; there's kinda no fallback here. Can probably delete the comment.

dijanin-brat: [4fd4f94](https://github.com/across-protocol/relayer/pull/3424/commits/4fd4f94fb98cfd71be5688bee88f38e7b21b56e0)

src/common/Constants.ts
dijanin-brat: [4fd4f94](https://github.com/across-protocol/relayer/pull/3424/commits/4fd4f94fb98cfd71be5688bee88f38e7b21b56e0)

src/clients/ProfitClient.ts
dijanin-brat: i think we should do it like this until we have real weird case with some new chain and see what exactly we want/need from that util function to do [4fd4f94](https://github.com/across-protocol/relayer/pull/3424/commits/4fd4f94fb98cfd71be5688bee88f38e7b21b56e0)

pxrl (approved):

PR #3437, opened 2026-05-30
fix(inventory): isolate per-adapter, per-bridge, and per-chain failures in inventory accounting

droplet-rl:
## Summary

A transient outage in one adapter/bridge/chain's read API used to propagate up through the inventory update chain and crash the relayer/rebalancer process. This PR isolates failures at three layers so a single venue's read-side outage no longer gates the entire fill loop, while preserving accounting safety.

## Motivation

Two recent incidents on the same theme:

1. `zion-across-relayer-primary` crashed on a 502 from Hyperliquid:
```
HttpRequestError: 502 Bad Gateway - nginx/1.22.1
  at HyperliquidStablecoinSwapAdapter._getInitiatedWithdrawalsFromHypercore
  at HyperliquidStablecoinSwapAdapter.getPendingRebalances
  at BaseRebalancerClient.getPendingRebalances
  at InventoryClient.update
```

2. `zion-across-fast-relayer-rebalancer` crashed on a 503 from the shared BridgeApi:
```
HttpError: HTTP 503: Service Unavailable
  at BridgeApiClient.getWithRetry
  at BridgeApiClient.getAllTransfersInRange
  at BridgeApi.queryL1BridgeInitiationEvents
  at TokenSplitterBridge.queryL1BridgeInitiationEvents
  at BaseChainAdapter (getOutstandingCrossChainTransfers)
```

Both bottom out in `InventoryClient.update`, which had no catch around either the rebalancer-client path or the cross-chain-transfer-accounting path. Treating any one venue's read API as a hard precondition for the whole relay loop is the root problem.

## Change

Three layers of isolation, finest-grain inward:

### 1. `BaseRebalancerClient.getPendingRebalances` (rebalancer-adapter level)

Wraps each `adapter.getPendingRebalances(account)` call in try/catch. On failure: log at `error` (`at: \"BaseRebalancerClient#getPendingRebalances\"`, includes adapter name), skip that adapter for the cycle. Aggregates whatever the healthy adapters returned.

### 2. `BaseChainAdapter.getOutstandingCrossChainTransfers` (per-bridge level)

Wraps the `(l1Token, monitoredAddress)` block in try/catch. The inner `Promise.all([queryL1BridgeInitiationEvents, queryL2BridgeFinalizationEvents])` stays all-or-nothing on purpose — partial success there would corrupt `outstanding = deposited − finalized` accounting — so the catch sits at the `(token, address)` granularity. On failure: log at `error` (`at: \"<adapterName>#getOutstandingCrossChainTransfers\"`), and the affected pair is simply absent from the cycle's result.

### 3. `CrossChainTransferClient.update` (per-chain safety net)

`Promise.all` → `Promise.allSettled` across chains. Successful chains overwrite with fresh state; **failed chains preserve their previously-recorded state** rather than being blanked. Stale state biases the InventoryClient toward under-rebalancing (safer) instead of duplicate-bridging (worse). Per-chain failures log at `error` (`at: \"CrossChainTransferClient\"`). This is defense-in-depth for any uncaught path inside `BaseChainAdapter`.

## Behavioral effect

- The HL 502 from `HyperliquidStablecoinSwapAdapter.getPendingRebalances` now logs and skips; `pendingRebalances` is partial (HL contribution absent) for that update cycle.
- The 503 from `TokenSplitterBridge` → `BridgeApi` now logs and skips at the per-bridge level; outstanding-transfer entries for the affected `(token, address)` pair are absent for the cycle.
- If a whole chain's accounting somehow rejects, `CrossChainTransferClient.update` preserves the prior cycle's state for that chain rather than blanking it.
- Next cycle retries in all cases.

For a sustained HL outage, the relayer will still attempt to initiate new HL swap rebalances since `rebalanceInventory` doesn't gate on read-side health. Out of scope here; will likely follow up with read-side health gating on HL rebalance initiation in a separate PR.

## Ops note

Silent-by-default is by design (relayer-up beats accounting-perfect), but operators should add Datadog alerts on the new error logs so a persistent outage doesn't quietly degrade inventory accounting:

- `at: \"BaseRebalancerClient#getPendingRebalances\"`
- `at: \"<adapterName>#getOutstandingCrossChainTransfers\"` (substring `#getOutstandingCrossChainTransfers`)
- `at: \"CrossChainTransferClient\"` with `Failed to fetch outstanding cross chain transfers`

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` + `prettier` clean on touched files
- [x] `hardhat test test/RebalancerClient.cumulativeRebalancing.ts test/CrossChainTransferClient.update.ts test/generic-adapters/BaseChainAdapter.isolation.ts test/generic-adapters/SplitBridgeTracking.ts` — 25/25 passing
- [x] `hardhat test test/InventoryClient.InventoryRebalance.ts test/InventoryClient.RefundChain.ts test/Monitor.ts` — 66/66 passing, no regressions
- [x] New unit tests:
  - `RebalancerClient.cumulativeRebalancing.ts` — `\"Isolates per-adapter failures when aggregating pending rebalances\"`
  - `CrossChainTransferClient.update.ts` — three cases: per-chain isolation + stale preservation, fresh replace on success, all-chains-fail does not throw
  - `BaseChainAdapter.isolation.ts` — `\"Isolates a failing bridge so other bridges' outstanding transfers still surface\"`
- [ ] Operator: confirm the three error logs surface in Datadog under the expected `at:` values and wire alerts.


src/rebalancer/clients/BaseRebalancerClient.ts
```diff
@@ -54,8 +54,21 @@ export abstract class BaseRebalancerClient implements RebalancerClient {
...
+      } catch (error) {
+        // Isolate per-adapter failures: a transient outage in one venue's read API (e.g. a 502
+        // from Hyperliquid) must not propagate up to callers like InventoryClient.update and take
+        // the relayer down. The caller operates on a partial view for this cycle.
+        this.logger.error({
+          at: "BaseRebalancerClient#getPendingRebalances",
+          message: `Failed to fetch pending rebalances from ${adapterName} adapter; skipping for this cycle`,
+          error: (error as Error)?.toString() ?? String(error),
+        });
+        return;
```
pxrl, line 70: Is it correct to return undefined here? Should we instead be returning 0 or something, in case the value is dereferenced?

droplet-rl: Good question — it's safe: the `return;` exits only the per-iteration `forEachAsync` callback, not `getPendingRebalances` itself.

- `forEachAsync` in the SDK is `await mapAsync(array, callback)` which is `Promise.all(array.map(callback))` — the callback's return value is discarded, so returning early just skips aggregation for that adapter.
- The outer `getPendingRebalances` always returns `pendingRebalances`, which is initialized to `{}` on entry and only ever gets keys added via `pendingRebalances[chainId][token] = ...` for adapters whose read succeeded. It is never `undefined`, and for the skipped adapter the corresponding `chainId`/`token` entry is simply absent (rather than zero) — i.e. a partial map.
- Consumers handle that shape naturally:
  - `InventoryClient` (`src/clients/InventoryClient.ts:307-317`) reads via `pendingRebalances[chainId]?.[l1TokenSymbol]` with `isDefined` guards, so a missing chain/token entry contributes 0 to the balance computation — exactly the same effect a literal `bnZero` would have, without forcing us to enumerate which chain/token pairs the failed adapter would have reported.
  - `Monitor` similarly indexes via `pendingRebalances[chainId]?.[token]` with optional chaining.

So from a caller's perspective the choice is between "missing key" and "key present with value 0" — both are dereferenced safely, and "missing key" has the advantage that we don't have to guess the set of chain/token pairs the failed adapter would have populated (it differs per adapter and per route config). Happy to switch to explicit zero-fill if you'd rather have the dictionary shape be uniform across success/failure cases, but the current form keeps the failure semantics minimal — "this adapter contributed nothing" rather than "this adapter contributed zero to chains X, Y, Z."

PR #3441, opened 2026-06-01
fix(RedisClient): debounce repeated socket error logs

droplet-rl:
## Summary

node-redis emits one `error` event per pending command the moment a socket resets, so a single Redis disconnect produces a burst of 10–25 identical `Redis error` warn lines under the same `run-identifier`. Examples from `#zbot-across-relayer` and friends show 5–25 messages per disconnect across `relayer-primary`, `secondary`, `usdh`, `sepolia`, `gasless-relayer-sepolia`, `inventory-manager-primary`, `arweave`, `persistent-address-handler`.

This PR collapses the burst:

- Coalesce `error` events with the same cause within a 1 s window into one log line, suffixed with the count when > 1 (e.g. `Redis error (x18)`).
- Flush any pending warning on `ready` and `end` so a recovery or clean teardown doesn't leak a trailing line. The two handlers share a single `flushOnTransition` callback.

No behavior change for callers — this is purely log-shape. The underlying root cause (a `BASIC`-tier Memorystore single-node restart taking out every client at once) is being addressed separately on the infra side.

## Test plan

This change touches only logging behavior on the `RedisClient` event handlers; there are no existing unit tests for `src/utils/Redis.ts` and the affected paths are exercised end-to-end across every bot that talks to Redis.

- [ ] Local smoke: run a bot pointed at a local Redis, restart Redis, verify the warn burst collapses to one line with `(xN)`.
- [ ] Confirm no regressions in the happy path (no extra logs on startup or graceful shutdown).

## Docs

No `AGENTS.md` / `README.md` updates — this change is internal to `RedisClient` event handling and does not change the module map or any public interface.

### DIFF
--- a/src/utils/Redis.ts
+++ b/src/utils/Redis.ts
@@ -68,18 +68,14 @@ export async function connectRedisClient(logger?: winston.Logger, url = REDIS_UR
         flushTimer = setTimeout(flushErrors, ERROR_DEBOUNCE_MS);
       }
     });
-    redisClient.on("ready", () => {
+    const flushOnTransition = (): void => {
       if (flushTimer) {
         clearTimeout(flushTimer);
         flushErrors();
       }
-    });
-    redisClient.on("end", () => {
-      if (flushTimer) {
-        clearTimeout(flushTimer);
-        flushErrors();
-      }
-    });
+    };
+    redisClient.on("ready", flushOnTransition);
+    redisClient.on("end", flushOnTransition);
 
     await redisClient.connect();
     return redisClient;
--- a/src/utils/Redis.ts
+++ b/src/utils/Redis.ts
@@ -73,7 +73,6 @@ export async function connectRedisClient(logger?: winston.Logger, url = REDIS_UR
         clearTimeout(flushTimer);
         flushErrors();
       }
-      logger?.info({ at: "RedisClient", message: "Redis client ready.", url });
     });
     redisClient.on("end", () => {
       if (flushTimer) {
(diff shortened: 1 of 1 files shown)


src/utils/Redis.ts
```diff
@@ -36,7 +36,52 @@ export async function connectRedisClient(logger?: winston.Logger, url = REDIS_UR
...
+      if (!flushTimer) {
+        flushTimer = setTimeout(flushErrors, ERROR_DEBOUNCE_MS);
+      }
+    });
+    redisClient.on("ready", () => {
+      if (flushTimer) {
+        clearTimeout(flushTimer);
+        flushErrors();
+      }
+      logger?.info({ at: "RedisClient", message: "Redis client ready.", url });
```
nicholaspai:
```suggestion
```

nicholaspai (commented):
Don't add info level logs unless absolutely vital for users to see

src/utils/Redis.ts
```diff
@@ -36,7 +36,51 @@ export async function connectRedisClient(logger?: winston.Logger, url = REDIS_UR
...
+      if (pendingErrorCause !== cause) {
+        flushErrors();
+        pendingErrorCause = cause;
+      }
+      pendingErrorCount += 1;
+      if (!flushTimer) {
+        flushTimer = setTimeout(flushErrors, ERROR_DEBOUNCE_MS);
+      }
+    });
+    redisClient.on("ready", () => {
```
nicholaspai: Can we consolidate the ready and end callbacks?

src/utils/Redis.ts
droplet-rl: Acknowledged — agree the ready log is not user-vital. Thanks for committing the removal directly.

src/utils/Redis.ts
droplet-rl: Consolidated — extracted a shared `flushOnTransition` handler and registered it on both `ready` and `end`.

PR #3449, opened 2026-06-03
fix(dataworker): pin minContextSlot on SVM refund-leaf dependency sends

droplet-rl:
## Summary

Closes the cross-slot race that's been paging oncall 1–2x/day for the past month on `zion-across-l2-executor-solana` (PD-325402, PD-325403, PD-325400, PD-325420, plus the May 23 storm of ~30).

## Root cause

`_executeRelayerRefundLeafSvm` issues a sequence of dependent SVM txs: `init / write fragments → LUT create → LUT extend → execute`. The dataworker GCP logs (`bots-across-3839`, `across-spoke-2c-8g`) show the execute's `sendTransaction` preflight runs ~73ms after the last data-write confirms — well inside Solana's ~400ms slot propagation window. If the RPC the relayer routes `sendTransaction` to is one slot behind the just-confirmed write, its preflight runs against stale instruction-params PDA state and returns `SolanaError: Transaction simulation failed`. The next dataworker cycle ~4 minutes later (when state has propagated) succeeds — every observed failure was followed by a success on the same `{rootBundleId, leafId}`.

## Fix

Pin every dependent send to `minContextSlot = slot-at-which-the-previous-dependency-confirmed`. The Solana JSON-RPC spec lets a receiving node reject with the explicit `MinContextSlotNotReached` code if its tip is behind the requested slot, instead of running preflight against stale state. The SDK's existing `RetrySolanaRpcFactory` already retries on transient RPC errors, so the retry happens transparently and the next cycle's state propagates before the retry fires.

## Changes

- **`src/utils/TransactionUtils.ts`**
  - Add `SolanaSendOptions { minContextSlot?: bigint }` and thread it through `signAndSendTransaction` to `provider.sendTransaction(..., { minContextSlot })`.
  - Add `sendAndConfirmSolanaTransactionWithSlot(...)` returning `{ signature, confirmedSlot }`. Reads the slot from `getSignatureStatuses[0].slot` (already available in the polling loop).
  - Existing `sendAndConfirmSolanaTransaction` keeps its `Promise<string>` signature; `minContextSlot` is opt-in via the new last param.

- **`src/dataworker/Dataworker.ts` `_executeRelayerRefundLeafSvm`**
  - Add a local `sendPinned()` closure that captures `confirmedSlot` from each send and threads it forward as `minContextSlot` to the next send. The chain only advances forward.
  - Route every dependency send (close/init params, write fragments, LUT create, both LUT extend loops, execute, executeDeferred, claim-account init, claim, deactivateLut) through `sendPinned`.

- **`test/TransactionUtils.svm.ts`** — 4 tests covering slot extraction, `minContextSlot` threading into options, opt-out when not supplied, `confirmedSlot=undefined` when polling never confirms, and backwards-compat for the existing API.

## Pairs with

[#3448](https://github.com/across-protocol/relayer/pull/3448) (fix C, surface SolanaError context in logs). Together these should eliminate the page class entirely; if a SolanaError does propagate post-B, the C diagnostic ensures it carries the program-level `__code` + RPC `value.err` + program logs.

## Scope

Deliberately narrow. Other SVM call sites (`_executeSlowFillLeafSvm`, `SvmFillerClient`, CCTP finalizer, claims) are untouched — they don't issue an init/write/execute sequence that exposes the same cross-slot race. If post-deploy logs surface similar `MinContextSlotNotReached` retries on other paths, they can be wired the same way in a follow-up.

## Test plan

- [x] `yarn typecheck:fast`
- [x] `yarn lint`
- [x] `RELAYER_TEST=true yarn hardhat test test/TransactionUtils.svm.ts` (4 passing)
- [ ] Post-deploy: confirm the `zion-across-l2-executor-solana` PD page class drops to zero. If a `MinContextSlotNotReached` propagates instead of being retried, it'd show up in Datadog at the same call site — that would indicate the SDK's retries are exhausted (RPC pool genuinely far behind) and warrants tuning retry count, not reverting this PR.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -137,6 +137,11 @@ const ERROR_DISPUTE_REASONS = new Set(["insufficient-dataworker-lookback", "out-
 // instruction is 900 bytes.
 const INSTRUCTION_PARAMS_MAX_WRITE_SIZE = 900;
 
+// Polling parameters for `sendAndConfirmSolanaTransactionWithSlot` when sending the chain of dependency-write txs that
+// precede an SVM relayer-refund-leaf execution. 25 cycles × 600 ms ≈ 15s, ~1.5 slots per cycle on Solana mainnet.
+const SVM_REFUND_LEAF_SEND_POLL_CYCLES = 25;
+const SVM_REFUND_LEAF_SEND_POLL_DELAY_MS = 600;
+
 const { getMessageHash, getRelayEventKey } = sdkUtils;
 
 // Create a type for storing a collection of roots
@@ -2957,9 +2962,13 @@ export class Dataworker {
     // SDK's `RetrySolanaRpcFactory` transparently retries.
     let minContextSlot: bigint | undefined;
     const sendPinned = async (tx: Parameters<typeof sendAndConfirmSolanaTransactionWithSlot>[0]): Promise<string> => {
-      const { signature, confirmedSlot } = await sendAndConfirmSolanaTransactionWithSlot(tx, provider, 25, 600, {
-        minContextSlot,
-      });
+      const { signature, confirmedSlot } = await sendAndConfirmSolanaTransactionWithSlot(
+        tx,
+        provider,
+        SVM_REFUND_LEAF_SEND_POLL_CYCLES,
+        SVM_REFUND_LEAF_SEND_POLL_DELAY_MS,
+        { minContextSlot }
+      );
       if (isDefined(confirmedSlot) && (minContextSlot === undefined || confirmedSlot > minContextSlot)) {
         minContextSlot = confirmedSlot;
       }
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -54,6 +54,7 @@ import {
   createDefaultTransaction,
   getNativeTokenAddressForChain,
   stringifyThrownValue,
+  sendAndConfirmSolanaTransactionWithSlot,
 } from "../utils";
 import { getRedisCache } from "../cache/Redis";
 import {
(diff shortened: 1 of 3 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -2946,6 +2945,24 @@ export class Dataworker {
...
+    // can pin subsequent `sendTransaction` preflights with `minContextSlot`.
+    // Without this, a receiving RPC that is one slot behind the just-confirmed
+    // write runs preflight against stale state (instruction-params PDA
+    // appears empty / wrong size) and rejects with `SolanaError: Transaction
+    // simulation failed` even though the on-chain state is correct. With it,
+    // a behind RPC returns `MinContextSlotNotReached` immediately, which the
+    // SDK's `RetrySolanaRpcFactory` transparently retries.
+    let minContextSlot: bigint | undefined;
+    const sendPinned = async (tx: Parameters<typeof sendAndConfirmSolanaTransactionWithSlot>[0]): Promise<string> => {
+      const { signature, confirmedSlot } = await sendAndConfirmSolanaTransactionWithSlot(tx, provider, 25, 600, {
```
nicholaspai: don't use hardcoded values (e.g. 25 and 600) here, use named variables to improve readability

droplet-rl: Good call — extracted `SVM_REFUND_LEAF_SEND_POLL_CYCLES` (25) and `SVM_REFUND_LEAF_SEND_POLL_DELAY_MS` (600) as module-level constants alongside `INSTRUCTION_PARAMS_MAX_WRITE_SIZE`.

droplet-rl: Addressed in d69a6dd — extracted `SVM_REFUND_LEAF_SEND_POLL_CYCLES` (25) and `SVM_REFUND_LEAF_SEND_POLL_DELAY_MS` (600) as module-level constants and passed them by name. Also rebased onto current master (which now includes #3448), so the merge conflict is gone.

nicholaspai (approved):

PR #3455, opened 2026-06-03
feat: Support tracking & broadcasting balance changes

pxrl:


src/libexec/BalancePublisher.ts
```diff
@@ -0,0 +1,244 @@
...
+    [blockNumber, , returnData] = result;
+  } catch (err) {
+    logger.warn({ at, message: `Failed to read ${chain} balances.`, cause: String(err) });
+    return;
+  }
+
+  await Promise.all(
+    tokens.map(async (token, i) => {
+      const balance = decodeFunctionResult({
+        abi: ERC20_ABI as Abi,
```
pxrl: @droplet-rl Can this type assertion be removed? We should avoid type assertions where possible.

src/libexec/BalancePublisher.ts
```diff
@@ -0,0 +1,244 @@
...
+        functionName: "balanceOf",
+        data: returnData[i].returnData,
+      }) as bigint;
```
pxrl: @droplet-rl Additional type assertion that we likely want to remove, or at least to handle more cleanly.

src/libexec/BalancePublisher.ts
droplet-rl: Yes — the `as Abi` was only there because `ERC20_ABI` was imported from JSON, which TypeScript widens. Switched all three call sites to viem's exported `erc20Abi` (fully typed at the source), which lets me drop:

- `as Abi` on the `decodeFunctionResult` call (this one)
- `as Abi` on the matching `encodeFunctionData` call
- `as bigint` on the `balanceOf` result (viem now infers `bigint` from the typed ABI)
- the runtime `getAbiItem` guard for the `Transfer` event, and the `: AbiEvent` annotation on `TRANSFER_EVENT` — `getAbiItem({ abi: erc20Abi, name: "Transfer" })` is typed precisely
- the local `MinimalERC20.json` import (no longer referenced in this file)

Left the `as \`0x${string}\`` casts on `target`/`args` alone — those come from `isAddress` checks that use Node's `assert` (not a typed assertion function), so the narrowing doesn't propagate without restructuring the validation. Happy to convert if you want it in this PR.

Pushed as ff404084. `tsc --noEmit` clean.

src/libexec/BalancePublisher.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (!isDefined(redis)) {
+    return;
+  }
+
+  const payload = {
+    type: "transfer",
+    schema_version: 1,
+    platform: `chainId:${chainId}`,
+    observed_at: new Date().toISOString(),
+    log,
```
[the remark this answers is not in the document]
pxrl: There's an implicitly-imported .toJSON() handler for bigint that ensures correct bigint stringification.

dijanin-brat (dismissed):

PR #3456, opened 2026-06-03
fix(rebalancer): measure swap spread against oracle fair price, not par

droplet-rl:
## Summary

Generalizes the spread-vs-par formula used inside `getEstimatedCost` on both `BinanceStablecoinSwapAdapter` and `HyperliquidStablecoinSwapAdapter` so it works for any spot market, not just dollar-pegged pairs. The HL adapter previously did `1 − latestPrice`, and BN was originally depth-only (and then in earlier commits on this branch, gated to stablecoin symbols). Both forms hardcoded $1 as the fair price, which:

1. **Missed real cost on stablecoin pairs.** USDCUSDT asks on Binance sit around ~$1.001; a market BUY pays the ~10 bps gap vs $1 even when slippage-vs-bestPx is ~0. BN's depth-only estimator reported ~0 bps on $200k+ swaps that actually cost 10-11 bps round-trip.
2. **Made the par assumption non-reusable.** A naive `latestPrice − 1` for ETH/USDC (`latestPrice ≈ 3500`) would report ~3500 bps of "spread" and dominate the cost-estimator's route selection.

## Approach

Measure the spread against an oracle-derived fair cross-token price `P_base_usd / P_quote_usd`. The ratio is dimensionless: for dollar-pegged pairs it ≈ 1 and the formula reduces to the prior par-based form; for non-par markets it correctly compares the order book's worst-fill price to the true cross-token rate.

```ts
const fairPrice = await this._getFairCrossPrice(
  spotMarketMeta.baseAssetName,
  spotMarketMeta.quoteAssetName
);
const slippageDecimal = spotMarketMeta.isBuy
  ? (latestPrice - fairPrice) / fairPrice
  : (fairPrice - latestPrice) / fairPrice;
spreadPct = Math.max(0, slippageDecimal /* × 100 for BN's percent units */);
```

## Infrastructure

Added two helpers on `BaseAdapter` (already holds the shared `PriceClient`):

- `_getTokenPriceUsd(symbol)` — mirrors `CumulativeBalanceRebalancerClient._getTokenPriceUsd`: resolves the hub-chain token address, supports `RELAYER_TOKEN_PRICE_FIXED_<address>` env override, returns an 18-decimal BigNumber USD price. Cached per-instance.
- `_getFairCrossPrice(baseToken, quoteToken)` — returns the dimensionless ratio `P_base / P_quote` as a plain number, ready for `latestPrice` comparison.

The cache scope matches the cumulative client (per-instance, no TTL) because the rebalancer is short-lived per cycle. Flagged in the source comment so a future long-lived runtime can add eviction.

## Adapter changes

- **BN**: removes the `STABLECOIN_SYMBOLS` gate (added in an earlier commit on this branch), now queries `_getFairCrossPrice` directly. Falls back to depth-only slippage if the oracle fails, with a warn log.
- **HL**: replaces the `1 − latestPrice` block with the same oracle-based form. Falls back to the original par-based formula on oracle failure (preserves current behavior for the stablecoin pairs HL routes today).
- Both adapters now clamp `spreadPct` to `≥ 0` so a favorably-priced book doesn't surface as a negative cost or crash `toBNWei`.

## Production evidence (from prior commits on this branch)

The BN cost estimator was reporting `expectedFees` ≈ 0 on $200k+ same-chain USDCUSDT swaps that realized ~10-11 bps cost. After this change, BN's `expectedFees` should land in the right ballpark when comparing against realized round-trip cost on USDCUSDT, USDC/DAI, and any future non-stable routing.

## Test plan

- [x] `yarn typecheck` passes
- [x] `yarn lint` + prettier clean on the three modified files
- [x] `test/BinanceAdapter.quotes.ts` — 4/4 existing tests pass
- [x] `test/RebalancerClient.cumulativeRebalancing.ts` — 17/17 existing tests pass (this exercises `getEstimatedCost` via the cheapest-route selection logic)
- [ ] Post-deploy: `RebalanceClient.rebalanceCumulativeInventory` logs should show `expectedFees` values in the ~10-15 bps range for BN-routed USDCUSDT swaps, matching realized cost
- [ ] Post-deploy: HL `expectedFees` should remain roughly stable (USDC/USDT pair fair price ≈ 1 so formula yields the same result)

## Doc updates

No `README.md` or `AGENTS.md` updates — the change is implementation-level and the public interface (`getEstimatedCost`) signature is unchanged. The `RebalancerAdapter` interface and `src/rebalancer/README.md` adapter-interface section stay accurate. If reviewers want a comment in `docs/rebalancer-mode-adapter-architecture.md` calling out the oracle-priced reference, happy to add it.

## Branch history note

This PR's commit history shows the evolution discussed in-thread:
1. `d2c1edd` — initial par-based fix for BN (correct for stables, but unsafe for non-stables)
2. `d4d53cf` — gated the par-based logic on a stablecoin symbol allowlist (review feedback)
3. `a0bf620` — generalizes via oracle fair price, removes the stablecoin gate, and applies the same treatment to the HL adapter

Happy to squash on merge.

### DIFF
--- a/src/rebalancer/README.md
+++ b/src/rebalancer/README.md
@@ -68,6 +68,23 @@ Implemented production swap adapters:
 Pending-status Redis sets are keyed by adapter status and signer address, so callers can request pending rebalance
 accounting for a specific EVM account while keeping independently operated accounts isolated from one another.
 
+#### Estimated cost: spread vs oracle fair price
+
+`getEstimatedCost` on the Binance and Hyperliquid adapters includes a spread-cost component that compares the worst-fill price an order would cross on the venue's order book against the oracle-derived fair cross-token price `P_base_usd / P_quote_usd`. For dollar-pegged pairs (e.g. `USDC/USDT`) the ratio is ≈ 1 and the formula reduces to the prior par-based form; for non-par markets (e.g. `WETH/USDC`, `WETH/USDT`) the same formula correctly compares the worst-fill price to the true cross-token rate without producing nonsense estimates.
+
+Oracle wiring:
+
+- `BaseAdapter._getTokenPriceUsd(symbol)` resolves a token's USD price via the same hub-chain-address lookup `CumulativeBalanceRebalancerClient` uses, with results cached per adapter instance for the rebalancer cycle.
+- Operators can override a token's USD price with the `RELAYER_TOKEN_PRICE_FIXED_<address>` env var (where `<address>` is the hex hub-chain ERC-20 address), matching the existing override the cumulative client honors. Use for incident response when the upstream oracle is degraded.
+- `BaseAdapter._getFairCrossPrice(baseToken, quoteToken)` is the helper both adapters call; it returns the quote-per-base ratio so it can be compared directly to the order book's `latestPrice`.
+
+Fallback behavior on oracle failure:
+
+- Binance falls back to depth-only order-book slippage (the prior `slippagePct` measure) — a conservative lower bound that under-estimates real cost when a stablecoin pair trades off par.
+- Hyperliquid falls back to the legacy `1 - latestPrice` par-based formula, which preserves current behavior for the stablecoin pairs HL routes today.
+- Both adapters emit a `warn` log on oracle failure (`Oracle fair price unresolved for <base>/<quote>; falling back to ...`) so operators can detect degraded estimates.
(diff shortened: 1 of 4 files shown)


nicholaspai (commented):
This doesn't work for any markets that don't trade at $1 like WETH-USDT

nicholaspai (approved):

PR #3468, opened 2026-06-08
feat(ProfitClient): Policy-based profitability configuration

droplet-rl:
Consolidates the per-route `MIN_RELAYER_FEE_PCT_<src>_<dst>_<o>_<d>` and `RELAYER_GAS_MULTIPLIER_<src>_<dst>_<o>_<d>` overrides into per-(token-pair) env vars:

```
RELAYER_RAMP_DESTINATIONS_<srcSymbol>_<dstSymbol>=<comma-separated chain IDs>
RELAYER_RAMP_ORIGINS_<srcSymbol>_<dstSymbol>=<comma-separated chain IDs>   # optional
RELAYER_RAMP_MIN_FEE_PCT=<decimal, default 0>
RELAYER_RAMP_GAS_MULTIPLIER=<decimal, default 0>
```

For each deposit, if the input/output token symbols match a configured pair, the destination is in that pair's allowlist, the origin is in the pair's optional origin allowlist (when set), and the origin chain supports unmetered fast rebalance for the input token, the resolved `minRelayerFeePct` and `gasMultiplier` are taken from these vars rather than the per-route lookups.

Token-symbol lookup uses the raw `getTokenSymbol(...)` output, bypassing the `PEGGED_TOKEN_PRICES` symbol-rewrite path that silently rewrote several existing per-route override keys (e.g. `MIN_RELAYER_FEE_PCT_USDT_USDC_42161_999` was never matched because `_getRemappedTokenSymbol("USDT")` returns `"USDC"`).

The fast-rebalance predicate is lifted from `InventoryClient` into a free function in `src/utils/FastRebalanceUtils.ts` so `ProfitClient` can call it without taking on `InventoryClient` as a dependency. `InventoryClient.isUnmeteredFastRebalance` becomes a one-line wrapper.

Existing per-route env vars remain functional for deposits that don't match the new policy.

## Test plan
- [x] `yarn tsc --noEmit` passes.
- [x] `yarn test test/ProfitClient.ConsiderProfitability.ts` passes (14 tests, including the new ramp-policy test).
- [x] `yarn test test/InventoryClient.RefundChain.ts` passes (47 tests; covers the `isUnmeteredFastRebalance` refactor).


src/utils/FastRebalanceUtils.ts
pxrl, line 1: @droplet-rl Find another file in src/utils to locate this. Maybe FillUtils.ts.

src/clients/ProfitClient.ts
```diff
@@ -178,6 +179,11 @@ export class ProfitClient {
     const { inputToken, originChainId, outputToken, destinationChainId } = deposit;
     const srcSymbol = this.getTokenSymbol(inputToken, originChainId);
     const dstSymbol = this.getTokenSymbol(outputToken, destinationChainId);
+
+    if (this.matchesRampPolicy(srcSymbol, dstSymbol, originChainId, destinationChainId, inputToken)) {
+      return toBNWei(process.env.RELAYER_RAMP_GAS_MULTIPLIER ?? "0");
```
pxrl: @droplet-rl This should revert to the default gas multiplier unless otherwise defined.

src/clients/ProfitClient.ts
```diff
@@ -417,6 +423,11 @@ export class ProfitClient {
     const { inputToken, originChainId, outputToken, destinationChainId } = deposit;
     const srcSymbol = this.getTokenSymbol(inputToken, originChainId);
     const dstSymbol = this.getTokenSymbol(outputToken, destinationChainId);
+
+    if (this.matchesRampPolicy(srcSymbol, dstSymbol, originChainId, destinationChainId, inputToken)) {
+      return toBNWei(process.env.RELAYER_RAMP_MIN_FEE_PCT ?? "0");
```
pxrl: @droplet-rl As above, unless defined, revert to the global default for the gas multiplier.

src/clients/ProfitClient.ts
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in 47cf85da. Mirrored the existing `0 <= multiplier <= 4` assertion from the per-route/token/chain override path, and added a test that exercises both the negative and >4 cases. Negative values would have flipped `tokenGasCost` to a profit boost in `estimateFillCost`, exactly as flagged.

src/clients/ProfitClient.ts
```diff
@@ -553,6 +564,40 @@ export class ProfitClient {
     }
   }
 
+  /**
+   * Per-(token-pair) override: a deposit matches when its destination is in
+   * `RELAYER_RAMP_DESTINATIONS_<srcSymbol>_<dstSymbol>`, optionally its origin is in
+   * `RELAYER_RAMP_ORIGINS_<srcSymbol>_<dstSymbol>`, and the origin chain supports unmetered fast
+   * rebalance for the input token. Matching deposits use `RELAYER_RAMP_MIN_FEE_PCT` and
+   * `RELAYER_RAMP_GAS_MULTIPLIER` instead of the per-route lookups.
```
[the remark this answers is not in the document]
droplet-rl: Added a "Ramp policy override" subsection to `src/clients/README.md` under the Profit Client section in 47cf85da. It covers the three predicate conditions (destinations allowlist, optional origins allowlist, `isUnmeteredFastRebalance` requirement), the raw-symbol behavior, the zero defaults, the `0 <= multiplier <= 4` constraint, and a worked USDT->USDC example.

src/utils/FastRebalanceUtils.ts
droplet-rl: Moved `isUnmeteredFastRebalance` into `src/utils/FillUtils.ts` and dropped the standalone `FastRebalanceUtils.ts` (1fbcf26a). InventoryClient/ProfitClient imports unchanged since both already pull it from the barrel.

src/clients/ProfitClient.ts
droplet-rl: Fixed in 1fbcf26a. The ramp branch now only short-circuits when `RELAYER_RAMP_GAS_MULTIPLIER` is actually defined; if the operator sets the destinations allowlist but not the override, it falls through to the existing per-route/token/chain lookup and the global `gasMultiplier` / `gasMessageMultiplier` defaults. Same treatment applied to `RELAYER_RAMP_MIN_FEE_PCT` below. Tests and README updated to match.

src/clients/ProfitClient.ts
droplet-rl: Fixed in 1fbcf26a — `minRelayerFeePct` now only short-circuits to `RELAYER_RAMP_MIN_FEE_PCT` when it is defined, otherwise falls through to the existing per-route/token/chain lookup and `defaultMinRelayerFeePct`.

src/clients/ProfitClient.ts
```diff
@@ -553,6 +577,66 @@ export class ProfitClient {
...
+  /**
+   * Per-(token-pair) override: returns the first policy from `RELAYER_POLICIES` whose
+   * `RELAYER_POLICY_<NAME>_DESTINATIONS_<srcSymbol>_<dstSymbol>` includes `destinationChainId`,
+   * whose optional `RELAYER_POLICY_<NAME>_ORIGINS_<srcSymbol>_<dstSymbol>` includes `originChainId`
+   * (or, when unset, whose origin chain supports unmetered fast rebalance for the input token).
+   * Callers additionally gate on the relevant override env var being defined
+   * (`RELAYER_POLICY_<NAME>_MIN_FEE_PCT` / `RELAYER_POLICY_<NAME>_GAS_MULTIPLIER`); if it isn't,
+   * they fall through to the standard per-route/token/chain lookup and global defaults.
+   */
+  protected matchingPolicy(
```
pxrl, line 592: @droplet-rl I think we should decode the policies once in the constructor, not each time we need to check for a policy.

droplet-rl: Refactored in e9596e09. `RELAYER_POLICIES` is now decoded once in the constructor into a structured registry: each policy gets a `Map<srcSymbol_dstSymbol, { destinations, origins? }>` plus pre-parsed `minFeePct` / `gasMultiplier` BigNumbers, and the `0 <= multiplier <= 4` range check happens at construct time so misconfiguration fails fast. `matchingPolicy` and the two callers now do O(1) lookups against the cached struct — no env reads or string splits on the hot path. Tests restructured to construct a fresh `MockProfitClient` per scenario (matching real operator usage where env is set before startup); the range-violation case now asserts that the constructor throws.

pxrl (approved):

PR #3475, opened 2026-06-10
fix(dataworker): lock L1 executor section

nicholaspai:
## Summary
- add a hub-chain Redis lock for the L1 executor section before building pool-rebalance execution txs
- hold the lock through hub-chain submission and receipt observation, deferring release to TTL on receipt timeout
- keep proposer/disputer/L2 executor flow outside the new lock, except for skipping the old submit path after the L1 helper handles hub-chain txs

## Test plan
- yarn typecheck
- yarn prettier --check src/dataworker/index.ts src/dataworker/README.md
- yarn eslint src/dataworker/index.ts
- git diff --check

### DIFF
--- a/src/dataworker/index.ts
+++ b/src/dataworker/index.ts
@@ -1,4 +1,5 @@
-import { TokenClient } from "../clients";
+import { randomUUID } from "node:crypto";
+import { SpokePoolClient, TokenClient } from "../clients";
 import {
   assert,
   blockExplorerLink,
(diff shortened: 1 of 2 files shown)


src/dataworker/index.ts
```diff
@@ -1,4 +1,5 @@
-import { TokenClient } from "../clients";
+import { randomUUID } from "node:crypto";
```
nicholaspai, line 1: why not use Date.now() as a random generator. Does true randomness really matter for this use case?

PR #3483, opened 2026-06-11
fix(InventoryClient): clamp excess L2 withdrawal to liquid on-chain balance

droplet-rl:
## Summary
- `InventoryClient#withdrawExcessBalances` sized the L2→L1 bridge call off virtual balance (actual + pending inbound rebalance credits + pending L2→L1 withdrawal credits) and submitted that amount without comparing it to the relayer's settled on-chain balance.
- When the swap-rebalancer has in-flight pending CCTP/OFT/Hyperliquid orders destined for a chain, `pendingRebalances[chainId][symbol]` inflates that chain's virtual balance, the `currentAllocPct - targetPct` sizing overshoots what's actually on-chain, and the bridge call reverts at simulation time with `ERC20: transfer amount exceeds balance`.
- Trigger still uses virtual balance (so pending inbound credits stop us from withdrawing every cycle until they settle); the transfer amount is now clamped to liquid balance via `tokenClient.getBalance(chainId, l2Token)`. If liquid is zero we skip the cycle and wait for the credits to actually land on-chain.

## Production symptom
The `zion-across-fast-relayer-rebalancer` bot has been emitting `MultiCallerClient#LogSimulationFailures` every ~5 minutes in `#zbot-across-error` (e.g. parent ts `1781167257.515209`). The reverting tx is the CCTP `depositForBurn` on the HyperEVM/Arb token messenger at `0x28b5a0e9C621a5BadaA536219b3a228C8168cf5d`. Today's snapshot:

- relayer EOA `0x07ae8551be970cb1cca11dd7a11f47ae82e70e67` on HyperEVM has **134,143.96 USDC** on-chain (`balanceOf` against `0xb88339CB…`)
- the bot's most recent `Executed excess L2 inventory withdrawal` log proposes **602,758.70 USDC** (current alloc 35.33%, target 8%, $2.2M cumulative)
- the ~644k phantom comes from `pendingRebalances[HYPEREVM]["USDC"]` — credits from in-flight CCTP/OFT orders that haven't settled

`rebalanceInventoryIfNeeded` already does the right thing on the L1→L2 side via the `unallocatedBalance` check (`InventoryClient.ts:1109`); this PR brings the L2→L1 path in line.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn hardhat test test/InventoryClient.InventoryRebalance.ts` — 18 passing, including two new tests:
  - `Clamps excess withdrawal to liquid on-chain balance when pending rebalance credits inflate the virtual balance`
  - `Skips excess withdrawal when there is no liquid on-chain balance`
- [ ] Deploy and watch `#zbot-across-error` — `fast-relayer-rebalancer` simulation reverts should stop and the bot should fall back to withdrawing whatever's actually on-chain each cycle.

## Docs
- `docs/inventory-virtual-balance-model.md` gets a new "Triggering vs sizing" section documenting the virtual-trigger-but-liquid-transfer rule so future inventory-side mutations don't regress this.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1547,13 +1547,24 @@ export class InventoryClient {
             return;
           }
 
-          // Cap the desired withdrawal at the relayer's liquid on-chain balance. Sizing off
-          // virtual balance would revert at simulation because pending inbound rebalance
-          // credits haven't actually settled on-chain yet.
+          // Cap the withdrawal so the post-withdrawal liquid balance stays at this chain's
+          // target allocation. Two failure modes are being avoided here:
+          //   1. `desiredWithdrawalAmount` is sized off virtual balance, which folds in pending
+          //      inbound rebalance credits — sending more than what's actually on-chain reverts
+          //      the bridge call at simulation with `ERC20: transfer amount exceeds balance`.
+          //   2. Clamping straight to `liquidL2Balance` could drain the chain to zero, undoing
+          //      the operator's intended target reserve and starving any in-flight fills until
+          //      pending inbound credits settle.
+          // Sizing off `liquid - target` covers both: the transfer is bounded by what's settled
+          // on-chain, and the chain retains at least its target allocation in liquid form.
           const liquidL2Balance = this.tokenClient.getBalance(chainId, l2Token);
-          const amountToWithdraw = desiredWithdrawalAmount.lte(liquidL2Balance)
+          const targetAmountInL2 = cumulativeBalanceInL2TokenDecimals.mul(targetPct).div(this.scalar);
+          const liquidAvailableForWithdraw = liquidL2Balance.gt(targetAmountInL2)
+            ? liquidL2Balance.sub(targetAmountInL2)
+            : bnZero;
+          const amountToWithdraw = desiredWithdrawalAmount.lte(liquidAvailableForWithdraw)
             ? desiredWithdrawalAmount
-            : liquidL2Balance;
+            : liquidAvailableForWithdraw;
           if (amountToWithdraw.lte(bnZero)) {
             return;
           }
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1539,14 +1539,33 @@ export class InventoryClient {
           if (pendingWithdrawalAmount.gte(maxL2WithdrawalVolume)) {
             return;
           }
+
+          // Skip any L2 withdrawal while the chain has an outstanding shortfall — pulling
+          // liquid back to L1 concurrently with promised fills means we either fail to fund
+          // those fills or undo the rebalancer's work once the pending inbound settles.
+          if (this.tokenClient.getShortfallTotalRequirement(chainId, l2Token).gt(bnZero)) {
+            return;
+          }
+
+          // Cap the desired withdrawal at the relayer's liquid on-chain balance. Sizing off
+          // virtual balance would revert at simulation because pending inbound rebalance
+          // credits haven't actually settled on-chain yet.
+          const liquidL2Balance = this.tokenClient.getBalance(chainId, l2Token);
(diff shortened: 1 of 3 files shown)


src/clients/InventoryClient.ts
```diff
@@ -1539,14 +1539,50 @@ export class InventoryClient {
...
+          // this L2 token. `desiredWithdrawalAmount` is sized off virtual balance, which folds
+          // in pending inbound rebalance credits (`pendingRebalances` from the rebalancer
+          // adapters via `getBalanceOnChain`) and, for the hub chain, pending L2->L1 withdrawal
+          // credits. The bridge call ultimately pulls tokens from the relayer EOA, so any
+          // amount above what's actually settled on-chain reverts at simulation time. Sizing
+          // off liquid balance keeps the trigger virtual-aware but the *transfer* liquid-only.
+          const liquidL2Balance = this.tokenClient.getBalance(chainId, l2Token);
+          const amountToWithdraw = desiredWithdrawalAmount.lte(liquidL2Balance)
+            ? desiredWithdrawalAmount
+            : liquidL2Balance;
```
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in the follow-up commit.

`getCurrentAllocationPct(..., ignoreShortfall=false)` discounts shortfall in the virtual world, so the unclamped `desiredWithdrawalAmount` is shortfall-safe. The bug was that `min(desired, liquid)` could reach back into liquid that was earmarked for outstanding fills once pending credits inflated `desired` above `liquid`. Now sizing off `max(0, liquid - shortfall)`:

- preserves the shortfall reservation already baked into the virtual-balance math
- still clamps below virtual so the bridge sim can't revert with `ERC20: transfer amount exceeds balance`
- skips the cycle when shortfall fully claims liquid

Added two regression tests covering the partial-shortfall (clamp to `liquid - shortfall`) and full-shortfall (skip entirely) cases, plus a docs update in `inventory-virtual-balance-model.md` spelling the rule out.

src/clients/InventoryClient.ts
```diff
@@ -1539,14 +1539,33 @@ export class InventoryClient {
...
+          // Skip any L2 withdrawal while the chain has an outstanding shortfall — pulling
+          // liquid back to L1 concurrently with promised fills means we either fail to fund
+          // those fills or undo the rebalancer's work once the pending inbound settles.
+          if (this.tokenClient.getShortfallTotalRequirement(chainId, l2Token).gt(bnZero)) {
+            return;
+          }
+
+          // Cap the desired withdrawal at the relayer's liquid on-chain balance. Sizing off
+          // virtual balance would revert at simulation because pending inbound rebalance
+          // credits haven't actually settled on-chain yet.
```
pxrl: This is wrong - doing so would leave the operator with 0 tokens on this chain, whereas the intention is to rebalance down to the desired threshold.

Therefore, the withdrawal amount should reduce the liquid token balance _down_ to the target allocation.

droplet-rl: Good catch — fixed in the follow-up commit.

Replaced `min(desired, liquid)` with `min(desired, max(0, liquid - targetAmount))`, so the post-withdrawal liquid balance stays at the chain's target allocation instead of going to zero. Still bounds the transfer below settled liquidity (so the bridge call doesn't revert at simulation), but no longer strips the operational reserve when the desired (virtual-derived) amount exceeds liquid.

Tests updated to assert the new clamp:
- `Caps the withdrawal at liquid-minus-target when pending rebalance credits inflate the virtual balance` — binding clamp, expected = `liquid - target`
- `Skips the withdrawal when liquid balance is at or below target allocation` — degenerate case where pending credits trigger the threshold but liquid is already inside the target reserve

Kept the skip-on-shortfall guard since that's a separate concern (don't withdraw at all while promised fills are outstanding). Happy to also fold that into the unified clamp (`liquid - max(target, shortfall)`) if you'd prefer — let me know.

PR #3493, opened 2026-06-12
fix(Binance): Filter rebalances on sender side from L2 to L1 via Binance adapter

nicholaspai:
We don't currently filter out outstanding deposits in the L2ToL1 Binance adapter by sender address in the same way that we do for L1ToL2

### DIFF
--- a/src/adapter/l2Bridges/BinanceCEXBridge.ts
+++ b/src/adapter/l2Bridges/BinanceCEXBridge.ts
@@ -120,7 +120,20 @@ export class BinanceCEXBridge extends BaseL2BridgeAdapter {
       );
     });
 
-    const unmatchedDeposits = getOutstandingBinanceDeposits(depositHistory, withdrawHistory, this.depositNetwork);
+    // FilterMap to remove all deposits which originated from another EOA.
+    const filteredDepositHistory = await filterAsync(depositHistory, async (deposit) => {
+      const txnReceipt = await this.getL1Bridge().provider.getTransactionReceipt(deposit.txId);
+      if (!compareAddressesSimple(txnReceipt.from, fromAddress.toNative())) {
+        return false;
+      }
+      return true;
+    });
+
+    const unmatchedDeposits = getOutstandingBinanceDeposits(
+      filteredDepositHistory,
+      withdrawHistory,
+      this.depositNetwork
+    );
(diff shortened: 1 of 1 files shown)


src/adapter/l2Bridges/BinanceCEXBridge.ts
```diff
@@ -108,7 +108,11 @@ export class BinanceCEXBridge extends BaseL2BridgeAdapter {
     // Remove any deposits and withdrawals that are marked as related to a swap.
     const depositHistory = await filterAsync(_depositHistory, async (deposit) => {
       const depositType = await getBinanceDepositType(deposit);
-      return deposit.coin === this.l1TokenInfo.symbol && depositType !== BinanceTransactionType.SWAP;
+      return (
+        deposit.network === this.depositNetwork &&
```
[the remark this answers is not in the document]
nicholaspai: We can assume that we only care about Binance deposits from the current L2

src/adapter/l2Bridges/BinanceCEXBridge.ts
```diff
@@ -120,7 +125,20 @@ export class BinanceCEXBridge extends BaseL2BridgeAdapter {
...
     });
 
-    const unmatchedDeposits = getOutstandingBinanceDeposits(depositHistory, withdrawHistory, this.depositNetwork);
+    // FilterMap to remove all deposits from this L2 which originated from another EOA.
+    const filteredDepositHistory = await filterAsync(depositHistory, async (deposit) => {
+      const txnReceipt = await this.getL2Bridge().provider.getTransactionReceipt(deposit.txId);
+      if (!isDefined(txnReceipt) || !compareAddressesSimple(txnReceipt.from, fromAddress.toNative())) {
+        return false;
+      }
+      return true;
```
pxrl: Minor simplification:

```suggestion
      return isDefined(txnReceipt) && compareAddressesSimple(txnReceipt.from, fromAddress.toNative()));
```

pxrl (dismissed):

src/adapter/l2Bridges/BinanceCEXBridge.ts
nicholaspai: Shouldn't this be `return isDefined(txnReceipt) && compareAddressesSimple(txnReceipt.from, fromAddress.toNative());`?

PR #3500, opened 2026-06-17
feat(deposit-address): add v3 refund-withdrawals for misrouted transfers

amateima:
Screenshot 2026-06-17 at 17 02 23
Screenshot 2026-06-17 at 17 00 45

## Summary

Adds the bot-side **v3** (upgradeable-counterfactual) refund-withdraw flow. Previously `processExecution` dropped every non-`correct_transfer` v3 message ("v3 refund-withdraw not yet supported"), stranding funds that landed on a v3 deposit address but could not be forwarded. `mis_route` classifications now flow to a new `initiateWithdrawV3`, which calls the quote-api `POST /deposit-addresses/sign-withdraw` endpoint ([quote-api #2790](https://github.com/across-protocol/quote-api/pull/2790)) and submits the signed `AdminWithdrawManager.signedWithdrawToUser` calldata.

The v1 path (`WITHDRAW_ENABLED` / `initiateWithdraw`) and the v1 endpoint are untouched.

## Behavior

- **Gated** behind a new `ENABLE_V3_WITHDRAWALS` env (must be exactly `"true"`), independent of v1's `WITHDRAW_ENABLED`.
- **Scope**: only `mis_route` is handled; `intent_refund` stays dropped on v3 for now.
- **Gas deduction**: sends `deductGasFromRefund: true` so refunds aren't operated at a loss.
- **Terminal 422**: a `GAS_EXCEEDS_REFUND` / `UNPRICEABLE_REFUND_TOKEN` response is treated as terminal — persisted to a new `deposit-address:skipped-withdraw-keys:<botId>` Redis set and never retried (survives handover). Every other failure (network, timeout, 5xx, transient 400) is retried, then skipped for the current poll and re-attempted next poll.
- Surfacing the 422 status required a non-swallowing `_postOrThrow` base-client method (`_post` collapses errors to `undefined`); classification is by `HttpError.status`.
- **Pub/Sub**: confirmed v3 withdraws publish the same `withdraw_executed` lifecycle event as v1 under the existing `ENABLE_DEPOSIT_ADDRESS_WITHDRAW_PUBLISHER` gate + topic. `buildWithdrawExecutedPayload` / `_publishWithdrawExecuted` generalized to `AnyDepositAddressMessage` (reads `refundAddress.address` for v3). When the publisher gate is off, publish is silently skipped — no pubsub env required for local runs.

### DIFF
--- a/src/clients/AcrossApiBaseClient.ts
+++ b/src/clients/AcrossApiBaseClient.ts
@@ -97,4 +97,29 @@ export abstract class BaseAcrossApiClient {
       return;
     }
   }
+
+  /**
+   * Like `_post`, but **rethrows** instead of collapsing failures to `undefined`. Callers that
+   * must classify the failure (e.g. distinguish a terminal 4xx from a retryable transient error)
+   * use this and inspect the thrown error — `postWithTimeout` throws a typed `HttpError` carrying
+   * the HTTP `status` on non-2xx responses. Still logs at warn for observability parity with `_post`.
+   */
+  protected async _postOrThrow<T>(endpoint: string, body: unknown): Promise<T> {
+    try {
+      const headers: Record<string, string> = { "Content-Type": "application/json", Accept: "application/json" };
+      if (this.apiKey) {
+        headers.Authorization = `Bearer ${this.apiKey}`;
+      }
+
+      return await postWithTimeout<T>(`${this.urlBase}/${endpoint}`, body, {}, headers, this.apiResponseTimeout);
+    } catch (err) {
+      this.logger.warn({
+        at: this.logContext,
+        message: `Failed to post to ${this.urlBase}`,
+        endpoint,
+        error: (err as Error).message,
+      });
+      throw err;
+    }
+  }
 }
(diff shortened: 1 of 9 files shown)


src/clients/AcrossApiBaseClient.ts
```diff
@@ -97,4 +97,29 @@ export abstract class BaseAcrossApiClient {
...
   }
+
+  /**
+   * Like `_post`, but **rethrows** instead of collapsing failures to `undefined`. Callers that
+   * must classify the failure (e.g. distinguish a terminal 4xx from a retryable transient error)
+   * use this and inspect the thrown error — `postWithTimeout` throws a typed `HttpError` carrying
+   * the HTTP `status` on non-2xx responses. Still logs at warn for observability parity with `_post`.
+   */
+  protected async _postOrThrow<T>(endpoint: string, body: unknown): Promise<T> {
+    try {
```
dijanin-brat, line 112: we can narrow down the try/catch for just the async call, we don't need try at the beginning of this function

amateima: Fixed in [31bf315](https://github.com/across-protocol/relayer/pull/3500/commits/31bf3155402e214f156fdce9a99815d6c90f370e)

dijanin-brat (approved):

PR #3524, opened 2026-06-26
chore: Exclude Universal Chains from Orbit path

dijanin-brat:


src/dataworker/Dataworker.ts
```diff
@@ -1884,8 +1885,12 @@ export class Dataworker {
     const fundedLeaves = await this._getExecutablePoolRebalanceLeaves(allLeaves, balanceAllocator);
     const executableLeaves: PoolRebalanceLeaf[] = [];
     for (const leaf of fundedLeaves) {
-      // For orbit leaves we need to check if we have enough gas tokens to pay for the L1 to L2 message.
-      if (!sdkUtils.chainIsArbitrum(leaf.chainId) && !sdkUtils.chainIsOrbit(leaf.chainId)) {
+      // Orbit chains relay roots via the Arbitrum inbox and need L1 ETH prefunding. Universal
+      // SpokePool chains (Helios messaging) finalize pool rebalances separately on L2.
+      const requiresOrbitL1ToL2MessageFees =
+        (sdkUtils.chainIsArbitrum(leaf.chainId) || sdkUtils.chainIsOrbit(leaf.chainId)) &&
+        !UNIVERSAL_CHAINS.includes(leaf.chainId);
```
pxrl, line 1892: How are the order of ops here? Should we maybe set (!UNIVERSAL_CHAINS.includes(leaf.chainId) as the first conditional and then bracket the 2nd conditional?

dijanin-brat: why? Its basically the same condition so it will aways be false if the chainId is included in UNIVERSAL_CHAINS?

pxrl (approved):

PR #3528, opened 2026-06-26
improve(InventoryClient): Broaden repayment eligibility for hub-origin deposits

droplet-rl:
## Motivation

For a deposit originating on the **hub chain**, the input liquidity lands on the hub — the rebalancing center. The relayer can therefore be repaid on *any* chain with a pool rebalance route for the token and have the hub source that rebalance cheaply. Today, hub-origin deposits are only evaluated for repayment on the **origin (hub)** and **destination** chains.

This surfaced on a USDT (Mainnet) → USDC (Base) deposit that went unfilled: the destination (Base) was over-allocated and the hub hit origin-commitment limits, so there was no acceptable repayment chain — even though the relayer was under-allocated on other chains it could have rebalanced into.

## Change

- `InventoryClient.getRepaymentChainsForL1Token()` — new helper enumerating every enabled L2 chain that has **both** inventory token config and a pool rebalance route for the L1 token.
- `determineRefundChainId()` — for hub-origin deposits, append those chains to the candidate set. The existing per-chain allocation check (`expectedPostRelayAllocation <= effectiveTargetPct`) keeps only under-allocated chains, so broadening can only *add* safe targets. Destination stays on the standard path; the hub remains the fallback.
- `getPossibleRepaymentChainIds()` — mirror the broadened set so `batchComputeLpFees` precomputes an LP fee per candidate. LP-fee differences (i.e. the cost of rebalancing to a third chain) are therefore priced into downstream profitability selection — an unprofitable third-chain repayment is filtered out.
- Token-agnostic: keyed on the deposit's L1 token, not any specific token.

### Notes
- **Cheap by construction.** This reuses the in-memory allocation check only; it does *not* reintroduce the running-balance computation removed in #3520.
- **Guard simplification.** The gate is just `originChainId === hubChainId`. `forceOriginRepayment` cannot be set at that point — a hub origin is always an unmetered fast-rebalance source, so `isUnmeteredFastRebalance()` is true and the forced-origin path short-circuits earlier in `determineRefundChainId()`.
- Fixes a stale `batchComputeLpFees` comment that still referenced the removed slow-withdrawal chains.
- **Interaction with the fast-rebalance short-circuit:** broadening engages for hub-origin deposits not already short-circuited to origin by the `canFastRebalanceFill` rule. Since the hub is typically absent from `tokenConfig`, that short-circuit doesn't fire and broadening runs (the real-world case here). If we later want broadening to also override an explicit hub token config, that's a follow-up.

## Testing

`yarn build` + lint clean. New cases in `test/InventoryClient.RefundChain.ts` (all 46 in the file pass, no regressions):
- hub-origin deposit surfaces every pool-rebalance-route chain in the possible list (and excludes a config-but-no-route chain);
- non-hub origin is unaffected (no broadening);
- a hub-origin fill refunds on under-allocated route chains that are neither origin nor destination.

Worth a sanity check on utilisation after enabling, in the same spirit as #3520.

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -527,7 +527,7 @@ export class InventoryClient {
 
   getPossibleRepaymentChainIds(deposit: Deposit): number[] {
     // Origin chain is always included in the repayment chain list.
-    const { originChainId, destinationChainId } = deposit;
+    const { originChainId, destinationChainId, inputToken } = deposit;
     const chainIds = new Set<number>();
     chainIds.add(originChainId);
 
@@ -542,6 +542,17 @@ export class InventoryClient {
       chainIds.add(destinationChainId);
     }
 
+    // For a hub-origin deposit the input liquidity lands on the hub, so the relayer can be repaid on any chain with
+    // a pool rebalance route for this token and have the hub source the rebalance cheaply. Surface them all here so
+    // their LP fees are precomputed; determineRefundChainId() applies the allocation check. forceOriginRepayment is
+    // necessarily false at this point (it returns above).
+    if (originChainId === this.hubPoolClient.chainId) {
+      const l1Token = this.getL1TokenAddress(inputToken, originChainId);
+      if (isDefined(l1Token)) {
+        this.getRepaymentChainsForL1Token(l1Token).forEach((chainId) => chainIds.add(chainId));
+      }
+    }
+
     // Check per-chain override first, then global override
     const chainOverride = this.getRepaymentChainOverride(originChainId);
     if (isDefined(chainOverride)) {
@@ -550,6 +561,23 @@ export class InventoryClient {
 
     chainIds.add(this.hubPoolClient.chainId);
     return [...chainIds];
+  }
+
+  /**
+   * @notice Returns every enabled L2 chain that has both inventory token config and a pool rebalance route for the
+   * given L1 token. Used to broaden repayment eligibility for hub-origin deposits: because the deposited liquidity
+   * is already on the hub, repaying the relayer on any of these chains doubles as a low-cost rebalance.
+   * @param l1Token L1 token to enumerate repayment routes for.
+   * @returns list of eligible L2 chain IDs.
+   */
+  getRepaymentChainsForL1Token(l1Token: EvmAddress): number[] {
+    return this.getEnabledL2Chains().filter((chainId) => {
+      if (!this._l1TokenEnabledForChain(l1Token, chainId)) {
+        return false;
+      }
+      const repaymentToken = this.getRemoteTokenForL1Token(l1Token, chainId);
+      return isDefined(repaymentToken) && this.hubPoolClient.l2TokenHasPoolRebalanceRoute(repaymentToken, chainId);
+    });
   }
 
   /**
@@ -716,7 +744,7 @@ export class InventoryClient {
     // To correctly compute destination-chain allocation, add upcoming refunds for all equivalents of this L1 token.
     const cumulativeVirtualBalancePostRefunds = this.getCumulativeBalanceWithApproximateUpcomingRefunds(l1Token);
 
-    // Repayment is evaluated on the origin and destination chains only.
+    // Standard repayment candidates are the origin and destination chains.
     const prioritizeOrigin = deposit.toLiteChain || originQuickRebalance;
     const standardChainOrder = prioritizeOrigin
       ? [originChainId, destinationChainId]
@@ -728,7 +756,18 @@ export class InventoryClient {
     const standardCandidateChains = [...standardChainOrder].filter((chainId) =>
       chainId === originChainId ? canEvaluateOrigin : canEvaluateDestination
     );
-    const rankedChains = dedupArray(standardCandidateChains);
+
+    // For a hub-origin deposit, also evaluate every other chain with a pool rebalance route for this token. The
+    // deposited liquidity is already on the hub, so repaying the relayer on an under-allocated chain is a cheap
+    // rebalance; the allocation check below keeps only under-allocated chains, and LP fee differences are priced
+    // into downstream profitability selection. Destination is handled by the standard path above. No
+    // forceOriginRepayment guard is needed: a hub origin is always an unmetered fast-rebalance source, so the
+    // forced-origin path has already returned above.
+    const broadenedCandidateChains =
+      originChainId === hubChainId
+        ? this.getRepaymentChainsForL1Token(l1Token).filter((chainId) => chainId !== destinationChainId)
+        : [];
+    const rankedChains = dedupArray([...standardCandidateChains, ...broadenedCandidateChains]);
 
     const eligibleChains: number[] = [];
     // At this point, all ranked chains have defined token configs or are destination chains that can be accepted
--- a/docs/repayment-eligibility.md
+++ b/docs/repayment-eligibility.md
@@ -69,6 +69,7 @@ This is an LP-fee precomputation helper, not final selection. It may include:
 
 - origin chain (always)
 - destination chain (if allowed)
+- for a hub-origin deposit, every chain with both inventory token config and a pool rebalance route for the deposit's L1 token (via `getRepaymentChainsForL1Token()`)
 - config override chain
 - hub chain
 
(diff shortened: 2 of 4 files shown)


src/clients/InventoryClient.ts
```diff
@@ -728,7 +756,18 @@ export class InventoryClient {
...
+    // For a hub-origin deposit, also evaluate every other chain with a pool rebalance route for this token. The
+    // deposited liquidity is already on the hub, so repaying the relayer on an under-allocated chain is a cheap
+    // rebalance; the allocation check below keeps only under-allocated chains, and LP fee differences are priced
+    // into downstream profitability selection. Destination is handled by the standard path above. No
+    // forceOriginRepayment guard is needed: a hub origin is always an unmetered fast-rebalance source, so the
+    // forced-origin path has already returned above.
+    const broadenedCandidateChains =
+      originChainId === hubChainId
+        ? this.getRepaymentChainsForL1Token(l1Token).filter((chainId) => chainId !== destinationChainId)
+        : [];
```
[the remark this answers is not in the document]
droplet-rl: `canTakeDestinationChainRepayment()` does not gate on input/output equivalence — it returns true whenever the input token's L1 mapping has a pool-rebalance route on the destination chain (`l2TokenEnabledForL1TokenAtBlock(l1Token, destinationChainId)`), regardless of the output token. So for `Mainnet USDT -> Base USDC` where Base has a USDT PoolRebalanceRoute, `canTakeDestinationChainRepayment` is true and the standard `canEvaluateDestination` path (`canTakeDestinationChainRepayment && _l1TokenEnabledForChain(l1Token, dest)`) already evaluates Base.

That standard predicate (config AND route on the destination) is exactly the predicate `getRepaymentChainsForL1Token()` uses for membership, so the destination is never a member of the broadened set that the standard path skips. The `chainId !== destinationChainId` filter only avoids double-counting the destination in the candidate list; it cannot silently drop an otherwise-eligible route. (Conversely, if Base lacked a USDT route or config, it would be excluded from *both* the standard and broadened sets — correctly, since the relayer can't manage USDT inventory there.) Leaving this as-is.

src/clients/InventoryClient.ts
```diff
@@ -542,6 +542,17 @@ export class InventoryClient {
       chainIds.add(destinationChainId);
     }
 
+    // For a hub-origin deposit the input liquidity lands on the hub, so the relayer can be repaid on any chain with
+    // a pool rebalance route for this token and have the hub source the rebalance cheaply. Surface them all here so
+    // their LP fees are precomputed; determineRefundChainId() applies the allocation check. forceOriginRepayment is
+    // necessarily false at this point (it returns above).
+    if (originChainId === this.hubPoolClient.chainId) {
```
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in 5f81d4a. Updated `docs/repayment-eligibility.md` to document the broadened hub-origin candidate set: the new `getRepaymentChainsForL1Token()` bullet under `getPossibleRepaymentChainIds()`, the candidate-build step in the `determineRefundChainId()` walkthrough, and a dedicated "hub-origin broadening" entry under Subtle behaviors covering the rationale, the allocation-check guard, and the LP-fee mirroring.

src/clients/InventoryClient.ts
```diff
@@ -552,6 +563,23 @@ export class InventoryClient {
...
+   * @param l1Token L1 token to enumerate repayment routes for.
+   * @returns list of eligible L2 chain IDs.
+   */
+  getRepaymentChainsForL1Token(l1Token: EvmAddress): number[] {
+    return this.getEnabledL2Chains().filter((chainId) => {
+      if (!this._l1TokenEnabledForChain(l1Token, chainId)) {
+        return false;
+      }
+      const repaymentToken = this.getRemoteTokenForL1Token(l1Token, chainId);
+      return isDefined(repaymentToken) && this.hubPoolClient.l2TokenHasPoolRebalanceRoute(repaymentToken, chainId);
```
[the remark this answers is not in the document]
droplet-rl: The helper deliberately mirrors the canonical-token resolution that the downstream allocation loop in `determineRefundChainId()` already uses: for each ranked non-origin chain the loop resolves `repaymentToken = getRemoteTokenForL1Token(l1Token, chainId)` (the same singular/canonical lookup) and then asserts `l2TokenHasPoolRebalanceRoute(repaymentToken, chainId)`.

If `getRepaymentChainsForL1Token()` instead enumerated alias tokens via `getRemoteTokensForL1Token()` and admitted a chain whose route lives only on a non-canonical alias (e.g. USDC.e/USDbC), that chain would enter the loop, the loop would still resolve the canonical token (which has no route), and the assert would throw — turning a silent omission into a hard crash. So the suggested change is unsafe unless the loop's repayment-token resolution is reworked to pick the alias that actually carries the route.

That canonical-only resolution is pre-existing across the entire origin/destination allocation path, not introduced by this PR, so broadening alias-token handling is a separate change with its own test surface. I'm keeping the helper aligned with the loop here, and have noted the alias limitation for a follow-up.

src/clients/InventoryClient.ts
```diff
@@ -542,6 +542,17 @@ export class InventoryClient {
...
     }
 
+    // For a hub-origin deposit the input liquidity lands on the hub, so the relayer can be repaid on any chain with
+    // a pool rebalance route for this token and have the hub source the rebalance cheaply. Surface them all here so
+    // their LP fees are precomputed; determineRefundChainId() applies the allocation check. forceOriginRepayment is
+    // necessarily false at this point (it returns above).
+    if (originChainId === this.hubPoolClient.chainId) {
+      const l1Token = this.getL1TokenAddress(inputToken, originChainId);
+      if (isDefined(l1Token)) {
+        this.getRepaymentChainsForL1Token(l1Token).forEach((chainId) => chainIds.add(chainId));
```
pxrl, line 552: @droplet-rl Do we subsequently dedup here? It'd be nice to avoid evaluating the same chainIds repeatedly.

src/clients/InventoryClient.ts
pxrl, line 579: @droplet-rl Can any of these methods throw exceptions?

src/clients/InventoryClient.ts
```diff
@@ -542,6 +542,17 @@ export class InventoryClient {
       chainIds.add(destinationChainId);
     }
 
+    // For a hub-origin deposit the input liquidity lands on the hub, so the relayer can be repaid on any chain with
+    // a pool rebalance route for this token and have the hub source the rebalance cheaply. Surface them all here so
+    // their LP fees are precomputed; determineRefundChainId() applies the allocation check. forceOriginRepayment is
+    // necessarily false at this point (it returns above).
+    if (originChainId === this.hubPoolClient.chainId) {
+      const l1Token = this.getL1TokenAddress(inputToken, originChainId);
```
pxrl, line 550: Given that the originChainId is known to be the hubPool chain, we already know that inputToken is equal to l1Token. Therefore, l1Token lookup is redundant. Agree?

src/clients/InventoryClient.ts
droplet-rl: Good question — I traced all four, and none throw on real inputs:

- `getEnabledL2Chains()` — pure `.filter()` over the chain-id list. Can't throw.
- `_l1TokenEnabledForChain()` — optional-chained `inventoryConfig` reads plus `isAliasConfig`/`Object.keys().some()`. Can't throw.
- `getRemoteTokenForL1Token()` (the protected wrapper → SDK `getRemoteTokenForL1Token`) — returns `undefined` for any unknown mapping rather than throwing (it uses the non-asserting 2-arg `resolveAcrossToken`, not the `assertExists` overload), and the `isDefined(repaymentToken)` guard handles that. The only theoretical throw is `toAddressType()` on a malformed address string, but that address comes from the curated `TOKEN_SYMBOLS_MAP`.
- `l2TokenHasPoolRebalanceRoute()` — just `isDefined(getL1TokenForL2TokenAtBlock(...))`; returns a boolean, never throws.

For the two SDK-backed ones this isn't new exception surface: `determineRefundChainId()` already calls `this.getRemoteTokenForL1Token(l1Token, chainId)` for every candidate chain (L778) and asserts `l2TokenHasPoolRebalanceRoute(...)` (L782), and `canTakeDestinationChainRepayment()` calls `l2TokenHasPoolRebalanceRoute()` (L654) — all unguarded. If either could throw on live data, that path would already be crashing.

The one token-resolution helper in this file that genuinely throws — `getInventoryEquivalentL1TokenAddress` — is exactly why `getL1TokenAddress()` wraps it in try/catch, and the `l1Token` passed into `getRepaymentChainsForL1Token()` has already been resolved through that guard (`getL1TokenAddress` at the `getPossibleRepaymentChainIds` call site; `getRequiredL1TokenAddress` after `validateOutputToken` in `determineRefundChainId`).

So I don't think a try/catch is warranted here — it would be inconsistent with the surrounding unguarded calls and risk silently swallowing a real misconfiguration. Happy to add a defensive guard if you'd prefer belt-and-suspenders, though.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=7696bc385f7bdc27f95f4fa1a154820e&trace_start_ts=1782486943&trace_end_ts=1782573643)

PR #3530, opened 2026-06-29
feat: Add Paxos Tranist API for rebalancing

dijanin-brat:


src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -121,11 +122,11 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     assert(toAddress.isEVM(), "Paxos Transit finalization events require an EVM destination address");
     const l2Provider = this.l2Bridge?.provider;
     assert(isDefined(l2Provider), "PaxosTransitBridge: l2 provider is required");
-    const transitStation = getPaxosTransitStationAddress(this.l2chainId);
+    const boringVault = getPaxosTransitBoringVaultAddress(this.l2chainId);
```
dijanin-brat, line 144: When we do rebalancing, we are sending tokens on origin chain to transitStation contract but we are receiving tokens on destination chain from boringVault contract

src/common/Constants.ts
```diff
@@ -729,7 +731,7 @@ export const CUSTOM_L2_BRIDGE: Record<number, Record<string, L2BridgeConstructor
     [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: L2SolanaUsdcCCTPBridge,
   },
   [CHAIN_IDs.ROBINHOOD]: {
-    // [TOKEN_SYMBOLS_MAP.USDG.addresses[CHAIN_IDs.MAINNET]]: PaxosTransitL2Bridge,
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: PaxosTransitL2Bridge,
```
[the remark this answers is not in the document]
dijanin-brat: [4c44c23](https://github.com/across-protocol/relayer/pull/3530/commits/4c44c233b723b34ac889e3c2bf680f6b554a63a0)

src/common/Constants.ts
```diff
@@ -546,7 +548,7 @@ export const CUSTOM_BRIDGE: Record<number, Record<string, L1BridgeConstructor<Ba
     [TOKEN_SYMBOLS_MAP.USDT.addresses[CHAIN_IDs.MAINNET]]: OFTBridge,
   },
   [CHAIN_IDs.ROBINHOOD]: {
-    // [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: PaxosTransitBridge,
+    [TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.MAINNET]]: PaxosTransitBridge,
```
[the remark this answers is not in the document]
dijanin-brat: @codex
i added new changes for your 2 previous comments: [4c44c23](https://github.com/across-protocol/relayer/pull/3530/commits/4c44c233b723b34ac889e3c2bf680f6b554a63a0)

Can you re-review this PR again, but keep in mind that:
- we will bridge both USDC and USDG from L1 to USDG on Robinhood
- We will bridge USDG from Robinhood to L1 USDC (always USDC)
- there is a fee for that bridging, so will this change help us resolve the amount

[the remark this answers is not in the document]
dijanin-brat: We will send USDG from L1 to RH only via refiller bot so this shouldn't affect it

src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -121,15 +180,55 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
...
+    const l1TokenDecimals = this.l1TokenInfo.decimals;
+    const routeParams = {
+      wantAsset: this.l2TokenAddress,
+      sourceChainId: this.hubChainId,
+      destinationChainId: this.l2chainId,
+      receiver: userAddress.toNative(),
+    };
+
+    const outstandingOrders = orders.filter(
+      (order) => paxosTransitOrderMatchesRoute(order, routeParams) && isPaxosTransitOrderOutstanding(order)
```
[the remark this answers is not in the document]
dijanin-brat: that is okey 👍

src/adapter/bridges/BaseBridgeAdapter.ts
```diff
@@ -102,6 +104,16 @@ export abstract class BaseBridgeAdapter {
...
 
+  /** Optional hook when on-chain initiation amount exceeds quoted destination receive amount. */
+  recordL1ToL2BridgeInitiation?(l1TxnHash: string, expectedDestinationReceiveAmount: BigNumber): Promise<void>;
+
+  /**
+   * When implemented, returns net outstanding L1→L2 bridge transfers from an off-chain API.
+   * Amounts are in L1 token decimals; finalization is already reflected in each event amount.
+   * On success, callers should skip on-chain initiation/finalization event queries.
+   */
+  getOutstandingTransfersFromApi?(l1Token: EvmAddress, monitoredAddress: Address): Promise<BridgeEvents>;
```
dijanin-brat, line 110: Maybe add this to the BridgeApi as well

src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -80,35 +92,82 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
...
+      spenderAddress: getPaxosTransitStationAddress(this.hubChainId),
     });
 
+    const expectedDestinationReceiveAmount = isDefined(orderData.amountOut)
+      ? BigNumber.from(orderData.amountOut)
+      : undefined;
+
     return {
       // @TODO: We should use the actual ABI of the contract. This is just a placeholder.
       contract: new Contract(orderData.transaction.to, [], this.l1Signer),
```
dijanin-brat, line 96: Not really sure if we should add ABI here because we are sending the calldata directly.

src/common/Constants.ts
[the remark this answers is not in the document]
bmzig: Will this still break when attempting to rebalance? Specifically because `USDG` is a `SUPPORTED_TOKEN` and this: https://github.com/across-protocol/relayer/blob/cbdd6e4de65c0781873982308e0acd7f5cd551ee/src/clients/bridges/AdapterManager.ts#L104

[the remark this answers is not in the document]
dijanin-brat: I think it shouldn't because we not have inventory config for USDG as standalone token (it will be inside the USDC inventory config)

src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -80,35 +92,82 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
...
     const l1Provider = this.l1Signer.provider;
     assert(isDefined(l1Provider), "PaxosTransitBridge: l1Signer must have a provider");
     const transitStation = getPaxosTransitStationAddress(this.hubChainId);
-    const tokenContract = new Contract(l1Token.toNative(), ERC20_ABI, l1Provider);
-    const events = await paginatedEventQuery(
-      tokenContract,
-      tokenContract.filters.Transfer(fromAddress.toNative(), transitStation),
-      eventConfig
-    );
+    const offerAssets = getPaxosTransitOfferAssetsForWantAsset(this.l2chainId, this.l2TokenAddress);
```
bmzig, line 118: Why do we have to track all this in Redis, add a new `expectedDestinationReceiveAmount` field, and two new functions for this bridge? Why couldn't we just track the transfers into the bridge, and then as we do on the L2 side, track the transfers out?

dijanin-brat: Because there will be a difference between input/output amount because of the bridge fee, so when we try to calculate the outstanding amount there will always be a difference between those (so the bots will wrongly assume there is still something that is not finalized). That is the reason for the new field as well.

For 2 new functions:
- `recordL1ToL2BridgeInitiation` - I wanted to create something that could be re-used in future so we can easily add new bridges that have fees
- `getOutstandingTransfersFromApi` - similar to the previous method, i wanted to add something we could re-use when we add bridges that could be queried via API (but if the API is down, we can always have a fallback with existing way where we look for the on-chain data)

bmzig: How big is this difference? We get trace amounts on the Binance bridge and can live with it usually since the amounts are a very small percentage of the total amount rebalanced. 

I'm basically just making sure that adding this is necessary since it does come with a considerably large complexity cost.

src/common/Constants.ts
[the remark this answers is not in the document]
bmzig: I think that might only work for L2 bridges, but not L1 bridges. But I suppose we can always address later if it is a problem.

src/adapter/bridges/PaxosTransitBridge.ts
dijanin-brat: hmm, that is a good question tho. I think the fees are around 3.5-5bps, so maybe we can just leave it i guess, wdyt?

For the API outstanding balance function, i still think that is useful and we should leave that

bmzig: I am not sure. I generally dislike using Redis for stuff like this so my opinion is to try and simplify this so we don't have to use Redis (and then getting some of those functions/extra fields out is a bonus), but this is just my opinion. If you feel strongly about keeping it in, then we can, I am just not a big fan of it.

dijanin-brat: Well for the `getOutstandingTransfersFromApi`, i think that is useful and that we should leave it

For the `recordL1ToL2BridgeInitiation` i agree it adds complexity and that we can remove it and live with the input/output amount difference.

dijanin-brat: here it is [d90fd2b](https://github.com/across-protocol/relayer/pull/3530/commits/d90fd2b08c0d69dd52e27d2bf3578b8011295b27)

bmzig (approved):

src/adapter/bridges/PaxosTransitBridge.ts
bmzig: Nice!

PR #3534, opened 2026-07-01
feat(relayer): per-destination-chain recipient allow-list

droplet-rl:
## What

Adds `RELAYER_ALLOWED_RECIPIENTS_<chainId>` — a per-destination-chain recipient allow-list for the relayer, as requested for the Robinhood (chainId 4663) launch.

The value is a JSON string array of recipient addresses:

```
RELAYER_ALLOWED_RECIPIENTS_4663=["0xAbc...","0xDef..."]
```

When an allow-list is configured for a chain, `Relayer::filterDeposit` drops every deposit to that chain whose `recipient` is not on the list. Chains with **no** configured allow-list are unrestricted (unchanged behavior). This mirrors the existing `addressFilter` blocklist pattern that `filterDeposit` already uses.

## The two things Paul asked us to consider

**1. Deposits that specify the MulticallHandler → dropped when an allow-list is active.**

A deposit carrying a message is executed on the destination chain by the *recipient contract* (typically the `MulticallHandler`), whose message is `tuple(tuple(address target, bytes callData, uint256 value)[] calls, address fallbackRecipient)`. The handler can forward the bridged funds to **arbitrary** addresses via its encoded `calls`, so the on-chain `recipient` field is not the effective beneficiary and the allow-list cannot be soundly enforced. The safe, correct behavior for an access-control gate is therefore to drop message deposits to an allow-listed chain.

A future enhancement *could* opt in to validating a `MulticallHandler` message's `fallbackRecipient` against the allow-list — but since the `calls` can still redirect funds regardless of `fallbackRecipient`, that would be best-effort, not a guarantee. I left it out deliberately so the control has no false-confidence bypass; happy to add it behind an explicit opt-in flag if we want composable flows for allow-listed users.

**2. Should we also filter on `depositor`? → No.**

The `depositor` is set by the user on the origin chain and is trivially spoofable (a blocked party can set `depositor` to an allow-listed address while keeping `recipient` as their own). Gating on it would provide **no** access-control guarantee and would create a false sense of security. Only the destination-chain `recipient` — which is enforced on-chain by the fill — is a meaningful control, so the filter keys exclusively on it.

## Implementation

- `RelayerConfig`: new `allowedRecipients: { [chainId: number]: Set<string> }`, populated in `validate()` from `RELAYER_ALLOWED_RECIPIENTS_<chainId>` using the existing per-chain env pattern (`RELAYER_MIN_FILL_TIME_<chainId>`, `SEND_MESSAGE_RELAYS_<chainId>`). Addresses are normalised via `toAddressType(addr, chainId).toNative()` for O(1) membership checks.
- `Relayer::filterDeposit`: new allow-list gate placed right after the `addressFilter` blocklist. Uses optional chaining so a config without the field (e.g. test harness) is a no-op. Rejections use `ignoreDeposit()` since the recipient can't change (permanently un-fillable), matching the blocklist.
- Docs: new "Restricting recipients on a destination chain" section in `src/relayer/README.md`.

## Tests

Added to `test/Relayer.BasicFill.ts` (all 22 passing):
- Drops a deposit to a non-allow-listed recipient.
- Fills a deposit to an allow-listed recipient.
- Drops a message deposit to an allow-list-restricted chain even when the recipient is allow-listed.

`yarn tsc --noEmit`, `yarn eslint`, and `yarn prettier` all clean.

### DIFF
--- a/src/relayer/README.md
+++ b/src/relayer/README.md
@@ -22,14 +22,21 @@ RELAYER_ALLOWED_RECIPIENTS_4663=["0xabc...","0xdef..."]
 ```
 
 When an allow-list is configured for a chain, `Relayer::filterDeposit` drops every deposit to that chain whose
-`recipient` is not on the list. Chains without a configured allow-list are unrestricted.
+recipient is not on the list. Chains without a configured allow-list are unrestricted. The check validates the
+*effective* recipient the fill will actually pay out to: a deposit that has been sped up is filled via
+`fillRelayWithUpdatedDeposit`, which delivers the funds to the depositor-signed `updatedRecipient` rather than the
+original `recipient`, so the allow-list is enforced against `updatedRecipient` when a speed-up is present. This
+prevents a depositor from depositing to an allow-listed recipient and then signing a speed-up that redirects the funds
+to a non-allow-listed address.
 
 Two deliberate design choices:
 
 - **Deposits carrying a message are always dropped when an allow-list is active.** Such deposits are executed on the
   destination chain by the recipient contract (typically the `MulticallHandler`), which can forward the funds to
   arbitrary addresses via its encoded calls. The on-chain `recipient` is therefore not the effective beneficiary, so
-  the allow-list cannot be enforced and the deposit is not filled. (A future enhancement could opt in to validating a
+  the allow-list cannot be enforced and the deposit is not filled. As with the recipient, the *effective* message is
+  evaluated — a speed-up that adds a message causes the deposit to be dropped, while one that nullifies the original
+  message allows it (subject to the recipient check). (A future enhancement could opt in to validating a
   `MulticallHandler` message's `fallbackRecipient`, but note that the message's calls can still redirect funds, so
   that check would be best-effort rather than a guarantee.)
 - **The filter never consults `depositor`.** The depositor is set by the user on the origin chain and is trivially
--- a/src/relayer/README.md
+++ b/src/relayer/README.md
@@ -14,34 +14,15 @@ The full config can be found in `RelayerConfig.ts`
 
 ### Restricting recipients on a destination chain
 
-`RELAYER_ALLOWED_RECIPIENTS_<chainId>` restricts which recipients the relayer will fill for on a given destination
-chain. The value is a JSON string array of recipient addresses, e.g.:
+`RELAYER_ALLOWED_RECIPIENTS_<chainId>` is a JSON array of recipient addresses the relayer will fill for on that
+destination chain, e.g. `RELAYER_ALLOWED_RECIPIENTS_4663=["0xabc...","0xdef..."]`. When set, `Relayer::filterDeposit`
+drops every deposit to that chain whose recipient is not listed; chains without a list are unrestricted.
 
-```
-RELAYER_ALLOWED_RECIPIENTS_4663=["0xabc...","0xdef..."]
-```
-
-When an allow-list is configured for a chain, `Relayer::filterDeposit` drops every deposit to that chain whose
-recipient is not on the list. Chains without a configured allow-list are unrestricted. The check validates the
-*effective* recipient the fill will actually pay out to: a deposit that has been sped up is filled via
-`fillRelayWithUpdatedDeposit`, which delivers the funds to the depositor-signed `updatedRecipient` rather than the
-original `recipient`, so the allow-list is enforced against `updatedRecipient` when a speed-up is present. This
-prevents a depositor from depositing to an allow-listed recipient and then signing a speed-up that redirects the funds
-to a non-allow-listed address.
-
-Two deliberate design choices:
-
-- **Deposits carrying a message are always dropped when an allow-list is active.** Such deposits are executed on the
-  destination chain by the recipient contract (typically the `MulticallHandler`), which can forward the funds to
-  arbitrary addresses via its encoded calls. The on-chain `recipient` is therefore not the effective beneficiary, so
-  the allow-list cannot be enforced and the deposit is not filled. As with the recipient, the *effective* message is
-  evaluated — a speed-up that adds a message causes the deposit to be dropped, while one that nullifies the original
-  message allows it (subject to the recipient check). (A future enhancement could opt in to validating a
-  `MulticallHandler` message's `fallbackRecipient`, but note that the message's calls can still redirect funds, so
-  that check would be best-effort rather than a guarantee.)
-- **The filter never consults `depositor`.** The depositor is set by the user on the origin chain and is trivially
-  spoofable, so gating on it would provide no access-control guarantee. Only the destination-chain `recipient` — which
-  is enforced on-chain by the fill — is a meaningful control.
+The check evaluates the *effective* recipient the fill pays out to — `updatedRecipient` for sped-up deposits, since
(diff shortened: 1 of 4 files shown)


src/relayer/Relayer.ts
```diff
@@ -281,6 +282,32 @@ export class Relayer {
...
+    // whose funds are provably delivered to an allow-listed recipient are permitted; everything else is dropped.
+    const chainAllowedRecipients = allowedRecipients?.[destinationChainId];
+    if (isDefined(chainAllowedRecipients)) {
+      // Deposits carrying a message are executed on the destination chain by the recipient contract (typically the
+      // MulticallHandler), which can forward the funds on to arbitrary addresses via its encoded calls. For those the
+      // `recipient` field is not the effective beneficiary, so the allow-list cannot be enforced and the deposit is
+      // dropped. The `depositor` is deliberately NOT consulted: it is set by the user on the origin chain and is
+      // trivially spoofable, so gating on it would provide no access-control guarantee.
+      const depositHasMessage = !isMessageEmpty(resolveDepositMessage(deposit));
+      if (depositHasMessage || !chainAllowedRecipients.has(recipient.toNative())) {
```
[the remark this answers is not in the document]
droplet-rl: Good catch — this is a real bypass. A sped-up deposit is filled via `fillRelayWithUpdatedDeposit`, which pays out to the depositor-signed `updatedRecipient`, so validating only the original `recipient` let a depositor deposit to an allow-listed address and then redirect the funds via a speed-up (empty message) to a non-allow-listed one.

Fixed in de188fd5: the check now validates the *effective* recipient the fill will actually use — `effectiveRecipient = isDepositSpedUp(deposit) ? deposit.updatedRecipient : recipient`. The message side was already correct: `resolveDepositMessage()` returns the updated message, so a speed-up that adds a message is still dropped.

Added two tests: one asserting a deposit to an allow-listed recipient sped up to a non-allow-listed `updatedRecipient` is dropped, and one asserting a speed-up to a *different* allow-listed recipient still fills.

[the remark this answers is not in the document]
droplet-rl: This is already addressed in `de188fd5` (this review ran against the original `33e9542d81`, before that commit landed). The allow-list check now validates the effective payout recipient: `effectiveRecipient = isDepositSpedUp(deposit) ? deposit.updatedRecipient : recipient`, so a deposit named to an allow-listed recipient but sped up to a non-allow-listed `updatedRecipient` is dropped. Covered by the `Drops deposits sped up to redirect funds to a non-allow-listed recipient` test in `Relayer.BasicFill`. See the earlier thread on this same line for details.

src/relayer/README.md
pxrl, line 1: @droplet-rl Commentary in README.md + comments in the ts files are way too verbose. Make them very concise - max a few lines per item.

droplet-rl: Trimmed in `ca16c229` — the README section is now two short paragraphs, and the code comments in `Relayer.ts` / `RelayerConfig.ts` are down to a few lines each (net −32 lines of commentary), keeping only the key points: effective/updated recipient, message deposits dropped, depositor not consulted. No behavior change; tests still green.

src/relayer/Relayer.ts
```diff
@@ -281,6 +282,31 @@ export class Relayer {
...
     }
 
+    // Enforce the per-destination-chain recipient allow-list, if configured. Validate the effective recipient/message
+    // the fill will actually use: sped-up deposits fill via fillRelayWithUpdatedDeposit, so gate on updatedRecipient
+    // (else a speed-up could redirect funds off the list). Message deposits are always dropped since they execute on
+    // the recipient contract (e.g. MulticallHandler), which can forward funds on. depositor is never consulted — it's
+    // user-set on the origin chain and spoofable.
+    const chainAllowedRecipients = allowedRecipients?.[destinationChainId];
+    if (isDefined(chainAllowedRecipients)) {
+      const depositHasMessage = !isMessageEmpty(resolveDepositMessage(deposit));
```
pxrl: @droplet-rl Whether there's a message is redundant here - by filtering EOAs we can control whether the MulticallHandler is an accepted recipient or not. Prefer to drop any consideration of message here - focus only on the recipients that are allowed.

src/relayer/RelayerConfig.ts
```diff
@@ -56,6 +56,9 @@ export class RelayerConfig extends CommonConfig {
   readonly relayerMessageGasMultiplier: BigNumber;
   readonly minRelayerFeePct: BigNumber;
   readonly minFillTime: { [chainId: number]: number } = {};
+  // Per-destination-chain recipient allow-lists from RELAYER_ALLOWED_RECIPIENTS_<chainId>. When set, only deposits to
+  // an allow-listed recipient on that chain are filled; unset chains are unrestricted. See Relayer::filterDeposit.
```
pxrl: Drop comment.

```suggestion
```

src/relayer/Relayer.ts
droplet-rl: Good point — done in `7b3123c5`. Dropped the message consideration entirely; the allow-list now checks only the effective recipient. Operators control whether message-recipient contracts (e.g. `MulticallHandler`) are accepted by choosing whether to allow-list their addresses, and messages remain gated separately by `sendingMessageRelaysEnabled`, so the special-case was redundant. Repurposed the message test to assert message deposits are recipient-gated like any other; full suite still green.

src/relayer/Relayer.ts
```diff
@@ -281,6 +277,25 @@ export class Relayer {
       return ignoreDeposit();
     }
 
+    // Enforce the per-destination-chain recipient allow-list, if configured. Gate on the effective recipient the fill
+    // pays out to — updatedRecipient for sped-up deposits, which fill via fillRelayWithUpdatedDeposit — so a speed-up
+    // can't redirect funds off the list. depositor is never consulted: it's user-set on the origin chain and spoofable.
+    const chainAllowedRecipients = allowedRecipients?.[destinationChainId];
+    if (isDefined(chainAllowedRecipients)) {
+      const effectiveRecipient = isDepositSpedUp(deposit) ? deposit.updatedRecipient : recipient;
+      if (!chainAllowedRecipients.has(effectiveRecipient.toNative())) {
```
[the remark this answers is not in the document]
droplet-rl: This is a deliberate design decision rather than an oversight — @pxrl explicitly asked to drop the message consideration from this filter in the prior review round (https://github.com/across-protocol/relayer/pull/3534#discussion_r3505719809), so I removed it in 7b3123c5.

The rationale: this is a *recipient* allow-list — it gates which addresses the relayer will pay out to. Whether an allow-listed contract subsequently forwards funds via an arbitrary-call message is the operator's opt-in: if you don't want onward-forwarding you simply don't allow-list `MulticallHandler` (or any other arbitrary-call recipient). Allow-listing such a contract is an explicit statement that its downstream behaviour is trusted. Messages are also independently gated by `SEND_MESSAGE_RELAYS` / `sendingMessageRelaysEnabled`, so an operator who wants no message fills at all already has that lever.

Rejecting all non-empty messages here would remove the operator's ability to allow-list a trusted handler, which is the behaviour @pxrl asked for. Leaving the code as-is; @pxrl please re-flag if you'd prefer to reinstate a message guard given Codex's framing.

dijanin-brat (approved):

PR #3536, opened 2026-07-01
fix: Wait for CCTP Solana mint confirmation before acking

dohaki:
## Problem

The CCTP finalizer acks the Pub/Sub burn message even when the Solana mint transaction was **dropped** and never landed on-chain.

Observed in prod — the finalizer logged a solscan URL for a tx that never confirmed, then reported success:
```
Processing burn transaction        13:31:09.737
Burn transaction processed successfully   13:31:25.779   (~16s later)
```

## Root cause

`_sendAndPoll` (`src/utils/TransactionUtils.ts`) exits its poll loop on *either* confirmation *or* timeout, and returns the (locally-derived) signature in both cases. A dropped tx never confirms, so the loop burns all 25 × 600ms ≈ 15s (matching the log gap) and then hands back the signature with no error. The finalizer logs success → returns `{ success: true }` → `app.ts` returns `204` → Pub/Sub acks → **no retry**.

Solana signatures are known before submission, which is why the exact signature shows up in the log for a tx that never landed.

## Fix (scoped to cctp-finalizer)

New `sendAndConfirmMintOrThrow` guard routes both the V1 and V2 mint sends through `sendAndConfirmSolanaTransactionWithSlot` and throws when the tx is never confirmed (`confirmedSlot === undefined`). The throw surfaces as a retryable `MintTransactionFailedError` → app returns `500` → Pub/Sub redelivers. `checkIfAlreadyProcessed` guards against a double-mint if the original was merely slow rather than dropped.

The shared `_sendAndPoll` is intentionally left untouched to keep the blast radius to the finalizer.

### Known limitation
Covers dropped/unconfirmed only. A landed-but-reverted tx (`entry.err` set with `confirmationStatus: "confirmed"`) still reads as success — flagged in a `ponytail:` comment; add an `err` check in the shared poller if that failure mode surfaces.

## Test

`test/cctpFinalizerSvm.confirm.ts` (3 passing):
- documents the root cause (shared helper resolves even when never confirmed)
- guard throws on a dropped tx
- guard returns the signature on confirmation

typecheck, eslint, prettier all clean.

### DIFF
--- a/src/cctp-finalizer/utils/svmUtils.ts
+++ b/src/cctp-finalizer/utils/svmUtils.ts
@@ -18,8 +18,9 @@ import { fetchAddressLookupTable } from "@solana-program/address-lookup-table";
 import {
   winston,
   SvmAddress,
-  sendAndConfirmSolanaTransaction,
+  sendAndConfirmSolanaTransactionWithSlot,
   SVMProvider,
+  SolanaTransaction,
   TOKEN_SYMBOLS_MAP,
   getAssociatedTokenAddress,
 } from "../../utils";
@@ -177,6 +178,36 @@ export function getSvmProvider(rpcUrl: string): SVMProvider {
   return createSolanaRpcFromTransport(transport);
 }
 
+/**
+ * Sends a mint tx and only returns once it is confirmed on-chain. The shared
+ * `sendAndConfirmSolanaTransaction*` helpers return the (locally-derived)
+ * signature even when the tx was never confirmed (e.g. dropped after blockhash
+ * expiry). In the finalizer that silently acks the Pub/Sub message for a mint
+ * that never landed, so treat an unconfirmed result (`confirmedSlot` undefined,
+ * which the poller only leaves unset when it never observed confirmed/finalized)
+ * as a hard failure and let Pub/Sub redeliver for a retry.
+ */
+// ponytail: covers dropped/unconfirmed only. A landed-but-reverted tx (entry.err
+// set with confirmationStatus "confirmed") still reads as success — add an err
+// check in the shared poller if that failure mode surfaces.
+export async function sendAndConfirmMintOrThrow(
+  tx: Parameters<typeof sendAndConfirmSolanaTransactionWithSlot>[0],
+  provider: SVMProvider,
+  cycles?: number,
+  pollingDelay?: number
+): Promise<string> {
+  const { signature, confirmedSlot } = await sendAndConfirmSolanaTransactionWithSlot(
+    tx,
+    provider,
+    cycles,
+    pollingDelay
+  );
+  if (confirmedSlot === undefined) {
+    throw new Error(`Solana mint transaction ${signature} was not confirmed on-chain (likely dropped)`);
+  }
+  return signature;
+}
+
 async function hasCCTPV2MessageBeenProcessed(message: string, provider: SVMProvider): Promise<boolean> {
   const parsedMessage = parseCctpV2Message(message);
   const { usedNonce } = await getMessageTransmitterAccounts(parsedMessage.nonceSeed);
@@ -184,7 +215,7 @@ async function hasCCTPV2MessageBeenProcessed(message: string, provider: SVMProvi
   return Boolean(nonceInfo.value);
 }
 
-async function sendAndConfirmCCTPV2ReceiveMessageTx(params: {
+export async function buildCctpV2ReceiveMessageTx(params: {
   provider: SVMProvider;
   signer: KeyPairSigner;
   attestation: string;
@@ -192,10 +223,8 @@ async function sendAndConfirmCCTPV2ReceiveMessageTx(params: {
   destinationChainId: number;
   originChainId: number;
   addressLookupTable: SolanaAddress;
-  logger: winston.Logger;
-}): Promise<string> {
-  const { provider, signer, attestation, message, destinationChainId, originChainId, addressLookupTable, logger } =
-    params;
+}): Promise<SolanaTransaction> {
+  const { provider, signer, attestation, message, destinationChainId, originChainId, addressLookupTable } = params;
 
   const parsedMessage = parseCctpV2Message(message);
   const [messageTransmitterAccounts, tokenMessengerAccounts] = await Promise.all([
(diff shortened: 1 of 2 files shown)


src/cctp-finalizer/utils/svmUtils.ts
```diff
@@ -337,7 +380,7 @@ export async function processMintSvm(
       1, // hubChainId
       SvmAddress.from(recipient)
     );
-    txSignature = await sendAndConfirmSolanaTransaction(receiveMessageTx, svmProvider);
+    txSignature = await sendAndConfirmMintOrThrow(receiveMessageTx, svmProvider);
```
amateima, line 383: I think our messages use CCTP V2, not V1

amateima: so maybe 395 it's the line that needs to be modified

droplet-rl (commented):
🤖 Babysit self-review — no blocking issues found. Validated locally on a fresh checkout: `tsc --build` ✓, `eslint` on both changed files ✓, and `hardhat test test/cctpFinalizerSvm.confirm.ts` → **3 passing** ✓.

**The fix is sound.** `sendAndConfirmMintOrThrow` keys off `confirmedSlot === undefined`, and that sentinel is trustworthy: in `_sendAndPoll` (`src/utils/TransactionUtils.ts`), `confirmedSlot` is set *only* inside the `if (confirmed)` branch (`confirmationStatus === "confirmed" | "finalized"`), and `context.slot` is always present on a real `getSignatureStatuses` response, so a confirmed tx always yields a defined slot. `undefined` therefore means the poll window elapsed with no confirmation → dropped, which is exactly the case being caught. Both mint paths (V1 in `processMintSvm`, V2 in `sendAndConfirmCCTPV2ReceiveMessageTx`) now route through the guard, and `MintTransactionFailedError.shouldRetry === true` → 500 → Pub/Sub redelivers as intended.

**Double-mint safety is adequately covered.** On redelivery, `CCTPService.processBurnTransaction` runs `checkIfAlreadyProcessed` before `processMint`, and if that check races (it swallows errors and returns `false`), the on-chain `usedNonce` PDA is the hard backstop — a duplicate `receiveMessage` reverts rather than double-minting. Worth being aware that the 25×600ms ≈ 15s poll window is shorter than Solana blockhash validity (~60–90s), so a tx flagged "dropped" can still land afterward; funds stay safe via the nonce PDA, and the reverted duplicate is the already-acknowledged landed-but-reverted limitation in the `ponytail:` comment. No change needed — just noting the boundary.

**Non-blocking housekeeping (not a code issue):** the branch is behind `master` — `master` gained `#3531` (`scripts/recoveries/2026-06-accounting/**`) after this branch forked, so the two-dot `git diff master` spuriously shows those recovery files as deleted. I simulated a merge into `master` and confirmed all three files are preserved, so there's **no** data-loss risk on merge. A rebase/merge of `master` would clean up the diff view for reviewers.

No inline comments — the diff is minimal and correct as written.

src/cctp-finalizer/utils/svmUtils.ts
dohaki: good point. @droplet-rl do the fix

dohaki: @amateima ah both go through this new helper https://github.com/across-protocol/relayer/pull/3536/changes#diff-f7b88e90f80405e6fcbff993c2f4cf036464aeb19d934f6e5e2b829db92be737R279

droplet-rl: You're right that prod messages are V2 — and good news, **the V2 path is already routed through the confirmation guard** in this PR. It's just not visible in the `processMintSvm` diff because the change lives one level down, inside the `sendAndConfirmCCTPV2ReceiveMessageTx` helper rather than inline at line 395.

Trace for the V2 branch:
- `processMintSvm` (line ~395, `version !== 0`) → `sendAndConfirmCCTPV2ReceiveMessageTx({...})`
- `sendAndConfirmCCTPV2ReceiveMessageTx` (line **279**) → `const signature = await sendAndConfirmMintOrThrow(tx, provider);`

That `sendAndConfirmMintOrThrow` call is exactly what this PR changed there — the diff swapped the old non-throwing `sendAndConfirmSolanaTransaction(tx, provider)` for the throwing guard on that same line. So a dropped V2 mint now throws → `MintTransactionFailedError` (`shouldRetry = true`) → 500 → Pub/Sub redelivers, same as V1.

I grepped the whole `src/cctp-finalizer/` and there are **no** remaining non-throwing `sendAndConfirmSolanaTransaction(` call sites — both branches (V1 at line 383, V2 via the helper at 279) go through the guard. The unit tests exercise `sendAndConfirmMintOrThrow` directly, which is the shared chokepoint both paths use.

So no send-path code change is needed for V2 — it's covered. I did sync one stale doc that this behavior change had left behind: `src/cctp-finalizer/README.md` still said `processMintSvm` "submits via `sendAndConfirmSolanaTransaction`"; updated it to reflect the `sendAndConfirmMintOrThrow` guard (committed this pass).

Leaving this thread open for you to confirm. If you'd prefer the V2 guard inlined at line 395 for symmetry with V1 (rather than tucked inside the helper) I'm happy to refactor it that way — just say the word.

droplet-rl (commented):
🤖 Babysit self-review (pass 2). Followed up on @amateima / @dohaki's V1-vs-V2 thread.

**Finding: the V2 (production) mint path is already guarded.** `processMintSvm`'s V2 branch calls `sendAndConfirmCCTPV2ReceiveMessageTx`, which at `svmUtils.ts:279` calls `sendAndConfirmMintOrThrow` — this PR swapped the old non-throwing `sendAndConfirmSolanaTransaction` there for the throwing guard. Grepped all of `src/cctp-finalizer/`: no non-throwing `sendAndConfirmSolanaTransaction(` call sites remain, so both V1 (line 383) and V2 (via helper) throw on a dropped mint. No send-path code change is warranted; replied on the thread with the trace and left it open for @dohaki to confirm.

**One doc fix this pass:** `src/cctp-finalizer/README.md` still described `processMintSvm` as submitting "via `sendAndConfirmSolanaTransaction`" — stale after this behavior change. Updated it to reflect the `sendAndConfirmMintOrThrow` guard (used by both paths). No functional/code changes; diff remains minimal and correct. No inline issues to flag.

pxrl (approved):

PR #3540, opened 2026-07-02
chore: Filter gasless deposits by integratorId

dijanin-brat:


src/gasless/GaslessRelayerConfig.ts
```diff
@@ -50,11 +59,14 @@ export class GaslessRelayerConfig extends CommonConfig {
       API_TIMEOUT_OVERRIDE,
       INITIALIZATION_RETRY_ATTEMPTS,
       RELAYER_GASLESS_REFUND_FLOW_TEST_ENABLED,
+      ENABLE_DEPOSITS_ONLY,
```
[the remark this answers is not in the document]
dijanin-brat: [c672232](https://github.com/across-protocol/relayer/pull/3540/commits/c67223205b40b7c4d3291ae99c1c325affbf44be)

src/gasless/GaslessRelayer.ts
```diff
@@ -1030,7 +1046,45 @@ export class GaslessRelayer {
...
+  protected _filterDepositsByIntegratorId(deposits: AnyGaslessDepositMessage[]): AnyGaslessDepositMessage[] {
+    const { allowedIntegratorIds, blockedIntegratorIds } = this.config;
+    if (!isDefined(allowedIntegratorIds) && !isDefined(blockedIntegratorIds)) {
+      return deposits;
+    }
+
+    return deposits.filter((deposit) => {
+      const integratorId = deposit.integratorId?.toLowerCase();
+      if (isDefined(allowedIntegratorIds)) {
+        const allowed = isDefined(integratorId) && allowedIntegratorIds.has(integratorId);
```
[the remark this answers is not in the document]
dijanin-brat: [1253b41](https://github.com/across-protocol/relayer/pull/3540/commits/1253b418800ecab671f0196a1f99c3d7b392a28f)

src/gasless/GaslessRelayer.ts
```diff
@@ -762,8 +768,13 @@ export class GaslessRelayer {
                 ? this._extractDepositFromTransactionReceipt(depositReceipt, originChainId)
                 : await this._findDeposit(bridgeMessage);
               if (isDefined(deposit)) {
-                log("info", `Verified deposit on ${origin}`);
-                nextState = MessageState.FILL_PENDING;
+                if (this.config.depositsOnlyEnabled) {
+                  log("info", `Verified deposit on ${origin}. Deposits-only mode: skipping fill.`);
+                  nextState = MessageState.FILLED;
```
pxrl: I wonder if we want to add a separate terminal state for this - i.e. `DONE` or something, because `FILLED` tends to mislead via the logs when we're actually skipping the fill (i.e. as for cctp).

src/gasless/GaslessRelayer.ts
```diff
@@ -778,6 +789,12 @@ export class GaslessRelayer {
           }
 
           case MessageState.FILL_PENDING: {
+            if (this.config.depositsOnlyEnabled) {
+              log("info", `Deposits-only mode: skipping fill on ${destination}.`);
+              setState(MessageState.FILLED);
+              break;
+            }
```
pxrl: There could also be an argument for just modularising the state machine, so that we specify purpose-specific transitions based on what we're doing. Probably not as part of this change, but I think it'd be a worthy follow-up.

src/gasless/GaslessRelayer.ts
```diff
@@ -1030,7 +1047,45 @@ export class GaslessRelayer {
...
+      const integratorId = isDefined(deposit.integratorId) ? normalizeIntegratorId(deposit.integratorId) : undefined;
+      if (isDefined(allowedIntegratorIds)) {
+        const allowed = isDefined(integratorId) && allowedIntegratorIds.has(integratorId);
+        if (!allowed) {
+          this.logger.debug({
+            at: "GaslessRelayer#_queryGaslessApi",
+            message: "Skipping gasless deposit with integratorId outside allow-list",
+            requestId: deposit.requestId,
+            depositId: deposit.depositId,
+            integratorId: deposit.integratorId,
```
pxrl: Probably worth factoring the common fields out to a `common` object and just bundling that into each log message - it seems like it'd cut a few lines out here.

src/gasless/GaslessRelayerConfig.ts
```diff
@@ -50,11 +60,14 @@ export class GaslessRelayerConfig extends CommonConfig {
...
       SPOKE_POOL_PERIPHERY_OVERRIDES,
       GASLESS_ALLOWED_PEGGED_PAIRS,
       SWAP_API_KEY,
       NO_PERMIT2_CONTRACT_CHAINS,
       RELAYER_GASLESS_DEPOSIT_USD_PAGE_THRESHOLD,
+      RELAYER_GASLESS_ALLOWED_INTEGRATOR_IDS,
+      RELAYER_GASLESS_BLOCKED_INTEGRATOR_IDS,
```
pxrl: You might be able to set defaults to avoid the `isDefined()` dance (repeating the long env var) below.

```suggestion
      RELAYER_GASLESS_ALLOWED_INTEGRATOR_IDS = "",
      RELAYER_GASLESS_BLOCKED_INTEGRATOR_IDS = "",
```

pxrl (commented):
Only minor comments - looks good otherwise

src/utils/GaslessUtils.ts
```diff
@@ -641,8 +654,13 @@ export function validateDeposit(
   allowRefundFlowTest = false,
   allowedPeggedPairs: AllowedPeggedPairs = {},
   logger?: winston.Logger,
-  depositUsdPageThreshold = 0
+  depositUsdPageThreshold = 0,
+  depositsOnlyEnabled = false
 ): boolean {
+  if (depositsOnlyEnabled) {
+    return true;
```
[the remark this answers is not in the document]
dijanin-brat: this is okey because these deposits will be filled only if its profitable

src/gasless/GaslessRelayerConfig.ts
dijanin-brat: [a1c82d9](https://github.com/across-protocol/relayer/pull/3540/commits/a1c82d9c1147e3a349438f96c843014ede385413)

src/gasless/GaslessRelayer.ts
dijanin-brat: [a1c82d9](https://github.com/across-protocol/relayer/pull/3540/commits/a1c82d9c1147e3a349438f96c843014ede385413)

src/gasless/GaslessRelayer.ts
dijanin-brat: YES! I wanted to add new status for deposit-only flows for a long time now..

[a1c82d9](https://github.com/across-protocol/relayer/pull/3540/commits/a1c82d9c1147e3a349438f96c843014ede385413)

src/gasless/GaslessRelayer.ts
dijanin-brat: totally agree, we should have some bigger refactor to the gasless relayer in general

src/gasless/GaslessRelayer.ts
```diff
@@ -424,6 +427,10 @@ export class GaslessRelayer {
     },
     spokePool: string
   ): boolean {
+    if (this.config.depositsOnlyEnabled) {
+      return false;
```
pxrl, line 431: wdyt about dropping a log message here to specify it's skipped?

src/utils/GaslessUtils.ts
```diff
@@ -641,8 +654,13 @@ export function validateDeposit(
   allowRefundFlowTest = false,
   allowedPeggedPairs: AllowedPeggedPairs = {},
   logger?: winston.Logger,
-  depositUsdPageThreshold = 0
+  depositUsdPageThreshold = 0,
+  depositsOnlyEnabled = false
 ): boolean {
+  if (depositsOnlyEnabled) {
```
pxrl: Minor, but rather than `depositsOnlyEnabled`, how about something like `fillsEnabled`, defaulting to `true` ? Then we could set RELAYER_GASLESS_FILLS_ENABLED=false in the GA instance.

src/gasless/GaslessRelayer.ts
dijanin-brat: Well we are not skipping it, it will just be regular flow for deposit and not instant fill. But if you think that is worth of doing, i am okey with it

src/utils/GaslessUtils.ts
dijanin-brat: i am okey with that. My thinking was that the flow with the filling should be "default" one, so that is why i did it like this.

dijanin-brat: [5c0ec86](https://github.com/across-protocol/relayer/pull/3540/commits/5c0ec8621951a37eb91e630ba88a794b4ea26b84)

src/utils/GaslessUtils.ts
```diff
@@ -641,8 +654,13 @@ export function validateDeposit(
   allowRefundFlowTest = false,
   allowedPeggedPairs: AllowedPeggedPairs = {},
   logger?: winston.Logger,
-  depositUsdPageThreshold = 0
+  depositUsdPageThreshold = 0,
+  fillsEnabled = true
 ): boolean {
+  if (!fillsEnabled) {
+    return true;
```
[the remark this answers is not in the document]
dijanin-brat: this is okey

pxrl (approved):

PR #3541, opened 2026-07-03
feat(DepositAddressHandler): Periodically kick env-defined watchdog

droplet-rl:
## Summary

Adds a watchdog heartbeat (dead-man's switch) to the deposit-address handler, as agreed in Slack.

- `kickWatchdog()` is scheduled via `scheduleTask` against the handler's `abortController` signal, firing every `WATCHDOG_INTERVAL` seconds (default 15) until handover/shutdown — same pattern as the relayer's periodic address-filter refresh in `src/relayer/index.ts`.
- Each tick GETs `DEPOSIT_BOT_HEARTBEAT_URL` (Checkly heartbeat). Unset disables the heartbeat entirely.
- Best-effort by design: fetch failures are swallowed so a heartbeat problem can never disrupt the bot, and each request is capped at 5s (`fetch` has no default timeout) so a hung endpoint can't stack pending requests.
- No internal throttle — cadence is owned by the scheduler (per Slack discussion). With a Checkly period of 30s + grace of 30s, a dead bot trips the alert within ~60s; cadence ≪ period, so a dropped ping is fine.
- Config/env documented in `.env.example` and `src/deposit-address/README.md`.

## Testing

- `tsc --noEmit` clean, eslint clean.
- `test/DepositAddressHandler.ts`: 30 passing. (1 pre-existing failure in `DepositAddressHandler.publish.ts` reproduces identically on unmodified `master` — unrelated.)
- Smoke-tested against a local HTTP server: ping lands on the endpoint; unset URL is a no-op; unreachable endpoint is swallowed without throwing.

### DIFF
--- a/src/deposit-address/DepositAddressHandler.ts
+++ b/src/deposit-address/DepositAddressHandler.ts
@@ -309,6 +309,20 @@ export class DepositAddressHandler {
       this.config.indexerPollingInterval,
       this.abortController.signal
     );
+    scheduleTask(() => this.runPeriodicTask(), this.config.periodicTaskInterval, this.abortController.signal);
+  }
+
+  /**
+   * Placeholder for a recurring background task, fired every `periodicTaskInterval` seconds
+   * alongside the indexer poll (same pattern as the relayer's periodic address-filter refresh).
+   * Runs until the handler's abortController fires on handover/shutdown.
+   */
+  private async runPeriodicTask(): Promise<void> {
+    this.logger.debug({
+      at: "DepositAddressHandler#runPeriodicTask",
+      message: "Periodic task fired (placeholder).",
+      intervalSeconds: this.config.periodicTaskInterval,
+    });
   }
 
   /*
--- a/src/deposit-address/DepositAddressHandler.ts
+++ b/src/deposit-address/DepositAddressHandler.ts
@@ -66,6 +66,13 @@ const INTEGRATOR_ID_REGEX = /^0x[0-9a-fA-F]{4}$/;
 const SUPPORTED_INDEXER_MESSAGE_VERSIONS = new Set([1, 3]);
 
 /**
+ * Cap on each watchdog heartbeat request. `fetch` has no default timeout, and the scheduler
+ * fires every ping fire-and-forget — without a cap, a hung heartbeat endpoint would stack
+ * pending requests indefinitely.
+ */
+const HEARTBEAT_TIMEOUT_MS = 5_000;
+
+/**
  * Independent relayer bot which processes EIP-3009 signatures into deposits and corresponding fills.
  */
 export class DepositAddressHandler {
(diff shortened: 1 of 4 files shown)


src/deposit-address/DepositAddressHandlerConfig.ts
```diff
@@ -28,6 +30,7 @@ export class DepositAddressHandlerConfig extends CommonConfig {
 
     const {
       INDEXER_API_POLLING_INTERVAL,
+      DEPOSIT_ADDRESS_WATCHDOG_INTERVAL,
```
pxrl:
```suggestion
      DEPOSIT_ADDRESS_WATCHDOG_INTERVAL = "60",
```

droplet-rl: Applied in 0e799d39 — default now set in the destructuring, and the redundant `?? 60` fallback below is removed.

src/deposit-address/DepositAddressHandler.ts
```diff
@@ -309,6 +316,26 @@ export class DepositAddressHandler {
...
+   */
+  private async kickWatchdog(): Promise<void> {
+    const url = this.config.heartbeatUrl;
+    if (!url) {
+      return;
+    }
+    try {
+      await fetch(url, { method: "GET", signal: AbortSignal.timeout(HEARTBEAT_TIMEOUT_MS) });
+    } catch {
+      /* best-effort: never let a heartbeat failure disrupt the bot */
```
pxrl: We probably want to log about this, perhaps if failures exceed some minimum number in a single run (i.e. 10 failures).

src/deposit-address/DepositAddressHandlerConfig.ts
```diff
@@ -28,6 +33,8 @@ export class DepositAddressHandlerConfig extends CommonConfig {
 
     const {
       INDEXER_API_POLLING_INTERVAL,
+      WATCHDOG_INTERVAL = "15",
```
pxrl, line 36: 15 -> 60

src/deposit-address/DepositAddressHandler.ts
droplet-rl: Done in 0351a41c. The handler now tracks consecutive heartbeat failures (reset on success) and logs a warning on every 10th consecutive failure — once per ~2.5min at the default 15s cadence during a sustained outage, rather than per-tick spam or a single line then silence. The log includes the failure count and last error.

Two notes on interpretation:
- I went with *consecutive* failures rather than cumulative-per-run, so sporadic one-off failures over a long-lived run don't accumulate into a false alarm.
- Non-2xx responses now count as failures too: `fetch` doesn't throw on HTTP error statuses, so a misconfigured heartbeat URL (persistent 404) would previously have been completely silent while the alert fired.

pxrl (approved):

pxrl (approved):

PR #3560, opened 2026-07-08
fix(rebalancer): tie Binance swap-deposit tag lifecycle to order state

droplet-rl:
## Problem

`_depositToBinance` tags each swap deposit `SWAP` in Redis so the Binance finalizer excludes it from its `amountToFinalize` ledger — but the tag expired on a **30-minute wall clock**, while the finalizer's deposit-history lookback is far longer. Once the tag lapsed, `getBinanceDepositType` returned `UNKNOWN` and the deposit was re-counted as a bridge deposit awaiting withdrawal — **even when its swap had completed and the funds were long gone**. This phantom debt was then paid out of the *next* order's just-credited deposit, starving that order until `REBALANCER_PENDING_ORDER_TTL` pruned it.

Observed 7 times over 2026-07-01 → 07-08 on USDT→USDC (HyperEVM) routes. Traced example (cloid `0xe42d273cb1ce29c53aad23e9241606cd`, 2026-07-08): its 82,385.72 USDT deposit ([tx](https://bscscan.com/tx/0x2a845bbb98e1347aa594e8fdde705d74d42d4ae22ed9884fdfe1d014603a60ff)) was credited at ~04:58 and withdrawn back to BSC by the finalizer at 04:59:27 against a 158k phantom backlog created by the *previous, successfully completed* order (`0x32e9b903`, swap filled 04:16, finalized 04:21, tag expired ~04:31). The starved order was pruned at 05:56:12 — the PagerDuty alert that kicked this off.

## Change

Tag lifecycle now follows **order state** instead of wall clock (the withdrawal-side SWAP tag already worked this way — set with the Redis default 1-week TTL):

- **Deposit** — `setBinanceDepositType(..., SWAP)` without explicit TTL → Redis default (1 week), which outlives the finalizer lookback. A `cloid → {chainId, transactionHash}` mapping is stored alongside the order state so later transitions can find the tag.
- **Order abandoned in `PENDING_DEPOSIT`** (TTL prune) — new `BaseAdapter._onExpiredOrderPruned` hook (no-op by default, wrapped so it can't break the shared read path); the Binance adapter deletes the deposit's tag, handing the unconsumed funds to the finalizer for reclaim on its next run.
- **Order progressed past `PENDING_DEPOSIT`** — the deposit was consumed by the spot order; the tag stays, so the finalizer's ledger never re-counts consumed deposits.

Behavior change: stuck-funds recovery starts when the order dies (`REBALANCER_PENDING_ORDER_TTL`, default 1h) rather than 30 minutes after deposit. This also removes the race where the finalizer could reclaim a *live* order's deposit that was legitimately waiting out Binance's deposit-unlock confirmations (RW00441).

Not touched: `scripts/swapOnBinance.ts` (interactive operator flow manages its own tag TTL); orders created before this change have no txn mapping — the prune hook logs and skips, falling back to today's behavior.

Replaces #3559, which protected the balance from the phantom debt but didn't stop the phantom debt from forming.

## Testing

- `yarn tsc --noEmit` clean
- `hardhat test test/BinanceAdapter.helpers.ts test/BinanceAdapter.withdrawals.ts test/BinanceUtils.ts test/BinanceFinalizer.ts` — 48 passing
- eslint + prettier clean on touched files

### DIFF
--- a/src/rebalancer/adapters/baseAdapter.ts
+++ b/src/rebalancer/adapters/baseAdapter.ts
@@ -293,8 +293,26 @@ export abstract class BaseAdapter implements RebalancerAdapter {
           message: `⏰ Pruning expired pending order ${cloid} from status set ${statusSetKey} without finalization. The order's REBALANCER_PENDING_ORDER_TTL elapsed before it could progress out of this status.`,
           account: account.toNative(),
         });
+        // The hook must not break this read path: it is called from every bot sharing the Redis state, so a
+        // venue/Redis hiccup during cleanup should be logged rather than propagated.
+        try {
+          await this._onExpiredOrderPruned(status, cloid, account);
+        } catch (error) {
+          this.logger.warn({
+            at: "BaseAdapter._redisCleanupPendingOrders",
+            message: `Expired-order cleanup hook failed for cloid ${cloid}`,
+            error,
+          });
+        }
       }
     });
+  }
+
+  // Hook invoked after a pending order is pruned because its details key TTL elapsed (a genuine expiry, not a
+  // concurrent finalize). Adapters override this to release venue-side state tied to the abandoned order — e.g.
+  // the Binance adapter untags the order's exchange deposit so the Binance finalizer can reclaim the funds.
+  protected async _onExpiredOrderPruned(_status: STATUS, _cloid: string, _account: EvmAddress): Promise<void> {
+    return;
   }
 
   protected async _redisGetPendingBridgesPreDeposit(account: EvmAddress): Promise<string[]> {
--- a/src/rebalancer/adapters/baseAdapter.ts
+++ b/src/rebalancer/adapters/baseAdapter.ts
@@ -293,17 +293,7 @@ export abstract class BaseAdapter implements RebalancerAdapter {
           message: `⏰ Pruning expired pending order ${cloid} from status set ${statusSetKey} without finalization. The order's REBALANCER_PENDING_ORDER_TTL elapsed before it could progress out of this status.`,
           account: account.toNative(),
         });
-        // The hook must not break this read path: it is called from every bot sharing the Redis state, so a
-        // venue/Redis hiccup during cleanup should be logged rather than propagated.
-        try {
-          await this._onExpiredOrderPruned(status, cloid, account);
-        } catch (error) {
-          this.logger.warn({
-            at: "BaseAdapter._redisCleanupPendingOrders",
-            message: `Expired-order cleanup hook failed for cloid ${cloid}`,
-            error,
-          });
-        }
+        await this._onExpiredOrderPruned(status, cloid, account);
       }
     });
   }
--- a/src/rebalancer/adapters/baseAdapter.ts
+++ b/src/rebalancer/adapters/baseAdapter.ts
@@ -298,12 +298,7 @@ export abstract class BaseAdapter implements RebalancerAdapter {
     });
   }
 
-  // Hook invoked after a pending order is pruned because its details key TTL elapsed (a genuine expiry, not a
-  // concurrent finalize). Adapters override this to release venue-side state tied to the abandoned order — e.g.
-  // the Binance adapter untags the order's exchange deposit so the Binance finalizer can reclaim the funds.
-  protected async _onExpiredOrderPruned(_status: STATUS, _cloid: string, _account: EvmAddress): Promise<void> {
-    return;
-  }
+  protected async _onExpiredOrderPruned(_status: STATUS, _cloid: string, _account: EvmAddress): Promise<void> {}
 
   protected async _redisGetPendingBridgesPreDeposit(account: EvmAddress): Promise<string[]> {
     await this._redisCleanupPendingOrders(STATUS.PENDING_BRIDGE_PRE_DEPOSIT, account);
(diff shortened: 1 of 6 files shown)


src/rebalancer/adapters/baseAdapter.ts
```diff
@@ -293,10 +293,28 @@ export abstract class BaseAdapter implements RebalancerAdapter {
           message: `⏰ Pruning expired pending order ${cloid} from status set ${statusSetKey} without finalization. The order's REBALANCER_PENDING_ORDER_TTL elapsed before it could progress out of this status.`,
           account: account.toNative(),
         });
+        // The hook must not break this read path: it is called from every bot sharing the Redis state, so a
+        // venue/Redis hiccup during cleanup should be logged rather than propagated.
```
nicholaspai: Remove the try-catch logic and comment

droplet-rl: Done in bdf7ede6 — removed the try-catch and the comment; hook errors now propagate.

src/rebalancer/adapters/binance.ts
```diff
@@ -540,9 +541,41 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
+        at: "BinanceStablecoinSwapAdapter._onExpiredOrderPruned",
+        message: `No deposit transaction mapping found for pruned order ${cloid}; cannot release its deposit tag`,
+        cloid,
+      });
+      return;
+    }
+    const { chainId, transactionHash } = JSON.parse(depositTxn) as { chainId: number; transactionHash: string };
+    await deleteBinanceDepositType(chainId, transactionHash);
+    await this.redisCache.del(depositTxnKey);
+    this.logger.info({
```
nicholaspai:
```suggestion
    this.logger.debug({
```

droplet-rl: Applied in 3236701e — downgraded to `logger.debug`.

src/rebalancer/adapters/binance.ts
```diff
@@ -1137,9 +1175,17 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
...
-    // Set the TTL to 30 minutes so that the Binance sweeper finalizer only attempts to pull back these deposited
-    // funds after 30 minutes. If the swap hasn't occurred in 30 mins then something has gone wrong.
-    await setBinanceDepositType(sourceChain, txnHash, BinanceTransactionType.SWAP, 30 * 60);
+    // The SWAP tag keeps this deposit out of the Binance finalizer's ledger. It must outlive the finalizer's
+    // deposit-history lookback (the Redis default TTL is 1 week), otherwise a completed swap's deposit would be
+    // re-counted as finalizable once the tag lapses, creating phantom `amountToFinalize` debt that the finalizer
+    // pays with the next order's credited funds. If this order is instead abandoned (TTL prune), the tag is
+    // deleted in _onExpiredOrderPruned so the finalizer can reclaim the funds — recovery keys off order state,
+    // not wall clock. The cloid -> deposit txn mapping below is what lets the prune path find this tag.
+    await setBinanceDepositType(sourceChain, txnHash, BinanceTransactionType.SWAP);
```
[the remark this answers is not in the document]
nicholaspai: we should use no TTL, similar to how we deal with binance withdrawals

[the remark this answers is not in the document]
droplet-rl: Good catch — confirmed: the SDK Redis default TTL is 1 week while `maxFinalizerLookback` defaults to `FINALIZER_TOKENBRIDGE_LOOKBACK` (14 days), so a consumed swap deposit aged 7–14 days would have re-entered the finalizer ledger. Fixed in 4843d2a0: the SWAP tag (and the cloid → deposit-txn mapping) now gets an explicit TTL of `2 * FINALIZER_TOKENBRIDGE_LOOKBACK` (28 days), leaving margin for `FINALIZER_MAX_TOKENBRIDGE_LOOKBACK` overrides.

nicholaspai (dismissed):

src/rebalancer/adapters/binance.ts
```diff
@@ -1137,9 +1168,14 @@ export class BinanceStablecoinSwapAdapter extends BaseAdapter {
       );
       txnHash = await this._submitTransaction(txn);
     }
-    // Set the TTL to 30 minutes so that the Binance sweeper finalizer only attempts to pull back these deposited
-    // funds after 30 minutes. If the swap hasn't occurred in 30 mins then something has gone wrong.
-    await setBinanceDepositType(sourceChain, txnHash, BinanceTransactionType.SWAP, 30 * 60);
+    // TTL must outlive the finalizer lookback so completed swaps are not re-counted as finalizable.
+    // The cloid -> deposit txn mapping lets the prune path find abandoned deposit tags.
+    await setBinanceDepositType(sourceChain, txnHash, BinanceTransactionType.SWAP, 2 * FINALIZER_TOKENBRIDGE_LOOKBACK);
```
droplet-rl, line 1173: Want to double-check this one before applying: dropping the explicit TTL falls back to the SDK's `DEFAULT_CACHING_TTL` (1 week), while the Binance finalizer's deposit window comes from `maxFinalizerLookback` (default `FINALIZER_TOKENBRIDGE_LOOKBACK` = 14 days, via `constructSpokePoolClientsWithLookback` in `src/finalizer/index.ts`). That reopens the gap Codex flagged: a consumed swap deposit aged 7–14 days loses its tag while still inside the finalizer's deposit history and gets re-counted as finalizable. It would also leave the call inconsistent with the mapping `set` right below (still 28d), the comment above it, and the README.

If prod always runs the finalizer with a sub-1-week lookback and you'd rather keep the default here, happy to apply this plus trim the comment/mapping TTL/README to match — just confirm and I'll push it.

PR #3568, opened 2026-07-13
fix(gasless): stop dropping long-tail swapToken deposits + surface swallowed task rejections (ACB-552)

mrice32:
## What

Fixes [ACB-552](https://linear.app/uma/issue/ACB-552): gasless swapAndBridge deposits whose user-signed origin `swapToken` isn't in the static `TOKEN_SYMBOLS_MAP` were **silently dropped** — state advanced `INITIAL → DEPOSIT_SUBMIT → DEPOSIT_CONFIRM` and then nothing, no broadcast, no log.

Latent today (production gasless uses USDC/USDT, both in the static map); becomes a hard blocker as soon as WS2 admits long-tail tokens.

## Root cause

`getTokenInfo(swapToken, originChainId)` throws for any token absent from `TOKEN_SYMBOLS_MAP`. In `initiateDeposit` it was called **only to build a cosmetic Slack `mrkdwn` line** — the deposit tx is already built above it. The throw rejected `initiateDeposit`, the rejection landed on the un-awaited `depositReceiptPromise`, propagated through `forEachAsync` → `evaluateApiSignatures`, and was swallowed by `fireAndForget`'s `.catch(() => undefined)` → zero logs.

## Fix

**1. `resolveTokenInfoForLog` (new, `src/utils/GaslessUtils.ts`)** — never-throwing resolution used at the log site:
- static `getTokenInfo` → Redis-cached on-chain ERC-20 probe (SDK `fetchTokenInfo`, 30-day TTL since token metadata is immutable) → neutral `{ symbol: "UNKNOWN", decimals: 18 }` placeholder.
- The result feeds **only** the Slack log line, never the on-chain deposit, so a placeholder/imprecise value cannot affect correctness — it just keeps logs readable.
- The on-chain probe is an injectable dependency, so the fallback logic is unit-tested without a network.

**2. `fireAndForget` logging (`src/utils/Tasks.ts`)** — swallowed rejections are now logged at error level when a logger is provided (still never rethrown — an unhandled rejection could crash the process). `scheduleTask` threads a logger through, and the gasless poller passes `this.logger`. Independent of the fix above, this alone would have surfaced the original drop in minutes.

## Tests

- `test/Tasks.ts`: 3 new `fireAndForget` cases (logs on reject; silent on resolve; never throws / swallows without a logger).
- `test/GaslessUtils.ts`: 4 new `resolveTokenInfoForLog` cases (static-map hit with no probe; on-chain probe + cache write on miss; cache-hit with no probe; probe-failure → placeholder, never throws).

`yarn typecheck` → 0 errors. `hardhat test test/Tasks.ts test/GaslessUtils.ts` → **27 passing**. Prettier clean.

## Notes

- No API/behavior change for tokens already in the static map (static path is tried first and returned unchanged).
- The "optional same-day mitigation" from the ticket (try/catch placeholder) is subsumed by the fuller probe-on-miss approach here.

https://claude.ai/code/session_01T2D1PN1krMHEz771tJ4PPG

### DIFF
--- a/src/utils/GaslessUtils.ts
+++ b/src/utils/GaslessUtils.ts
@@ -20,6 +20,8 @@ import {
   ConvertDecimals,
   convertRelayDataParamsToBytes32,
   getTokenInfo,
+  fetchTokenInfo,
+  getProvider,
   toBN,
   toBytes32,
   toAddressType,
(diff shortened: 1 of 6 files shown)


src/utils/GaslessUtils.ts
```diff
@@ -35,6 +37,103 @@ import { isStablecoin } from "./TokenUtils";
...
+  const probeOnChain = opts.probeOnChain ?? defaultOnChainTokenInfoProbe;
+  try {
+    const { symbol, decimals } = await probeOnChain(address, chainId);
+    const info = { symbol, decimals };
+    try {
+      await redisCache?.set(cacheKey, JSON.stringify(info), GASLESS_TOKEN_INFO_CACHE_TTL_SECONDS);
+    } catch {
+      // Best-effort cache write — a failure here only costs a future re-probe.
+    }
+    return info;
```
dijanin-brat: Can we remove this inner try/catch and just add if (isDefined(info)) return info in outer catch. I personally just don't like nesting try/catch and would like to avoid it if possible, but its not that problematic here so we can leave as is if you think this is better solution.

dijanin-brat: [218cd86](https://github.com/across-protocol/relayer/pull/3568/commits/218cd86d940332edb90628df1c38715e9aa9477c)

test/Tasks.ts
```diff
@@ -103,4 +103,38 @@ describe("Tasks", function () {
...
+  describe("fireAndForget", function () {
+    // A macrotask tick to let the wrapped promise's `.catch` run before assertions.
+    const tick = () => new Promise((resolve) => setTimeout(resolve, 0));
+
+    it("logs a rejected task at error level when a logger is provided (ACB-552)", async function () {
+      const errors: Record<string, unknown>[] = [];
+      const logger = { error: (info: Record<string, unknown>) => errors.push(info) };
+      const boom = new Error("boom");
+
+      fireAndForget(() => Promise.reject(boom), logger, "TestTask")();
```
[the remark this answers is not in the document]
dijanin-brat: [5dc801e](https://github.com/across-protocol/relayer/pull/3568/commits/5dc801e668e197a4ff3efce56935342bce71ba32)

src/utils/GaslessUtils.ts
```diff
@@ -35,6 +37,105 @@ import { isStablecoin } from "./TokenUtils";
...
+    } catch {
+      // Corrupt cache entry — ignore and re-probe.
+    }
+  }
+
+  const probeOnChain = opts.probeOnChain ?? defaultOnChainTokenInfoProbe;
+  let info: { symbol: string; decimals: number } | undefined = undefined;
+  try {
+    const { symbol, decimals } = await probeOnChain(address, chainId);
+    info = { symbol, decimals };
```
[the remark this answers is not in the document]
dijanin-brat: [7692d26](https://github.com/across-protocol/relayer/pull/3568/commits/7692d2635fe9ac3089c89154b380e6d2636a9a11)

src/utils/GaslessUtils.ts
```diff
@@ -35,6 +37,105 @@ import { isStablecoin } from "./TokenUtils";
...
+  try {
+    cached = (await redisCache?.get<string>(cacheKey)) ?? null;
+  } catch {
+    // Best-effort cache read — fall through to the probe on any error.
+  }
+  if (isDefined(cached)) {
+    try {
+      const parsed = JSON.parse(cached) as { symbol: string; decimals: number };
+      if (typeof parsed.symbol === "string" && typeof parsed.decimals === "number") {
+        return parsed;
```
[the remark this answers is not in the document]
dijanin-brat: [ce2bffc](https://github.com/across-protocol/relayer/pull/3568/commits/ce2bffcbf89cf9b39655cbe852ec8c8d22c2f331)

src/utils/GaslessUtils.ts
```diff
@@ -35,6 +37,125 @@ import { isStablecoin } from "./TokenUtils";
...
+  }
+  const { symbol, decimals } = info as { symbol?: unknown; decimals?: unknown };
+  return (
+    typeof symbol === "string" &&
+    typeof decimals === "number" &&
+    Number.isInteger(decimals) &&
+    Number.isFinite(decimals) &&
+    decimals >= 0 &&
+    decimals <= 255
+  );
```
pxrl: @droplet-rl Does superstruct make sense here?

droplet-rl: Yes — good call. superstruct is already a direct dependency and the established idiom for this kind of untrusted-shape validation (`TypeGuards.ts`, `SuperstructUtils.ts`), so I replaced the hand-rolled checks with a struct in bc5f3ece:

```ts
const TokenInfoForLog = type({
  symbol: string(),
  decimals: max(min(integer(), 0), 255),
});
```

`integer()` already rejects NaN/Infinity/fractional values, so the explicit `Number.isInteger`/`isFinite` checks collapse into the refinements. I used `type()` rather than `object()` to keep the guard's existing tolerance of extra keys in cached entries. `isValidTokenInfoForLog` stays as a thin named wrapper around `is()` so the call site keeps its intent-revealing name.

Tests (incl. the out-of-range-decimals re-probe case), eslint, and tsc all pass.

pxrl (approved):

PR #3569, opened 2026-07-14
feat(deposit-address): relay erc20Transfer provenance on v3 execute

amateima:
Attach the new optional erc20Transfer object ({ chainId, blockNumber, transactionHash, logIndex }) to POST /deposit-addresses/execute requests. When present, the quote-api wraps executeTx in a Multicall3 bundle that also emits a version-2 provenance blob via AcrossEventEmitter, giving the indexer an on-chain sweep <-> funding-transfer link in the execute receipt.

The bot always has these fields on the indexer message, so it always sends them (chainId is coerced to an integer as the endpoint requires).

### DIFF
--- a/src/deposit-address/DepositAddressHandler.ts
+++ b/src/deposit-address/DepositAddressHandler.ts
@@ -1192,10 +1192,10 @@ export class DepositAddressHandler {
       // Provenance reference to the inbound funding transfer. The API folds this into a Multicall3
       // bundle that emits a version-2 provenance blob, giving the indexer an on-chain sweep ↔
       // funding-transfer link in the execute receipt. The endpoint requires chainId as a
-      // non-negative integer; coerce only when the indexer message hands it to us as a string, so a
-      // future switch to a numeric chainId in the polled response keeps working unchanged.
+      // non-negative integer; Number() is a no-op on an already-numeric value, so this keeps working
+      // if the indexer later types chainId as a number.
       erc20Transfer: {
-        chainId: typeof erc20Transfer.chainId === "number" ? erc20Transfer.chainId : Number(erc20Transfer.chainId),
+        chainId: Number(erc20Transfer.chainId),
         blockNumber: erc20Transfer.blockNumber,
         transactionHash: erc20Transfer.transactionHash,
         logIndex: erc20Transfer.logIndex,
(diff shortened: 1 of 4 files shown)


src/deposit-address/DepositAddressHandler.ts
```diff
@@ -1181,6 +1181,16 @@ export class DepositAddressHandler {
       amount: erc20Transfer.amount,
       executionFeeRecipient: this.signerAddress.toNative(),
       integratorId,
+      // Provenance reference to the inbound funding transfer. The API folds this into a Multicall3
+      // bundle that emits a version-2 provenance blob, giving the indexer an on-chain sweep ↔
+      // funding-transfer link in the execute receipt. chainId is stringly-typed on the indexer
+      // message; the endpoint requires a non-negative integer.
+      erc20Transfer: {
```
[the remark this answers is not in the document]
amateima: Good catch — agreed, the exact `deep.equal` in the `_getExecuteTx request mapping` test would have failed deterministically. Fixed in 50f5923e: the expected request body now includes the `erc20Transfer` provenance object (with `chainId` coerced to a number), plus a new case asserting a numeric `chainId` passes through without re-coercion. Suite passes locally.

src/deposit-address/DepositAddressHandler.ts
```diff
@@ -1189,6 +1189,17 @@ export class DepositAddressHandler {
       amount: erc20Transfer.amount,
       executionFeeRecipient: this.signerAddress.toNative(),
       integratorId,
+      // Provenance reference to the inbound funding transfer. The API folds this into a Multicall3
+      // bundle that emits a version-2 provenance blob, giving the indexer an on-chain sweep ↔
+      // funding-transfer link in the execute receipt. The endpoint requires chainId as a
+      // non-negative integer; coerce only when the indexer message hands it to us as a string, so a
+      // future switch to a numeric chainId in the polled response keeps working unchanged.
+      erc20Transfer: {
+        chainId: typeof erc20Transfer.chainId === "number" ? erc20Transfer.chainId : Number(erc20Transfer.chainId),
```
dijanin-brat: can we just always do Number(erc20Transfer.chainId) here?

amateima: yes. I actually wrote it this way in the first place. I don't know why I thought Number(number) would error 🤦

amateima: https://github.com/across-protocol/relayer/pull/3569/changes/92ba93f36ff4c25a10d0881aa5802f019835b2c5

dijanin-brat (approved):

PR #3588, opened 2026-07-17
fix(rebalancer): skip status updates for orders on routes an instance does not support

droplet-rl:
## Problem

The `swapRebalancer` and `sameAssetRebalancer` bots typically share a base signer and the Redis order store, so each bot's `updateRebalanceStatuses()` pass lists pending orders created by the other mode. Progressing another mode's order throws `Route is not supported: <token> <sourceChain> -> <token> <destinationChain>` on one of its bridge legs (observed: the `sameAssetRebalancer` crashing on `USDC 999 -> USDC 1` while progressing a swap-mode Hyperliquid order's final CCTP leg out of HyperEVM), aborting the entire run.

## Fix

Per review feedback, each client now only updates statuses of orders for routes it supports:

- `BaseAdapter._filterOrdersOnSupportedRoutes` filters a pending-order set down to orders whose stored route is in this instance's `availableRoutes`, and emits one warn per adapter pass listing the skipped orders.
- All four adapters (`cctp`, `oft`, `hyperliquid`, `binance`) apply the filter to every pending-order set read in `updateRebalanceStatuses()`. Orders on unsupported routes are left pending for a properly-configured instance to progress, so one unroutable order can no longer take down a run after future config drift either.
- `supportsRoute` now accepts a route-shaped object without requiring the `adapter` tag (the tag is irrelevant there — `availableRoutes` is already filtered per adapter at initialize()), so order details can be checked directly.

This replaces the previous approach on this branch (initializing the CCTP/OFT adapters with the union of all modes' route catalogs), which has been reverted.

## Tests

`test/RebalancerAdapter.updateStatusRouteFiltering.ts`:
- CCTP adapter skips pending orders on unconfigured routes and warns, without touching the attestation lookup.
- CCTP adapter still progresses orders on supported routes.
- Regression for the original crash: an instance with no Hyperliquid routes observing a swap-mode Hyperliquid order pending its final withdrawal leaves it pending and never calls the CCTP/OFT `initializeRebalance` final bridge leg.

### DIFF
--- a/src/rebalancer/README.md
+++ b/src/rebalancer/README.md
@@ -249,6 +249,8 @@ High-level flow:
 
 The runtime entrypoint is `runSameAssetRebalancer`, exposed as the `sameAssetRebalancer` bot. It is independent of cumulative mode: operators enable destination token/chain pairs under `sameAssetBalances`, while the support catalog controls which of those pairs can become routes.
 
+Cross-mode order progression: the `swapRebalancer` and `sameAssetRebalancer` bots typically share a base signer and the Redis order store, so each bot's `updateRebalanceStatuses()` pass can pick up and progress pending orders created by the other mode. To make the intermediate/final CCTP and OFT bridge legs of those orders executable in either bot, `constructInitializedRebalancerClient` initializes the CCTP/OFT bridge adapters with `buildAllExecutableRebalanceRoutes` — the union of the cumulative (swap) and SameAsset route catalogs — while each mode's client still initiates new rebalances only from its own mode-specific route list.
+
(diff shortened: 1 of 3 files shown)


nicholaspai (changes requested):
I think a better fix that can replace this is to implement this:

"updateRebalanceStatuses could check supportsRoute before a final bridge leg and warn + skip (leaving the order pending for a properly-configured instance) instead of throwing, so one unroutable order can't take down an entire run after future config drift."

Essentially, each client should only update statuses of orders for routes it supports

PR #3591, opened 2026-07-18
fix(AcrossApiClient): retain last known limits when the limits update fails

droplet-rl:
## Motivation

The relayer logged `😱 Skipping deposit with greater unfilled amount than API suggested limit` with `limit: 0` for WETH deposits (e.g. depositId 21632, origin 4326 → destination 4663), skipping deposits that were actually fillable.

Root cause: when the `/liquid-reserves` query fails (timeout, HTTP error, or invalid response), `callLimits()` returned `bnZero` for every queried token, and `update()` unconditionally overwrote the stored limits with those zeroes. A single transient API failure therefore clobbered known-good limits and blocked all fills (for non-hub-chain origins) until the next successful refresh.

## Changes

- `callLimits()` now returns `undefined` on failure instead of an array of zeroes. The invalid-response path previously fell through to a `TypeError` (`result[...]` on a null result) that was swallowed by the catch block; it now returns cleanly after logging.
- `update()` only overwrites stored limits when the query succeeds. On failure it logs a warning and retains the most recently fetched limits, so the relayer keeps operating on the last known values until the next refresh (retry cadence is unchanged: next attempt after `API_UPDATE_RETENTION_TIME`).
- Behavior when *no* update has ever succeeded is unchanged: limits default to 0 and deposits from non-hub-chain origins are skipped (conservative, same as before).
- `callLimits`/`updatedAt` relaxed from `private` to `protected` to allow the test mock to stub the network call and the retention window.
- Adds `test/AcrossApiClient.ts` covering successful update, fallback-on-failure, recovery after failure, first-update failure, and the hub-chain-origin exemption.

## Docs

No `README.md`/`AGENTS.md` updates needed: `AcrossApiClient` limits behavior is not currently documented in `src/clients/README.md` or `src/relayer/README.md`.

## Testing

- `npx hardhat test test/AcrossApiClient.ts` — 4 passing
- `npx hardhat test test/Relayer.UnfilledDeposits.ts` — 8 passing
- `tsc --noEmit` and `eslint` clean on changed files


src/clients/AcrossAPIClient.ts
```diff
@@ -84,13 +84,21 @@ export class AcrossApiClient {
...
-    });
+    if (isDefined(liquidReserves)) {
+      tokens.forEach((token, i) => (this.limits[token.toEvmAddress()] = liquidReserves[i]));
+      this.logger.debug({
+        at: "AcrossAPIClient",
+        message: "🏁 Fetched HubPool liquid reserves",
+        limits: this.limits,
+      });
+    } else {
+      // The query failed, so fall back to the most recently fetched limits rather than zeroing them out.
```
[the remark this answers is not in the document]
droplet-rl: Good catch — added an "Across API Client" section to `src/clients/README.md` (the module doc for shared runtime clients) covering the limits flow and the new fallback: on a failed `/liquid-reserves` query the relayer retains the last successfully fetched limits, and tokens with no successful fetch default to a limit of 0.

src/clients/AcrossAPIClient.ts
```diff
@@ -84,13 +84,21 @@ export class AcrossApiClient {
...
+      this.logger.warn({
+        at: "AcrossAPIClient",
+        message: "Failed to fetch HubPool liquid reserves. Retaining last known limits.",
+        limits: this.limits,
+      });
+    }
     this.updatedLimits = true;
     this.updatedAt = now;
```
pxrl: In this case we will set `updatedLimits` true even if it fails. How does this work on the first call to update limits?

droplet-rl: Intentional — `updatedLimits` gates whether the relayer *enforces* the limit check at all (`Relayer.ts`: `if (acrossApiClient.updatedLimits && inputAmount.gt(limit))`). Leaving it `false` on failure would bypass the check entirely and let fills proceed unconstrained (fail-open).

On the first call, a failed query leaves `this.limits` empty, so `getLimit()` warns and returns 0 — non-hub-origin fills for those tokens are skipped until a query succeeds. That is the same fail-closed outcome as before this PR, where a failure stored explicit zeros; the change only affects failures *after* a successful update, which now retain the last good values instead of zeroing them. Covered by the "Defaults to a limit of 0 when no update has succeeded" test.

Added a comment in 6689fda5 to make this explicit.

src/clients/AcrossAPIClient.ts
pxrl: @droplet-rl Also bump timeout to 10 seconds.

src/clients/AcrossAPIClient.ts
```diff
@@ -84,13 +84,23 @@ export class AcrossApiClient {
...
+        message: "🏁 Fetched HubPool liquid reserves",
+        limits: this.limits,
+      });
+    } else {
+      // The query failed, so fall back to the most recently fetched limits rather than zeroing them out.
+      this.logger.warn({
+        at: "AcrossAPIClient",
+        message: "Failed to fetch HubPool liquid reserves. Retaining last known limits.",
+        limits: this.limits,
+      });
```
pxrl: @droplet-rl These two debug messages are mostly equivalent. Factor out the common fields and then supplement with the message where they are different. Try to reduce the number of lines required.

src/clients/AcrossAPIClient.ts
droplet-rl: Done in b0cc48af — constructor default `timeout` is now `10_000` (was 3000). No caller overrides it (`RelayerClientHelper.ts:162` and the `test/Relayer.*` constructions all rely on the default), so this applies everywhere.

Fits within the existing cadence: the relayer schedules `update()` every 30s (`src/relayer/index.ts:73`) and `API_UPDATE_RETENTION_TIME` is 60s, so a worst-case 10s hang cannot overlap the next query. It does add up to ~10s to `Relayer.init()` on a hung API, but that call is already `Promise.all`-ed alongside the config and token-client updates, so it only bites if the API itself is the slowest leg.

src/clients/AcrossAPIClient.ts
```diff
@@ -84,13 +84,23 @@ export class AcrossApiClient {
...
+    }
+    // Flag limits as updated even on failure so that the relayer keeps enforcing the limit check. If the first
+    // query fails, limits is empty and getLimit() defaults to 0, so fills are skipped until a query succeeds.
```
pxrl: I think we should be trying to enforce that limits succeeds before we return from the first update.

droplet-rl: Agreed — done in ba4d021a. `update()` now throws when the query fails and nothing has ever been fetched:

```ts
} else if (this.updatedLimits) {
  // A previous update succeeded, so fall back to those limits rather than zeroing them out.
  this.logger.warn({ ... });
} else {
  throw new Error("Failed to fetch HubPool liquid reserves");
}
```

`updatedAt` is deliberately left unset on the throw, so the next call retries immediately instead of short-circuiting on the 60s retention window. `updatedLimits` now also carries a stronger guarantee: true implies at least one successful fetch.

This surfaces at startup — `Relayer.init()` awaits `update()` inside `Promise.all` (`Relayer.ts:108`) before `scheduleTask` is wired up, so a first-fetch failure fails relayer startup loudly rather than filling nothing. Post-startup failures go through the retain path and never throw, and `scheduleTask` uses `fireAndForget` anyway.

Test swapped from "Defaults to a limit of 0 when no update has succeeded" to "Requires the initial limits update to succeed" (asserts the rejection, `updatedLimits === false`, and that the immediately following update still succeeds). Also fixed the hub-chain test, which had been relying on a failed first update returning normally. Confirmed the `test/Relayer.*` suites never call `update()`/`init()`, so the new throw does not reach them — 35 passing there, 4 in `test/AcrossApiClient.ts`, typecheck and lint clean. README updated to match.

src/clients/AcrossAPIClient.ts
pxrl: @droplet-rl This is undesirable. We should perhaps just escalate without throwing (i.e. return false on update(), and let the caller retry. Then we could impose that in the top-level client.update() call, such that we repeatedly call in before giving up.

droplet-rl: Reworked in 1383512e — the throw is gone.

`update()` now returns `Promise<boolean>`: `false` when the query fails (previously-fetched limits are retained, `updatedAt` left unset so a retry queries again immediately rather than waiting out the 60s retention window), `true` when limits are current — including the skip paths (`ignoreLimits`, retention window, no non-hub chains).

Retry is imposed at the top-level call in `Relayer.init()`, reusing the SDK `retry` helper already used in `HyperliquidUtils`:

```ts
retry(
  async () => {
    assert(await acrossApiClient.update(this.config.ignoreLimits), "Failed to update Across API limits");
  },
  API_UPDATE_ATTEMPTS, // 3
  API_UPDATE_BACKOFF   // 2s, exponential
),
```

The periodic `scheduleTask` caller ignores the return, which is correct — limits are retained and the next 30s tick refreshes.

One design point I want to confirm with you. `updatedLimits` only flips true on a successful fetch, so it truthfully means "limits are known" — but that also means `Relayer.ts:449`'s `updatedLimits && inputAmount.gt(limit)` check is *bypassed* while limits are unknown. So if retries are exhausted and we let init continue, the relayer would fill unconstrained (fail-open, risking 7-day stuck refunds). To avoid that, exhausted retries currently propagate out of `retry()` and fail init, so the bot restarts instead of filling blind. That's a throw at the call site rather than inside the client — if you'd rather "giving up" mean something else (e.g. proceed fail-closed by enforcing a limit of 0 until a fetch succeeds), say so and I'll switch it.

Tests: 5 passing in `test/AcrossApiClient.ts` (added coverage for the false return, the retained-limits false return, and the `ignoreLimits` skip); 40 passing across those plus the four `Relayer.*` suites. Typecheck/eslint/prettier clean, README updated. Note the `init()` retry glue itself isn't unit-tested — no test constructs a `Relayer` and calls `init()` — so that path rests on the client-level contract plus the SDK helper.

src/relayer/Relayer.ts
```diff
@@ -107,7 +110,15 @@ export class Relayer {
     const { acrossApiClient, tokenClient } = this.clients;
     await Promise.all([
       this.config.update(this.logger), // Update address filter.
-      acrossApiClient.update(this.config.ignoreLimits),
+      // The relayer does not enforce fill limits until the API limits have been fetched at least once, so retry
+      // before giving up and failing init rather than starting up and filling without limits.
+      retry(
+        async () => {
+          assert(await acrossApiClient.update(this.config.ignoreLimits), "Failed to update Across API limits");
```
pxrl: We probably don't need to assert here, but should warn loudly. With origin-only repayment options we can actually continue without limits. We are independently working to add that exception in a separate PR.

droplet-rl: Done in 93c2a2d6 — assert dropped, no throw anywhere in the path now. Moved the retry into a small `Relayer.updateLimits()` helper so `init()` stays readable:

```ts
private async updateLimits(): Promise<void> {
  const { acrossApiClient } = this.clients;

  for (let attempt = 1; attempt <= API_UPDATE_ATTEMPTS; attempt++) {
    if (await acrossApiClient.update(this.config.ignoreLimits)) {
      return;
    }

    if (attempt < API_UPDATE_ATTEMPTS) {
      await delay(API_UPDATE_BACKOFF ** attempt);
    }
  }

  this.logger.error({
    at: "Relayer::init",
    message: "😱 Unable to fetch Across API limits. Proceeding without deposit limits.",
    attempts: API_UPDATE_ATTEMPTS,
  });
}
```

3 attempts with 2s/4s backoff (≤6s added to startup worst case), then error-level and continue. I went with `logger.error` rather than `warn` for "loudly" — 46 of the 50 `logger.error` call sites in `src/` omit `notificationPath`, so plain error looked like the established loud tier. Easy to downgrade to `warn` if that is too noisy for a startup API hiccup.

Noted on the origin-only repayment exception landing separately — this leaves the relayer running with `updatedLimits === false`, so `Relayer.ts`'s limit check is simply skipped until a query succeeds, which is the state your separate PR will make explicitly safe.

40 tests passing (`test/AcrossApiClient.ts` + the four `Relayer.*` suites), typecheck/eslint/prettier clean. README adjusted — it had said init fails startup.

src/relayer/Relayer.ts
```diff
@@ -124,6 +127,31 @@ export class Relayer {
...
 
+  /**
+   * @description Refresh the Across API deposit limits, retrying on failure.
+   * @dev The relayer only enforces limits once they have been fetched successfully, so warn loudly if they never
+   * are. Filling without limits risks fills whose refunds are unavailable until the HubPool is replenished.
+   */
+  private async updateLimits(): Promise<void> {
+    const { acrossApiClient } = this.clients;
+
+    for (let attempt = 1; attempt <= API_UPDATE_ATTEMPTS; attempt++) {
```
pxrl:
```suggestion
    for (let attempt = 1; attempt <= API_UPDATE_ATTEMPTS; ++attempt) {
```

droplet-rl: Applied in 95285594.

src/clients/AcrossAPIClient.ts
droplet-rl: Done in 8d21d423 — folded into one call, so `at` and `limits` are stated once and only the level and message vary:

```ts
const updated = isDefined(liquidReserves);
if (updated) {
  tokens.forEach((token, i) => (this.limits[token.toEvmAddress()] = liquidReserves[i]));
}

this.logger[updated ? "debug" : "warn"]({
  at: "AcrossAPIClient",
  message: updated
    ? "🏁 Fetched HubPool liquid reserves"
    : `Failed to fetch HubPool liquid reserves.${this.updatedLimits ? " Retaining last known limits." : ""}`,
  limits: this.limits,
});
if (!updated) {
  return false;
}
```

10 lines of logging down to 7, and the failure message keeps the conditional "Retaining last known limits." suffix so the retain and never-fetched cases stay distinguishable. `this.logger[cond ? "warn" : "debug"]` matches the existing pattern in `Relayer.ts` (e.g. `Relayer.ts:389`, `:1091`).

Note the store-then-log order changed: `limits` is now assigned before the log rather than after, so the logged `limits` is the post-update set in both branches — same as before for each branch individually.

src/relayer/Relayer.ts
droplet-rl: Applied in 8d21d423.

pxrl (dismissed):

src/clients/AcrossAPIClient.ts
droplet-rl: Correction: this is **not** addressed on the current branch. I reverted it at Paul's request.

`8d21d423` (the consolidation quoted above) has been dropped — the branch is back to `95285594`, the exact commit you approved. Rationale: `8d21d423` landed on top of your approval and triggered the stale-review rule, which dismissed it and left the PR `BLOCKED`. Restoring the approved commit was the only way back to the tree you actually reviewed.

Note that restoring the SHA did not un-dismiss the review — GitHub keeps `4859154804` as `DISMISSED`, so `reviewDecision` is still `REVIEW_REQUIRED`. The code is byte-identical to what you approved, so a re-approve should be a formality.

So this thread's ask is still open. Two options, your call:

1. Leave it — the two log calls stay as they are, and I close this out as won't-do.
2. Re-apply the consolidation in a follow-up PR, so it doesn't disturb the approval on this one.

Happy either way; just don't want the thread reading "Done" when it isn't.

pxrl (approved):

PR #3592, opened 2026-07-19
improve(monitor): align balance report values on the decimal point

droplet-rl:
> **Stacked on #3598** — merge that first. This PR's base is #3598's branch; once #3598 lands, this retargets to master with only the formatting diff.

## Motivation

Requested by Paul in Slack: the relayer balance report padded values by raw string length and mixed precisions, so decimal points drifted between lines and the report was hard to sum by eye. Follow-up requests renamed the categories and folded the separate refunds rows into the per-token rows.

## Changes

- Add `createDecimalAligner` to `src/utils/NumberUtils.ts`: given every value that shares a column, returns a padder that left-pads integer parts and right-pads fractional parts to fixed widths, so all values align on the decimal point.
- Add `formatFixedDecimals`: `createFormatFunction(2, 4, ...)` renders values ≥ 1 with exactly 2 decimals (`toFixed(2)`) but values < 1 with 4 *significant digits* (`toPrecision(4)`), so a zero pending printed as `0.000` next to `69,539.59`. The new helper truncates every value to an exact decimal count (comma-grouped), giving uniform precision.
- Anchor the report on 4 decimal places for every token: uniform within a table, resolves ~cents on stables and ~$10 on BTC. Totals are still summed from full-precision BigNumbers.
- Rework `Monitor#reportRelayerBalances` into a ledger layout: detail lines share a left value column, each chain's `Total` sits in a second column offset to the right, and the bottom `TOTAL` (chain-cumulative) aligns under that column so it sums straight down. Labels occupy a fixed 22-column field (spaces rather than literal tabs, which render inconsistently in Slack).
- Rename `Pending` → `In-Transit`, and replace the separate per-chain `refunds` rows with a `Pending Repayment` line on each row, attributed to the exact L2 token the refund leaf pays out in via the `repaymentToken` filter from #3598. Every row now renders the same four lines (`Current` / `In-Transit` / `Pending Repayment` / `Total`), which also removes the `-` placeholder special case.
- The machine-readable Datadog debug log still reports current + in-transit only (upcoming refunds were never part of that metric).
- Unit tests for both formatting helpers in `test/NumberUtils.ts`.

## Before / after

Before:

```
Mainnet — DAI
  Current:  69,539.59
  Pending:      0.000
  Total:    69,539.59

Mainnet — refunds
  Total:     1,111.00

Blast — USDB
  Current:     329.16
  Pending:      0.000
  Total:       329.16

  TOTAL:   219,989.73
```

After:

```
Mainnet — DAI
  Current:            69,539.5937
  In-Transit:              0.0000
  Pending Repayment:   1,111.0000
  Total:                            70,650.5937

Blast — USDB
  Current:               329.1600
  In-Transit:              0.0000
  Pending Repayment:       0.0000
  Total:                               329.1600

                                   ------------
  TOTAL:                           219,989.7337
```

`Monitor#reportSpokePoolRunningBalances` has a similar misalignment; left untouched to keep this scoped, but it can adopt the same helpers in a follow-up if wanted.

No README/AGENTS updates needed: no behavior, config, or interface changes beyond the mrkdwn layout of one log message.

## Testing

- `yarn hardhat test test/BundleDataApproxClient.ts test/NumberUtils.ts test/Monitor.ts test/InventoryClient.InventoryRebalance.ts test/InventoryClient.RefundChain.ts` — 87 passing
- `yarn build`, eslint + prettier clean on touched files

### DIFF
--- a/src/monitor/Monitor.ts
+++ b/src/monitor/Monitor.ts
@@ -15,7 +15,9 @@ import {
   bnUint32Max,
   Contract,
   convertFromWei,
+  createDecimalAligner,
   createFormatFunction,
+  formatFixedDecimals,
   ERC20,
   blockExplorerLink,
   blockExplorerLinks,
@@ -474,10 +476,19 @@ export class Monitor {
 
       for (const l1Token of allL1Tokens) {
         const l1TokenDecimals = l1Token.decimals;
-        const formatWei = createFormatFunction(2, 4, false, l1TokenDecimals);
+        // Fixed 4 decimals for every token so all values share the same precision;
+        // createFormatFunction renders sub-1 values with variable precision.
+        const formatWei = (amount: string) => formatFixedDecimals(amount, l1TokenDecimals, 4);
 
         // Collect all rows first so we can compute column widths for alignment.
-        type Row = { chain: string; token: string; current: string; pending: string; total: string };
+        type Row = {
+          chain: string;
+          token: string;
+          current: string;
+          inTransit: string;
+          pendingRepayment: string;
+          total: string;
+        };
         const rows: Row[] = [];
         let tokenTotal = bnZero;
 
@@ -491,6 +502,8 @@ export class Monitor {
             continue;
           }
 
+          const canonicalL2Token = getRemoteTokenForL1Token(l1Token.address, chainId, hubChainId);
+
           for (const l2Token of l2Tokens) {
             const { symbol: l2Symbol, decimals: l2Decimals } = this.getTokenInfo(l2Token, chainId);
             const toL1Decimals = ConvertDecimals(l2Decimals, l1TokenDecimals);
@@ -499,8 +512,8 @@ export class Monitor {
             const rawBalance = balanceIndex[chainId]?.[l2Token.toNative()] ?? bnZero;
             const currentBalance = toL1Decimals(rawBalance);
 
-            // Pending: cross-chain transfers + pending L2 withdrawals (hub chain only) + pending swap rebalances.
-            let pending = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
+            // In-transit: cross-chain transfers + pending L2 withdrawals (hub chain only) + pending swap rebalances.
+            let inTransit = this.clients.crossChainTransferClient.getOutstandingCrossChainTransferAmount(
               relayer,
               chainId,
               l1Token.address,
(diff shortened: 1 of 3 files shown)


src/monitor/Monitor.ts
pxrl, line 1: Comments far too verbose. Make them concise. Max 1 - 2 lines in most cases; 3 lines or more should be exceptional.

src/utils/NumberUtils.ts
```diff
@@ -14,3 +17,45 @@ export function truncate(number: number, decimals: number): number {
...
+ * columns of values with differing decimal precision (e.g. "69,539.59" and "0.000") line up. Values
+ * with no decimal point (e.g. the "-" placeholder) are right-aligned to the integer column.
+ * @param values Every value that will share the column, used to compute the fixed widths.
+ * @returns A function padding any of those values to the common width (all outputs are equal length).
+ */
+export function createDecimalAligner(values: string[]): (value: string) => string {
+  const split = (value: string): [string, string] => {
+    const idx = value.indexOf(".");
+    return idx === -1 ? [value, ""] : [value.slice(0, idx), value.slice(idx + 1)];
+  };
```
pxrl: Is there not a native js string method that would split on the decimal point?

src/monitor/Monitor.ts
droplet-rl: Trimmed in 53e3dba — all comment blocks are now 1-2 lines (also tightened the new JSDoc prose in NumberUtils.ts).

src/utils/NumberUtils.ts
pxrl: @droplet-rl Review the above question.

droplet-rl: Good call — replaced the custom helper with native `String.prototype.split(".")` in 47e3f70. The helper only existed to normalize the no-decimal case, which a `fraction = ""` destructuring default handles. Existing tests in test/NumberUtils.ts pass unchanged.

pxrl: @droplet-rl Are you sure we don't need to handle no-decimal cases? Are we still robust to that?

droplet-rl: Yes — no-decimal values are still handled. `"1,000".split(".")` returns a single-element array, so index 1 is `undefined`: the width calculation catches that with `value.split(".")[1]?.length ?? 0`, and the align function with the `fraction = ""` destructuring default. Behavior is identical to the removed helper for every input that contains at most one `.` (always true here — inputs are pre-formatted amounts or the `"-"` placeholder).

The existing tests cover both no-decimal paths and pass against the new code: "right-aligns values without a decimal point to the integer column" (`["123.45", "-"]`) and "handles integer-only value sets" (`["7", "1,000"]`).

PR #3604, opened 2026-07-20
improve(clients): track upcoming refunds only for configured relayers

droplet-rl:
## Motivation

Prerequisite to #3598, split out per Paul in Slack.

#3602 (remove spoke pool running-balance reporter) deleted the last consumer that needed all-relayer refund sums from `BundleDataApproxClient`. Every remaining consumer queries a relayer that is known at construction time: `InventoryClient` always queries its own relayer, and the Monitor queries each of its configured `MONITORED_RELAYERS`. Tracking every relayer on every spoke is now wasted work — and #3598 is about to make the per-fill bookkeeping heavier (repayment-chain/token resolution per fill), so shrinking the fill set first keeps that PR cheap.

## Changes

- `BundleDataApproxClient` takes the set of relayers to track and skips all other relayers' fills. `InventoryClient` passes `[this.relayer]`; the Monitor passes `monitoredRelayers`. Internal per-relayer keying is unchanged (the Monitor tracks several relayers).
- Remove upcoming-deposits tracking outright: `getUpcomingDeposits` lost its only consumer in #3602. The serialized `BundleDataState` shape is **unchanged** — `upcomingDeposits` is exported empty — so Redis handover state remains compatible with older/newer relayer versions in both directions. The inventory cache key embeds the relayer address (`InventoryClient#getInventoryCacheKey`), so imported state always belongs to the same, tracked relayer.
- `getUnexecutedBundleStartBlocks` drops the `requireExecution` parameter; the non-execution path existed only for deposit counting.

No README/AGENTS updates needed: internal client behavior only; `src/clients/README.md` describes the refund-tracking role, which is unchanged.

## Testing

- `yarn hardhat test test/BundleDataApproxClient.ts test/InventoryClient.InventoryRebalance.ts test/InventoryClient.RefundChain.ts test/Monitor.ts test/Rebalancer.loadCumulativeModeBalances.ts test/JussiGraph.buildGraph.ts` — 109 passing
- New test: fills from untracked relayers are ignored; existing all-relayer sum test updated to the tracked-set semantics
- `yarn build`, eslint + prettier clean on touched files

### DIFF
--- a/src/clients/BundleDataApproxClient.ts
+++ b/src/clients/BundleDataApproxClient.ts
@@ -19,18 +19,18 @@ import { Address, bnZero } from "../utils/SDKUtils";
 import { HubPoolClient } from "./HubPoolClient";
 
 export type BundleDataState = {
+  // Deposit tracking was removed when its last consumer went away, but the field is retained (exported empty)
+  // so the serialized shape stays compatible across relayer versions.
   upcomingDeposits: { [l1Token: string]: { [chainId: number]: BigNumber } };
   upcomingRefunds: { [l1Token: string]: { [chainId: number]: { [relayer: string]: BigNumber } } };
 };
 
-// This client is used to approximate running balances and the refunds and deposits for a given L1 token. Running balances
-// can easily be estimated by taking the last validated running balance for a chain and subtracting the total deposit amount
-// on that chain since the last validated end block and adding the total refund amount on that chain since the last validated
-// end block.
+// This client approximates the upcoming refunds owed to a set of tracked relayers for a given L1 token:
+// every fill sent by a tracked relayer since the last executed bundle is assumed to be valid and refunded
+// on its selected repayment chain. Fills from other relayers are ignored.
 export class BundleDataApproxClient {
   private upcomingRefunds: { [l1Token: string]: { [chainId: number]: { [relayer: string]: BigNumber } } } | undefined =
     undefined;
-  private upcomingDeposits: { [l1Token: string]: { [chainId: number]: BigNumber } } | undefined = undefined;
   private readonly spokePoolManager: SpokePoolManager;
 
   private readonly protocolChainIdIndices: number[];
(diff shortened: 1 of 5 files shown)


src/clients/BundleDataApproxClient.ts
```diff
@@ -145,12 +143,6 @@ export class BundleDataApproxClient {
             return false;
           }
 
-          // If we don't require execution verification, return the last relayed root bundle directly.
-          // This is used for deposit counting where the boundary is bundle validation/relay, not leaf execution.
-          if (!requireExecution) {
-            return true;
-          }
```
pxrl, line 152: Why is this being removed? It seems irrelevant to the relayer cleanup.

src/clients/InventoryClient.ts
```diff
@@ -136,6 +136,7 @@ export class InventoryClient {
       this.hubPoolClient,
       this.chainIdList,
       Array.from(allL1Tokens.values()).map((l1Token) => EvmAddress.from(l1Token)),
+      [this.relayer],
```
pxrl, line 139: Does this work OK with non-EVM chain families? i.e. SVM and maybe TVM?

src/clients/BundleDataApproxClient.ts
droplet-rl: This branch existed only for deposit counting: deposits are bounded by bundle validation/relay rather than leaf execution, so `getApproximateUpcomingDepositsForToken` was the one caller passing `requireExecution = false`. #3602 removed the last consumer of `getUpcomingDeposits` (the running-balance reporter), so this PR drops upcoming-deposit tracking, and after that the only remaining caller (the refunds path) passes `true` — making this branch unreachable. The refund path is unaffected: it always required leaf execution and still does.

Happy to split the deposit-tracking removal (and this dead branch) into a separate PR if you'd rather keep this one strictly to the relayer filter.

PR #3610, opened 2026-07-21
Require exact poolRebalanceLeafCount in bundle validation

droplet-rl:
`validateRootBundle` compared the reconstructed pool rebalance root's leaf count against the live `unclaimedPoolRebalanceLeafCount` with `<`. That counter decrements as leaves execute, so the bound had to stay lenient — meaning it only rejected counts that were too high, never too low.

Compare against the immutable proposed `poolRebalanceLeafCount` instead and require exact equality. Correct in both callers (pre-liveness dispute path and post-liveness execute path, where the live counter has already decremented); no change for well-formed bundles, where reconstructed count == proposed count.

### DIFF
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1058,9 +1058,17 @@ export class Dataworker {
     };
 
     // Compare against the leaf count committed at proposal time, not the live unclaimed count (which decrements
-    // as leaves execute). The reconstructed root must contain exactly the proposed number of leaves.
-    const proposedPoolRebalanceLeafCount =
-      this.clients.hubPoolClient.getLatestProposedRootBundle().poolRebalanceLeafCount;
+    // as leaves execute). Look up the proposal matching the root bundle under validation rather than assuming it's
+    // the latest proposal, since this function can validate historical bundles (e.g. via the validateRootBundle
+    // script). The reconstructed root must contain exactly the proposed number of leaves.
+    const matchingProposal = this.clients.hubPoolClient
+      .getProposedRootBundles()
+      .find((proposal) => proposal.blockNumber === rootBundle.proposalBlockNumber);
+    assert(
+      isDefined(matchingProposal),
+      "validateRootBundle: no proposal found at rootBundle.proposalBlockNumber; increase HubPoolClient lookback"
+    );
+    const proposedPoolRebalanceLeafCount = matchingProposal.poolRebalanceLeafCount;
 
     if (
       expectedPoolRebalanceRoot.leaves.length !== proposedPoolRebalanceLeafCount ||
@@ -1097,6 +1105,7 @@ export class Dataworker {
         expectedSlowRelayRoot: expectedSlowRelayRoot.tree.getHexRoot(),
         pendingRoot: rootBundle.poolRebalanceRoot,
         pendingPoolRebalanceLeafCount: rootBundle.unclaimedPoolRebalanceLeafCount,
+        proposedPoolRebalanceLeafCount,
       });
     } else if (expectedRelayerRefundRoot.tree.getHexRoot() !== rootBundle.relayerRefundRoot) {
       this.logger.debug({
--- a/src/dataworker/Dataworker.ts
+++ b/src/dataworker/Dataworker.ts
@@ -1057,11 +1057,13 @@ export class Dataworker {
       slowRelayTree: expectedSlowRelayRoot,
     };
 
+    // Compare against the leaf count committed at proposal time, not the live unclaimed count (which decrements
+    // as leaves execute). The reconstructed root must contain exactly the proposed number of leaves.
+    const proposedPoolRebalanceLeafCount =
+      this.clients.hubPoolClient.getLatestProposedRootBundle().poolRebalanceLeafCount;
+
     if (
-      // Its ok if there are fewer unclaimed leaves than in the reconstructed root, because some of the leaves
-      // might already have been executed, but its an issue if the reconstructed root expects fewer leaves than there
-      // are left to execute because it means that the unclaimed count can never drop to 0.
-      expectedPoolRebalanceRoot.leaves.length < rootBundle.unclaimedPoolRebalanceLeafCount ||
+      expectedPoolRebalanceRoot.leaves.length !== proposedPoolRebalanceLeafCount ||
       expectedPoolRebalanceRoot.tree.getHexRoot() !== rootBundle.poolRebalanceRoot
     ) {
       this.logger.debug({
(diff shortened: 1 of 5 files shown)


src/dataworker/Dataworker.ts
```diff
@@ -1057,11 +1057,13 @@ export class Dataworker {
       slowRelayTree: expectedSlowRelayRoot,
     };
 
+    // Compare against the leaf count committed at proposal time, not the live unclaimed count (which decrements
+    // as leaves execute). The reconstructed root must contain exactly the proposed number of leaves.
+    const proposedPoolRebalanceLeafCount =
+      this.clients.hubPoolClient.getLatestProposedRootBundle().poolRebalanceLeafCount;
```
pxrl: @droplet-rl Does this also work when run via the validateRootBundle script? In that case it is likely evaluating a historical deposit, so the latest proposed root bundle is very likely to be a different proposal.

droplet-rl: Good catch — it did not. `getLatestProposedRootBundle()` is only correct when validating the live pending bundle; via the `validateRootBundle` script the bundle under validation is likely historical, so the latest proposal would be a different bundle and the leaf-count comparison would use the wrong reference.

Fixed by looking up the proposal at `rootBundle.proposalBlockNumber` instead. That field resolves to the correct proposal on both paths: the SDK builds the pending bundle with `proposalBlockNumber` taken from its matching `ProposeRootBundle` event, and the script sets it from the historical proposal event it reconstructs. Also added an assert for the (currently unreachable) case where the HubPoolClient lookback does not cover the proposal block, and included `proposedPoolRebalanceLeafCount` in the dispute debug log.

src/dataworker/Dataworker.ts
```diff
@@ -1057,11 +1057,21 @@ export class Dataworker {
       slowRelayTree: expectedSlowRelayRoot,
     };
 
+    // Compare against the leaf count committed at proposal time, not the live unclaimed count (which decrements
+    // as leaves execute). Look up the proposal matching the root bundle under validation rather than assuming it's
+    // the latest proposal, since this function can validate historical bundles (e.g. via the validateRootBundle
+    // script). The reconstructed root must contain exactly the proposed number of leaves.
+    const matchingProposal = this.clients.hubPoolClient
+      .getProposedRootBundles()
+      .find((proposal) => proposal.blockNumber === rootBundle.proposalBlockNumber);
```
pxrl: There could theoretically be multiple proposals in a single block number - i.e. propose, dispute, propose in a single block.

droplet-rl: True — and no combination of event fields can fully disambiguate that case either, since two same-block proposals share the block timestamp (hence `challengePeriodEndTimestamp`) and could even share identical roots while claiming different leaf counts, which is exactly the field this check depends on.

So I removed the lookup entirely and instead pass the `ProposeRootBundle` event into `validateRootBundle` as a parameter. Every caller already knows unambiguously which proposal it is validating: the disputer/executor paths pass `getLatestProposedRootBundle()` (the same event the SDK builds the pending bundle from in `HubPoolClient.update`), and the `validateRootBundle` script passes the `precedingProposeRootBundleEvent` it already resolved.

src/dataworker/Dataworker.ts
```diff
@@ -1095,6 +1102,7 @@ export class Dataworker {
         expectedSlowRelayRoot: expectedSlowRelayRoot.tree.getHexRoot(),
         pendingRoot: rootBundle.poolRebalanceRoot,
         pendingPoolRebalanceLeafCount: rootBundle.unclaimedPoolRebalanceLeafCount,
```
pxrl: We probably want to drop this.

src/dataworker/Dataworker.ts
```diff
@@ -1580,6 +1588,8 @@ export class Dataworker {
       hubPoolChainId,
       widestPossibleExpectedBlockRange,
       pendingRootBundle,
+      // The pending root bundle is always constructed from the most recent proposal (see HubPoolClient.update).
+      this.clients.hubPoolClient.getLatestProposedRootBundle(),
```
pxrl: Why is `pendingRootBundle` above not sufficient here?

src/dataworker/Dataworker.ts
droplet-rl: Done — dropped it. The check no longer references the unclaimed count, so only `proposedPoolRebalanceLeafCount` remains in the log.

src/dataworker/Dataworker.ts
droplet-rl: `pendingRootBundle` is a `PendingRootBundle`, which only carries `unclaimedPoolRebalanceLeafCount` — the *live* on-chain counter that decrements as leaves execute. The check needs the count the proposer originally committed, and that only exists on the `ProposeRootBundle` event as `poolRebalanceLeafCount`; it isn't a field on `PendingRootBundle`.

In this executor path specifically, the two differ: by the time we re-validate, some leaves may already have executed, so `unclaimedPoolRebalanceLeafCount < poolRebalanceLeafCount`. Passing the event gives us the invariant proposed count.

(If we'd rather not thread the event through, the alternative is to add `poolRebalanceLeafCount` to `PendingRootBundle` in the SDK so `pendingRootBundle` alone would suffice — happy to do that instead if you prefer.)

src/dataworker/Dataworker.ts
```diff
@@ -1057,11 +1057,12 @@ export class Dataworker {
...
+    // Compare against the leaf count committed at proposal time, not the live unclaimed count (which decrements
+    // as leaves execute). The reconstructed root must contain exactly the proposed number of leaves.
+    const proposedPoolRebalanceLeafCount = rootBundle.poolRebalanceLeafCount;
+
     if (
-      // Its ok if there are fewer unclaimed leaves than in the reconstructed root, because some of the leaves
-      // might already have been executed, but its an issue if the reconstructed root expects fewer leaves than there
-      // are left to execute because it means that the unclaimed count can never drop to 0.
-      expectedPoolRebalanceRoot.leaves.length < rootBundle.unclaimedPoolRebalanceLeafCount ||
+      expectedPoolRebalanceRoot.leaves.length !== proposedPoolRebalanceLeafCount ||
```
pxrl: I don't think we need a separate variable here - just reference the one directly in rootBundle. Also update this down at line 1098.

```suggestion
      expectedPoolRebalanceRoot.leaves.length !== rootBundle.proposedPoolRebalanceLeafCount ||
```

droplet-rl: Done — dropped the variable and inlined the reference in both the condition and the debug log. One note: the field on `PendingRootBundle` in sdk 4.4.14 is `poolRebalanceLeafCount` (no `proposed` prefix), so the code uses `rootBundle.poolRebalanceLeafCount` rather than the suggestion's `rootBundle.proposedPoolRebalanceLeafCount`; the log key keeps the more descriptive `proposedPoolRebalanceLeafCount` name.

pxrl (dismissed):

pxrl (approved):

PR #3614, opened 2026-07-22
feat: Support Tron origins on the v3 deposit-execute path

amateima:
## Summary

Adds Tron (TVM) support to the deposit-address bot's **v3 deposit-execute path**. The bot stays a thin submitter — the quote-api builds all calldata; the bot relays funding context, validates the response, and submits through the existing TVM path in `TransactionClient`.

- **Namespace guard** (`initiateDepositV3`) now validates namespaces against the origin chain's family (`evm` ⇒ EVM chains, `tron` ⇒ Tron). svm and cross-family anomalies (e.g. a `tron` namespace on an EVM chainId) keep skipping with a warn.
- **On-chain reads and log matching convert base58 → 0x-hex** via `getEthersCompatibleAddress`: `getDepositAddressBalance` (Tron providers speak eth-JSON-RPC) and `buildDepositExecutedPayload` (`hexZeroPad` throws on base58 *after* the execute is committed, which would have rejected the whole poll batch).
- **`_getExecuteTx`** re-encodes `executionFeeRecipient` to the origin-native form (base58 on Tron; the TVM submitter reuses the same EVM key, so the fee recipient stays the bot). All other Tron fields relay verbatim from the indexer.
- **`_validateExecuteResponse`** compares deposit addresses canonically through the hex form (base58 is case-sensitive; a hex echo of the same address also passes) and requires `executeTx.ecosystem` to match the chain family (`"tvm"` on Tron).
- **`getExecuteContract`** now connects the provider in dispatcher mode — raw-tx simulation (`willSucceed`) and `_runTransactionTvm` both read `contract.provider`, which was previously null in that mode — and defensively hex-converts the target address.
- `DepositAddressExecuteResponse.executeTx.ecosystem` widened to `"evm" | "tvm"`.

Out of scope: the v3 refund-withdraw path stays EVM-only (Tron mis_routes keep being skipped).

## Rollout

- No new env gate: add `728126428` to `RELAYER_ORIGIN_CHAINS` to activate (providers + dedup sets are provisioned from that list).
- The bot signer's derived Tron account must hold TRX for fees (default feeLimit 100 TRX, `TVM_GAS_LIMIT` to override).
- Prod runs deposit executions with `ENABLE_EXECUTE_ERC20_TRANSFER_METADATA=true`, so provenance is on-chain and no `deposit_executed` Pub/Sub is published; the payload-builder fix is a safety net for other gate combinations.

## Verification

- 74 tests passing, including 12 new Tron cases (namespace guard, request mapping, response validation, payload builder). Lint clean.
- Probed the quote-api preview (`deposit-addresses/execute`) with the exact request the new code builds: HTTP 200, `ecosystem: "tvm"`, 0x-hex `to`, base58 `depositAddress` echo, base58 `userAddress`/`executionFeeRecipient` accepted.
- Pre-existing `tsc` errors on master (`poolRebalanceLeafCount` in dataworker/scripts vs installed SDK) are unrelated; no errors in changed files.

### DIFF
--- a/src/deposit-address/DepositAddressHandler.ts
+++ b/src/deposit-address/DepositAddressHandler.ts
@@ -1194,7 +1194,7 @@ export class DepositAddressHandler {
       });
       return false;
     }
-    const expectedEcosystem = chainIsTvm(originChainId) ? "tron" : "evm";
+    const expectedEcosystem = chainIsTvm(originChainId) ? "tvm" : "evm";
     if (executeTx.ecosystem !== expectedEcosystem) {
       this.logger.warn({
         ...logContext,
--- a/src/deposit-address/DepositAddressHandler.ts
+++ b/src/deposit-address/DepositAddressHandler.ts
@@ -26,6 +26,9 @@ import {
   TransactionReceipt,
   getCurrentTime,
   isHttpError,
+  chainIsEvm,
+  chainIsTvm,
+  getEthersCompatibleAddress,
 } from "../utils";
 import { getRedisCache, RedisCacheInterface } from "../cache/Redis";
 import {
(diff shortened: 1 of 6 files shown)


src/deposit-address/DepositAddressHandler.ts
```diff
@@ -1155,7 +1174,11 @@ export class DepositAddressHandler {
       chainId: originChainId,
     };
 
-    if (executeResponse.depositAddress.toLowerCase() !== depositAddress.toLowerCase()) {
+    // Compare canonically: base58 is case-sensitive, so map both sides through the 0x-hex form
+    // (no-op on EVM, where this degrades to the previous lowercase compare). Also accepts a
+    // hex-encoded response for a base58-funded address.
+    const canonicalAddress = (address: string) => getEthersCompatibleAddress(originChainId, address).toLowerCase();
+    if (canonicalAddress(executeResponse.depositAddress) !== canonicalAddress(depositAddress)) {
```
pxrl, line 1181: fwiw we have the Address class for this - it handles comparisons despite encoding differences. I think that'd be worth looking at - also when passing address types around internally, since I see `string` typng on functions here.

pxrl (approved):

src/deposit-address/DepositAddressHandler.ts
amateima: I will follow up with a PR that parses all addresses with the SDK utility type 👍

PR #3653, opened 2026-08-04
improve(TransactionClient): Bump fees on timeout resubmission

droplet-rl:
## Motivation

During the 2026-08-04 PDA fill-latency breach a mainnet sweep sat unmined for 10 blocks. The confirmation-timeout path resubmitted four times with no fee bump — each replacement was economically identical to the transaction it was meant to displace. Fee escalation only triggers on `REPLACEMENT_UNDERPRICED`, which privately-routed RPC submissions (QuickNode/Alchemy private routes) never return.

Slack thread: https://risklabs.slack.com/archives/C08E4H3FSRX/p1785847036698639 · Companion: UMAprotocol/bot-configs#4256

## Changes (+24/−11)

The existing `REPLACEMENT_UNDERPRICED` escalation is extracted into `_bumpRetryScaler()` — `max(priorityFeeScaler, 1.1)` compounding per attempt, capped by `MAX_GAS_RETRY_SCALER_DEFAULT[_<chainId>]` — and timeout replacements now run the same escalation, carried as `AugmentedTransaction.retryScaler` into the existing `retryScaler` mechanics in `_scaleGasPrice` (delta applied to both fee caps, satisfying nodes' dual-bump replacement rule). With prod config that's 1.5×/attempt on mainnet. Also guards a latent NaN when a chain has no priority-fee scaler configured.

## Testing

`hardhat test test/TransactionClient.ts` 16 passing (scaler assertion folded into the transient-RPC resubmission test); tsc/eslint/prettier clean.

Note: a prior revision also floored replacement caps at 1.1× the outgoing transaction's fees (guarding a falling gas estimate between attempts). Dropped for diff budget — recoverable from `e6683a1` if wanted.


src/clients/TransactionClient.ts
```diff
@@ -238,11 +244,17 @@ export class TransactionClient {
               if (blockNumber < startBlock + CONFIRMATION_BLOCKS) {
                 continue; // Loop: retry the wait (bounded by maxTries).
               }
+              // Bump fees on the replacement: the outgoing transaction was evidently not priced for
+              // inclusion, and nodes holding it will reject a replacement that doesn't raise both fee
+              // caps. Don't rely on REPLACEMENT_UNDERPRICED to escalate — RPCs that route transactions
+              // privately (no public mempool insertion) never emit it.
+              const bumpedScaler = _bumpRetryScaler(chainId, retryScaler);
```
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in e6683a15. The outgoing fee caps (`gasPrice`/`maxFeePerGas`/`maxPriorityFeePerGas`) from the timed-out `TransactionResponse` are now threaded through `_submit` → `_runTransaction` → `_scaleGasPrice`, which floors each replacement cap at 1.1x its predecessor (geth's default price-bump threshold, matching its floored integer arithmetic). The scaled fresh estimate still wins whenever it clears the floor, and the floor survives `_runTransaction`'s internal retries. Covered by new unit tests on `_scaleGasPrice` (falling estimate, rising estimate, legacy gasPrice) plus a threading test on the timeout path.

src/clients/TransactionClient.ts
```diff
@@ -242,7 +244,9 @@ export class TransactionClient {
                 ...common,
                 message: `Transaction on ${chain} timed out at nonce ${txnResponse.nonce}, resubmitting...`,
               });
-              return this._submit(txn, { nonce: txnResponse.nonce, maxTries: maxTries - 1 });
+              // Bump fees on the replacement; private RPC routes never emit REPLACEMENT_UNDERPRICED.
```
pxrl: It's speculation that private RPC routes haven't emitted REPLACEMENT_UNDERPRICED. I don't think this can be taken as absolute truth. There could be other scenarios where it doesn't hold up - i.e. a pending txn in the public mempool and an attempted private replacement. We haven't verified this, so don't be so concrete.

src/clients/TransactionClient.ts
```diff
@@ -242,7 +244,9 @@ export class TransactionClient {
...
+              const bumped = { ...txn, retryScaler: _bumpRetryScaler(chainId, txn.retryScaler ?? 1.0) };
+              return this._submit(bumped, { nonce: txnResponse.nonce, maxTries: maxTries - 1 });
```
pxrl: Why do we loop via this._submit() instead of recursing via runTransaction, as is done for replacements?

src/clients/TransactionClient.ts
droplet-rl: Fair — unproven. Softened in bc2027cf to state only what was observed (code comment + README).

src/clients/TransactionClient.ts
droplet-rl: `_runTransaction` recursion handles submission-time rejections, where nothing was accepted. Here a live transaction is outstanding, so the replacement needs the confirmation flow re-armed against the new hash (wait binding, fresh block baseline, ability to replace again) — that machinery lives in `_submit`, and recursing there reuses it with `maxTries` as the single bound. Same pattern as the pre-existing TRANSACTION_REPLACED branch above (from #3643); this PR only adds the scaler.

PR #3655, opened 2026-08-04
fix(relayer): evaluate HubPool liquid reserves per repayment chain

droplet-rl:
## Motivation

Reported by @pxrl in Slack: when the relayer is forced to take origin chain repayment, it should not refuse fills just because the API reports `maxDeposit`/liquid reserves of 0.

The `/liquid-reserves` limit exists so the relayer doesn't take a refund the HubPool can't fund, which would leave that refund stuck for as long as it takes liquidity to return. That risk only applies to refunds taken **away from the origin chain** — an origin chain refund is funded by the deposit itself. So the constraint is a property of the repayment chain, not of the deposit, and it can't be evaluated before a repayment chain has been resolved.

`filterDeposit()` evaluated it at the filter stage and exempted only deposits that the *protocol* forces onto origin repayment (`fromLiteChain`, or no PoolRebalanceRoute for the input token, via `depositForcesOriginChainRepayment()`). Deposits forced onto origin repayment by **inventory config** — `forceOriginRepayment` / `forceOriginRepaymentPerChain`, via `InventoryClient.shouldForceOriginRepayment()` — were still refused when reserves were exhausted, even though `getPossibleRepaymentChainIds()` then returns `[originChainId]` alone and no HubPool liquidity is ever required.

## Changes

- **`filterFundableRepaymentChains()`** (`src/relayer/Relayer.ts`) drops candidate repayment chains whose refund can't currently be funded by the HubPool. Origin chain repayment always survives, so a forced-origin deposit is never refused for want of reserves. Where inventory ranked a non-origin chain first and reserves are short, the fill now proceeds with origin repayment instead of being dropped outright. The hub-origin exemption in `getLimit()` (`uint256Max`, JIT-bridgeable) is unchanged, and the limit isn't even resolved when origin is the only candidate.
- **An unfundable repayment is handled as an unprofitable fill.** When the filter empties a non-empty eligible list, `resolveRepaymentChain()` reports no repayment chain — exactly as for any other failure to identify one — and `evaluateFill()` routes it through the existing mechanism: `captureUnprofitableFill()` plus suppression in `ignoredDeposits` until the next `runMaintenance()` flush. Now that a failed `/liquid-reserves` query retains the last known limits rather than reporting zero (#3591), utilisation high enough to refuse every eligible refund is a real condition rather than a transient blip, and a refund the HubPool can't fund may stall settlement. The unfundable/over-allocated distinction survives only as a local in `resolveRepaymentChain()`, to select the log level and message.
- **Unmapped input tokens keep origin-only repayment.** An input token with no L1 token mapping has no limit to compare against, so there's no basis to call a non-origin refund fundable. This is reachable rather than hypothetical: `filterDeposit()` drops an unmapped input token only when no swap route covers it, and the mapping is symbol-based rather than PoolRebalanceRoute-based.

The `fillRelay()` submit-time forced-origin guard is untouched.

> **Note:** this branch originally also fixed `callLimits()` returning all-zeroes on a failed query. That fix landed independently as #3591 and is now on master; merging master left only a `callLimits()` JSDoc and a corrected invalid-response message here (`/liquid-reserves` returns one entry per l1Token, not a `maxDeposit` field).

## Testing

`test/Relayer.BasicFill.ts`:

- fills a deposit that must take origin chain repayment while reserves report 0 (the reported bug)
- drops an unfundable preferred chain and fills with origin repayment instead
- treats a deposit whose only eligible chain is unfundable as an unprofitable fill, and confirms it stays suppressed once captured
- drops non-origin repayment chains for an input token with no L1 token mapping
- doesn't interpret a failed `/liquid-reserves` query as zero available liquidity (drives the real `fetchWithTimeout` failure path)

45 passing across `Relayer.BasicFill`, `AcrossApiClient`, `Relayer.UnfilledDeposits`, `Relayer.TokenShortfall` and `Relayer.SlowFill`. `yarn typecheck`, eslint and prettier clean.

## Docs

`docs/relayer-fill-decision-flow.md` and `docs/repayment-selection.md` updated per the repo's documentation-maintenance rule, including the caveat that origin repayment is only strictly self-funding when the deposit and refund land in the same bundle.


src/relayer/Relayer.ts
```diff
@@ -523,6 +503,61 @@ export class Relayer {
...
+   */
+  protected filterFundableRepaymentChains(deposit: DepositWithBlock, repaymentChainIds: number[]): number[] {
+    const { acrossApiClient, inventoryClient } = this.clients;
+    const { depositId, inputAmount, inputToken, originChainId, txnRef } = deposit;
+
+    // There's nothing to constrain, so don't bother resolving a limit. Deposits that are forced to take origin chain
+    // repayment always land here.
+    if (repaymentChainIds.every((chainId) => chainId === originChainId)) {
+      return repaymentChainIds;
+    }
```
pxrl: Would we expect this to be [originChainId] alone, or could there be duplicates in the array?

docs/repayment-selection.md
pxrl, line 1: @droplet-rl far too verbose. Keep this concise.

src/relayer/Relayer.ts
droplet-rl: No duplicates are possible today — `determineRefundChainId()` never produces them:

- `InventoryClient.ts:739` builds `rankedChains` via `dedupArray(standardCandidateChains)`, and `standardChainOrder` only ever holds origin + destination anyway.
- Both fallback pushes are `includes()`-guarded (`:834` origin-on-quick-rebalance, `:839` hub chain), so neither can re-add a chain.
- Every other exit returns at most one element: `[]`, `[originChainId]`, `[destinationChainId]`, or `[chainOverride]`.

So the input here is either `[]` or `[originChainId]` — i.e. `[originChainId]` alone, as you suspected.

I kept `every()` rather than tightening to `length === 1 && [0] === originChainId` for two reasons: it also folds the empty case into the same branch (an empty candidate list is unconstrained, and resolving a limit for it cannot change the outcome), and this is a `protected` helper whose contract is "return a subset of the input" — that reads better if it does not silently depend on the caller having deduped. `every()` costs nothing over the length check for a 1-2 element array.

Pushed e4f1840 to record the invariant in the comment, since it was not obvious from the call site. Happy to switch to the explicit length check if you would rather have the tighter assertion.

src/relayer/Relayer.ts
```diff
@@ -523,6 +503,62 @@ export class Relayer {
...
+
+    if (this.config.ignoreLimits || !acrossApiClient.updatedLimits) {
+      return repaymentChainIds;
+    }
+
+    // Note: the relayer only queries limits for supported tokens, so an unmapped token has no constraint to apply.
+    const l1Token = inventoryClient.getL1TokenAddress(inputToken, originChainId);
+    if (!isDefined(l1Token)) {
+      return repaymentChainIds;
+    }
```
pxrl: Should this be [originChainId] ?

src/relayer/Relayer.ts
```diff
@@ -523,6 +503,62 @@ export class Relayer {
...
+        originChainId,
+        inputToken,
+        inputAmount,
+        l1Token,
+        limit,
+        repaymentChainIds,
+        fundableChainIds,
+        txnRef,
+      });
+    }
```
pxrl: Too convoluted - we don't care about one repayment chain vs. another, with the exception of originChainId.

src/relayer/Relayer.ts
droplet-rl: Good catch — yes, and it turns out this is reachable rather than merely defensive, so I have changed it in e7362c3.

I initially assumed `filterDeposit()` had already dropped unmapped input tokens, making the branch dead. It has not: `Relayer.ts:309-315` only drops them when **no swap route covers the deposit** (`!swapSupported && !isDefined(l1Token)`). And the two mappings involved are different lookups:

- `inventoryClient.getL1TokenAddress()` → `getInventoryEquivalentL1TokenAddress()` (`TokenUtils.ts:71-83`), which resolves via `TOKEN_SYMBOLS_MAP` / `TOKEN_EQUIVALENCE_REMAPPING`.
- `canTakeDestinationChainRepayment()` (`InventoryClient.ts:627-644`), which reads on-chain PoolRebalanceRoutes.

So the symbol-based mapping can be absent while the HubPool does permit destination chain repayment — a newly-listed token looks exactly like this before the SDK constant lands. In that state the old code returned every candidate unconstrained, waving a non-origin refund through on precisely the deposits whose limit we could not check. That is the risk this filter exists to prevent.

One deliberate deviation from your suggestion: I used `repaymentChainIds.filter((chainId) => chainId === originChainId)` rather than a literal `[originChainId]`. A literal would satisfy the downstream `assert(isDefined(fee))` — `batchComputeLpFees()` covers `getPossibleRepaymentChainIds()`, which always includes origin — but it can inject a chain that `determineRefundChainId()` deliberately excluded (e.g. origin dropped for over-allocation), silently overriding inventory from inside a liquidity filter. The filter keeps the documented "subset of `repaymentChainIds`" contract; if origin was not a candidate it yields `[]`, which `resolveRepaymentChain()` already handles as `unfundableRepayment` with a warn and a retryable skip.

Trade-off worth naming: in the inventory-management-disabled case, candidates are `[destinationChainId]` alone, so the filter yields `[]` and we defer the fill instead of taking origin repayment. A literal `[originChainId]` would keep that fill, and with inventory management off there is no allocation objection to origin. If you would rather preserve the fill there, say so and I will special-case it — I opted for the conservative reading since we cannot verify the limit at all in that state.

Also folded the drop into the existing debug log with a distinct message so it is not silent, and added a regression test (verified failing before the change, passing after). `docs/repayment-selection.md` gained a bullet under "Step 2a".

docs/repayment-selection.md
pxrl, line 1: @droplet-rl Still too verbose. Keep it super concise.

droplet-rl: Cut in f68b702. Step 2a is now one paragraph — 28 lines down to 5:

> `filterFundableRepaymentChains()` restricts repayment to the origin chain when `AcrossApiClient`'s `/liquid-reserves` limits show the HubPool can't fund a refund elsewhere. An origin chain refund is funded by the deposit itself, so it always survives; forced-origin deposits are therefore never refused for want of reserves. The filter is inert when `ignoreLimits` is set or limits have not been fetched. If nothing survives, the deposit follows the normal unprofitable-fill path.

Apologies for the second round — I missed your first comment on this file entirely and then made the section *longer*, which is the opposite of what you asked for.

Also corrected two things that went stale when `unfundableRepayment` was dropped: the mermaid node still named `UnfundableRepaymentSkipFill`, and the Failure outcomes bullet still claimed the deposit is retried.

src/relayer/Relayer.ts
droplet-rl: Agreed, and sorry — this comment sat unaddressed for two days, I only caught it now while reviewing the full thread. Simplified in f68b702.

Taking "we don't care about one repayment chain vs. another, with the exception of originChainId" literally removed most of the block:

- the `fundableChainIds.length !== repaymentChainIds.length` comparison is gone. The two early returns already establish that a non-origin chain is present and that the limit is exceeded, so by the time we reach the log the filter always drops something — there was nothing to compare.
- the two-way log message is gone, replaced by one line.
- the unmapped-`l1Token` case folded into the limit rather than getting its own branch: no mapping means no limit, which is `bnZero`, which restricts to origin repayment by the same path as an exhausted reserve. That also removed the `BigNumber | undefined` plumbing.

Net: 57 lines to 40, and the tail is now just a log plus `repaymentChainIds.filter((chainId) => chainId === originChainId)`.

src/relayer/Relayer.ts
```diff
@@ -551,6 +531,49 @@ export class Relayer {
...
   }
 
+  /**
+   * Restrict repayment to the origin chain when the HubPool can't currently fund a refund anywhere else. An origin
+   * chain refund is funded by the deposit itself; a refund on any other chain draws on HubPool liquidity.
+   * @param deposit Deposit under evaluation.
+   * @param repaymentChainIds Candidate repayment chain IDs, ordered from highest to lowest priority.
+   * @returns The subset of repaymentChainIds whose refund can be funded, preserving the input ordering.
+   */
+  protected filterFundableRepaymentChains(deposit: DepositWithBlock, repaymentChainIds: number[]): number[] {
```
pxrl:
```suggestion
  protected filterRepaymentChains(deposit: DepositWithBlock, repaymentChainIds: number[]): number[] {
```

src/relayer/Relayer.ts
```diff
@@ -551,6 +531,49 @@ export class Relayer {
...
+      return repaymentChainIds;
+    }
+
+    // Only non-origin repayment draws on HubPool liquidity, so there's nothing to constrain.
+    if (repaymentChainIds.every((chainId) => chainId === originChainId)) {
+      return repaymentChainIds;
+    }
+
+    // An unmapped input token has no limit to compare against, so treat it as no liquidity available.
+    const l1Token = inventoryClient.getL1TokenAddress(inputToken, originChainId);
```
pxrl: @droplet-rl COnfirm that this function _only_ returns tokens mapped via a HubPool pool rebalance route, and does not include any additional token mappings.

src/relayer/Relayer.ts
```diff
@@ -1311,16 +1334,33 @@ export class Relayer {
     const destinationChain = getNetworkName(destinationChainId);
 
     const mark = this.profiler.start("resolveRepaymentChain");
-    const preferredChainIds = await inventoryClient.determineRefundChainId(deposit);
+    const eligibleChainIds = await inventoryClient.determineRefundChainId(deposit);
+
+    // Origin chain repayment survives this filter unconditionally, so a deposit that is forced to take origin chain
+    // repayment is never refused on the basis of HubPool liquid reserves.
+    const preferredChainIds = this.filterFundableRepaymentChains(deposit, eligibleChainIds);
```
pxrl:
```suggestion
    const availableChainIds = await inventoryClient.determineRefundChainId(deposit);

    // Origin chain repayment survives this filter unconditionally, so a deposit that is forced to take origin chain
    // repayment is never refused on the basis of HubPool liquid reserves.
    const eligibleChainIds = this.filterFundableRepaymentChains(deposit, availableChainIds);
```

src/relayer/Relayer.ts
droplet-rl: Renamed in 8845572, including the two doc references and the log `at:` field.

src/relayer/Relayer.ts
```diff
@@ -1283,19 +1326,34 @@ export class Relayer {
...
```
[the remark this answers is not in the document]
droplet-rl: This one holds up, and is the only one of the three bot comments I think is worth a decision.

The fallback branch fires on `!preferredChainIds.includes(destinationChainId)`, which cannot distinguish "destination was never eligible" from "destination was just dropped for liquidity". In the latter case, with origin surviving and unprofitable, it sets `preferredChain = preferredChainIds[0]` (origin) and fills at a known loss, justified by a destination quote for a chain we just declined to fund. No stuck-refund risk — the refund is taken on origin — but it converts a temporary liquidity deferral into a deliberate loss.

@pxrl a fix does not need the "which chains were dropped" plumbing you asked me to remove; the existing helper answers it directly by adding one condition to the fallback `if`:

```ts
this.filterRepaymentChains(deposit, [destinationChainId]).length > 0
```

`[]` when destination is unfundable, and unchanged behaviour when `ignoreLimits` is set or limits are unfetched. I have deliberately not applied it — it is a behaviour change beyond the rename you asked for, and it is your call on relayer economics. Happy to add it here or leave it for a follow-up.

src/relayer/Relayer.ts
```diff
@@ -1283,19 +1326,34 @@ export class Relayer {
...
+    const unfundableRepayment = preferredChainIds.length === 0 && eligibleChainIds.length > 0;
```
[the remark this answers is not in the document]
droplet-rl: Partly stale, partly deliberate. `unfundableRepayment` no longer exists — 0d071d6 removed it in favour of the ordinary unprofitable-fill path, so there is no longer an expression here to get wrong.

The underlying observation stands but "not retried" overstates it: `evaluateFill()` records the deposit in `ignoredDeposits` (`Relayer.ts:871`), and that map is flushed wholesale in `runMaintenance()` (`Relayer.ts:201`). So the deposit is re-evaluated on the next maintenance cycle rather than blacklisted for the run. Accepting that delay was the deliberate trade for not carrying per-deposit liquidity state.

src/relayer/Relayer.ts
```diff
@@ -810,6 +846,12 @@ export class Relayer {
       },
     } = await this.resolveRepaymentChain(deposit, lpFees);
 
+    // A repayment that the HubPool can't currently fund is not an unprofitable fill; HubPool liquidity recovers as LPs
+    // deposit and as bundles execute, so leave this deposit eligible for re-evaluation on subsequent loops.
+    if (unfundableRepayment) {
+      return;
```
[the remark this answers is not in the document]
droplet-rl: Accurate but minor. `computeRequiredDepositConfirmations(unfilledDeposits, ...)` (`Relayer.ts:1077`) does run before `evaluateFills()`, so a deposit that later turns out to need unavailable HubPool liquidity still contributes to its origin chain MDC bucket for that loop.

Fixing it means classifying fundability before the aggregate pass, i.e. hoisting the liquidity check back to filter stage — which is precisely what this PR moves downstream, and for good reason. Not worth reintroducing that coupling to shave one loop of confirmation sizing; noting it rather than acting on it.

src/relayer/Relayer.ts
droplet-rl: Confirmed — and the answer is **no**. `getL1TokenAddress()` is a symbol-map lookup, not a PoolRebalanceRoute lookup, and it deliberately includes additional mappings.

The chain:

- `InventoryClient.getL1TokenAddress()` (`InventoryClient.ts:570-576`) → `getInventoryEquivalentL1TokenAddress()`, which is **this repo's** `src/utils/TokenUtils.ts:71-83`, not the SDK's.
- That first tries the SDK's `getL1TokenAddress()` (`@across-protocol/sdk` `utils/TokenUtils.ts:147`): scan `TOKEN_SYMBOLS_MAP` for a token whose `addresses[chainId]` matches, return its mainnet/sepolia address. No HubPool involvement at all.
- On throw it falls back to `getTokenInfo()` + `TOKEN_EQUIVALENCE_REMAPPING[symbol]`. That table is 19 entries today: `USDC.e`/`USDbC`/`USDzC`/`USDH`/`USDG`/`pathUSD`/`USDC-BNB`/`USDC-PERPS`/`USDC-SPOT` → `USDC`, `USDT-BNB`/`USDT-SPOT` → `USDT`, `WBTC-BNB` → `WBTC`, `USDB` → `DAI`, `ETH` → `WETH`, `BNB` → `WBNB`, `HYPE` → `WHYPE`, `XPL` → `WXPL`, `AVAX` → `WAVAX`, `LGHO` → `WGHO`.
- Its own docstring says as much: *"Returns the L1 token that is equivalent to the `l2Token` within the context of the inventory. This is used to link tokens that are **not** linked via pool rebalance routes, for example."*
- The `hubChainId` argument is the only HubPool-derived input and it's just a chain id, used to index `TOKEN_SYMBOLS_MAP`. The route-based lookup is a different function entirely: `hubPoolClient.getL1TokenForL2TokenAtBlock()` / `l2TokenHasPoolRebalanceRoute()`, which is what `canTakeDestinationChainRepayment()` uses (`InventoryClient.ts:627-644`).

Three things that bound how much this matters here.

**1. It is not a change of behaviour.** The check this PR moves used the identical resolution on master: `filterDeposit()` computed `l1Token = inventoryClient.getL1TokenAddress(inputToken, originChainId)` — still there at `Relayer.ts:330` for the `relayerTokens` filter — and passed it straight into `acrossApiClient.getLimit(originChainId, l1Token)`. Same resolver, same key. The PR changes *where* the limit is evaluated, not *which token* it resolves.

**2. The map it indexes is route-keyed, and a miss fails safe.** `AcrossApiClient.limits` is populated from `tokensQuery ∩ hubPoolClient.getL1Tokens()` (`AcrossAPIClient.ts:81-82`, `:105`) — i.e. HubPool-registered L1 tokens. So a symbol-derived key is looked up in a route-derived map. On a miss `getLimit()` warns and returns `bnZero` (`:124-130`), which restricts to origin repayment. Wrong-direction failure isn't possible from a *missing* key.

**3. A divergence is only reachable on one path.** `determineRefundChainId()` resolves the same symbol-based token itself at `InventoryClient.ts:694` via `getRequiredL1TokenAddress()`, which **asserts** it is defined; then per candidate chain it asserts `_l1TokenEnabledForChain()` (`:745`) and, for every non-origin chain, `l2TokenHasPoolRebalanceRoute(repaymentToken, chainId)` (`:751`). So on every path that can return a non-origin candidate, the symbol-derived L1 token has already been tied back to a real PoolRebalanceRoute before we ever look up a limit — a bogus mapping throws loudly there rather than silently reaching `getLimit()`.

The one exception is the inventory-management-disabled early return (`:666`), which returns `[destinationChainId]` without resolving an L1 token at all. That is exactly the trade-off I flagged on [r3716271792](https://github.com/across-protocol/relayer/pull/3655#discussion_r3716271792): in that mode a token with a PoolRebalanceRoute but no `TOKEN_SYMBOLS_MAP` entry yet resolves to `undefined` → `bnZero` → the fill is deferred rather than taking destination repayment. Still your call whether to special-case it.

If you'd rather the limit key be route-derived — `hubPoolClient.getL1TokenForL2TokenAtBlock(inputToken, originChainId)` is the semantically exact key for "which pooled token funds this refund", and it matches how `limits` is keyed — I'm happy to switch it. It would diverge from the l1Token `determineRefundChainId()` itself uses, and from master, so I haven't done it unprompted.

src/relayer/Relayer.ts
droplet-rl: Applied in e6f96b4a, with the downstream rename the suggestion implies — `preferredChainIds` no longer exists, so the whole function had to follow or it wouldn't compile.

Settled vocabulary in `resolveRepaymentChain()`:

- `availableChainIds` — raw output of `determineRefundChainId()`
- `eligibleChainIds` — what survives `filterRepaymentChains()`
- `preferredChain` — the one finally selected (unchanged, along with `preferredChainIndex` / the `preferredChainId` callback param, which are all singular)

Two notes on things the rename touches beyond the three lines:

- The profiler `mark.stop()` payload field changes name, `preferredChainIds` → `eligibleChainIds` (`Relayer.ts:1382`). Flagging in case anything greps for it.
- I left the log *prose* alone ("Preferred chains [...] are not profitable", `:1492`, `:1518`, `:1544`) to keep those greppable. Only string I did retouch is the unfundable message, which said "no eligible repayment chain" while interpolating what is now `availableChainIds`; it now reads "no available repayment chain".

One overlap worth naming: `determineRefundChainId()`'s own docstring already uses a four-stage vocabulary — possibleChains → candidateChains → rankedChains → **eligibleChains** — so `InventoryClient` calls its return value "eligible" while the Relayer now calls the same array "available". Happy to reconcile if that bothers you; I've kept your naming as given.

`tsc --noEmit`, eslint and prettier clean; 45 passing across `Relayer.BasicFill`, `AcrossApiClient`, `Relayer.UnfilledDeposits`, `Relayer.TokenShortfall`, `Relayer.SlowFill`.

docs/repayment-selection.md
droplet-rl: Closing the loop on this one directly — I answered your repeat of it on [r3732042872](https://github.com/across-protocol/relayer/pull/3655#discussion_r3732042872) but never replied here, which is how it came to need repeating in the first place.

Fixed in f68b702. Step 2a went 28 lines → 5, one paragraph, and is unchanged since:

> `filterRepaymentChains()` restricts repayment to the origin chain when `AcrossApiClient`'s `/liquid-reserves` limits show the HubPool can't fund a refund elsewhere. An origin chain refund is funded by the deposit itself, so it always survives; forced-origin deposits are therefore never refused for want of reserves. The filter is inert when `ignoreLimits` is set or limits have not been fetched. If nothing survives, the deposit follows the normal unprofitable-fill path.

`docs/repayment-selection.md` is 133 lines total; this PR's net contribution to it is those five lines plus a mermaid node rename.

src/relayer/Relayer.ts
```diff
@@ -551,6 +531,49 @@ export class Relayer {
...
   }
 
+  /**
+   * Restrict repayment to the origin chain when the HubPool can't currently fund a refund anywhere else. An origin
+   * chain refund is funded by the deposit itself; a refund on any other chain draws on HubPool liquidity.
+   * @param deposit Deposit under evaluation.
+   * @param repaymentChainIds Candidate repayment chain IDs, ordered from highest to lowest priority.
+   * @returns The subset of repaymentChainIds whose refund can be funded, preserving the input ordering.
+   */
+  protected filterRepaymentChains(deposit: DepositWithBlock, repaymentChainIds: number[]): number[] {
```
pxrl, line 541: If originChainId is the hub chain then we don't want to avoid taking repayment on a valid pool rebalance route. In general this is what helps to keep utilisation low.

droplet-rl: Agreed, and there was a real hole here. Fixed in b329876.

`getLimit()` already encodes exactly your point — hub chain origin returns `uint256Max` with the comment *"Funds can be JIT-bridged from mainnet to anywhere, so don't apply any constraint"* (`AcrossAPIClient.ts:118-122`) — so for a mapped token a hub chain origin was already unconstrained.

The hole is that my unmapped-token fallback never calls `getLimit()`:

```ts
const limit = isDefined(l1Token) ? acrossApiClient.getLimit(originChainId, l1Token) : bnZero;
```

`bnZero` short-circuits straight past the exemption, so a hub chain origin with an unmapped input token was being restricted to origin repayment — precisely the case where we most want the refund taken out on a pool rebalance route. Backwards, as you say.

Fixed by checking the hub chain up front rather than relying on `getLimit()` to do it, so the exemption holds whether or not the token resolves:

```ts
// Funds can be JIT-bridged from the hub chain to anywhere, so a hub chain origin is never constrained; taking
// repayment out on a pool rebalance route is what keeps utilisation low. getLimit() encodes the same exemption,
// but an unmapped input token never reaches it.
if (originChainId === hubPoolClient.chainId) {
  return repaymentChainIds;
}
```

It does duplicate `getLimit()`'s first branch. I judged that worth it over threading a sentinel through the unmapped case, but if you'd rather have one source of truth I can drop the `bnZero` fallback and have the unmapped case resolve its limit some other way.

Added a regression test (`Doesn't constrain hub chain origins`), verified failing without the fix. The doc sentence now reads "inert for hub chain origins ..., when `ignoreLimits` is set, and until limits have been fetched" — no net lines added. 52 passing across `Relayer.*` + `AcrossApiClient`; tsc/eslint/prettier clean.

src/relayer/Relayer.ts
pxrl: @droplet-rl I think we should apply the change. The existing behaviour seems incorrect even though I haven't seen any instances of it misbehaving.

droplet-rl: Applied in 5b432f0.

```ts
const l1Token = hubPoolClient.getL1TokenForL2TokenAtBlock(
  inputToken,
  originChainId,
  hubPoolClient.latestHeightSearched
);
const limit = isDefined(l1Token) ? acrossApiClient.getLimit(originChainId, l1Token) : bnZero;
```

`getL1TokenForL2TokenAtBlock()` returns `undefined` rather than throwing, and `l2TokenHasPoolRebalanceRoute()` is literally `isDefined()` of the same call, so no separate route check is needed. I passed `latestHeightSearched` explicitly rather than taking the `Number.MAX_SAFE_INTEGER` default, so it resolves at the same hub block `depositForcesOriginChainRepayment()` uses — that matters below.

**Consequence worth knowing: the unresolved-token branch is now unreachable.** An input token with no pool rebalance route makes `depositForcesOriginChainRepayment()` true, so `determineRefundChainId()` returns `[originChainId]` or `[]` and we return at the all-origin check before any limit is resolved. Since both now read the same mapping at the same block they cannot disagree. I kept the `bnZero` fallback as a fail-safe and marked it `@dev` as defensive.

**I deleted a test.** `Drops non-origin repayment chains for an input token with no L1 token mapping` existed only to cover the symbol-map/route divergence — the case where `getL1TokenAddress()` returned `undefined` while a route existed. That is exactly what this change eliminates, and reaching the branch now requires stubbing `getL1TokenForL2TokenAtBlock()` to `undefined`, which simultaneously flips the deposit to forced-origin and makes the assertion pass for the wrong reason. Keeping it would have been theatre.

Also simplified `Doesn't constrain hub chain origins`: its `getL1TokenAddress` / `isSwapSupported` stubs were there to fake an unmapped token, which is no longer the relevant lever. Re-ran the mutation check after trimming — still fails without the hub chain early return.

One honest caveat on verification: no test distinguishes the two resolvers, because in the fixtures both resolve to the same L1 token. This is a semantic hardening against a divergence that the test setup cannot produce, so the suite confirms no regression but does not prove the new key is better. Your reasoning is the justification here, not the tests.

51 passing across `Relayer.*` + `AcrossApiClient` (was 52; the deleted test accounts for the difference). tsc/eslint/prettier clean.

src/relayer/Relayer.ts
```diff
@@ -551,6 +531,62 @@ export class Relayer {
...
+      originChainId,
+      inputToken,
+      inputAmount,
+      l1Token,
+      limit,
+      repaymentChainIds,
+      txnRef,
+    });
+
+    return repaymentChainIds.filter((chainId) => chainId === originChainId);
```
pxrl, line 589: Why not just return [originChainId] here?

droplet-rl: Because `[originChainId]` would fill on an origin chain that inventory deliberately withheld, and the codebase already has a rule against exactly that.

At this point in the function we know at least one non-origin candidate exists — but not that origin is among them. `determineRefundChainId()` routinely returns candidate sets without it: `[destinationChainId]` when inventory management is off, `[chainOverride]`, or `[destinationChainId, hubChainId]` when origin is over-allocated. So the two forms differ whenever origin was withheld:

- `.filter(...)` → `[]` → no repayment chain → the ordinary unprofitable-fill path. No fill.
- `[originChainId]` → fill, taking repayment on the chain inventory just declined to offer.

`InventoryClient.ts:824-831` already decides this case, and decides it the other way:

```ts
// ... if the deposit originated on a lite chain, which forces fillers to take repayment on the origin
// chain, and the origin chain is not an eligible repayment chain, then we shouldn't fill this deposit otherwise
// the filler will be forced to be over-allocated on the origin chain, which could be very difficult to withdraw
// funds from.
if (forceOriginRepayment && (eligibleChains.length !== 1 || !eligibleChains.includes(originChainId))) {
  return [];
}
```

This filter puts a deposit in the same position — forced onto origin repayment — so returning `[originChainId]` would contradict that rule for the liquidity case specifically, and pile onto a chain that is hard to withdraw from. The `.filter()` reaches `[]` and the deposit is skipped, which is what that code already prefers.

Mechanically the literal is safe, to be clear — `batchComputeLpFees()` covers `getPossibleRepaymentChainIds()`, which always includes origin, so the downstream `assert(isDefined(fee))` would hold. The objection is policy, not crashes.

Since this is the second time it has come up, I added the reasoning at the return in 67abe40 so the next reader gets it without asking. Comment only — behaviour unchanged, 51 passing, tsc/eslint/prettier clean.

Still happy to switch if you want the fill preserved over the allocation guard; it is a one-word change and your call. I just did not want to make it silently.

pxrl (approved):

PR #3659, opened 2026-08-05
fix(finalizer): stop holding zkSync withdrawals for 2.6h after they are claimable

droplet-rl:
## Problem

The zkSync finalizer ignores any `TokensBridged` event newer than 6 hours:

```ts
// Zksync takes ~6 hours to finalize so ignore any events
// earlier than that.
getCurrentTime() - 60 * 60 * 6,
```

That assumption is stale. zkSync withdrawals are claimable in **~3.6h**, so the filter holds every withdrawal back for ~2.6h after the funds are already claimable on L1.

## Measurement

21 zkSync withdrawals between 2026-07-16 and 2026-08-04, taken from `evm.tokens_bridged` on chain 324. For each, I timed the L1 batch lifecycle via `zks_getTransactionDetails` plus the **mainnet block timestamps** of the commit/prove/execute txns (the node's own `committedAt`/`provenAt`/`executedAt` strings are DB-write times and lag L1 by 13-17 min, so they are not a valid basis):

| stage | delay after L2 withdrawal |
|---|---|
| batch commit | ~0.3h |
| batch prove | ~1.0h |
| **batch execute — claimable here** | **3.4-3.8h, median 3.6h** |

The spread is ~25 minutes across three weeks, i.e. a fixed delay rather than variable proving load.

Observed end to end in `#zbot-across-finalizer`: withdrawals became claimable at ~3.6h, but the finalizer did not announce them until 6.1-6.5h — consistent with this filter, and too large to be run-cadence jitter (the sweeper runs every ~30 min).

## Change

Threshold 6h → 2h, named and env-overridable.

The filter is only a cheap pre-filter to avoid querying status for withdrawals that cannot possibly be ready. `sortWithdrawals()` (requires L2 status `finalized`) and `filterMessageLogs()` (`isWithdrawalFinalized`) are the authoritative readiness checks, so a low value costs a few extra RPC calls per run and **cannot** produce a bad finalization attempt. I kept 2h rather than 0 so the common case still skips status queries for brand-new withdrawals, and made it overridable via `ZKSYNC_MIN_WITHDRAWAL_AGE` so a future change in zkSync's batch cadence doesn't need a redeploy.

Expected effect: finalization latency drops from ~6.3h to ~3.6h plus up to one run interval.

## Notes for the reviewer

- **This will not by itself make any zkSync withdrawal finalize today.** Separately, no finalization has landed on the hub chain since 2026-07-31 21:31 UTC — 26.54 WETH + 25,903.85 USDT of zkSync withdrawals are currently claimable-but-unclaimed, plus Lens/Mode/Optimism withdrawals and OP-stack proofs. That is a submission-path problem in the shared chain-1 multicall batch, not a zkSync one, and is reported separately. This PR is the latency fix only.
- Data-quality note for anyone repeating the measurement: `receivedAt` from `zks_getTransactionDetails` was wrong on 5 of the 21 txs (stamped with re-ingestion wall-clock, producing negative latencies of -210 to -259h). Use the L2 block timestamp as t0.
- No `README.md`/`AGENTS.md` updates proposed: no doc in the repo documents this threshold, and the constant is self-describing at its definition. Happy to add one if you'd like it surfaced in `src/finalizer/`.
- Verified locally: `tsc --noEmit` clean, prettier clean, eslint clean. There is no existing unit test covering `zkSyncFinalizer`, so nothing to extend; the change is a constant and a comment.


src/finalizer/utils/zkSync.ts
```diff
@@ -40,6 +40,13 @@ const IGNORED_WITHDRAWALS = [
   "0xe93642e22eec21ead2abb20f23a1dc3033b41274cdfe7439cf3ada3dfa1dff06", // Lens USDC 2025-06-13 @todo remove
 ];
 
+// Lower bound on the age of a zkSync withdrawal before it is worth checking for finalization. This is only a cheap
+// pre-filter to avoid querying status for withdrawals that cannot possibly be ready; sortWithdrawals() and
+// isWithdrawalFinalized() below are the authoritative readiness checks, so setting this too low costs a few extra
+// RPC calls per run, never a bad finalization attempt. Overridable so a change in zkSync's batch cadence doesn't
+// need a redeploy.
+const ZKSYNC_MIN_WITHDRAWAL_AGE = Number(process.env["ZKSYNC_MIN_WITHDRAWAL_AGE"] ?? 2 * 60 * 60);
```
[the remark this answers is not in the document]
droplet-rl: Good catch — taken your second option: dropped the override in 4dcb4c8, keeping a plain `MIN_WITHDRAWAL_AGE_SECONDS = 2 * 60 * 60` with the units in the name.

Reasoning for removing rather than documenting:

- There is no documented home for finalizer knobs today — `LZ_API_CHUNK_SIZE`, `SVM_COMPUTE_UNIT_OVERRIDE`, `HELIOS_EXECUTE_MESSAGE_GAS_LIMIT_<chain>` and `BINANCE_API_BASE` are all undocumented, so adding a README section for this one knob alone would have been inconsistent.
- The override was speculative. This filter sits in front of the authoritative readiness checks (`sortWithdrawals()`, and `isWithdrawalFinalized()` in `filterMessageLogs()`), so mistuning it in either direction only shifts RPC volume — it cannot cause a bad finalization attempt. If zkSync's batch cadence changes, editing the constant is a one-line PR.
- Beyond the units ambiguity you flagged, it was unvalidated: a non-numeric value made `Number(...)` return `NaN`, and `getCurrentTime() - NaN` would have reached `getBlockForTimestamp()` as a `NaN` timestamp and failed confusingly.

The 6h → 2h threshold change itself — the actual fix — is unchanged.

src/finalizer/utils/zkSync.ts
```diff
@@ -40,6 +40,15 @@ const IGNORED_WITHDRAWALS = [
   "0xe93642e22eec21ead2abb20f23a1dc3033b41274cdfe7439cf3ada3dfa1dff06", // Lens USDC 2025-06-13 @todo remove
 ];
 
+// Lower bound on the age of a zkSync withdrawal before it is worth querying its finalization status. A withdrawal
+// becomes finalizable once the L1 batch containing it is executed on the hub chain. Measured over 21 withdrawals
+// between 2026-07-16 and 2026-08-04, execution lands 3.4-3.8h after the L2 withdrawal (batch commit ~0.3h, prove
+// ~1h, execute ~3.6h) -- not the ~6h this filter previously assumed, which held back every withdrawal for ~2.6h
+// after it was already claimable. This is only a cheap pre-filter: sortWithdrawals() and the isWithdrawalFinalized()
+// checks in filterMessageLogs() are authoritative, so a value below zkSync's true batch cadence costs a few extra
+// RPC calls per run, never a bad finalization attempt.
```
pxrl: Make comment more concise.

Additionally, instead of hardcoding the min withdrawal age, can it be read or inferred dynamically from either L1 or L2?

droplet-rl: Yes — it can be read directly, and that turns out to be strictly better than any constant. Done in a818323; comment trimmed to three lines in the same commit.

A withdrawal is claimable exactly when the L1 batch containing it has been executed on the hub chain. That is not a proxy for readiness, it *is* the condition the L1 contract enforces: executing a batch is what stores the root that `proveL2MessageInclusion()` checks. So the cutoff is now:

```ts
const mainContract = await wallet.getMainContract();
const executedBatch = (await mainContract.getTotalBatchesExecuted()).toNumber();
const batchBlockRange = await l2Provider.getL1BatchBlockRange(executedBatch);
const latestBlockToFinalize = batchBlockRange[1];
```

Verified against mainnet rather than assumed:

- `getTotalBatchesExecuted()` on `0x3240...0324` returned **515161**. `zks_getL1BatchDetails` shows 515161 with `executedAt: 2026-08-05T11:41:18Z` and 515162 with `executedAt: null` — so the counter is exactly the highest executed batch.
- Note `zks_L1BatchNumber` was 515168 at the same moment, i.e. the latest *sealed* batch, 7 ahead of executed. That one would have been the wrong thing to read.
- The filter still compares L2 block numbers, which relies on batch ranges being monotonic. They are contiguous: 515160 ends `0x4423a70`, 515161 spans `0x4423a71`-`0x4423afd`, 515162 starts `0x4423afe`. So `block <= last block of batch N` ⟺ `batch <= N`.
- Same semantics on Lens, which shares this finalizer: `zks_getMainContract` → `0xc29d...3cef`, executed 8962, batch 8963 not yet executed.

Two side benefits: it drops the `getBlockForTimestamp()` binary search and its Redis dependency in favour of two direct reads, and there is no longer anything to retune when zkSync's batch cadence moves — which is what caused the original 6h value to go stale.

pxrl (approved):

PR #3661, opened 2026-08-05
fix(finalizer): finalize Lens asset-router withdrawals via finalizeDeposit

droplet-rl:
## Problem

A Lens WETH withdrawal — **19.2128 ETH**, `0xa964075c…12cec8`, initiated 2026-07-31 15:17 UTC — cannot be finalized. The call reverts `InvalidProof()` (`0x09bde339`).

It is also the **poison pill that has blocked every hub-chain finalization since 2026-07-31 21:31 UTC**, because the finalizer bundles them all into one all-or-nothing `Multicall3.aggregate()`. That bundling problem is fixed separately in #3660; this PR fixes the call itself.

## Cause

`prepareFinalization()` picks the L1 entrypoint **by chain**:

```ts
const isLegacyBridge = [CHAIN_IDs.LENS].includes(l2ChainId);
```

Lens therefore goes through `finalizeWithdrawal()`, which does not carry `l2Sender` and leaves the L1 contract to assume it. That assumption doesn't hold for withdrawals initiated by the **L2 asset router** — every ERC-20, and on Lens, whose base token is GHO rather than ETH, that includes WETH. The reconstructed message hash mismatches and the proof check fails.

## Fix

Select the entrypoint by the withdrawal's actual `l2Sender`:

| `l2Sender` | entrypoint | why |
|---|---|---|
| asset router `0x…010003` | `finalizeDeposit(struct)` | passes `l2Sender` explicitly |
| standalone ZkStack USDC bridge | `finalizeWithdrawal()` | the only entrypoint it implements |
| Lens base token (GHO) | `finalizeWithdrawal()` | unchanged |

## Verification

Simulated both entrypoints against mainnet state on 2026-08-05, using params derived from `zksync-ethers` 5.11.2 (the version this repo pins), i.e. the same library and calls the bot itself makes:

```
Lens WETH   (asset router)  finalizeWithdrawal -> InvalidProof()   finalizeDeposit -> OK (211,646 gas)
Lens USDC   (usdcBridge)    finalizeWithdrawal -> OK               finalizeDeposit -> no such function
zkSync ETH  (base token)    finalizeDeposit    -> OK   (5 withdrawals)
zkSync USDT (asset router)  finalizeDeposit    -> OK
```

- Extracted the rule into an exported `useLegacyFinalizeWithdrawal()` and pinned all five verified combinations in a new `test/ZkSyncFinalizer.ts` (6 passing).
- **Confirmed the test fails against the previous chain-based behaviour** (2 failing: `expected true to equal false`) and passes with the fix.
- `tsc --noEmit`, prettier, eslint clean.

## Reviewer notes

- **Deliberately conservative on Lens base-token (GHO) withdrawals.** Every historical one is already finalized, so there is no pending case to simulate, and I did not want to change a path I cannot verify. Given that base-token withdrawals from zkSync finalize fine via `finalizeDeposit`, I suspect the Lens legacy special-case is obsolete entirely and could be deleted — but that wants one pending GHO withdrawal to confirm first. Happy to follow up.
- No `README.md`/`AGENTS.md` update proposed: no doc describes the entrypoint choice, and the rule is now documented on the exported function it lives in. Say the word if you'd like it surfaced in `src/clients/README.md` alongside the finalizer batching notes.
- Related: `IGNORED_WITHDRAWALS` at the top of `zkSync.ts` hardcodes a Lens USDC withdrawal with `@todo remove`, which looks like an earlier instance of a poison pill being papered over. Left alone here — with #3660 merged a reverting withdrawal is excluded and reported loudly rather than blocking the batch, so that entry can probably go. Worth its own PR.

### DIFF
--- a/src/finalizer/utils/zkSync.ts
+++ b/src/finalizer/utils/zkSync.ts
@@ -1,6 +1,6 @@
 import { interfaces, utils as sdkUtils } from "@across-protocol/sdk";
 import { Contract, Signer } from "ethers";
-import { Provider as zksProvider, Wallet as zkWallet } from "zksync-ethers";
+import { Provider as zksProvider, Wallet as zkWallet, utils as zksUtils } from "zksync-ethers";
 import { HubPoolClient, SpokePoolClient } from "../../clients";
 import { CONTRACT_ADDRESSES, getContractEntry } from "../../common";
 import {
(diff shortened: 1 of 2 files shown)


src/finalizer/utils/zkSync.ts
```diff
@@ -214,6 +214,31 @@ async function getWithdrawalParams(
...
+ * @param requiresCustomUsdcBridge Whether this withdrawal routes via the standalone ZkStack USDC bridge.
+ * @returns true if the withdrawal must be finalized via the legacy finalizeWithdrawal() entrypoint.
+ */
+export function useLegacyFinalizeWithdrawal(
+  l2ChainId: number,
+  l2Sender: string,
+  requiresCustomUsdcBridge: boolean
+): boolean {
+  const isAssetRouterWithdrawal = l2Sender.toLowerCase() === zksUtils.L2_ASSET_ROUTER_ADDRESS.toLowerCase();
+  return requiresCustomUsdcBridge || ([CHAIN_IDs.LENS].includes(l2ChainId) && !isAssetRouterWithdrawal);
```
pxrl: `requiresCustomUsdcBridge` seems totally redundant here; if the caller knows it needs a custom bridge then it shouldn't be calling `useLegacyFinalizedWithdrawal`. Refactor at the call site to avoid that.

droplet-rl: Agreed — done in 64b60ce. `useLegacyFinalizeWithdrawal(l2ChainId, l2Sender)` is now just the L1Nullifier sender rule, and the USDC-bridge case is decided in `prepareFinalizations()`, which already calls `withdrawalRequiresCustomUsdcBridge()` to pick the contract:

```ts
const useLegacyEntrypoint =
  withdrawalRequiresCustomUsdcBridge(l1ChainId, chainId, l2TokenAddress) ||
  useLegacyFinalizeWithdrawal(l2ChainId, withdrawal.sender);
```

`prepareFinalization()` takes the resolved flag rather than re-deriving it.

pxrl (approved):

PR #3662, opened 2026-08-05
fix(finalizer): pad the gas limit on tryAggregate finalization batches

droplet-rl:
## Problem

`eth_estimateGas` must not be trusted on `Multicall3.tryAggregate(requireSuccess=false, ...)`. The estimator finds the lowest limit at which the *outer* transaction succeeds — and a batch whose inner calls all ran out of gas still succeeds, because `tryAggregate` catches their reverts and reports `success=false` per call. **The estimate is a floor, not a requirement.**

The finalizer submitted it raw. No `gasLimit` was set, so `willSucceed()` filled in the bare estimate, and `MULTICALL3_AGGREGATE_GAS_MULTIPLIER` never applied — it only reaches bundles built by `buildMultiSenderBundle()`, and a single-element multisender chunk passes through untouched.

On 2026-08-05 this discarded a **76,064.59 USDC** CCTP v2 mint on Arbitrum ([`0x6c8809…e3d01`](https://arbiscan.io/tx/0x6c880910dcf00e269be73d7ceef584bd59abe10e1c7b33ae4521a0f5c07e3d01)), in a transaction that mined with `status: 1` and no logs. Decoding its return value gives `success = false`; `callTracer` shows the mint executing in a sub-frame and being rolled back by an OOG in the enclosing frame.

Worth dwelling on: **the limit of 165,215 exceeded the batch's true cost of 162,070 and still starved the inner call.** EIP-150 caps what a frame may forward at 63/64 of its remaining gas, so only 128,606 reached `receiveMessage`, which needed ~131k. The shortfall is about how gas is *distributed across frames*, not how much is available overall — so "just give it more" only works if the padding is explicit. Omitting the limit does not do that; it means "spend exactly the estimate", which is what shipped.

## Fix

Declare `gasLimitMultiplier` on the batch, so the existing `TransactionClient` padding applies. **Three lines in `src/finalizer/index.ts` plus the constant.**

A fixed multiplier is sound rather than merely hopeful because the shortfall is bounded: it is the reserve withheld from the largest call, hence at most 1/64 (~1.6%) of the batch, and it *shrinks* as the batch grows. Equivalently, the true requirement never exceeds `estimate × 64/63` (~1.016×). **1.1×** keeps ~6× headroom over that bound.

Deliberately not higher: the finalizer does not chunk a batch's inner calls, so a large batch's limit has to stay clear of chain block gas limits. Finalization gas isn't the cost to optimise; an unsubmittable transaction is. This revision lowered the multiplier from 1.5× to 1.1× on review — 1.5× was ~32× the measured requirement, and that unneeded excess was what created the block-limit exposure (crossover moves from ~67% of a block limit to ~91%).

## Verification

`test/MultiCallerClient.TryAggregateGas.test.ts` characterises the estimator against **real Multicall3 bytecode** (via `deployMulticall3`) across three batch shapes — one large call, one large among small, uniformly sized — bisecting for the true requirement and asserting:

1. the estimate never exceeds the true requirement (it really is a floor),
2. the shortfall stays within the EIP-150 1/64 reserve (the bound the multiplier relies on),
3. `estimate × multiplier` clears the requirement.

Measured shortfalls track `(largest single call)/64`: a 3.07M single call is short by 47,600 (1/64 = 48.0k); five 267k calls short by 4,307 (1/64 = 4.2k).

**The padding is load-bearing:** at `1.0×` — today's raw-estimate behaviour — every case fails. Re-running the incident's own calldata on Arbitrum agrees: `success = false` at 165,215, `success = true` once padded.

Measured `required / estimate` per shape: **1.01550**, **1.01374**, **1.00323** — all under the `64/63` = 1.01587 bound, which a fourth case now asserts against the configured multiplier directly, so the guarantee covers shapes the three measured ones don't reach.

`yarn tsc`, `yarn lint`, and the existing MultiCaller/TryMulticall/finalizer suites pass.

## Diff size

97 added lines, of which **85 are the test and its gas-burner mock**; the runtime change is 10. Happy to split the test into a follow-up PR if you'd rather review the fix on its own — say the word.

An earlier revision computed the limit bottom-up from per-call `estimateGas` results in `preflightFinalizations()`. Measurement showed that was unjustified: the estimator's error is bounded at ~1.6%, so it bought no real precision while adding an RPC per call, a new drop-on-unestimable failure mode, and *more* over-provisioning than a multiplier on large batches. Replaced with the multiplier; the measurement survives as the test.

## Not addressed here

The reason this went unnoticed is a **separate** bug, deliberately out of scope: `submission()` in `src/finalizer/index.ts` derives success from `txnRefLookup[destinationChainId]?.length > 0` — "some tx landed on this chain" — not from whether *this message's* call succeeded. So the batch logged `info` + `Finalized … 🪃` for a no-op. This PR makes the failure much less likely; it does not make it *detectable*. Until `tryAggregate`'s per-call `success[]` flags (or the expected events) are checked, alerts of this shape remain unfalsifiable. Happy to follow up.


src/clients/README.md
```diff
@@ -102,6 +102,14 @@ This client is responsible for submitting transactions on-chain and therefore fo
...
+
+Callers that build such a batch must supply their own `gasLimit`, sized bottom-up from per-call estimates via `computeTryAggregateGasLimit()` (`src/utils/TransactionUtils.ts`). Note that supplying `gasLimit` also causes `willSucceed()` to skip simulating the batch — which is the intent, since the caller is expected to have simulated each call individually, and an individual simulation is strictly more informative than simulating the wrapper that swallows the reverts. The finalizer does this in `preflightFinalizations()`.
+
+The correction is not merely a safety margin. `computeTryAggregateGasLimit()` scales by 64/63 because EIP-150 caps what a frame may forward at 63/64 of its remaining gas, so the reserve must sit *on top of* the inner requirement. In the 2026-08-05 incident that motivated this, the naive estimate (165,215) exceeded the batch's true cost (162,070) and still starved the inner call: a limit that merely covers total cost is not sufficient.
```
pxrl: Too verbose. Keep it concise.

pxrl (approved):

src/clients/README.md
droplet-rl: Cut it down to four sentences: the estimator returns a floor, `tryAggregate` batches must therefore declare a multiplier, and covering total cost isn't sufficient because the shortfall is the EIP-150 reserve. Dropped the incident narrative and the restated rationale — that lives in the commit message.

Also trimmed the `MULTICALL3_TRY_AGGREGATE_GAS_MULTIPLIER` comment from 5 lines to 3 for the same reason; it was duplicating the README.

src/common/Constants.ts
```diff
@@ -229,6 +229,13 @@ export const DEFAULT_RELAYER_GAS_MESSAGE_MULTIPLIER = "1.0"; // Multiplier on pr
 
 export const DEFAULT_MULTICALL_CHUNK_SIZE = 50;
 
+// Padding for a Multicall3 tryAggregate() batch, which must not be submitted on a bare eth_estimateGas: a batch
+// whose inner calls all ran out of gas still "succeeds", so the estimate is a floor, not a requirement. The
+// shortfall is the EIP-150 reserve withheld from the largest call, hence <= 1/64 of the batch; this is ~30x that.
+// Not higher: the finalizer doesn't chunk inner calls, so large batches must stay under block gas limits.
+// See src/clients/README.md, and test/MultiCallerClient.TryAggregateGas.test.ts for the measurements.
+export const MULTICALL3_TRY_AGGREGATE_GAS_MULTIPLIER = 1.5;
```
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in f42a22a, lowered 1.5x → 1.1x.

You are right that nothing caps the padded value (`TransactionClient.ts:301-310` multiplies and submits; there is no block-gas-limit awareness anywhere in the repo), and right that the inner calls are not chunked — the finalizer enqueues one `tryAggregate` per chain, and `MultiCallerClient` passes a single-element multisender chunk through untouched, so the batch reaches the wire whole.

But the more useful conclusion is that **1.5x was never justified in the first place**, and the unneeded excess was precisely what created the exposure you describe. The only shortfall `eth_estimateGas` leaves unaccounted for is the reserve the *outer* frame withholds — everything inside the inner call is already simulated — so the true requirement never exceeds `estimate * 64/63` (~1.016x). Measured over the test's three shapes:

| shape | required | estimate | shortfall | required/estimate |
|---|---|---|---|---|
| single large call | 3,118,599 | 3,070,999 | 47,600 | 1.01550 |
| large among small | 3,136,198 | 3,093,687 | 42,511 | 1.01374 |
| uniformly sized | 1,338,363 | 1,334,056 | 4,307 | 1.00323 |

All under 64/63 = 1.01587. So 1.5x was ~32x the needed padding. 1.1x keeps ~6x headroom over the bound, matches the sibling `MULTICALL3_AGGREGATE_GAS_MULTIPLIER` (`MultiCallerClient.ts:68`), and moves your crossover from ~67% of a block gas limit to ~91% — at which point the batch is too large to submit unchunked regardless of padding, so a cap would not save it either.

I also added a fourth case asserting the multiplier against the `64/63` bound directly, so the guarantee holds for shapes the three measured ones don't reach instead of only for those three.

Not doing the capped-multiplier or gas-based chunking you suggest: a cap needs per-chain block-limit plumbing that does not exist yet, and chunking inner calls is a real change to finalizer batching. Both are worth doing if backlogs ever approach that size — but they address unsubmittable-batch sizing, which is a separate problem from this estimator bug.

PR #3671, opened 2026-08-06
fix(finalizer): size finalization batches from their calls' own estimates

droplet-rl:
Supersedes #3666 and #3667, collapsed into one change at @pxrl's suggestion: estimate the calls individually, sum them, slice at a gas boundary.

## Why the estimate of a batch is useless

`tryAggregate()` catches inner reverts, so a batch whose calls all ran out of gas still succeeds and `eth_estimateGas` prices the *failure*. #3662 padded that estimate, on the premise that the shortfall is bounded by the EIP-150 reserve — which holds only while a call's requirement follows what it consumes. OP-stack `SafeCall.callWithMinGas` gates on `gasleft() >= minGas * 64/63`, where `minGas` was declared by the withdrawal on L2 and is unrelated to consumption, so no multiplier reaches it.

Measured on mainnet `0x1b7974b8`, which logged "Batch finalized 3 txns" and finalized one:

| | declared `minGas` | headroom required | gas forwarded | |
|---|---|---|---|---|
| Optimism, 18,866.83 USDT | 5,393,173 | 5,478,982 | 1,075,273 | ✗ |
| Lisk, 14,506.34 LSK | 516,982 | 525,391 | 938,486 | ✓ |
| Mode, 200.27 ETH | 5,392,149 | 5,477,941 | 747,280 | ✗ |

The `tryAggregate()` estimate was **757,755**, shipped at ×1.5 = 1,136,632. The batch needed **6,323,632** and consumed 749,930 of it.

## Fix

`buildFinalizationBatches()` estimates each call alone, sums, and packs the calls under a ceiling below EIP-7825's 2^24 = 16,777,216 per-transaction cap. Submission is unchanged — still `tryAggregate(false, ...)`, so a call that starts reverting after the pre-flight can't take the rest of the batch with it.

For the same batch the calls estimate at 5,830,076 / 674,835 / 5,823,461 = **12,328,372** against a requirement of 6,323,632. Loose by 1.95x, and that looseness is a *limit*, not a spend — the batch is still charged the 749,930 it consumes.

Two properties this has that an `aggregate()`-derived size did not: nothing can fail to estimate as a whole, and a per-call rejection names the finalization that stopped being valid.

## The sum alone is short

Multicall3 forwards only 63/64 to each call, which for a batch of one costs more than the intrinsic gas a standalone estimate carries. Measured on the Optimism withdrawal: it estimates at **5,830,076** and needs **5,866,277** in a batch. `MULTICALL3_BATCH_GAS_MULTIPLIER` (1.1) covers it, and a test pins that the bare sum fails where the padded one succeeds, so it can't be quietly dropped.

Each batch is enqueued `nonMulticall`; without it `MultiCallerClient` merges them back into a `multicall(bytes[])` Multicall3 doesn't expose, which throws on encode and abandons the chain's whole batch. (Found by @codex on #3667.)

## Tests

`test/Finalizer.BatchBuilding.test.ts` — four cases against real Multicall3 bytecode, with `MockGasBurner.requireGas(minGas)` modelling `callWithMinGas`: a call that reverts unless *given* `minGas` while consuming almost none of it. They assert that padding the batch's own estimate by 1.5x still leaves that shape failing (pinning the shipped bug), that the sized limit lands every call, that the bare sum is short for a single call where the padded one isn't, and that an oversized batch splits with every padded limit under the ceiling and the calls partitioned in order.

`blockGasLimit` is held under the 2^24 cap in `hardhat.config.ts`: `eth_estimateGas` bisects up towards the block gas limit for any call gated on `gasleft()`, and a probe above the cap errors instead of estimating — which would silently size those calls at zero. This is also what made #3667's suite unrunnable.

## Deliberately not here

- **Partial-submission reporting.** `TransactionClient#submit` stops at the first failure, so a chain that enqueued several batches can land only some and still read as submitted. Splitting makes that reachable, but it was already inaccurate and the fix belongs with the bigger gap below.
- **Reading `tryAggregate`'s `success[]`.** Nothing does, so a batch that mines having done nothing is still reported as finalized — the reason this ran ~10 times unnoticed. One `eth_call` at the sized limit would isolate the failing calls *and* validate the limit: on the batch above it returns `[false, true, false]` at the 1,136,632 that shipped and `[true, true, true]` at 6,323,632. Worth doing properly in one follow-up.


src/finalizer/index.ts
```diff
@@ -396,23 +477,35 @@ export async function finalize(
...
-            multicallerClient.enqueueTransaction(txnToSubmit);
+            const { batches, dropped: unestimated } = await buildFinalizationBatches(
+              logger,
+              Number(chainId),
+              multisender,
+              multicallTxns
+            );
+            unestimated.forEach((crossChainMessage) => droppedMessages.add(crossChainMessage));
+            batches.forEach(({ calls, gasLimit }) =>
+              multicallerClient.enqueueTransaction({
```
pxrl, line 488: We also want to enforce awaiting confirmation on these.

pxrl (commented):
Just one comment.

src/finalizer/index.ts
droplet-rl: Done — `ensureConfirmation: true` on the batch transactions, in "fix(finalizer): await confirmation on finalization batches".

Worth noting it does more here than await a receipt: it also closes a reporting hole this PR opened. A batch that reverts outright otherwise comes back as a hash like any other, and `submissionStatus()` counts hashes — so the chain's expected count would be met and every message on it would read as finalized. With `ensureConfirmation`, the revert surfaces as a `CALL_EXCEPTION`, `TransactionClient#submit` stops there and returns the hashes it already has, and that short list is exactly what `submissionStatus()` reads as *"only n/m transactions submitted; this message may not be in one of them"*.

The other benefit is sequencing. A chain can now enqueue more than one batch, so each waiting for its receipt means a stuck early nonce gets repriced at that nonce rather than leaving the later batches queued behind it.

Cost is wall-clock: batches submit sequentially and each awaits a receipt (6 s sampling, 24 s on mainnet), so a chain with several batches takes proportionally longer per run. Seems clearly the right trade for finalizations, but flagging it in case you disagree.

One thing I did **not** change: the `AugmentedTransaction` finalizations enqueued directly just above this (`txn.nonMulticall = true; multicallerClient.enqueueTransaction(txn)`). Those come from the individual chain finalizers with their own config, and your comment was anchored on the batches — happy to set it there too if you meant both.

pxrl (approved):

PR #3676, opened 2026-08-06
fix(inventory): treat Paxos Transit routes as unmetered fast rebalance

droplet-rl:
## Why

Robinhood USDG has no PoolRebalanceRoute, so `depositForcesOriginChainRepayment()` is true for every RH deposit — repayment can only be taken on Robinhood. Inventory there therefore only ratchets up, with no HubPool path to move it off. Once the allocation crosses its ceiling, `determineRefundChainId()` finds no eligible repayment chain and returns `[]`, which the relayer books as an unprofitable fill and skips.

That is what happened to deposit `36295` (100,000.67 USDG, Robinhood → mainnet, `0x8d6903238bc813525a2c7003a66a9990a641d912bd9aacfdccee2a698e8b435f`). From `zion-across-relayer-primary`:

```
15:11:35 InventoryClient  Evaluated taking repayment on origin chain 4663 for deposit 36295: OVERALLOCATED ❌
15:11:35 Relayer::resolveRepaymentChain  Deposit 36295 forces origin chain repayment and has an over-allocated origin chain Robinhood
15:11:35 ProfitClient     Handling unprofitable fill
```

`rankedChains: [4663]`, `expectedPostRelayAllocation` 41.73% against `targetPct` 0.1 × `targetOverage` 3.5 = 35%. Six RH fills totalling about 460,042 USDG between 13:33 and 14:48 had pushed it over. The deposit was re-evaluated about ten times and rejected identically each time until its destination-swap deadline expired. The fill itself was valid and paid a $70 fee against under $1 of gas — we refused the repayment, not the fill.

## What

`isUnmeteredFastRebalance()` already exempts force-origin deposits from the allocation check when the origin can be drained quickly. It only knew about CCTP, OFT and the hub chain. Paxos Transit qualifies on the same terms:

- **Fast.** Measured from `evm.paxos_transit_order_{submitted,executed}` over 7 days: 4663→1 is n=11, p50 430s, p90 590s, max 686s. Faster than CCTP, and far from the ~12h HyperEVM OFT case the function explicitly excludes.
- **Effectively unmetered at our size.** Paxos caps orders at `TRANSIT_MAX_USD` = $50mm; the largest order already settled on this route is 567,066 USDG — more than the entire accumulation that tipped us over.
- **Already wired.** `CUSTOM_BRIDGE[ROBINHOOD][USDC-mainnet] = PaxosTransitBridge` and `CUSTOM_L2_BRIDGE[ROBINHOOD][USDC-mainnet] = PaxosTransitL2Bridge`. No new adapter.

The check is derived from the existing `PAXOS_TRANSIT_DESTINATION_TOKENS` registry via `getPaxosTransitOfferAssetsForWantAsset()` rather than hardcoding `(ROBINHOOD, USDG)`, so new Paxos routes are picked up automatically. Both modules live in `src/utils`, so no new cross-layer dependency.

Gated on `PAXOS_API_KEY`, mirroring how `hasBinanceRoute()` gates on `binanceCredentialsConfigured()` — `createPaxosTransitClient()` throws without it, so there would be no route to claim.

## Two things for the reviewer

1. **Does the relayer process carry `PAXOS_API_KEY`?** If only the refiller/rebalancer has it, this change is a no-op in the relayer and the gate needs to move (or the key needs adding). I could not verify the deployed env from here.
2. **`ProfitClient` blast radius.** `isUnmeteredFastRebalance()` also backs the `RELAYER_POLICIES` origin default (`src/clients/ProfitClient.ts`), so RH/USDG will now match policy routes that leave `RELAYER_POLICY_<NAME>_ORIGINS_*` unset. That reads as consistent with treating RH like a CCTP origin, but it is a behaviour change beyond the inventory path. README updated accordingly.

## Note on the trade-off

This removes the only backstop on Robinhood inventory — the veto exists precisely because force-origin repayment on a chain you cannot drain is a trap. Safety now rests entirely on the rebalancer actually issuing the withdrawals, so an alert on RH USDG balance/age is worth adding separately.

The structural alternative is to give USDG@4663 a PoolRebalanceRoute, which kills `forceOriginRepayment` at the root and needs no special-casing at all. Heavier (HubPool governance) but cleaner, and it would fix this for third-party relayers running this same code — deposit 36295 went unfilled market-wide, not just by us.

## Testing

- `test/FillUtils.isUnmeteredFastRebalance.ts` — new, 3 cases: route recognised with credentials, fails closed without them, does not leak to non-Paxos tokens on the same chain.
- `yarn typecheck` clean; `test/InventoryClient.RefundChain.ts` + `test/ProfitClient.ConsiderProfitability.ts` pass (57 passing).

### DIFF
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1733,7 +1733,8 @@ export class InventoryClient {
     );
   }
 
-  // True for unmetered fast-rebalance routes (CCTP, OFT, hub). Binance is excluded — it's quota-gated.
+  // True for unmetered fast-rebalance routes (CCTP, OFT, Paxos Transit, hub). Binance is excluded —
+  // it's quota-gated.
   private isUnmeteredFastRebalance(repaymentChainId: number, repaymentToken: Address): boolean {
     return isUnmeteredFastRebalance(repaymentChainId, repaymentToken, this.hubPoolClient.chainId);
   }
--- a/src/clients/InventoryClient.ts
+++ b/src/clients/InventoryClient.ts
@@ -1733,8 +1733,7 @@ export class InventoryClient {
     );
   }
 
-  // True for unmetered fast-rebalance routes (CCTP, OFT, Paxos Transit, hub). Binance is excluded —
-  // it's quota-gated.
+  // True for unmetered fast-rebalance routes (CCTP, OFT, hub). Binance is excluded — it's quota-gated.
   private isUnmeteredFastRebalance(repaymentChainId: number, repaymentToken: Address): boolean {
     return isUnmeteredFastRebalance(repaymentChainId, repaymentToken, this.hubPoolClient.chainId);
   }
--- a/src/utils/FillUtils.ts
+++ b/src/utils/FillUtils.ts
@@ -2,7 +2,8 @@ import { utils as sdkUtils } from "@across-protocol/sdk";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
 import { HubPoolClient, SpokePoolClient } from "../clients";
 import { FillStatus, FillWithBlock, SpokePoolClientsByChain, DepositWithBlock, RelayData } from "../interfaces";
-import { Address, compareAddressesSimple, EMPTY_MESSAGE } from "../utils";
+import { Address, compareAddressesSimple, EMPTY_MESSAGE, isDefined } from "../utils";
+import { getPaxosTransitOfferAssetsForWantAsset } from "./PaxosTransitUtils";
 
 export type RelayerUnfilledDeposit = {
   deposit: DepositWithBlock;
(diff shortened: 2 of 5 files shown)


src/utils/FillUtils.ts
```diff
@@ -84,7 +85,13 @@ export function isUnmeteredFastRebalance(chainId: number, token: Address, hubCha
     sdkUtils.chainIsOFTEnabled(chainId) &&
     compareAddressesSimple(TOKEN_SYMBOLS_MAP.USDT.addresses[chainId], token.toNative()) &&
     chainId !== CHAIN_IDs.HYPEREVM;
-  return cctp || oft || chainId === hubChainId;
+  // Paxos Transit withdrawals settle in ~7-11 min and cap at $50mm/order, so they're fast and
+  // effectively unmetered at relayer size. Gated on credentials, since createPaxosTransitClient()
+  // requires PAXOS_API_KEY — without it there is no route to claim.
+  const paxosTransit =
+    getPaxosTransitOfferAssetsForWantAsset(chainId, token.toNative()).length > 0 &&
+    isDefined(process.env.PAXOS_API_KEY);
```
pxrl: @droplet-rl I think we probably drop this requirement.

droplet-rl: Dropped. You are right that it did not belong — the two branches above it (CCTP, OFT) are pure over the static registries, and the docstring claims the whole function is, so reading `process.env` here broke that for no real gain. If a relayer is configured to fill USDG routes it has the key; if it does not, the fill fails on the missing route rather than on a stale allocation check.

Also refreshed the `InventoryClient.isUnmeteredFastRebalance()` comment, which still listed only CCTP/OFT/hub.

src/clients/InventoryClient.ts
```diff
@@ -1733,7 +1733,8 @@ export class InventoryClient {
     );
   }
 
-  // True for unmetered fast-rebalance routes (CCTP, OFT, hub). Binance is excluded — it's quota-gated.
+  // True for unmetered fast-rebalance routes (CCTP, OFT, Paxos Transit, hub). Binance is excluded —
+  // it's quota-gated.
```
pxrl: @droplet-rl Revert this change.

droplet-rl: Reverted. `InventoryClient.ts` now matches master exactly — that comment refresh was unrelated to the fix and I should not have folded it in.

pxrl (dismissed):

pxrl (approved):

PR #3677, opened 2026-08-06
feat(l2Bridges): add ZK Stack L2 -> L1 withdrawal support

droplet-rl:
Requested by @paul in [Slack](https://risklabs.slack.com/archives/C0BHMM63D9Q). Follow-up to the USDT allocation review, where zkSync turned out to be one of four chains the relayer can bridge *into* but never *out of*.

## Problem

zkSync Era (324) and Lens (232) have an L1→L2 bridge but no L2→L1 route for any token. `CUSTOM_L2_BRIDGE[ZK_SYNC]` held only a Binance route for USDC and `CANONICAL_L2_BRIDGE` had no ZK Stack entry at all, so `InventoryClient.withdrawExcessBalances` short-circuited at the `isSupportedL2Bridge` check (`InventoryClient.ts:1330-1342`) and skipped them **silently** — no log. Inventory accumulated permanently.

Current stranded balances for `0x07aE…0E67`:

| chain | token | balance | route today |
|---|---|---|---|
| zkSync | USDT | 48,932.17 | none |
| zkSync | WETH | 38.72 | none |
| Lens | ETH (as ERC20) | 22.95 | none |

The zkSync USDT position has a standard deviation of **0.00** across 289 inventory samples over 24h — it has not moved a cent.

## Approach

Two adapters, because the stack has two genuinely different exits.

**`ZKStackBridge`** — ERC20s via the L2 asset router (`0x…010003`), using `withdraw(bytes32 assetId, bytes assetData)`. This is chosen over the legacy `withdraw(address,address,uint256)` for two reasons:

1. the legacy entrypoint reverts `TokenNotLegacy()` for L2-origin tokens;
2. the two paths emit **different L2→L1 messages from different senders**. The legacy path sends via the legacy shared bridge; the assetId path sends via the asset router. `useLegacyFinalizeWithdrawal` (`finalizer/utils/zkSync.ts:231-234`) keys off exactly that sender to pick `finalizeDeposit` (which passes `l2Sender` explicitly) over `finalizeWithdrawal` (which has to assume it). Using one L2 path keeps one correct L1 path.

`assetId` is resolved at runtime from the native token vault. Two guards return no transaction: an unregistered token (`assetId == 0`), and the chain's own `WETH_TOKEN`, which the vault refuses to burn with `BurningNativeWETHNotSupported`. The second guard is separate because a token can be registered and still be unburnable.

**`ZKStackNativeBridge`** — unwraps the wrapped native token and withdraws the base token via `L2BaseToken.withdraw` (`0x…800A`), which is the only way out for zkSync WETH. Funds arrive on L1 unwrapped, which is fine on zkSync because `MAINNET` is in `AdapterManager.chainsToWrapEtherOn`.

**Finalizer.** `zkSyncFinalizer` previously read only `spokePoolClient.getTokensBridged()`, and its signature took 4 of the 6 `ChainFinalizer` params so `addressesToFinalize` was silently discarded. Rebalancer withdrawals would therefore have initiated on L2 and **never finalized**. It now also discovers `BridgeBurn` on the L2 native token vault and `Withdrawal` on L2BaseToken, filtered by sender — the same pattern `opStack.ts:218-231` and `linea/l2ToL1.ts:232-245` already use. Unknown tokens are skipped rather than thrown, since one would otherwise abort finalization for the whole chain.

## Deliberately not wired

- **Lens WGHO** (11,719.41). Withdrawing it delivers unwrapped **L1 GHO** `0x40D16FC0…`, but the inventory book is keyed on L1 WGHO/LGHO `0x1ff1dC3c…` and there is no GHO→LGHO equivalent of the ETH wrap. Needs that gap closed first — then it is a one-line wiring change.
- **Lens USDC** (3,130.37). Uses the standalone `L2USDCBridge` with approval to the bridge itself; needs its own adapter. The L1 finalization side already handles it via `withdrawalRequiresCustomUsdcBridge`.
- **zkSync USDC** stays on `L2BinanceCEXBridge`. `hasBinanceRoute` compares constructors by reference, so replacing it would flip fast-rebalance eligibility.

⚠️ Related, and worth flagging separately: `BINANCE_NETWORKS` lists `BASE` and `ZKSYNCERA`, but Binance's live coin list shows **USDT is not offered on either network**. Since `hasBinanceRoute` is a purely static check, wiring `L2BinanceCEXBridge` for USDT on those chains would compile, pass every guard, and send funds to an uncredited deposit address. Not touched here.

## Verification

Beyond unit tests, I built each adapter's **real calldata** and `eth_call`-simulated it from the relayer address against live chain state:

| case | result |
|---|---|
| zkSync USDT via asset router | ✅ simulation OK |
| zkSync WBTC via asset router | ✅ simulation OK |
| zkSync WETH via asset router | ✅ correctly guarded, 0 txns |
| zkSync WETH via unwrap + base token | ✅ both txns simulate OK |
| Lens ETH via asset router | ✅ simulation OK |
| Lens WGHO via asset router | ✅ correctly guarded, 0 txns |

The `assetId` values and the exact `(uint256,address,address)` `assetData` encoding were confirmed against the vault on both chains and against a production withdrawal on zkSync Era (`0x038b81dd…`).

Tests: `yarn test` gives **950 passing** vs **942** on `master` (+8, the new file). The same 3 failures occur on both — `Dataworker.loadData.deposit` (order-dependent; passes in isolation), `disputer.watchdog`, and `finalizeCCTPV1Messages` (missing local `solana-test-validator`). Typecheck and lint clean.

## Enabling it

This adds the capability only. Each token still needs `withdrawExcessPeriod` in the inventory config before the sweep will run, and the relayer address must be in `FINALIZER_WITHDRAWAL_TO_ADDRESSES` or withdrawals will initiate and never finalize.


src/common/Constants.ts
```diff
@@ -467,7 +469,9 @@ export const CANONICAL_BRIDGE = resolveCanonicalBridges();
 
 export const CANONICAL_L2_BRIDGE: Record<number, L2BridgeConstructor<BaseL2BridgeAdapter>> = {
   [CHAIN_IDs.BSC]: L2BinanceCEXBridge,
+  [CHAIN_IDs.LENS]: L2ZKStackBridge,
```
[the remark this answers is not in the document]
droplet-rl: Right conclusion, and fixed in 749390c — though the mechanism is a step earlier than described.

Lens USDC never reached the asset router: `NativeTokenVault.assetId(0x88F0…f884)` on Lens returns `0x00…00` (it arrives over the standalone bridge and was never registered), so `ZKStackBridge.constructWithdrawToL1Txns` hits the `assetId === ZERO_BYTES` guard and returns no transaction. Nothing gets stuck on L1 because nothing is withdrawn. Same for WGHO.

But you are right that the config was too broad, and the actual damage is worse than a latent trap. `CANONICAL_L2_BRIDGE[LENS]` covered all three of Lens's supported tokens, and `src/jussi/topology/edges.ts:60` creates an outbound edge for any `(chain, token)` with a defined constructor, while `quoteNativeBridgeFeeUsd` prices an empty transaction list at zero. So Lens USDC and WGHO were advertised to the rebalancer as *free* withdrawal routes — maximally attractive to a cost-minimising planner — that silently execute nothing.

Holding them back on a live `assetId` lookup rather than on config also means your scenario becomes reachable the moment either token is registered in the vault, since `withdrawalRequiresCustomUsdcBridge()` classifies on `(l1Chain, l2Chain, l2Token)` alone.

Fix is to make the exclusion static: Lens moves out of `CANONICAL_L2_BRIDGE` and into `CUSTOM_L2_BRIDGE` with a WETH-only entry, so USDC and WGHO resolve to no bridge at all rather than to one that constructs nothing. zkSync keeps its canonical entry — all five of its supported tokens are intentionally routed (I checked DAI's assetId on-chain too, alongside the USDT/WBTC verified earlier). Added a config test pinning the resolved bridge for every supported token on both chains so a future `SUPPORTED_TOKENS` addition cannot silently inherit a route.

src/adapter/l2Bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,193 @@
...
+  winston,
+  ZERO_BYTES,
+} from "../../utils";
+import { BaseL2BridgeAdapter } from "./BaseL2BridgeAdapter";
+import { AugmentedTransaction } from "../../clients/TransactionClient";
+
+// A ZK Stack withdrawal is only claimable on L1 once the batch containing it has been executed. The finalizer
+// waits 6 hours before it even attempts to finalize (see src/finalizer/utils/zkSync.ts), so allow generous
+// headroom on top of that so that in-flight withdrawals are not double-counted as spendable inventory.
+const ZK_STACK_WITHDRAWAL_LOOKBACK_SECONDS = 24 * 60 * 60;
```
pxrl: @droplet-rl This is no longer true - a change has just been merged. Merge it in and update this accordingly; might make sense to refactor for a more consistent implementation.

src/adapter/l2Bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * @dev One bridge instance is constructed per l1Token, so both of these are constant for the lifetime of the
+   * instance and are memoised to keep withdrawal construction cheap on the hot path.
+   */
+  protected resolveAssetId(l2Token: EvmAddress): Promise<string> {
+    return (this.assetId ??= this.getL2Bridge().assetId(l2Token.toNative()));
+  }
+
+  protected resolveNativeWrapper(): Promise<string> {
+    return (this.nativeWrapper ??= this.getL2Bridge().WETH_TOKEN());
```
pxrl: Is this also valid for Lens?

pxrl: Could the resolver also return an Address class, for easier comparisons later? i.e. instead of using compareAddressSimple().

src/adapter/l2Bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,193 @@
...
+      });
+      return [];
+    }
+
+    // NativeTokenVault.decodeBridgeBurnData expects exactly (uint256 amount, address l1Receiver, address l2Token);
+    // anything else reverts with InvalidNTVBurnData().
+    const assetData = ethers.utils.defaultAbiCoder.encode(
+      ["uint256", "address", "address"],
+      [amount, toAddress.toNative(), l2Token.toNative()]
+    );
```
pxrl, line 130: Worth saving this in src/common/abi ?

src/adapter/l2Bridges/ZKStackBridge.ts
droplet-rl: Good catch, and the answer is more interesting than I expected.

The call is valid on Lens — same system contract, same address — but it does not return what you might assume. `NativeTokenVault.WETH_TOKEN()` is the chain's wrapped *base* token, so on Lens it resolves to `0xB66F040B005Fad43b645794d2c53F51a74D327eC`, symbol `WLGHO`, `totalSupply() == 0`. That is a different contract from the WGHO Across tracks (`0x6bDc36E20D267Ff0dd6097799f82e78907105e2F`, the `wrappedNativeToken` entry in ContractAddresses).

So the guard never shielded Lens WGHO. What actually stopped it was the other branch — `assetId(0x6bDc…) == 0x00…00`, because WGHO was never registered in the vault — which is live chain state rather than a decision. As of 749390c that exclusion is static: Lens is wired per-token in `CUSTOM_L2_BRIDGE` with a WETH-only entry, so WGHO and USDC resolve to no bridge at all.

I have added a `@dev` block on the resolver spelling this out, so nobody later reads the `WETH_TOKEN` check as "the wrapped native token Across holds" — on zkSync Era those coincide, on Lens they do not.

droplet-rl: Done in 2f62f16 — `resolveNativeWrapper()` now returns `Promise<EvmAddress>` and the guard reads `l2Token.eq(nativeWrapper)`.

Left `resolveAssetId()` returning a raw `string`: it is a bytes32 assetId, not an address, so there is no Address class to hand back.

Also left the `compareAddressesSimple()` in `getL2PendingWithdrawalAmount()` alone — that one compares a raw `l1Args.receiver` off an event, per reduce iteration, and matches how the sibling adapters handle event args. Happy to convert it too if you would rather the whole file be uniform.

src/adapter/l2Bridges/ZKStackBridge.ts
droplet-rl: I would lean towards leaving this inline, but tell me if you disagree and I will move it.

`src/common/abi` is all contract ABI JSON artifacts, loaded through `getContractEntry()`. This is not a contract interface — it is the payload shape `NativeTokenVault.decodeBridgeBurnData()` expects, encoded ad hoc for a single call site, and there is no existing precedent for an encoding shape living in that directory.

It also matches the deposit-leg counterpart directly: `src/adapter/bridges/ZKStackBridge.ts:264` builds its bridgehub second-bridge calldata the same way, with an inline literal tuple. There are ~8 similar sites (opStack, helios, hyperliquid, arbitrum).

If the concern is that the tuple is a magic literal, the comment above it names the exact function it has to match and the error it reverts with when wrong (`InvalidNTVBurnData()`), which is what I would want to see when debugging it. But if you would prefer a named export I will add one.

src/adapter/l2Bridges/ZKStackBridge.ts
droplet-rl: Merged master in (45a16a3) and fixed this in 2f62f16.

You're right that the rationale was stale — #3659 replaced the fixed 6h lookback with the executed L1 batch boundary read from chain, so "the finalizer waits 6 hours" is no longer a thing to hang a number on. The merge also needed a manual resolution: this branch's EOA withdrawal discovery hangs off the same `latestBlockToFinalize`, so it now inherits the dynamic boundary rather than a fixed offset, and `getRedisCache` is no longer needed in that file.

On the refactor: I looked at making the adapter follow the executed-batch boundary the way the finalizer now does, and decided against it. `pendingWithdrawalLookbackPeriodSeconds()` is a synchronous seconds value consumed by `BaseChainAdapter`, so tracking it dynamically means making that interface async across all nine implementations. More to the point, precision isn't what this window wants — it needs *slack*, not accuracy. Over-counting pending withdrawals only suppresses a duplicate withdrawal; under-counting reports the same inventory as spendable on both sides at once. So 24h over an observed 3-4h execution is deliberate headroom that also covers the finalizer being down for a few cycles, and I've rewritten the comment to say that instead of citing a delay that no longer exists.

What I did take from "more consistent": the constant was declared twice, once in each ZK Stack adapter, with `ZKStackNativeBridge` deferring to a comment in `ZKStackBridge` that was itself wrong. It's now a single exported constant with one rationale.

Replied to your three other comments on their own threads — the Lens one is worth a look, `WETH_TOKEN()` there resolves to WLGHO rather than the WGHO we track, so that guard never did what its name suggests.

Typecheck, lint, and the adapter + zkSync finalizer suites pass.

src/common/Constants.ts
```diff
@@ -722,6 +740,10 @@ export const CUSTOM_L2_BRIDGE: Record<number, Record<string, L2BridgeConstructor
   },
   [CHAIN_IDs.ZK_SYNC]: {
```
pxrl, line 741: Why are USDT withdrawals not supported on zkSync?

droplet-rl: They are supported — zkSync USDT was the headline motivation for the PR (48.9k USDT stranded on L2 with no exit).

`CUSTOM_L2_BRIDGE` is an exceptions list, not a whitelist. Resolution is `CUSTOM_L2_BRIDGE[chainId]?.[l1Token] ?? CANONICAL_L2_BRIDGE[chainId]`, so for zkSync:

| token | resolves to | why |
|---|---|---|
| USDT | `L2ZKStackBridge` (canonical) | asset router; `assetId` is registered (`0xdde1b8f8…`) |
| WBTC | `L2ZKStackBridge` (canonical) | same |
| USDC | `L2BinanceCEXBridge` | pre-existing, left alone so `hasBinanceRoute()` is unchanged |
| WETH | `L2ZKStackNativeBridge` | vault refuses to burn its own `WETH_TOKEN`, so it unwraps and exits as ETH |

Only the two that deviate need an entry. This mirrors the deposit leg, where `CANONICAL_BRIDGE[ZK_SYNC] = ZKStackBridge` and `CUSTOM_BRIDGE[ZK_SYNC]` likewise only overrides WETH.

Since the block reads like a whitelist at a glance, the config test pins the resolved bridge for all four supported tokens explicitly (`test/generic-l2-adapters/ZKStack.ts`, "resolves the intended L2 bridge for every supported token"), including `USDT: ZKStackBridge`. It also asserts the pinned set equals `SUPPORTED_TOKENS[ZK_SYNC]`, so a token added there cannot silently inherit the canonical route unreviewed — that assertion is what caught DAI in 339472e.

src/finalizer/utils/zkSync.ts
```diff
@@ -37,14 +39,145 @@ const IGNORED_WITHDRAWALS = [
...
+    return [
+      {
+        ...rest,
+        amountToReturn,
+        chainId: l2ChainId,
+        leafId: 0,
+        l2TokenAddress: EvmAddress.from(l2TokenAddress),
+        txnRef: transactionHash,
+        txnIndex: transactionIndex,
+      } as TokensBridged,
```
pxrl: Is this type assertion needed?

src/finalizer/utils/zkSync.ts
```diff
@@ -37,14 +39,145 @@ const IGNORED_WITHDRAWALS = [
...
+  senderAddresses: AddressesToFinalize
+): Promise<TokensBridged[]> {
+  assert(isEVMSpokePoolClient(spokePoolClient));
+  const { chainId: l2ChainId } = spokePoolClient;
+  const { chainId: l1ChainId } = hubPoolClient;
+
+  // Withdrawals sent by the SpokePool are already covered by getTokensBridged().
+  const senders = Array.from(senderAddresses.keys())
+    .filter((address) => address.isEVM())
+    .map((address) => address.toEvmAddress())
```
pxrl:
```suggestion
    .map((address) => address.toNative())
```

src/common/ContractAddresses.ts
pxrl, line 1: Given the contract addresses here basically the same for each chain, how about we set these as defaults for elastic chains?

src/adapter/l2Bridges/ZKStackBridge.ts
```diff
@@ -0,0 +1,211 @@
...
+ * observed to take 3-4 hours and drifts with the chain's batch cadence. The finalizer no longer assumes a fixed
+ * delay — it reads the executed-batch boundary from chain (see src/finalizer/utils/zkSync.ts) — so there is no
+ * constant to mirror here. This window does not need to track that boundary precisely, it only has to be long
+ * enough that a burn which has not yet been minted on L1 is still visible, and erring long is the safe direction:
+ * over-counting pending withdrawals suppresses a duplicate withdrawal, whereas under-counting would report bridged
+ * inventory as spendable on both sides at once. A day gives the finalizer room to be down for a few cycles.
+ *
+ * @dev Shared with ZKStackNativeBridge; the same batch-execution delay governs base token withdrawals.
+ */
+export const ZK_STACK_WITHDRAWAL_LOOKBACK_SECONDS = 24 * 60 * 60;
```
pxrl, line 34: Why is this a constant? Might want to make it configurable?

src/finalizer/utils/zkSync.ts
droplet-rl: No — and it was worse than redundant. Removed in e5ea02f.

`getEOAWithdrawals()` is annotated `Promise<TokensBridged[]>`, so tsc already checks the literal structurally. The assertion was widening it first and suppressing that check.

Verified both directions rather than just deleting it and seeing green:

- Assertion removed, otherwise untouched → `yarn typecheck` passes, so nothing depended on it.
- Assertion removed **and** `leafId: 0` deleted → tsc fails with `Property 'leafId' is missing in type ... but required in type 'TokensBridged'`.

That second one is the point. With the assertion in place, the same omission compiles clean. This object is hand-assembled field by field out of raw event args — `blockNumber`/`logIndex` ride in on `...rest`, the rest are written out individually — so a missing or misnamed field is precisely the mistake worth catching, and the assertion was the one thing preventing it.

One caveat while I was in there: `amountToReturn` still infers as `any`, because it comes off `event.args`. So the check covers every field except that one. Not something I changed here, but worth knowing the guarantee is not total.

src/finalizer/utils/zkSync.ts
droplet-rl: Applied in 52eeb17a — behaviour-identical, and `toNative()` is what the rest of the adapter code uses.

`.filter((address) => address.isEVM())` narrows to `EvmAddress`, and `EvmAddress.toNative()` is defined as `return this.toEvmAddress()` (sdk `AddressUtils.ts`), so the two produce the same checksummed string here. Only the base `Address`/`SvmAddress`/`TvmAddress` implementations diverge, and the filter has already excluded those.

src/common/ContractAddresses.ts
droplet-rl: Done in 7cb34d73 — but as an explicit spread rather than a family-driven default, because `chainIsZkStack()` doesn't cover Era.

At the pinned `@across-protocol/constants` (3.1.123), `PUBLIC_NETWORKS[324].family` is `ChainFamily.NONE`; only Lens (232) and Lens Sepolia (37111) are `ZK_STACK`. Checked directly: `chainIsZkStack(324) === false`. So auto-populating over `Object.values(CHAIN_IDs).filter(chainIsZkStack)` — the `MIN_DEPOSIT_CONFIRMATIONS` / `DEFAULT_GAS_FEE_SCALERS` pattern in `Constants.ts` — would skip zkSync Era, and it would fail quietly: the finalizer reads a missing `nativeTokenVault`/`l2BaseToken` entry as "not a ZK Stack deployment" and skips EOA withdrawal discovery rather than throwing. Silent, on the one chain this PR exists for.

What's there instead: `ZK_STACK_SYSTEM_CONTRACTS` (`nativeTokenVault`, `assetRouter`, `l2BaseToken` — all three are protocol predeploys at fixed addresses) defined once and spread into Era, Lens and Lens Sepolia. Chain-specific entries (`nativeToken`, `weth`, `wrappedNativeToken`, Lens `usdcBridge`) stay inline. I diffed the resolved `CONTRACT_ADDRESSES` for 324/232/37111 before and after: identical, modulo key order.

Happy to move it to `chainIsZkStack()` once constants marks Era as `ZK_STACK`. That looks like an upstream bug worth fixing regardless — want me to open the PR against `across-protocol/constants`?

src/adapter/l2Bridges/ZKStackBridge.ts
droplet-rl: Left as a constant for now, but say the word and I'll add the knob.

Precedent is that these aren't configurable: `OpStackBridge` hardcodes 7d + 1h, `BinanceCEXBridge` 1h, `BaseL2BridgeAdapter` defaults to 7200s. `ArbitrumOrbitBridge` is the only one that derives it, per chain via `getArbitrumOrbitFinalizationTime()`. None read env.

It also isn't a value that needs to be right, only long enough. It bounds the window over which we count burned-but-not-yet-minted withdrawals, and the error is asymmetric: too short reports bridged inventory as spendable on both sides at once; too long only suppresses a duplicate withdrawal. 24h against an observed 3-4h batch execution is deliberate slack, not a tuned number — it exists so the finalizer can be down for a few cycles without the rebalancer double-spending.

The cost of the slack is RPC: it widens the paginated event query on both L1 and L2 in `getL2PendingWithdrawalAmount()`. If that's the concern, either shape works — a per-chain map in `Constants.ts` mirroring `getArbitrumOrbitFinalizationTime()` (Era and Lens do have different batch cadences), or an env override in the `RELAYER_OFT_MIN_WITHDRAWAL_PCT` style from `OFTL2Bridge`. Tell me which and I'll push it.

pxrl (approved):

PR #3682, opened 2026-08-06
refactor(paxos): carry Paxos Transit token routes as Address, not string

droplet-rl:
Follow-up to #3676, requested by @pxrl in Slack. **Stacked on #3676** — base is that branch, so the diff here is just the refactor. #3676 has since squash-merged as c62814933fbf310ee334f673bfc9eacb712fcd66, so retargeting to `master` now needs a rebase to drop those commits — say the word and I'll do it.

## Why

`getPaxosTransitOfferAssetsForWantAsset()` took a raw `string` and compared it case-insensitively against `PAXOS_TRANSIT_DESTINATION_TOKENS`, so callers holding an `Address` had to unwrap first — #3676 added `getPaxosTransitOfferAssetsForWantAsset(chainId, token.toNative())` in `isUnmeteredFastRebalance()`, which is what prompted this.

The end state, after @pxrl asked whether the pattern applied elsewhere: **no address-typed parameter or return value on the Paxos interface is a bare `string` any more.**

### Route resolution

- `getPaxosTransitOfferAssetsForWantAsset(dstChainId, wantAsset: Address): EvmAddress[]`. Offer assets are always keyed on mainnet, hence `EvmAddress` rather than `Address`. Matching is `Address.eq()` against `toAddressType(registryValue, dstChainId)` instead of `toLowerCase()` string equality.
- `getPaxosTransitDestinationToken()` returns `Address | undefined` for the same reason: its result is what feeds the function above, and both Paxos adapters were re-wrapping it with `toAddressType()` at each use.
- `paxosTransitOrderMatchesRoute()` takes `Address` for `wantAsset`/`receiver`, so the caller stops unwrapping two `Address`es to build `routeParams`.

### Address resolvers

`getMainnetUsdgAddress()` → `EvmAddress`; `getPaxosTransitStationAddress()` and `getPaxosTransitBoringVaultAddress()` → `Address`. These three were the only `get*Address()` helpers in `src/` returning a bare string — `getSpokePoolAddress`, `getHubPoolAddress`, `getNativeTokenAddressForChain` and `getWrappedNativeTokenAddress` all return `Address`/`EvmAddress`.

### Request surfaces

`listPaxosTransitOrders`, `listOutstandingPaxosTransitOrders`, `resolvePaxosTransitAuthorization`, `buildPaxosTransitSubmitOrderTxn`, `submitPaxosTransitApproval` and the three `PaxosTransitClient` request-param objects all take `Address`. The client serialises with `toNative()` into `URLSearchParams`, so strings are produced at the wire boundary and nowhere earlier.

### Adapters

- `PaxosTransitBridge.l2TokenAddress` and `PaxosTransitL2Bridge.l2TokenAddress` become `Address`, letting the three `l2Token` identity checks use `eq()` rather than `===` on `toNative()`.
- Both adapters wrap the signer address as `EvmAddress.from(await signer.getAddress())`, matching the eleven other sites in `src/` that do this.
- `PaxosTransitBridge` holds `l1Token`, so the L1-token assert uses `eq()` instead of comparing `toNative()` against the ethers `Contract` address.
- The `offerAsset` log field passes the `Address` directly — `Address.toJSON()` delegates to `toNative()`.

### Deliberately still `string`

- Response types (`PaxosTransitOrder`, `PaxosTransitAuthorizationResponse`, `PaxosTransitOrderQuoteResponse`, `PaxosTransitPermitData`) are deserialised JSON. `toAddressType()` throws on a malformed value where a string compare merely fails to match, so `paxosTransitOrderMatchesRoute` keeps using `compareAddressesSimple` on the order side.
- `PAXOS_TRANSIT_DESTINATION_TOKENS` is built from `TOKEN_SYMBOLS_MAP` literals; the conversion belongs in the accessors.
- Remaining `toNative()` calls are the real boundaries: ethers `Contract` constructors and event filters, `BridgeEvents` record keys (`{ [l2Token: string]: BridgeEvent[] }`), the `switch` on `token.toNative()` in `Refiller#refillBalances`, and the Paxos HTTP params.

## Behaviour

None intended. The comparison is equivalent for these routes: registry values and callers' tokens are both EVM addresses on the same chain, so `toAddressType()` yields `EvmAddress` on both sides and `eq()` compares checksummed forms where the old code compared lowercased ones.

## Testing

- `yarn typecheck`, `yarn lint` clean.
- `test/generic-adapters/*`, `test/generic-l2-adapters/*`, `test/Refiller.ts`, `test/FillUtils.isUnmeteredFastRebalance.ts` — 113 passing.
- `test/InventoryClient.RefundChain.ts`, `test/ProfitClient.ConsiderProfitability.ts` — 57 passing.
- `test/generic-adapters/PaxosTransitBridge.ts` updated for the new types, plus a case asserting a token with no Paxos route returns `[]`.

Note `tsconfig.json` scopes `typecheck` to `./src`, `index.ts` and `./scripts` — test files are only type-checked when the suite runs, so the suites above are what covers the test-side changes.

No `README.md`/`AGENTS.md` updates: the two docs that mention Paxos (`src/refiller/README.md`, `src/clients/README.md`) describe behaviour and env config, neither of which changes.

### DIFF
--- a/src/adapter/bridges/PaxosTransitBridge.ts
+++ b/src/adapter/bridges/PaxosTransitBridge.ts
@@ -35,6 +35,7 @@ import ERC20_ABI from "../../common/abi/MinimalERC20.json";
 export class PaxosTransitBridge extends BaseBridgeAdapter {
   protected client: PaxosTransitClient;
   protected l1TokenInfo: TokenInfo;
+  protected l1TokenAddress: EvmAddress;
   protected l2TokenAddress: Address;
 
   constructor(
@@ -59,6 +60,7 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     this.l2Bridge = new Contract(l2TokenAddress.toNative(), ERC20_ABI, l2SignerOrProvider);
     this.client = createPaxosTransitClient(logger);
     this.l1TokenInfo = getTokenInfo(l1Token, this.hubChainId);
+    this.l1TokenAddress = l1Token;
     this.l2TokenAddress = l2TokenAddress;
   }
 
@@ -70,18 +72,18 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     _optionalParams?: TransferTokenParams
   ): Promise<BridgeTransactionDetails> {
     assert(l2Token.eq(this.l2TokenAddress), `Attempting to bridge unsupported l2 token ${l2Token}`);
-    assert(l1Token.toNative() === this.l1Bridge?.address, "L1 token mismatch for Paxos Transit bridge");
+    assert(l1Token.eq(this.l1TokenAddress), "L1 token mismatch for Paxos Transit bridge");
 
     if (amount.lt(PAXOS_TRANSIT_MINIMUMS[this.hubChainId]?.[this.l2chainId] ?? bnZero)) {
       throw new Error(`Cannot bridge to ${getNetworkName(this.l2chainId)} due to invalid amount ${amount}`);
     }
 
-    const userAddress = await this.l1Signer.getAddress();
+    const userAddress = EvmAddress.from(await this.l1Signer.getAddress());
     const orderData = await buildPaxosTransitSubmitOrderTxn(this.client, this.l1Signer, {
       userAddress,
       offerAmount: amount,
-      offerAsset: l1Token.toNative(),
-      wantAsset: this.l2TokenAddress.toNative(),
+      offerAsset: l1Token,
+      wantAsset: this.l2TokenAddress,
       sourceChainId: this.hubChainId,
       destinationChainId: this.l2chainId,
       spenderAddress: getPaxosTransitStationAddress(this.hubChainId),
@@ -117,7 +119,7 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
         const tokenContract = new Contract(offerAsset.toNative(), ERC20_ABI, l1Provider);
         const events = await paginatedEventQuery(
           tokenContract,
-          tokenContract.filters.Transfer(fromAddress.toNative(), transitStation),
+          tokenContract.filters.Transfer(fromAddress.toNative(), transitStation.toNative()),
           eventConfig
         );
         return events.map((event) => processEvent(event, "value"));
@@ -141,7 +143,7 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     const tokenContract = new Contract(this.l2TokenAddress.toNative(), ERC20_ABI, l2Provider);
     const events = await paginatedEventQuery(
       tokenContract,
-      tokenContract.filters.Transfer(boringVault, toAddress.toNative()),
+      tokenContract.filters.Transfer(boringVault.toNative(), toAddress.toNative()),
       eventConfig
     );
     return {
@@ -153,7 +155,7 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     l1Token: EvmAddress,
     userAddress: EvmAddress
   ): Promise<BridgeEvents> {
-    const orders = await listOutstandingPaxosTransitOrders(this.client, userAddress.toNative());
+    const orders = await listOutstandingPaxosTransitOrders(this.client, userAddress);
     const l2TokenDecimals = getTokenInfo(this.l2TokenAddress, this.l2chainId).decimals;
     const l1TokenDecimals = this.l1TokenInfo.decimals;
     const routeParams = {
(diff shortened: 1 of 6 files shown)


src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -70,10 +69,7 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
     amount: BigNumber,
     _optionalParams?: TransferTokenParams
   ): Promise<BridgeTransactionDetails> {
-    assert(
-      l2Token.toNative() === this.l2TokenAddress,
-      `Attempting to bridge unsupported l2 token ${l2Token.toNative()}`
-    );
+    assert(l2Token.eq(this.l2TokenAddress), `Attempting to bridge unsupported l2 token ${l2Token.toNative()}`);
```
pxrl: .toNative() is redundant - it's implied.

src/adapter/l2Bridges/PaxosTransitL2Bridge.ts
```diff
@@ -63,7 +62,7 @@ export class PaxosTransitL2Bridge extends BaseL2BridgeAdapter {
     l1Token: EvmAddress,
     amount: BigNumber
   ): Promise<AugmentedTransaction[]> {
-    assert(l2Token.toNative() === this.l2TokenAddress, `Unsupported L2 token ${l2Token.toNative()}`);
+    assert(l2Token.eq(this.l2TokenAddress), `Unsupported L2 token ${l2Token.toNative()}`);
     assert(l1Token.eq(this.l1Token), `Unsupported L1 token ${l1Token.toNative()}`);
```
pxrl: .toNative() is redundant.

src/adapter/l2Bridges/PaxosTransitL2Bridge.ts
```diff
@@ -107,12 +106,12 @@ export class PaxosTransitL2Bridge extends BaseL2BridgeAdapter {
   ): Promise<BigNumber> {
     const { l2Signer } = this;
     assert(isDefined(l2Signer), "PaxosTransitL2Bridge: l2Signer is required");
-    assert(l2Token.toNative() === this.l2TokenAddress, `Unsupported L2 token ${l2Token.toNative()}`);
+    assert(l2Token.eq(this.l2TokenAddress), `Unsupported L2 token ${l2Token.toNative()}`);
```
pxrl: .toNative() redundant.

src/adapter/bridges/PaxosTransitBridge.ts
droplet-rl: Done — `${l2Token}` now. Confirmed `Address.toString()` delegates to `toNative()` (SDK `AddressUtils.ts:220`), so the subclass override still resolves correctly.

src/adapter/l2Bridges/PaxosTransitL2Bridge.ts
droplet-rl: Done. Also dropped it from the adjacent `l1Token` assert on the next line so the pair stays consistent.

src/adapter/l2Bridges/PaxosTransitL2Bridge.ts
droplet-rl: Done.

pxrl (approved):

src/adapter/bridges/PaxosTransitBridge.ts
```diff
@@ -142,29 +140,29 @@ export class PaxosTransitBridge extends BaseBridgeAdapter {
...
     );
     return {
-      [this.l2TokenAddress]: events.map((event) => processEvent(event, "value")),
+      [this.l2TokenAddress.toNative()]: events.map((event) => processEvent(event, "value")),
     };
   }
 
   private async buildOutstandingTransfersFromApiOrders(
     l1Token: EvmAddress,
     userAddress: EvmAddress
```
pxrl: s/userAddress/receiver ?

droplet-rl: Done in `a440796` — renamed to `receiver`, and `routeParams` now uses the shorthand.

Agreed it was the odd one out: within the helper the address is the order recipient (feeds `routeParams.receiver`, matched against `PaxosTransitOrder.receiver`). The signer-derived `userAddress` locals in `constructL1ToL2Txn`/`constructWithdrawToL1Txns` keep their name — those are the order *submitter* passed to `buildPaxosTransitSubmitOrderTxn`, a genuinely different role.

One alternative if you prefer it: `monitoredAddress`, which is what `BaseBridgeAdapter.ts:110`, `BaseChainAdapter.ts:595` and the public `getOutstandingTransfersFromApi` override all call this value — so the name would stay constant across the call boundary. I went with your `receiver` since it is more specific about the role inside the helper. Easy to flip.

dijanin-brat (approved):

pxrl (approved):

PR #3684, opened 2026-08-07
chore: Remove DAI routing mappings

droplet-rl:
## Summary

DAI is no longer a supported token. The routes API returns **0 DAI routes** across all 1,456 live routes, so this removes the relayer's DAI routing mappings and the code they orphan.

Checked on-chain before removing anything: every spoke is drained of DAI (Optimism, Arbitrum, Linea, Base, zkSync all zero; Polygon holds 0.03 dust), and the July 2026 recovery's DAI leaves (zkSync 70,005 / Base 104.1) have landed. Dropping the bridge adapters strands nothing.

## Removed

**`src/common/Constants.ts`**
- `SUPPORTED_TOKENS` — DAI on Arbitrum, Base, Linea, Optimism, Polygon, zkSync
- `CUSTOM_BRIDGE[OPTIMISM][DAI]` → `DaiOptimismBridge`
- `CUSTOM_ARBITRUM_GATEWAYS[ARBITRUM][DAI]`

**`DaiOptimismBridge`** — that `CUSTOM_BRIDGE` entry was its only non-test reference, so it and its dependencies are now dead:
- `src/adapter/bridges/DaiOptimismBridge.ts` + its `bridges/index.ts` export
- `src/common/abi/DaiOptimismBridgeL{1,2}.json`
- `CONTRACT_ADDRESSES.daiOptimismBridge` (mainnet L1 + Optimism L2)

**Tests**
- Dropped the `Custom bridge: DAI` suite in `test/generic-adapters/OpStack.ts`. It duplicated the adjacent `Default ERC20 bridge` suite — same `OpStackStandardBridge` mock, same assertions — so this is not a coverage loss.
- Dropped the per-chain DAI cases in `AdapterManager.SendTokensCrossChain.ts`. Each chain keeps equivalent coverage of the same bridge path via WBTC / USDC / WETH.
- Swapped the zkSync adapter test's generic-ERC20 fixture from DAI to WBTC, since it models `SUPPORTED_TOKENS[ZK_SYNC]`. Amounts there are decimal-agnostic, so the 18→8 decimal difference is inert.

**Comments** — `InventoryManagement.ts` config example DAI → WBTC; dropped the stale "DAI on Optimism" example in `opStack.ts`.

## Deliberately out of scope

- **Blast's DAI entries** (`SUPPORTED_TOKENS[BLAST]`, both `CUSTOM_BRIDGE` Blast keys, `BlastDaiRetriever`) — already removed by #3647, which drops Blast wholesale. Left untouched so the two PRs stay independently mergeable. They do conflict on one line: `SUPPORTED_TOKENS` `BASE` and `BLAST` are adjacent, so whichever lands second needs a one-line resolution (keep the DAI-free `BASE`, drop `BLAST`). I test-merged to confirm that is the only overlap between the two branches.
- **`TOKEN_SYMBOLS_MAP.DAI` in `@across-protocol/constants`** — must stay for now. The HubPool DAI pool is still enabled with 501,007 `liquidReserves` + 20,414 `utilizedReserves`, and relayer / sdk / quote-api / dapp all resolve DAI through that map, so removing it upstream first would break them at compile time. Usages first, constants last.
- **DAI as an abstract test fixture** (`RebalancerClient.*`, `RebalancerConfig`, `TokenUtils`, `ProfitClient.PriceRetrieval`) — these use DAI as an arbitrary token symbol rather than as route config.
- `scripts/recoveries/2026-07-accounting/*` — historical record, left as-is.

Downstream repos (`dapp` chain-configs, `quote-api`, `sdk` `STABLE_COIN_SYMBOLS`) also still carry DAI and are not touched here.

## Testing

- `yarn lint` clean, `yarn typecheck` clean
- Full suite: **948 passing**, 1 pending, 2 failing — against a `master` baseline of **950 passing, 2 failing**

Both failures are the same two on `master`, so they are pre-existing and unrelated:
- `finalizeCCTPV1Messages` — `spawn solana-test-validator ENOENT` (binary absent in this environment)
- `Dataworker: Load bundle data / Filters expired deposits` — cross-file suite pollution; passes in isolation and in its own Dataworker group, and references none of the symbols touched here

The 950 → 948 delta is exactly the two `it` blocks removed from the duplicated `Custom bridge: DAI` suite.

## Docs

No `AGENTS.md` / `README.md` updates needed — no doc references DAI or enumerates supported tokens.


src/finalizer/utils/opStack.ts
```diff
@@ -198,9 +198,11 @@ export async function opStackFinalizer(
   const minimumFinalizationTime = getCurrentTime() - 7 * 3600 * 24;
   const latestBlockToProve = await getBlockForTimestamp(logger, chainId, minimumFinalizationTime, undefined, redis);
 
-  // OP Stack chains have several tokens that do not go through the standard ERC20 withdrawal process (e.g. DAI
-  // on Optimism, USDC.e on Worldchain, etc) so the easiest way to query for these
-  // events is to use the TokenBridged event emitted by the Across SpokePool on every withdrawal.
+  // OP Stack chains have several tokens that do not go through the standard ERC20 withdrawal process
+  // (e.g. DAI on Optimism via the Maker bridge, USDC.e on Worldchain, etc) so the easiest way to query
```
pxrl: WorldChain migrated to native USDC via CCTP, so don't mention that here. DAI is also a dying token so I don't think it's worth flagging it.

droplet-rl: Good catch — dropped both examples and left the comment token-agnostic:

```
// OP Stack chains may have tokens that do not go through the standard ERC20 withdrawal process, so the
// easiest way to query for these events is to use the TokensBridged event emitted by the Across SpokePool
// on every withdrawal.
```

Also corrected `TokenBridged` -> `TokensBridged` to match the actual event name. The EOA withdrawal caveat below it is unchanged. 2de7cab

pxrl (approved):

PR #3703, opened 2026-08-11
fix(finalizer): bound Arbitrum withdrawal search at the confirmed send root

droplet-rl:
## Motivation

`arbStackFinalizer` bounded its event search at `now - challengePeriodSeconds`:

```ts
// Arbitrum orbit takes 7 days to finalize withdrawals, so don't look up events younger than that.
const latestBlockToFinalize = await getBlockForTimestamp(
  logger, chainId, getCurrentTime() - getArbitrumOrbitFinalizationTime(chainId), undefined, redis
);
```

That estimates *when* withdrawals ought to have become confirmed. The Outbox only settles a leaf beneath a confirmed send root, so the estimate is a proxy for the real gate rather than the gate itself.

**Measured on Arbitrum One, 2026-08-11 11:16Z:**

| bound | L2 block |
|---|---|
| `now - challengePeriodSeconds` (current) | 490,993,567 |
| Outbox latest `SendRootUpdated` (actual) | 491,188,161 |
| **gap** | **194,594 blocks ≈ 14.4h** |

Both bounds advance at chain rate, so this is systematic, not a one-off: every Arbitrum withdrawal was discovered ~14h after it became executable.

This was found with four SpokePool→HubPool withdrawals (28.29 WBTC) sitting confirmed-but-undiscovered, while the finalizer logged `Found 0 Arbitrum One messages (0 withdrawals | 0 deposits | 0 misc txns) for finalization` on every 30-minute run and `statusesGrouped: {EXECUTED: 6}`. Nothing was broken — the items were simply past the window's edge:

| outbox position | L2 block | blocks past bound |
|---|---|---|
| 164347 | 491,004,395 | ~11k |
| 164350 | 491,071,735 | ~78k |
| 164355 | 491,125,869 | ~132k |
| 164357 | 491,200,970 | ~207k |

## Change

Read the frontier from the Outbox. `SendRootUpdated(bytes32 outputRoot, bytes32 l2BlockHash)` fires on every confirmed assertion and carries the L2 block hash the root covers — exactly what `executeTransaction()` enforces.

- `getLatestConfirmedL2Block()` resolves that hash to a block number.
- The wall-clock estimate is kept as a **fallback** for when no confirmation is visible in the lookback (e.g. a halted Orbit chain — Aleph Zero's rollup has not confirmed since 2025-09-16).
- The `TokensBridged event filter` log line now records `bound: "outboxSendRoot" | "challengePeriodEstimate"`, so a silent regression to the estimate is visible in prod.
- Adds `SendRootUpdated` to `ArbitrumOutbox.json`, which previously carried no events.

## Why not just shrink the constant

Shaving the offset narrows the gap but re-guesses the same unknown, and it errs the wrong way. If assertions ever confirm *slower* than the configured period, a shorter offset pushes the bound **past** the frontier and the finalizer repeatedly attempts leaves that cannot settle. Reading the Outbox is correct in both directions.

## Scope and risk

- Applies to every Orbit chain served by `arbStack.ts`, not just Arbitrum One.
- Strictly widens the search window, so it cannot cause a withdrawal to be missed relative to today.
- The lookup is wrapped in try/catch and falls back to current behavior on any failure, so it cannot break finalization.
- One extra `eth_getLogs` (bounded to one challenge period of L1 blocks) plus one `eth_getBlockByHash` per finalizer run.

## Testing

`yarn lint` and `yarn typecheck` clean.

I did **not** add unit coverage — there is no existing `arbStack` test to extend, and exercising this needs a mocked Outbox plus L1/L2 providers. Flagging that rather than implying coverage. The mechanism was verified directly against mainnet: the Outbox's latest `SendRootUpdated` at L1 block 25,731,242 resolves to L2 block 491,188,161, which is the number the helper returns.

Per `AGENTS.md`: no `README.md` / `AGENTS.md` updates proposed — `src/finalizer/` has no module doc, and the change is an internal correctness fix with no config, interface, or runtime-flow change.

### DIFF
--- a/src/finalizer/utils/arbStack.ts
+++ b/src/finalizer/utils/arbStack.ts
@@ -28,6 +28,7 @@ import {
   isDefined,
   isEVMSpokePoolClient,
   EvmAddress,
+  Provider,
 } from "../../utils";
 import { getRedisCache } from "../../cache/Redis";
 import { TokensBridged } from "../../interfaces";
@@ -75,19 +76,31 @@ export async function arbStackFinalizer(
   const { chainId } = spokePoolClient;
   const networkName = getNetworkName(chainId);
 
-  // Arbitrum orbit takes 7 days to finalize withdrawals, so don't look up events younger than that.
+  // A withdrawal is only executable once the Outbox holds a confirmed send root covering it, so bound the
+  // event search at that block rather than at a wall-clock estimate of when confirmation ought to have
+  // happened. See getLatestConfirmedL2Block() for why the estimate is not good enough.
   const redis = await getRedisCache(logger);
-  const latestBlockToFinalize = await getBlockForTimestamp(
+  const latestConfirmedBlock = await getLatestConfirmedL2Block(
     logger,
     chainId,
-    getCurrentTime() - getArbitrumOrbitFinalizationTime(chainId),
-    undefined,
-    redis
+    hubPoolProvider,
+    spokePoolClient.spokePool.provider
   );
+  const latestBlockToFinalize =
+    latestConfirmedBlock ??
+    (await getBlockForTimestamp(
+      logger,
+      chainId,
+      getCurrentTime() - getArbitrumOrbitFinalizationTime(chainId),
+      undefined,
+      redis
+    ));
   logger.debug({
     at: `Finalizer#${networkName}Finalizer`,
     message: `${networkName} TokensBridged event filter`,
     to: latestBlockToFinalize,
+    // Surfaced so a regression back to the estimate is visible in the logs rather than silent.
+    bound: isDefined(latestConfirmedBlock) ? "outboxSendRoot" : "challengePeriodEstimate",
   });
   const withdrawalEvents: TokensBridged[] = [];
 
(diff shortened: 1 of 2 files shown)


src/finalizer/utils/arbStack.ts
```diff
@@ -192,6 +205,57 @@ export async function arbStackFinalizer(
...
+  try {
+    const { address, abi } = getContractEntry(CHAIN_IDs.MAINNET, `orbitOutbox_${chainId}`);
+    const outbox = new Contract(address, abi, l1Provider);
+
+    // One challenge period of L1 blocks contains at least one confirmation on a live chain.
+    const lookbackBlocks = Math.ceil(getArbitrumOrbitFinalizationTime(chainId) / MAINNET_BLOCK_TIME);
+    const sendRootEvents = await paginatedEventQuery(outbox, outbox.filters.SendRootUpdated(), {
+      from: Math.max(LATEST_MAINNET_BLOCK - lookbackBlocks, 0),
+      to: LATEST_MAINNET_BLOCK,
+    });
```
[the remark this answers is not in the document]
droplet-rl: Good catch, fixed in aa6b453.

Confirmed the mechanism: `getPaginatedBlockRanges` short-circuits to `[[from, to]]` when `maxLookBack` is `undefined`, so the whole range went out as one `eth_getLogs`. For Arbitrum One the lookback is `604800 / ~12s` ≈ 50k L1 blocks, 5x the 10k default in `CHAIN_MAX_BLOCK_LOOKBACK`.

The consequence is worse than a slow query, which is why this is worth fixing rather than leaving: the rejection would be swallowed by the `catch` and the function would return `undefined`, so the caller falls back to the challenge-period estimate — silently reinstating the exact ~14h discovery lag this PR removes, with the `bound: "challengePeriodEstimate"` log line as the only signal.

Now passes `maxLookBack: CHAIN_MAX_BLOCK_LOOKBACK[CHAIN_IDs.MAINNET]`, matching `opStack.ts:875` and `linea/l2ToL1.ts:197`. Ordering is unaffected — `getPaginatedBlockRanges` returns ranges ascending and `paginatedEventQuery` flattens them in order, so `.at(-1)` is still the newest root.

src/finalizer/utils/arbStack.ts
```diff
@@ -192,6 +205,59 @@ export async function arbStackFinalizer(
...
+ */
+async function getLatestConfirmedL2Block(
+  logger: winston.Logger,
+  chainId: number,
+  l1Provider: Provider,
+  l2Provider: Provider
+): Promise<number | undefined> {
+  const at = `Finalizer#${getNetworkName(chainId)}Finalizer`;
+  try {
+    const { address, abi } = getContractEntry(CHAIN_IDs.MAINNET, `orbitOutbox_${chainId}`);
```
pxrl, line 235: This is fragile because it assumes mainnet. Can we derive the L1 chainId instead?

src/finalizer/utils/arbStack.ts
```diff
@@ -192,6 +205,59 @@ export async function arbStackFinalizer(
...
+    const outbox = new Contract(address, abi, l1Provider);
+
+    // One challenge period of L1 blocks contains at least one confirmation on a live chain. That is ~50k blocks
+    // for a 7-day period, so the range has to be paginated to stay under provider eth_getLogs limits.
+    const lookbackBlocks = Math.ceil(getArbitrumOrbitFinalizationTime(chainId) / MAINNET_BLOCK_TIME);
```
pxrl: Is `MAINNET_BLOCK_TIME` able to be derived instead of assumed?

droplet-rl: Good catch — it isn't. Fixed in 49e6158.

`MAINNET_BLOCK_TIME` comes from `arch.evm.averageBlockTime()`, which *looks* derived but returns a hardcoded seed here. The SDK seeds its cache at module load and returns it while the TTL holds:

```ts
const now = getCurrentTime();                                        // module load
const blockTimes = { [CHAIN_IDs.MAINNET]: { average: 12.5, timestamp: now, ... } };
const cacheTTL = 60 * 15;
// averageBlockTime():
if (isDefined(cache) && now < cache.timestamp + cacheTTL) return { average: cache.average, ... };
```

A finalizer run that starts and finishes inside 15 minutes of process start never touches the RPC — it gets 12.5s.

It's also the wrong constant, in the unsafe direction. Mainnet runs ~12s, so `604800 / 12.5` = ~48.4k blocks ≈ **6.7 days**, not the 7 the "one challenge period contains at least one confirmation" invariant depends on. The shortfall is silent: a chain confirming near the period boundary falls out of the window, the function returns `undefined`, and the caller drops back to the challenge-period estimate this PR exists to replace.

Now resolved from real block timestamps:

```ts
const fromBlock = await getBlockForTimestamp(
  logger, CHAIN_IDs.MAINNET, getCurrentTime() - getArbitrumOrbitFinalizationTime(chainId), undefined, redis
);
```

`getBlockForTimestamp` binary-searches actual timestamps and is redis-cached across runs, so the window is a true challenge period regardless of block-time drift — and it's the same helper the fallback path already uses. Costs one extra lookup, cached.

Left `MAINNET_BLOCK_TIME` alone for its other two callers (`confirmPeriodBlocks` at L69 and the finalization-ETA log at L433). Both are estimates where a seeded 12.5s is defensible; happy to revisit separately if you'd rather it were measured everywhere.

`yarn build` + eslint + prettier clean.

PR #3706, opened 2026-08-11
fix(finalizer): tolerate unrecognised tokens in Arbitrum withdrawal discovery

droplet-rl:
## Problem

`arbStackFinalizer` discovers withdrawals by querying the canonical Arbitrum gateways, then resolves each event's L1 token to an L2 address before it can build any finalization:

```ts
const l2Token = getL2TokenAddresses(e.args.l1Token)?.[chainId];
assert(isDefined(l2Token), `Missing L2 token mapping ...`);
```

`getL2TokenAddresses` returns `undefined` for any token absent from `TOKEN_SYMBOLS_MAP`, so the assert throws inside the `.map()` — before `multicallArbitrumFinalizations` is reached. The finalizer returns nothing, and `finalize()` catches it per-chain, so *every* Arbitrum → Ethereum withdrawal in that run is skipped rather than just the one unrecognised token. Because the input is recomputed each loop, the condition persists until the event ages out of the lookback window.

The `try/catch` a few lines below, whose `catch` logs `"Skipping ERC20 withdrawal event for unknown token"`, was intended to cover this — but it wraps only object construction and an `Array.push`, neither of which can throw, so that path was unreachable.

## Why finalize rather than skip

The token is not an input to the finalization. `finalizeArbitrum` builds the Outbox `executeTransaction` calldata entirely from the Arbitrum SDK message (`getOutboxProof()` plus `nitroWriter.event`); `getUniqueLogIndex` keys on `txnRef`. The only real consumer is `getTokenInfo` in `multicallArbitrumFinalizations`, which supplies `symbol`/`decimals` for the log line.

So an unrecognised token is a labelling gap, not a reason to withhold a withdrawal — and skipping would drop a legitimate withdrawal whenever a real token is missing from `constants` (removed or renamed while a withdrawal sits mid-challenge-period). These withdrawals are already destined for an address we finalize for; the gateway queries filter on `to ∈ {HubPool, SpokePool, AtomicDepositor, userAddresses}`.

## Changes

- Discovery no longer treats token resolution as a precondition. An unmapped token falls back to the L1 address so the withdrawal stays identifiable in logs.
- One aggregated `logger.warn` per run listing the unrecognised withdrawals (`l1Token`, `to`, `amount`, `txnRef`), so these stay visible.
- `getTokenInfo` at the log-line site is wrapped in `describeToken()`, degrading to `UNKNOWN`/18. This was a second throw site — fixing only the assert would have relocated the failure, not removed it.
- Removed the dead `try/catch`.

`zkSync.ts` already carries the equivalent guard ("a single unknown token would otherwise abort finalization for every withdrawal on this chain"); `arbStack` was the gap. Polygon and Linea source their events from the SpokePool and sender-filtered queries respectively, so neither is affected.

## Testing

`yarn build`, `eslint` and `prettier` pass. No automated regression test — exercising the discovery path needs Arbitrum SDK receipts, Redis and block-time mocking. Happy to add one by extracting the event → `TokensBridged` mapping into an exported pure helper (the way `buildFinalizationBatches` is exported for its test) if reviewers want it.

## Note for reviewers

The warn is deliberately **not** routed to `notificationPath: "across-error"`. Reasoning in the code comment; easy to flip if you disagree.

## Credit

Reported by Jagdish, who also offered to help test the fix. Thanks!

### DIFF
--- a/src/finalizer/utils/arbStack.ts
+++ b/src/finalizer/utils/arbStack.ts
@@ -158,6 +158,8 @@ export async function arbStackFinalizer(
         unknownTokenWithdrawals.push({
           l1Token: e.args.l1Token,
           to: e.args._to,
+          // @dev Stringified deliberately: the logger stringifies BigNumbers itself, but does so via
+          // Object.fromEntries(), which would collapse this array into an object keyed "0", "1", ...
           amount: e.args._amount.toString(),
           txnRef: e.transactionHash,
         });
--- a/src/finalizer/utils/arbStack.ts
+++ b/src/finalizer/utils/arbStack.ts
@@ -27,6 +27,7 @@ import {
   assert,
   isDefined,
   isEVMSpokePoolClient,
+  Address,
   EvmAddress,
 } from "../../utils";
 import { getRedisCache } from "../../cache/Redis";
@@ -146,14 +147,27 @@ export async function arbStackFinalizer(
       to: latestBlockToFinalize,
     }
   );
+  // @dev Anyone can withdraw an arbitrary token to an address we finalize for, so an unrecognised token must not
+  // stall the queue. The token is only used to label the finalization; it is not an input to the Outbox proof.
+  const unknownTokenWithdrawals: { l1Token: string; to: string; amount: string; txnRef: string }[] = [];
   const _withdrawalEvents = [
     ...withdrawalErc20Events.map((e) => {
+      const l1Token = EvmAddress.from(e.args.l1Token);
       const l2Token = getL2TokenAddresses(e.args.l1Token)?.[chainId];
-      assert(isDefined(l2Token), `Missing L2 token mapping for L1 token ${e.args.l1Token} on chain ${chainId}`);
+      if (!isDefined(l2Token)) {
+        unknownTokenWithdrawals.push({
+          l1Token: e.args.l1Token,
+          to: e.args._to,
+          amount: e.args._amount.toString(),
+          txnRef: e.transactionHash,
+        });
+      }
       return {
         ...e,
         amount: e.args._amount,
-        l2TokenAddress: EvmAddress.from(l2Token),
+        // @dev An unmapped token has no known L2 address, so fall back to the L1 address to keep the withdrawal
+        // identifiable in logs. Nothing downstream reads this to build the finalization.
+        l2TokenAddress: isDefined(l2Token) ? EvmAddress.from(l2Token) : l1Token,
       };
     }),
     ...withdrawalNativeEvents.map((e) => {
@@ -166,30 +180,45 @@ export async function arbStackFinalizer(
       };
     }),
   ];
+  // @dev Deliberately not routed to across-error: anyone can mint this condition on demand, so paging on it would
+  // hand out a notification-spam primitive. It stays a warn so log-based alerting can still key off it if wanted.
+  if (unknownTokenWithdrawals.length > 0) {
+    logger.warn({
+      at: `Finalizer#${networkName}Finalizer`,
+      message: `Finalizing ${unknownTokenWithdrawals.length} ${networkName} withdrawal(s) of unrecognised token(s) 🚩`,
+      unknownTokenWithdrawals,
+    });
+  }
+
   // If there are any found withdrawal initiated events, then add them to the list of TokenBridged events we'll
   // submit proofs and finalizations for.
   _withdrawalEvents.forEach(({ transactionHash, transactionIndex, ...event }) => {
-    try {
-      const tokenBridgedEvent: TokensBridged = {
-        ...event,
-        amountToReturn: event.amount,
-        chainId,
-        leafId: 0,
-        l2TokenAddress: event.l2TokenAddress,
-        txnRef: transactionHash,
-        txnIndex: transactionIndex,
-      };
-      withdrawalEvents.push(tokenBridgedEvent);
-    } catch {
-      logger.debug({
-        at: `Finalizer#${networkName}Finalizer`,
-        message: `Skipping ERC20 withdrawal event for unknown token ${event.l2TokenAddress} on chain ${networkName}`,
-        event: event,
-      });
-    }
+    const tokenBridgedEvent: TokensBridged = {
+      ...event,
+      amountToReturn: event.amount,
+      chainId,
+      leafId: 0,
+      l2TokenAddress: event.l2TokenAddress,
+      txnRef: transactionHash,
+      txnIndex: transactionIndex,
+    };
+    withdrawalEvents.push(tokenBridgedEvent);
   });
 
   return await multicallArbitrumFinalizations(withdrawalEvents, signer, hubPoolClient, logger, chainId);
+}
+
+/**
+ * Resolve token metadata for the finalization log line. The Outbox proof is built from the L2 message alone, so a
+ * token we can't describe must degrade the label rather than abort the batch. The fallback decimals only keep the
+ * logged amount readable; treat it as indicative for an unrecognised token.
+ */
+function describeToken(l2TokenAddress: Address, chainId: number): { symbol: string; decimals: number } {
+  try {
+    return getTokenInfo(l2TokenAddress, chainId);
+  } catch {
+    return { symbol: "UNKNOWN", decimals: 18 };
+  }
 }
 
 async function multicallArbitrumFinalizations(
@@ -202,7 +231,7 @@ async function multicallArbitrumFinalizations(
   const finalizableMessages = await getFinalizableMessages(logger, tokensBridged, hubSigner, chainId);
   const callData = await Promise.all(finalizableMessages.map((message) => finalizeArbitrum(message.message, chainId)));
   const crossChainTransfers = finalizableMessages.map(({ info: { l2TokenAddress, amountToReturn } }) => {
-    const { symbol, decimals } = getTokenInfo(l2TokenAddress, chainId);
+    const { symbol, decimals } = describeToken(l2TokenAddress, chainId);
     const amountFromWei = convertFromWei(amountToReturn.toString(), decimals);
     const withdrawal: CrossChainMessage = {
       originationChainId: chainId,


src/finalizer/utils/arbStack.ts
```diff
@@ -146,14 +147,27 @@ export async function arbStackFinalizer(
...
   const _withdrawalEvents = [
     ...withdrawalErc20Events.map((e) => {
+      const l1Token = EvmAddress.from(e.args.l1Token);
       const l2Token = getL2TokenAddresses(e.args.l1Token)?.[chainId];
-      assert(isDefined(l2Token), `Missing L2 token mapping for L1 token ${e.args.l1Token} on chain ${chainId}`);
+      if (!isDefined(l2Token)) {
+        unknownTokenWithdrawals.push({
+          l1Token: e.args.l1Token,
+          to: e.args._to,
+          amount: e.args._amount.toString(),
```
pxrl: Is `e.args._amount` known to be a BigNumber? If so, we don't need the .toString() conversion.

src/finalizer/utils/arbStack.ts
```diff
@@ -146,14 +147,27 @@ export async function arbStackFinalizer(
       to: latestBlockToFinalize,
     }
   );
+  // @dev Anyone can withdraw an arbitrary token to an address we finalize for, so an unrecognised token must not
+  // stall the queue. The token is only used to label the finalization; it is not an input to the Outbox proof.
+  const unknownTokenWithdrawals: { l1Token: string; to: string; amount: string; txnRef: string }[] = [];
   const _withdrawalEvents = [
     ...withdrawalErc20Events.map((e) => {
+      const l1Token = EvmAddress.from(e.args.l1Token);
       const l2Token = getL2TokenAddresses(e.args.l1Token)?.[chainId];
```
pxrl: Are we able to query the l2Token address via an l1 or l2 contract, rather than looking it up locally? Can we alternatively derive the l2 address statically/deterministically without any queries?

src/finalizer/utils/arbStack.ts
```diff
@@ -166,32 +180,47 @@ export async function arbStackFinalizer(
...
-        at: `Finalizer#${networkName}Finalizer`,
-        message: `Skipping ERC20 withdrawal event for unknown token ${event.l2TokenAddress} on chain ${networkName}`,
-        event: event,
-      });
-    }
+    const tokenBridgedEvent: TokensBridged = {
+      ...event,
+      amountToReturn: event.amount,
+      chainId,
+      leafId: 0,
```
pxrl: Why do we assume leafId 0 here? What's its relevance?

src/finalizer/utils/arbStack.ts
droplet-rl: Yes, it's a BigNumber (`_amount` is `uint256` in the gateway ABI, and we pass the same value straight through as `amount: e.args._amount` two lines down).

The `.toString()` isn't a type conversion though, it's for log legibility: ethers v5 `BigNumber.toJSON()` emits `{"type":"BigNumber","hex":"0x…"}`, and `src/utils/extensions.ts` only patches `BigInt.prototype.toJSON`, not ethers' BigNumber. `zkSync.ts:253` stringifies `amountToReturn` in its equivalent per-withdrawal log payload for the same reason. (`convertFromWei` isn't an option here — an unrecognised token has no known decimals.)

Happy to drop it if you'd rather have the object in the log.

pxrl: In my experience these log fine without the .toString(), but there might be something here that causes a different codepath to run. Try to verify either way if you can.

droplet-rl: Verified, and you were right — my earlier explanation was wrong. `@risk-labs/logger` does stringify BigNumbers itself (`Formatters.js`, `iterativelyReplaceBigNumbers`), so the amount logs fine either way. Sorry for the confident bad answer.

But there *is* a different codepath, and it bites. That helper rebuilds the object with `Object.fromEntries(...)`, and only when something actually changed:

```js
const copyNeeded = replacements.some(([key, value]) => obj[key] !== value);
return copyNeeded ? Object.fromEntries(replacements) : obj;
```

So an array containing a BigNumber gets round-tripped through `Object.entries`/`Object.fromEntries` and comes back as a plain object. An array with no BigNumber hits the `copyNeeded === false` path and is returned untouched. Ran both against the real `Logger`:

```
// amount: bn                      // amount: bn.toString()
"unknownTokenWithdrawals": {       "unknownTokenWithdrawals": [
  "0": {                             {
    "amount": "1000000000000000000"     "amount": "1000000000000000000"
  }                                  }
}                                  ]
```

Amount renders identically; the array degrades to `{"0": ...}`. Reproduced with a bare `[bn, bn]` too, and a top-level (non-array) BigNumber logs perfectly — which matches your experience, since that case never loses anything.

Keeping the `.toString()` so the payload stays an array, and added a comment recording why (0f6d9a9) so it does not get tidied away later. Happy to drop it if you would rather take the object form — the amount itself is fine either way.

src/finalizer/utils/arbStack.ts
droplet-rl: `leafId` belongs to the SpokePool event: `TokensBridged` is emitted at the end of `_executeRelayerRefundLeaf`, and `leafId` identifies the relayer refund leaf that caused the bridge-back. arbStack never sees that event — it discovers withdrawals from the gateway's `WithdrawalInitiated` and `ArbSys.L2ToL1Tx` — so there is no leaf, and the `0` was filler to satisfy the type. It predates this PR (#1909); the diff only de-indented it out of the dead `try/catch`.

Nothing read it, here or anywhere: repo-wide there is no read of `.leafId` off a `TokensBridged` (the dataworker hits are `RelayerRefundLeaf.leafId`, a different field).

Fixed in c5b3ac8 by typing the path to the subset it actually uses, per your suggestion:

```ts
type ArbitrumWithdrawal = Pick<TokensBridged, "amountToReturn" | "l2TokenAddress" | "txnRef">;
```

`txnRef` is the only load-bearing field — it fetches the receipt the Outbox proof is rebuilt from, and keys `getUniqueLogIndex`; the other two just label the log line. Excess-property checking still fires on explicitly-written props next to a spread, so the fictional `leafId` and the unread `chainId`/`txnIndex` became compile errors rather than something that could quietly return. The push-into-a-preallocated-array became a `map`, since the callback is now a pure projection.

Two notes:

- arbStack is the only finalizer where this is possible. opStack, zkSync and linea all `.concat()` their synthetic events onto `spokePoolClient.getTokensBridged()`, so they're pinned to the real type and keep their own `leafId: 0`. arbStack never calls it — every event on that path is discovered, so nothing forces the full shape.
- One small loss: `getMessageOutboxStatusAndProof` logs `event` wholesale, and that payload was previously the entire raw ethers log (`args`, `topics`, `blockNumber`). It's now the three fields. `txnRef` still identifies the transaction, and `logIndex` is logged separately alongside it, so I think that's fine — say the word if you'd rather keep `blockNumber`.

Two pre-existing inaccuracies I left alone to keep the diff tight, both now more visible: the debug at line 96 calls this a "TokensBridged event filter", and the warn at line 357 says "Transaction that emitted TokensBridged event…" — neither is a TokensBridged event on this path. The second one pages via `across-error`, so it's the one worth rewording; happy to do it here or separately, but log-message edits can break log-based alerting so I didn't assume.

`yarn build`, `eslint` and `prettier` pass.

src/finalizer/utils/arbStack.ts
```diff
@@ -164,43 +187,52 @@ export async function arbStackFinalizer(
...
+  if (unknownTokenWithdrawals.length > 0) {
+    logger.warn({
+      at: `Finalizer#${networkName}Finalizer`,
+      message: `Finalizing ${unknownTokenWithdrawals.length} ${networkName} withdrawal(s) of unrecognised token(s) 🚩`,
+      unknownTokenWithdrawals,
+    });
+  }
+
+  // Reduce the discovered events to the fields the finalization path actually reads.
+  const withdrawalEvents: ArbitrumWithdrawal[] = _withdrawalEvents.map(
```
pxrl: Is this extra type annotation required?

droplet-rl: No, removed in 1ff6c2e. `withdrawalEvents` is passed straight to `multicallArbitrumFinalizations(withdrawals: ArbitrumWithdrawal[], ...)` on the next line, so the parameter type already pins the shape.

Checked the call site still catches drift — renaming `txnRef` to `txnReference` in the mapped literal fails without the annotation:

```
error TS2345: Argument of type "{ amountToReturn: any; l2TokenAddress: EvmAddress; txnReference: string; }[]"
is not assignable to parameter of type "ArbitrumWithdrawal[]".
```

Only real difference: the local infers `amountToReturn` as `any` rather than `BigNumber`, since ethers' `Result` args are `any`. That does not escape the function — the callee receives `ArbitrumWithdrawal[]`, so it is a `BigNumber` everywhere it is actually read. Bonus, the map now fits without the extra wrapper.

src/finalizer/utils/arbStack.ts
```diff
@@ -144,14 +152,29 @@ export async function arbStackFinalizer(
...
-      assert(isDefined(l2Token), `Missing L2 token mapping for L1 token ${e.args.l1Token} on chain ${chainId}`);
+      if (!isDefined(l2Token)) {
+        unknownTokenWithdrawals.push({
+          l1Token: e.args.l1Token,
+          to: e.args._to,
+          // @dev Stringified deliberately: the logger stringifies BigNumbers itself, but does so via
+          // Object.fromEntries(), which would collapse this array into an object keyed "0", "1", ...
+          amount: e.args._amount.toString(),
+          txnRef: e.transactionHash,
+        });
```
pxrl, line 171:
```suggestion
        const { l1Token, _to: to, _amount: amount } = e.args;
        unknownTokenWithdrawals.push({
          l1Token,
          to,
          // @dev Stringified deliberately: the logger stringifies BigNumbers itself, but does so via
          // Object.fromEntries(), which would collapse this array into an object keyed "0", "1", ...
          amount: amount.toString(),
          txnRef: e.transactionHash,
        });
```

src/finalizer/utils/arbStack.ts
```diff
@@ -144,14 +152,29 @@ export async function arbStackFinalizer(
...
     ...withdrawalErc20Events.map((e) => {
+      const l1Token = EvmAddress.from(e.args.l1Token);
       const l2Token = getL2TokenAddresses(e.args.l1Token)?.[chainId];
-      assert(isDefined(l2Token), `Missing L2 token mapping for L1 token ${e.args.l1Token} on chain ${chainId}`);
+      if (!isDefined(l2Token)) {
+        unknownTokenWithdrawals.push({
+          l1Token: e.args.l1Token,
+          to: e.args._to,
+          // @dev Stringified deliberately: the logger stringifies BigNumbers itself, but does so via
+          // Object.fromEntries(), which would collapse this array into an object keyed "0", "1", ...
```
pxrl: Comment too noisy - shrink it to 1 line.

src/finalizer/utils/arbStack.ts
droplet-rl: Applied in 248d608, with one adjustment: taken as written the destructure sits inside the `if` block and shadows the outer `const l1Token = EvmAddress.from(e.args.l1Token)` with the raw string, so `l1Token` would mean two different types either side of the brace. No `no-shadow` rule to catch it either.

Hoisted it to the top of the callback instead, which avoids the shadow and lets the `EvmAddress` binding go entirely:

```ts
const { l1Token, _to: to, _amount: amount } = e.args;
const l2Token = getL2TokenAddresses(l1Token)?.[chainId];
...
l2TokenAddress: EvmAddress.from(l2Token ?? l1Token),
```

`isDefined` is `input !== null && input !== undefined`, so `l2Token ?? l1Token` is exactly the old ternary, and the address is now only constructed once. tsc/eslint/prettier clean.

Also re-verified the `.toString()` note against the 1.3.14 logger that came in with your master merge — `iterativelyReplaceBigNumbers` still ends in `Object.fromEntries(replacements)`, and the array still collapses to `{"0": ...}` without it. Comment stays accurate.

src/finalizer/utils/arbStack.ts
```diff
@@ -202,7 +231,7 @@ async function multicallArbitrumFinalizations(
   const finalizableMessages = await getFinalizableMessages(logger, tokensBridged, hubSigner, chainId);
   const callData = await Promise.all(finalizableMessages.map((message) => finalizeArbitrum(message.message, chainId)));
   const crossChainTransfers = finalizableMessages.map(({ info: { l2TokenAddress, amountToReturn } }) => {
-    const { symbol, decimals } = getTokenInfo(l2TokenAddress, chainId);
+    const { symbol, decimals } = describeToken(l2TokenAddress, chainId);
```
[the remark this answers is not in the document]
droplet-rl: Confirmed — both halves reproduce. Fixed in aca68d4.

`getL2TokenAddresses(l1Token, l1ChainId = CHAIN_IDs.MAINNET)` returns the first `TOKEN_SYMBOLS_MAP` entry whose `addresses[l1ChainId]` matches, without checking that the entry covers the target chain. Ran both cases:

```
USDC sepolia -> ArbSepolia          old=undefined  new=0x75faf114...
WETH mainnet -> Robinhood(4663)     old=undefined  new=0x0Bd7D308...
USDC mainnet -> Arbitrum            old=new=0xaf88d065...   (unchanged)
WETH mainnet -> Arbitrum            old=new=0x82aF4944...   (unchanged)
unknown token                       old=new=undefined
```

The Sepolia one is live — `arbStackFinalizer` is wired up for `ARBITRUM_SEPOLIA` in the finalizer registry. The alias one is worse than it looks: `ETH` and `WETH` share `0xC02aaA39...` on mainnet and `ETH` sorts first in the map but has no chain-4663 entry, so *every* Robinhood WETH withdrawal was resolving to `undefined` and would have tripped the new unknown-token warn.

Switched to `getRemoteTokenForL1Token(l1Token, chainId, hubPoolClient.chainId)`, which takes the hub chain explicitly and requires `isDefined(addresses[remoteChainId])`, so the alias is skipped. Genuinely unrecognised tokens still return `undefined` and keep the fallback.

src/finalizer/utils/arbStack.ts
droplet-rl: Shrunk to one line in aca68d4:

```ts
// @dev Keeps this an array; the logger's BigNumber pass rebuilds via Object.fromEntries().
```

src/finalizer/utils/arbStack.ts
droplet-rl: Queryable, yes: the L2 gateway router exposes `calculateL2TokenAddress(l1Token)`, and `@arbitrum/sdk` wraps the L1 side as `Erc20Bridger.getChildErc20Address`. Two caveats:

- Our `erc20GatewayRouter` ABI only carries `outboundTransfer`, `getGateway` and `TransferRouted`, so it would need an ABI entry.
- Not statically derivable in general. The SDK queries rather than computes; standard-gateway tokens are CREATE2 and derivable in principle, but custom gateways hold an arbitrary l1→l2 mapping, so there is no offline answer that covers both.

It also would not buy much here: the address alone gives no symbol/decimals, which is all `describeToken` needs for the label, so we would still hit `fetchTokenInfo` — a couple of extra RPCs per token to decorate a log line that already degrades gracefully.

What this was really guarding against was the local lookup missing *recognised* tokens, and that turned out to be real — see the codex thread. Fixed properly in aca68d4 by resolving against the hub chain. Resolving this one on that basis; reopen if you would still like the on-chain query.

PR #3709, opened 2026-08-11
feat(deposit-address-service): execute v3 refund withdrawals

amateima:
Part of #3663 — PR 5 of the standalone deposit-address service. Stacked on #3701.

Executes the v3 refund withdrawal: the path for a `mis_route`, and for a `correct_transfer` the execute endpoint rejected as `AMOUNT_BELOW_MINIMUM`. Replaces PR 4's two placeholder throws at the lines they sat on, so the deposit lock is held across both actions via `processUnderLock`'s existing `try/finally`. `WithdrawRouteNotImplementedError` disappears with them, taking the last identifier carrying the removed `route` vocabulary.

**Review focus: one lock held across both actions; terminal 422 handling.**

## What is genuinely new

- `executeWithdraw` in `depositHandler.ts`, ported guard-for-guard from `initiateWithdrawV3` per the issue's parity matrix: the `ENABLE_V3_WITHDRAWALS` gate (same env var the polling bot reads, NACK while off), **EVM-only** namespaces (stricter than the deposit path — `assertSupportedNamespace` allows `tron` on TVM chains, so it could not be reused), the withdraw leaf materials check, and a smaller response assertion (`assertValidWithdrawResponse`: signed `chainId` against the **refund** chain `erc20Transfer.chainId`, and the now+60s signature deadline — `assertValidExecuteResponse` deliberately does not carry over).
- Terminal classification of a sign-withdraw failure **on the HTTP status alone**, exactly as production's `_getSignedWithdrawV3` does (`isHttpError(err) && err.status === 422`): persist `withdraw_failed`, ACK. Everything else NACKs. No client change; `_postOrThrow` discards the API's error code, so `withdraw_failed.code` becomes `optional(...)` and is unset — every existing PR 3 state test passes unchanged.
- `deductGasFromRefund: true`, deliberately unlike v1's full-amount refund — not to be unified in the v1 PR.

## Reused unchanged

The lock, both state reads, `assertSupportedOriginChain` (already runs on `erc20Transfer.chainId` before routing), canonicality-then-balance in that order and for the same reason, `onBroadcast` + `maxTries`, the pending-write retry, and `resolvePendingTransaction` (whose `operation: "withdraw"` mapping already existed). `broadcast()` is parameterised over `{operation, to, data, value, message, mrkdwn}` rather than duplicated. One adjustment there: a confirmed **withdraw** is not expected to carry the `MetadataEmitted` provenance event, so the missing-metadata warning is now gated to deposits.

## Not in this PR

`withdraw_executed` is recorded but **not published** — lifecycle publishing and its recovery are PR 6, which lands before PR 7 enables execution anywhere.

## Verification

- `yarn tsc --build --force` clean; `yarn lint` clean.
- 190 passing across the service suites + `TransactionClient` (163 on the base branch; +27 here).
- Every new guard and branch was red-checked by reintroducing the bug: the metadata gate, `deductGasFromRefund`, the signed-chainId check, the `operation` label, 422-only terminal classification, the code-gated below-minimum fallthrough, the lock held across both actions (the fake API records the lock token it observed inside each call), the withdraw gate, the leaf `kind` filter, EVM-only strictness, and canonicality-before-balance ordering — each failed exactly the expected tests, then passed on revert.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,7 +51,7 @@ The main bot types in `src/`:
 - `monitor`: Runs monitoring and reporting checks.
 - `gasless`: Handles gasless relay flows.
 - `deposit-address`: Polls the across-indexer for counterfactual deposit-address transfers and executes the resulting deposits or refund withdraws.
-- `deposit-address-service`: Standalone Express service doing the same work driven by GCP Pub/Sub push instead of polling. Not part of the `index.ts` CLI dispatch — it runs as its own entrypoint, like `cctp-finalizer`. Intended to replace `deposit-address`. Executes v3 deposits; v3 withdrawals and v1 are not ported yet, and both NACK so nothing is discarded.
+- `deposit-address-service`: Standalone Express service doing the same work driven by GCP Pub/Sub push instead of polling. Not part of the `index.ts` CLI dispatch — it runs as its own entrypoint, like `cctp-finalizer`. Intended to replace `deposit-address`. Executes v3 deposits and refund withdrawals; v1 is not ported yet and stays with the polling bot.
 
 ## Directory tree
 
--- a/src/deposit-address-service/README.md
+++ b/src/deposit-address-service/README.md
@@ -5,11 +5,11 @@ and executes them, replacing the polling bot in [`../deposit-address`](../deposi
 left untouched until cutover. Why a service rather than a poller:
 [#3663](https://github.com/across-protocol/relayer/issues/3663).
 
-> **v3 deposits execute; v3 withdrawals do not yet.** A `mis_route` transfer, and a `correct_transfer`
-> the execute endpoint rejects as below the minimum, both **NACK** until the withdraw path lands — they
-> are preserved, not discarded. With no handler configured, or `EXECUTION_ENABLED` unset, the service
-> NACKs every delivery and `/ready` stays `503`, so nothing is discarded if a subscription is attached
-> early. v1 messages stay with the polling bot.
+> **v3 deposits and refund withdrawals execute.** A withdrawal's `withdraw_executed` is recorded but not
+> yet announced — lifecycle publishing and its recovery land next, before execution is enabled anywhere.
+> With no handler configured, or `EXECUTION_ENABLED` unset, the service NACKs every delivery and `/ready`
+> stays `503`, so nothing is discarded if a subscription is attached early. v1 messages stay with the
+> polling bot.
 
 ## Contracts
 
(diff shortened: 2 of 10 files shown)


pxrl (dismissed):

pxrl:
@droplet-rl Review changes since last approval. No change?

droplet-rl:
Confirmed — no change. The commit patch is byte-identical (64,692 bytes) and the message unchanged across all three force-pushes since your approval at `3348163f`. The dismissal was stale-on-push, not new content.

Only tree delta vs the approved commit is `TransactionClient.ts`, pulled in from master by the rebase (#3766) — not this PR.

Worth flagging though: it's now CONFLICTING. The base branch gained `c06a8062` (the `AcrossApiHttpError` arg-order test fix) and this PR deletes that very test, superseding it with "falls through to the withdraw when the execute endpoint rejects the amount as below minimum". Resolution is take-this-PR's-side; I verified locally that the rebased tree is byte-identical to the current tip, typecheck clean, 79/79 deposit-address-service tests pass. Want me to push the rebase?

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=fb261422a0aac5c6d2a06912c470b56c&trace_start_ts=1788801794&trace_end_ts=1788888494)

pxrl (approved):

dookynme-bit:
Approved

AGENTS.md
```diff
@@ -52,7 +52,7 @@ The main bot types in `src/`:
 - `monitor`: Runs monitoring and reporting checks.
 - `gasless`: Handles gasless relay flows.
 - `deposit-address`: Polls the across-indexer for counterfactual deposit-address transfers and executes the resulting deposits or refund withdraws.
-- `deposit-address-service`: Standalone Express service doing the same work driven by GCP Pub/Sub push instead of polling. Not part of the `index.ts` CLI dispatch — it runs as its own entrypoint, like `cctp-finalizer`. Intended to replace `deposit-address`. Executes v3 deposits; v3 withdrawals and v1 are not ported yet, and both NACK so nothing is discarded.
+- `deposit-address-service`: Standalone Express service doing the same work driven by GCP Pub/Sub push instead of polling. Not part of the `index.ts` CLI dispatch — it runs as its own entrypoint, like `cctp-finalizer`. Intended to replace `deposit-address`. Executes v3 deposits and refund withdrawals; v1 is not ported yet and stays with the polling bot.
 
 ## Directory tree
```
dookynme-bit: Approved

